PR #47, opened 2021-12-13
Add contribution documentation

### LINKED ISSUE #39: Add docs on contributing to the performance plugin
This issue is about creating a `/docs/Getting-started.md` file with the most important information needed to contribute to the performance plugin. That documentation should follow a similar structure like Gutenberg's https://github.com/WordPress/gutenberg/blob/trunk/docs/contributors/code/getting-started-with-code-contribution.md, e.g.:
* explain how to clone the repository
* which local tools are required (Node.js, Docker)
* any commands that need to be run (e.g. linting, tests)
* high-level code requirements (e.g. follow WordPress coding standards and WordPress core coding patterns in general)

We should also include there with which WordPress and PHP versions the code should be compatible, which ties into a greater product question: **Which WordPress and PHP version should we require and how far back do we support?**

^ That is probably the main question to answer here before we can get started with the documentation.

In addition to the above file, which is the main part of this issue, we should also add the regular `CONTRIBUTING.md` file. That one can be really short and mostly link to the above documentation file. At a high level, we can look at Gutenberg's https://github.com/WordPress/gutenberg/blob/trunk/docs/contributors/code/README.md and https://github.com/WordPress/gutenberg/blob/trunk/CONTRIBUTING.md for inspiration what to include here.

The main part of this issue will be coming up with the above documentation in a PR and reviewing it.

JustinyAhin:
This PR adds the documentation to get start with contribution to the performance plugin. It also adds the new documentation page to the `/docs/README.md` file.

See https://github.com/WordPress/performance/blob/add/39-contributing-documentation/docs/Getting-started.md for the new doc.

Fixes #39.

### DIFF
--- a/docs/Getting-started.md
+++ b/docs/Getting-started.md
@@ -25,6 +25,11 @@ First, make sure Docker is installed, up and running on your machine. Then, run 
 The following command will stop the WordPress environment:
 `npm run wp-env stop`
 
+## Open an issue
+Opening an issue is the first step to contributing to the Performance plugin. This allows to discuss, review or validate an idea before implementation. Opening an issue is not strict requirement, but it is highly recommended in most cases, unless the change you are trying to make is really minor or is already covered by an existing issue.
+
+Issues should be labeled to facilitate browsing and filtering. Here are some common labels: `[Type] Feature`, `[Type] Discussion`, `Needs Decision`, `[Focus] Images`. The full list of labels can be found [here](https://github.com/WordPress/performance/labels).
+
 ## Create a pull request
 A pull request must be created to submit changes to the Performance plugin. Pull requests should refer to an issue in the [repository issue tracker](https://github.com/WordPress/performance/issues).
 
--- a/docs/Getting-started.md
+++ b/docs/Getting-started.md
@@ -10,11 +10,15 @@ This guide focuses on how to contribute to the Performance plugin, from setting 
 
 ### Clone the repository
 Start by cloning the repository into your local machine. To do this, run the following command in your terminal:
-`git clone https://github.com/WordPress/performance.git`
+```
+git clone https://github.com/WordPress/performance.git
+```
 
 ### Install the local development dependencies
 To install the local development dependencies, run the following command in your terminal:
-`npm install`
+```
+npm install
+```
 
 ### Start the local WordPress environment
 The Performance plugin uses Docker and [the wp-env](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env) to start a local WordPress environment.
@@ -23,7 +27,9 @@ First, make sure Docker is installed, up and running on your machine. Then, run 
 
 ### Stop the local WordPress environment
 The following command will stop the WordPress environment:
-`npm run wp-env stop`
+```
+npm run wp-env stop
+```
 
 ## Open an issue
 Opening an issue is the first step to contributing to the Performance plugin. This allows to discuss, review or validate an idea before implementation. It is not a strict requirement, but it is highly recommended in most cases, unless the change you are trying to make is really minor or is already covered by an existing issue.
@@ -51,7 +57,12 @@ npms run format-php
 ```
 
 ## PHPUnit tests
-To execute PHPUnit tests, run the `npm run test-php` in your terminal.
+To execute PHPUnit tests, run this in your terminal:
+```
+npm run test-php
+```
 
 And to execute the tests in multisite, run:
-`npm run test-php-multisite`
+```
+npm run test-php-multisite
+```
--- a/docs/Getting-started.md
+++ b/docs/Getting-started.md
@@ -1,9 +1,9 @@
 # Start contributing to the Performance plugin
-This guide is intended to give details on how to contribute to the Performance plugin. From setting up the development environment to creating pull requests.
+This guide focuses on how to contribute to the Performance plugin, from setting up the development environment to creating pull requests.
 
 ## Prerequisites
-- Node.js
-- Docker
+- [Node.js](https://nodejs.org)
+- [Docker](https://www.docker.com/products/docker-desktop)
 - [Git](https://git-scm.com)
 
 ## Setup the development environment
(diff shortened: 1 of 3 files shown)


docs/README.md
```diff
@@ -1,3 +1,4 @@
 # Documentation
 
+* [Getting Started](./getting-started.md)
```
felixarntz: Nit-pick: Since the other Markdown file uses "regular" capitalization, let's align this one and name it `Getting-started.md`?

docs/Getting-started.md
```diff
@@ -0,0 +1,52 @@
+# Start contributing to the Performance plugin
+This guide is intended to give details on how to contribute to the Performance plugin. From setting up the development environment to creating pull requests.
+
+## Prerequisites
+- Node.js
+- Docker
```
felixarntz: Can we put links to those two here as well?
* https://nodejs.org/
* https://www.docker.com/products/docker-desktop

docs/Getting-started.md
```diff
@@ -0,0 +1,52 @@
+# Start contributing to the Performance plugin
+This guide is intended to give details on how to contribute to the Performance plugin. From setting up the development environment to creating pull requests.
```
felixarntz: I think those need to be connected since the latter is not a full sentence by itself. Maybe like this?
```suggestion
This guide focuses on how to contribute to the Performance plugin, from setting up the development environment to creating pull requests.
```

docs/Getting-started.md
```diff
@@ -0,0 +1,52 @@
...
+### Clone the repository
+Start by cloning the repository into your local machine. To do this, run the following command in your terminal:
+`git clone https://github.com/WordPress/performance.git`
+
+### Install the local development dependencies
+To install the local development dependencies, run the following command in your terminal:
+`npm install`
+
+### Start the local WordPress environment
+The Performance plugin uses Docker and `[wp-env](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env) to start a local WordPress environment.
```
felixarntz: Either there's a missing ` character here or one too many? :)
```suggestion
The Performance plugin uses Docker and [the wp-env package](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env) to start a local WordPress environment.
```

docs/Getting-started.md
```diff
@@ -0,0 +1,52 @@
...
+## Prerequisites
+- Node.js
+- Docker
+- [Git](https://git-scm.com)
+
+## Setup the development environment
+
+### Clone the repository
+Start by cloning the repository into your local machine. To do this, run the following command in your terminal:
+`git clone https://github.com/WordPress/performance.git`
```
felixarntz: I think for a better reading flow, it would help to put all full commands in their own line, wrapping it with ``` instead of `. Comparing https://github.com/WordPress/gutenberg/blob/trunk/docs/contributors/code/getting-started-with-code-contribution.md with https://github.com/WordPress/performance/blob/add/39-contributing-documentation/docs/Getting-started.md, I think the Gutenberg doc is easier to follow with its commands broken out in individual lines.

If you agree, let's apply the change to this whole doc?

docs/Getting-started.md
```diff
@@ -0,0 +1,52 @@
...
+First, make sure Docker is installed, up and running on your machine. Then, run `npm run wp-env start` to start the WordPress environment. The WordPress development site will be available at `http://localhost:8080`.
+
+### Stop the local WordPress environment
+The following command will stop the WordPress environment:
+`npm run wp-env stop`
+
+## Create a pull request
+A pull request must be created to submit changes to the Performance plugin. Pull requests should refer to an issue in the [repository issue tracker](https://github.com/WordPress/performance).
+
+For better triaging, it is recommended that each pull request receive a `[Type] xyz`, `[Focus] xyz` or `[Infrastructure]` labels.
```
felixarntz: I think this needs to be clarified a bit: Every pull-request should receive both a `[Type] xyz` label _and_ either a `[Focus] xyz` or `Infrastructure` label.

The first is to classify what type of task this is, the second is which area it belongs to.

docs/Getting-started.md
```diff
@@ -0,0 +1,52 @@
...
+The Performance plugin uses Docker and `[wp-env](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env) to start a local WordPress environment.
+
+First, make sure Docker is installed, up and running on your machine. Then, run `npm run wp-env start` to start the WordPress environment. The WordPress development site will be available at `http://localhost:8080`.
+
+### Stop the local WordPress environment
+The following command will stop the WordPress environment:
+`npm run wp-env stop`
+
+## Create a pull request
+A pull request must be created to submit changes to the Performance plugin. Pull requests should refer to an issue in the [repository issue tracker](https://github.com/WordPress/performance).
```
felixarntz: Nit-pick, but maybe put the link for `https://github.com/WordPress/performance/issues` instead, since that's where the issues are?

Also I wonder if we should call this out a bit more explicitly, that it's recommended to open an issue to first discuss/review an idea _before_ opening a pull-request. It's not a strict requirement, but in most cases it makes sense, unless the PR is e.g. to fix a typo or something really minor.

I would go as far as saying maybe we should put another brief section about `Open an issue` above the `Create a pull request` section.

felixarntz (commented):
@JustinyAhin This looks great, I've added a few comments with minor suggestions, mostly around possible ways to improve the reading flow.

One more thing: Can you also add a `CONTRIBUTING.md` file in this PR? That is the standard place for people to look, so we should have this file and was also mentioned in the issue #39. The file can be super minimal and mostly just link to the other documentation file that you're adding here. Maybe take [Gutenberg's one](https://github.com/WordPress/gutenberg/blob/trunk/CONTRIBUTING.md) for inspiration, although I think ours could even be simpler than that.

docs/Getting-started.md
JustinyAhin: That's what I initially thought. But I wondered about documentation PRs for instance. Are they about `Infrastructure`?

docs/Getting-started.md
JustinyAhin: Yeah, it makes sense to have an `Open an issue` section.

docs/Getting-started.md
JustinyAhin: I'm not sure if I correctly understand this @felixarntz. Can you add a snippet to illustrate?

felixarntz: @JustinyAhin I mean like this:
```
git clone https://github.com/WordPress/performance.git
```

If you put ``` around code and split it between lines, it renders in its own code block rather than inline, which makes it easier to spot.

docs/Getting-started.md
felixarntz: @JustinyAhin Good question - I'd say yes, but for documentation PRs it wouldn't be as critical since they would likely not appear in changelogs. Still, I think it's a good pattern to always add those labels.

docs/Getting-started.md
JustinyAhin: Oh you are right, this is much cleaner.

felixarntz (approved):
🎉

Issue #40, opened 2021-12-03
Add plugin `readme.txt`


felixarntz:
Like every plugin, the performance plugin needs a `readme.txt`. We already have a plugin name (see #1) and a preliminary plugin description (https://github.com/WordPress/performance/blob/trunk/load.php#L5), and we are currently defining WP and PHP version support as part of #39.

What should our `readme.txt` contain? Any ideas on the detailed outline, what to put in the FAQ etc? Let's brainstorm.

dainemawer:
@felixarntz couple of ideas here to get this thread going:

1. First off should be a list of the improvement modules as well as why they improve performance (motivation)
2. Where we see this plugin headed (dreams/aspirations)
3. Any specific call outs (Gutenberg, Media Library etc)
4. Link to contributing guidelines
5. Perhaps a detailed section about what Web Vitals this plugin will have an impact on?
6. I think most importantly, what to do and who to reach out to if you (the user) doesn't see immediate impacts - could be a number of reasons - this is obviously very specific to each user, so maybe we should cover our bases here.
7. Links to CHANGELOG and Issues
8. Perhaps some direction on auditing / measuring metrics holistically? https://web.dev/vitals-tools-workflow/
9. FAQ's I think could be quite far reaching, here are a few ideas:

- How does this plugin effect caching?
- How well does it play with existing Media Library plugins (SmushIt / Regenerate Thumbnails)
- Can data be sent to Google Analytics / Google Tag Manager / Search Console integrations?
- Any Gutenberg specific functionality?
- Any effects on the server? DB queries / effects on Core / wp-uploads?

Happy to create a PR outlining this in coherent English!

felixarntz:
@dainemawer I opened a PR with an initial `readme.txt` draft. It also includes a script to facilitate maintenance of it, but specifically the readme content itself is merely a first draft, so please review and iterate on it as you please. That also applies to anyone else, we should collaborate on the content as much as possible. With that said, let's keep the first version somewhat simple, it should have everything that we think should be there, but it doesn't need to be perfect and document every detail etc. - we can always iterate in follow-up issues/PRs :)

Here's the initial proposal: https://github.com/WordPress/performance/pull/72/files#diff-9e6e4772050998a5c0dc3c61acf3dab0a7e594566171fa5746d6b62f9598efb6

dainemawer:
Sorry @felixarntz was out when this was merged - I'll take a quick look at the merged code and open up any further items in issues. Nice work!

PR #51, opened 2021-12-16
Add changelog generator script

### LINKED ISSUE #41: Define changelog approach / tooling
As the performance plugin will be released in the WordPress plugin repository at some point (and also in general 😄), we should think about how to create and maintain a changelog for new features, enhancements, and bug fixes.

Some high level questions:
* Should we maintain the changelog manually or through automated tooling (e.g. like https://github.com/WordPress/gutenberg/blob/trunk/bin/plugin/commands/changelog.js)? Manually is probably going to read more nicely, but it's also only maintainable to a limited degree. Potentially we could start with a manual approach, since it's hard to foresee how exactly an automated tooling approach should be structured for our purposes?
* Do we want to have a separate changelog file (e.g. `CHANGELOG.md` on GitHub), have the changelog only on the release tags on GitHub, or should the changelog be within the `readme.txt` file (visible on wordpress.org directly)?

The decision for the latter will also tie into #40 - the `readme.txt` file either needs to continue the changelog itself or a link to where it is at.

felixarntz:
Fixes #41 

This PR takes over the code from [Gutenberg's changelog generator script](https://github.com/WordPress/gutenberg/blob/trunk/bin/plugin/commands/changelog.js) to this repository, with a few adjustments for our purposes. Overall, our implementation can be a bit simpler than theirs since we don't have as many labels and complex requirements to our changelog - at least not yet 🙂

In order to test this PR, pull the branch and run the following command (make sure to update dependencies first):
```
npm run changelog -- -m "Initial plugin release"
```
You can provide any milestone title as argument, but the above is the only one we have so far (see https://github.com/WordPress/performance/milestones).

The above should give an output like the following (which can later be directly copied/pasted into the `readme.txt` file):
```
= Initial plugin release =

**Features**

* Images: Add WebP for uploads module. ([32](https://github.com/WordPress/performance/pull/32))
* Infrastructure: Add settings screen to toggle modules. ([30](https://github.com/WordPress/performance/pull/30))

**Documentation**

* Infrastructure: Add contribution documentation. ([47](https://github.com/WordPress/performance/pull/47))
* Infrastructure: Define module specification in documentation. ([26](https://github.com/WordPress/performance/pull/26))
```

Our milestones will eventually be named with actual versions, i.e. instead of `= Initial plugin release =` the output would start with something like `= 1.0.0 =`.

More in-depth overview of the changes:
* Add `bin/plugin` folder with `cli.js` entrypoint to the plugin's CLI framework.
    * Mostly copied from Gutenberg, with a few simplifications.
* Add `bin/plugin/commands/changelog.js` with the implementation of the changelog command. That is going to be the only command we need for now.
    * Mostly copied from Gutenberg, with a few simplifications.
    * Passing a `--milestone` or `-m` argument is required in our case.
    * The changelog is generated based on the specification in https://github.com/WordPress/performance/issues/41#issuecomment-987120029.
        * It gets all closed pull-requests for the given milestone.
        * It groups them by their `[Type] xyz` label.
        * It then uses the pull-request title and prefixes it with the focus (from `[Focus] xyz` label) or the word "Infrastructure" (from `Infrastructure` label).
        * The output format is different than the one from the Gutenberg script because the changelog we generate needs to be compatible with the wordpress.org `readme.txt` format (which Gutenberg is not using for its changelog, it uses regular Markdown instead).
* Add `bin/plugin/lib` folder with some common utility functions.
    * Pretty much an exact copy of the same Gutenberg files.
* Add `npm run changelog` script that calls the above changelog command via the new CLI tool.
* Add ESLint configuration via `@wordpress/scripts` package, with `format-js` and `lint-js` commands.
    * This is only added in a very simple way to have basic linting and formatting for the new JS files in the `bin` folder, which at this point are the only JS files in the entire project.
    * Any enhancements to the configuration and a GitHub action to enforce linting can be added in a follow-up issue.
* Add relevant packages for all the above as `devDependencies` to `package.json`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,14 @@
+/**
+ * External dependencies
+ */
+const wpConfig = require( '@wordpress/scripts/config/.eslintrc.js' );
+
+const config = {
+	...wpConfig,
+	rules: {
+		...( wpConfig?.rules || {} ),
+		'jsdoc/valid-types': 'off',
+	},
+};
+
+module.exports = config;
--- a/.gitignore
+++ b/.gitignore
@@ -44,7 +44,6 @@ vendor/
 ############
 
 tests/logs
-bin/
 tmp/
 *.tmp
 *.bak
--- a/bin/plugin/cli.js
+++ b/bin/plugin/cli.js
@@ -1,0 +1,39 @@
+#!/usr/bin/env node
+
+/**
+ * External dependencies
+ */
+const program = require( 'commander' );
+
+const withOptions = ( command, options ) => {
+	options.forEach( ( { description, argname } ) => {
+		command = command.option( argname, description );
+	} );
+	return command;
+};
+
+const catchException = ( handler ) => {
+	return async ( ...args ) => {
+		try {
+			await handler( ...args );
+		} catch ( error ) {
+			console.error( error ); // eslint-disable-line no-console
+			process.exitCode = 1;
+		}
+	};
+};
+
+/**
+ * Internal dependencies
+ */
+const {
+	handler: changelogHandler,
+	options: changelogOptions,
+} = require( './commands/changelog' );
+
+withOptions( program.command( 'release-plugin-changelog' ), changelogOptions )
+	.alias( 'changelog' )
+	.description( 'Generates a changelog from merged pull requests' )
+	.action( catchException( changelogHandler ) );
+
+program.parse( process.argv );
--- a/bin/plugin/commands/changelog.js
+++ b/bin/plugin/commands/changelog.js
@@ -1,0 +1,318 @@
+/**
+ * External dependencies
+ */
+const { groupBy } = require( 'lodash' );
+const Octokit = require( '@octokit/rest' );
+
+/**
+ * Internal dependencies
+ */
+const { log, formats } = require( '../lib/logger' );
+const {
+	getMilestoneByTitle,
+	getIssuesByMilestone,
+} = require( '../lib/milestone' );
+const config = require( '../config' );
+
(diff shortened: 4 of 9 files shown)


JustinyAhin (approved):
Implementation wise, this looks fine to me @felixarntz. Tested and it worked as expected. 

I was wondering, when the `npm run changelog -- -m "Initial plugin release"` command is run, would it make sense to paste the output in the clipboard, or even paste it in the README.txt file directly? We can anticipate that with time, the generated changelog will be bigger and bigger, and at least the first option will come handy.

tillkruss (approved):

felixarntz:
@JustinyAhin 
> I was wondering, when the `npm run changelog -- -m "Initial plugin release"` command is run, would it make sense to paste the output in the clipboard, or even paste it in the README.txt file directly? We can anticipate that with time, the generated changelog will be bigger and bigger, and at least the first option will come handy.

Great idea! I think we could eventually add this, but since we don't have the `readme.txt` yet, I think it would be helpful to do that separately once it's there. Maybe another command like `npm run update-changelog` or `npm run readme-changelog` that would rely on the above `npm run changelog` command and then inject the output into the `readme.txt`. That way we'd have both commands and they could be run individually, since sometimes it may also be useful to only see the "current" changelog output, e.g. to double-check.

Maybe worth a follow-up issue (which would depend on both #40 and #41 to be completed)?

JustinyAhin:
> Maybe worth a follow-up issue (which would depend on both #40 and #41 to be completed)?

Yes definitely. Let's merge this one first when it's ready, and move forward.

PR #53, opened 2021-12-16
 Add GitHub Action to validate PR labels and milestone

### LINKED ISSUE #52: Implement GitHub action to ensure necessary labels and milestone are present on pull requests
Based on https://github.com/WordPress/performance/issues/41#issuecomment-994746944, we should have a GitHub workflow that ensures that every pull request has the necessary labels and milestones. This can later be set up as a requirement that has to be met before a pull request can be merged.

Here's the specifications that the GitHub workflow should implemented (based on #41 as well):
* Every pull request must have one of the `[Type] xyz` labels. This is used to group pull requests in the changelog.
* Every pull request must have _either_ one of the `[Focus] xyz` labels _or_ the `Infrastructure` label. This is used to further categorize and prefix pull requests in the changelog.
* Every pull request must have _either_ a milestone set _or_ be marked with the `no milestone` label. This is used to define which release a pull request belongs to, or to explicitly define that a certain issue is "release-irrelevant" and therefore does not need a milestone.

JustinyAhin:
Fixes #52.

This PR adds a workflow to validate the labels and milestone added to a PR (see the mentioned issue description for the requirements).


JustinyAhin:
Right now, the workflow is not triggered when a milestone is added or removed on a PR. I haven't found a way to add this event to the action.
 Still digging for that.

JustinyAhin:
Some details about how the action work:
- Every time a PR is opened, reopened, or when a label is added or removed on a PR, the workflow is triggered
- If the PR contains none of the `[Type]` labels, the workflow fails with this error message https://github.com/WordPress/performance/blob/cb35ef0ca1bbda0972e2fc897cfe18b09158b446/.github/workflows/pr-validation.yml#L32
- If the PR contains none of the `[Focus]` or `[Infrastructure]` label, the workflow fails with https://github.com/WordPress/performance/blob/cb35ef0ca1bbda0972e2fc897cfe18b09158b446/.github/workflows/pr-validation.yml#L47
- If the PR does not have a milestone and is not labeled `no milestone` the workflow fails with https://github.com/WordPress/performance/blob/cb35ef0ca1bbda0972e2fc897cfe18b09158b446/.github/workflows/pr-validation.yml#L54


There is one scenario where the action will not work as we would want:

Let's say the PR has initially a milestone set and does not have the `no milestone` label. The workflow will success. When the milestone is removed, the workflow should normally fail. But it will not because GitHub does not associate milestone changes events with a PR. So we can not trigger the workflow on that.

I do think we can merge the PR as it is now, and figure out in another issue how to fix that.

.github/workflows/pr-validation.yml
```diff
@@ -0,0 +1,54 @@
...
+      - uses: actions/checkout@v2
+      - name: Check Type Labels
+        if: |
+          !contains( github.event.pull_request.labels.*.name, '[Type] Bug' ) &&
+          (
+            !contains( github.event.pull_request.labels.*.name, '[Type] Feature' ) &&
+            !contains( github.event.pull_request.labels.*.name, '[Type] Documentation' ) &&
+            !contains( github.event.pull_request.labels.*.name, '[Type] Discussion' ) &&
+            !contains( github.event.pull_request.labels.*.name, '[Type] Enhancement' )
+          )
```
felixarntz: Why are all but the first check here in parentheses? That shouldn't be necessary I'd think? It looks a bit confusing since it makes one think the first condition is in any way special, but it's similar to the ones below.

Anyway, I'm wondering whether we could simplify this. This code would be easier to maintain if we only checked whether there is any label that starts with `[Type] ...` - otherwise we would need to update the code every time we'd add a new type label. Maybe we can do a regex or some sort of `startsWith` matching here instead?

.github/workflows/pr-validation.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Check Focus or Infrastructure Labels
+        if: |
+          !contains( github.event.pull_request.labels.*.name, '[Focus] Images' ) &&
+          (
+            !contains( github.event.pull_request.labels.*.name, '[Focus] JavaScript' ) &&
+            !contains( github.event.pull_request.labels.*.name, '[Focus] Measurement' ) &&
+            !contains( github.event.pull_request.labels.*.name, '[Focus] Object caching' ) &&
+            !contains( github.event.pull_request.labels.*.name, '[Focus] Site Health' ) &&
+            !contains( github.event.pull_request.labels.*.name, 'Infrastructure' )
+          )
```
felixarntz: Similar to my comment above: I think this could be simplified and made easier to maintain by checking the following:
* the PR does not have any label starting with `[Focus] ...`
* _and_ the PR also does not have the `Infrastructure` label

That way we don't have to update this code whenever we add a new focus.

.github/workflows/pr-validation.yml
```diff
@@ -0,0 +1,54 @@
...
+on:
+  pull_request:
+    branches:
+      - trunk
+    types:
+      - "labeled"
+      - "unlabeled"
+      - "opened"
+      - "reopened"
+      - "synchronize"
```
felixarntz, line 12: Too bad indeed that there's no `milestoned` (or similar) activity type in https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request. I guess for now one would have to add/remove/add labels to re-trigger "manually".

felixarntz (commented):
Thanks @JustinyAhin! I left a few code comments, but other than those it looks good to me.

While it's unfortunate we can't trigger this when a milestone is added/removed, I agree it's okay to move forward without this, since there's no way to fix it for now, and also since at least it's not possible to get the workflow to pass without having never added a milestone or the "no milestone" label - which should at least raise enough awareness of these requirements.

The one other case where it would be problematic is that the workflow would fail if you first remove the "no milestone" label and then add a milestone. But at least you can always retrigger it e.g. by adding a label and removing it again.

JustinyAhin:
> The one other case where it would be problematic is that the workflow would fail if you first remove the "no milestone" label and then add a milestone. But at least you can always retrigger it e.g. by adding a label and removing it again.

Yes, that one is also quite tricky

.github/workflows/pr-validation.yml
JustinyAhin: > Why are all but the first check here in parentheses?

This seems to be how the `.yml` syntax works in GitHub Actions. The other option would have been to have all the expressions on the same line (see https://github.community/t/how-to-write-multi-line-condition-in-if/128477/2).

> Maybe we can do a regex or some sort of startsWith matching here instead?

That's the first option I actually explored 😅. Yes, I agree it'll definitely make the code more maintainable. I'll do another tries with https://github.com/marketplace/actions/actions-ecosystem-action-regex-match.

eugene-manuilov: Maybe we should use something like this? https://github.com/marketplace/actions/label-checker-for-pull-requests

JustinyAhin:
@eugene-manuilov you can start by looking at what I've done here https://github.com/WordPress/performance/blob/add/52-github-actions-pr-validation/.github/workflows/pr-validation.yml.

Pretty much everything is implemented, with the use of regex.

Every time a PR is opened, or a label is added or removed, the workflow is run. It checks the presence of Type, Focus and Infrastructure labels. It also checks if a milestone is attached to the PR.

The only things not 100% ok yet is that the workflow will not run on changes on the PR milestone, as GitHub does not have a PR event for that.

I think we can still merge this as that for now though.

eugene-manuilov:
@JustinyAhin, oups... I have already created a new PR #85 using your template as an example. Hope that's fine... 🤔

JustinyAhin:
> @JustinyAhin, oups... I have already created a new PR #85 using your template as an example. Hope that's fine... 🤔

It's okay @eugene-manuilov :). Giving a look to your PR. And I'll add feedbacks on it.

JustinyAhin:
Closing this PR in favor of https://github.com/WordPress/performance/pull/85 which has a much cleaner implementation.

Issue #25, opened 2021-12-01
Create Site Health Audit Enqueued Assets module


ThierryA:
Similar to #4 and #5, would be great to include to work started by @audrasjb [here](https://github.com/audrasjb/site-health-audit-enqueued-assets/blob/master/site-health-audit-enqueued-assets.php) as module 😄

manuelRod:
Shouldn't we define messaging? Maybe tailor it for end-users?
Also, where this info is gonna show in the Site Health Screen? 
Maybe all modules should be added under a custom tab?

ThierryA:
> Shouldn't we define messaging? Maybe tailor it for end-users?

+1 to that

> Also, where this info is gonna show in the Site Health Screen?

Stacked with all other items and with a "Performance" badge, although that is not set in stone I guess.

Meta note, raising awareness of too much CSS/JS etc is a good start, doing so and providing the source of the "offenders" is one step more helpful, doing so AND providing an action to solve the issue is the pinnacle. The later is extremely hard without risking to break something indeed.

cc @westonruter who has a lot of experience with sourcing offenders 😉

westonruter:
Yeah, identifying the theme/plugin responsible for enqueuing a script or stylesheet is something the AMP plugin does heavily. It also goes to the extent of sourcing where inline scripts/styles are coming from (e.g. via `wp_localize_script()`, `wp_add_inline_script()`, and `wp_set_script_translations()`, and `wp_add_inline_style()`). Beyond `WP_Scripts` and `WP_Styles`, it also goes to the extent of identifying the source of scripts/styles that are printed manually, such as in `wp_head` or `wp_footer`. Not all of this will be applicable or important perhaps in the beginning, but I think it's all doable for core.

manuelRod:
guys, I can work bringing https://github.com/audrasjb/site-health-audit-enqueued-assets/blob/master/site-health-audit-enqueued-assets.php to a module, and then we can discuss further messaging/features?

ThierryA:
@manuelRod thanks for raising your hand. Sure you can either open a PR bringing the code in a module and we discuss further on it (I have some thoughts about the `> 10` asset as a signal) or alternatively edit this issue description with more details about the approach and we continue discussing in this issue. Either war works 😄

manuelRod:
@ThierryA here is the PR: https://github.com/WordPress/performance/pull/54

We can continue the conversation here.

[...]

Issue #46, opened 2021-12-08
Module Slugs Naming Proposal


manuelRod:
As the number of modules is gonna grow, having just /modules/{module-slug} can lead to difficulties when trying to find one.

Maybe a different approach would help us in the future to easily identify, to which focus area they belong to, for instance:
/modules/{focus}/{module-slug} 
/modules/{focus}-{module-slug}

mehigh:
+1 for a flat structure like /modules/{focus}-{module-slug}

ThierryA:
Generally agree with organizing modules by focus area, I would personally prefer `/modules/{focus}/{module-slug}` as `/modules/{focus}-{module-slug}` by not be obvious to everyone, having folders for focus area makes it more obvious that modules are "grouped".

JustinyAhin:
I agree on having a` /modules/{focus}/{module-slug}` structure for more readability.

adamsilverstein:
makes sense, good suggestion @manuelRod!

felixarntz:
I'm in favor of the nested structure `/modules/{focus}/{module-slug}` as well.

However, note that this will require code changes (e.g. module parsing) and documentation changes, so we need a PR. It would also mean that modules should no longer need to specify the focus in the module header, since it would already be defined by the folder name.

Let's implement the changes to make it work, but we should then only start using that new approach once an eventual PR is merged. In other words, modules added now should still rely on the flat structure without any mention of the "focus" because that's how our code and documentation work now. The PR for this one would then need to migrate any existing modules to the new structure, and after merge all new modules would need to follow that.

ThierryA:
I would be in favor of doing it rather sooner than later to avoid having more modules needing to migrate.

adamsilverstein:
👍🏼  I will create a PR to update the webp-uploads module and adjust our docs to reflect the new naming convention.

felixarntz:
@adamsilverstein Make sure to also updated the way the modules are parsed and loaded, so that the plugin continues to work, e.g.:
* https://github.com/WordPress/performance/blob/trunk/load.php#L101
* https://github.com/WordPress/performance/blob/trunk/admin/load.php#L194

We may need to change the data in the option then so that it stores an array of strings, where each string is the focus slug and module slug separated by a slash, e.g. `images/webp-uploads`.

adamsilverstein:
[...]

PR #54, opened 2021-12-17
Add module to alert about excessive JS and CSS assets

manuelRod:
Addresses #25


modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,270 @@
...
+		),
+		'description' => sprintf(
+			'<p>%s</p>',
+			esc_html__( 'The amount of enqueued JS assets is acceptable.', 'performance-lab' )
+		),
+		'actions'     => '',
+		'test'        => 'enqueued_js_assets',
+	);
+
+	if ( $enqueued_scripts > 10 ) {
```
ThierryA: This signal needs to be more robust. The number of scripts is somehow a week signal, what really matter is the totally size. We need to grab and sum the size of all scripts and define a threshold for the warning.

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,270 @@
...
+		),
+		'description' => sprintf(
+			'<p>%s</p>',
+			esc_html__( 'The amount of enqueued CSS assets is acceptable.', 'performance-lab' )
+		),
+		'actions'     => '',
+		'test'        => 'enqueued_css_assets',
+	);
+
+	if ( $enqueued_styles > 10 ) {
```
ThierryA: Same comment as for the scripts, we need to grab and sum the total size of assets loaded.

ThierryA (changes requested):
Thanks for working on this 😄 I see this is a draft but I left a few early comments without doing a detailed review yet. The approach to trigger an asset warning is quite weak at the moment and need some more work.

modules/audit-enqueued-assets/load.php
manuelRod: Hello @ThierryA, should we combine both signals? number + size?  I think it stills makes sense to know the number of assets, so we can recommend concatenating them for instance. 

Should we define the size threshold and messaging?

manuelRod: @ThierryA I've implemented the logic to calculate sizes and so, we just need to define the thresholds and messaging.

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+/**
+ * Convert full URL paths to absolute paths.
+ *
+ * @since 1.0.0
+ *
+ * @param string $resource_url URl resource link.
+ * @return string Returns abosulte path to the resource.
+ */
+function perflab_get_path_from_resource_url( $resource_url ) {
+	return ABSPATH . wp_make_link_relative( $resource_url );
```
ThierryA: Note: this needs to be tested with various WP setup, url to path is a tricky thing to nail.

modules/audit-enqueued-assets/load.php
ThierryA: Great, let's discuss with the broader group about the thresholds for scripts and stylesheets size as well as the number if we decide to have one.

https://almanac.httparchive.org/en/2021/javascript#how-much-javascript-do-we-load and https://almanac.httparchive.org/en/2021/css can help giving us an understanding of the current state of the web.

We could base the threshold on CWV guidance if that make sense to everyone.

@felixarntz @adamsilverstein paging you here as well, would love more people thoughts on this too.

ThierryA: Replied on [the issue](https://github.com/WordPress/performance/issues/25#issuecomment-998639169) to avoid duplicate conversation.

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+
+/**
+ * Audit enqueued scripts in the frontend. Ignore /wp-includes scripts.
+ *
+ * It will save information in a transient for 12 hours.
+ *
+ * @since 1.0.0
+ */
+function perflab_aea_audit_enqueued_scripts() {
+	if ( ! is_admin() && ! get_transient( 'aea_enqueued_scripts' ) ) {
```
felixarntz: Three things about the conditions here:
* It needs to use a stricter check for the transient, since the transient could contain an empty array - it's unlikely that no scripts are enqueued, but it's certainly possible, so we need to change this to a `false === get_transient( 'aea_enqueued_scripts' )` instead.
* More importantly though, this code can set a transient when no user is logged in, which is problematic from both a security and frontend caching perspective. I think we will need to change this so that it only runs when a user is logged in, potentially even only when a user with a certain capability is logged in.
* Last but not least, right now this will record the enqueued scripts data from a "random" page, whatever a user accesses when the transient has expired. On different types of pages, different scripts may be enqueued. I understand we can't reasonably cover all the different scenarios, and especially for the first one we should keep it simple, but I strongly suggest to always use the same context to gather this data, otherwise it would be inconsistent and not very meaningful when comparing a before/after. I suggest we go with `is_front_page()`, i.e. only collecting data from the home page of the site.

I suggest we go with something like the following:
```suggestion
	if ( ! is_admin() && is_front_page() && current_user_can( 'view_site_health_checks' ) && false === get_transient( 'aea_enqueued_scripts' ) ) {
```

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+		foreach ( $wp_scripts->queue as $handle ) {
+			$src = $wp_scripts->registered[ $handle ]->src;
+			if ( $src && ! strpos( $src, 'wp-includes' ) ) {
+				$enqueued_scripts[] = array(
+					'src'  => $src,
+					'size' => perflab_get_resource_file_size( perflab_get_path_from_resource_url( $src ) ),
+				);
+			}
+		}
+		$expiration = apply_filters( 'perflab_aea_audit_enqueued_scripts_expiration_in_seconds', 12 * HOUR_IN_SECONDS );
```
felixarntz: Do we need a filter here? Not opposed if there's a good reason for it, but I think we should only introduce it if we have a good use-case for it, and keep things simple where we can.

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+		set_transient( 'aea_enqueued_scripts', $enqueued_scripts, $expiration );
```
felixarntz: Per my above comment, _if_ we agree that this should only collect data for a specific type of page, I suggest we name the transient accordingly, e.g. based on my above suggestion it could be something like `aea_enqueued_front_page_scripts`.

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+
+/**
+ * Audit enqueued styles in the frontend. Ignore /wp-includes styles.
+ *
+ * It will save information in a transient for 12 hours.
+ *
+ * @since 1.0.0
+ */
+function perflab_aea_audit_enqueued_styles() {
+	if ( ! is_admin() && ! get_transient( 'aea_enqueued_styles' ) ) {
```
felixarntz: Similar to my above comment, maybe this should be something closer to:
```suggestion
	if ( ! is_admin() && is_front_page() && current_user_can( 'view_site_health_checks' ) && false === get_transient( 'aea_enqueued_styles' ) ) {
```

Let's discuss on my comment above and once we've decided on something make the change in both places.

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+		foreach ( $wp_styles->queue as $handle ) {
+			$src = $wp_styles->registered[ $handle ]->src;
+			if ( $src && ! strpos( $src, 'wp-includes' ) ) {
+				$enqueued_styles[] = array(
+					'src'  => $src,
+					'size' => perflab_get_resource_file_size( perflab_get_path_from_resource_url( $src ) ),
+				);
+			}
+		}
+		$expiration = apply_filters( 'perflab_aea_audit_enqueued_styles_expiration_in_seconds', 12 * HOUR_IN_SECONDS );
```
felixarntz: See my above comment, do we really need a filter here?

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+		set_transient( 'aea_enqueued_styles', $enqueued_styles, $expiration );
```
felixarntz: See my above comment, depending on the decision we make, maybe this should be named `aea_enqueued_front_page_styles`?

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+
+/**
+ * Gets total size in bytes of Enqueued Scripts.
+ *
+ * @since 1.0.0
+ *
+ * @return int Byte Total size.
+ */
+function perflab_aea_get_total_size_bytes_enqueued_scripts() {
+	$total_size            = 0;
```
felixarntz: I think it would be more accurate to make this function behave similar to the above `perflab_aea_get_total_enqueued_scripts()` and return `false` if the transient hasn't been set yet. Right now it will always return `0` which makes it impossible to know whether there are no scripts being enqueued or whether the data just hasn't been collected yet.

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+
+/**
+ * Gets total size in bytes of Enqueued Styles.
+ *
+ * @since 1.0.0
+ *
+ * @return int Byte Total size.
+ */
+function perflab_aea_get_total_size_bytes_enqueued_styles() {
+	$total_size           = 0;
```
felixarntz: See my above comment, I think it would be more accurate if this function also returned `false` if the transient hasn't been set yet.

modules/audit-enqueued-assets/load.php
felixarntz: +1, we should warn if _either_ the number of scripts _or_ the overall size of the scripts is "too high". But it is indeed a complicated question what our thresholds should be for both of those, let's continue discussing that on the issue.

Maybe for now it would still be good to already change the code here to make the intention clear - maybe we start with a threshold of 100KB, just to have something to begin with and further prepare the code?

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+		'actions'     => '',
+		'test'        => 'enqueued_js_assets',
+	);
+
+	if ( $enqueued_scripts > 10 ) {
+		$result['status']         = 'recommended';
+		$result['badge']['color'] = 'orange';
+		$result['description']    = sprintf(
+			/* translators: %s: Number of enqueued scripts */
+			esc_html__( 'Your website enqueues %s scripts. Try to reduce the number of JS assets, or to concatenate them.', 'performance-lab' ),
```
felixarntz: Based on the above, I think we should also always include the overall script file size here, such as `Your website enqueues %1$s scripts (size: %2$s).`? We could use the [`size_format()` function](https://developer.wordpress.org/reference/functions/size_format/) to then fill in the size there.

Also, shouldn't this be wrapped in a `p` tag similar to how you have the `description` above?

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+	$result = array(
+		'label'       => esc_html__( 'Enqueued JS assets', 'performance-lab' ),
+		'status'      => 'good',
+		'badge'       => array(
+			'label' => esc_html__( 'Performance', 'performance-lab' ),
+			'color' => 'blue',
+		),
+		'description' => sprintf(
+			'<p>%s</p>',
+			esc_html__( 'The amount of enqueued JS assets is acceptable.', 'performance-lab' )
```
felixarntz: I think it would be worth still informing about the exact numbers here as well, i.e. providing the number of scripts and their overall file size here. Maybe the sentence could be like this: `The amount of %1$s enqueued scripts (size: %2$s) is acceptable.`

Nit-pick, but we should also use a consistent term throughout all messaging, i.e. "scripts" instead of "JS assets".

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+	$result = array(
+		'label'       => esc_html__( 'Enqueued CSS assets', 'performance-lab' ),
+		'status'      => 'good',
+		'badge'       => array(
+			'label' => esc_html__( 'Performance', 'performance-lab' ),
+			'color' => 'blue',
+		),
+		'description' => sprintf(
+			'<p>%s</p>',
+			esc_html__( 'The amount of enqueued CSS assets is acceptable.', 'performance-lab' )
```
felixarntz: Same comment as above, how about we make this: `The amount of %1$s enqueued styles (size: %2$s) is acceptable.`

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+		'actions'     => '',
+		'test'        => 'enqueued_css_assets',
+	);
+
+	if ( $enqueued_styles > 10 ) {
+		$result['status']         = 'recommended';
+		$result['badge']['color'] = 'orange';
+		$result['description']    = sprintf(
+			/* translators: %s: Number of enqueued styles */
+			esc_html__( 'Your website enqueues %s styles. Try to reduce the number of CSS assets, or to concatenate them.', 'performance-lab' ),
```
felixarntz: See my above comment, how about including the size here as well, e.g. `Your website enqueues %1$s styles (size: %2$s).` Also, similar to the above this needs to be wrapped in a `p` tag.

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+
+/**
+ * Invalidate both transients/cache on user clean_aea_audit action.
+ * Redirects to site-health.php screen adter clean up.
+ *
+ * @since 1.0.0
+ */
+function perflab_clean_aea_audit_action() {
+	if ( isset( $_GET['action'] ) && 'clean_aea_audit' === $_GET['action'] ) {
+		check_admin_referer( 'clean_aea_audit' );
```
felixarntz: This should probably also have a capability check for security purposes since it writes data. I suggest we check for `view_site_health_checks`.

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+	return $result;
+}
+
+/**
+ * Invalidate both transients/cache on user clean_aea_audit action.
+ * Redirects to site-health.php screen adter clean up.
+ *
+ * @since 1.0.0
+ */
+function perflab_clean_aea_audit_action() {
```
felixarntz: Nit-pick, but let's make sure all functions here are properly prefixed with `perflab_aea_*`.

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+	}
+}
+add_action( 'admin_init', 'perflab_clean_aea_audit_action' );
+
+/**
+ * Invalidate both transients/cache.
+ *
+ * @since 1.0.0
+ */
+function perflab_invalidate_cache_transients() {
```
felixarntz: See above, this should be prefixed with `perflab_aea_*`.

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+
+/**
+ * Convert full URL paths to absolute paths.
+ *
+ * @since 1.0.0
+ *
+ * @param string $resource_url URl resource link.
+ * @return string Returns abosulte path to the resource.
+ */
+function perflab_get_path_from_resource_url( $resource_url ) {
```
felixarntz: See above, let's prefix this with `perflab_aea_*`.

modules/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,312 @@
...
+
+/**
+ * If file exists, returns its size.
+ *
+ * @since 1.0.0
+ *
+ * @param string $file_src Path to the file.
+ * @return int Returns size if file exists, 0 if it doesn't.
+ */
+function perflab_get_resource_file_size( $file_src ) {
```
felixarntz: See above, let's prefix this with `perflab_aea_*`.

felixarntz (commented):
Great work so far @manuelRod! I agree the most important decision we need to make here is find good thresholds for the amount of assets and their filesize when we should start warning users.

Other than that, I left a few specific code comments below for things I think can be improved. My most important point would be that we also need to add appropriate capability checks and collect the enqueued assets data in a single consistent scenario instead of "anywhere in the frontend".

felixarntz (commented):
@manuelRod Two higher-level points of feedback:
* Now that #46 / #58 is merged, please refresh this against latest `trunk` and update all the module- and test-related files here to be in a `site-health` subdirectory, according to the updated module specification.
* It may help to split the module itself into a few smaller files, right now there is just a single `load.php` which makes it hard to overview the code. A good starting point could be to have only the main hooks and their functions in the `load.php` file and the other lower-level functions could be in separate files that are `require_once`d in `load.php`.

modules/audit-enqueued-assets/load.php
manuelRod: @felixarntz this was introduced in case some webmaster wants to control programmatically the expiration of this cache, in case of experiments or whatever the need may arise for them. Would it have a real use case? hard to know, but I think it doesn't hurt.

modules/audit-enqueued-assets/load.php
manuelRod: nice finding! solved

modules/audit-enqueued-assets/load.php
manuelRod: let's discuss this in the related one.

modules/audit-enqueued-assets/load.php
manuelRod: @felixarntz I agree with the extra checks you propose. Let's go for **is_front_page** for simplicity (for now) maybe we can even update the messaging to mention this? I would say in future iterations, we can capture information for other important pages too, like **is_shop** for woo (if installed) and others, since that will be also really valuable info.

manuelRod: I've implemented this approach.

modules/audit-enqueued-assets/load.php
manuelRod: refactored

modules/audit-enqueued-assets/load.php
manuelRod: refactored

modules/audit-enqueued-assets/load.php
manuelRod: added check

manuelRod:
@felixarntz I've implemented all your requests and updated the module to the new structure, let's continue the conversation in the issue for thresholds and messaging.

modules/audit-enqueued-assets/load.php
manuelRod: agree and implemented

modules/audit-enqueued-assets/load.php
manuelRod: implemented

JustinyAhin:
@manuelRod I went through the code, and the general implementation feels good.

I also discussed this with @aristath and there's a concern we had. For a scenario like this one:

> You visit a site front page, then go to the site health page and see I have a problem (let's say excessive load scripts size). 
Then you deactivate some plugins, and go back to the site health page. No change... it will still show the previous warning, because the user haven’t visited the front page to regenerate the values.

They are also similar scenarios with themes, and also when the front page or the front page template is edited (new blocks that add new styles/scripts for instance).

We should clean up the transients after these events.

manuelRod:
> @manuelRod I went through the code, and the general implementation feels good.
> 
> I also discussed this with @aristath and there's a concern we had. For a scenario like this one:
> 
> > You visit a site front page, then go to the site health page and see I have a problem (let's say excessive load scripts size).
> > Then you deactivate some plugins, and go back to the site health page. No change... it will still show the previous warning, because the user haven’t visited the front page to regenerate the values.
> 
> They are also similar scenarios with themes, and also when the front page or the front page template is edited (new blocks that add new styles/scripts for instance).
> 
> We should clean up the transients after these events.

@JustinyAhin You have good points there, but some of your concerns are already covered [here](https://github.com/WordPress/performance/pull/54/files#diff-a728eb492f98421696bcf274525576b551e1903782081652543907c198cc0cc5R230), there are 3 hooks to clean transients ( the previous warning will disappear ):
```
add_action( 'switch_theme', 'perflab_aea_invalidate_cache_transients' );
add_action( 'activated_plugin', 'perflab_aea_invalidate_cache_transients' );
add_action( 'deactivated_plugin', 'perflab_aea_invalidate_cache_transients' );
```

Also, we offer the possibility to clean the test cache on demand (we maybe should word it better), this would cover people trying to make different changes and resetting the test.

Screen Shot 2022-01-19 at 10 30 25 AM

JustinyAhin:
Hello @manuelRod. 

While doing some testing, I came across something.

When Query Monitor plugin (https://wordpress.org/plugins/query-monitor) is activated on the site, `get_transient( 'aea_enqueued_front_page_styles' )` also included this style `/wp-content/plugins/query-monitor/assets/query-monitor.css`.

manuelRod:
@JustinyAhin Would you mind explaining what would be the issue? 
If query monitor plugin enqueues that CSS in the frontend, then the plugin includes it. That would be the right behavior. 
Or I am missing your point?

JustinyAhin:
> @JustinyAhin Would you mind explaining what would be the issue? If query monitor plugin enqueues that CSS in the frontend, then the plugin includes it. That would be the right behavior. Or I am missing your point?

@manuelRod the query monitor CSS is only enqueued when logged in as an admin. I was wondering if we should include these kinds of assets in the checks.

aristath:
Maybe we could have a filter to allow changing the list of scripts/styles that the implementation detects?
Or - ideally - only update the transient for anonymous visits? 🤔

manuelRod:
@JustinyAhin @aristath I added the check of capabilities to save transients per @felixarntz request [here](https://github.com/WordPress/performance/pull/54#discussion_r776509031).

> More importantly though, this code can set a transient when no user is logged in, which is problematic from both a security and frontend caching perspective. I think we will need to change this so that it only runs when a user is logged in, potentially even only when a user with a certain capability is logged in.

aristath:
Hmmm I don't see how setting a transient could be considered a security risk... 🤔 
I do see the point about it being an issue from a frontend caching perspective though.

What if we added a button in the module, something like "refresh stats/caches/whatever"? On click, we could load the frontend in an invisible iframe using a nonce & a special param in the URL to trigger the transient update. In the request, we can also "fake" the user-ID to be an anonymous user so it doesn't load all the extra scripts/styles normally associated with an admin user.

manuelRod:
@aristath there is already a "reset" button to clean the transient (but that's about it).
Maybe we can discuss different approaches at this afternoon's meeting.

modules/site-health/audit-enqueued-assets/helper.php
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * @return int|false Byte Total size or false if transient hasn't been set.
+ */
+function perflab_aea_get_total_size_bytes_enqueued_styles() {
+	$total_size           = false;
+	$list_enqueued_styles = get_transient( 'aea_enqueued_front_page_styles' );
+	if ( $list_enqueued_styles ) {
+		foreach ( $list_enqueued_styles as $enqueued_style ) {
+			$total_size += $enqueued_style['size'];
+		}
```
felixarntz: I think it's safer to set `$total_size` to `0` before starting to add to it. I think this will work, but it's technically problematic, and I could see how this may eventually cause a warning (you shouldn't add numeric values to a boolean).
```suggestion
	if ( $list_enqueued_styles ) {
		$total_size = 0;
		foreach ( $list_enqueued_styles as $enqueued_style ) {
			$total_size += $enqueued_style['size'];
		}
```

modules/site-health/audit-enqueued-assets/helper.php
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * @return int|false Byte Total size or false if transient hasn't been set.
+ */
+function perflab_aea_get_total_size_bytes_enqueued_scripts() {
+	$total_size            = false;
+	$list_enqueued_scripts = get_transient( 'aea_enqueued_front_page_scripts' );
+	if ( $list_enqueued_scripts ) {
+		foreach ( $list_enqueued_scripts as $enqueued_script ) {
+			$total_size += $enqueued_script['size'];
+		}
```
felixarntz: See other comment:
```suggestion
	if ( $list_enqueued_scripts ) {
		$total_size = 0;
		foreach ( $list_enqueued_scripts as $enqueued_script ) {
			$total_size += $enqueued_script['size'];
		}
```

modules/site-health/audit-enqueued-assets/helper.php
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Convert full URL paths to absolute paths.
+ *
+ * @since 1.0.0
+ *
+ * @param string $resource_url URl resource link.
+ * @return string Returns absolute path to the resource.
+ */
+function perflab_aea_get_path_from_resource_url( $resource_url ) {
+	return ABSPATH . wp_make_link_relative( $resource_url );
```
felixarntz, line 88: This is okay for a start, but it's a bit oversimplified:
* It is possible for URLs to be in a location that is not within `ABSPATH`, for example several sites have their `wp-content` directory separate, which is supported by WordPress.
* WordPress may be in a subdirectory of the domain.

For most sites what you have here will work, but we should make sure to also accomodate those other sites.

I suggest we expand this as follows to at least cover for these two cases a bit better:
```suggestion
	if ( 0 === strpos( $resource_url, WP_CONTENT_URL ) ) {
		return WP_CONTENT_DIR . substr( $resource_url, strlen( WP_CONTENT_URL ) );
	}

	$site_url = untrailingslashit( site_url() );
	if ( 0 === strpos( $resource_url, $site_url ) ) {
		return ABSPATH . substr( $resource_url, strlen( $site_url ) );
	}

	return ABSPATH . wp_make_link_relative( $resource_url );
```

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,232 @@
...
+		foreach ( $wp_scripts->queue as $handle ) {
+			$src = $wp_scripts->registered[ $handle ]->src;
+			if ( $src && ! strpos( $src, 'wp-includes' ) ) {
+				$enqueued_scripts[] = array(
+					'src'  => $src,
+					'size' => perflab_aea_get_resource_file_size( perflab_aea_get_path_from_resource_url( $src ) ),
+				);
+			}
+		}
+		$expiration = apply_filters( 'perflab_aea_enqueued_front_page_scripts_expiration_in_seconds', 12 * HOUR_IN_SECONDS );
```
felixarntz: Following up on the other comment thread in https://github.com/WordPress/performance/pull/54#discussion_r776510603, since we don't have a use-case for it in mind, I suggest we remove it? We can always add it in later if something comes up where it's useful, but for maintenance it's better not to arbitrarily add filters for something just because we can.

If we think there is a use-case for it, I think at least we should not have multiple filters for this in the different functions, since they are _way_ too specific, which is also clear by looking at the very long filter name. Again, I'd prefer to remove it, but if you feel strongly that we should keep it, maybe let's instead introduce a single new function like `perflab_aea_get_transient_expiration()` which applies this filter in a consistent way throughout this module (e.g. the filter could have its name simplified to `perflab_aea_transient_expiration`). Then the code here could be simplified to
```suggestion
		$expiration = perflab_aea_get_transient_expiration();
```

Also, for any of these filters, if we keep them, they still need to receive a doc-block, which is currently missing.

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,232 @@
...
+		foreach ( $wp_styles->queue as $handle ) {
+			$src = $wp_styles->registered[ $handle ]->src;
+			if ( $src && ! strpos( $src, 'wp-includes' ) ) {
+				$enqueued_styles[] = array(
+					'src'  => $src,
+					'size' => perflab_aea_get_resource_file_size( perflab_aea_get_path_from_resource_url( $src ) ),
+				);
+			}
+		}
+		$expiration = apply_filters( 'perflab_aea_enqueued_front_page_styles_expiration_in_seconds', 12 * HOUR_IN_SECONDS );
```
felixarntz: See my above comment. If we go with that, we could in here e.g. simplify to
```suggestion
		$expiration = perflab_aea_get_transient_expiration();
```

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,232 @@
...
+			'<p>' . esc_html__( 'The amount of %1$s enqueued %2$s (size: %3$s) is acceptable.', 'performance-lab' ) . '</p>',
+			$enqueued_scripts,
+			_n( 'script', 'scripts', $enqueued_scripts, 'performance-lab' ),
+			size_format( perflab_aea_get_total_size_bytes_enqueued_scripts() )
+		),
+		'actions'     => '',
+		'test'        => 'enqueued_js_assets',
+	);
+
+	if ( $enqueued_scripts > 30 || perflab_aea_get_total_size_bytes_enqueued_scripts() > 300000 ) {
```
felixarntz: Contrary to the above filters, I think all of these thresholds may benefit from being filterable, especially since we're still unclear about them. I just want to mention that here, we can also leave it as is for now and add filters later, if we need them.

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,232 @@
...
+		$result['status']         = 'recommended';
+		$result['badge']['color'] = 'orange';
+		$result['description']    = sprintf(
+		/* translators: 1: Number of enqueued scripts. 2: "script" word. 3.Scripts size. */
+			'<p>' . esc_html__( 'Your website enqueues %1$s %2$s (size: %3$s). Try to reduce the number or to concatenate them.', 'performance-lab' ) . '</p>',
+			$enqueued_scripts,
+			_n( 'script', 'scripts', $enqueued_scripts, 'performance-lab' ),
```
felixarntz: This doesn't work for localization, since it results in a sentence being split. While it may work for several languages, it's not appropriately translatable. Please integrate it into the full sentence in line 130 (i.e. you need to use `esc_html( _n( ... ) )` there). We need to fill in the numbers with `sprintf`, but this shouldn't be used for words that are part of the sentence.

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,232 @@
...
+		'status'      => 'good',
+		'badge'       => array(
+			'label' => esc_html__( 'Performance', 'performance-lab' ),
+			'color' => 'blue',
+		),
+		'description' => sprintf(
+		/* translators: 1: Number of enqueued scripts. 2: "script" word. 3.Scripts size. */
+			'<p>' . esc_html__( 'The amount of %1$s enqueued %2$s (size: %3$s) is acceptable.', 'performance-lab' ) . '</p>',
+			$enqueued_scripts,
+			_n( 'script', 'scripts', $enqueued_scripts, 'performance-lab' ),
```
felixarntz: See my other comment, this needs to be integrated into the translation string in line 116.

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,232 @@
...
+		'status'      => 'good',
+		'badge'       => array(
+			'label' => esc_html__( 'Performance', 'performance-lab' ),
+			'color' => 'blue',
+		),
+		'description' => sprintf(
+		/* translators: 1: Number of enqueued styles. 2: "styles" word. 3.Styles size. */
+			'<p>' . esc_html__( 'The amount of %1$s enqueued %2$s (size: %3$s) is acceptable.', 'performance-lab' ) . '</p>',
+			$enqueued_styles,
+			_n( 'style', 'styles', $enqueued_styles, 'performance-lab' ),
```
felixarntz: See my other comment, this needs to be integrated into the translation string in line 172.

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,232 @@
...
+	);
+
+	if ( $enqueued_styles > 10 || perflab_aea_get_total_size_bytes_enqueued_styles() > 100000 ) {
+		$result['status']         = 'recommended';
+		$result['badge']['color'] = 'orange';
+		$result['description']    = sprintf(
+		/* translators: 1: Number of enqueued styles. 2: "style" word. 3.Styles size. */
+			'<p>' . esc_html__( 'Your website enqueues %1$s %2$s (size: %3$s). Try to reduce the number or to concatenate them.', 'performance-lab' ) . '</p>',
+			$enqueued_styles,
+			_n( 'style', 'styles', $enqueued_styles, 'performance-lab' ),
```
felixarntz: See my other comment, this needs to be integrated into the translation string in line 186.

felixarntz (commented):
@manuelRod First of all, apologies for the long delay in review - I'll make sure to follow up more quickly next time.

This is mostly looking good now, I left a few more comments, but after those are addressed, LGTM from my end.

felixarntz:
@manuelRod @aristath @JustinyAhin How do you feel about merging this PR once it's good to go with what there is for now? This is a fair amount of code, and once the last quirks are resolved, I think this is a good first proof of concept. While in WP core we only commit things we really think are ready, this makes code review very complicated and unfocused, and I don't think we need to operate like that for this plugin (it's not the case in Gutenberg either).

I agree that we definitely have to find a solution for excluding the admin-exclusive assets for the numbers here, but I think we can explore that in a follow-up issue that is only focused on that and not with everything else in this module. Same goes for the thresholds which we may still iterate on.

How does that sound to you?

modules/audit-enqueued-assets/load.php
felixarntz: Followed up in https://github.com/WordPress/performance/pull/54#discussion_r791704370

I still prefer to remove it since we don't have a good use-case in mind for it. But the other comment has some more nuance, let's follow up there.

aristath:
I agree, as a POC this is fine. I'm good with merging this one as-is, and then iterate and improve things as we go 👍

manuelRod:
Agree, I think this version is a good starting point and we can iterate it later. Also good to have it shipped with beta v1.0.0 so this plugin gets more _stuff_ and exposure.

@felixarntz thanks for the new feedback, I'll check

modules/audit-enqueued-assets/load.php
manuelRod: closing since they are outdated

modules/site-health/audit-enqueued-assets/helper.php
manuelRod: @felixarntz can you give me an example of the first case? I cannot see it clearly.

Regarding _WordPress may be in a subdirectory of the domain._
Do you mean something like this? 
https://example.com/blog/wp-content/themes/test-theme/style.css ?
If that's the case, doesn't ABSPATH . wp_make_link_relative( $resource_url ) covers it too?

modules/site-health/audit-enqueued-assets/load.php
manuelRod: Ok, agree, let's remove it for simplicity since we don't have a real use case here, and we want to roll this out soon.

modules/site-health/audit-enqueued-assets/helper.php
felixarntz: @manuelRod Some popular environments like https://github.com/roots/bedrock have their `wp-content` directory outside of the (rest of the) WordPress directory, so that URLs would be e.g. `https://example.com/content/themes/...` and `https://example.com/wp/wp-includes/...`. Both conditions I'm suggesting here cover that use-case, the first is for assets that come from the `wp-content` directory in a custom location, the second one for when the WordPress directory is in a subdirectory.

> If that's the case, doesn't ABSPATH . wp_make_link_relative( $resource_url ) covers it too?

I don't think it does, since `wp_make_link_relative` simply removes the protocol and domain. Looking at my example above, that means you get `/wp/wp-includes/...`, and when you append that to `ABSPATH`, the `wp` directory would basically be in there twice. That's why for these cases we have to use the `site_url()` function, which is the "URL equivalent" of `ABSPATH` (the same way that `WP_CONTENT_URL` is the "URL equivalent" to `WP_CONTENT_DIR`).

JustinyAhin:
@felixarntz I also agree on merging this as it is now, and iterate after that.

modules/site-health/audit-enqueued-assets/helper.php
manuelRod: good catch.

modules/site-health/audit-enqueued-assets/load.php
manuelRod: Filters added, you are right, these make more sense.

modules/site-health/audit-enqueued-assets/load.php
manuelRod: You are right, I've fixed all the occurrences.

manuelRod:
@felixarntz there is only one outstanding point, [see](https://github.com/WordPress/performance/pull/54#discussion_r791698929) after this we can merge!

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -112,27 +110,65 @@ function perflab_aea_enqueued_js_assets_test() {
...
 		'description' => sprintf(
-		/* translators: 1: Number of enqueued scripts. 2: "script" word. 3.Scripts size. */
-			'<p>' . esc_html__( 'The amount of %1$s enqueued %2$s (size: %3$s) is acceptable.', 'performance-lab' ) . '</p>',
-			$enqueued_scripts,
-			_n( 'script', 'scripts', $enqueued_scripts, 'performance-lab' ),
-			size_format( perflab_aea_get_total_size_bytes_enqueued_scripts() )
+			'<p>%s</p>',
+			esc_html(
+				sprintf(
+				/* translators: 1: Number of enqueued styles. 2.Styles size. */
```
felixarntz: Nit-pick, but can you indent one tab more this so that it is aligned with the below `_n` call that it applies to?

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -112,27 +110,65 @@ function perflab_aea_enqueued_js_assets_test() {
...
 
-	if ( $enqueued_scripts > 30 || perflab_aea_get_total_size_bytes_enqueued_scripts() > 300000 ) {
+	/**
+	 * Filters number of enqueued scripts to trigger warning.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param int $scripts_limit Scripts limit number. Default '30'.
+	 */
+	$scripts_limit = apply_filters( 'perflab_aea_enqueued_scripts_limit', 30 );
```
felixarntz: Just an idea: What do you think about using the term "threshold" rather than "limit"? Since we don't really "limit" how many scripts can be used.

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -112,27 +110,65 @@ function perflab_aea_enqueued_js_assets_test() {
...
 		'test'        => 'enqueued_js_assets',
 	);
 
-	if ( $enqueued_scripts > 30 || perflab_aea_get_total_size_bytes_enqueued_scripts() > 300000 ) {
+	/**
+	 * Filters number of enqueued scripts to trigger warning.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param int $scripts_limit Scripts limit number. Default '30'.
```
felixarntz: Nit-pick, but since the value is not a string, this would be more clear just as `Default 30.`

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -112,27 +110,65 @@ function perflab_aea_enqueued_js_assets_test() {
...
+	 */
+	$scripts_limit = apply_filters( 'perflab_aea_enqueued_scripts_limit', 30 );
+
+	/**
+	 * Filters size of enqueued scripts to trigger warning.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param int $scripts_size_limit Enqueued Scripts size (in bytes) limit. Default '300000'.
```
felixarntz: See above, since this is an integer, better to use `Default 300000.`

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -112,27 +110,65 @@ function perflab_aea_enqueued_js_assets_test() {
...
+	 */
+	$scripts_size_limit = apply_filters( 'perflab_aea_enqueued_scripts_byte_size_limit', 300000 );
```
felixarntz: See above, maybe use "threshold"?

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -112,27 +110,65 @@ function perflab_aea_enqueued_js_assets_test() {
...
-			'<p>' . esc_html__( 'Your website enqueues %1$s %2$s (size: %3$s). Try to reduce the number or to concatenate them.', 'performance-lab' ) . '</p>',
-			$enqueued_scripts,
-			_n( 'script', 'scripts', $enqueued_scripts, 'performance-lab' ),
-			size_format( perflab_aea_get_total_size_bytes_enqueued_scripts() )
+
+		$result['description'] = sprintf(
+			'<p>%s</p>',
+			esc_html(
+				sprintf(
+				/* translators: 1: Number of enqueued styles. 2.Styles size. */
```
felixarntz: See above, can you indent this by one more tab so that it's aligned with the function call below?

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -168,28 +204,64 @@ function perflab_aea_enqueued_css_assets_test() {
...
 		'description' => sprintf(
-		/* translators: 1: Number of enqueued styles. 2: "styles" word. 3.Styles size. */
-			'<p>' . esc_html__( 'The amount of %1$s enqueued %2$s (size: %3$s) is acceptable.', 'performance-lab' ) . '</p>',
-			$enqueued_styles,
-			_n( 'style', 'styles', $enqueued_styles, 'performance-lab' ),
-			size_format( perflab_aea_get_total_size_bytes_enqueued_styles() )
+			'<p>%s</p>',
+			esc_html(
+				sprintf(
+				/* translators: 1: Number of enqueued styles. 2.Styles size. */
```
felixarntz: See above.

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -168,28 +204,64 @@ function perflab_aea_enqueued_css_assets_test() {
...
 
-	if ( $enqueued_styles > 10 || perflab_aea_get_total_size_bytes_enqueued_styles() > 100000 ) {
+	/**
+	 * Filters number of enqueued styles to trigger warning.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param int $styles_limit Styles limit number. Default 10.
+	 */
+	$styles_limit = apply_filters( 'perflab_aea_enqueued_styles_limit', 10 );
```
felixarntz: See above, maybe use "threshold"?

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -168,28 +204,64 @@ function perflab_aea_enqueued_css_assets_test() {
...
+
+	/**
+	 * Filters size of enqueued styles to trigger warning.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param int $styles_size_limit Enqueued styles size (in bytes) limit. Default 100000.
+	 */
+	$styles_size_limit = apply_filters( 'perflab_aea_enqueued_styles_byte_size_limit', 100000 );
```
felixarntz: See above.

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -168,28 +204,64 @@ function perflab_aea_enqueued_css_assets_test() {
...
-			'<p>' . esc_html__( 'Your website enqueues %1$s %2$s (size: %3$s). Try to reduce the number or to concatenate them.', 'performance-lab' ) . '</p>',
-			$enqueued_styles,
-			_n( 'style', 'styles', $enqueued_styles, 'performance-lab' ),
-			size_format( perflab_aea_get_total_size_bytes_enqueued_styles() )
+
+		$result['description'] = sprintf(
+			'<p>%s</p>',
+			esc_html(
+				sprintf(
+				/* translators: 1: Number of enqueued styles. 2.Styles size. */
```
felixarntz: See above.

tests/testdata/modules/site-health/audit-enqueued-assets/class-site-health-mock-responses.php
```diff
@@ -45,11 +45,20 @@ public static function return_aea_enqueued_js_assets_test_callback_less_than_lim
...
 			'description' => sprintf(
-			/* translators: 1: Number of enqueued scripts. 2: "script" word. 3.Scripts size. */
-				'<p>' . esc_html__( 'The amount of %1$s enqueued %2$s (size: %3$s) is acceptable.', 'performance-lab' ) . '</p>',
-				$enqueued_scripts,
-				_n( 'script', 'scripts', $enqueued_scripts, 'performance-lab' ),
-				size_format( perflab_aea_get_total_size_bytes_enqueued_scripts() )
+				'<p>%s</p>',
+				esc_html(
+					sprintf(
+					/* translators: 1: Number of enqueued styles. 2.Styles size. */
```
felixarntz: See above.

tests/testdata/modules/site-health/audit-enqueued-assets/class-site-health-mock-responses.php
```diff
@@ -69,13 +78,22 @@ public static function return_aea_enqueued_js_assets_test_callback_more_than_lim
...
 		$result['description']    = sprintf(
-		/* translators: 1: Number of enqueued scripts. 2: "script" word. 3.Scripts size. */
-			'<p>' . esc_html__( 'Your website enqueues %1$s %2$s (size: %3$s). Try to reduce the number or to concatenate them.', 'performance-lab' ) . '</p>',
-			$enqueued_scripts,
-			_n( 'script', 'scripts', $enqueued_scripts, 'performance-lab' ),
-			size_format( perflab_aea_get_total_size_bytes_enqueued_scripts() )
+			'<p>%s</p>',
+			esc_html(
+				sprintf(
+				/* translators: 1: Number of enqueued styles. 2.Styles size. */
```
felixarntz: See above.

tests/testdata/modules/site-health/audit-enqueued-assets/class-site-health-mock-responses.php
```diff
@@ -102,11 +120,20 @@ public static function return_aea_enqueued_css_assets_test_callback_less_than_li
...
 			'description' => sprintf(
-			/* translators: 1: Number of enqueued styles. 2: "styles" word. 3.Styles size. */
-				'<p>' . esc_html__( 'The amount of %1$s enqueued %2$s (size: %3$s) is acceptable.', 'performance-lab' ) . '</p>',
-				$enqueued_styles,
-				_n( 'style', 'styles', $enqueued_styles, 'performance-lab' ),
-				size_format( perflab_aea_get_total_size_bytes_enqueued_styles() )
+				'<p>%s</p>',
+				esc_html(
+					sprintf(
+					/* translators: 1: Number of enqueued styles. 2.Styles size. */
```
felixarntz: See above.

tests/testdata/modules/site-health/audit-enqueued-assets/class-site-health-mock-responses.php
```diff
@@ -126,13 +153,22 @@ public static function return_aea_enqueued_css_assets_test_callback_more_than_li
...
 		$result['description']    = sprintf(
-		/* translators: 1: Number of enqueued styles. 2: "style" word. 3.Styles size. */
-			'<p>' . esc_html__( 'Your website enqueues %1$s %2$s (size: %3$s). Try to reduce the number or to concatenate them.', 'performance-lab' ) . '</p>',
-			$enqueued_styles,
-			_n( 'style', 'styles', $enqueued_styles, 'performance-lab' ),
-			size_format( perflab_aea_get_total_size_bytes_enqueued_styles() )
+			'<p>%s</p>',
+			esc_html(
+				sprintf(
+				/* translators: 1: Number of enqueued styles. 2.Styles size. */
```
felixarntz: See above.

felixarntz (commented):
@manuelRod Thanks for all the updates! This is pretty much good to go from my perspective, I just left some _super minor_ nit-pick comments on things that would be great to address, but not critical. The only other thing I think needs to be addressed still is https://github.com/WordPress/performance/pull/54#discussion_r791698929.

Would love for someone else to review this as well - any volunteers? :)

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -0,0 +1,304 @@
...
+				$enqueued_styles[] = array(
+					'src'  => $src,
+					'size' => perflab_aea_get_resource_file_size( perflab_aea_get_path_from_resource_url( $src ) ),
+				);
+			}
+		}
+		set_transient( 'aea_enqueued_front_page_styles', $enqueued_styles, 12 * HOUR_IN_SECONDS );
+	}
+}
+add_action( 'wp_print_styles', 'perflab_aea_audit_enqueued_styles' );
```
aristath, line 65: A few notes for this one:

* Styles can be printed in the header or the footer. I'm not sure `wp_print_styles` would cover all scenarios
* Stylesheets can have additional styles added inline - which should also be accounted for when getting the size
* For some styles we already know their path (it was part of a refactor for block styles which was merged in 5.9)
* `$wp_styles->queue` contains styles that are to be enqueued. However, a stylesheet can be enqueued _after_  this has run, and if I'm not mistaken any styles that have already been printed will not be included here. To avoid a priorities battle and over-complicaing the implementation, it would be better if we ran this as the last action in `wp_footer`, and check which stylesheets were done.

The below code fixes the above issues:

```php

add_action( 'wp_footer', function() {
	global $wp_styles;
	$enqueued_styles = array();

	// Loop through the styles that were printed, and get their details.
	foreach ( $wp_styles->done as $style_handle ) {
		$style = $wp_styles->registered[ $style_handle ];

		if ( ! empty( $style->extra ) && ! empty( $style->extra['path'] ) ) { // Check if we already have the style's path.
			$path = $style->extra['path'];
		} else { // Fallback to getting the path from the style's src.
			$path = perflab_aea_get_path_from_resource_url( $style->src );
		}

		// Add any extra data that was passed with the style.
		$extras_size = 0;
		if ( ! empty( $style->extra ) && ! empty( $style->extra['after'] ) ) {
			foreach ( $style->extra['after'] as $extra ) {
				$extras_size += strlen( $extra );
			}
		}
		
		$enqueued_styles[] = array(
			'src' => $style->src,
			'size' => perflab_aea_get_resource_file_size( $src ) + $extras_size,
		);
	}

	set_transient( 'aea_enqueued_front_page_styles', $enqueued_styles, 12 * HOUR_IN_SECONDS );

}, PHP_INT_MAX );
```

modules/site-health/audit-enqueued-assets/load.php
manuelRod: agree, I've updated terminology, also in tests.

modules/site-health/audit-enqueued-assets/load.php
manuelRod: thanks, @aristath for this valuable feedback, would this apply also to scripts? or just to styles?

aristath: > would this apply also to scripts? or just to styles?

Some (but not all) parts also apply to scripts, yes. Scripts can get added in the head or the footer, and they can get added while the page renders. For example, a block can enqueue its own script on-render. This will make the script get added only when the block gets rendered, but it won't be visible to `$wp_scripts` before we get to the footer.
However, scripts don't (yet) have a `path` defined, this is something we plan on doing in WP 6.0 - so that part can be skipped 👍

felixarntz: Good catch @aristath! Since considering this is also adding an extra "dimension" of code to this though, based on my earlier comment, would it be okay to enhance/fix this in a follow-up issue and PR?

felixarntz (approved):
Approved from my end - awesome work so far @manuelRod!

Per some of the comments, there are definitely a number of things to follow up on - the most important one from my personal perspective is collecting the asset information in a way that doesn't include admin-specific assets.

modules/site-health/audit-enqueued-assets/load.php
aristath: Of course! We can merge this one as-is, and improve in follow-up PRs 👍

aristath (approved):
Approving this one as it's a good starting point and we can iterate as we go in followup PRs 👍

manuelRod:
💯

PR #58, opened 2021-12-21
Update module directories to be within their focus directory

adamsilverstein:
Addresses https://github.com/WordPress/performance/issues/46

* Move the `webp-uploads` module into a `images` folder in `modules`.
* Update `perflab_get_modules` to parse through all the focus folders to look for modules.  
* Modules stored in settings now include their focus as part of the module name, for example `webp-uploads` changes to `images/webp-uploads` - for this reason, existing users will need to re-enable the webp-uploads module. Not a big concern since only developers are testing this so far.
* Update the docs to reflect the new file/folder structure for modules.
* Remove `Focus:` from the headers of all sample code and from the `webp-uploads` module header (only change to module itself)
* Add logic in `perflab_get_module_data` to use a regex to extract the module type from the module filename (which includes the focus slug as part of its path). Feedback on regex always welcome!


admin/load.php
```diff
@@ -200,39 +200,49 @@ function perflab_get_modules( $modules_root = null ) {
...
+	// Modules are organized as {focus}/{module-slug} in the modules folder.
 	if ( $modules_dir ) {
-		// phpcs:ignore WordPress.CodeAnalysis.AssignmentInCondition.FoundInWhileCondition
-		while ( ( $file = readdir( $modules_dir ) ) !== false ) {
-			if ( '.' === substr( $file, 0, 1 ) ) {
-				continue;
-			}
+		// Load modules from each focus area folder.
+		$focus_areas = perflab_get_focus_areas();
+		foreach ( array_keys( $focus_areas ) as $focus ) {
```
felixarntz: Can you keep the previous approach here where we go through all directories instead? The original code here was agnostic of which modules and focuses exist, which makes it more flexible and also easier to test, because it's not coupled to the focus areas that exist - which we've even talked about potentially removing them to not be hard-coded, which would be another reason to keep the "just go over the directories" approach here.

I would suggest you reinstate the original `while` loop, but then you check for `is_dir` and consider every top-level directory within a focus directory; then open that dir and iterate through it to get the modules similar to today.

admin/load.php
```diff
@@ -291,5 +300,11 @@ function perflab_get_module_data( $module_file ) {
 		$module_data['experimental'] = false;
 	}
 
+	// Extract the module focus from its filename.
+	preg_match( '/.*\/modules\/([^\/]*)\/.*/', $module_file, $matches );
+	if ( isset( $matches[1] ) ) {
+		$module_data['focus'] = $matches[1];
+	}
```
felixarntz: Further above, we already have a `dirname( $module_file )` which parses this into a simpler format of just `{focus}/{module-slug}`. It would be a bit simpler/more performant to assign that to a variable (e.g. `$module_dir`) instead and then parse the two parts using `explode`, e.g.:
```php
list( $focus, $slug ) = explode( '/', $module_dir );
```

Since now the array keys are the module dirs, I think it would be worth including the `slug` separately within `$module_data` as well.

docs/Writing-a-module.md
```diff
@@ -13,12 +13,11 @@ Every module surfaces on the admin settings page of the performance plugin, wher
 
 ## Module requirements
 
-* The production code for a module must all be located in a directory `/modules/{module-slug}` where `{module-slug}` is the module's slug.
-* The entry point file must be called `load.php` and per the above be located at `/modules/{module-slug}/load.php`.
+* The production code for a module must all be located in a directory `/modules/{focus}/{module-slug}` where `{module-slug}` is the module's slug and `{focus}` is the focus area: an identifier of a single focus (e.g. `images`). This should correspond to a section on the performance plugin's settings page. [See the `perflab_get_focus_areas()` function for the currently available focus areas.](../admin/load.php#L161)
+* The entry point file must be called `load.php` and per the above be located at `/modules/{focus}/{module-slug}/load.php`.
```
felixarntz, line 17: This change has to also be made in a few other places in this file, e.g. in the tests instructions and also in the location for the demo file. Please double-check to ensure they are all updated.

tests/testdata/demo-modules/demo-module-1/load.php
```diff
@@ -2,7 +2,6 @@
 /**
  * Module Name: Demo Module 1
  * Description: This is the description for demo module 1.
- * Focus: javascript
```
felixarntz, line 5: This file still needs to be moved accordingly.

tests/testdata/demo-modules/demo-module-2/load.php
```diff
@@ -2,7 +2,6 @@
 /**
  * Module Name: Demo Module 2
  * Description: This is the description for demo module 2.
- * Focus: something
```
felixarntz, line 5: See above, this file still needs to be moved accordingly (to a `something` directory, which of course is a focus that doesn't actually exist). This relates to my other comment above, parsing modules shouldn't be coupled to which focuses we have explicitly defined.

tests/testdata/demo-modules/demo-module-3/load.php
```diff
@@ -2,7 +2,6 @@
 /**
  * Module Name: Demo Module 3
  * Description: This is the description for demo module 3.
- * Focus: images
```
felixarntz, line 5: See above, this file still needs to be moved accordingly.

modules/images/webp-uploads/load.php
```diff
@@ -2,7 +2,6 @@
 /**
```
felixarntz, line 2: Can you also move all `webp-uploads`-related test files accordingly?

felixarntz (commented):
@adamsilverstein This looks good except for the logic change for parsing modules, which is now coupled to which focus areas we have defined and also causes tests to fail. Can you update that? I believe some related tests need to be updated as well.

Other than that, only minor comments for e.g. a few oversights that still need to be updated.

felixarntz:
@adamsilverstein Also, please add the appropriate PR labels and milestone based on #52.

admin/load.php
adamsilverstein: sure, I actually had it that way, then switched to using the specific focus names instead. in [`9ecb11a` (#58)](https://github.com/WordPress/performance/pull/58/commits/9ecb11a1dd6224c271c90269c35194cbad7dcdb8) which I will revert

docs/Writing-a-module.md
adamsilverstein: 👍🏼  I wasn't sure we wanted to change test structure as well; will update

tests/testdata/demo-modules/demo-module-2/load.php
adamsilverstein: 👍🏼

tests/testdata/demo-modules/demo-module-3/load.php
adamsilverstein: 👍🏼

modules/images/webp-uploads/load.php
adamsilverstein: 👍🏼

adamsilverstein:
Thanks for the feedback @felixarntz - I'll work on addressing these points.

admin/load.php
adamsilverstein: updated the approach to use `$module_dir` in [`06a5a78` (#58)](https://github.com/WordPress/performance/pull/58/commits/06a5a786bd327afd137655471b301ae0d4ed946d)

adamsilverstein: added slug to module data (plus data validation) in [`1d4984d` (#58)](https://github.com/WordPress/performance/pull/58/commits/1d4984d4a31ca2b8e09512acb29fcf37d07b8bfb)

admin/load.php
adamsilverstein: reverted that commit in [`fa5adc4` (#58)](https://github.com/WordPress/performance/pull/58/commits/fa5adc419c169385355b5841c514a17868637a3a), should be fixed

adamsilverstein:
Addressed feedback, updating tests next to work with new setup.

adamsilverstein:
@felixarntz ready for review:
* module loading now parses all folders in `modules`
* moved all module tests into `focus` folders
* used module_dir variable and store slug in module_data
* ensure all tests are updated to pass

admin/load.php
```diff
@@ -266,10 +280,11 @@ function( $a, $b ) {
  * @since 1.0.0
  *
  * @param string $module_file Absolute path to the main module file.
+ * @param string $module_dir  The module directory ({focus}/{module-slug}).
  * @return array|bool Associative array of parsed module data, or false on failure. Fields for every module include
  *                    'name', 'description', 'focus', and 'experimental'.
  */
-function perflab_get_module_data( $module_file ) {
+function perflab_get_module_data( $module_file, $module_dir ) {
```
felixarntz: I think passing `$module_dir` here is unnecessary, we could also get that inside this function from the `$module_file` via using a cobination of `dirname()` and `basename()`. That makes using the function simpler.

admin/load.php
```diff
@@ -266,10 +280,11 @@ function( $a, $b ) {
...
 	$default_headers = array(
```
felixarntz, line 292: This still needs to be adjusted to remove the `focus` field since it's no longer in the module header.

docs/Writing-a-module.md
```diff
@@ -13,18 +13,17 @@ Every module surfaces on the admin settings page of the performance plugin, wher
...
 * The module must neither rely on any PHP code from outside its directory nor on any external PHP code. If relying on an external PHP dependency is essential for a module, the approach should be evaluated and discussed with the wider team.
 * The module must use the `performance-lab` text domain for all of its localizable strings.
 * All global code structures in the module PHP codebase must be prefixed (e.g. with a string based on the module slug) to avoid conflicts with other modules or plugins.
-* All test code for a module (e.g. PHPUnit tests) must be located in a directory `/tests/modules/{module-slug}` where `{module-slug}` is the module's slug (i.e. the same folder name used above).
-    * If tests require some test-specific structures (e.g. dummy data or mock classes), those should be implemented in a directory `/tests/testdata/modules/{module-slug}`.
+* All test code for a module (e.g. PHPUnit tests) must be located in a directory `/tests/modules/{focus}/{module-slug}` where `{module-slug}` is the module's slug (i.e. the same folder name used above).
```
felixarntz:
```suggestion
* All test code for a module (e.g. PHPUnit tests) must be located in a directory `/tests/modules/{focus}/{module-slug}` where `{module-slug}` is the module's slug and `{focus}` is the focus area (i.e. the same folder names used above).
```

docs/Writing-a-module.md
```diff
@@ -35,14 +34,13 @@ Every module surfaces on the admin settings page of the performance plugin, wher
 
 ## Example
 
-The following is a minimum module entry point file `/modules/my-module/load.php` (i.e. the module slug is "my-module"):
+The following is a minimum module entry point file `/modules/focus/my-module/load.php` (i.e. the module slug is "my-module"):
```
felixarntz: Nit-pick, but maybe put an actual focus here such as `images`, to clarify we shouldn't just arbitrarily define focuses this way. This is different for the module slugs which can be any name.
```suggestion
The following is a minimum module entry point file `/modules/images/my-module/load.php` (i.e. the module slug is "my-module", and the focus is "images"):
```

tests/admin/load-tests.php
```diff
@@ -183,7 +186,7 @@ public function test_perflab_get_modules() {
 	public function test_perflab_get_module_data() {
 		// Use test data directory with demo modules that match the modules declared on top of this file.
 		foreach ( self::$demo_modules as $module_slug => $expected_module_data ) {
-			$module_data = perflab_get_module_data( TESTS_PLUGIN_DIR . '/tests/testdata/demo-modules/' . $module_slug . '/load.php' );
+			$module_data = perflab_get_module_data( TESTS_PLUGIN_DIR . '/tests/testdata/demo-modules/' . $module_slug . '/load.php', $module_slug );
```
felixarntz: See my above comment, I think it would be better to keep the parameters for that function as is and handle getting the focus and module slug internally.

felixarntz (commented):
@adamsilverstein Almost lgtm now! Only a few last minor comments.

felixarntz:
@adamsilverstein Note there is also #60 being implemented right now. Depending on which of these two PRs is merged first, the other one may need an update.

felixarntz:
@adamsilverstein I've pushed two small commits here just to rebase and adjust the code from #60 to work with this, since that was just merged (see https://github.com/WordPress/performance/pull/58#issuecomment-999920207).

admin/load.php
adamsilverstein: 👍🏼  Adjusted in [`f126464` (#58)](https://github.com/WordPress/performance/pull/58/commits/f1264644d7de67eb010224b0e10b49ad508adbb1)

admin/load.php
adamsilverstein: removed in [`86f7553` (#58)](https://github.com/WordPress/performance/pull/58/commits/86f755346278b3473f727a836af52a99a354576e)

tests/admin/load-tests.php
adamsilverstein: 👍🏼  done

adamsilverstein:
@felixarntz - ready for another review when you have a chance

felixarntz (approved):
@adamsilverstein Excellent work!

PR #60, opened 2021-12-21
Ensure that module header fields can be translated

### LINKED ISSUE #59: Module header information is not translatable
#2, #27, and #46 have established the required infrastructure for modules in the performance plugin. While this approach is what we want to use going forward, at this point there is still a crucial problem with it, which is that the user-facing strings within the module header (most importantly the module description) are not translatable, because they are a PHP comment and therefore don't go through any WordPress translation function (e.g. `__`).

Therefore, wordpress.org currently won't recognize these strings as translatable if we upload the plugin in that state. This issue is about implementing a similar approach to translate that data that is also used to translate plugin header values in WordPress core.

The main difference here will be that we need to generate a PHP file with those translations somewhere. That file doesn't even need to be `require`d anywhere, but it must be included in the deployment to the WordPress plugin repository so that it can recognize the strings accordingly. This is somewhat similar to what Gutenberg used to do early in its development phase where JS translations were not yet available via WordPress core and wordpress.org.

felixarntz:
Fixes #59

This PR implements an `npm run translations` command (via the JS infrastructure introduced in #51), which makes the two module header fields `Module Name` and `Description` translatable by extracting the values from all module files and putting them into a generated `module-i18n.php` file. This file allows wordpress.org to recognize these translations as if they were normally used in the plugin. The PR then uses a similar approach to WordPress core when it comes to the headers themselves in the production code: By calling `translate` it will technically go through the same function, which in combination with the generated file ensures that those are translatable in GlotPress.

The implementation of the script is heavily based on a combination of https://github.com/WordPress/gutenberg/pull/5310 and https://github.com/wp-pot/wp-pot/commit/c46b8cbc2c8f494405a71dd8ebaee2e7ce8bd79b, both of which do in parts similar tasks.

The generated file is `.gitignore`d, since it doesn't need to be in the repository (it isn't even `require`d anywhere, it just needs to "exist" when the plugin is deployed to wordpress.org).

To test the script locally, run `npm run translations`.

### DIFF
--- a/bin/plugin/commands/translations.js
+++ b/bin/plugin/commands/translations.js
@@ -11,19 +11,21 @@ const readline = require( 'readline' );
  * Internal dependencies
  */
 const { log, formats } = require( '../lib/logger' );
+const config = require( '../config' );
 
 /**
  * @typedef WPTranslationsCommandOptions
  *
  * @property {string=} directory Optional directory, default is the root `/modules` directory.
- * @property {string=} output    Optional output file, default is the root `/module-i18n.php` file.
+ * @property {string=} output    Optional output PHP file, default is the root `/module-i18n.php` file.
  */
 
 /**
  * @typedef WPTranslationsSettings
  *
- * @property {string=} directory Optional directory, default is the root `/modules` directory.
- * @property {string=} output    Optional output file, default is the root `/module-i18n.php` file.
+ * @property {string} textdomain Plugin textdomain.
+ * @property {string} directory  Modules directory.
+ * @property {string} output     Output PHP file.
  */
 
 const options = [
@@ -44,7 +46,9 @@ const options = [
  */
 async function handler( opt ) {
 	await createTranslations( {
-		directory: opt.directory,
+		textdomain: config.textdomain,
+		directory: opt.directory || 'modules',
+		output: opt.output || 'module-i18n.php',
 	} );
 }
 
@@ -52,6 +56,19 @@ module.exports = {
 	options,
 	handler,
 };
+
+const TAB = '\t';
+const NEWLINE = '\n';
+const FILE_HEADER =
+	[
+		'<?php',
+		'/* THIS IS A GENERATED FILE. DO NOT EDIT DIRECTLY. */',
+		'$generated_i18n_strings = array(',
+	].join( NEWLINE ) + NEWLINE;
+const FILE_FOOTER =
+	NEWLINE +
+	[ ');', '/* THIS IS THE END OF THE GENERATED FILE */' ].join( NEWLINE ) +
+	NEWLINE;
 
 /**
  * Parses module header translation strings.
@@ -103,26 +120,41 @@ async function getTranslations( settings ) {
 }
 
 /**
+ * Parses module header translation strings.
+ *
+ * @param {[]string}               translations List of translation strings.
+ * @param {WPTranslationsSettings} settings     Translations settings.
+ */
+function createTranslationsPHPFile( translations, settings ) {
+	const output = translations.map( ( translation ) => {
+		// Escape single quotes.
+		return (
+			TAB +
+			`__( '${ translation.replace( /'/g, "\\'" ) }', '${
+				settings.textdomain
+			}' ),`
+		);
+	} );
+
+	const fileOutput = FILE_HEADER + output.join( NEWLINE ) + FILE_FOOTER;
+	fs.writeFileSync( path.join( '.', settings.output ), fileOutput );
+}
+
+/**
  * Parses module header translation strings and generates a PHP file with them.
  *
  * @param {WPTranslationsSettings} settings Translations settings.
  */
 async function createTranslations( settings ) {
-	const fullSettings = {
-		...settings,
-		directory: settings.directory || 'modules',
-		output: settings.output || 'module-i18n.php',
-	};
-
 	log(
 		formats.title(
-			`\n💃Preparing module translations for "${ fullSettings.directory }" in "${ fullSettings.output }"\n\n`
+			`\n💃Preparing module translations for "${ settings.directory }" in "${ settings.output }"\n\n`
 		)
 	);
 
 	try {
-		const translations = await getTranslations( fullSettings );
-		log( translations );
+		const translations = await getTranslations( settings );
+		createTranslationsPHPFile( translations, settings );
 	} catch ( error ) {
 		if ( error instanceof Error ) {
 			log( formats.error( error.stack ) );
@@ -131,7 +163,7 @@ async function createTranslations( settings ) {
 
 	log(
 		formats.success(
-			`\n💃Module translations successfully set in "${ fullSettings.output }"\n\n`
+			`\n💃Module translations successfully set in "${ settings.output }"\n\n`
 		)
 	);
 }
--- a/bin/plugin/commands/translations.js
+++ b/bin/plugin/commands/translations.js
@@ -63,7 +63,8 @@ const FILE_HEADER = `<?php
 $generated_i18n_strings = array(
 `;
 const FILE_FOOTER = `
-'/* THIS IS THE END OF THE GENERATED FILE */'
+);
+/* THIS IS THE END OF THE GENERATED FILE */
 `;
 
 /**
--- a/bin/plugin/commands/translations.js
+++ b/bin/plugin/commands/translations.js
@@ -1,7 +1,6 @@
 /**
  * External dependencies
  */
-const { flatten } = require( 'lodash' );
 const path = require( 'path' );
 const glob = require( 'fast-glob' );
 const fs = require( 'fs' );
@@ -113,7 +112,7 @@ async function getTranslations( settings ) {
 		} )
 	);
 
-	return flatten( moduleTranslations );
+	return moduleTranslations.flat();
 }
 
 /**
--- a/bin/plugin/commands/translations.js
+++ b/bin/plugin/commands/translations.js
@@ -62,9 +62,7 @@ module.exports = {
  */
 async function getTranslations( settings ) {
 	const moduleFilePattern = path.join( settings.directory, '*/load.php' );
-	const moduleFiles = await glob(
-		path.resolve( '../../..', moduleFilePattern )
-	);
+	const moduleFiles = await glob( path.resolve( '.', moduleFilePattern ) );
 
 	const moduleTranslations = await Promise.all(
 		moduleFiles.map( async ( moduleFile ) => {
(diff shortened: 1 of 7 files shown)


bin/plugin/config.js
```diff
@@ -11,6 +11,7 @@
 const config = {
 	githubRepositoryOwner: 'WordPress',
 	githubRepositoryName: 'performance',
+	textdomain: 'performance-lab',
```
mitogh: To be consistent with the rest of the config this one should be:

```suggestion
	textDomain: 'performance-lab',
```

bin/plugin/commands/translations.js
```diff
@@ -0,0 +1,169 @@
...
+};
+
+const TAB = '\t';
+const NEWLINE = '\n';
+const FILE_HEADER =
+	[
+		'<?php',
+		'/* THIS IS A GENERATED FILE. DO NOT EDIT DIRECTLY. */',
+		'$generated_i18n_strings = array(',
+	].join( NEWLINE ) + NEWLINE;
```
mitogh: I think we can go with template strings instead due this is executed by Node.

```suggestion
const FILE_HEADER = `<?php
/* THIS IS A GENERATED FILE. DO NOT EDIT DIRECTLY. */
$generated_i18n_strings = array(
`;
```

bin/plugin/commands/translations.js
```diff
@@ -0,0 +1,169 @@
...
+const FILE_FOOTER =
+	NEWLINE +
+	[ ');', '/* THIS IS THE END OF THE GENERATED FILE */' ].join( NEWLINE ) +
+	NEWLINE;
```
mitogh: Same for this one as described above.

```suggestion
const FILE_FOOTER = `
'/* THIS IS THE END OF THE GENERATED FILE */'
`;
```

Or

```suggestion
const FILE_FOOTER = `${NEWLINE}/* THIS IS THE END OF THE GENERATED FILE */${NEWLINE}`;
```

bin/plugin/commands/translations.js
```diff
@@ -0,0 +1,169 @@
...
+						}
+					}
+				} );
+			}
+
+			return translationStrings;
+		} )
+	);
+
+	return flatten( moduleTranslations );
```
mitogh: `flat()` method since `Node 11` so this can be.

```suggestion
	return moduleTranslations.flat();
```

bin/plugin/commands/translations.js
```diff
@@ -0,0 +1,169 @@
...
+		// Escape single quotes.
+		return (
+			TAB +
+			`__( '${ translation.replace( /'/g, "\\'" ) }', '${
+				settings.textdomain
+			}' ),`
+		);
+	} );
+
+	const fileOutput = FILE_HEADER + output.join( NEWLINE ) + FILE_FOOTER;
```
mitogh: Using template literals.

```suggestion
	const fileOutput = `${FILE_HEADER}${ NEWLINE }${FILE_FOOTER}`;
```

bin/plugin/commands/translations.js
```diff
@@ -0,0 +1,169 @@
...
+				input: fileStream,
+			} );
+
+			const headers = [ 'Module Name', 'Description' ];
+			const translationStrings = [];
+
+			for await ( const line of rl ) {
+				headers.forEach( ( header, index ) => {
+					const regex = new RegExp(
+						'^(?:[ \t]*<?php)?[ \t/*#@]*' + header + ':(.*)$',
```
mitogh: Looks like the capturing group can be adjusted to only get valid words.

```suggestion
						'^(?:[ \t]*<?php)?[ \t/*#@]*' + header + ':\s*([\w|\s]*).*$',
```

It also removes the need to call `trim()` on the next iteration.

mitogh (changes requested):

bin/plugin/commands/translations.js
felixarntz: Updated (keeping the `output.join( NEWLINE )` though, the suggestion here is slightly off in that regard :)

bin/plugin/commands/translations.js
felixarntz: Hmm, I get the idea, but there's something off with that regex. I tried to add it, but afterwards no matches were found anymore - note that the fields may contain spaces. If that's crucial, feel free to revise.

felixarntz:
@mitogh Addressed all your feedback, except https://github.com/WordPress/performance/pull/60#discussion_r773490623 (see comment).

package.json
```diff
@@ -24,10 +24,12 @@
     "@wordpress/scripts": "^19.0",
     "chalk": "4.1.1",
     "commander": "4.1.0",
+    "fast-glob": "^3.2.7",
```
eugene-manuilov, line 27: @felixarntz, we also need to update the `package-lock.json` to make sure everybody uses the same version of this dependency.

bin/plugin/commands/translations.js
```diff
@@ -0,0 +1,168 @@
...
+		textDomain: config.textDomain,
+		directory: opt.directory || 'modules',
+		output: opt.output || 'module-i18n.php',
+	} );
+}
+
+module.exports = {
+	options,
+	handler,
+};
```
eugene-manuilov: We can use the default `exports` object that is available within the module's file-level scope instead of redefining it:

```diff
  ...

- const options = [
+ exports.options = [
      ...
  ];

  ...

- async function handler( opt ) {
+ exports.handler = async ( opt ) => {
      ...
  }

- module.exports = {
-     options,
-     handler,
- };

  ...
```

bin/plugin/commands/translations.js
```diff
@@ -0,0 +1,168 @@
...
+const TAB = '\t';
+const NEWLINE = '\n';
+const FILE_HEADER = `<?php
+/* THIS IS A GENERATED FILE. DO NOT EDIT DIRECTLY. */
+$generated_i18n_strings = array(
+`;
+const FILE_FOOTER = `
+);
+/* THIS IS THE END OF THE GENERATED FILE */
+`;
```
eugene-manuilov: Probably worth moving these constants to the top of the file, right after all imports. In this case, it will be easier to see which constants are used in the file.

bin/plugin/commands/translations.js
```diff
@@ -0,0 +1,168 @@
...
+	} );
+}
+
+module.exports = {
+	options,
+	handler,
+};
+
+const TAB = '\t';
+const NEWLINE = '\n';
```
eugene-manuilov: Node.js has `EOL` constant in the `os` module that we can use instead of defining our own:

```js
const { EOL } = require( 'os' );
```

bin/plugin/commands/translations.js
```diff
@@ -0,0 +1,168 @@
...
+			const fileStream = fs.createReadStream( moduleFile );
+
+			const rl = readline.createInterface( {
+				input: fileStream,
+			} );
+
+			const headers = [ 'Module Name', 'Description' ];
+			const translationStrings = [];
+
+			for await ( const line of rl ) {
```
eugene-manuilov: @felixarntz why do we need to use `readline` and `for await`? Why not just `const file = fs.readFileSync( moduleFile, ... )` and then `const match = regex.exec( file );`? This will let us avoid using promises.

bin/plugin/commands/translations.js
eugene-manuilov: Why not just `^\s*\*\s*${ header }:\s*(.*)$`? https://regex101.com/r/BpmfsY/1
It will help to avoid doing an additional cleanup afterwards:

```diff
 					if ( match ) {
-						const headerValue = match[ 1 ]
-							.replace( /\s*(?:\*\/|\?>).*/, '' )
-							.trim();
-						if ( headerValue ) {
-							headers.splice( index, 1 );
-							translationStrings.push( headerValue );
-						}
+						translationStrings.push( match[1] );
 					}
```

bin/plugin/commands/translations.js
```diff
@@ -0,0 +1,168 @@
...
+ * @param {WPTranslationsSettings} settings     Translations settings.
+ */
+function createTranslationsPHPFile( translations, settings ) {
+	const output = translations.map( ( translation ) => {
+		// Escape single quotes.
+		return (
+			TAB +
+			`__( '${ translation.replace( /'/g, "\\'" ) }', '${
+				settings.textDomain
+			}' ),`
```
eugene-manuilov: Perhaps would be better if we use `_x` to be able to provide an additional context for translators?

```js
_x( '${ translation.replace( /'/g, "\\'" ) }', 'Module name', '${ settings.textDomain }' ),`
```
or
```js
_x( '${ translation.replace( /'/g, "\\'" ) }', 'Module description', '${ settings.textDomain }' ),`
```

bin/plugin/commands/translations.js
```diff
@@ -0,0 +1,168 @@
...
+		)
+	);
+
+	try {
+		const translations = await getTranslations( settings );
+		createTranslationsPHPFile( translations, settings );
+	} catch ( error ) {
+		if ( error instanceof Error ) {
+			log( formats.error( error.stack ) );
+		}
```
eugene-manuilov: We need to exit the function if an error is thrown to avoid displaying the success message after the error:

```suggestion
		if ( error instanceof Error ) {
			log( formats.error( error.stack ) );
		}

		return;
```

eugene-manuilov (changes requested):

package.json
felixarntz: I may have been doing something wrong, but when I tried it, the `package-lock.json` was unchanged. It looks like the same version of this dependency was already previously required via another dependency.

bin/plugin/commands/translations.js
felixarntz: Hmm, can you please try that locally? That suggestion doesn't work for me either.

bin/plugin/commands/translations.js
felixarntz: Updated in both `translations.js` and `changelog.js`.

bin/plugin/commands/translations.js
felixarntz: Moved constants to the top for both `translations.js` and `changelog.js`.

bin/plugin/commands/translations.js
felixarntz: I notably simplified this in https://github.com/WordPress/performance/pull/60/commits/430e98f5c4345769ac74e2794ea7d53232430b51, to not use promises and also to use just a single regex for all module headers.

I've also added a `filter` call to ensure that if a file doesn't have any matches it doesn't cause an error later.

bin/plugin/commands/translations.js
felixarntz: Great idea! I've refactored this to include a context and use `_x` throughout. Also see https://github.com/WordPress/performance/pull/60/commits/848f0d2d103f897f308e2f68087eff660f171d8e, which I had to update as a result of this as well. I've clarified in a comment that this must match.

felixarntz:
@eugene-manuilov Updated based on your feedback!

mitogh (approved):
LGTM :heavy_check_mark:

ThierryA (approved):

JustinyAhin (approved):

PR #72, opened 2021-12-30
Add `readme.txt` and related update script

### LINKED ISSUE #40: Add plugin `readme.txt`

felixarntz:
Fixes #40

This PR is a starting point for the plugin's `readme.txt` file. It is a first content **proposal** for the readme itself, but in addition to that the PR includes a new related script callable via `npm run readme` that will help automate `readme.txt` updates in the future. Here's what that script does:

* Since the readme should contain an up-to-date list of all the available modules in the plugin description, the script automatically parses all module names and descriptions from the codebase and updates that bulleted list in the readme accordingly. This way we can easily always keep the readme up to date without double-maintaining these descriptions.
    * Of course we can always add more detailed information about a module somewhere else in the readme, but IMO having a quick overview list somewhere in the beginning of the plugin description is helpful, and for that purpose this script simplifies maintenance.
* The script also optionally updates the changelog for a given milestone within the readme, relying on the `npm run changelog` logic that was previously implemented in #51.

Here's how to use the script:
* `npm run readme`: This way it'll only update the module list. Try changing it to something else in the `readme.txt` file and then run this command, to see how it will always update the content to the names and descriptions of all modules within the codebase.
* `npm run readme -- -m "1.0.0-beta.1"`: This way it'll _also_ (in addition to the module list, see above) update the changelog section for the given milestone (right now we only have `1.0.0-beta.1`). If that changelog section doesn't exist yet, it'll add it right below `== Changelog ==`, otherwise it'll replace the existing section with the latest content. Try changing the content for that release to something else or add some dummy changelogs for other (non-existing) versions and then run the script to see how it always puts the latest changelog into `readme.txt`.

This PR needs two types of reviews: On the one hand, it needs engineering review and testing for the new `npm run readme` script. On the other hand, it needs content review for the initial proposal for the readme itself. Let's keep in mind that this doesn't need to be perfect or the final state of the readme (we can always add more details later), but it should at least be good for a start and a minimum foundation with which we _could_ potentially release a first version of the plugin.

### DIFF
--- a/bin/plugin/commands/common.js
+++ b/bin/plugin/commands/common.js
@@ -1,0 +1,55 @@
+/**
+ * External dependencies
+ */
+const path = require( 'path' );
+const glob = require( 'fast-glob' );
+const fs = require( 'fs' );
+
+/**
+ * @typedef WPModuleDescription
+ *
+ * @property {string} name        Module name.
+ * @property {string} description Module description.
+ */
+
+/**
+ * Returns a promise resolving to the module description list string for the `readme.txt` file.
+ *
+ * @param {string} modulesDir Modules directory.
+ *
+ * @return {Promise<[]WPModuleDescription>} Promise resolving to module description list.
+ */
+exports.getModuleDescriptions = async ( modulesDir ) => {
+	const moduleFilePattern = path.join( modulesDir, '*/*/load.php' );
+	const moduleFiles = await glob( path.resolve( '.', moduleFilePattern ) );
+
+	return moduleFiles
+		.map( ( moduleFile ) => {
+			// Map of module header => object property.
+			const headers = {
+				'Module Name': 'name',
+				Description: 'description',
+			};
+			const moduleData = {};
+
+			const fileContent = fs.readFileSync( moduleFile, 'utf8' );
+			const regex = new RegExp(
+				`^(?:[ \t]*<?php)?[ \t/*#@]*(${ Object.keys( headers ).join(
+					'|'
+				) }):(.*)$`,
+				'gmi'
+			);
+			let match = regex.exec( fileContent );
+			while ( match ) {
+				const content = match[ 2 ].trim();
+				const prop = headers[ match[ 1 ] ];
+				if ( content && prop ) {
+					moduleData[ prop ] = content;
+				}
+				match = regex.exec( fileContent );
+			}
+
+			return moduleData;
+		} )
+		.filter( ( moduleData ) => moduleData.name && moduleData.description );
+};
--- a/bin/plugin/commands/readme.js
+++ b/bin/plugin/commands/readme.js
@@ -1,0 +1,177 @@
+/**
+ * External dependencies
+ */
+const path = require( 'path' );
+const fs = require( 'fs' );
+
+/**
+ * Internal dependencies
+ */
+const { log, formats } = require( '../lib/logger' );
+const config = require( '../config' );
+const { getModuleDescriptions } = require( './common' );
+const { getChangelog } = require( './changelog' );
+
+/**
+ * @typedef WPReadmeCommandOptions
+ *
+ * @property {string=} directory Optional directory, default is the root `/modules` directory.
+ * @property {string=} milestone Optional milestone title, to update the changelog in the readme.
+ * @property {string=} token     Optional personal GitHub access token, only relevant for changelog updates.
+ */
+
+/**
+ * @typedef WPReadmeSettings
+ *
+ * @property {string}  owner     GitHub repository owner.
+ * @property {string}  repo      GitHub repository name.
+ * @property {string}  directory Modules directory.
+ * @property {string=} milestone Optional milestone title, to update the changelog in the readme.
+ * @property {string=} token     Optional personal GitHub access token, only relevant for changelog updates.
+ */
+
+exports.options = [
+	{
+		argname: '-d, --directory <directory>',
+		description: 'Modules directory',
+	},
+	{
+		argname: '-m, --milestone <milestone>',
+		description: 'Milestone title, to update the changelog',
+	},
+	{
+		argname: '-t, --token <token>',
+		description: 'GitHub token',
+	},
+];
+
+/**
+ * Command that updates the `readme.txt` file.
+ *
+ * @param {WPReadmeCommandOptions} opt
+ */
+exports.handler = async ( opt ) => {
+	await updateReadme( {
+		owner: config.githubRepositoryOwner,
+		repo: config.githubRepositoryName,
+		directory: opt.directory || 'modules',
+		milestone: opt.milestone,
+		token: opt.token,
+	} );
+};
+
+/**
+ * Returns a promise resolving to the module description list string for the `readme.txt` file.
+ *
+ * @param {WPReadmeSettings} settings Readme settings.
+ *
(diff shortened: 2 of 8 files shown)


felixarntz:
To anybody reviewing the content here, I specifically want to encourage you to update this PR directly to iterate on the content. People often shy away from updating someone else's PR, but the `readme.txt` content I've added so far is an initial **proposal** - so don't just make review suggestions, feel free to update and rewrite it as you please!

bin/plugin/commands/common.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+/**
+ * Returns a promise resolving to the module description list string for the `readme.txt` file.
+ *
+ * @param {string} modulesDir Modules directory.
+ *
+ * @return {Promise<[]WPModuleDescription>} Promise resolving to module description list.
+ */
+exports.getModuleDescriptions = async ( modulesDir ) => {
```
felixarntz, line 22: This is now in here since pretty much the same code was needed to extract the module header translations (`npm run translations` command) as it was needed to update the list of modules within the readme (the new `npm run readme` command).

bin/plugin/commands/readme.js
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * Updates the `readme.txt` file with the given module description list.
+ *
+ * @param {string} moduleList Module description list in markdown, with trailing newline.
+ */
+function updateReadmeModuleDescriptionList( moduleList ) {
+	const readmeFile = path.join( '.', 'readme.txt' );
+	const fileContent = fs.readFileSync( readmeFile, 'utf8' );
+	const newContent = fileContent.replace(
+		/(the following performance modules:\s+)((\*.*\n)+)/,
```
felixarntz, line 93: Note that this must always match the exact wording preceding the module bulleted list in the `readme.txt` file. So if you update that bit in the `readme.txt`, make sure to update it here as well.

bin/plugin/commands/readme.js
```diff
@@ -0,0 +1,177 @@
...
+const { getChangelog } = require( './changelog' );
+
+/**
+ * @typedef WPReadmeCommandOptions
+ *
+ * @property {string=} directory Optional directory, default is the root `/modules` directory.
+ * @property {string=} milestone Optional milestone title, to update the changelog in the readme.
+ * @property {string=} token     Optional personal GitHub access token, only relevant for changelog updates.
+ */
+
```
mitogh, line 22: Maybe missing the object defined above?

bin/plugin/commands/readme.js
```diff
@@ -0,0 +1,177 @@
...
+ */
+async function getModuleDescriptionList( settings ) {
+	const moduleDescriptions = await getModuleDescriptions(
+		settings.directory
+	);
+
+	return moduleDescriptions
+		.map( ( moduleData ) => {
+			return `* **${ moduleData.name }:** ${ moduleData.description }`;
+		} )
```
mitogh: The `return` can be removed from this statement.

```suggestion
		.map( ( moduleData ) => ( `* **${ moduleData.name }:** ${ moduleData.description }` ) )
```

bin/plugin/commands/readme.js
```diff
@@ -0,0 +1,177 @@
...
+ * @param {string} moduleList Module description list in markdown, with trailing newline.
+ */
+function updateReadmeModuleDescriptionList( moduleList ) {
+	const readmeFile = path.join( '.', 'readme.txt' );
+	const fileContent = fs.readFileSync( readmeFile, 'utf8' );
+	const newContent = fileContent.replace(
+		/(the following performance modules:\s+)((\*.*\n)+)/,
+		( match, prefix ) => {
+			return `${ prefix }${ moduleList }`;
+		}
```
mitogh: Similar as above, as well the `return` can be removed.

```suggestion
		( match, prefix ) => `${ prefix }${ moduleList }`
```

bin/plugin/commands/readme.js
```diff
@@ -0,0 +1,177 @@
...
+				`\n💃Updating readme.txt for "${ settings.directory }"\n\n`
+			)
+		);
+	} else {
+		log(
+			formats.title(
+				`\n💃Updating readme.txt for "${ settings.directory }" and changelog for milestone "${ settings.milestone }"\n\n`
+			)
+		);
+	}
```
mitogh: Probably a personal, preference but inverting the logic on this conditional would remove the need to have a `!` operator.


```suggestion
	if ( settings.milestone ) {
		log(
			formats.title(
				`\n💃Updating readme.txt for "${ settings.directory }" and changelog for milestone "${ settings.milestone }"\n\n`
			)
		);
	} else {
		log(
			formats.title(
				`\n💃Updating readme.txt for "${ settings.directory }"\n\n`
			)
		);
	}
```

mitogh (commented):
Nothing major, just minor stuff, overall looks great 👍

readme.txt
```diff
@@ -0,0 +1,84 @@
...
+* Translate the plugin into your language at [translate.wordpress.org](https://translate.wordpress.org/projects/wp-plugins/performance-lab)
+* Join the weekly chat (Tuesdays at 16:00 UTC) in the [#performance channel on Slack](https://wordpress.slack.com/archives/performance)
+
+== Changelog ==
+
+= 1.0.0-beta.1 =
+
+**Features**
+
+* Images: Add WebP for uploads module. ([32](https://github.com/WordPress/performance/pull/32))
```
ThierryA, line 69: The description could be a bit more user friendly, "add WebP for uploads module" probably doesn't speak much to many folks

ThierryA (approved):
I only reviewed the readme wording and took a quick pass a the code + current review by @mitogh. Minor comment, other than that LGTM.

readme.txt
JustinyAhin: I think these are generated from the PR themselves. Should we update the PR then @felixarntz?

JustinyAhin (approved):

readme.txt
felixarntz: @JustinyAhin Correct, in that case we should change the PR title.

@ThierryA Do you have a suggestion on what it could be while still being concise? I'm personally okay with it, maybe we can clarify what the module does, but I also think we should keep it short.

bin/plugin/commands/readme.js
felixarntz: Hmm, can you clarify what you mean by this?

felixarntz:
@mitogh Updated based on your feedback! Also replied to your other comments (cc @ThierryA @JustinyAhin).

bin/plugin/commands/readme.js
mitogh: Sure, I see a definition for an object `WPReadmeCommandOptions` and properties in this case `milestone` and `token` but I don't see the object itself in the doc, is this maybe document as part of the command ?

mitogh (approved):

Issue #33, opened 2021-12-02
Allow programmatic control of which modules are enabled


felixarntz:
Based on the settings screen from #27 which allows to toggle individual modules on/off via UI, it should alternatively be possible to programmatically control which modules should be enabled. This can be useful for advanced users or environments which would like to use certain modules from the plugin without confronting end users with control over them.

Suggested approach:
* Introduce a filter `perflab_active_modules` that filters the return value of the `perflab_get_active_modules()` function.
* Skip adding the plugin's settings screen (see `perflab_add_modules_page()`) if the above filter is being used (check via WP core's `has_filter()` function).

felixarntz:
cc @ThierryA I opened this based on your comment in https://github.com/WordPress/performance/issues/27#issuecomment-984407800.

eugene-manuilov:
> Suggested approach:
> 
> * Introduce a filter `perflab_active_modules` that filters the return value of the `perflab_get_active_modules()` function.
> * Skip adding the plugin's settings screen (see `perflab_add_modules_page()`) if the above filter is being used (check via WP core's `has_filter()` function).

@felixarntz, how about using the theme support feature instead of the custom filter? In this case, if someone programmatically activates a module using the module slug (for example, `add_theme_support( 'webp-uploads' )`), they won't need to re-activate it in the future when this module makes it to the core. 

I have prepared a draft PR #82 that implements this idea. Do you mind taking a look at it?

P.S: that PR contains some refactoring for the general utility functions defined in the `admin/load.php` and `load.php` files. Nothing special, just re-organised them a little bit and made it possible to use settings and module related functions not only in the admin.

eclarke1:
@eugene-manuilov can we remove the Needs Dev label from here too, as it is in Progress now? @bethanylang

eugene-manuilov:
Done.

felixarntz:
@eugene-manuilov I don't think `add_theme_support` is appropriate for this, since these modules aren't theme features. While it would work, as the name of that function states, it should be used for theme features provided by a plugin or core, which would be enabled from within the theme's codebase. But the use-case here will be most likely must-use plugins, e.g. injected by an agency or hosting provider that wants the site to use a predefined list of modules and not allow the user to change them.

The typical core approach for modifying an option like this would be a filter as suggested in the issue description.

[...]

PR #82, opened 2022-01-06
Allow programmatic control of active modules

eugene-manuilov:
* Implements the ability to activate a module using `add_theme_support( 'module-slug' )` call.
* Refactors module and settings related functions.

To test the theme support feature, add `add_theme_support( 'webp-uploads' )` to your theme's `functions.php` file. Once you add this, the module will be programmatically activated and you won't be able to disable it in the settings:

image


includes/modules.php
```diff
@@ -0,0 +1,220 @@
...
+/**
+ * Gets the active performance modules.
+ *
+ * @since 1.0.0
+ *
+ * @return array List of active module slugs.
+ */
+function perflab_get_active_modules() {
+	$module_settings = perflab_get_module_settings();
+	$all_modules     = perflab_get_modules();
```
felixarntz, line 182: This function shouldn't be used outside of the admin, intentionally, as getting all available modules is considered an "expensive" function. We don't need to know which modules exist to get the active modules, we can just look at the option. With the originally proposed approach of a filter we also don't need to. The filter would control the entire list of which modules should be active.

FYI the entire functionality here is based on how plugins work in WordPress core: Active plugins are simply read from the option, while parsing all plugins that are available is limited to the admin as it should be avoided e.g. on every page load.

felixarntz (commented):
@eugene-manuilov I left most of my feedback already in https://github.com/WordPress/performance/issues/33#issuecomment-1011022979. The main technical problem with the approach is that we should not use `perflab_get_modules` outside of the admin. And there is _way_ too much refactoring going on here, most of which isn't needed for this and in the case of the `perflab_get_modules` function for example isn't even desirable - this function should remain limited to the admin.

Can you simplify this to apply a filter in `perflab_get_active_modules`? Last but not least, as far as I can tell the second bullet point in the #33 issue description isn't covered here yet.

eugene-manuilov:
@felixarntz, I have created a new PR #87. Do you mind looking at it? Going to close this one.

PR #85, opened 2022-01-11
GitHub Action to validate flags and milestones.

eugene-manuilov:
## Summary

Addresses issue:

- #52 

## Checklist

- [x] PR has `[Focus]` and `[Type]` labels.
- [x] PR has a milestone.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,5 +11,5 @@ Addresses issue:
 
 ## Checklist
 
-- [ ] `[Focus]` and `[Type]` labels are added.
-- [ ] A milestone has been assigned.
+- [ ] PR has `[Focus]` and `[Type]` labels.
+- [ ] PR has a milestone.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,9 +1,7 @@
 ## Summary
 
-<!-- Please reference the issue this PR addresses in the following list. -->
-Addresses issue:
-
-- #
+<!-- Please reference the issue this PR fixes. If this PR does not fix the entire issue, change this to Addresses #... instead. -->
+Fixes #
 
 ## Relevant technical choices
 
@@ -11,5 +9,6 @@ Addresses issue:
 
 ## Checklist
 
-- [ ] PR has `[Focus]` and `[Type]` labels.
-- [ ] PR has a milestone.
+- [ ] PR has either `[Focus]` or `Infrastructure` label.
+- [ ] PR has a `[Type]` label.
+- [ ] PR has a milestone or the `no milestone` label.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,15 @@
+## Summary
+
+<!-- Please reference the issue this PR addresses in the following list. -->
+Addresses issue:
+
+- #
+
+## Relevant technical choices
+
+<!-- Please describe your changes. -->
+
+## Checklist
+
+- [ ] `[Focus]` and `[Type]` labels are added.
+- [ ] A milestone has been assigned.
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -17,8 +17,8 @@ jobs:
     name: Check [Type] Label
     runs-on: ubuntu-latest
     steps:
-      - if: contains( join( github.event.issue.labels.*.name, ' ' ), '[Type]' ) == false
-        run: echo "❌ Pull request must have a [Type] Label." && exit 1
+      - run: |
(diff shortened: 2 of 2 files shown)


JustinyAhin:
@eugene-manuilov I've looked at your PR. And there are a couple of things that could be improved.

Here https://github.com/WordPress/performance/blob/a2404304700e8073d09a424f51a51449d6c0b15b/.github/workflows/pr-validation.yml#L22 and here https://github.com/WordPress/performance/blob/a2404304700e8073d09a424f51a51449d6c0b15b/.github/workflows/pr-validation.yml#L30 where you check the labels, it'll be better to use a regex to match them. This will allow us not to manually update the workflow every time a label is added or removed from the repository. Same for the focus.

That's why we went with the regex solution (see this comment https://github.com/WordPress/performance/pull/53#discussion_r772646682). And it's what I've implemented here:

https://github.com/WordPress/performance/blob/408df89fa80ba834193e7fa3eae1d5050185321a/.github/workflows/pr-validation.yml#L18-L24

and here:

https://github.com/WordPress/performance/blob/408df89fa80ba834193e7fa3eae1d5050185321a/.github/workflows/pr-validation.yml#L33-L39.


Other than that, your PR looks good, and I like the PR template file idea.

eugene-manuilov:
Thanks, @JustinyAhin. I have updated it. Do you mind to take another look?

.github/workflows/pr-validation.yml
```diff
@@ -0,0 +1,36 @@
...
+env:
+  LABELS: ${{ join( github.event.pull_request.labels.*.name, ' ' ) }}
+
+jobs:
+  check-type-label:
+    name: Check [Type] Label
+    runs-on: ubuntu-latest
+    steps:
+      - if: contains( env.LABELS, '[Type]' ) == false
+        run: exit 1
```
JustinyAhin, line 24: @eugene-manuilov do you think we can add a message with the exit?

JustinyAhin:
@eugene-manuilov I've added a comment and made a few tests, and the PR works fine.

.github/workflows/pr-validation.yml
eugene-manuilov: @JustinyAhin I think echoing messages is redundant for those checks since it's clear what is wrong by looking at the title of the job.

JustinyAhin: Make sense yes.

JustinyAhin (approved):
I think we can merge the PR @eugene-manuilov

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
+## Summary
+
+<!-- Please reference the issue this PR addresses in the following list. -->
+Addresses issue:
+
+- #
```
felixarntz: I'm not sure this is exactly how we want the format here. For example, most PRs fix an issue, and to simplify maintenance it may be a better approach to have _Fixes #..._ by default. Potentially there could be a comment above that includes "If this PR does not fix the entire issue, change this to _Addresses #..._ instead."

Anyway, I think this needs a bit more definition and doesn't really belong in here.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Relevant technical choices
+
+<!-- Please describe your changes. -->
+
+## Checklist
+
+- [ ] PR has `[Focus]` and `[Type]` labels.
+- [ ] PR has a milestone.
```
felixarntz: These are not entirely accurate now.

* We should clarify that instead of `[Focus] ...`, it's also okay to use the `Infrastructure` label. 
* We should clarify that instead of adding a milestone for a release, it's also okay to use the `no milestone` label if this PR/issue are irrelevant of any release (e.g. not part of the release plugin ZIP).

Similar to the above, maybe it's a bit preliminary to include this here.

felixarntz (commented):
@eugene-manuilov The GitHub workflow looks great!
However, I think adding a PR template here is a bit out of scope and preliminary, also see my comments below. I'd prefer if we removed it and discussed/added it via a separate issue.

.github/PULL_REQUEST_TEMPLATE.md
eugene-manuilov: Updated.

.github/PULL_REQUEST_TEMPLATE.md
eugene-manuilov: Updated.

eugene-manuilov:
> However, I think adding a PR template here is a bit out of scope and preliminary, also see my comments below. I'd prefer if we removed it and discussed/added it via a separate issue.

@felixarntz I have added the pull request template to be able to trigger milestone checks manually since there is no an appropriate activity type for the `pull_request` event. I have updated the template to address your feedback. If you still think it should be added via a separate ticket, I'll remove it from my PR and add a new ticket for it. Please, let me know what you think.

felixarntz:
@eugene-manuilov 
> I have added the pull request template to be able to trigger milestone checks manually since there is no an appropriate activity type for the `pull_request` event.

Thanks for clarifying, that makes sense.

felixarntz (approved):
Great!

PR #87, opened 2022-01-12
Introduce `perflab_active_modules` filter to control which modules are active

eugene-manuilov:
Addresses #33

### DIFF
--- a/load.php
+++ b/load.php
@@ -95,7 +95,7 @@ function perflab_get_active_modules() {
 	/**
 	 * Filters active modules to allow programmatically control which modules are active.
 	 *
-	 * @since n.e.x.t
+	 * @since 1.0.0
 	 *
 	 * @param array An array of the currently active modules.
 	 */
--- a/load.php
+++ b/load.php
@@ -83,7 +83,7 @@ function perflab_get_module_settings() {
  * @return array List of active module slugs.
  */
 function perflab_get_active_modules() {
-	return array_keys(
+	$modules = array_keys(
 		array_filter(
 			perflab_get_module_settings(),
 			function( $module_settings ) {
@@ -91,6 +91,10 @@ function perflab_get_active_modules() {
 			}
 		)
 	);
+
+	$modules = apply_filters( 'perflab_active_modules', $modules );
+
+	return $modules;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


load.php
```diff
@@ -83,14 +83,25 @@ function perflab_get_module_settings() {
...
 			function( $module_settings ) {
 				return isset( $module_settings['enabled'] ) && $module_settings['enabled'];
 			}
 		)
 	);
+
+	/**
+	 * Filters active modules to allow programmatically control which modules are active.
+	 *
+	 * @since n.e.x.t
```
felixarntz: We don't have the `n.e.x.t` convention here so far, but it's worth evaluating. Do you mind opening an issue where we can propose/discuss it? It would be important to document this somewhere eventually. We haven't published a release yet, but once we get to that, we'll need to have documentation anyway.

Maybe we could even have a `npm` script that replaces all `n.e.x.t` occurrences with a specific version.

felixarntz (approved):
@eugene-manuilov Looks great!

I left one comment below that doesn't block this from being merged, but we'll need to keep this in mind. Maybe it's better for now to put `1.0.0` like we have elsewhere. However, I think it'd be great to have an issue bringing up the `n.e.x.t` proposal.

load.php
eugene-manuilov: Updated and added a new ticket: #90

felixarntz (approved):
Re-approving for the additional (mini-)change.

tillkruss (approved):

JustinyAhin:
@eugene-manuilov could provide a snippet of how to test the filter, in a mu-plugin for instance? Thanks :)

aristath (approved):

tillkruss:
> @eugene-manuilov could provide a snippet of how to test the filter, in a mu-plugin for instance? Thanks :)

```php
add_filter( 'perflab_active_modules', function ( $modules ) {
    return array_diff( $modules, [ "foo", "bar" ] ); // modules you want to remove
} );

add_filter( 'perflab_active_modules', function ( $modules ) {
    return array_merge( $modules, [ "foo", "bar" ] ); // modules you want to add
} );
```

JustinyAhin:
Thanks, @tillkruss

JustinyAhin (approved):

PR #89, opened 2022-01-14
GitHub action for automated plugin deployment to wordpress.org

### LINKED ISSUE #42: Configure GitHub action for automated plugin deployment to wordpress.org
Once the plugin is released on the wordpress.org plugin repository, we will need to regularly push updates. Since this is somewhat of a tedious and potentially error-prone process, it should be automated: We need a GitHub action that deploys the plugin release to wordpress.org when a GitHub release is published.

Some potential approaches / inspiration:
* https://github.com/10up/action-wordpress-plugin-deploy
* https://github.com/WordPress/gutenberg/blob/trunk/.github/workflows/upload-release-to-plugin-repo.yml

We don't need this _yet_, but it certainly doesn't hurt starting to already think about and discuss how we should go about it.

JustinyAhin:
Fixes #42.

This PR implements a GitHub action for deploying the performance plugin to W.org.

It uses 10up https://github.com/10up/action-wordpress-plugin-deploy to deploy on new releases on the repository.

The action also uses the `SVN_PASSWORD` and `SVN_USERNAME` secrets (not set yet in the repository) for the credentials.

Next steps:

- [x] Attach a zip file to the release
This is usually done using this action https://github.com/actions/upload-release-asset (see https://github.com/marketplace/actions/wordpress-plugin-deploy#deploy-on-publishing-a-new-release-and-attach-a-zip-file-to-the-release).


JustinyAhin:
https://github.com/actions/upload-release-asset is not maintained any more. Worth trying with https://github.com/softprops/action-gh-release#%EF%B8%8F-uploading-release-assets instead.

.github/workflows/deploy-dotorg.yml
```diff
@@ -0,0 +1,21 @@
...
+      - name: WordPress plugin deploy
+        uses: 10up/action-wordpress-plugin-deploy@stable
+        env:
+          SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
+          SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+      - name: Upload release assets
+        uses: softprops/action-gh-release@v0.1.14
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          files: ${{github.workspace}}/${{ github.event.repository.name }}.zip
```
felixarntz, line 23: I think we should probably use a custom "ZIP build" workflow, since there's a bunch of things in the GitHub repository that shouldn't be in the plugin ZIP. At a minimum, I think we should have a `.gitattributes` file (or `.distignore`) that ensures the ZIP only includes what we need in a production ZIP (e.g. no tests, no docs, no development tooling etc.). See also https://github.com/10up/action-wordpress-plugin-deploy#sample-baseline-files

.github/workflows/deploy-dotorg.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  release:
+    name: New Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: WordPress plugin deploy
+        uses: 10up/action-wordpress-plugin-deploy@stable
+        env:
+          SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
+          SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
```
felixarntz, line 17: We'll also need to specify `SLUG: performance-lab` here, since this is the actual plugin slug, but the repo here is just named "performance". See https://github.com/10up/action-wordpress-plugin-deploy#optional-environment-variables

felixarntz (commented):
@JustinyAhin This looks solid for a start, just two things that I think we need to improve for the first version.

.github/workflows/deploy-dotorg.yml
JustinyAhin: Added a `.distignore` file https://github.com/WordPress/performance/blob/add/42-actions-automatic-deployment-wordpressdotorg/.distignore

.distignore
```diff
@@ -0,0 +1,30 @@
...
+/composer.lock
+/package.json
+/package-lock.json
+/phpcs.xml.dist
+/phpunit.xml.dist
+
+/CODE_OF_CONDUCT.md
+/CONTRIBUTING.md
+/SECURITY.md
+/LICENSE
```
felixarntz: I think we need to remove this file from here since every wordpress.org plugin needs to include a license file.

felixarntz (commented):
@JustinyAhin All good now from my end, except one more tiny thing.

JustinyAhin:
Now we will need to decide whether we will use a dedicated w.org account, or one of the committer's account to deploy the plugin to WordPress. Can we mention this during the next meeting @bethanylang?

See https://github.com/WordPress/performance/issues/42#issuecomment-1000428780.

.github/workflows/deploy-dotorg.yml
```diff
@@ -0,0 +1,22 @@
+name: Deploy to WordPress.org
+
+on:
+  release:
+    types: [published]
```
eugene-manuilov: I think we should listen to the `released` type instead, since `published` is not triggered when a pre-release is released. See [here](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-38): 

> * `published`: a release, pre-release, or draft of a release is published
> * `released`: a release or draft of a release is published, or a pre-release is changed to a release

```suggestion
    types:
      - released
```

eugene-manuilov (changes requested):

.github/workflows/deploy-dotorg.yml
JustinyAhin: Good catch @eugene-manuilov. Thanks.

aristath (approved):
Great work here!
This will definitely make things easy for releases 👍

felixarntz (approved):
Great work @JustinyAhin!

eugene-manuilov (approved):
LGTM too.

PR #91, opened 2022-01-17
Update CODEOWNERS to use the new module naming convention

### LINKED ISSUE #23: Set up code owners
As per @felixarntz suggestion in #2, should we set up a [`CODEOWNERS`](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) file to automatically assign issues and PRs?

It would help reduce the noise.

JustinyAhin:
Fixes #23. Follows up on https://github.com/WordPress/performance/issues/23#issuecomment-1013140822.

This PR update the CODEOWNERS file to reflect the module naming convention. It also updates the code owners based on the latest changes on [the focus spreadsheet](https://docs.google.com/spreadsheets/d/16N5oZ9wE6AkiqMz7b_707eh24vvpjMwsEG67XFAbxy8/).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,7 @@
 # Documentation/Infrastructure
 /docs                                       @justinahinon @felixarntz
 /admin                                      @justinahinon @felixarntz
+/tests/admin                                @justinahinon @felixarntz
 /bin                                        @justinahinon @felixarntz
 /.github                                    @justinahinon @felixarntz
 
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,15 +5,21 @@
 /.github                                    @justinahinon @felixarntz
 
 # Focus: Images
-/modules/webp-uploads                       @adamsilverstein
-/tests/modules/webp-uploads                 @adamsilverstein
+/modules/images                             @adamsilverstein
+/tests/modules/images                       @adamsilverstein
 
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -5,15 +5,21 @@
 /.github                                    @justinahinon @felixarntz
 
 # Focus: Images
-/modules/webp-uploads                       @adamsilverstein
-/tests/modules/webp-uploads                 @adamsilverstein
+/modules/images                             @adamsilverstein
+/tests/modules/images                       @adamsilverstein
```
felixarntz, line 10: Maybe in addition to those two locations, we should also add `/tests/testdata/modules/{focus}` here for each module? That's the third focus-specific location, which is more minor but probably still worth capturing here for completeness.

felixarntz (approved):
@JustinyAhin Looks good, just one tiny suggestion to "technically" complete this.

.github/CODEOWNERS
JustinyAhin: That makes sense, yes.

JustinyAhin:
One last thing @felixarntz. Should I add you as owner for `/tests/admin`?

felixarntz:
@JustinyAhin Yeah, probably best if you add both of us since that should IMO be the same as the production code for that, which falls into _Infrastructure_.

Issue #92, opened 2022-01-17
Create Site Health Audit for Autoloaded options


manuelRod:
Create a new module for site health to audit and measure heavy autoloaded data usage from the options table.

We could discuss different approaches for this or we can create and evolve the module further in different steps.

1. Informative module, we just gather the information and warn the user if needed (ex, you have more than X autoloaded options), we show them useful information of what it means or how it can be solved ([exmple](https://kinsta.com/knowledgebase/wp-options-autoloaded-data/) )
2. Functional module, we can go a step further and keep track of used/unused options, and then give the ability to users to clean up the table from the site health screen itself.

What do you think?

@spacedmonkey @pbearne

mehigh:
❤️ 
If you don't know about the problem, you won't be lifting a finger to fix it. This is a great idea!

pbearne:
I have been working on some filters that would help provide detailed stats

yogeshbeniwal:
Nice idea, may be rather than X autoload options, we can keep it as more than x MB size of autoload.

> 1. you have more than X autoloaded options

tillkruss:
> Nice idea, may be rather than X autoload options, we can keep it as more than x MB size of autoload.

Yeah, byte size and total count would be very useful.

manuelRod:
should we define the thresholds?

mehigh:
750KB for warning (yellow background), 1MB+ for error (red background)

manuelRod:
From WPEngine

> While autoloaded data is useful in some cases, it is simply not necessary for most information to be load on every single request. It’s recommended to keep your total autoloaded data below 800,000 bytes (0.8Mb) for optimal performance.

From Kinsta

> How much is too much-autoloaded data? This can vary of course, but ideally, you want this to be between 300 KB to 1MB. Once you start approaching the 3-5 MB range or more, there are most likely things that can be optimized or removed from being autoloaded. And anything above 10 MB should be addressed right away. This doesn’t always mean it’s going to cause an issue, but it’s a good place to start.

spacedmonkey:
I think whatever values we have for sizes would need a filter. It is made clear by these quotes, different hosts have different levels of acceptable autoload sizes.

tillkruss:
@manuelRod: I've got a PR nearly ready for #5 that almost identical to this. What would you suggest to the user if the autoloaded options are too big, an object cache, or to reduce it?

From @spacedmonkey:

[...]

PR #111, opened 2022-01-19
Add Site Health check module for persistent object cache

### LINKED ISSUE #35: Promote object caching in Site Health
Recommend the use of an object cache under Site Health for sites with non-trivial amounts of data.

- Detect available extensions and make recommendations
  - [See AMP](https://github.com/ampproject/amp-wp/blob/5405daa38e65f0ec16ffc920014d0110b03ee773/src/Admin/SiteHealth.php#L227-L330)
- What are the criteria for running the check?
- Allow WooCommerce (etc.) to enable running the check?

>Would you see this as being something that checks for APCu, or a reachable memcache / redis install?

tillkruss:
## Summary

This PR adds the `object-cache/health` module containing a single health check to suggest the usage of a persistent object cache backend, if needed.

The module adds the direct `persistent_object_cache` health check.

- Plugins (like WooCommerce) can use the `site_status_suggest_persistent_object_cache` filter to inject their own criteria checks to suggest the use of an object cache, as well as bypass the default core health check
- Hosts may use the `site_status_persistent_object_cache_url` filter to link to their own support docs
- Hosts may want to use the `site_status_persistent_object_cache_notes` filter to recommend their preferred object caching solution
- Plugin authors may want to append notes using the `site_status_persistent_object_cache_notes` filter on why object caching is recommended for their plugin

Resolves #35.


modules/object-cache/health/load.php
```diff
@@ -0,0 +1,225 @@
...
+function oc_health_persistent_object_cache() {
+	/**
+	 * Filter the action URL for the persistent object cache health check.
+	 *
+	 * @param string $action_url Learn more link for persistent object cache health check.
+	 */
+	$action_url = apply_filters(
+		'site_status_persistent_object_cache_url',
+		/* translators: Localized Support reference. */
+		__( 'https://wordpress.org/support/article/optimization/#object-caching', 'performance-lab' )
```
eugene-manuilov: There is no `object-caching` section on the Optimization page. Is it yet to be added? :thinking: If not, then we probably need to update it to something like `https://wordpress.org/support/article/optimization/#server-side-caching`?

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,225 @@
...
+		'posts_count'      => 1000,
+		'terms_count'      => 1000,
+		'users_count'      => 1000,
+	);
+
+	$alloptions = wp_load_alloptions();
+
+	if ( $thresholds['alloptions_count'] < count( $alloptions ) ) {
+		return true;
+	}
```
eugene-manuilov: This will always be `true` with the `alloptions_count` threshold equal `1`. Do we need to change it to equal at least `100`?

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,225 @@
...
+
+	if ( $thresholds['alloptions_bytes'] < strlen( serialize( wp_load_alloptions() ) ) ) {
```
eugene-manuilov: 1. All options are already loaded, so we can use `$alloptions` instead of calling the `wp_load_alloptions` function once again.
2. We should use `mb_strlen( ..., '8bit' )` to get the size of the serialized data more accurately.

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,225 @@
...
+ * @return array The list of available persistent object cache services.
+ */
+function oc_health_available_persistent_object_cache_services() {
+	$extensions = array_map(
+		'extension_loaded',
+		array(
+			'Redis'     => 'redis',
+			'Relay'     => 'relay',
+			'Memcached' => 'memcache', // The `memcached` extension seems unmaintained.
+			'APCu'      => 'apcu',
```
eugene-manuilov: Memcached, Relay (to some degree), and APCu are **impersistent** in-memory object cache storages. Do we really need to suggest only **persistent** object cache storages?

tillkruss: In WordPress `persistent` means longer than the request. Redis can persist data between restarts JFYI.

modules/object-cache/health/load.php
tillkruss: Great catch.

modules/object-cache/health/load.php
tillkruss: This was a debug value. Good catch!

modules/object-cache/health/load.php
tillkruss: Yeah, it's mentioned in the PR description as a TODO.

modules/object-cache/health/load.php
tillkruss: @eugene-manuilov Can you elaborate why 8bit strlen would be more accurate with serialized data?

eugene-manuilov: > Can you elaborate why 8bit strlen would be more accurate with serialized data?

@tillkruss `8bit` is the same as `binary` encoding which means that `mb_strlen` will calculate every byte of a string and return the actual memory size taken by it. But now I realise that this is the default behaviour of the `strlen` function, so no changes are needed then 🙈, thus, please, disregard the 2nd point.

modules/object-cache/health/load.php
eugene-manuilov: > In WordPress `persistent` means longer than the request.

This is confusing since in non-WordPress world it has a different meaning. Maybe we should remove `persistent` from the function name and use just `oc_health_available_object_cache_services` to avoid confusion? Anyway the filter that is applied to the returned services list doesn't have `persistent` keyword either.

> Redis can persist data between restarts JFYI.

Yes, i know it. That's why Redis considered as persistent object cache storage.

modules/object-cache/health/load.php
tillkruss: Thanks, good to know it's the default behaviour!

pbearne (approved):

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	/**
+	 * Filter the second paragraph of the health check's description
+	 * when suggesting the use of a persistent object cache.
+	 *
+	 * Hosts may want to replace the notes to recommend their preferred object caching solution.
+	 *
+	 * Plugin authors may want to append notes (not replace) on why object caching is recommended for their plugin.
+	 *
```
spacedmonkey: Add a since comment here.

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,225 @@
...
+	 * @param string $description
+	 * @param array $available_services
```
spacedmonkey: Missing descriptions.

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,225 @@
...
+ *
+ * @param mixed $should_suggest Whether to suggest using a persistent object cache.
+ * @return bool
+ */
+function oc_health_should_persistent_object_cache( $should_suggest ) {
+	global $wpdb;
+
+	$thresholds = array(
+		'alloptions_count' => 500,
+		'alloptions_bytes' => 100000,
```
spacedmonkey: Does this step on the toes of #92 CC @manuelRod @pbearne

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,225 @@
...
+ *
+ * Determines whether to suggest using a persistent object cache.
+ *
+ * @param mixed $should_suggest Whether to suggest using a persistent object cache.
+ * @return bool
+ */
+function oc_health_should_persistent_object_cache( $should_suggest ) {
+	global $wpdb;
+
+	$thresholds = array(
```
spacedmonkey: Should we have a filter for these values?

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,225 @@
...
+ * Module Name: Health Check
+ * Description: Recommend a persistent object cache for sites with non-trivial amounts of data.
+ * Experimental: No
+ *
+ * @package performance-lab
+ * @since 1.0.0
+ */
+
+add_filter( 'site_status_tests', 'oc_health_add_tests' );
+add_filter( 'site_status_persistent_object_cache', 'oc_health_should_persistent_object_cache' );
```
spacedmonkey: Move these to line after the function like this

```
function oc_health_add_tests( $tests ) {
	$tests['direct']['persistent_object_cache'] = array(
		'label' => 'persistent_object_cache',
		'test'  => 'oc_health_persistent_object_cache',
	);

	return $tests;
}
add_filter( 'site_status_tests', 'oc_health_add_tests' );
```

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,225 @@
...
+	);
+
+	$services = array_keys( array_filter( $extensions ) );
+
+	/**
+	 * Filter the persistent object cache services available to the user.
+	 *
+	 * This can be useful to hide or add services not included in the defaults.
+	 *
+	 * @param array $suggest
```
spacedmonkey: Missing description

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,225 @@
...
+		)
+	);
+
+	$services = array_keys( array_filter( $extensions ) );
+
+	/**
+	 * Filter the persistent object cache services available to the user.
+	 *
+	 * This can be useful to hide or add services not included in the defaults.
+	 *
```
spacedmonkey: Add since

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,225 @@
...
+		return true;
+	}
+
+	if ( $thresholds['terms_count'] < $results[ $wpdb->terms ]->rows ) {
+		return true;
+	}
+
+	if ( $thresholds['users_count'] < $results[ $wpdb->users ]->rows ) {
+		return true;
+	}
```
spacedmonkey: If we tweak the logic here, we make this into a loop. 

Like this 

```
$mapping = array(
         'comments_count'   => array( 'table' => $wpdb->comments, 'threshold' => 1000),
		'options_count'    => array( 'table' => $wpdb->options, 'threshold' => 1000),
		'posts_count'      => array( 'table' => $wpdb->posts, 'threshold' => 1000),
		'terms_count'      => array( 'table' => $wpdb->terms, 'threshold' =>  1000),
		'users_count'      => array( 'table' => $wpdb->users, 'threshold' =>  1000),
	);
```

spacedmonkey:
@tillkruss Can you explain what I am meant to see in site health. I checked two sites and I see any new options in site health?

Can you provide details to test this and maybe a screenshot?

modules/object-cache/health/load.php
tillkruss: In a sense yes, but #92 only warns the user about it.

This module explicitly recommends and object cache to the user.

tillkruss:
For testing:

- Ensure that no `object-cache.php` drop-in is enabled on the test site
- Go to `Settings > Performance` and enable this module
- Then go to `Tools > Site Health` to see the recommendation

If the test site doesn't have much data to trigger the suggestion, lower the post threshold:

```php
add_filter( 'site_status_persistent_object_cache_thresholds', function ( $thresholds ) {
    return array_merge( $thresholds, [ 'posts_count' => 1 ] );
} );
```

spacedmonkey:
Should this message show on all environments? In WordPress 5.5, the `wp_get_environment_type` was added. This message should only really show on 'production'? I think it is okay if there is no object caching on staging / development / local.

tillkruss:
@felixarntz: This is ready to be merged.

spacedmonkey (changes requested):
Some level of unit tests would be useful before merge. 

I would recommend testing `oc_health_should_persistent_object_cache` at the very least.

tillkruss:
> I would recommend testing `oc_health_should_persistent_object_cache` at the very least.

I added tests for it 👍

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,244 @@
...
+
+/**
+ * Adds a health check testing for and suggesting a persistent object cache backend.
+ *
+ * @since 1.0.0
+ *
+ * @param array $tests An associative array of direct and asynchronous tests.
+ * @return array
+ */
+function oc_health_add_tests( $tests ) {
```
felixarntz: Nit-pick, but let's use the `perflab_` prefix for all functions to have the plugin-specific prefix. I'd suggest using `perflab_oc_*`, to reduce the chance of conflicts with other plugins, but also with other modules in the performance plugin.

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,244 @@
...
+ * @return array
+ */
+function oc_health_persistent_object_cache() {
+	/**
+	 * Filter the action URL for the persistent object cache health check.
+	 *
+	 * @param string $action_url Learn more link for persistent object cache health check.
+	 */
+	$action_url = apply_filters(
+		'site_status_persistent_object_cache_url',
```
felixarntz: Related to the above, let's prefix all hooks with `perflab_oc_` as well.

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,244 @@
...
+ * Callback for `persistent_object_cache` health check.
+ *
+ * @since 1.0.0
+ *
+ * @return array
+ */
+function oc_health_persistent_object_cache() {
+	/**
+	 * Filter the action URL for the persistent object cache health check.
+	 *
```
felixarntz: This needs a `@since` annotation.

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,244 @@
+<?php
+/**
+ * Module Name: Health Check
```
felixarntz: Two notes here:
* Let's always use a module slug based on the "sanitized" name of the module: For example since this is called "Health Check", the module slug/directory should be `health-check`.
* I suggest to make the name (and per the above also the slug) a bit more precise, e.g. "Persistent Object Cache Health Check".

Overall I'm still unsure whether this should be in the Object cache focus or the Site Health focus. IMO the latter is more suitable, since this doesn't improve anything around object caching on the WordPress site, it just informs about the state of object caching - which IMO makes it more suitable alongside other Site Health audits, in that focus area.

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,244 @@
...
+ * @since 1.0.0
+ */
+
+/**
+ * Adds a health check testing for and suggesting a persistent object cache backend.
+ *
+ * @since 1.0.0
+ *
+ * @param array $tests An associative array of direct and asynchronous tests.
+ * @return array
```
felixarntz: Missing return description.

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,244 @@
...
+	return $tests;
+}
+add_filter( 'site_status_tests', 'oc_health_add_tests' );
+
+/**
+ * Callback for `persistent_object_cache` health check.
+ *
+ * @since 1.0.0
+ *
+ * @return array
```
felixarntz: Missing return description.

tests/modules/object-cache/health-check/health-check-test.php
```diff
@@ -0,0 +1,142 @@
+<?php
+/**
+ * Tests for webp-uploads module.
+ *
+ * @package performance-lab
+ * @group webp-uploads
```
felixarntz: Not the right group :)

Let's use the module slug.

modules/object-cache/health/load.php
```diff
@@ -0,0 +1,244 @@
...
+			__( "WordPress performs at its best when a persistent object cache is used. A persistent object cache helps to reduce load on your SQL server significantly and allows WordPress to retrieve your site's content and settings much faster.", 'performance-lab' )
+		),
+		'actions'     => sprintf(
+			'<p><a href="%s" target="_blank" rel="noopener">%s <span class="screen-reader-text">%s</span><span aria-hidden="true" class="dashicons dashicons-external"></span></a></p>',
+			esc_url( $action_url ),
+			__( 'Learn more about persistent object caching.', 'performance-lab' ),
+			/* translators: Accessibility text. */
+			__( '(opens in a new tab)', 'performance-lab' )
+		),
+	);
```
felixarntz: This may be a personal preference, but IMO the logic in this function is a bit hard to follow. For example, why not include all fields in their "default" version in this `$result` variable right away? It's a bit confusing and error-prone that the `label` key is set below, I think it would be more appropriate to have all fields in some "default" version set right here, and then the conditions below should only adjust their values as needed.

felixarntz (commented):
@tillkruss This looks good at a high level, I left a few comments, mostly minor.

One higher-level comment is that I'm not sure we need all these filters -some of them seem a bit excessive to have. I suggest we keep filters minimal and only add them based on demand. What would be the use cases for all of these filters?

tillkruss:
> Should this message show on all environments? In WordPress 5.5, the `wp_get_environment_type` was added. This message should only really show on 'production'? I think it is okay if there is no object caching on staging / development / local.

@spacedmonkey Addressed in https://github.com/WordPress/performance/pull/111/commits/aca709ad22671413391883dcf67aedd360165308

tillkruss:
@felixarntz

I've added missing `@since` tags in https://github.com/WordPress/performance/pull/111/commits/80d8a8c6f4561a8bdd437c38f556f56bbafe99fc.

And documented return values in https://github.com/WordPress/performance/pull/111/commits/25e040cdb469ac215783741be6576c9a9940edcb.

modules/object-cache/health/load.php
tillkruss: I've renamed the module to `health-check` and updated the module name in https://github.com/WordPress/performance/pull/111/commits/2e3816950af4a4a1ddea778ee390c38ca51cdb3c.

Did you mean the slug should be `persistent-object-cache-health-check`? I did that in https://github.com/WordPress/performance/pull/111/commits/831c85ce6144314002636708c77c5b06554e55b6.

tillkruss: >Overall I'm still unsure whether this should be in the Object cache focus or the Site Health focus. IMO the latter is more suitable, since this doesn't improve anything around object caching on the WordPress site,

At this point I don't think we'll have any object caching modules, since all improvements are so far have taken place in core itself. I'll see if there are non-core issues.

>it just informs about the state of object caching - which IMO makes it more suitable alongside other Site Health audits, in that focus area.

I agree with that, however IMO the _Object Cache_ team should be responsible for the code. I'm thinking code-ownership at this point.

Maybe we move this into _Site Health_ once we it's stable and ready for merge into 6.0? There is still #161 to tackle, likely as part of this.

modules/object-cache/health/load.php
tillkruss: Done in https://github.com/WordPress/performance/pull/111/commits/c992a7c81b97d072e276c0e259d7ff11239d2601.

modules/object-cache/health/load.php
tillkruss: Fixed all `@since` tags in https://github.com/WordPress/performance/commit/80d8a8c6f4561a8bdd437c38f556f56bbafe99fc.

modules/object-cache/health/load.php
tillkruss: Added docs for all `@return` directives in https://github.com/WordPress/performance/commit/25e040cdb469ac215783741be6576c9a9940edcb.

tests/modules/object-cache/health-check/health-check-test.php
tillkruss: Fixed in https://github.com/WordPress/performance/pull/111/commits/831c85ce6144314002636708c77c5b06554e55b6.

modules/object-cache/health/load.php
tillkruss: Prefixed all filters in https://github.com/WordPress/performance/pull/111/commits/3502b2821840ffb329975b2461dcdd303611db45.

modules/object-cache/health/load.php
tillkruss: I'm not sure I understand, could you elaborate or give a pseudo code example?

I followed what some other core health checks are doing.

modules/object-cache/health/load.php
felixarntz: Yeah I think the module name is better now. 👍 

I don't have a strong opinion on the focus area, I agree with your point on code ownership that it makes sense to have it in the Object Cache team.

modules/object-cache/health/load.php
felixarntz: I mean that, just in the one definition of the `$result` variable (https://github.com/WordPress/performance/pull/111/files#diff-954eabafd12fa670f1c78ae66a34667c831db04ac1984af3ef1591fbc21ebe2eR52-R70), all relevant fields should be included already, for easier overview. Right now the code does not include the required `label` field, it only gets set later. When I reviewed your code I first thought "wait why is there no `label`".

I think it's better to include all fields in the initial `$result =` assignment, in one of the potential scenarios that makes sense.

modules/object-cache/persistent-object-cache-health-check/load.php
```diff
@@ -0,0 +1,252 @@
...
+			/* translators: Accessibility text. */
+			__( '(opens in a new tab)', 'performance-lab' )
+		),
+	);
+
+	if ( wp_using_ext_object_cache() ) {
+		$result['label'] = __( 'A persistent object cache is being used', 'performance-lab' );
+
+		return $result;
+	}
```
felixarntz: This is what I meant in https://github.com/WordPress/performance/pull/111#discussion_r807420816:
```suggestion
	$result = array(
		'test'        => 'persistent_object_cache',
		'status'      => 'good',
		'badge'       => array(
			'label' => __( 'Performance', 'performance-lab' ),
			'color' => 'blue',
		),
		'label'       => __( 'A persistent object cache is being used', 'performance-lab' ),
		'description' => sprintf(
			'<p>%s</p>',
			__( "WordPress performs at its best when a persistent object cache is used. A persistent object cache helps to reduce load on your SQL server significantly and allows WordPress to retrieve your site's content and settings much faster.", 'performance-lab' )
		),
		'actions'     => sprintf(
			'<p><a href="%s" target="_blank" rel="noopener">%s <span class="screen-reader-text">%s</span><span aria-hidden="true" class="dashicons dashicons-external"></span></a></p>',
			esc_url( $action_url ),
			__( 'Learn more about persistent object caching.', 'performance-lab' ),
			/* translators: Accessibility text. */
			__( '(opens in a new tab)', 'performance-lab' )
		),
	);

	if ( wp_using_ext_object_cache() ) {
		return $result;
	}
```

modules/object-cache/persistent-object-cache-health-check/load.php
```diff
@@ -0,0 +1,252 @@
...
+	$available_services = perflab_oc_health_available_object_cache_services();
+
+	$notes = __( 'Speak to your web host about what persistent object caches are available and how to enable them.', 'performance-lab' );
+
+	if ( ! empty( $available_services ) ) {
+		$notes .= sprintf(
+			/* translators: Available object caching services. */
+			__( ' Your host appears to support the following object caching services: %s.', 'performance-lab' ),
+			implode( ', ', $available_services )
+		);
```
felixarntz: Translation strings shouldn't have any trailing spaces, as those get easily missed.
```suggestion
		$notes .= ' ' . sprintf(
			/* translators: Available object caching services. */
			__( 'Your host appears to support the following object caching services: %s.', 'performance-lab' ),
			implode( ', ', $available_services )
		);
```

modules/object-cache/persistent-object-cache-health-check/load.php
```diff
@@ -0,0 +1,252 @@
...
+	 * when suggesting the use of a persistent object cache.
+	 *
+	 * Hosts may want to replace the notes to recommend their preferred object caching solution.
+	 *
+	 * Plugin authors may want to append notes (not replace) on why object caching is recommended for their plugin.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param string $notes The notes appended to the health check description.
+	 * @param array $available_services The list of available persistent object cache services.
```
felixarntz:
```suggestion
	 * @param string $notes              The notes appended to the health check description.
	 * @param array  $available_services The list of available persistent object cache services.
```

felixarntz (commented):
@tillkruss LGTM, just a few last nit-picks.

tillkruss:
@felixarntz: All done. Any filter you'd like me to remove?

spacedmonkey (approved):
Good first PR.

modules/object-cache/persistent-object-cache-health-check/load.php
```diff
@@ -0,0 +1,251 @@
...
+	/**
+	 * Filter whether to suggest using a persistent object cache.
+	 *
+	 * Plugin and theme authors should NOT use this filter to discourage the use of an object cache.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param bool $suggest Whether to suggest using a persistent object cache.
+	 */
+	if ( ! apply_filters( 'perflab_oc_site_status_suggest_persistent_object_cache', false ) ) {
```
felixarntz: I think this filter would be better to remove, and instead we can call our implementation below directly. The `perflab_oc_site_status_persistent_object_cache_thresholds` filter can already be used to more granularly control the same thing, and seeing that `Plugin and theme authors should NOT use this filter to discourage the use of an object cache.` makes me think even more we should get rid of it.
```suggestion
	if ( ! perflab_oc_health_should_persistent_object_cache() ) {
```

modules/object-cache/persistent-object-cache-health-check/load.php
```diff
@@ -0,0 +1,251 @@
...
+ * @param mixed $should_suggest Whether to suggest using a persistent object cache.
+ * @return bool Whether to suggest using a persistent object cache.
+ */
+function perflab_oc_health_should_persistent_object_cache( $should_suggest ) {
+	global $wpdb;
+
+	// Bypass expensive calls if `$should_suggest` was already set truthy.
+	if ( $should_suggest ) {
+		return true;
+	}
```
felixarntz: Based on my above comment, let's make this a simple "check" function instead of a filter callback.
```suggestion
/**
 * Determines whether to suggest using a persistent object cache.
 *
 * @since 1.0.0
 *
 * @return bool Whether to suggest using a persistent object cache.
 */
function perflab_oc_health_should_persistent_object_cache() {
	global $wpdb;
```

felixarntz (commented):
@tillkruss I think all filters currently present make sense, except for the one I'm flagging below. Particularly the filters to modify the Site Health contents are great to have.

If you feel strongly about keeping this filter, I'd say at a minimum we should put that into the `perflab_oc_health_should_persistent_object_cache()` function as a "short-circuit" filter, rather than hooking that function onto the filter.

modules/object-cache/persistent-object-cache-health-check/load.php
tillkruss: Resolved in 873dc12d832ebb6ecd7a3237fc19662145cb0718.

tests/modules/object-cache/persistent-object-cache-health-check/health-check-test.php
```diff
@@ -0,0 +1,132 @@
...
+class Object_Cache_Health_Check_Tests extends WP_UnitTestCase {
+
+	function test_object_cache_default_thresholds() {
+		$result = perflab_oc_health_should_persistent_object_cache();
+
+		$this->assertFalse( $result );
+	}
+
+	function test_object_cache_thresholds_check_can_be_bypassed() {
+		add_filter( 'perflab_oc_site_status_persistent_object_cache_thresholds', '__return_true' );
```
felixarntz: Shouldn't this use `perflab_oc_site_status_suggest_persistent_object_cache` instead?

tests/modules/object-cache/persistent-object-cache-health-check/health-check-test.php
```diff
@@ -0,0 +1,132 @@
...
+
+		$result = perflab_oc_health_should_persistent_object_cache();
+
+		$this->assertTrue( $result );
+
+		remove_all_filters( 'perflab_oc_site_status_persistent_object_cache_thresholds' );
```
felixarntz: This can be removed, all hooks added during a test method are automatically cleared by the WP test suite. That applies to all other similar occurrences below as well.

tests/modules/object-cache/persistent-object-cache-health-check/health-check-test.php
```diff
@@ -0,0 +1,132 @@
...
+				return array_merge( $thresholds, array( 'alloptions_bytes' => 1000 ) );
+			}
+		);
+
+		$result = perflab_oc_health_should_persistent_object_cache();
+
+		$this->assertTrue( $result );
+
+		remove_all_filters( 'perflab_oc_site_status_persistent_object_cache_thresholds' );
+	}
```
felixarntz: Not critical, but all of these could be replaced with a single test method with parameters and an associated`@dataProvider` method.

There's a ton of duplicate code here that we could get rid of by doing that.

modules/object-cache/persistent-object-cache-health-check/load.php
```diff
@@ -0,0 +1,248 @@
...
+		)
+	);
+
+	return $result;
+}
+
+/**
+ * Callback for `site_status_persistent_object_cache` filter.
+ *
+ * Determines whether to suggest using a persistent object cache.
```
felixarntz: The first line here is no longer accurate.
```suggestion
 * Determines whether to suggest using a persistent object cache.
```

felixarntz (commented):
@tillkruss Production code lgtm, leaving a few recommendations on tests. At a minimum, we should fix the one filter usage there that's using the wrong filter.

tillkruss:
> @tillkruss Production code lgtm, leaving a few recommendations on tests. At a minimum, we should fix the one filter usage there that's using the wrong filter.

@felixarntz: All resolved in 6291513ddd924a871b656d7e063b3cac92ed72bd.

felixarntz (approved):
@tillkruss Awesome stuff! 🎉

PR #124, opened 2022-01-24
Add module to warn about excessive amount of autoloaded options

### LINKED ISSUE #92: Create Site Health Audit for Autoloaded options

manuelRod:
## Summary

Adds check into Site Health screen for autoloaded options.
Fixes [#92](https://github.com/WordPress/performance/issues/92)

## Relevant technical choices

There is a limit by default to trigger warnings (800kb)
There are 3 filters to modify the default behavior ( limit, description, and actions to take).
Those would be used by hosting companies to adapt to their requirements.
Filters:
```
$limit = apply_filters( 'perflab_aao_autoloaded_options_limit_size_in_bytes', 800000 );
$result['description'] = apply_filters( 'perflab_aao_autoloaded_options_limit_description', $result['description'] );
$result['actions'] = apply_filters( 'perflab_aao_autoloaded_options_action_to_perform', $result['actions'] );
```
Screen Shot 2022-02-01 at 4 47 08 PM
## Checklist

- [] PR has either `[Focus]` or `Infrastructure` label.
- [] PR has a `[Type]` label.
- [] PR has a milestone or the `no milestone` label.


manuelRod:
How I can add Focus/Type/Milestone? It seems I don't have permissions

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Autoloaded options counter.
+ *
+ * @since 1.0.0
+ *
+ * @return int
+ */
+function perflab_aao_autoloaded_options_count() {
+	return count( wp_load_alloptions() );
+}
```
spacedmonkey: I am not sure what value, this being it's own function is? Seems simple enough to just call a count.

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Calculate total amount of autoloaded data.
+ *
+ * @since 1.0.0
+ *
+ * @return int autoloaded data in bytes.
+ */
+function perflab_aao_autoloaded_options_size() {
+	global $wpdb;
+	$autoloaded_options_size = $wpdb->get_row( 'SELECT SUM(LENGTH(option_value)) FROM ' . $wpdb->prefix . 'options WHERE autoload = \'yes\'', ARRAY_N );
```
spacedmonkey: Instead of `get_row` should this use `get_var`

spacedmonkey:
A screenshot of what this looks like might be useful.

modules/site-health/audit-autoloaded-options/load.php
manuelRod: you are right, let's simplify

modules/site-health/audit-autoloaded-options/load.php
manuelRod: if we are expecting just one value, what is the benefit?

manuelRod:
This is how would look
Screen Shot 2022-02-01 at 4 47 08 PM

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,99 @@
...
+	return current( $autoloaded_options_size );
```
spacedmonkey:
```suggestion
	return $wpdb->get_var( 'SELECT SUM(LENGTH(option_value)) FROM ' . $wpdb->prefix . 'options WHERE autoload = \'yes\'' );
```

The benefit would be do this. @manuelRod

tests/modules/site-health/audit-autoloaded-options/audit-autoloaded-options-test.php
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	/**
+	 * Tests perflab_aao_autoloaded_options_size()
+	 */
+	public function test_perflab_aao_autoloaded_options_size() {
+		global $wpdb;
+		$autoloaded_options_size = $wpdb->get_row( 'SELECT SUM(LENGTH(option_value)) FROM ' . $wpdb->prefix . 'options WHERE autoload = \'yes\'', ARRAY_N );
+		$autoloaded_options_size = current( $autoloaded_options_size );
+		$this->assertEquals( $autoloaded_options_size, perflab_aao_autoloaded_options_size() );
```
spacedmonkey: Not sure about this test. This is just testing that logic inside function returns the same as function.

tests/testdata/modules/site-health/audit-autoloaded-options/class-autoloaded-options-set.php
```diff
@@ -0,0 +1,40 @@
...
+ * @package performance-lab
+ * @since 1.0.0
+ */
+
+/**
+ * Class Autoloaded_Options_Set sets extra options.
+ *
+ * @since 1.0.0
+ */
+class Autoloaded_Options_Set {
```
spacedmonkey: This should be remove it a method on `Audit_Autoloaded_Options_Tests` class. No need to new class here.

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,99 @@
...
+			$autoloaded_options_count,
+			size_format( $autoloaded_options_size )
+		),
+		'actions'     => '',
+		'test'        => 'autoloaded_options',
+	);
+
+	/**
+	 * Hostings can modify its limits depending on their own requirements.
+	 */
```
spacedmonkey: Comment style should match core. Add params and since here.

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,99 @@
...
+
+		$result['actions'] = sprintf(
+		/* translators: 1: HelpHub URL. 2: Link description. */
+			'<p><a target="_blank" href="%1$s">%2$s</a></p>',
+			esc_url( __( 'https://wordpress.org/support/article/optimization/', 'performance-lab' ) ),
+			esc_html__( 'More info about performance optimization', 'performance-lab' )
+		);
+		/**
+		 * Hostings can add actions to take to reduce size of autoloaded options linking to their own guides.
+		 */
```
spacedmonkey: See https://github.com/WordPress/performance/pull/124/files#r801877521

tests/modules/site-health/audit-autoloaded-options/audit-autoloaded-options-test.php
```diff
@@ -0,0 +1,72 @@
...
+
+	/**
+	 * Tests perflab_aao_autoloaded_options_test() when autoloaded options more than warning size.
+	 */
+	public function test_perflab_aao_autoloaded_options_test_warning() {
+		Autoloaded_Options_Set::set_autoloaded_option( self::WARNING_AUTOLOADED_SIZE_LIMIT_IN_BYTES );
+		$autoloaded_options_size  = perflab_aao_autoloaded_options_size();
+		$autoloaded_options_count = count( wp_load_alloptions() );
+
+		if ( $autoloaded_options_size > self::WARNING_AUTOLOADED_SIZE_LIMIT_IN_BYTES ) {
```
spacedmonkey: Why is there a conditional test here? This trigger a risk warning error, as no assets.

modules/site-health/audit-autoloaded-options/load.php
manuelRod: right! 👯

tests/testdata/modules/site-health/audit-autoloaded-options/class-autoloaded-options-set.php
manuelRod: ok, agree

tests/modules/site-health/audit-autoloaded-options/audit-autoloaded-options-test.php
manuelRod: that's the point, right? testing that function against that query, if that function would be modified, this test will fail

tests/modules/site-health/audit-autoloaded-options/audit-autoloaded-options-test.php
manuelRod: this conditional check is redundant, i will remove it ( together with the other one for less than limit threshold)

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,116 @@
...
+ * Calculate total amount of autoloaded data.
+ *
+ * @since 1.0.0
+ *
+ * @return int autoloaded data in bytes.
+ */
+function perflab_aao_autoloaded_options_size() {
+	global $wpdb;
+	$autoloaded_options_size = $wpdb->get_var( 'SELECT SUM(LENGTH(option_value)) FROM ' . $wpdb->prefix . 'options WHERE autoload = \'yes\'' );
+	return $autoloaded_options_size;
```
spacedmonkey:
```suggestion
	return $wpdb->get_var( 'SELECT SUM(LENGTH(option_value)) FROM ' . $wpdb->prefix . 'options WHERE autoload = \'yes\'' );
```

spacedmonkey (approved):

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,115 @@
...
+
+		/**
+		 * Hostings can add actions to take to reduce size of autoloaded options linking to their own guides.
+		 *
+		 * @since 1.0.0
+		 *
+		 * @param string $actions Call to Action to be used to point to the right direction to solve the issue.
+		 */
+		$result['actions'] = apply_filters( 'perflab_aao_autoloaded_options_action_to_perform', $result['actions'] );
+	}
```
mitogh:
```suggestion
	if ( $autoloaded_options_size <= $limit ) {
	  return $result;
	}
	
		$result['status']         = 'critical';
		$result['badge']['color'] = 'red';
		$result['description']    = sprintf(
		/* translators: 1: Number of autoloaded options. 2.Autoloaded options size. */
			'<p>' . esc_html__( 'Your website uses %1$s autoloaded options (size: %2$s). Try to reduce the number of autoloaded options or performance will be affected.', 'performance-lab' ) . '</p>',
			$autoloaded_options_count,
			size_format( $autoloaded_options_size )
		);

		/**
		 * Hostings can modify description to be shown on site health screen.
		 *
		 * @since 1.0.0
		 *
		 * @param string $description Description message when autoloaded options bigger than treshold.
		 */
		$result['description'] = apply_filters( 'perflab_aao_autoloaded_options_limit_description', $result['description'] );

		$result['actions'] = sprintf(
		/* translators: 1: HelpHub URL. 2: Link description. */
			'<p><a target="_blank" href="%1$s">%2$s</a></p>',
			esc_url( __( 'https://wordpress.org/support/article/optimization/', 'performance-lab' ) ),
			esc_html__( 'More info about performance optimization', 'performance-lab' )
		);

		/**
		 * Hostings can add actions to take to reduce size of autoloaded options linking to their own guides.
		 *
		 * @since 1.0.0
		 *
		 * @param string $actions Call to Action to be used to point to the right direction to solve the issue.
		 */
		$result['actions'] = apply_filters( 'perflab_aao_autoloaded_options_action_to_perform', $result['actions'] );
```

With an early return and inverting the condition here allow to remove an indentation level.

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Calculate total amount of autoloaded data.
+ *
+ * @since 1.0.0
+ *
+ * @return int autoloaded data in bytes.
+ */
+function perflab_aao_autoloaded_options_size() {
+	global $wpdb;
+	return $wpdb->get_var( 'SELECT SUM(LENGTH(option_value)) FROM ' . $wpdb->prefix . 'options WHERE autoload = \'yes\'' );
```
mitogh: Casting to an `int` would be ideal to match the expected return type.

```suggestion
	return (int) $wpdb->get_var( 'SELECT SUM(LENGTH(option_value)) FROM ' . $wpdb->prefix . 'options WHERE autoload = \'yes\'' );
```

tests/modules/site-health/audit-autoloaded-options/audit-autoloaded-options-test.php
```diff
@@ -0,0 +1,93 @@
...
+	 */
+	public static function set_autoloaded_option( $bytes = 800000 ) {
+		$heavy_option_string = self::random_string_generator( $bytes );
+		add_option( 'test_set_autoloaded_option', $heavy_option_string );
+	}
+
+	/**
+	 * Generate random string with certain $length.
+	 *
+	 * @param int $length Length of string to create.
```
mitogh: Would be worth mentioned that this length is in Bytes not in the number of characters of the string.

tests/modules/site-health/audit-autoloaded-options/audit-autoloaded-options-test.php
```diff
@@ -0,0 +1,93 @@
...
+		);
+	}
+
+	/**
+	 * Tests perflab_aao_autoloaded_options_size()
+	 */
+	public function test_perflab_aao_autoloaded_options_size() {
+		global $wpdb;
+		$autoloaded_options_size = $wpdb->get_var( 'SELECT SUM(LENGTH(option_value)) FROM ' . $wpdb->prefix . 'options WHERE autoload = \'yes\'' );
+		$this->assertEquals( $autoloaded_options_size, perflab_aao_autoloaded_options_size() );
```
mitogh, line 52: To make sure the returned type matches as well.

```suggestion
		$this->assertSame( $autoloaded_options_size, perflab_aao_autoloaded_options_size() );
```

tests/testdata/modules/site-health/audit-autoloaded-options/class-autoloaded-options-mock-responses.php
```diff
@@ -0,0 +1,91 @@
...
+ * @package performance-lab
+ * @since 1.0.0
+ */
+
+/**
+ * Class Autoloaded_Options_Mock_Responses mock site status health data for autoloaded options.
+ *
+ * @since 1.0.0
+ */
+class Autoloaded_Options_Mock_Responses {
```
mitogh: I see the reasoning to have this is in a separate class, however it seems like the only reason to exists is to return an array in all cases, it seems like a `@dataprovider` would be enough for all the method using this class:

- https://phpunit.de/manual/3.7/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers

That would remove the need to have this class in the first place.

mitogh (commented):
Great work just few minor comments / suggestions.

modules/site-health/audit-autoloaded-options/load.php
manuelRod: good point

tests/modules/site-health/audit-autoloaded-options/audit-autoloaded-options-test.php
manuelRod: fixed

tests/testdata/modules/site-health/audit-autoloaded-options/class-autoloaded-options-mock-responses.php
manuelRod: @mitogh from the 3 methods of that class, 2 accept arguments. Though this can be achievable with providers, I think for simplicity and readability's sake, I prefer the class choice.

mitogh: > from the 3 methods of that class, 2 accept arguments. 

A provider can or can't have arguments so I think this can be fine.

> Though this can be achievable with providers, I think for simplicity and readability's sake, I prefer the class choice.

I still think that a provider makes more sense tho as the only usage of this class is constrained to those 3 methods, currently, there's no other value for this class other than to pass data to a test function, which falls perfectly into the use case of a data provider.

manuelRod: @mitogh I moved them to providers and saved the mock response class.

mitogh: Thank you @manuelRod let me take a look

tests/modules/site-health/audit-autoloaded-options/audit-autoloaded-options-test.php
```diff
@@ -0,0 +1,170 @@
...
+		}
+		return $string;
+	}
+
+	/**
+	 * This is the information we are adding into site_status_tests hook.
+	 *
+	 * @return array
+	 */
+	function provider_added_test_info_site_health() {
```
mitogh:
```suggestion
	public function provider_added_test_info_site_health() {
```

mitogh (commented):
Just one minor thing other than that looks great thanks @manuelRod for your work on this one.

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,116 @@
...
+
+	if ( $autoloaded_options_size < $limit ) {
+		return $result;
+	}
+
+	$result['status']         = 'critical';
+	$result['badge']['color'] = 'red';
+	$result['description']    = sprintf(
+	/* translators: 1: Number of autoloaded options. 2.Autoloaded options size. */
+		'<p>' . esc_html__( 'Your website uses %1$s autoloaded options (size: %2$s). Try to reduce the number of autoloaded options or performance will be affected.', 'performance-lab' ) . '</p>',
```
felixarntz, line 74: How can the user reduce the number of autoloaded options? The average end user will have no idea what that even means. We have introduced extensive documentation even for high-level technical topics like [what PHP is and how to update it](https://wordpress.org/support/update-php/#why-php-matters-to-you), and I'd argue the topic here goes way beyond that.

Also, there is no way for the user right now to see the autoloaded options, so even if they had an idea about what it means, how can they reason about it or fix it? In the "Audit Enqueued Assets" module for example, people can go to the frontend and see which CSS and JS assets are being enqueued and then based on that get an idea which plugins or themes are responsible. But that is not possible here.

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,116 @@
...
+		$autoloaded_options_count,
+		size_format( $autoloaded_options_size )
+	);
+
+	/**
+	 * Hostings can modify description to be shown on site health screen.
```
felixarntz: That's not a good filter description, we shouldn't include something overly specific like "Hostings can...". Let's keep descriptions generic and follow the core standards here, e.g. this can become "Filters description to be shown...".

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,116 @@
...
+			'<p>' . esc_html__( 'The amount of %1$s autoloaded options (size: %2$s) in options table is acceptable.', 'performance-lab' ) . '</p>',
+			$autoloaded_options_count,
+			size_format( $autoloaded_options_size )
+		),
+		'actions'     => '',
+		'test'        => 'autoloaded_options',
+	);
+
+	/**
+	 * Hostings can modify its limits depending on their own requirements.
```
felixarntz: See above, this would be better as "Filters the limits...". Also, which limits? The message right now does not mention autoloaded options in any way, so that's a bit too generic.

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,116 @@
...
+
+	$result['actions'] = sprintf(
+	/* translators: 1: HelpHub URL. 2: Link description. */
+		'<p><a target="_blank" href="%1$s">%2$s</a></p>',
+		esc_url( __( 'https://wordpress.org/support/article/optimization/', 'performance-lab' ) ),
+		esc_html__( 'More info about performance optimization', 'performance-lab' )
+	);
+
+	/**
+	 * Hostings can add actions to take to reduce size of autoloaded options linking to their own guides.
```
felixarntz: See above, this would be better as "Filters actions to take...".

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,116 @@
+<?php
+/**
+ * Module Name: Audit Autoloaded Options
+ * Description: Adds Autoloaded options checker in Site Health checks.
+ * Experimental: No
```
felixarntz: If we merge it like this, I'd be in favor of marking it as experimental based on my previous comment that this isn't reasonably actionable for end users.

felixarntz (commented):
While the code looks technically solid, I have some concerns about this module as it is right now, as it is not actionable. What can we reasonably expect an end user to do here? While other Site Health status warnings are also technical, this is another level, there is no way for someone other than a developer to fix this. A way to make this more actionable for site owners would be to give some indication which plugins or themes are adding the majority of these, and then the end user could decide whether one of those isn't critical to them.

If there is consensus that this should be merged even in the state that it is, I'd be in favor of marking it as experimental.

modules/site-health/audit-autoloaded-options/load.php
manuelRod: Hello @felixarntz those are valid concerns, my initial idea was to flag this problem so hostings could link to actionable guides on how to solve it ( you can check the original discussion on its [issue](https://github.com/WordPress/performance/issues/92))

Most, if not all, hosting providers have specific articles pointing to this problem, see [Kinsta](https://kinsta.com/knowledgebase/wp-options-autoloaded-data/) one.

It happened to me in the past that some clients were having death screens that were caused by this, opening a ticket to the hosting, it ended up sending me a link to its guide on how to solve it. These warnings would allow hosting to tweak its limits and point the client in the right direction. (that's why threshold + actionable text is filterable).

Regarding default behavior, @tillkruss mentioned in the issue that this would be a nice place [https://wordpress.org/support/article/optimization/](https://wordpress.org/support/article/optimization/) though it would need a new topic to cover the issue.

As for next iteration, indeed, we should expand functionality and give more information or even offer some actionable solution, see again point 2 in the issue.
> Functional module, we can go a step further and keep track of used/unused options, and then give the ability to users to clean up the table from the site health screen itself.

tests/modules/site-health/audit-autoloaded-options/audit-autoloaded-options-test.php
manuelRod: good eye :)

modules/site-health/audit-autoloaded-options/load.php
manuelRod: updated

modules/site-health/audit-autoloaded-options/load.php
manuelRod: updated

modules/site-health/audit-autoloaded-options/load.php
manuelRod: updated. 

The message right now is the following:
> Your website uses %1$s autoloaded options (size: %2$s). Try to reduce the number of autoloaded options or performance will be affected.

We don't mention the default limit, since in theory this feature would be used by different hosting providers that will set up their own limit.

modules/site-health/audit-autoloaded-options/load.php
felixarntz: Let's continue this discussion on the issue as you've already started doing. The main thing if we don't add the guidance here is that there should be another issue to do it.

Will resolve this comment in favor of conversation on the issue (where there's also the vote going on).

felixarntz:
The code here looks solid technically now. Whether or how we should merge it depends on the vote in https://github.com/WordPress/performance/issues/92#issuecomment-1068215411 which is open until this Friday.

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -0,0 +1,116 @@
...
+	 * @since 1.0.0
+	 *
+	 * @param string $description Description message when autoloaded options bigger than treshold.
+	 */
+	$result['description'] = apply_filters( 'perflab_aao_autoloaded_options_limit_description', $result['description'] );
+
+	$result['actions'] = sprintf(
+	/* translators: 1: HelpHub URL. 2: Link description. */
+		'<p><a target="_blank" href="%1$s">%2$s</a></p>',
+		esc_url( __( 'https://wordpress.org/support/article/optimization/', 'performance-lab' ) ),
```
adamsilverstein, line 91: this link should go right to a support section explaining how to optimize autoload options, eg `https://wordpress.org/support/article/optimization/#autoload`. The link as is is not helpful to users -   especially since the page doesn't even mention autoload options.

tests/modules/site-health/audit-autoloaded-options/audit-autoloaded-options-test.php
```diff
@@ -0,0 +1,170 @@
...
+		delete_option( self::AUTOLOADED_OPTION_KEY );
+	}
+
+	/**
+	 * Generate random string with certain $length.
+	 *
+	 * @param int $length Length ( in bytes ) of string to create.
+	 * @return string
+	 */
+	protected static function random_string_generator( $length ) {
```
adamsilverstein: Can you use `wp_generate_password` instead of adding this function?

modules/site-health/audit-autoloaded-options/load.php
manuelRod: I agree with this, https://wordpress.org/support/article/optimization/#autoload should be written and we can link in there.

tests/modules/site-health/audit-autoloaded-options/audit-autoloaded-options-test.php
manuelRod: changed for simplicity sake

felixarntz:
@manuelRod Updated this PR to be against the `1.0.0-beta.2` branch.

aristath (approved):

mitogh (approved):

felixarntz (approved):

PR #136, opened 2022-01-28
Enhance detection of enqueued frontend assets

### LINKED ISSUE #135: Enhance reliability of how Site Health enqueued assets module detects assets
The initial implementation for the Site Health enqueued assets module (see #25) is a solid start, but there's room for enhancement in how reliably it detects enqueued assets. This was already pointed out in the [initial PR](https://github.com/WordPress/performance/pull/54) review, but it was decided to merge the large first PR first and then implement this enhancement separately, to allow for a more focused review.

This issue is primarily to address the following review comment: https://github.com/WordPress/performance/pull/54#discussion_r793498014

There are potentially more aspects we can improve around this, let's explore this here.

manuelRod:
## Summary

Enhances for Site Health Enqueued assets module.
Fixes #135

## Relevant technical choices

Follow up of the previous closed PR [#54 ](https://github.com/WordPress/performance/pull/54)

## Checklist

### DIFF
--- a/modules/site-health/audit-enqueued-assets/helper.php
+++ b/modules/site-health/audit-enqueued-assets/helper.php
@@ -78,6 +78,8 @@ function perflab_aea_get_total_size_bytes_enqueued_styles() {
 
 /**
  * Convert full URL paths to absolute paths.
+ * Covers Standard WP configuration, wp-content outside WP directories and subdirectories.
+ * Ex: https://example.com/content/themes/, https://example.com/wp/wp-includes/
  *
  * @since 1.0.0
  *
@@ -85,7 +87,20 @@ function perflab_aea_get_total_size_bytes_enqueued_styles() {
  * @return string Returns absolute path to the resource.
  */
 function perflab_aea_get_path_from_resource_url( $resource_url ) {
-	return ABSPATH . wp_make_link_relative( $resource_url );
+
+	// Different content folder ex. /content/.
+	if ( 0 === strpos( $resource_url, content_url() ) ) {
+		return WP_CONTENT_DIR . substr( $resource_url, strlen( content_url() ) );
+	}
+
+	// wp-content in a subdirectory. ex. /blog/wp-content/.
+	$site_url = untrailingslashit( site_url() );
+	if ( 0 === strpos( $resource_url, $site_url ) ) {
+		return untrailingslashit( ABSPATH ) . substr( $resource_url, strlen( $site_url ) );
+	}
+
+	// Standard wp-content configuration.
+	return untrailingslashit( ABSPATH ) . wp_make_link_relative( $resource_url );
 }
 
 /**
--- a/modules/site-health/audit-enqueued-assets/helper.php
+++ b/modules/site-health/audit-enqueued-assets/helper.php
@@ -87,6 +87,9 @@ function perflab_aea_get_total_size_bytes_enqueued_styles() {
  * @return string Returns absolute path to the resource.
  */
 function perflab_aea_get_path_from_resource_url( $resource_url ) {
+	if ( ! $resource_url ) {
+		return '';
+	}
 
 	// Different content folder ex. /content/.
 	if ( 0 === strpos( $resource_url, content_url() ) ) {
--- a/modules/site-health/audit-enqueued-assets/load.php
+++ b/modules/site-health/audit-enqueued-assets/load.php
@@ -28,22 +28,24 @@ function perflab_aea_audit_enqueued_scripts() {
 		foreach ( $wp_scripts->done as $handle ) {
 			$script = $wp_scripts->registered[ $handle ];
 
-			if ( $script->src && ! strpos( $script->src, 'wp-includes' ) ) {
-
-				// Add any extra data (inlined) that was passed with the script.
-				$inline_size = 0;
-				if ( ! empty( $script->extra ) && ! empty( $script->extra['after'] ) ) {
-					foreach ( $script->extra['after'] as $extra ) {
-						$inline_size += ( is_string( $extra ) ) ? mb_strlen( $extra, '8bit' ) : 0;
-					}
-				}
-
-				$path               = perflab_aea_get_path_from_resource_url( $script->src );
-				$enqueued_scripts[] = array(
-					'src'  => $script->src,
-					'size' => perflab_aea_get_resource_file_size( $path ) + $inline_size,
-				);
+			if ( ! $script->src || false !== strpos( $script->src, 'wp-includes' ) ) {
+				continue;
 			}
+
+			// Add any extra data (inlined) that was passed with the script.
+			$inline_size = 0;
+			if ( ! empty( $script->extra ) && ! empty( $script->extra['after'] ) ) {
+				foreach ( $script->extra['after'] as $extra ) {
+					$inline_size += ( is_string( $extra ) ) ? mb_strlen( $extra, '8bit' ) : 0;
+				}
+			}
+
+			$path               = perflab_aea_get_path_from_resource_url( $script->src );
+			$enqueued_scripts[] = array(
+				'src'  => $script->src,
+				'size' => perflab_aea_get_resource_file_size( $path ) + $inline_size,
+			);
+
 		}
 		set_transient( 'aea_enqueued_front_page_scripts', $enqueued_scripts, 12 * HOUR_IN_SECONDS );
 	}
@@ -63,28 +65,30 @@ function perflab_aea_audit_enqueued_styles() {
 		$enqueued_styles = array();
 		foreach ( $wp_styles->done as $handle ) {
 			$style = $wp_styles->registered[ $handle ];
-			if ( $style->src && ! strpos( $style->src, 'wp-includes' ) ) {
 
-				// Check if we already have the style's path ( part of a refactor for block styles from 5.9 ).
-				if ( ! empty( $style->extra ) && ! empty( $style->extra['path'] ) ) {
-					$path = $style->extra['path'];
-				} else { // Fallback to getting the path from the style's src.
-					$path = perflab_aea_get_path_from_resource_url( $style->src );
-				}
-
-				// Add any extra data (inlined) that was passed with the style.
-				$inline_size = 0;
-				if ( ! empty( $style->extra ) && ! empty( $style->extra['after'] ) ) {
-					foreach ( $style->extra['after'] as $extra ) {
-						$inline_size += ( is_string( $extra ) ) ? mb_strlen( $extra, '8bit' ) : 0;
-					}
-				}
-
-				$enqueued_styles[] = array(
-					'src'  => $style->src,
-					'size' => perflab_aea_get_resource_file_size( $path ) + $inline_size,
-				);
+			if ( ! $style->src || false !== strpos( $style->src, 'wp-includes' ) ) {
+				continue;
 			}
+
+			// Check if we already have the style's path ( part of a refactor for block styles from 5.9 ).
+			if ( ! empty( $style->extra ) && ! empty( $style->extra['path'] ) ) {
+				$path = $style->extra['path'];
+			} else { // Fallback to getting the path from the style's src.
+				$path = perflab_aea_get_path_from_resource_url( $style->src );
+			}
+
+			// Add any extra data (inlined) that was passed with the style.
+			$inline_size = 0;
+			if ( ! empty( $style->extra ) && ! empty( $style->extra['after'] ) ) {
+				foreach ( $style->extra['after'] as $extra ) {
+					$inline_size += ( is_string( $extra ) ) ? mb_strlen( $extra, '8bit' ) : 0;
+				}
+			}
+
+			$enqueued_styles[] = array(
+				'src'  => $style->src,
+				'size' => perflab_aea_get_resource_file_size( $path ) + $inline_size,
+			);
 		}
 		set_transient( 'aea_enqueued_front_page_styles', $enqueued_styles, 12 * HOUR_IN_SECONDS );
 	}
--- a/modules/site-health/audit-enqueued-assets/load.php
+++ b/modules/site-health/audit-enqueued-assets/load.php
@@ -40,7 +40,11 @@ function perflab_aea_audit_enqueued_scripts() {
 				}
 			}
 
-			$path               = perflab_aea_get_path_from_resource_url( $script->src );
+			$path = perflab_aea_get_path_from_resource_url( $script->src );
+			if ( ! $path ) {
+				continue;
+			}
+
 			$enqueued_scripts[] = array(
 				'src'  => $script->src,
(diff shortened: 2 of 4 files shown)


felixarntz (approved):
@manuelRod This looks solid to me!

Would love to get someone else to review as well, can you maybe give this a pass @aristath?

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -25,22 +25,33 @@ function perflab_aea_audit_enqueued_scripts() {
 		global $wp_scripts;
 		$enqueued_scripts = array();
 
-		foreach ( $wp_scripts->queue as $handle ) {
-			$src = $wp_scripts->registered[ $handle ]->src;
-			if ( $src && ! strpos( $src, 'wp-includes' ) ) {
+		foreach ( $wp_scripts->done as $handle ) {
+			$script = $wp_scripts->registered[ $handle ];
+
+			if ( $script->src && ! strpos( $script->src, 'wp-includes' ) ) {
```
mitogh: Personally adding an early return (`continue` in this case) would allow the removal for the indentation level here, giving a simpler way to follow the code.

```suggestion
			if ( ! $script->src || strpos( $script->src, 'wp-includes' ) === false ) {
					continue;
			}
```

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -25,22 +25,33 @@ function perflab_aea_audit_enqueued_scripts() {
...
+
+				// Add any extra data (inlined) that was passed with the script.
+				$inline_size = 0;
+				if ( ! empty( $script->extra ) && ! empty( $script->extra['after'] ) ) {
```
mitogh: Similar to the above.

```suggestion
				if ( empty( $script->extra['after'] ) || ! is_array( $script->extra['after'] )  ) {
					continue;
				}
				$inline_size = 0;
				
```

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -50,19 +61,35 @@ function perflab_aea_audit_enqueued_styles() {
 	if ( ! is_admin() && is_front_page() && current_user_can( 'view_site_health_checks' ) && false === get_transient( 'aea_enqueued_front_page_styles' ) ) {
 		global $wp_styles;
 		$enqueued_styles = array();
-		foreach ( $wp_styles->queue as $handle ) {
-			$src = $wp_styles->registered[ $handle ]->src;
-			if ( $src && ! strpos( $src, 'wp-includes' ) ) {
+		foreach ( $wp_styles->done as $handle ) {
+			$style = $wp_styles->registered[ $handle ];
+			if ( $style->src && ! strpos( $style->src, 'wp-includes' ) ) {
```
mitogh: Similar comment as described above.

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -50,19 +61,35 @@ function perflab_aea_audit_enqueued_styles() {
...
+
+				// Check if we already have the style's path ( part of a refactor for block styles from 5.9 ).
+				if ( ! empty( $style->extra ) && ! empty( $style->extra['path'] ) ) {
+					$path = $style->extra['path'];
+				} else { // Fallback to getting the path from the style's src.
+					$path = perflab_aea_get_path_from_resource_url( $style->src );
+				}
```
mitogh: By inverting the conditional here allows the removal of the `!` operator, allowing for a more natural flow.

```suggestion
				// default to getting the path from the style's src.
				if ( empty( $style->extra ) || empty( $style->extra['path'] ) ) {
					$path = perflab_aea_get_path_from_resource_url( $style->src );				
				} else { 
					$path = $style->extra['path'];
				}
```

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -50,19 +61,35 @@ function perflab_aea_audit_enqueued_styles() {
...
+
+				// Add any extra data (inlined) that was passed with the style.
+				$inline_size = 0;
+				if ( ! empty( $style->extra ) && ! empty( $style->extra['after'] ) ) {
```
mitogh: Similar comment to the above.

modules/site-health/audit-enqueued-assets/load.php
felixarntz: @mitogh Not sure I agree in this case, the original code here explicitly looks at whether the `$style->extra` value is not empty, and the _other_ function use would be the fallback. I don't have a strong preference on the order, but with your suggestion here the comment in line 71 is no longer accurate.

mitogh: Oh, right let me update line 71.

Mostly this pattern avoids the usage of `!` which translates 1 operation into 2.

felixarntz: I'd argue that's a bit of a micro optimization. IMO the code path checking `! empty( $var )` and then using `$var` is more straightforward to read.

mitogh: The suggestion goes more towards optimization for reading rather than performance so it becomes a personal preference I would say either way is fine, but IMO this seems easier to read:

- If empty then
- Else

Rather than 

- If not empty then
- else

modules/site-health/audit-enqueued-assets/helper.php
```diff
@@ -78,14 +78,29 @@ function perflab_aea_get_total_size_bytes_enqueued_styles() {
...
  * Convert full URL paths to absolute paths.
+ * Covers Standard WP configuration, wp-content outside WP directories and subdirectories.
+ * Ex: https://example.com/content/themes/, https://example.com/wp/wp-includes/
  *
  * @since 1.0.0
  *
  * @param string $resource_url URl resource link.
  * @return string Returns absolute path to the resource.
  */
 function perflab_aea_get_path_from_resource_url( $resource_url ) {
```
dainemawer, line 89: Im noticing quite a few varying function namespaces - maybe this function should be called `perflab_get_path_from_resource_url`?

modules/site-health/audit-enqueued-assets/helper.php
```diff
@@ -78,14 +78,29 @@ function perflab_aea_get_total_size_bytes_enqueued_styles() {
...
-	return ABSPATH . wp_make_link_relative( $resource_url );
+
+	// Different content folder ex. /content/.
+	if ( 0 === strpos( $resource_url, content_url() ) ) {
```
dainemawer, line 95: Do we perhaps need to handle a situation where `$resource_url` is `undefined`? Maybe a:

```
if(! $resource_url ) {
  return;
}
```

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -25,22 +25,33 @@ function perflab_aea_audit_enqueued_scripts() {
...
+						$inline_size += ( is_string( $extra ) ) ? mb_strlen( $extra, '8bit' ) : 0;
+					}
+				}
+
+				$path               = perflab_aea_get_path_from_resource_url( $script->src );
 				$enqueued_scripts[] = array(
-					'src'  => $src,
-					'size' => perflab_aea_get_resource_file_size( perflab_aea_get_path_from_resource_url( $src ) ),
+					'src'  => $script->src,
+					'size' => perflab_aea_get_resource_file_size( $path ) + $inline_size,
```
dainemawer: Does this size take into account the size on the server / disk - or the transferred / gzipped size? Probably worth noting as the 2 can be different and could skew results.

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -25,22 +25,33 @@ function perflab_aea_audit_enqueued_scripts() {
...
 		}
 		set_transient( 'aea_enqueued_front_page_scripts', $enqueued_scripts, 12 * HOUR_IN_SECONDS );
 	}
 }
-add_action( 'wp_print_scripts', 'perflab_aea_audit_enqueued_scripts' );
+add_action( 'wp_footer', 'perflab_aea_audit_enqueued_scripts', PHP_INT_MAX );
 
 /**
- * Audit enqueued styles in the frontend. Ignore /wp-includes styles.
+ * Audit enqueued and printed styles in the frontend. Ignore /wp-includes styles.
```
dainemawer, line 60: Why would we ignore `wp-includes`? Gutenberg for instance adds a couple of stylesheets that arent exactly optimized. My opinion is that we should give the user the full picture, even if it means flagging WP for enqueuing unnecessary stylesheets.

dainemawer (commented):
Nice work overall, left a few questions!

aristath (approved):
Apologies it took me so long to get to this PR.
The code looks good to me and makes sense 👍

felixarntz:
@manuelRod FYI Since we've branched off from `trunk` into a `release/1.0.0-beta.1` branch, I've updated the base branch of this PR to that new branch. So going forward, please make sure to work off `release/1.0.0-beta.1`, and specifically **not merge `trunk`** into this again.

modules/site-health/audit-enqueued-assets/load.php
manuelRod: @dainemawer those assets are included by default by wp, and this module focus on an overall overview of external assets being enqueued (plugins, theme), I think it would make more noise to include wp-includes, since most of the consumers won't be techies and won't know how to react on it.
I agree that going forward and as we iterate over this module, we can show more useful information, at that point, we can add more info (like wp-includes) and tips on how to improve it, but for now, I would just ignore them.

modules/site-health/audit-enqueued-assets/load.php
manuelRod: makes sense

modules/site-health/audit-enqueued-assets/load.php
manuelRod: makes sense

modules/site-health/audit-enqueued-assets/load.php
manuelRod: this would break functionality. If empty extra['after'] $inline_size will be 0 and that's it, we don't need to skip the iteration with continue, since the loop is not finished.

modules/site-health/audit-enqueued-assets/load.php
manuelRod: same as above, it would break functionality.

modules/site-health/audit-enqueued-assets/load.php
manuelRod: I also don't agree with this (since is a personal preference), to me, seems more natural and easier to read the way it is right now.

modules/site-health/audit-enqueued-assets/helper.php
manuelRod: hello, @dainemawer all functions are prepending "perflab_aea", what are the function namespaces varying? maybe I'm missing some

modules/site-health/audit-enqueued-assets/helper.php
manuelRod: you are right, not sure if that case would happen, but it's worth nothing to add it.

modules/site-health/audit-enqueued-assets/load.php
manuelRod: it's the disk/server size. It uses filesize() to get the file size.

manuelRod:
applied some of the comments! not sure if anyone wants to review them again to be ready for the release @felixarntz @aristath

modules/site-health/audit-enqueued-assets/helper.php
felixarntz: I think the approach of having module specific namespaces (all of them starting with `perflab_` though due to the overall plugin) is a solid approach, which avoids us running into accidental conflicts between individual modules.

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -25,22 +25,35 @@ function perflab_aea_audit_enqueued_scripts() {
...
+
+			// Add any extra data (inlined) that was passed with the script.
+			$inline_size = 0;
+			if ( ! empty( $script->extra ) && ! empty( $script->extra['after'] ) ) {
+				foreach ( $script->extra['after'] as $extra ) {
+					$inline_size += ( is_string( $extra ) ) ? mb_strlen( $extra, '8bit' ) : 0;
+				}
 			}
+
+			$path               = perflab_aea_get_path_from_resource_url( $script->src );
```
felixarntz: Since you've changed `perflab_aea_get_path_from_resource_url()` to support empty values (which may not be necessary), we need to make sure this here can also deal with that change properly, supporting an empty return value.
```suggestion
			$path = perflab_aea_get_path_from_resource_url( $script->src );
			if ( ! $path ) {
				continue;
			}

```

modules/site-health/audit-enqueued-assets/load.php
```diff
@@ -50,19 +63,37 @@ function perflab_aea_audit_enqueued_styles() {
...
+
+			if ( ! $style->src || false !== strpos( $style->src, 'wp-includes' ) ) {
+				continue;
 			}
+
+			// Check if we already have the style's path ( part of a refactor for block styles from 5.9 ).
+			if ( ! empty( $style->extra ) && ! empty( $style->extra['path'] ) ) {
+				$path = $style->extra['path'];
+			} else { // Fallback to getting the path from the style's src.
+				$path = perflab_aea_get_path_from_resource_url( $style->src );
```
felixarntz, line 81: See above.
```suggestion
				$path = perflab_aea_get_path_from_resource_url( $style->src );
				if ( ! $path ) {
					continue;
				}
```

felixarntz (commented):
@manuelRod Two more tiny changes needed here based on the change from https://github.com/WordPress/performance/pull/136/files#r819490106, I can quickly add them.

felixarntz (approved):
Added the two tiny changes, all the rest already lgtm before. Thank you, great work @manuelRod!

modules/site-health/audit-enqueued-assets/load.php
dainemawer: @manuelRod - that makes sense - just so Im understanding correctly, would that size be more or less the same than transferred / gzipped size?

modules/site-health/audit-enqueued-assets/load.php
dainemawer: Thats a fair statement, thanks @manuelRod !

modules/site-health/audit-enqueued-assets/helper.php
dainemawer: @manuelRod I was thinking more globally across the plugin - I've noticed other functions are all prefixed with `perflab_` but not necessarily `perflab_aea` - that being said it sounds like @felixarntz is good with this approach - Im not swayed either way - just wanted to call it out for consistency reasons :)

PR #139, opened 2022-01-31
Update the plugin contributors

JustinyAhin:
## Summary

Following https://github.com/WordPress/performance/issues/42#issuecomment-1024067465, this PR updates the contributor to the plugin to use https://profiles.wordpress.org/performanceteam handle.

## Checklist

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -1,6 +1,6 @@
 === Performance Lab ===
 
-Contributors:      wordpressdotorg
+Contributors:      performanceteam
 Requires at least: 5.8
 Tested up to:      5.9
 Requires PHP:      5.6


felixarntz (commented):
@JustinyAhin I'm not sure we should make this change. Per the conversation in #42, the `performanceteam` account is purely for operational purposes and shouldn't really be used for anything else. I don't think we should draw attention to it, the same way you don't really see the `gutenbergplugin` wordpress.org account surface anywhere.

IMO the [`wordpressdotorg`](https://profiles.wordpress.org/wordpressdotorg/) account is more appropriate to use here, since this plugin is a wordpress.org project and e.g. plugins like [Classic Editor](https://wordpress.org/plugins/classic-editor/) use that too.

Let's see what other people think here, but my vote would be to leave it as `wordpressdotorg` and close this PR.

PR #141, opened 2022-02-01
Add WebP support in site health

### LINKED ISSUE #130: Expose WebP support under Site Health status tab
Since not every server environment supports translating JPEG uploads to WebP format, it's worth indicating this to site administrators in an easy-to-understand way. While we already have some basic PHP extension information about image capabilities in the Site Health info tab, it may be worth adding something explicitly about WebP in the Site Health "Status" tab.

For example a message like "WebP images are (not) supported by your server environment", with a brief description on why this is relevant. And potentially a note to "Ask your hosting provider to enable/update xyz"

Also see https://wordpress.slack.com/archives/C02KGN5K076/p1643112555307200?thread_ts=1643041077.295600&cid=C02KGN5K076

cc @adamsilverstein @getsource @spacedmonkey

kirtangajjar:
## Summary
Fixes #130. 

## Relevant technical choices

Adds a module in site health to check for WebP support and shows a warning if support is not present. 
If either GD or Imagick has WebP support, the WebP support warning is not shown.

The code of checking support in GD and Imagick is same as how it's shown in site health info's media handling section.

## Screenshots
Screenshot from 2022-02-03 18-19-09

### DIFF
--- a/modules/site-health/webp-support/load.php
+++ b/modules/site-health/webp-support/load.php
@@ -48,7 +48,7 @@ function webp_uploads_check_webp_supported_test() {
 		'test'        => 'is_webp_uploads_enabled',
 	);
 
-	$webp_supported = wp_image_editor_supports( array( 'mime_type' => 'image/webp' ) ) ;
+	$webp_supported = wp_image_editor_supports( array( 'mime_type' => 'image/webp' ) );
 
 	if ( ! $webp_supported ) {
 		$result['status']  = 'critical';
--- a/modules/site-health/webp-support/load.php
+++ b/modules/site-health/webp-support/load.php
@@ -42,7 +42,7 @@ function webp_uploads_check_webp_supported_test() {
 		),
 		'description' => sprintf(
 			'<p>%s</p>',
-			__( 'WebP image format is used by WordPress to improve the performance of your site by generating smaller images than it usually could with JPEG format. This means your pages will load faster and consume less bandwidth from users.', 'performance-lab' )
+			__( 'WebP image format is used by WordPress to improve the performance of your site by generating smaller images than it usually could with the JPEG format. This means your pages will load faster and consume less bandwidth.', 'performance-lab' )
 		),
 		'actions'     => '',
 		'test'        => 'is_webp_uploads_enabled',
--- a/modules/site-health/webp-support/load.php
+++ b/modules/site-health/webp-support/load.php
@@ -51,7 +51,7 @@ function webp_uploads_check_webp_supported_test() {
 	$webp_supported = wp_image_editor_supports( array( 'mime_type' => 'image/webp' ) );
 
 	if ( ! $webp_supported ) {
-		$result['status']  = 'critical';
+		$result['status']  = 'recommended';
 		$result['label']   = __( 'Your site does not support WebP', 'performance-lab' );
 		$result['actions'] = sprintf(
 			'<p>%s</p>',
--- a/modules/site-health/webp-support/load.php
+++ b/modules/site-health/webp-support/load.php
@@ -56,7 +56,7 @@ function webp_uploads_check_webp_supported_test() {
 		$result['actions'] = sprintf(
 			'<p>%s</p>',
 			/* translators: Accessibility text. */
-			__( 'Please contact your host and ask them to add WebP support', 'performance-lab' )
+			__( 'Please contact your host and ask them to add WebP support.', 'performance-lab' )
 		);
 	}
 
--- a/modules/site-health/webp-support/load.php
+++ b/modules/site-health/webp-support/load.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Module Name: Web Uploads Support
+ * Module Name: Web Support
  * Description: Add WebP support check in Site Health checks.
  * Experimental: No
  *
--- a/modules/site-health/webp-uploads-enabled/load.php
+++ b/modules/site-health/webp-uploads-enabled/load.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * Module Name: Web Uploads Support
+ * Description: Add WebP support check in Site Health checks.
+ * Experimental: No
+ *
+ * @package performance-lab
+ * @since 1.0.0
+ */
+
+/**
+ * Adds tests to site health.
+ *
+ * @since 1.0.0
+ *
+ * @param array $tests Site Health Tests.
+ * @return array
+ */
+function webp_uploads_add_is_webp_supported_test( $tests )
+{
+	$tests['direct']['webp_supported']  = array(
+		'label' => esc_html__( 'WebP Support', 'performance-lab' ),
+		'test'  => 'webp_uploads_check_webp_supported_test',
+	);
+	return $tests;
+}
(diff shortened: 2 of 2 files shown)


pbearne:
it's nice to see the output But most users won't know what to look for
I feel we need to add a "score" as to how good the support is. 
ie a nice green 100% support if all is good and messages, if support expected/desired, is missing

kirtangajjar:
@pbearne Can you look at the screenshot again? Earlier I added a screenshot in the last line to indicate where I copied the code of checking WebP support from, but I saw how it can be mistaken as the output of the code, so I have uploaded the actual screenshot of output. 

Do you still feel we need to show percentage-wise support? I think now since the user has a clear picture to see that they have an action to take here, I feel we don't need to show partial support indication.

modules/site-health/webp-uploads-enabled/load.php
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Callback for webp_enabled test.
+ *
+ * @since 1.0.0
+ *
+ * @return array
+ */
+function webp_uploads_check_webp_supported_test() {
+	$result = array(
+		'label'       => __( 'Your site supports WebP' ),
```
mitogh: Missing the domain name on the translation.

modules/site-health/webp-uploads-enabled/load.php
```diff
@@ -0,0 +1,80 @@
...
+		'status'      => 'good',
+		'badge'       => array(
+			'label' => __( 'Performance' ),
```
mitogh: Same as above, missing translation domain.

modules/site-health/webp-uploads-enabled/load.php
```diff
@@ -0,0 +1,80 @@
...
+			'color' => 'blue',
+		),
+		'description' => sprintf(
+			'<p>%s</p>',
+			__( 'WebP image format is used by WordPress to boost performance of your site by generating smaller images than it usually could with default JPEG format. This means your pages will load faster and consume less data from users.' )
```
mitogh: Missing domain as well. Some suggestions for the wording of this section.

```suggestion
			__( 'WebP image format is used by WordPress to improve the performance of your site by generating smaller images than it usually could with JPEG format. This means your pages will load faster and consume less bandwidth from users.' )
```

modules/site-health/webp-uploads-enabled/load.php
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	try {
+		$formats = Imagick::queryFormats( '*' );
+
+		if( in_array('WEBP', $formats ) ) {
+			$gd_supported = true;
+		}
+	} catch (Exception $e) {
+	}
```
mitogh: Looks like this section of code can be replaced with: 

```php
wp_image_editor_supports( array( 'mime_type' => 'image/webp' ) ) 
```

As that would use the supported image editor if Imagemagic is supported or GD if that's the case.

mitogh (changes requested):

modules/site-health/webp-uploads-enabled/load.php
kirtangajjar: Fixed

modules/site-health/webp-uploads-enabled/load.php
kirtangajjar: Fixed

modules/site-health/webp-uploads-enabled/load.php
kirtangajjar: Thanks for the tip. I've made the change and the code looks much cleaner :)

mitogh (commented):
This looks good to me, let's wait on other people to chime in case they have any additional feedback.

Thank you for putting this together :1st_place_medal: :bow:

mmuyskens:
I wouldn't call it critical. Kick it down to recommendations.

modules/site-health/webp-support/load.php
```diff
@@ -0,0 +1,64 @@
...
+
+	$webp_supported = wp_image_editor_supports( array( 'mime_type' => 'image/webp' ) );
+
+	if ( ! $webp_supported ) {
+		$result['status']  = 'critical';
+		$result['label']   = __( 'Your site does not support WebP', 'performance-lab' );
+		$result['actions'] = sprintf(
+			'<p>%s</p>',
+			/* translators: Accessibility text. */
+			__( 'Please contact your host and ask them to add WebP support', 'performance-lab' )
```
adamsilverstein: Since this is a full sentence, it should end in a period.

adamsilverstein:
Nice work @kirtangajjar - code looks good, I'll give this a test.

adamsilverstein:
This worked well in my testing! 

After testing though, I feel like the notice belongs under "recommended improvements" instead of "critical issues". Having WebP support is nice to have, it isn't really critical.

modules/site-health/webp-support/load.php
```diff
@@ -0,0 +1,64 @@
...
+			__( 'WebP image format is used by WordPress to improve the performance of your site by generating smaller images than it usually could with JPEG format. This means your pages will load faster and consume less bandwidth from users.', 'performance-lab' )
+		),
+		'actions'     => '',
+		'test'        => 'is_webp_uploads_enabled',
+	);
+
+	$webp_supported = wp_image_editor_supports( array( 'mime_type' => 'image/webp' ) );
+
+	if ( ! $webp_supported ) {
+		$result['status']  = 'critical';
```
adamsilverstein: Suggest changing 'critical' to 'recommended'
```suggestion
		$result['status']  = 'recommended';
```

kirtangajjar:
@adamsilverstein Done. LMK if you find any additional changes.

mitogh (approved):

modules/site-health/webp-support/load.php
```diff
@@ -0,0 +1,64 @@
...
+	$result = array(
+		'label'       => __( 'Your site supports WebP', 'performance-lab' ),
+		'status'      => 'good',
+		'badge'       => array(
+			'label' => __( 'Performance', 'performance-lab' ),
+			'color' => 'blue',
+		),
+		'description' => sprintf(
+			'<p>%s</p>',
+			__( 'WebP image format is used by WordPress to improve the performance of your site by generating smaller images than it usually could with JPEG format. This means your pages will load faster and consume less bandwidth from users.', 'performance-lab' )
```
adamsilverstein:
```suggestion
			__( 'WebP image format is used by WordPress to improve the performance of your site by generating smaller images than it usually could with the JPEG format. This means your pages will load faster and consume less bandwidth.', 'performance-lab' )
```
Slight text edit suggestion

adamsilverstein:
@kirtangajjar I left one additional tiny suggestion on the text, then we can merge this. Thanks for your work here!

kirtangajjar:
@adamsilverstein I've committed your suggestion. Thanks for giving it a second look!

adamsilverstein:
👍🏼  Looks good, nice work @kirtangajjar - going to go ahead and merge this. We'll get another chance to review the exact wording when we propose this feature for core.

adamsilverstein (approved):

Issue #142, opened 2022-02-01
When uploading, generate both WebP and jpeg format images by default


adamsilverstein:
## Feature Description

Based on the decision in https://github.com/WordPress/performance/issues/96#issuecomment-1016458629 we need to work on enabling generating both formats by default. 

Specifically we need to address:
```
1. When the user uploads a JPEG image, we generate the smaller image sizes in both JPEG and WebP by default.
2. When inserting an image into a post or including an image in the frontend through any other means, we use the WebP version whenever it is available.
3. Developers can opt out of generating JPEG or WebP, e.g. if they are certain they won't need one of these versions.
```
----

1. Generate both formats
  * We can hook on `wp_generate_attachment_metadata` to generate the additional mime type images (some example [code](https://gist.github.com/erikyo/f0aca6c34fd1d6bc5f73f953bb342a65) from @erikyo). 
  * If the user uploads a jpeg and generates jpeg sub sizes, we would hook in and also generate WebP sizes. If the user uploads a WebP core will generate WebP sub sizes, we would hook in and generate jpeg sub sizes
  * The behavior should be filterable, I propose a new filter named `image_editor_output_formats` which would return an array of mime types. By default this would return `['image/webp', 'image/jpeg']`.
    * The first/default mime type would be used for the default sub-sizes generated by default and stored in the `sizes` array (If only one mime type is given, only that type is generated).
    * When more than one mime type is returned, the second and subsequent types are used to generate additional sub sizes and their data stored in the `additional_sizes` array...

 
1b. In addition to generating the new images, we also need to store meta data for new mime type sizes
  * for backwards compatibility reasons, I suggest we store a single mime type in the `sizes` array
  * ~additional mime type sizes we generate should be stored separately, in a key I propose naming `additional_sizes`~
  * mime type data will go into a `sources` array inside each size
  * additional handling will be required for these additional sizes:
    * when the original image is deleted, these sub sizes should also be deleted 
    * when missing sub sizes are regenerated, this should include all mime types
    * when the image is cropped/rotated/flipped in the media library, the action should apply to these images as well
    * eventually, functions should be extended or added that leverage these new sizes to generate a `picture` element 
    
2. Use WebP on front end 
  * This can be achieved by having the default `sizes` sub sizes be WebP by default, which will be the case with `['image/webp', 'image/jpeg']` as the default value for `image_editor_output_formats`
[...]

Issue #122, opened 2022-01-21
`*.jpg` is not generated for scaled full size image.


mitogh:
## Steps to replicate

- Enable the WebP module
- Upload an image bigger than the `2560` or `big_image_size_threshold`   (Use attached image)
  - https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/includes/image.php#L275
- Upload the image. 
- Observe the full-size image was created with a `*webp`  extension for the full size image.

![2022-01-20_19-02](https://user-images.githubusercontent.com/3921289/150447221-54f92016-940a-4234-9856-3e9de68c6bef.png)

## Image meta data stored in WordPress

<details>
  <summary>Metadata of the image.</summary>

```
=> array(6) {
  ["width"]=>
  int(1707)
  ["height"]=>
  int(2560)
  ["file"]=>
  string(55) "2022/01/philip-myrtorp-hDBdvjk1b0s-unsplash-scaled.webp"
  ["sizes"]=>
  array(7) {
    ["medium"]=>
    array(4) {
      ["file"]=>
      string(48) "philip-myrtorp-hDBdvjk1b0s-unsplash-200x300.webp"
      ["width"]=>
      int(200)
      ["height"]=>
      int(300)
      ["mime-type"]=>
      string(10) "image/webp"
    }
    ["large"]=>
    array(4) {
      ["file"]=>
      string(49) "philip-myrtorp-hDBdvjk1b0s-unsplash-683x1024.webp"
      ["width"]=>
      int(683)
      ["height"]=>
      int(1024)
      ["mime-type"]=>
      string(10) "image/webp"
    }
    ["thumbnail"]=>
    array(4) {
      ["file"]=>
      string(48) "philip-myrtorp-hDBdvjk1b0s-unsplash-150x150.webp"
      ["width"]=>
      int(150)
      ["height"]=>
      int(150)
      ["mime-type"]=>
      string(10) "image/webp"
    }
    ["medium_large"]=>
    array(4) {
      ["file"]=>
      string(49) "philip-myrtorp-hDBdvjk1b0s-unsplash-768x1152.webp"
      ["width"]=>
      int(768)
      ["height"]=>
      int(1152)
      ["mime-type"]=>
      string(10) "image/webp"
    }
    ["1536x1536"]=>
    array(4) {
      ["file"]=>
      string(50) "philip-myrtorp-hDBdvjk1b0s-unsplash-1024x1536.webp"
      ["width"]=>
      int(1024)
[...]

PR #143, opened 2022-02-01
Ensure the `-scaled` image remains in the original uploaded format

### LINKED ISSUE #67: Image mime type is not set correctly when a `webp` image is generated.
## Note

Make sure the `Performance` plugin is enabled and activated, as well with the `webp` modules for images at: `Settings > 
Performance`

Related with issue: 

- #16

## Steps to recreate:

1. Open the WordPress dashboard.
2. Go to `Media > Add New`
2. Upload the image
3. Go to `Media > Library` and open the image, observe the image has an extension `image/jpeg` instead of `image/webp`

## Original image to replicate the issue.

gantas-vaiciulenas-ZLMHrnSyJpM-unsplash

adamsilverstein:
## Summary

When uploading very large images (over the `big_image_size_threshold` filter threshold), WordPress creates a scaled version of the image with the word`-scaled` appended to the file name. It stores a reference to the original image in the image meta `original_image` field, see [this make post](https://make.wordpress.org/core/2019/10/09/introducing-handling-of-big-images-in-wordpress-5-3/#:~:text=The%20default%20threshold%20value%20is,as%20the%20largest%20available%20size.) for more details.

WordPress subsequently uses the "scaled" image as if it were the original in most use cases (only going back to the true original image when editing the image). This causes the mime type shown in the media library for a large jpeg upload to show as image/webp because the scaled image is a WebP.

To fix that, in this PR I skip the `-scaled` image, leaving it in the original format of the uploaded image.

This matches the behavior of uploading a smaller image that isn't over the `big_image_size_threshold`, improving consistency.

Fixes https://github.com/WordPress/performance/issues/67

## Relevant technical choices

* Makes sense that the `scaled` image, which WordPress largely treats as the "original" would remain in the mime type of the upload
* Ensure the behavior is consistent between large and smaller image uploads.
 
## Checklist

### DIFF
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -32,7 +32,7 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
 	}
 
 	// Skip conversion when creating the `-scaled` image (for large image uploads).
-	if ( 1 === preg_match( '/.*-scaled\..{3}.?$/', $filename ) ) {
+	if ( preg_match( '/.*-scaled\..{3}.?$/', $filename ) ) {
 		return $output_format;
 	}
 
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -32,7 +32,7 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
 	}
 
 	// Skip conversion when creating the `-scaled` image (for large image uploads).
-	if ( preg_match( '/.*-scaled\..{3}.?$/', $filename ) ) {
+	if ( preg_match( '/-scaled\..{3}.?$/', $filename ) ) {
 		return $output_format;
 	}
 
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -31,6 +31,11 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
 		return $output_format;
 	}
 
+	// Skip conversion when creating the `-scaled` image (for large image uploads).
+	if ( 1 === preg_match( '/.*-scaled\..{3}.?$/', $filename ) ) {
+		return $output_format;
+	}
+
 	$output_format['image/jpeg'] = 'image/webp';
 
 	return $output_format;
--- a/tests/modules/images/webp-uploads/webp-uploads-test.php
+++ b/tests/modules/images/webp-uploads/webp-uploads-test.php
@@ -8,11 +8,6 @@
 
 class WebP_Uploads_Tests extends WP_UnitTestCase {
 
-	// Filter callback to limit output to WebP for the single output type tests.
-	function return_webp_array() {
-		return array( 'image/webp' );
-	}
-
 	/**
 	 * Test if webp-uploads applies filter based on system support of WebP.
 	 */
@@ -23,9 +18,6 @@ class WebP_Uploads_Tests extends WP_UnitTestCase {
 		} else {
 			$expect = array( 'image/jpeg' => 'image/webp' );
 		}
-		// The image_editor_output_format filter is only applied if image_editor_output_format
-		// has fewer than two types.
-		add_filter( 'image_editor_output_formats', array( $this, 'return_webp_array' ) );
 		$output_format = apply_filters( 'image_editor_output_format', array(), '', 'image/jpeg' );
 		$this->assertEquals( $expect, $output_format );
 	}
@@ -36,7 +28,6 @@ class WebP_Uploads_Tests extends WP_UnitTestCase {
 	function test_webp_uploads_filter_image_editor_output_format_with_support() {
 		// Mock a system that supports WebP.
 		add_filter( 'wp_image_editors', array( $this, 'mock_wp_image_editor_supports' ) );
-		add_filter( 'image_editor_output_formats', array( $this, 'return_webp_array' ) );
 		$output_format = webp_uploads_filter_image_editor_output_format( array(), '', 'image/jpeg' );
 		$this->assertEquals( array( 'image/jpeg' => 'image/webp' ), $output_format );
 	}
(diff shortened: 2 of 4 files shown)


modules/images/webp-uploads/load.php
```diff
@@ -31,6 +31,11 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
 		return $output_format;
 	}
 
+	// Skip conversion when creating the `-scaled` image (for large image uploads).
+	if ( 1 === preg_match( '/.*-scaled\..{3}.?$/', $filename ) ) {
+		return $output_format;
+	}
```
felixarntz, line 37: Can we add a test for this, with a couple different file names?

modules/images/webp-uploads/load.php
```diff
@@ -31,6 +31,11 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
 		return $output_format;
 	}
 
+	// Skip conversion when creating the `-scaled` image (for large image uploads).
+	if ( 1 === preg_match( '/.*-scaled\..{3}.?$/', $filename ) ) {
```
felixarntz: The `1 === ` is maybe a bit overly strict, we can probably remove it. I'm not sure, but I'm thinking maybe some obscure PHP version returns an integer here, another a boolean - in any case, for simple boolean(-ish) checks I don't think we need to have strict checks, similar to how it's done in other such places in WP core.

felixarntz (commented):
Looks good, just two comments.

modules/images/webp-uploads/load.php
mitogh: Similar comment to @felixarntz would be ideal to check `!== false`  instead.

Additionally, the first part of the REGEX can be removed.

```suggestion
	if ( false !== preg_match( '/-scaled\..{3}.?$/', $filename ) ) {
```

Another approach as well can be using substr.

```php
if ( substr( $filename, -11, 7 ) === '-scaled' ) {
```

mitogh:
Additionally if we are going with this route once this is merged we should be able to close: 

- #122

modules/images/webp-uploads/load.php
adamsilverstein: i wanted to avoid substring which might make it harder to match files with .jpeg vs .jpg

adamsilverstein: 👍🏼  updated

adamsilverstein: updated regex. return of false is a regex failure, 0 not found, 1 found, so i changed to a "truthy" check, i'll make sure that works

adamsilverstein:
> Additionally if we are going with this route once this is merged we should be able to close:
> 
> * [`*.jpg` is not generated for scaled full size image. #122](https://github.com/WordPress/performance/issues/122)

Thinking we should generate both, see https://github.com/WordPress/performance/issues/142#issuecomment-1028354600

modules/images/webp-uploads/load.php
adamsilverstein: Added in [`98e1003` (#143)](https://github.com/WordPress/performance/pull/143/commits/98e100336738c0fd4b1ad8c56536b10d1a3825fc)

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -3,9 +3,12 @@
  * Tests for webp-uploads module.
  *
  * @package performance-lab
+ * @group webp-uploads
```
adamsilverstein, line 6: Adding this enables running just these tests locally

felixarntz: Great! Maybe worth adding such an annotation to the other module test class(es) as well? Doesn't really belong here, but might make sense to just add.

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -3,9 +3,12 @@
...
  */
 
 class WebP_Uploads_Tests extends WP_UnitTestCase {
+	function return_webp_array() {
+		return array( 'image/webp' ); }
```
felixarntz: The syntax here looks off? Can you move the closing brace below?

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -17,7 +20,9 @@ function test_webp_uploads_applies_filter_based_on_system_support() {
 		} else {
 			$expect = array( 'image/jpeg' => 'image/webp' );
 		}
-
+		// The image_editor_output_format filter is only applied if image_editor_output_format
+		// has fewer than two types.
+		add_filter( 'image_editor_output_formats', array( $this, 'return_webp_array' ) );
```
felixarntz: Why did you add this here? This existing test shouldn't be affected by this change, or is it?

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -28,6 +33,7 @@ function test_webp_uploads_applies_filter_based_on_system_support() {
 	function test_webp_uploads_filter_image_editor_output_format_with_support() {
 		// Mock a system that supports WebP.
 		add_filter( 'wp_image_editors', array( $this, 'mock_wp_image_editor_supports' ) );
+		add_filter( 'image_editor_output_formats', array( $this, 'return_webp_array' ) );
```
felixarntz: See above, why does this need to be here?

tests/modules/images/webp-uploads/webp-uploads-test.php
felixarntz: Ah I see you changed it now, nevermind.

adamsilverstein: oof phpcbf did that, i was shooting for a single line function which phpcs didn't like; fixed.

tests/modules/images/webp-uploads/webp-uploads-test.php
adamsilverstein: oh, you are right, I thought this was breaking the tests, removing.

tests/modules/images/webp-uploads/webp-uploads-test.php
adamsilverstein: removing

PR #147, opened 2022-02-03
Create image sub-sizes in additional MIME types using `sources` for storage.

### LINKED ISSUE #142: When uploading, generate both WebP and jpeg format images by default

mitogh:
## Summary

The property is added to every single image size available, each value inside the properties array is a reference to
a different image mime type.

If the image size was created as a JPEG version a WebP version would be created, if a WebP version exists a JPEG version
would be created instead based on the details from ticket #142

This mechanism would allow extending the sources property for more additional mime types.

Fixes: 

- #142
- #8 

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
 
 	return $output_format;
 }
-add_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
```
felixarntz, line 43: Nit-pick, but let's keep the `add_filter` / `add_action` calls right below the callback function since that is somewhat of a standard in functional plugins and makes it easy to find which hook a function belongs to.

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+ * sizes. Each property uses the default for each property `null` for height and width and `false`
+ * for crop the same way as `make_subsize` the editor image.
+ *
+ * @since n.e.x.t
+ *
+ * @see   WP_Image_Editor_Imagick::make_subsize()
+ * @see   WP_Image_Editor_GD::make_subsize()
+ * @return array An array with the details of all available image sizes: width, height and crop.
+ */
+function webp_uploads_get_image_sizes() {
```
felixarntz: I don't think this function is needed. WordPress core already has [`wp_get_registered_image_subsizes()`](https://developer.wordpress.org/reference/functions/wp_get_registered_image_subsizes/) which should allow us to get exactly this data.

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+ *
+ * @see   wp_generate_attachment_metadata
+ * @see   webp_uploads_valid_image_mime_types
+ *
+ * @param array $metadata      An array with the metadata from this attachment.
+ * @param int   $attachment_id The ID of the attachment where the hook was dispatched.
+ *
+ * @return array An array with the updated structure for the metadata before is stored in the database.
+ */
+function webp_uploads_create_sources_property( array $metadata, $attachment_id ) {
```
felixarntz, line 28: Since the `wp_generate_attachment_metadata` filter is used for any kind of attachment, we need to include a check on top of this function, to only do something if the attachment is actually an image. If not, we need to just short-circuit the callback and do nothing. It makes sense to use the same check as core does for that, see https://github.com/WordPress/wordpress-develop/blob/5.9.0/src/wp-admin/includes/image.php#L484.

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+			if ( is_wp_error( $image ) ) {
+				continue;
+			}
+
+			// TODO: Store the file size of the created image.
+			// $image['filesize'] = filesize( $image['path'] );
+			// Remove duplicated properties from the size image.
+			unset( $image['path'], $image['height'], $image['width'], $image['mime-type'] );
+
+			$sources[ $mime ] = $image;
```
felixarntz: A bit safer than unsetting all "assumed" fields would be to create the array from scratch below, to ensure that it really only includes the `file` and `filesize` fields.

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+		if ( ! array_key_exists( $current_mime, $valid_mime_types ) ) {
+			continue;
+		}
+
+		$sources[ $current_mime ] = array(
+			'file' => array_key_exists( 'file', $current_size ) ? $current_size['file'] : '',
+			// TOOD: Add filesize from the original version of this image.
+		);
+
+		$formats = array_diff_assoc( $valid_mime_types, array( $current_mime => $valid_mime_types[ $current_mime ] ) );
```
felixarntz: This can be simplified to
```suggestion
		$formats = array_diff_key( $valid_mime_types, array( $current_mime => true ) );
```
It doesn't matter what the value is, and it's redundant to look up the value in `$valid_mime_types`.

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+ * @since n.e.x.t
+ *
+ * @todo Add a filter to support more mime types.
+ *
+ * @return string[] An array of valid mime types, where the key is the mime type and the value is the extension type.
+ */
+function webp_uploads_valid_image_mime_types() {
+	return array(
+		'image/jpeg' => 'jpg',
+		'image/webp' => 'wepb',
```
felixarntz: Typo: --> `webp`

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+	);
+}
+
+add_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
+add_filter( 'wp_generate_attachment_metadata', 'webp_uploads_create_sources_property', 10, 2 );
```
felixarntz: See my comment above, let's move the `add_filter` calls right below their respective function.

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+		$sizes[ $size ]          = $current_size;
+	}
+
+	$metadata['sizes'] = $sizes;
+
+	return $metadata;
+}
+
+/**
+ * List all the available image sizes as well with the properties for each, the properties included
```
felixarntz: Nit-pick, but any function doc-block descriptions in WordPress core should start with a 3rd-person verb:
```suggestion
 * Lists all the available image sizes as well with the properties for each, the properties included
```

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+			'height' => get_option( $size . '_size_h', $height ),
+			'crop'   => (bool) get_option( $size . '_crop', $crop ),
+		);
+	}
+
+	return $sizes;
+}
+
+/**
+ * Return an array with the list of valid mime types for a sources' property.
```
```suggestion
 * Returns an array with the list of valid mime types for a sources' property.
```

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+ *
+ * @since n.e.x.t
+ *
+ * @todo Add a filter to support more mime types.
+ *
+ * @return string[] An array of valid mime types, where the key is the mime type and the value is the extension type.
+ */
+function webp_uploads_valid_image_mime_types() {
```
felixarntz: We should think about a better name for this function, it's not really about _valid_ image MIME types. For example, PNG is also a valid MIME type.

Also, I think it would be more appropriate to have this function return a _map_ of which image MIME types can be transformed to another image MIME types, since that's what we're actually looking for here. Storing the extensions here is duplicated information, we can already get this from WordPress core's `wp_get_mime_types()` function.

We need to make sure this function also makes sense for the future, and what we really care about here is which image sizes core should support generating versions in another MIME type for.

Because of the above, I think we should change this to a function that returns a map of supported MIME type "transformations", e.g.
```php
return array(
	'image/jpeg' => array( 'image/webp' ),
	// And potentially, if we also decide to support that:
	'image/webp' => array( 'image/jpeg' ),
);
```

That would easily allow to in the future also control which file types should be transformed to AVIF, or to also transform PNG images to WebP if that becomes possible.

Of course the logic above that consumes this function would then need to be adjusted: We could check whether the original MIME type is one of the keys in this "transformation" array, and then we generate sub-sizes for all MIME types that are in the value array.

The function could be called something like `webp_uploads_get_supported_image_mime_transforms()`.

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+	// All subsizes are created out of the original image.
+	$file = wp_get_original_image_path( $attachment_id, true );
+
+	// File does not exist.
+	if ( ! file_exists( $file ) ) {
+		return $metadata;
+	}
+
+	// Prevent to convert JPEG to WebP if we are creating JPEG versions of the image.
+	remove_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format' );
```
felixarntz: I'm not sure about this here - shouldn't it be in a conditional? If the user uploads a JPEG image, we _do_ need to create WebP sub-sizes here by default, so why remove the filter?

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+		}
+
+		$sources[ $current_mime ] = array(
+			'file' => array_key_exists( 'file', $current_size ) ? $current_size['file'] : '',
+			// TOOD: Add filesize from the original version of this image.
+		);
+
+		$formats = array_diff_assoc( $valid_mime_types, array( $current_mime => $valid_mime_types[ $current_mime ] ) );
+
+		foreach ( $formats as $mime => $extension ) {
```
felixarntz: Before going through this loop of formats, we should filter out any formats not supported by the environment, similar to how the above existing filter callback does, e.g. `wp_image_editor_supports( array( 'mime_type' => 'image/webp' ) )`. If the format is not supported, we shouldn't even run the logic to try generating image sizes for it.

This could mean that there are no more image sizes to generate, in that case the `sources` item should only be updated with the data for the original MIME type (e.g. JPEG).

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+
+		// Make sure the current mime is consider a valid mime type.
+		if ( ! array_key_exists( $current_mime, $valid_mime_types ) ) {
+			continue;
+		}
+
+		$sources[ $current_mime ] = array(
+			'file' => array_key_exists( 'file', $current_size ) ? $current_size['file'] : '',
+			// TOOD: Add filesize from the original version of this image.
+		);
```
felixarntz: Before populating this, we should first check that `$sources[ $current_mime ]` isn't already set. We only need to do this the very first time we process the image size data.

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,186 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+
+		$formats = array_diff_assoc( $valid_mime_types, array( $current_mime => $valid_mime_types[ $current_mime ] ) );
+
+		foreach ( $formats as $mime => $extension ) {
+			// Editor needs to be recreated every time as there is not flush() or clear() function that can be used after we created an image.
```
felixarntz: Similar to my above comment, we should only do anything here if this image size does not yet exist in this MIME type. Otherwise we'll endlessly run duplicate processes to create image sizes. This is critical because the `wp_generate_attachment_metadata` filter is also invoked when missing image sizes are being created (see also WordPress core's `wp_get_missing_image_subsizes()` function), so we have to also only create the image sizes here that _do not_ yet exist in the format.

Basically add this:
```suggestion
			if ( isset( $sources[ $mime ] ) ) {
				continue;
			}

			// Editor needs to be recreated every time as there is not flush() or clear() function that can be used after we created an image.
```

felixarntz (commented):
@mitogh This looks like a good start, however there are a couple things that need to be refined. I left some comments.

modules/images/webp-uploads/load.php
mitogh: Great I wasn't aware of that one.

felixarntz:
Summarizing a chat conversation on this between @mitogh, @adamsilverstein and myself:
* We're in agreement to proceed with this approach since it's a more appropriate way to store the additional MIME type images, we can probably close the alternative approach #154 then.
* One high-level enhancement that still needs to be added to this PR before it can be merged is ensure that the new files are also being deleted when the image is deleted. We need to hook into the `delete_attachment` action for that.
* The other thing that needs to be added is that for every new file generated, we need to update the attachment metadata immediately after, to allow for the process to fail gracefully and to be able to proceed with only the missing sizes on next invokation:
    * Introduce a function to check for which sizes are still missing in the extra MIME type(s) to generate. For this we can probably create a copy of WP core's `wp_get_missing_image_subsizes()` function and add a MIME type parameter to it. For an eventual core merge, the real function could then be enhanced similarly.
    * Adjust the logic in our `wp_generate_attachment_metadata` filter callback to always only generate versions for the missing MIME types, using the above function.
    * Call `wp_update_attachment_metadata()` every time a new image sub-size file is generated, i.e. store the `sources` field just for that image sub-size file generated and immediately store it.
* In a separate PR later, we need to integrate this approach also with the feature of editing images, i.e. when an image is edited within WordPress, we also need to generate versions in the additional MIME types for the _edited_ versions, and we need to make sure the original `sources` are then included in the `_wp_attachment_backup_sizes`.

modules/images/webp-uploads/load.php
mitogh: Correct this has been included already here: https://github.com/WordPress/performance/pull/147/files/25479a1530536631ca39d5bf21e2c63202d506ce#diff-71ed188a16b79d179868ee80e76f775462a451e871fc3d6214fcacca3edcea8cR62

modules/images/webp-uploads/load.php
mitogh: Good idea :+1:

modules/images/webp-uploads/load.php
mitogh: Right, the way the filter works it would convert any JPEG to WebP, so in the scenario when a user uploads a WebP image we need to generate JPEG subsizes instead. 

If the filter remains as it is, when we aim to create a JPEG effectively is going to create a WebP image instead, so when this is the case we need to make sure that does not happen.

Agree I think this can be more robust in terms of only removing it when we are actually creating a JPEG image.

modules/images/webp-uploads/load.php
mitogh: I think this would be true for new uploads, subsequent edits into an image need to override this key if present to include information about the newly uploaded image instead, but good point. We can add a condition before the image is uploaded.

modules/images/webp-uploads/load.php
mitogh: Agree, There's a third parameter that indicates if this comes from `create` or `update` called context that we can use to optimize this condition for those use cases.

modules/images/webp-uploads/load.php
mitogh: Oh, absolutely, great point.

modules/images/webp-uploads/load.php
mitogh: Agree, having a map makes more sense in the long term to support additional mime types for a defined mime type.

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
-function webp_uploads_filter_image_editor_output_format( $output_format, $filename, $mime_type ) {
-	// Only enable if the server supports WebP.
-	if ( ! wp_image_editor_supports( array( 'mime_type' => 'image/webp' ) ) ) {
-		return $output_format;
+function webp_uploads_create_sources_property( array $metadata, $attachment_id ) {
+	// This should take place only on the JPEG image.
+	$valid_mime_transforms = webp_uploads_get_supported_image_mime_transforms();
+
+	// Not a supported mime type to create the sources property.
+	if ( ! array_key_exists( get_post_mime_type( $attachment_id ), $valid_mime_transforms ) ) {
```
felixarntz: Nit-pick, but `isset` is a tiny bit faster than `array_key_exists`, and the latter is not needed here.
```suggestion
	$mime_type = get_post_mime_type( $attachment_id );
	if ( ! isset( $valid_mime_transforms[ $mime_type ] ) ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+	$image_sizes = array();
+	if ( array_key_exists( 'sizes', $metadata ) && is_array( $metadata['sizes'] ) ) {
+		$image_sizes = $metadata['sizes'];
+	}
+
+	$sizes = array();
+
+	foreach ( wp_get_registered_image_subsizes() as $size_name => $properties ) {
+		// This image size does not exist on the defined sizes.
+		if ( ! array_key_exists( $size_name, $image_sizes ) || ! is_array( $image_sizes[ $size_name ] ) ) {
```
```suggestion
		if ( ! isset( $image_sizes[ $size_name ] ) || ! is_array( $image_sizes[ $size_name ] ) ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+			continue;
+		}
+
+		$current_size = $image_sizes[ $size_name ];
+		$sources      = array();
+		if ( array_key_exists( 'sources', $current_size ) && is_array( $current_size['sources'] ) ) {
```
```suggestion
		if ( isset( $current_size['sources'] ) && is_array( $current_size['sources'] ) ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+		return $metadata;
+	}
+
+	$dirname     = pathinfo( $file, PATHINFO_DIRNAME );
+	$image_sizes = array();
+	if ( array_key_exists( 'sizes', $metadata ) && is_array( $metadata['sizes'] ) ) {
+		$image_sizes = $metadata['sizes'];
+	}
+
+	$sizes = array();
```
felixarntz: Why an extra variable here? This seems risky and error-prone. It would be safer to modify the `$image_sizes` variable at all time. Otherwise any tiny bug would result in the original data being wiped. In addition to that, using an extra variable here complicates the code.

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+		// Try to find the mime type of the image size.
+		$current_mime = '';
+		if ( array_key_exists( 'mime-type', $current_size ) ) {
+			$current_mime = $current_size['mime-type'];
+		} elseif ( array_key_exists( 'file', $current_size ) ) {
+			$current_mime = wp_check_filetype( $current_size['file'] )['type'];
+		}
+
+		if ( empty( $current_mime ) ) {
+			$sizes[ $size_name ] = $current_size;
```
felixarntz: See above, let's get rid of the extra `$sizes` variable. So this line here can be removed for example.

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+
+		$current_size = $image_sizes[ $size_name ];
+		$sources      = array();
+		if ( array_key_exists( 'sources', $current_size ) && is_array( $current_size['sources'] ) ) {
+			$sources = $current_size['sources'];
+		}
+
+		// Try to find the mime type of the image size.
+		$current_mime = '';
+		if ( array_key_exists( 'mime-type', $current_size ) ) {
```
```suggestion
		if ( isset( $current_size['mime-type'] ) ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+			$current_mime = $current_size['mime-type'];
+		} elseif ( array_key_exists( 'file', $current_size ) ) {
```
```suggestion
		} elseif ( isset( $current_size['file'] ) ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+		$sources[ $current_mime ] = array(
+			'file'     => array_key_exists( 'file', $current_size ) ? $current_size['file'] : '',
+			'filesize' => 0,
+		);
+
+		// Set the filesize from the current mime image.
+		$file_location = path_join( $dirname, $sources[ $current_mime ]['file'] );
+		if ( file_exists( $file_location ) ) {
+			$sources[ $current_mime ]['filesize'] = filesize( $file_location );
+		}
```
felixarntz: This whole block here should be wrapped in a condition, we only need to run it `if ( ! isset( $sources[ $current_mime ] )`.

Otherwise it will run this logic over and over again although the data is already set.

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+
+		$formats = array_key_exists( $current_mime, $valid_mime_transforms )
+			? $valid_mime_transforms[ $current_mime ]
+			: array();
```
```suggestion
		$formats = isset( $valid_mime_transforms[ $current_mime ] ) ? $valid_mime_transforms[ $current_mime ] : array();
```

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+
+		foreach ( $formats as $mime ) {
+			wp_schedule_single_event( time(), 'webp_uploads_create_image', array( $attachment_id, $size_name, $mime ) );
+		}
```
felixarntz: The code here is still missing the critical check for whether we actually need to generate the format still. Since the process may fail at any point, we need to always make sure we only generate the sizes which are missing, see also my feedback in https://github.com/WordPress/performance/pull/147#issuecomment-1043145067.

Also, why a cron event? WordPress core doesn't use a cron event to generate its image sizes, and I think we should follow how it generates its default sub-sizes, which happens in a single request, just allowing for failures as far as I can tell, and then it retries until everything has been completed (see the `_wp_make_subsizes()` function and how it is used).

We also need to make sure then that we keep using the latest version of `$sources`, since it could be updated within the function, so your `webp_uploads_generate_image_size()` function below should probably return the source array that it has added.

How about this:
```suggestion
		foreach ( $formats as $mime ) {
			if ( isset( $sources[ $mime ] ) ) {
				continue;
			}
			$new_source = webp_uploads_generate_image_size( $attachment_id, $size_name, $mime );
			if ( is_wp_error( $new_source ) ) {
				continue;
			}
			$sources[ $mime ] = $new_source;
		}
```

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+
+		$current_size['sources'] = $sources;
+		$sizes[ $size_name ]     = $current_size;
+	}
+
+	$metadata['sizes'] = $sizes;
```
felixarntz: See above, this can be simplified to
```suggestion
		$image_sizes[ $size_name ]['sources'] = $sources;
	}

	$metadata['sizes'] = $image_sizes;
```

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+		return new WP_Error( 'image_mime_type_invalid_metadata', __( 'The image does not have a valid metadata.', 'performance-lab' ) );
+	}
+
+	// All subsizes are created out of the attached file.
+	$file = get_attached_file( $attachment_id );
+
+	// File does not exist.
+	if ( ! file_exists( $file ) ) {
+		return new WP_Error( 'image_file_size_not_found', __( 'The provided size does not have a valid image file.', 'performance-lab' ) );
+	}
```
felixarntz, line 137: This is a bit unnecessary. How about we make the `$file` a required parameter of the function, since we have it available outside anyway? That's also similar to WP core's `_wp_make_subsizes` where the file is passed to as well.

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+		$metadata['sizes'][ $size ]['sources'] = array();
 	}
 
-	$output_format['image/jpeg'] = 'image/webp';
+	$metadata['sizes'][ $size ]['sources'][ $mime ] = array(
+		'file'     => $image['file'],
+		'filesize' => array_key_exists( 'path', $image ) ? filesize( $image['path'] ) : 0,
+	);
+
+	return wp_update_attachment_metadata( $attachment_id, $metadata );
```
felixarntz: See my comment above, we should return the new source here so that it can also be processed outside, so that the data in the above function remains intact.
```suggestion
	$new_source = array(
		'file'     => $image['file'],
		'filesize' => isset( $image['path'] ) ? filesize( $image['path'] ) : 0,
	);

	// Update metadata right away with the new source in case the process fails later.
	$metadata['sizes'][ $size ]['sources'][ $mime ] = $new_source;
	wp_update_attachment_metadata( $attachment_id, $metadata );

	return $new_source;
```

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+/**
+ * Hook fired when an attachment is deleted, this hook is in charge of removing any
+ * additional mime types created by this plugin besides the original image. Any source
+ * with the same as the main image would not be removed by this hook due this file would
+ * be removed by WordPress when the attachment is deleted, usually this happens after this
+ * hook is executed.
+ *
+ * @since n.e.x.t
+ *
+ * @see   wp_delete_attachment
```
```suggestion
 * @see wp_delete_attachment
```

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+ *
+ * @param int $attachment_id The ID of the attachment the sources are going to be deleted.
+ *
+ * @return void
```
felixarntz: `void` return type docs are typically omitted in WP core.
```suggestion
 * @param int $attachment_id The ID of the attachment the sources are going to be deleted.
```

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+ */
+function webp_uploads_remove_sources_files( $attachment_id ) {
+	$metadata = wp_get_attachment_metadata( $attachment_id );
+
+	if (
+		! isset( $metadata['sizes'], $metadata['file'] )
+		|| empty( $metadata['file'] )
+		|| ! is_array( $metadata['sizes'] )
+	) {
```
felixarntz: Let's use `get_attached_file()` instead of looking at `$metadata['file']` as this is how core does it, and that function is filtered so we should stick to it.

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+		return;
+	}
+
+	$upload_path = wp_get_upload_dir();
+	if ( empty( $upload_path['basedir'] ) ) {
+		return;
+	}
+
+	$file_directory = dirname( $metadata['file'] );
+	$directory      = path_join( $upload_path['basedir'], $file_directory );
```
felixarntz: Nit-pick, but IMO it would be a good idea to stick to WP core names (see `wp_delete_attachment_files()` function) to clarify the variable names:
```suggestion
	$intermediate_dir = path_join( $upload_path['basedir'], dirname( $metadata['file'] ) );
```

modules/images/webp-uploads/load.php
```diff
@@ -5,40 +5,294 @@
...
+				continue;
+			}
+
+			if ( ! is_array( $properties ) || empty( $properties['file'] ) ) {
+				continue;
+			}
+
+			$file_deletion = path_join( $file_directory, $properties['file'] );
+			$file_deletion = path_join( $upload_path['basedir'], $file_deletion );
+			wp_delete_file_from_directory( $file_deletion, $directory );
```
felixarntz: See above - also, more importantly, the logic here looks different from WP core's - are you sure that works? Here's how it would be, following WP core's `wp_delete_attachment_files()`:
```suggestion
			$intermediate_file = str_replace( wp_basename( $metadata['file'] ), $properties['file'], $metadata['file'] );
			if ( ! empty( $intermediate_file ) ) {
				$intermediate_file = path_join( $uploadpath['basedir'], $intermediate_file );
				wp_delete_file_from_directory( $intermediate_file, $intermediate_dir );
			}
```

felixarntz (commented):
@mitogh This looks really close now, awesome work! I left a few more comments. Most are minor, but there are a few more substantial things that need to be revisited, for example the usage of a cron job which seems unnecessary and potentially problematic to me. See individual comments for details.

modules/images/webp-uploads/load.php
mitogh: Good point.

modules/images/webp-uploads/load.php
mitogh: Mostly to increase readability, but I see the point here so let me update it to do that.

modules/images/webp-uploads/load.php
mitogh: Actually, this is already covered here: 

https://github.com/WordPress/performance/pull/147/files/6e3afeef0648073c2615aa15f4d0e090f0b3b93f#diff-71ed188a16b79d179868ee80e76f775462a451e871fc3d6214fcacca3edcea8cR47

As if the size has not been created yet the additional mime types for that image won't be created. 

> Since the process may fail at any point, we need to always make sure we only generate the sizes which are missing, see also my feedback in https://github.com/WordPress/performance/pull/147#issuecomment-1043145067.

Once the image is retried and finally generated it would trigger this hook again, which at that point would contain the key for that particular image and then the image would be scheduled for creation. 

> Also, why a cron event? 

To split the work for every single size in a different process, and only use as minimal resources as possible per image size so we can decrease the number of potential failures when creating an image in a constrained environment.

> WordPress core doesn't use a cron event to generate its image sizes

It does not because it has to process the image in a single request and display the image right away, we as progressive enchantment can do that as we know a fallback image was already created. And we can create a background process (Cron) for any additional mime images.

This mechanism would ensure existing users would continue to use the same number of resources when uploading an image as they currently do, without the overhead of additional resources introduced by a new mime type. 

> just allowing for failures as far as I can tell, and then it retries until everything has been completed (see the _wp_make_subsizes() function and how it is used).

The failures would be handled by this hook as described at the beginning every time an image fails won't be added into the image sizes, and every time a new image size is added it would trigger this hook, ensuring that the image is only scheduled for creation when the main image was created.

> We also need to make sure then that we keep using the latest version of $sources, since it could be updated within the function, so your webp_uploads_generate_image_size() function below should probably return the source array that it has added.

This is already handled here by just returning the updated metadata and the hook dispatched by this function would update the meta once is executed. 

- https://github.com/WordPress/performance/pull/147/files/6e3afeef0648073c2615aa15f4d0e090f0b3b93f#diff-71ed188a16b79d179868ee80e76f775462a451e871fc3d6214fcacca3edcea8cR103

modules/images/webp-uploads/load.php
mitogh: Right this is actually helpful as if this run when the image is updated we ensure that the `sources` property points to the right image instead of the older one, as this won't be updated in that scenario.

Subsequent updates from:

- #158 

Would handle additional cases like _should previous sources be stored in the backup sizes or delete those.?_

modules/images/webp-uploads/load.php
mitogh: That could be a possibility, however by introducing a new parameter we are allowing external callers to define a different base image for the additional mime types, and in this particular case we want to make sure the additional mime types are created out of the attached file.

In addition to that the current API allows a simplified call from external callers to create new additional mime types without much complexity.

Examples:

- https://github.com/WordPress/performance/blob/0e94c15cc5d6164e14950c19847d370738422190/tests/modules/images/webp-uploads/webp-uploads-test.php#L229
- https://github.com/WordPress/performance/blob/0e94c15cc5d6164e14950c19847d370738422190/tests/modules/images/webp-uploads/webp-uploads-test.php#L246
- https://github.com/WordPress/performance/blob/0e94c15cc5d6164e14950c19847d370738422190/tests/modules/images/webp-uploads/webp-uploads-test.php#L261

modules/images/webp-uploads/load.php
mitogh: Right, this behavior is to allow an `async` mechanism rather than a single request.

modules/images/webp-uploads/load.php
mitogh: The current logic works, but I can verify that particular case from core.

modules/images/webp-uploads/load.php
adamsilverstein: Although the `wp_schedule_single_event` approach works in my testing, I wonder if it will work across WordPress installs or might create problems for some users, for example does this work if `define( 'DISABLE_WP_CRON', true ); ` is set?

I think we should test Felix's suggestion of trying to generate all images in one go, tracking the completed ones in meta . Core's upload process (in both media and the block editor) will catch incomplete generation runs (a timeout error) and send a retry request (which we may need to hook into?). I will try testing this approach with a low server timeout to verify.

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,290 @@
...
+ *
+ * @since n.e.x.t
+ *
+ * @param int    $attachment_id The ID of the attachment we are going to use as a reference to create the image.
+ * @param string $size          The size name that would be used to create this image, out of the registered subsizes.
+ * @param string $mime          A mime type we are looking to use to create this image.
+ *
+ * @return bool|int|WP_Error
  */
-function webp_uploads_filter_image_editor_output_format( $output_format, $filename, $mime_type ) {
```
adamsilverstein, line 23: rather than deleting the existing function, we may want to keep it around for sites that want to generate **only** WebP sub sized images (which is what the current code does)

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,290 @@
...
+		$height = $sizes[ $size ];
+	}
+
+	if ( array_key_exists( 'crop', $sizes[ $size ] ) ) {
+		$crop = (bool) $sizes[ $size ]['crop'];
+	}
+
+	$editor->resize( $width, $height, $crop );
+	$filename = $editor->generate_filename( null, null, $extension );
+	$filename = preg_replace( '/-(scaled|rotated|imagifyresized)/', '', $filename );
```
adamsilverstein, line 180: when the source file is large we should generate a WebP version of the `-scaled` image as well, we need that for markup. Currently only the jpeg version is generated.

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,290 @@
...
+
+			$intermediate_file = str_replace( $basename, $properties['file'], $file );
+			if ( ! empty( $intermediate_file ) ) {
+				$intermediate_file = path_join( $upload_path['basedir'], $intermediate_file );
+				wp_delete_file_from_directory( $intermediate_file, $intermediate_dir );
+			}
+		}
+	}
 }
-add_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
```
adamsilverstein, line 43: leave this for when `webp_uploads_get_supported_image_mime_transforms()` returns only webp.

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,290 @@
...
+ *
+ * @return array An array with the updated structure for the metadata before is stored in the database.
+ */
+function webp_uploads_create_sources_property( array $metadata, $attachment_id ) {
+	// This should take place only on the JPEG image.
+	$valid_mime_transforms = webp_uploads_get_supported_image_mime_transforms();
+
+	// Not a supported mime type to create the sources property.
+	$mime_type = get_post_mime_type( $attachment_id );
+	if ( ! isset( $valid_mime_transforms[ $mime_type ] ) ) {
```
adamsilverstein, line 34: if the mime_type is jpeg and webp_uploads_get_supported_image_mime_transforms returns only webp, i would expect only webp images to be output (current behavior)

modules/images/webp-uploads/load.php
adamsilverstein: i encountered a big with the callback approach: after uploading an image, I deleted it before WebP creation completed. A final WebP was generated after the deletion.

testing with a short server timeout, WordPress's retry mechanism worked as expected, and the cron callback was set up after the core image process completed.

mitogh: > Although the wp_schedule_single_event approach works in my testing, I wonder if it will work across WordPress installs or might create problems for some users, for example does this work if define( 'DISABLE_WP_CRON', true ); is set?

Usually when this is the case, is to run the corn with an actual CRON via the server which will still execute any remaining cron job. However, if the CRON is completely disabled from the server any additional functionality from WordPress won't work either like: 

- Scheduling publishing 
- Automatic updates from WP and plugins.

> I think we should test Felix's suggestion of trying to generate all images in one go, tracking the completed ones in meta . Core's upload process (in both media and the block editor) will catch incomplete generation runs (a timeout error) and send a retry request (which we may need to hook into?). I will try testing this approach with a low server timeout to verify.

Can you share your setup PHP.ini so I can try to replicate it?

> i encountered a big with the callback approach: after uploading an image, I deleted it before WebP creation was completed. A final WebP was generated after the deletion.

Oh, great point I think this can be an edge case but still would be ideal to make sure any remaining cron is removed when the image is removed as well.

modules/images/webp-uploads/load.php
mitogh: I think this is being discussed here: 

- https://github.com/WordPress/performance/issues/174

modules/images/webp-uploads/load.php
mitogh: > rather than deleting the existing function, we may want to keep it around for sites that want to generate only WebP sub sized images (which is what the current code does)

This is a great point. Question how can we answer this one from the code?

> sites that want to generate only WebP sub sized images (which is what the current code does

Should we rely on the existing setting from the plugin for this scenario?

modules/images/webp-uploads/load.php
mitogh: Well the transform won't define which mime types are allowed instead allows to define in which mime types the provided mime type can be transformed, if the mime is not present on the array no additional mime type would be created for that image but the image would be uploaded as it was uploaded with no additional change.

modules/images/webp-uploads/load.php
mitogh: See the answer above.

modules/images/webp-uploads/load.php
felixarntz: @mitogh 
> > Also, why a cron event?
>
> To split the work for every single size in a different process, and only use as minimal resources as possible per image size so we can decrease the number of potential failures when creating an image in a constrained environment.

While I get the thinking behind this, we have to keep in mind we're building this for WordPress core, which already has an existing working solution for generating sub-sizes in a scalable way. WP core does not use cron for this, and if we wanted to change that, it would open up a big discussion on whether that makes sense or not, which is not the point we're trying to make here - we want to add support for adding sub-sizes also in WebP.

We should really minimize any sort of "refactoring" that is unrelated to the problem we're trying to solve. We can tie into the existing logic and follow the same approach that WP core has here without running into problems - and as @adamsilverstein has pointed out, using a different approach here potentially causes other problems and it will require a lot of extra testing - in contrast to using WP core's existing approach, which we know works.

For the above reasons, I still think it would be better to drop the cron usage here and just call the function right away, as that follows the existing approach in WP core.

modules/images/webp-uploads/load.php
felixarntz: @mitogh That's a fair point. The decision depends a bit on whether we consider this function a "public" API function, or whether it's more for internal usage, like WP core's `_wp_make_subsizes()` - if it's the latter, my suggestion would make sense.

Given that this is currently more of a "public" API function, I agree we can leave this as is - the overhead of looking up the attached file should be minimal.

modules/images/webp-uploads/load.php
felixarntz: See my comment above - I think we should change that to use a single request while supporting retry attempts, thus following how WP core does it.

modules/images/webp-uploads/load.php
felixarntz: @mitogh @adamsilverstein I think for now it's okay to remove the function since at this point we're focusing on the default implementation. However, we need to iterate on this in a follow-up issue (and later PR): We have to define a filter where developers can determine which formats they want to generate their images in. Once we introduce such a filter, we can reintroduce a filter for the output format, based on that filter's value.

modules/images/webp-uploads/load.php
felixarntz: Yeah, let's keep that out of this PR and review in #174 to follow up if needed.

modules/images/webp-uploads/load.php
felixarntz: See my response in https://github.com/WordPress/performance/pull/147#discussion_r813265136 - right now this is without a purpose because the new default behavior is different. We should reintroduce something like this again once we have actually defined the API for developers to tweak the behavior. We can then use this function as a starting point to tweak.

modules/images/webp-uploads/load.php
felixarntz: See https://github.com/WordPress/performance/pull/147#discussion_r813265136, this PR changes the default implementation, we don't yet have any proper way for developers to indicate which formats they'd like to generate, so I think we should explore that in a follow-up issue to not overcomplicate this (already complex) issue/PR further.

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,290 @@
...
+ * `webp_uploads_valid_image_mime_types` the image is just added to the `sources` property and  not
+ * created again. If the uploaded attachment is not a valid image this function does not alter the
+ * metadata of the attachment, on the other hand a `sources` property is added.
  *
- * For uploaded JPEG images, map the default output format to WebP.
+ * @since n.e.x.t
  *
- * @since 1.0.0
+ * @see   wp_generate_attachment_metadata
+ * @see   webp_uploads_valid_image_mime_types
```
```suggestion
 * @see   wp_generate_attachment_metadata()
 * @see   webp_uploads_valid_image_mime_types()
```

modules/images/webp-uploads/load.php
adamsilverstein: I did some testing with a low php timeout threshold using both approaches. Unfortunately, the way we hook in now is _after core completes its retries_, so our generation code only gets one run. With a low timeout, this means not all sizes will be generated. We could either change how we hook in so it happens earlier (before completion so the error is caught), or we could hook in later and return an error if mime/sizes are still missing:

For Ajax hooking in with the `wp_prepare_attachment_for_js` [filter](url), which is the last thing the ajax callback [calls](https://github.com/WordPress/wordpress-develop/blob/74d7ef73b361fff82605291282588dc05059dc27/src/wp-admin/includes/ajax-actions.php#L2476) (returning false when mime/sizes are missing would trigger another retry)

I found the same issue in Gutenberg, the retry 'create-image-subsizes' was only called until all the jpg images are generated, then one attempt at webp images, the retry did not continue because the [code](https://github.com/WordPress/gutenberg/blob/ea471c0161a3e3bb9af7424c85985f7b8be8aa67/packages/api-fetch/src/middlewares/media-upload.js#L45-L60) only retries until a success return. We may be able to use the `rest_prepare_attachment` filter (in `src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php::L843). Returning an Error when the attachment has missing mime/sizes woulds trigger the retry.

Maybe we can also bump these retries up on the trac ticket.

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,290 @@
...
  *
- * @param string $output_format The image editor default output format mapping.
- * @param string $filename      Path to the image.
- * @param string $mime_type     The source image mime type.
- * @return string The new output format mapping.
+ * @param array $metadata      An array with the metadata from this attachment.
+ * @param int   $attachment_id The ID of the attachment where the hook was dispatched.
+ *
+ * @return array An array with the updated structure for the metadata before is stored in the database.
```
felixarntz: There shouldn't be blank lines between `@param` and `@return` in WP core.
```suggestion
 * @param int   $attachment_id The ID of the attachment where the hook was dispatched.
 * @return array An array with the updated structure for the metadata before is stored in the database.
```

modules/images/webp-uploads/load.php
felixarntz: Let's handle that part once we actually get to #158 - at this point we don't support editing images, so it's better to focus on exclusively the goal of #142. Right now, executing this logic on every call of this function is unnecessary - we only need to populate the `$sources` value for the existing (primary) image format once, only _if_ it doesn't exist yet.

#158 will show if this is really needed or not - probably the implementation there will have to use different hooks or behavior anyway.

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,290 @@
...
+ * this image would be stored in the same place as the provided size name inside the
+ * metadata of the attachment.
+ *
+ * @since n.e.x.t
+ *
+ * @param int    $attachment_id The ID of the attachment we are going to use as a reference to create the image.
+ * @param string $size          The size name that would be used to create this image, out of the registered subsizes.
+ * @param string $mime          A mime type we are looking to use to create this image.
+ *
+ * @return bool|int|WP_Error
```
```suggestion
 * @param string $mime          A mime type we are looking to use to create this image.
 * @return bool|int|WP_Error
```

Also though, related to my other comments of dropping the cron usage, we may need to update the return docs, if we return the new source from this function (see https://github.com/WordPress/performance/pull/147/files#r812448741).

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,290 @@
...
+
+	/**
+	 * Filter to allow the definition of a custom mime types, in which a defined mime type
+	 * can be transformed and provide a wide range of mime types.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array<string, array<string>> An array with the valid mime transforms.
+	 *
+	 * @return array<string, array<string>> An array with the valid mime transformation
```
felixarntz: This is missing the variable name, which needs to be included in filter param docs. What it does _not_ need is the `@return` doc, filters don't have that. Also arrays in WP core can simply be documented as `array`, or if they are a list of just one data type like `[]string`.
```suggestion
	 * @param array $image_mime_transforms A map with the valid mime transforms.
```

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,290 @@
...
+/**
+ * Hook fired when an attachment is deleted, this hook is in charge of removing any
+ * additional mime types created by this plugin besides the original image. Any source
+ * with the same as the main image would not be removed by this hook due this file would
+ * be removed by WordPress when the attachment is deleted, usually this happens after this
+ * hook is executed.
+ *
+ * @since n.e.x.t
+ *
+ * @see wp_delete_attachment
```
```suggestion
 * @see wp_delete_attachment()
```

felixarntz (commented):
@mitogh I left a few last nit-pick comments. Other than those, the one critical thing here is that I still think the cron usage isn't really appropriate here, since we should tie into WP core behavior, which doesn't have that. See my comment https://github.com/WordPress/performance/pull/147/files#r813258590 for more details.

modules/images/webp-uploads/load.php
adamsilverstein: 👍🏼

modules/images/webp-uploads/load.php
adamsilverstein: 👍🏼

modules/images/webp-uploads/load.php
mitogh: Add parenthesis to method references.

modules/images/webp-uploads/load.php
adamsilverstein: I'm fine merging Felix's suggested code, then working on ensuring error catching (and retries) work correctly for secondary types as well. 

It might be worth looking at hooking on `wp_update_attachment_metadata`, which is called for each image core generates, passing the full meta with each call, with the last item in the sizes array the most recently generated image. If our image generation threw a timeout error here it could trigger the retry.

`wp_update_attachment_metadata` is also called by image edits which could be handy if we want to handle those in alternate mime types.

felixarntz: @adamsilverstein Based on your feedback, I agree we may need to hook into `wp_update_attachment_metadata`. That would require some tweaks for our callback though, since that hook is fired every time a single new sub-size is generated in core. We had also discussed the potential usage of this hook in our call last week - the tricky part would be to figure out whether it is used for an image size being generated so that our code is only run in that case. Also, the logic for which sub-sizes to generate may need to be reworked.

Before we go towards doing that though, are we sure the current implementation is not working in a scalable way? I understand your argumentation @adamsilverstein, but why is that actually failing? Looking at the code, WP core uses `wp_update_image_subsizes()`, both via AJAX and the REST API. Before that function succeeds, it fires the `wp_generate_attachment_metadata` filter, which we are using here, so I don't follow how the code would return a success response without running our added filter callback 🤔 

In any case, I think it makes sense to drop the cron hook, replace it with a direct call, and explore further from there on whether the `wp_generate_attachment_metadata` filter is okay or whether we need an alternative solution using the `wp_generate_attachment_metadata` filter. cc @mitogh

felixarntz: Summarizing a chat with @mitogh and @adamsilverstein here:
* We're going to switch to "direct" generation of sub-sizes, essentially replace the cron hook with a direct call to the function as suggested above.
* The remaining problem with that approach is that `wp_update_image_subsizes()` will no longer trigger the `wp_generate_attachment_metadata` filter in any subsequent requests after the WP core sizes have been generated (since `wp_get_missing_image_subsizes()` returns an empty array, it short-circuits) - that's why the implementation will only make one more request after a failure in _our_ code, but that won't trigger the hook, so WP thinks the sub-sizes generation has succeeded.
* We can work around that by hooking into the `wp_get_missing_image_subsizes` filter, and essentially misusing it as an action, as follows:
    * If the list of sizes is empty _and_ we are in some sort of "upload" request, we manually call our function to generate the missing sub-sizes in the additional formats.
    * We just return the original filtered value without changing it.
    * In order to know whether we are in an "upload" request, we have to set a temporary meta key in the attachment as soon as our logic triggers, to flag that this attachment is basically "awaiting additional image MIME sub-sizes generation". Once our function has completed, we remove that same meta key flag again.
    * This approach will work regardless of whether image sizes are being generated via WP-AJAX or REST.
    * It is quite hacky, but it's what we need to do to work with the limits of not having a proper filter/action in WP core to use here. We should add a comment to that code indicating how that would behave differently in an eventual WP core patch.
* Let's limit this PR to going with just the simple removal of the cron hook. The above "hack" we can then add in a follow-up PR to the same issue #142 - that allows the review to be simpler, since already a lot of work has gone into this, and it's a good stage to break the work into two PRs.

modules/images/webp-uploads/load.php
mitogh: Sounds good this has been introduced as an `empty` check inside of the `foreach` loop.

adamsilverstein (approved):
Excellent work!

felixarntz (commented):
@mitogh All great now! Except can you please replace all the remaining usages of `array_key_exists` here with `isset` as outlined before?

mitogh:
Sure @felixarntz just updated the code based on your feedback.

felixarntz (approved):
Awesome work @mitogh! 🎉 

Let's proceed in #188 💪

christianmagill:
I see a lot of these issues are merged and closed and I'm not sure where the trail ends. Was this ever introduced into a release?

mxbclang:
Hi @christianmagill! The easiest way to tell if a PR/issue was released is via the Milestone field in the right sidebar – this was included in v1.0.0.

christianmagill:
Much thanks Bethany!

Christian Magill

PR #148, opened 2022-02-03
Added CT.CSS with Debug Constant

### LINKED ISSUE #118: Implement CT.CSS
Harry Roberts (CSS Wizardry) brought out a nifty tool that we recently included into our `wp-scaffold` at 10up, called `ct.css`
At 10up we included this CSS file behind a query param, such as: `?debug_perf`. 

The CSS file trawls the `<head>` showcasing:
- Which scripts / css files are render block
- Where async / defer attributes are redundant
- Other issues that could effect the parsing of HTML on the page.

Demo: https://csswizardry.com/ct/test/
Motivation: https://speakerdeck.com/csswizardry/get-your-head-straight
10up Scaffold Source: https://github.com/10up/wp-scaffold/blob/fc6ed94d596ecf0e4667249263a72b46d7d8fe5b/themes/10up-theme/includes/core.php#L263

dainemawer:
## Summary
As discussed in the "Implement CT.CSS" issue this is a `proof of concept` that implements CT.CSS into the plugin as a module. I do understand that this may not be part of the plugins intended use cases, but thought it was a worthwhile contribution nevertheless.

Fixes https://github.com/WordPress/performance/issues/118

## Relevant technical choices
- CT.CSS is a useful debugging tool that comes with minimal overhead that helps engineers debug render-blocking scripts in the `<head>`
- The feature can be toggled on/off through the plugins default module system.
- The CSS itself only runs when the `WP_DEBUG_PERFORMANCE` constant is set to true
- The CSS is inlined to the `<head>` saving a server request to any external stylesheet.

More information: https://csswizardry.com/ct/test/
Demo: https://speakerdeck.com/csswizardry/get-your-head-straight

## Checklist


modules/measurement/ct/load.php
```diff
@@ -0,0 +1,24 @@
...
+ * @since 1.0.0
+ */
+
+ require_once __DIR__ . '/helper.php';
+
+ function perflab_embed_ct_css() {
+
+	$perflab_debug_performance = perflab_should_debug_performance();
+
+	if ( ! $perflab_debug_performance ) return;
```
mitogh: I think a filter would be ideal instead and then move `perflab_should_debug_performance` to hook into that filter.

```suggestion
	if ( apply_filters( 'perflab_debug_performance_ct', false ) ) return;
```

Then the function above can hook into this filter.

```php
add_filter( 'perflab_debug_performance_ct', 'perflab_should_debug_performance' );

function perflab_should_debug_performance( $debug_performance ) {

	if( $debug_performance ) {
		return $debug_performance;
	}

	if ( function_exists( 'getenv' ) ) {
        return (bool) getenv( 'WP_DEBUG_PERFORMANCE' );
    }

	if ( defined( 'WP_DEBUG_PERFORMANCE' ) ) {
        return (bool) WP_DEBUG_PERFORMANCE;
    }

	return false;
}
```

modules/measurement/ct/load.php
```diff
@@ -0,0 +1,24 @@
...
...t:"The async attribute is redundant on inline scripts"}head script:not([src])[defer]::before{content:"The defer attribute is redundant on inline scripts"}head [rel=stylesheet][href^="//"],head [rel=stylesheet][href^=http],head script[src][src][src^="//"],head script[src][src][src^=http]{border-style:var(--ct-is-problematic);border-color:var(--ct-error)}head script[src][src][src^="//"]::before,head script[src][src][src^=http]::before{content:"[Third Party Blocking Script – " attr(src) "]"}head [rel=stylesheet][href^="//"]::before,head [rel=stylesheet][href^=http]::before{content:"[Third Party Blocking Stylesheet – " attr(href) "]"}head script~meta[http-equiv=content-security-policy]{border-style:var(--ct-is-problematic);border-color:var(--ct-error)}head script~meta[http-equiv=content-security-policy]::before{content:"[Meta CSP defined after JS]"}head>meta[charset]:not(:nth-child(-n+5)){border-style:var(--ct-is-problematic);border-color:var(--ct-warn)}head>meta[charset]:not(:nth-child(-n+5))::before{content:"[Charset should appear as early as possible]"}link[rel=stylesheet].ct,link[rel=stylesheet][media=print],script[async],script[defer],script[type=module],style.ct{display:none}' );
```
mitogh: Feels like this can go inside of a hook for `wp_head` removing the need for an empty stylesheet.

```php
add_action( 'wp_head', 'render_ct' );

function render_ct() {
?>
   <style>
       head{--ct-is-problematic:solid;--ct-is-affected:dashed;--ct-notify
       .....
   </style>
<?php
}
```

mitogh (commented):
Great work @dainemawer added some suggestions, questions .

modules/measurement/ct/load.php
dainemawer: I like this! I have updated, thanks @mitogh !

modules/measurement/ct/load.php
dainemawer: Yup, agreed! Updated!

westonruter:
Just seeing this PR now. Unfortunately this PR introduces a new module, but since the 2+ years this PR was opened, the modules have gone away in favor of standalone plugins in this monorepo. If we want to go ahead with this, we'll need to open a new PR that creates a plugin instead. Nevertheless, since the Performance Lab repo intends to incubate plugins for merging into WordPress core, I don't think CT.CSS will be a good fit for the Performance Lab plugin.

PR #151, opened 2022-02-04
[WIP] Generate additional mime types on upload

### LINKED ISSUE #142: When uploading, generate both WebP and jpeg format images by default

adamsilverstein:
## Summary
* Add a new filter `image_editor_output_formats` that returns the output formats and their order. The first type is used as the default output format
* The default `image_editor_output_format` filter applies for the first mime type, which also creates a full sized image if not the dame mime type as the uploaded image.
* New sub sizes **and** a full sized image are created for each mime type in the returned array

## Usage
By default, this code will generate WebP and jpeg sub sizes for every uploaded image, with WebP as the default.
* Adding add_filter( 'image_editor_output_formats', '__return_false' ); would ensure only the uploaded mime type would be use for sub sizes (current WordPress behavior).
* Adding add_filter( 'image_editor_output_formats', function() { return array( 'image/jpeg', 'image/webp' ); } ); would ensure jpeg and WebP were generated, with jpeg stored in sizes and used as the default output format
* Adding add_filter( 'image_editor_output_formats', function() { return array( 'image/webp' ); } ); would ensure only WebP sub sizes were generated (current behavior of the webp-uploads module).

Fixes #

## Checklist

### DIFF
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -32,7 +32,7 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
 	}
 
 	// Skip conversion when creating the `-scaled` image (for large image uploads).
-	if ( 1 === preg_match( '/.*-scaled\..{3}.?$/', $filename ) ) {
+	if ( preg_match( '/.*-scaled\..{3}.?$/', $filename ) ) {
 		return $output_format;
 	}
 
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -11,7 +11,8 @@
 /**
  * Filter the image editor default output format mapping.
  *
- * For uploaded JPEG images, map the default output format to WebP.
+ * Only applies when the image_editor_output_formats filter returns a single mime
+ * type. In that case, the mime type is used as the default output format.
  *
  * @since 1.0.0
  *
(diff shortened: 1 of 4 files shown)


felixarntz:
@adamsilverstein We need to be careful here to not introduce this logic entirely on the server-side, to avoid timeouts. This relates to the enhancements from https://core.trac.wordpress.org/ticket/40439 - the additional sizes in the other format have to somehow be kicked off from the client-side, so that the server can take care of them through individual requests.

felixarntz:
@adamsilverstein Do we still need this PR to be open? I'm thinking it's probably somewhat of a duplicate of #147 which is being worked on now.

adamsilverstein:
Closing, this was an experimental PR.

Issue #149, opened 2022-02-03
Update `the_content` with the correct image format


mitogh:
## Feature Description

Once an image attachment has more than one type of image available we need to make sure the correct images are rendered:

## Scenario 1 - Render WebP by default

- If the requested browser support WebP continue with the flow, otherwise render a JPEG version.
- If the attachment has a WebP version of the image replace any JPEG version with the exact WebP version of the image. 
- If no WebP version exists keep serving the JPEG version

## Scenario 2 - Render JPEG version

- If the browser does not accepts WebP version render a JPEG version
- if the current attachment does not have an exact WebP version for the requested image deliver the JPEG version
- If the WebP is disabled via a filter keep rendering JPEG by default.

This filter would be placed in `the_content` so when the article or articles are rendered the delivered content to the users uses the appropriate image format.

This work is a subsequent step for:

- #142

felixarntz:
@mitogh 
> If the requested browser support WebP continue with the flow

I thought the idea was to always load WebP in the frontend (if the image exists in that version), based on the previous discussions on the Slack chat and the decision made in https://github.com/WordPress/performance/issues/96#issuecomment-1027147051

We decided to keep generating JPEG versions not for the frontend, but for other usages that are more niche but relevant, e.g. email newsletters.

In other words, I don't think we need to worry about scenario 2 here - we can rely on WordPress core's definition of officially supported browsers, all of which support WebP.

mitogh:
Correct @felixarntz , the main idea would be to render WebP by default, if the browser supports it which is the majority of browsers at this point but for the cases where the browser does not support WebP the JPEG version would be displayed instead. By doing this we allow displaying the image at all times for older and more modern browsers, and this does not require a lot of extra work as is just checking against the `HTTP_ACCEPT` header which I think is a great benefit so modern and older browsers can still work as they are where modern browsers would take the WebP version by default (as WebP is supported there).

For example, if we load WebP by default this is how it would look in Safari when this validation is not present.

![2022-02-04_08-07](https://user-images.githubusercontent.com/3921289/152543845-593a23ec-9350-42fe-8baa-c4f5a49fc3ac.png)

And this is how it looks when the validation is present, the JPEG version is loaded instead.

![2022-02-03_19-28](https://user-images.githubusercontent.com/3921289/152544434-bca175ce-da3b-49a3-9b49-9ccd92beb809.png)

And for modern browsers, a WebP version is loaded instead.

[...]

PR #152, opened 2022-02-04
Update `the_content` with the appropiate image format

mitogh:
## Summary

Uses `image/webp` if supported by the client and updates the `image/jpeg`, `image/jpg` formats with the modern format if available.

References in the content are updated with the WebP format if available.

Fixes: 

- #149

Related with:

- #147

## Checklist


adamsilverstein:
This worked well in my testing across a variety of themes. The only issue I noticed was when inserting an image at the "large" size with a large image upload that results in a `-scaled` image, the jpeg is used for the image `src`, since the `webp` version isn't available yet (see https://github.com/WordPress/performance/issues/174).

image

mitogh:
Thanks for taking the time to test it, currently this is expected and described here:

https://github.com/WordPress/performance/blob/7b38de8bd880e09349bc8a441f09ef1be6bc30e3/modules/images/webp-uploads/load.php#L414-L417

Based on the decision from:

- #174 

The code would be updated accordingly if required.

adamsilverstein:
> Based on the decision from #174 ...

Right, this answers that question for me - we need the full size `webp` available for the correct markup when users select the full size image.

adamsilverstein:
> Right, this answers that question for me - we need the full size webp available for the correct markup when users select the full size image.


Note this applies even for smaller upload sizes that don't generate a `-scaled` image, the jpg is still used when selecting the "Full" size in the editor:

image

modules/images/webp-uploads/load.php
```diff
@@ -340,3 +340,112 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
 }
 
 add_filter( 'wp_get_missing_image_subsizes', 'webp_uploads_wp_get_missing_image_subsizes', 10, 3 );
+
+/**
+ * Filters on `the_content` to update the references for supported mime of images into the
+ * `webp_uploads_preferred_mime_type()` for the most part `image/webp` if the current
+ * attachment contains the targeted mime type.
```
felixarntz: The reference to `webp_uploads_preferred_mime_type()` is old, so let's remove it. We can figure out how to make all of this customizable in #187. Also, we should point out that a lot of the code here is duplicated from WP core's `wp_filter_content_tags()`, and later in WP core we wouldn't need to do that.

```suggestion
 * Filters `the_content` to update images so that they use the preferred MIME type where possible.
 *
 * By default, this is `image/webp`, if the current attachment contains the targeted MIME
 * type. In the near future this will be filterable.
 *
 * Note that most of this function will not be needed for an eventual core implementation as it
 * would rely on `wp_filter_content_tags()`.
```

modules/images/webp-uploads/load.php
```diff
@@ -340,3 +340,112 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
...
+ *
+ * @since n.e.x.t
+ *
+ * @see the_content
```
felixarntz: IMO not worth focusing on that filter, but what would be helpful is a pointer to the very related `wp_filter_content_tags()`.
```suggestion
 * @see wp_filter_content_tags()
```

mitogh:
Right in this particular case, this image regardless of `scaled` prefix is the `full` image size stored in the `file`  meta key, so it behaves as a virtual size defined during runtime. 

If the decision from #174 would be to add support for multiple mime types for this virtual type I think we need to support a `sources` property to the root of the metadata with all the available mime types for each image.

modules/images/webp-uploads/load.php
```diff
@@ -340,3 +340,112 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
...
+
+		foreach ( $urls as $url ) {
+			$images[ $attachment_id ][ $url ] = true;
+		}
+	}
+
+	// TODO: Add a filterable option to change the selected mime type. See https://github.com/WordPress/performance/issues/187.
+	$target_mime = 'image/webp';
+	$replacement = array();
+	foreach ( $images as $attachment_id => $urls ) {
```
felixarntz: Instead of having all this in a single function and using a variable storage like `$images[ $attachment_id ][ $url ]`, I think it would be great to implement this in a way that aligns more with [WP core's `wp_filter_content_tags()`](https://developer.wordpress.org/reference/functions/wp_filter_content_tags/) which we should use eventually in a core patch, since it already does all the content parsing of images. We can't use that here since it's lacking filters, but later in core we definitely should.

To prepare the implementation here for a smoother transition to core, I suggest the following:
* Split it into two functions. One that does the content image parsing (which we won't need later in core) and the other to modify an individual image, similar to e.g. `wp_img_tag_add_srcset_and_sizes_attr()`.
* The variable should become a simple map like `$image => $attachment_id` (image HTML tag and attachment ID), the exact same how WP core's `wp_filter_content_tags()` has it. This way we're already building a compatible implementation which we can later just plug into core.
* For example, the second function could be something like `webp_uploads_img_tag_choose_mime( $image, $context, $attachment_id )`, and it would return the modified `$image` tag. All the logic you have below here should then be moved into that function. The above `preg_match_all` to find all URLs would also need to be included there, instead of having it above, which isn't really necessary.

Basically the code would become something like
```php
foreach ( $images as $image => $attachment_id ) {
	$filtered_image = webp_uploads_img_tag_choose_mime( $image, 'the_content', $attachment_id );
	if ( $filtered_image !== $image ) {
		$content = str_replace( $image, $filtered_image, $content );
	}
}
```

modules/images/webp-uploads/load.php
```diff
@@ -340,3 +340,112 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
...
+			$src_filename = wp_basename( $url );
+			$extension    = wp_check_filetype( $src_filename );
+
+			// Extension was not set properly no action possible or extension is already in the expected mime.
+			if ( empty( $extension['type'] ) || $extension['type'] === $target_mime ) {
+				continue;
+			}
+
+			// Find the appropriate size for the provided URL.
+			foreach ( $metadata['sizes'] as $name => $size_data ) {
```
felixarntz: Instead of nesting two `foreach` loops here (_O(n²)_), we can optimize performance a bit by having them as two separate `foreach` loops (_O(2n)_):
* First, loop through all `$urls` only to find each `$src_filename` (only if the extension isn't already the target MIME, like you have it above), and through that start populating the replacements map like `$replacement[ $src_filename ] = ''`. This way you just set an empty string, the following loop can then populate the actual replacements.
* The second loop (after the first loop) can then **just once** go through `$metadata['sizes']` and check for each size whether a matching `$replacement[ $src_filename ]` exists. If not, `continue`, if yes, set the replacement value like you have below.

At the end we have to call `array_filter( $replacement )`, to ignore any keys where no replacements could be found for.

This way you get the same `$replacement` data, but without increased complexity from nested `foreach` loops.

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -385,4 +385,163 @@ public function it_should_remove_the_webp_version_of_the_image_if_the_image_is_f
...
 	}
+
+	/**
+	 * Avoid the change of URLs of images that are not part of the media library
+	 *
+	 * @group webp_uploads_update_image_references
+	 *
+	 * @test
+	 */
+	public function it_should_avoid_the_change_of_ur_ls_of_images_that_are_not_part_of_the_media_library() {
```
felixarntz:
```suggestion
	public function it_should_avoid_the_change_of_urls_of_images_that_are_not_part_of_the_media_library() {
```

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -385,4 +385,163 @@ public function it_should_remove_the_webp_version_of_the_image_if_the_image_is_f
...
+		$attachment_id = $this->factory->attachment->create_upload_object( $image_path );
+
+		$tag          = wp_get_attachment_image( $attachment_id, 'medium', false, array( 'class' => "wp-image-{$attachment_id}" ) );
+		$expected_tag = $tag;
+		$metadata     = wp_get_attachment_metadata( $attachment_id );
+		foreach ( $metadata['sizes'] as $size => $properties ) {
+			$this->assertArrayHasKey( 'image/jpeg', $properties['sources'] );
+			$this->assertArrayHasKey( 'file', $properties['sources']['image/jpeg'] );
+			$this->assertArrayHasKey( 'image/webp', $properties['sources'] );
+			$this->assertArrayHasKey( 'file', $properties['sources']['image/webp'] );
```
felixarntz: IMO this doesn't need to be asserted here, since this is a test for `webp_uploads_update_image_references()`, not for the other function that updates the attachment metadata.

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -385,4 +385,163 @@ public function it_should_remove_the_webp_version_of_the_image_if_the_image_is_f
...
+			$expected_tag = str_replace( $properties['sources']['image/jpeg']['file'], $properties['sources']['image/webp']['file'], $expected_tag );
+		}
+
+		$this->assertNotEmpty( $expected_tag );
+		$this->assertNotSame( $tag, $expected_tag );
```
felixarntz, line 466: These seem unnecessary as well.

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -385,4 +385,163 @@ public function it_should_remove_the_webp_version_of_the_image_if_the_image_is_f
...
+	public function it_should_prevent_replacing_an_image_with_no_available_sources() {
+		add_filter( 'webp_uploads_supported_image_mime_transforms', '__return_empty_array' );
+
+		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/car.jpeg' );
+
+		$tag      = wp_get_attachment_image( $attachment_id, 'full', false, array( 'class' => "wp-image-{$attachment_id}" ) );
+		$metadata = wp_get_attachment_metadata( $attachment_id );
+		foreach ( $metadata['sizes'] as $size => $properties ) {
+			$this->assertArrayNotHasKey( 'sources', $properties );
+		}
```
felixarntz: See above, this shouldn't be relevant to assert here.

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -385,4 +385,163 @@ public function it_should_remove_the_webp_version_of_the_image_if_the_image_is_f
...
+
+		$this->assertSame( $tag, webp_uploads_update_image_references( $tag ) );
+	}
+
+	public function data_provider_not_supported_webp_images() {
+		yield 'PNG image' => array( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/dice.png' );
+		yield 'GIFT image' => array( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/earth.gif' );
+	}
```
felixarntz: I think this test is unnecessary. It tests exactly the same as the above `it_should_prevent_replacing_an_image_with_no_available_sources()`, which is that an image without `sources` is not being replaced. Whether there are no `sources` because of the filter returning an empty value or because the image is a PNG/GIF is irrelevant to `webp_uploads_update_image_references()` - if anything, that should be tested for `webp_uploads_generate_image_size()` or `webp_uploads_get_supported_image_mime_transforms()` which are actually responsible for that.

felixarntz (commented):
@mitogh Functionally, this looks great! I left some comments, most importantly I think we should restructure the code a bit to align it more with WP core's `wp_filter_content_tags()`, so that it is prepared to be more easily transferable into a core patch later.

tests/modules/images/webp-uploads/webp-uploads-test.php
mitogh: Good point, the main reason for PNG and GIFt was to make sure those URL extensions were not affected by the regex inside of the function that does the replacement. Which is slightly different from `it_should_prevent_replacing_an_image_with_no_available_sources` which mimics an image that was uploaded before the plugin and module was activated.

tests/modules/images/webp-uploads/webp-uploads-test.php
mitogh: This mostly ensures that the expected tag was modified out of the original tag and make sure the following assertion is not a false positive.

modules/images/webp-uploads/load.php
mitogh: > Split it into two functions. One that does the content image parsing (which we won't need later in core) and the other to modify an individual image, similar to e.g. wp_img_tag_add_srcset_and_sizes_attr().

Agree on this one.

> The variable should become a simple map like $image => $attachment_id (image HTML tag and attachment ID), the exact same how WP core's wp_filter_content_tags() has it. This way we're already building a compatible implementation which we can later just plug into core.

Actually on this one having a hash map makes more sense and is more optimal due if an image is included 10 times on the content the logic to replace the same image would take place 10 times instead of a single time if we are using a hash map as we currently have this logic will ensure every single iteration is unique based on the image ID, Using the core version currently does not consider this scenario. 

> For example, the second function could be something like webp_uploads_img_tag_choose_mime( $image, $context, $attachment_id ), and it would return the modified $image tag. All the logic you have below here should then be moved into that function. The above preg_match_all to find all URLs would also need to be included there, instead of having it above, which isn't really necessary.

Agree on that one.

modules/images/webp-uploads/load.php
mitogh: Actually, this approach would take `O(3N*M)` the current approach is `N*M` core approach is very close the main difference is the fact that core uses functions to hide the complexity see `wp_image_src_get_dimensions` is called for every image which internally triggers a loop for every single size. 

The fact that we iterate on every single size 2 more times 1 to loop metadata and another during the `array_filter` makes the current approach complete in fewer steps. We can do this because we are using a hash map where we know each iteration is unique and we are do not repeat the same step for the same attachment id.

I would split the code into 2 functions as you described above to see if that would make things a bit easier to visualize. But let me know otherwise.

modules/images/webp-uploads/load.php
felixarntz: > Actually on this one having a hash map makes more sense and is more optimal due if an image is included 10 times on the content the logic to replace the same image would take place 10 times instead of a single time if we are using a hash map as we currently have this logic will ensure every single iteration is unique based on the image ID, Using the core version currently does not consider this scenario.

Can you clarify how you mean that? If an image is included 10 times in the same content, with my suggestion above (`$image => $attachment_id` map) the logic would still only be run once, unless the same image was included in 10 different versions (e.g. if someone included it as `medium`, `large`, `full`, `post-thumbnail` etc., all in a single post), which is an edge-case.

Good point on the URL lookup though, it makes sense to find a solution where we don't need to look up the replacement for the same URL multiple times. But still, let's decouple that storage from the `$image => $attachment_id` map, which is needed for seamless integration in core. We can still store the URLs separately, in a way that the lookup doesn't need to happen more than once per URL.

modules/images/webp-uploads/load.php
felixarntz: > Actually, this approach would take `O(3N*M)`

What is N and M in this, could you clarify? I was specifically referring to the two `foreach` loops in your code here, not in comparison to how it's today in WP core or anything. Good point on considering the `array_filter`, but still from my understanding that would still mean less complexity.

Let's keep this open and revisit after you've split into two functions, that we should definitely do to align the high-level implementation with core so that it can be integrated later.

felixarntz:
@adamsilverstein 
> This worked well in my testing across a variety of themes. The only issue I noticed was when inserting an image at the "large" size with a large image upload that results in a `-scaled` image, the jpeg is used for the image `src`, since the `webp` version isn't available yet (see #174).
> 
> image

I noticed that as well, it depends on the viewport size and pixel density. Since we currently don't have WebP versions for the "full" size (whether `-scaled` or original), it's best to currently test this PR on a mobile viewport since there it's most likely to get served a WebP image.

The problem is that most modern screens have a high pixel density like `@2x`, which means if I look at an image of 600px (or more accurately in a viewport of 600px), the browser will look for a size >=1200px, which would already be more than WordPress core's `large` by default.

If you want to test this more specifically without the above caveats, you can put a temporary filter in place to ensure WordPress never chooses a size greater than the image size you actually specified (which may cause blurry images on larger or high-density devices though):
```php
add_filter(
	'wp_calculate_image_srcset',
	function( $sources, $size_array ) {
		foreach ( $sources as $index => $source ) {
			if ( 'w' === $source['descriptor'] && $source['value'] > $size_array[0] ) {
				unset( $sources[ $index ] );
			}
		}
		return $sources;
	},
	10,
	2
);
```

There's certainly room for optimization in core's default behavior there, but that's a separate rabbit hole :)

> Right, this answers that question for me - we need the full size `webp` available for the correct markup when users select the full size image.

I agree this gives us an answer for #174, that we should also generate a WebP version of the `full` size.

modules/images/webp-uploads/load.php
mitogh: > What is N and M in this, could you clarify? I was specifically referring to the two foreach loops in your code here, not in comparison to how it's today in WP core or anything. Good point on considering the array_filter, but still from my understanding that would still mean less complexity.

M - Being the number of images processed
M - The number of sizes that we need to look for each image in might be the case that both are the same size but technically those two for each are different in terms of the number of sizes.

> Let's keep this open and revisit after you've split into two functions, that we should definitely do to align the high-level implementation with core so that it can be integrated later.

Sure just updated the code based on your feedback let me know in case you have additional concers.

modules/images/webp-uploads/load.php
mitogh: > Can you clarify how you mean that? If an image is included 10 times in the same content, with my suggestion above ($image => $attachment_id map) the logic would still only be run once, unless the same image was included in 10 different versions (e.g. if someone included it as medium, large, full, post-thumbnail etc., all in a single post), which is an edge-case.

The fact that we need to look up for URLS besides just `` tags required a way to uniquely identity each URL to avoid duplicate work, I've updated the code based on your feedback as I think it makes more sense as you describe now that the logic is placed in 2 places.

I've omited the `context` and `attachment_id` on the second function as those were never used I feel like we can add those once the migration in core lands due it would make sense in that context but not so much under our current filter, let me know otherwise.

modules/images/webp-uploads/load.php
```diff
@@ -340,3 +340,130 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
...
+		// Create an array if attachment_id has not been added to the images array.
+		if ( empty( $images[ $attachment_id ] ) ) {
+			$images[ $attachment_id ] = array();
+		}
+
+		$images[ $attachment_id ][] = $img;
+	}
+
+	foreach ( $images as $attachment_id => $images_tag ) {
+		$metadata = wp_get_attachment_metadata( $attachment_id );
```
felixarntz: Something we need to include before here is a call to `_prime_post_caches()`, similar to how it's present in [`wp_filter_content_tags()`](https://developer.wordpress.org/reference/functions/wp_filter_content_tags/). That function ensures that all attachment metadata for these images gets requested at once. Without that, this code will get a performance problem, since on an average environment without object caching every `wp_get_attachment_metadata()` call would cause an individual database query.
```suggestion
	if ( count( $images ) > 1 ) {
		/*
		 * Warm the object cache with post and meta information for all found
		 * images to avoid making individual database calls.
		 */
		_prime_post_caches( array_keys( $images ), false, true );
	}
	foreach ( $images as $attachment_id => $images_tag ) {
		$metadata = wp_get_attachment_metadata( $attachment_id );
```

modules/images/webp-uploads/load.php
```diff
@@ -340,3 +340,130 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
...
+ * Finds all the urls with *.jpg and *.jpeg extension and updates with *.webp version for the provided image
+ * for the specified image sizes, the *.webp references are stored inside of each size.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $image An <img> tag where the urls would be updated.
+ * @param array  $metadata An associative array with the metadata for the image.
+ * @return string The updated img tag.
+ */
+function webp_uploads_img_tag_update_mime_type( $image, array $metadata ) {
```
felixarntz: While I see your point that it's technically not necessary, I think it makes sense to align this function already to be how it can easily be integrated into WordPress core, i.e. we could use the same parameters `$image, $context, $attachment_id`.

Currently you're calling `wp_get_attachment_metadata()` outside of this function, but that doesn't really have any performance benefit. Even if an attachment is present 10 times, these additional calls will all use the same data that is being cached, even on a system without object cache it will only be a single DB request.

Especially with the `_prime_post_caches()` call, there should be no concern of moving the `wp_get_attachment_metadata()` call into this function, which then aligns it with other core functions (which also call `wp_get_attachment_metadata()` for every image, since that's not really a problem).

modules/images/webp-uploads/load.php
```diff
@@ -340,3 +340,130 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
...
+		if ( ! $attachment_id ) {
+			continue;
+		}
+
+		// Create an array if attachment_id has not been added to the images array.
+		if ( empty( $images[ $attachment_id ] ) ) {
+			$images[ $attachment_id ] = array();
+		}
+
+		$images[ $attachment_id ][] = $img;
```
felixarntz: I'm still not sure about the benefits of this structure. Why not `$images[ $img ] = $attachment_id;`?

Related to my other comments, the call to `wp_get_attachment_metadata()` doesn't really have a notable performance cost due to caching. Also, the way your code is structure here, it will take more cycles:
* With your current `images[ $attachment_id ][] = $img;`, if the same `img` tag is included 10 times in a post, `webp_uploads_img_tag_update_mime_type()` will be called 10 times, and the same `str_replace` will happen 10 times, which will not be necessary, since the first time will already replace all of these same occurrences.
* With `$images[ $img ] = $attachment_id;`, if the same `img` tag is included 10 times in a post, `webp_uploads_img_tag_update_mime_type()` will still only be called once and `str_replace` will also only run once.
* Both approaches will bring the same end result, but the latter does it more performantly.

modules/images/webp-uploads/load.php
```diff
@@ -340,3 +340,130 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
...
+ * @param array  $metadata An associative array with the metadata for the image.
+ * @return string The updated img tag.
+ */
+function webp_uploads_img_tag_update_mime_type( $image, array $metadata ) {
+	if ( empty( $metadata['file'] ) ) {
+		return $image;
+	}
+
+	// Creates a regular extension to find all the URLS with the provided extension for img tag.
+	preg_match_all( '/[^\s"]+\.jpe?g/i', $image, $matches );
```
felixarntz: While this works, let's put these extensions in a variable like you had it before, that makes it easier to follow and later also to customize. You can then still put them into this regex dynamically, something like `(jpg|jpeg)`.

We should then have a `TODO` comment like below (referring to #187) for that variable as well.

modules/images/webp-uploads/load.php
```diff
@@ -340,3 +340,130 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
...
+	if ( empty( $matches ) ) {
+		return $image;
+	}
+
+	$urls = reset( $matches );
```
felixarntz: Nit-pick, but IMO it's safer to not use an iterator and instead have
```suggestion
	$urls = $matches[0];
```

modules/images/webp-uploads/load.php
```diff
@@ -340,3 +340,130 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
...
+
+		$src_filename = wp_basename( $url );
+		$extension    = wp_check_filetype( $src_filename );
+		// Extension was not set properly no action possible or extension is already in the expected mime.
+		if ( empty( $extension['type'] ) || $extension['type'] === $target_mime ) {
+			continue;
+		}
+
+		// Find the appropriate size for the provided URL.
+		foreach ( $metadata['sizes'] as $name => $size_data ) {
```
felixarntz, line 455: Ahh now I see where my previous thinking was flawed. This is all good then :)

modules/images/webp-uploads/load.php
```diff
@@ -340,3 +340,130 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
...
+			if ( empty( $size_data['sources'][ $target_mime ]['file'] ) ) {
+				continue;
+			}
+
+			// This is the same as the file we want to replace nothing to do here.
+			if ( $size_data['sources'][ $target_mime ]['file'] === $src_filename ) {
+				continue;
+			}
+
+			$image = str_replace( $src_filename, $size_data['sources'][ $target_mime ]['file'], $image );
```
felixarntz, line 470: Not sure what's more performant. Now you have many individual `str_replace` calls while before you had a single `str_replace` call for all values in arrays. Was there a reason why you changed to this approach? I'd be curious to see which one is faster.

Anyway, that's kind of a micro-optimization, so not super-critical for now.

modules/images/webp-uploads/load.php
```diff
@@ -340,3 +340,130 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
...
+		if ( ! preg_match( '/wp-image-([\d]+)/i', $img, $class_name ) ) {
+			continue;
+		}
+
+		if ( empty( $class_name ) ) {
+			continue;
+		}
+
+		// Make sure we use the last item on the list of matches.
+		$attachment_id = (int) end( $class_name );
```
felixarntz: Nit-pick, but IMO safer to not use an iterator:
```suggestion
		$attachment_id = (int) $class_name[1];
```

felixarntz (commented):
@mitogh Great! I left a few follow-up comments, mostly minor though. This is almost ready.

modules/images/webp-uploads/load.php
mitogh: i see the point, the previous iteration used more space (because it was creating an array to store) it had to replace every single occurrence in the array tho in the same we are doing individually inside of the loop, agree we can improve as we move forward but seems like results might be similar.

modules/images/webp-uploads/load.php
mitogh: Good call.

modules/images/webp-uploads/load.php
mitogh: By using this structure we avoid the need to do what core does here:

```php
	// Reduce the array to unique attachment IDs.
	$attachment_ids = array_unique( array_filter( array_values( $images ) ) );
```

This structure would ensure only unique attachment IDs are stored, and avoid the need to use other functions.

One thing I'm having second thoughts is the fact that we need an array to store all the `img` tags as you describe if we have one should work for the majority of the cases as that would replace all instances on the content. However, I'm thinking in the use case where the same attachment ID uses 2 different markup elements like when you embed a particular size in the content and then embed another size the current implementation would be able to support that use case as would traverse all `img` tags for that particular attachment ID however the suggested approach won't be able to handle the following scenario:

> The content contains 2 images, at the top one image uses a small image size, at the bottom of the content the same image the same image uses only bigger sizes where the smaller size was not referenced.

In the scenario described above only the last element tag would be updated leaving the first image tag without modifications, unless the image uses any of the images from the same image at the bottom of the content.

felixarntz: @mitogh Either of the implementations works. Something that is missing from your current implementation though is that every array of `$images[ $attachment_id ]` may include duplicate tags. In that case, it will run the logic below for the same tag multiple times which wouldn't do anything, it would just run logic without even making more replacements, because the first run per tag would already have made the replacement. So there would need to be some sort of `array_unique()` call for every `$images[ $attachment_id ]` to avoid it.

Using WP core's existing approach avoids that `array_unique()` call since `$images[ $image ]` automatically ensures there are no duplicates. But again, in that case we need `$attachment_ids = array_unique( array_filter( array_values( $images ) ) );` like you're saying.

IMO no approach here is really more suitable than the other. Therefore I generally would be okay going with your suggested approach, however since we're building towards WP core, I'd say let's use the same approach that WP core follows, to avoid having a discussion about that later, which would distract from what we want to achieve here.

felixarntz (commented):
This one open discussion is the only thing we need to finalize before we can merge this.

mitogh:
Thanks, @felixarntz just updated to code to follow what WP Core uses in order to reduce the friction in the scenario this would be merged into WP Core.

Let me know in case you have any additional feedback.

modules/images/webp-uploads/load.php
```diff
@@ -340,3 +340,138 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
...
+
+/**
+ * Finds all the urls with *.jpg and *.jpeg extension and updates with *.webp version for the provided image
+ * for the specified image sizes, the *.webp references are stored inside of each size.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $image An <img> tag where the urls would be updated.
+ * @param string $context The context where this is function is being used.
+ * @param int    $attachment_id The ID of the attachment being modified.
```
```suggestion
 * @param string $image         An <img> tag where the urls would be updated.
 * @param string $context       The context where this is function is being used.
 * @param int    $attachment_id The ID of the attachment being modified.
```

felixarntz (approved):
Awesome work @mitogh!

adamsilverstein (approved):
Code looks great to me, thanks for the structural changes that will make prepping this for core more straightforward.  Also retested and everything worked well in my testing.

Issue #150, opened 2022-02-03
Progressive transition into WebP formats


mitogh:
## Feature Description

Based on:

- #142 
- #149 

At this point, we can assume that every image would have at least 2 formats when the WebP module is enabled in this case the well-known and supported JPEG and the WebP format, (eventually leading into other formats as those are supported in WP). So this issue is mostly to define the strategy behind this image generation and decide into the approach into the transition into a modern format and still have a fallback for older browsers or users using this plugin but not necessarily interested in test/using this section of the performance plugin.

### Strategy 1 - Default JPEG and WebP as an enhancement 

This approach suggests keeping using JPEG in the sizes definition of the metadata of each image, creating the WebP version after all the JPEG versions were created and WebP images are created only as an enchantment for the existing images. 

#### Benefits

- WebP would be used as an enchantment for JPEG replacing existing JPEG images with WebP versions 
- Backwards compatible this approach allows for a transition into WebP formats incrementally, meaning older attachments where WebP version was not created would keep working as they are until a WebP version is created for that attachment see #24  and then WebP version would be used instead.

#### Problems

- Requires content modification like #149 in order to render the new image format
- New image format is not used by default and requires filtering in core functions in order to ensure a more modern image format is used appropriately.

### Strategy 2 - WebP by default and JPEG as a fallback

This strategy suggests instead a logic similar to the current behavior of the plugin which is for new image uploads only create the WebP version and create the JPEG version of the images as fallback, and WebP format is used as the default image format for all new image uploads. JPEG versions for the images are created as a fallback approach.

### Benefits

- Core functions like `wp_get_attachment_image_src` and `the_conent` would render automatically the modern image format, in this case WebP
- Benefits from modern images out of the box due this image is already rendered by default for more core functions.

#### Problems

- If a user of the plugin opts out of not using WebP version a filter should be used in order to adjust the content to render the JPEG version for each image instead. 
- When moving or deactivating the plugin WebP versions would remain as the main image format as the logic to render JPEG won't be available at this point. 
- Requires an additional logic in place to migrate JPEG versions as the main images for each image size instead of WebP

## Reasoning

The main reason to discuss this approach is to have a standard way of doing things, due if we have WebP versions for some attachments as the main images and JPEG for others things start to become more complex due we need to deal with multiple combinations, in some cases JPEG version is used an in others WebP. So knowing before approaching #149 or other places where we need to replace or update references to images we know which mime type is the main image type and which one is the one used as an enchantment.

[...]

PR #154, opened 2022-02-04
Store the backup JPEG images in `_wp_attachment_backup_sizes`

mitogh:
## Summary

Creates the JPEG version of the WebP images and store the references into `_wp_attachment_backup_sizes` similar to image transformations in order to take advantage of benefits from WP Core like files removed automatically by WordPress and uploaded into CDNs

Fixes:

- #142 
- #150 

## Relevant technical choices

The changes introduced in this PR replicate the same behavior from WordPress core to store edited versions of additional mime types and the additional mime types themselves in the meta `_wp_attachment_backup_sizes`.

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,170 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+			if ( is_wp_error( $image ) ) {
+				continue;
+			}
+
+			// TODO: Store the file size of the created image.
+			// $image['filesize'] = filesize( $image['path'] );
+			// Remove the path of the image to follow the same pattern as core.
+			unset( $image['path'] );
+
+			$backup_sizes[ $size . '_' . $extension ] = $image;
```
```suggestion
			$backup_sizes[ $size . '-' . $extension ] = $image;
```
we use a dash everywhere else for filename modification

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,170 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+			$editor->resize( (int) $properties['width'], (int) $properties['height'], $properties['crop'] );
+			$filename = $editor->generate_filename( null, null, $extension );
+			$image    = $editor->save( $filename, $mime );
+
+			if ( is_wp_error( $image ) ) {
+				continue;
+			}
+
+			// TODO: Store the file size of the created image.
+			// $image['filesize'] = filesize( $image['path'] );
```
adamsilverstein, line 198: I'm in favor of storing this for all sizes, this is useful for performance analysis and potentially other things.

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,170 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+ * Return an array with the list of valid mime types for a sources' property.
+ *
+ * @since n.e.x.t
+ *
+ * @todo Add a filter to support more mime types.
+ *
+ * @return string[] An array of valid mime types, where the key is the mime type and the value is the extension type.
+ */
+function webp_uploads_valid_image_mime_types() {
+	return array(
```
adamsilverstein: maybe should be filtered here?

modules/images/webp-uploads/load.php
```diff
@@ -35,5 +35,170 @@ function webp_uploads_filter_image_editor_output_format( $output_format, $filena
...
+	if ( ! file_exists( $file ) ) {
+		return $metadata;
+	}
+
+	// Prevent to convert JPEG to WebP if we are creating JPEG versions of the image.
+	remove_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format' );
+	$backup_sizes = get_post_meta( $attachment_id, '_wp_attachment_backup_sizes', true );
+	$backup_sizes = is_array( $backup_sizes ) ? $backup_sizes : array();
+
+	foreach ( webp_uploads_get_image_sizes() as $size => $properties ) {
```
adamsilverstein: instead of trying to recreate sizes, can we just copy the meta 'sizes'?

modules/images/webp-uploads/load.php
mitogh: Thanks, great flag.

modules/images/webp-uploads/load.php
mitogh: Great. for now would keep the TODO, and we can came back to this once we start working on: 

- https://github.com/WordPress/performance/issues/8

modules/images/webp-uploads/load.php
mitogh: Agree, added a todo at the top of the function, but let me add the filter right away instead.

modules/images/webp-uploads/load.php
mitogh: We can certainly do that but aside from that, we need to know if the newly created image should be `cropped` or not and that's the main reason why we need this function. 

I think something we can do at the top of the `foreach` loop to make sure this `$size` exists in the original metadata before creating a targeted image instead.

mitogh: Actually, this is already the case on this line :smiley:  

https://github.com/WordPress/performance/blob/84eb3ec30e532939cf81c65d4fc46120ae1aac8c/modules/images/webp-uploads/load.php#L81

modules/images/webp-uploads/load.php
```diff
@@ -4,36 +4,434 @@
...
+/**
+ * Callback used to return a callback used by `webp_uploads_ajax_wp_die_handler` filter.
+ *
+ * @see webp_uploads_ajax_wp_die_handler
+ *
+ * @return string A function that would be used as a callback.
+ */
+function webp_uploads_wp_die_ajax_handler() {
+	return 'webp_uploads_ajax_wp_die_handler';
+}
```
eugene-manuilov: This hook should execute the callback function passed to it via filter instead of calling the `_ajax_wp_die_handler` function:

```php
function webp_uploads_wp_die_ajax_handler( $callback ) {
	return function( $message, $title = '', $args = array() ) use ( $callback ) {
		if ( ... ) {
			webp_uploads_restore_image_from_backup( absint( $_REQUEST['postid'] ) );
		}

		call_user_func( $callback, $message, $title, $args );
	};
}
```

Also when you register this hook it should use the biggest priority number to make sure it is executed at the very end:

```php
add_filter( 'wp_die_ajax_handler', 'webp_uploads_wp_die_ajax_handler', PHP_INT_MAX );
```

mitogh: > This hook should execute the callback function passed to it via filter instead of calling the _ajax_wp_die_handler function:

Right, great point let me update the code accordingly :) 

> Also when you register this hook it should use the biggest priority number to make sure it is executed at the very end:

Well in this case we know even if is executed first on the set of hooks won't make any difference so the default seems reasonable. 

Ideally we should have a filter in the ajax via core for restore and edited images.

modules/images/webp-uploads/load.php
adamsilverstein: One thing I noticed in testing when I uploading a very large image, I got a jpeg created with a `-scaled` suffix, then **all** of the generated `.webp` sub size files also included the `-scaled` suffix:

image

In addition, I noticed there was no full sized `.webp` created (that is `{filename}-scaled.webp`) - we need that full sized file for the image markup (only applies when you upload images larger than `big_image_size_threshold`)

mitogh: Thanks for testing and for the feedback. Some follow-up comments.

> One thing I noticed in testing when I uploading a very large image, I got a jpeg created with a -scaled suffix, then all of the generated .webp sub size files also included the -scaled suffix.

Great point mostly because this is created out of the scaled version. But let me fix this one :) 

> In addition, I noticed there was no full sized .webp created (that is {filename}-scaled.webp) - we need that full sized file for the image markup (only applies when you upload images larger than big_image_size_threshold)

Actually, this follows the same logic applied here #122  and #143   this behavior would go against this solved issue, as this `scaled` version is the `full` size image. Based on the issue `scaled` and `full` size always remains in the original image format. (not converted)

Let me know if you think we should change it instead.

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,414 @@
...
+	}
+
+	// If no size is present there's no need to generate any additional image as a backup.
+	if ( empty( $metadata['sizes'] ) ) {
+		// TODO: Maybe delete the meta key and files depending on `IMAGE_EDIT_OVERWRITE` ? if no size is present.
+		return;
+	}
+
+	// All subsizes are created out of the `file` property and not the original image.
+	$file = get_attached_file( $attachment_id, true );
```
adamsilverstein: Why do you use get_attached_file here instead of the original image, is this to use the -scaled image?

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,414 @@
...
+		// Generate backups only for the missing mime types.
+		$formats = webp_uploads_get_remaining_image_mimes( $metadata, $size );
+
+		foreach ( $formats as $mime => $extension ) {
+			$key = $extension . '-' . $size;
+
+			// The file already exists as part of the backup sizes in the same key.
+			if ( array_key_exists( $key, $backup_sizes ) ) {
+				// This case would remove the file if is an edited version.
+				if ( defined( 'IMAGE_EDIT_OVERWRITE' ) && IMAGE_EDIT_OVERWRITE ) {
```
adamsilverstein, line 166: Nice!

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,414 @@
...
+					$file_name     = empty( $backup_sizes[ $key ]['file'] ) ? '' : $backup_sizes[ $key ]['file'];
+					$file_location = path_join( $dirname, $file_name );
+					// Test to make sure this is an edited version, if so we can remove it safely.
+					preg_match( '/-e[\d]{13}-/', $file_name, $matches );
```
adamsilverstein: consider storing this regex in a constant since you reuse it

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,414 @@
...
+						wp_delete_file( $file_location );
+					}
+				} else {
+					$backup_sizes[ "{$key}-{$hash}" ] = $backup_sizes[ $key ];
+				}
+				// Clear the key, so the subsequent section can add the new image there.
+				unset( $backup_sizes[ $key ] );
+			}
+
+			// Editor needs to be recreated every time as there is not flush() or clear() function that can be used after we created an image.
```
adamsilverstein, line 180: Did you try`$editor->__destruct` (or even just $editor = null which should trigger `__destruct`)?  For both Imagick and GD this should clear the image.

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,414 @@
...
  *
- * @since 1.0.0
+ * @param int $attachment_id ID of the attachment being modified.
+ */
+function webp_uploads_restore_image_from_backup( $attachment_id ) {
+	$metadata = wp_get_attachment_metadata( $attachment_id );
+	// This should take place only on the JPEG image.
+	$valid_mime_types = webp_uploads_valid_image_mime_types();
+
+	// Not a supported mime type to create the sources' property.
```
adamsilverstein: stray `'` in this comment

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,414 @@
  * Description: Uses WebP as the default format for new JPEG image uploads if the server supports it.
  * Experimental: No
  *
+ * @since   1.0.0
  * @package performance-lab
- * @since 1.0.0
  */
 
+add_filter( 'wp_generate_attachment_metadata', 'webp_uploads_create_images_with_additional_mime_types', 10, 3 );
```
adamsilverstein, line 11: Generating the additional mime types immediately might fail because it will double execution time, right? I can test this locally, but I think this might break uploads for some users.

Can we instead add an immediately one-time executing cron callback (`wp_schedule_single_event`)? that way the images will get generated on the next callback as part of a separate request?

modules/images/webp-uploads/load.php
mitogh: Correct any edits to an image are performed to the `scaled` version (if present) by WordPress so any subsize should be based out of the modified image and not the original image instead. 

The original image is only used to create the scaled version if not required it would be used as the attached file instead.

modules/images/webp-uploads/load.php
mitogh: Great point, let me do that instead.

modules/images/webp-uploads/load.php
mitogh: No, let me try that thanks for the suggestion.

modules/images/webp-uploads/load.php
```diff
@@ -4,41 +4,414 @@
...
+ * @return string[] An array of valid mime types, where the key is the mime type and the value is the extension type.
+ */
+function webp_uploads_valid_image_mime_types() {
+	$valid_formats = array(
+		'image/jpeg' => 'jpg',
+		'image/webp' => 'webp',
+	);
+
+	/**
+	 * An array representing all the valid mime types where multiple images would be created if
```
adamsilverstein, line 273: We should make sure that by returning only `webp` in the array **only webp images are generated** - that is, the uploads would work the way they do now in the plugin. 

Although we decided this shouldn't be the default approach, it may turn out to be the best approach for most users, and over time it may become our choice for default. Code wise, we should be able to keep the existing plugin approach, but only use it in the case when a single mime type is specified.

adamsilverstein:
@mitogh this looks good so far, I left some small comments and questions.

This PR has grown quite large because in addition to handling image generation you are supporting image edit and restore. I realize these are related because of the data storage, but still wonder if we could break them out so we can focus solely on the image generation and meta data storage. Once we land that, we can tackle the edit/restore bits in a separate PR.

modules/images/webp-uploads/load.php
mitogh: Actually calling `__destruct` seems like the same as creating a new variable altogether as essentially this is called by the engine itself so, for now, will keep it like that. 

One thing that I've noted out of the class `WP_Image_Editor` is the fact that has a property `get_size` but not a `set_size` which would be the one we can use here to reset the size to the appropriate dimensions. Any thoughts on that one ?

modules/images/webp-uploads/load.php
mitogh: Agree I'm going to explore this as part of enchantment on top of this PR [as you describe](https://github.com/WordPress/performance/pull/154#issuecomment-1040833914) has become quite large at this point.

modules/images/webp-uploads/load.php
mitogh: Great point. 

Let me think about how I can go about this use case when only a single mime type is present either JPEG or WebP so I can update this PR accordingly. Another use case we need to consider is the fact that due this is filterable it can be empty as well. Under that scenario we can either:

- Generate all subsizes with the same mime-type as the original image regardless if this filter is empty
- Not generate any subsize to follow what was specified on the filter.

It feels like the last point would be the way to go for consistency but let me know your thoughts on that one.

modules/images/webp-uploads/load.php
mitogh: Actually I just updated this PR to handle this case.

modules/images/webp-uploads/load.php
mitogh: Fixed this one and added a TODO to handle this last point.

felixarntz:
@mitogh @adamsilverstein I think the approach in #147 makes more sense than this one. `_wp_attachment_backup_sizes` is used for edited/cropped versions of an image, while the `sizes` field of the regular image metadata is what is being used for the registered sizes. Since we are focusing on generating the sub-sizes in progressive image formats, I think our data should be in the `sizes` array.

Regarding backward-compatibility, I don't think the approach here justifies going with it given the unexpected data storage, specifically for the following reasons:
* It shouldn't be too hard deleting the "custom" sub-sizes generated in the approach from #147, we just have to use the `delete_attachment` action to delete any of those image files that exist.
* Regarding CDNs, that indeed is a valid concern. However, we also don't break anything in these plugins or WordPress core with the `sources` implementation in #147, and the eventual dev note post about this feature in WordPress core should include guidance on how to update such type of plugins to support the new `sources` field. It's okay to introduce a new API for plugins to use, and since we're adding a new feature to WordPress core, it's reasonable to expect plugin developers that want to support the feature to adopt the new "API" (which here really is just newly stored data).

mitogh:
Closing this one in favor of: 

- #147 

Details https://github.com/WordPress/performance/pull/147#issuecomment-1043145067

PR #157, opened 2022-02-08
WIP: Liveprof

eugene-manuilov:
## Summary

This PR is a PoC for backend metrics measurement that we can use on our local machines to see how our changes affect the performance. This PR uses [liveprof](https://github.com/badoo/liveprof) extension created by Badoo to save metrics collected by [xhprof](https://github.com/tideways/php-xhprof-extension) into the database and [liveprof-ui](https://github.com/badoo/liveprof-ui) GUI to let us view metrics. 

This PR is still WIP and not ready for the final review. Feel free to use it if you want to play with it and see how it works on your end. Any feedback is appreciated.

## Relevant technical choices

* Unfortunately, the `wp-env` tool doesn't allow us to modify the docker-compose environment that it creates, thus we can't install the xhprof and liveprof extensions to the WordPress container that wp-env uses. So, I had to create a separate docker-compose environment that has all required extensions and services.
* I have added four scripts to the package.json file that help working with the new env:
  * `npm start` - starts the new docker-compose env, builds images, installs WordPress and additional database tables if they are missing.
  * `npm stop` - stops the environment
  * `npm run build-images` - builds custom images for the new env. This command is probably less needed and eventually we can delete it.
  * `npm run aggregate` - aggregates metrics collected by xhprof/liveprof extensions to get stats of average value, minimum, maximum, 95 percentile and so on.
* Once the environment is started, there are 3 sites available to use:
  * `http://localhost:8080` - WordPress instance
  * `http://localhost:8081` - phpMyAdmin (probably not needed and can be removed later)
  * `http://lcoalhost:8082` - liverpof-ui

One thing that is nice in the new environment is that we control what we get in the WordPress that is created on initial run. It means that we can generate certain content that we need for testing and everyone will have the same data for testing certain things on certain pages. It also means that we can technically run some test in GitHub actions using this environment as well to make sure our changes doesn't make the performance worth.

## Checklist


adamsilverstein:
> Unfortunately, the wp-env tool doesn't allow use to modify the docker-compose environment that it creates, thus we can't install the xhprof and liveprof extensions to the WordPress container that wp-env uses. So, I had to create a separate docker-compose environment that has all required extensions and services.

Would it be possible to extend wp-env to remove this limitation? I haven't used wp-env much because of it's docker dependency.

> npm run aggregate - aggregates metrics collected by xhprof/liveprof extensions to get stats of average value, minimum, maximum, 95 percentile and so on.

How do you imagine using this? can we export or expose the raw collected data?

eugene-manuilov:
> Would it be possible to extend wp-env to remove this limitation? I haven't used wp-env much because of it's docker dependency.

Unfortunately, the latest version of `wp-env` doesn't allow us to do it.

> How do you imagine using this?

In theory, we can run tests for certain pages in GitHub actions. Let's say we want to make sure that when a user uploads an image, webp generation doesn't reduce the performance of the upload request more than by X%. We create a new workflow file that starts the environment and sends Y post requests to upload an image. Then we run the aggregate command and check the 95 percentile for the tested URL. If it is bigger than expected result more than X%, the workflow throws an error and fails.

> can we export or expose the raw collected data?

All data is stored in the database. So, we can used it however we want to.

adamsilverstein:
> Unfortunately, the latest version of wp-env doesn't allow us to do it.

I meant more could we modify wp-env to support this (or should we pursue another environment?). Since it is part of Gutenberg (https://github.com/WordPress/gutenberg/tree/trunk/packages/env) we could open an issue there to add this feature? I don't fully understand yet how you propose to "allow ... to modify the docker-compose environment that it creates" (since I haven't used wp-env much)

Also, for `xhprof` I feel like this is generally useful and we could open an issue to add that by default to `wp-env`, what do you think? I'm not as sure about `liveprof` since it is third party, not part of core PHP.

eugene-manuilov:
>  I don't fully understand yet how you propose to "allow ... to modify the docker-compose environment that it creates" (since I haven't used wp-env much)

@adamsilverstein there was a typo, I meant "doesn't allow **us** to modify the docker-compose environemnt...". As to modifications:
* we need an additional service that displays metrics (liveprof-ui)
* we need to change the image created for WordPress to install additional extensions
* we need to add additional volumes for the WP container to be able to start profiling in the `wp-config.php` file.

> Since it is part of Gutenberg (https://github.com/WordPress/gutenberg/tree/trunk/packages/env) we could open an issue there to add this feature?

It would be nice, but I suspect it will take a while before it will be implemented and published.

> Also, for `xhprof` I feel like this is generally useful and we could open an issue to add that by default to `wp-env`, what do you think? I'm not as sure about `liveprof` since it is third party, not part of core PHP.

It's useful only if you need to measure PHP metrics. I don't think it's widely needed.

mxbclang:
@eugene-manuilov Just checking on next steps on this one!

eugene-manuilov:
@bethanylang, I put this PR on hold because we focused on webp stuff. I am not whether we are going to continue looking at it or not.

PR #170, opened 2022-02-14
Add performance budget calculator to generate budget.json file

### LINKED ISSUE #145: Performance Budget Calculator and support for `budget.json`
## Feature Description

This feature request is to create/add support for a UI where the user can define their own metrics as some metrics are more important than others depending on the use case or needs of a user. 

The UI would be similar to: https://perf-budget-calculator.firebaseapp.com/

Where a user would be able to define their own metrics, once these metrics have been defined a file called `budget.json` would be created to store all the values from the budget. 

The `budget.json` can be used to perform continuous reviews of the state of the site and can be used in conjunction with

-  #63  

To provide a reporting mechanism to the owner of the site.

Additionally, the tool should support an existing `budget.json` similar to how `theme.json` works for the blocks editor.

gabriel-detassigny:
## Summary

This allows users to generate a budget.json file via an option page.
Range inputs allow to select the budget for each resource type.
Screenshot 2022-02-14 at 14 36 03

Fixes #145 

## Relevant technical choices

The budget.json is generated on the fly when saving the options page.
One issue this current has is it won't trigger the file generation + download if no option has changed since last save. 
This could be solved with a hidden input containing a timestamp, but there might be a better approach? I'm open for suggestions.

## Checklist


gabriel-detassigny:
@mitogh This is a small PoC attempting to solve #145 . It's still a draft, but I would appreciate any feedback on if the approach makes sense or if you envisioned something else.

modules/measurement/budget-calculator/load.php
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Display the content of the budget calculator option page.
+ *
+ * @since 1.0.0
+ */
+function budget_calc_page_content() {
+	echo '<div class="wrap">
+	<h1>Budget Calculator</h1>
+	<form method="post" action="options.php">';
```
mitogh: A `nonce` here would be great.

mitogh: If there's no additional logic here this can be split into

```suggestion
?>
	<div class="wrap">
	<h1>Budget Calculator</h1>
	<form method="post" action="options.php">
<?php
```

modules/measurement/budget-calculator/load.php
```diff
@@ -0,0 +1,165 @@
+<?php
+/**
+ * Module Name: Budget Calculator
+ * Description: Allows users to define their own metrics and generate a budget.json file from them.
+ * Experimental: No
+ *
+ * @package performance-lab
+ * @since 1.0.0
```
mitogh: Replace `1.0.0` with `n.e.x.t` instead here and the the rest of the file to make sure when this is released is replaced with the right version.

modules/measurement/budget-calculator/load.php
```diff
@@ -0,0 +1,165 @@
...
+			);
+		}
+	}
+
+	$data = array( array( 'resourceSizes' => $resource_sizes ) );
+
+	header( 'Content-disposition: attachment; filename=budget.json' );
+	header( 'Content-type: application/json' );
+	echo wp_json_encode( $data );
+	die;
```
mitogh: Replace with `wp_die` instead.

modules/measurement/budget-calculator/load.php
```diff
@@ -0,0 +1,165 @@
...
+				'resourceType' => $type,
+				'budget'       => $value[ $range ],
+			);
+		}
+	}
+
+	$data = array( array( 'resourceSizes' => $resource_sizes ) );
+
+	header( 'Content-disposition: attachment; filename=budget.json' );
+	header( 'Content-type: application/json' );
```
mitogh: Looks like this can be replaced with `wp_send_json`

modules/measurement/budget-calculator/load.php
```diff
@@ -0,0 +1,165 @@
...
+ * @param array  $options  The settings value already set in the database, if any.
+ * @param string $field_id The ID of the field in the database.
+ * @param int    $max      The maximum value that can be set on the range input.
+ */
+function budget_calc_render_range_field( $options, $field_id, $max ) {
+	$text     = isset( $options[ $field_id ] ) ? esc_attr( $options[ $field_id ] ) : 0;
+	$field_id = esc_attr( $field_id );
+	$max      = esc_attr( $max );
+
+	echo "<input type='range' id='$field_id' name='budget_calc_options[$field_id]' min='0' max='$max' step='10'
```
mitogh: Make sure to escape the values of each attribute and probably using the approach described before splitting the HTML and the PHP.

modules/measurement/budget-calculator/load.php
```diff
@@ -0,0 +1,165 @@
...
+ * Render a range input field for the budget calculator options page.
+ *
+ * @since 1.0.0
+ *
+ * @param array  $options  The settings value already set in the database, if any.
+ * @param string $field_id The ID of the field in the database.
+ * @param int    $max      The maximum value that can be set on the range input.
+ */
+function budget_calc_render_range_field( $options, $field_id, $max ) {
+	$text     = isset( $options[ $field_id ] ) ? esc_attr( $options[ $field_id ] ) : 0;
```
mitogh:
```suggestion
	$text     = isset( $options[ $field_id ] ) ? *(int) $options[ $field_id ]  : 0;
```

mitogh (commented):
Great job, some suggestions but overall looks promising @gabriel-detassigny 

Thank you for taking the time to put this together.

modules/measurement/budget-calculator/load.php
gabriel-detassigny: There is already one generated by calling `settings_fields` below 🙂  
Screenshot 2022-03-01 at 14 26 45

It's also already verified on submission too (tried changing the value of the field and I get a "link expired" error from WP)

modules/measurement/budget-calculator/load.php
gabriel-detassigny: I've followed your other suggestion of using `wp_send_json` and that calls `wp_die` already, so no need for a call to `die` there anymore!

gabriel-detassigny:
@mitogh I've addressed your feedback (thanks!), handled a few more edge cases and also added some unit tests.
I think this should now be ready for review.
PS: I don't have access to add the required labels mentioned by the checklist

eclarke1:
@gabriel-detassigny thanks so much - I've added the labels to this PR and to #145 :)

modules/measurement/budget-calculator/load.php
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * Display the content of the budget calculator option page.
+ *
+ * @since n.e.x.t
+ */
+function budget_calc_page_content() {
+	?>
+	<div class="wrap">
+	<h1>Budget Calculator</h1>
```
mukeshpanchal27: Please use the escape function.

`esc_html_e( 'Budget Calculator', 'performance-lab' );`

modules/measurement/budget-calculator/load.php
```diff
@@ -0,0 +1,237 @@
...
+}
+
+/**
+ * Register the option page.
+ *
+ * @since n.e.x.t
+ */
+function budget_calc_menu_page() {
+	add_menu_page(
+		'Budget Calculator',
```
mukeshpanchal27: Use escape function.

`__( 'Budget Calculator', 'performance-lab' ),`

modules/measurement/budget-calculator/load.php
```diff
@@ -0,0 +1,237 @@
...
+function budget_calc_register_settings() {
+	register_setting(
+		'budget_calc_settings',
+		'budget_calc_options',
+		array(
+			'sanitize_callback' => 'budget_calc_sanitize_settings',
+		)
+	);
+
+	add_settings_section( 'budget_calc_settings_id', 'Budget Calculator Metrics', '', 'budget-calculator' );
```
mukeshpanchal27: `add_settings_section( 
	'budget_calc_settings_id',
	__( 'Budget Calculator Metrics', 'performance-lab' ),
	'',
	'budget-calculator'
);`

modules/measurement/budget-calculator/load.php
```diff
@@ -0,0 +1,237 @@
...
+/**
+ * Callback to render an updated hidden field.
+ * This field saves a timestamp of the current time.
+ * Its purpose is to still trigger an update event to download the budget.json file in case nothing was modified.
+ *
+ * @since n.e.x.t
+ */
+function budget_calc_render_hidden_updated_field() {
+	?>
+		<input type="hidden" value="<?php echo time(); ?>" name="budget_calc_options[updated]">
```
mukeshpanchal27: `<input type="hidden" value="<?php echo time(); ?>" name="budget_calc_options[updated]" />`

modules/measurement/budget-calculator/load.php
```diff
@@ -0,0 +1,237 @@
...
+		array(
+			'class' => 'hidden',
+		)
+	);
+}
+add_action( 'admin_init', 'budget_calc_register_settings' );
+
+/**
+ * Callback to sanitize the input from the settings form.
+ *
```
mukeshpanchal27, line 111: `* @since n.e.x.t`

mukeshpanchal27 (changes requested):

modules/measurement/budget-calculator/load.php
gabriel-detassigny: 👍 All done now

mukeshpanchal27: Thank you for fix it @gabriel-detassigny

aristath:
Not everyone is familiar with the concept of a "performance budget". 
Could we explain it somehow, describe what this "budget calculator" is, and how it can/should be used? Otherwise, if I see "budget calculator" on my dashboard, my mind will immediately go to a _financial_ budget calculator, which is obviously completely unrelated.

gabriel-detassigny:
@aristath Thanks for the feedback. You're right, it lacked a bit of explanation. I've changed the module's description a bit to give more context as that will be an editor's first encounter with this. Then included a brief paragraph with more details on the form itself. If you have any wording suggestion let me know!

PR #177, opened 2022-02-18
Run phpcbf and phpcs before commit

### LINKED ISSUE #175: Automate PHPCS/PHPCBF run after commit with Husky
## Feature Description

We're expecting developers to run `phpcs/phpcbf` manually before creating a PR. This can be automated easily via [husky](https://typicode.github.io/husky/). Husky is also [used by Gutenberg](https://github.com/WordPress/gutenberg/blob/trunk/.husky/pre-commit) to lint before commit.

So here are things that we need to do:
1. Add a pre-commit hook via husky to run `phpcbf` before commit to auto-fix phpcs errors. 
2. Run `phpcs`  so that users can manually resolve any errors which are not automatically fixable.

kirtangajjar:
## Summary

Fixes #175 

Adds support for running `phpcbf` and `phpcs` before commit using `husky` and `lint-staged` npm packages.

[Husky](https://github.com/typicode/husky) is a tool used to manage git hooks and [lint-staged](https://github.com/okonet/lint-staged) ensures that phpcs/phpcbf is only run on staged files, not the whole repo.

This git hook will first run `phpcbf` to fix all phpcs errors it can fix. Then it will run `phpcs` to display errors that it cannot fix.

If there are phpcs issues that the author has not fixed, it won't allow them to commit.


kirtangajjar:
JS linting is failing on this PR since it seems the implementation is a little broken. 

[Linting uses prettier](https://github.com/WordPress/gutenberg/blob/trunk/packages/scripts/scripts/format.js#L31), but it isn't been defined in `package.json` [as it's been done in Gutenberg](https://github.com/WordPress/gutenberg/blob/trunk/package.json#L197). 

I can create a separate PR to fix it or fix it in this PR itself whichever seems to be a better way.

kirtangajjar:
@eclarke1 Can you please add a "needs review" label and add @felixarntz as a reviewer?

eclarke1:
@kirtangajjar absolutely! All updated, thank you so much for reaching out.

package.json
```diff
@@ -40,6 +42,13 @@
...
     "pretest-php-multisite": "wp-env run composer 'install --no-interaction'",
     "test-php-multisite": "wp-env run phpunit 'WP_MULTISITE=1 phpunit -c /var/www/html/wp-content/plugins/performance/tests/multisite.xml --verbose'",
-    "wp-env": "wp-env"
+    "wp-env": "wp-env",
+    "prepare": "husky install"
+  },
+  "lint-staged": {
+    "*.php": [
+      "phpcbf --standard=phpcs.xml.dist",
+      "phpcs --standard=phpcs.xml.dist"
```
mitogh: I think this one should be `npm run lint-php` instead

Same as the above ideally the command should run from within the docker image in order to ensure it uses the right version from `composer.lock` file.

package.json
```diff
@@ -40,6 +42,13 @@
     "test-php": "wp-env run phpunit 'phpunit -c /var/www/html/wp-content/plugins/performance/phpunit.xml.dist --verbose'",
     "pretest-php-multisite": "wp-env run composer 'install --no-interaction'",
     "test-php-multisite": "wp-env run phpunit 'WP_MULTISITE=1 phpunit -c /var/www/html/wp-content/plugins/performance/tests/multisite.xml --verbose'",
-    "wp-env": "wp-env"
+    "wp-env": "wp-env",
+    "prepare": "husky install"
```
mitogh, line 56: Wondering the `prepare` as I don't think we are actually packaging the plugin as a new version.

package.json
dainemawer: @kirtangajjar we probably do not want to run `phpcbf` on staged files - that would make changes in the track. And potentially unwanted changes that we may not be able to get around. I would say just running `phpcs` or `npm run lint-php` would suffice here.

package.json
dainemawer: Where is `prepare` being run?

dainemawer (commented):
Nice work!

package.json
erikyo: I guess it would be more useful a precommit hook with husky that runs test lint and so on
https://khalilstemmler.com/blogs/tooling/enforcing-husky-precommit-hooks/#Configuring-Husky

Anyhow this is the same husky setup used in the Gutenberg repo, which is fine.

kirtangajjar: @mitogh @dainemawer `prepare` is a [lifecycle event](https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-scripts) in NPM. [Here](https://docs.npmjs.com/cli/v8/using-npm/scripts#npm-install) is the sequence in which events are run in when running `npm install`. I have put `husky install` in package.json as [mentioned in](https://typicode.github.io/husky/#/?id=install) husky's docs. 

Basically, we need to run `husky install` once after cloning the repo. [What it does is](https://stackoverflow.com/a/66786333/2867422), add the `.husky` directory to git's `core.hooksPath`

package.json
kirtangajjar: Fixed.

kirtangajjar:
@dainemawer @mitogh Thank you for the feedback! I have made the changes. Please have a look at it again.

erikyo:
@kirtangajjar about https://github.com/WordPress/performance/pull/177#issuecomment-1044602984 I guess the checks was failing because was missing the .eslintignore file with this content:
```
*.min.js
node_modules
```

kirtangajjar:
@erikyo I think that's happening because we haven't added prettier in `package.json` https://github.com/WordPress/performance/runs/5303982931?check_suite_focus=true#step:8:13.

P.S. Gutenberg [has it too](https://github.com/WordPress/gutenberg/blob/trunk/package.json#L197).

erikyo:
that strange since prettier is a direct dependency of [@wordpress/eslint-plugin](https://github.com/WordPress/gutenberg/blob/b63aa744a3610a5ec881e220876d904bf9301783/packages/eslint-plugin/package.json)

the `"lint-js": "wp-scripts lint-js ./bin"`, is targeting [this who is looking for eslint files](https://github.com/WordPress/gutenberg/blob/trunk/packages/scripts/scripts/lint-js.js) and that the reason why i guess so.

mitogh (approved):

erikyo:
@kirtangajjar [this](https://github.com/erikyo/woocommerce-radio-buttons/actions) is a working setup for actions inspired to the performance repo i've done today, and the only dependency is the `"@wordpress/scripts": "^21.0.2"`

maybe isn't working because the outdated wpscript version? prettier is a dependency of multiple modules and to me it is not the cause of the failed test

kirtangajjar:
@erikyo As you can see above, I just updated the version of `@wordpress/scripts` as you mentioned and it didn't solve the issue. Even on my local I nuked `node_modules` and I reinstalled it and it didn't install prettier. Perhaps because of this error that was thrown in the console:
```
npm WARN eslint-plugin-prettier@3.4.1 requires a peer of prettier@>=1.13.0 but none is installed. You must install peer dependencies yourself.
```
As per [their blog](https://blog.npmjs.org/post/110924823920/npm-weekly-5)

> We will also be changing the behaviour of peerDependencies in npm@3. We won’t be automatically downloading the peer dependency anymore. Instead, we’ll warn you if the peer dependency isn’t already installed. This requires you to resolve peerDependency conflicts yourself, manually, but in the long run, this should make it less likely that you’ll end up in a tricky spot with your packages’ dependencies.

My guess is this is why Gutenberg has prettier explicitly in its dependency too for this exact same reason.

I have used [wp-prettier](https://www.npmjs.com/package/wp-prettier) instead of prettier as it's a fork of prettier which is also [used by Gutenberg](https://github.com/WordPress/gutenberg/blob/trunk/package.json#L197). The changes made by WP team is documented in its readme:
```
This is a fork of Prettier that adds a new command line option --paren-spacing which inserts many extra spaces inside parentheses, the way how projects in the WordPress ecosystem (Calypso, Gutenberg, etc.) like to format their code.
```

dainemawer (approved):

kirtangajjar:
I found out that the current implementation wasn't working for me in my local env. The problem was that we were trying to run the lint command inside of the container with `husky` and `lint-staged`. 

When we commit any change, `husky` runs the `lint-staged` command, which runs the specified command, but only on the files that are changed. So the command becomes `wp-env run composer run-script lint '/home/kirtan/performance.local/app/public/wp-content/plugins/performance/load.php'`. This command is being run inside the container so the container will try to find the path to file inside the container and will throw an error at this point as that path is not present inside the container. It's only present on the host.

To remediate this, I think the best solution is to just run the command directly on the host. I think it's fair to assume most people will have composer and phpcs installed globally. So I have gone ahead with the change for now. 

@mitogh @dainemawer LMK if you have any feedback regarding this.

mitogh:
I think that's a fair assumption @kirtangajjar, however I tested this locally and the husky command was not running when I staged some files. 

When running the script  `npx lint-stagged` the following output is returned:

> Configuration could not be found.

dainemawer:
Yeah agreed, I think thats solid. Nice work @kirtangajjar !

kirtangajjar:
@mitogh 
> the husky command was not running when I staged some files.

The husky command will run after you commit anything(because of the pre-commit hook). However, for testing purposes, staging some changes and running `npx lint-staged` will have the same effect.

> When running the script npx lint-stagged the following output is returned:
> > Configuration could not be found.


That's really strange. The configuration of `lint-staged` is present in [`package.json`]( https://github.com/WordPress/performance/pull/177/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R50). Are you sure that you checked out the correct branch? 

I performed the following steps in a newly cloned repo and it was running fine for me:
```
git clone https://github.com/WordPress/performance
git checkout -b feature/add-husky
git pull https://github.com/kirtangajjar/performance.git feature/add-husky
# Change admin/load.php
git add --all
npx lint-staged
```

I tried these same steps on a remote machine as well and it worked there too.

mitogh:
> The husky command will run after you commit anything(because of the pre-commit hook)

Then this is a post-commit hook, not a pre-commit as the title of this PR indicates, from my perspective, this should run before the commit to ensuring all the appropriate lints are executed before a commit.  @kirtangajjar 

Looks like had to pull out from your fork as your branch was not found locally, the linting happen correctly before the commit and when running the standalone script `lint-stagged`.

So all good from my end  @kirtangajjar

package.json
```diff
@@ -25,7 +25,10 @@
     "chalk": "4.1.1",
     "commander": "4.1.0",
     "fast-glob": "^3.2.7",
-    "lodash": "4.17.21"
+    "husky": "^7.0.4",
+    "lint-staged": "^12.3.4",
+    "lodash": "4.17.21",
+    "prettier": "npm:wp-prettier@^2.2.1-beta-1"
```
felixarntz: What's the reasoning for using a `wp-prettier` beta version here for the `prettier` dependency? That looks potentially problematic to me 🤔

felixarntz (commented):
@kirtangajjar One question here, other than that lgtm.

package.json
kirtangajjar: Hey @felixarntz, 

I explained that choice in [this comment](https://github.com/WordPress/performance/pull/177#issuecomment-1049756635). Pasting relevant contents here for easy reference:

> I have used [wp-prettier](https://www.npmjs.com/package/wp-prettier) instead of prettier as it's a fork of prettier which is also [used by Gutenberg](https://github.com/WordPress/gutenberg/blob/50b5bfa7de660d41a3b0e0693bf3fc9fc56836d8/package.json#L200). The changes made by the WP team is documented in its readme:
> ```
> This is a fork of Prettier that adds a new command-line option --paren-spacing which inserts many extra spaces inside parentheses, the way how projects in the WordPress ecosystem (Calypso, Gutenberg, etc.) like to format their code.
> ``` 

Basically. we needed prettier for dependency and then I checked [Gutenberg repo](https://github.com/WordPress/gutenberg/blob/50b5bfa7de660d41a3b0e0693bf3fc9fc56836d8/package.json#L200) and they were using this exact version, so I used the same one. 

I don't think the beta tag in the version should be a problem as this package is an internal fork of `prettier` maintained by WordPress. Also,  the given version, `2.2.1-beta-1`, was released 1 year ago and has been used by Gutenberg since then so I think it's safe to use.

package.json
```diff
@@ -41,6 +44,12 @@
     "test-php": "wp-env run phpunit 'phpunit -c /var/www/html/wp-content/plugins/performance/phpunit.xml.dist --verbose'",
     "pretest-php-multisite": "wp-env run composer 'install --no-interaction'",
     "test-php-multisite": "wp-env run phpunit 'WP_MULTISITE=1 phpunit -c /var/www/html/wp-content/plugins/performance/tests/multisite.xml --verbose'",
-    "wp-env": "wp-env"
+    "wp-env": "wp-env",
+    "prepare": "husky install"
+  },
+  "lint-staged": {
+    "*.php": [
+      "composer run-script lint"
```
eugene-manuilov, line 60: It should be `npm run lint-php`, shouldn't it?

kirtangajjar: Running `npm run lint-php` executes the commands in the docker environment. It is primarily used in Gutenberg for [running unit tests in CI](https://github.com/WordPress/gutenberg/blob/74a99b2cfce30a49d4c007960f24314ba1a1983e/.github/workflows/unit-test.yml#L78).  We were using it earlier, but we faced an issue as discussed [here](https://github.com/WordPress/performance/pull/177#issuecomment-1059080228). That's why we went with `composer run-script lint` as it ran on host.

eugene-manuilov: Ok, then, i think, we need to update the `lint-php` command to use `composer run-script lint` for consistency. Right?

kirtangajjar: I don't think it's necessary as we may need to run the linting command in the docker environment in future during unit testing(as done by Gutenberg). 

So I think it's okay to keep it as-is. Similar to what's here, even Gutenberg has different commands in `package.json` and `composer.json`. We don't need to emulate them every time, but I don't see any harm in keeping it as-is.

felixarntz: I agree with @kirtangajjar. `npm run lint-php` acts in the Docker container, like most of the other `npm` scripts, so that it's easier to set up and doesn't depend on the local environment. So we shouldn't change that.

Based on the comment https://github.com/WordPress/performance/pull/177#issuecomment-1059080228, it's a fair point to use `composer run-script lint` here, but it feels also like a workaround to me. That may be okay, since linting is basic enough so that it would still work from just running `composer install` and `npm install`.

The main thing here is that we shouldn't require contributors to do anything else in order to use the critical features. For example, `composer run-script test` must never become the only command to run tests, as that would require a lot more to set up on your environment.

felixarntz (commented):
@kirtangajjar This looks good to me. Can you resolve the merge conflict here?

kirtangajjar:
@felixarntz Thank you. The merge conflicts have been resolved!

mitogh (approved):

felixarntz:
@kirtangajjar Apologies, this just now got a second approval and there is now another merge conflict to resolve due to the version number change. @eugene-manuilov has rightfully raised this in #252 so that this won't happen again in the future.

mitogh (approved):

felixarntz (approved):

PR #178, opened 2022-02-18
NPM script to update `@since n.e.x.t` tags

### LINKED ISSUE #90: Automate @since tags generation using the `n.e.x.t` placeholders
Currently, we use `1.0.0` for `@since` tags in the codebase. This is fine for now since we haven't released an actual version yet. Later on it may not work because some already implemented tickets might be postponed and included into later releases which means that we will need amend `@since` tags manually to match the release number.

What we can do is to start using `@since n.e.x.t` tags for new functions, methods, filters and actions instead of entering the exact release number. Then we can implement an npm script that will replace `n.e.x.t` with the correct version.

eugene-manuilov:
## Summary

Fixes #90

## Checklist

### DIFF
--- a/bin/plugin/commands/since.js
+++ b/bin/plugin/commands/since.js
@@ -47,7 +47,7 @@ exports.handler = async ( opt ) => {
 		ignore: [ __filename, '**/node_modules', '**/vendor' ],
 	} );
 
-	const regexp = new RegExp( '@since n.e.x.t' );
+	const regexp = new RegExp( '@since n.e.x.t', 'g' );
 	files.forEach( ( file ) => {
 		const content = fs.readFileSync( file, 'utf-8' );
 		if ( regexp.test( content ) ) {
--- a/bin/plugin/commands/since.js
+++ b/bin/plugin/commands/since.js
@@ -1,4 +1,9 @@
 /**
+ * Internal dependencies
+ */
+const { log, formats } = require( '../lib/logger' );
+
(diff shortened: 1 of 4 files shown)


bin/plugin/commands/since.js
```diff
@@ -0,0 +1,60 @@
...
+	const patterns = [
+		path.resolve( __dirname, '../../../**/*.php' ),
+		path.resolve( __dirname, '../../../**/*.js' ),
+	];
+
+	const files = await glob( patterns, {
+		ignore: [ __filename, '**/node_modules', '**/vendor' ],
+	} );
+
+	const regexp = new RegExp( '@since n.e.x.t' );
```
felixarntz: Doesn't this need to have a global modifier? Otherwise it'll only replace one occurrence of this in each file.

felixarntz (commented):
@eugene-manuilov Looks great, only one point of feedback.

bin/plugin/commands/since.js
eugene-manuilov: Good catch. You are absolutely right, it should have the global modifier. Updated.

felixarntz (approved):
@eugene-manuilov Thanks, this looks great!

cc @JustinyAhin for the updated release instructions

JustinyAhin (approved):
This looks good to me too.
Thanks @eugene-manuilov

PR #188, opened 2022-02-24
Support retry mechanism for generating sub-sizes in additional MIME types on constrained environments.

mitogh:
## Summary

Fixes:

-  #142 

Requires:

- #147 

## Relevant technical choices

The following code uses the filter `wp_get_missing_image_subsizes` as an action in order to trigger the execution of additional images when those images do not exist yet when the image is created under constrained environments.  If the server timeouts WordPress will retry until a maximum of 5 times. Every retry would only process non-existing images on the metadata.

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -42,60 +46,55 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
-		$image_sizes = $metadata['sizes'];
-	}
-
-	foreach ( wp_get_registered_image_subsizes() as $size_name => $properties ) {
-		// This image size does not exist on the defined sizes.
-		if ( ! isset( $image_sizes[ $size_name ] ) || ! is_array( $image_sizes[ $size_name ] ) ) {
+	$dirname = pathinfo( $file, PATHINFO_DIRNAME );
+	foreach ( $metadata['sizes'] as $size_name => $properties ) {
+		// This image size is not defined or not an array.
+		if ( ! isset( $metadata['sizes'][ $size_name ] ) || ! is_array( $metadata['sizes'][ $size_name ] ) ) {
```
felixarntz: At least the first check here is now redundant since we're iterating through exactly this array.
```suggestion
		if ( ! is_array( $properties ) ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -42,60 +46,55 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+		if ( isset( $properties['mime-type'] ) ) {
+			$current_mime = $properties['mime-type'];
+		} elseif ( isset( $properties['file'] ) ) {
+			$current_mime = wp_check_filetype( $properties['file'] )['type'];
 		}
 
+		// The mime type can't be determined.
 		if ( empty( $current_mime ) ) {
 			continue;
 		}
```
felixarntz, line 67: Let's move all this block of code up, above the logic in lines 56-59 where it adds the `sources` key. We shouldn't change anything if the MIME type cannot be determined.

In theory it's okay to still add the `sources` key, but if we're not gonna add anything to it, we also don't need to add it.

modules/images/webp-uploads/load.php
```diff
@@ -42,60 +46,55 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
 
-		$current_size = $image_sizes[ $size_name ];
-		$sources      = array();
-		if ( isset( $current_size['sources'] ) && is_array( $current_size['sources'] ) ) {
-			$sources = $current_size['sources'];
+		// Ensure a `sources` property exists on the existing size.
+		if ( empty( $metadata['sizes'][ $size_name ]['sources'] ) || ! is_array( $metadata['sizes'][ $size_name ]['sources'] ) ) {
+			$metadata['sizes'][ $size_name ]['sources'] = array();
 		}
```
felixarntz: Throughout this entire loop, the code would be much easier to read if we always looked at the `$properties` variable which is the same as `$metadata['sizes'][ $size_name ]`. The only difference is that at the very end we have to reassign the whole variable like `$metadata['sizes'][ $size_name ] = $properties;` so that the data is also updated outside of this loop.

This feedback applies to the entire `foreach` loop, here specifically it would become
```suggestion
		if ( empty( $properties['sources'] ) || ! is_array( $properties['sources'] ) ) {
			$properties['sources'] = array();
		}
```

modules/images/webp-uploads/load.php
```diff
@@ -42,60 +46,55 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+				'file'     => array_key_exists( 'file', $properties ) ? $properties['file'] : '',
+				'filesize' => 0,
+			);
+			// Set the filesize from the current mime image.
+			$file_location = path_join( $dirname, $properties['file'] );
+			if ( file_exists( $file_location ) ) {
+				$metadata['sizes'][ $size_name ]['sources'][ $current_mime ]['filesize'] = filesize( $file_location );
+			}
+			wp_update_attachment_metadata( $attachment_id, $metadata );
 		}
```
felixarntz: See above, we can make this easier to read:
```suggestion
		if ( empty( $properties['sources'][ $current_mime ] ) ) {
			$properties['sources'][ $current_mime ] = array(
				'file'     => isset( $properties['file'] ) ? $properties['file'] : '',
				'filesize' => 0,
			);
			// Set the filesize from the current mime image.
			$file_location = path_join( $dirname, $properties['file'] );
			if ( file_exists( $file_location ) ) {
				$properties['sources'][ $current_mime ]['filesize'] = filesize( $file_location );
			}
			$metadata['sizes'][ $size_name ] = $properties;
			wp_update_attachment_metadata( $attachment_id, $metadata );
		}
```

modules/images/webp-uploads/load.php
```diff
@@ -42,60 +46,55 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
 
 		$formats = isset( $valid_mime_transforms[ $current_mime ] ) ? $valid_mime_transforms[ $current_mime ] : array();
 
 		foreach ( $formats as $mime ) {
-			if ( empty( $sources[ $mime ] ) ) {
+			if ( empty( $metadata['sizes'][ $size_name ]['sources'][ $mime ] ) ) {
```
```suggestion
			if ( empty( $properties['sources'][ $mime ] ) ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -42,60 +46,55 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
 				$source = webp_uploads_generate_image_size( $attachment_id, $size_name, $mime );
 				if ( is_array( $source ) ) {
-					$sources[ $mime ] = $source;
+					$metadata['sizes'][ $size_name ]['sources'][ $mime ] = $source;
+					wp_update_attachment_metadata( $attachment_id, $metadata );
```
```suggestion
					$properties['sources'][ $mime ] = $source;
					$metadata['sizes'][ $size_name ] = $properties;
					wp_update_attachment_metadata( $attachment_id, $metadata );
```

modules/images/webp-uploads/load.php
```diff
@@ -42,60 +46,55 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
 				}
 			}
 		}
-
-		$current_size['sources']         = $sources;
-		$metadata['sizes'][ $size_name ] = $current_size;
```
felixarntz, line 98: Based on the above, just to be safe, we should probably still include a line `$metadata['sizes'][ $size_name ] = $properties;` here.

Using `$properties` throughout makes the code in the foreach loop easier to follow, and only at the end (or when updating data) we need to use the `$metadata` variable and update it.

felixarntz (commented):
@mitogh Left a few minor points of initial feedback, specifically to simplify the code for readability.

In any case, we should _either_ always use `$metadata['sizes'][ $size_name ]` _or_ always use `$properties`, but not mixed like it is currently here, sometimes checking one, other times the other.

modules/images/webp-uploads/load.php
```diff
@@ -285,3 +284,53 @@ function webp_uploads_remove_sources_files( $attachment_id ) {
...
+function _webp_uploads_is_valid_rest_for_post_process( $body = '' ) {
+	$matches     = array();
+	$body        = json_decode( $body, true );
+	$valid_rest  = defined( 'REST_REQUEST' ) && REST_REQUEST;
+	$valid_route = $valid_rest && isset( $GLOBALS['wp'], $GLOBALS['wp']->query_vars['rest_route'] ) && preg_match( '/media\/\d+\/post-process/', $GLOBALS['wp']->query_vars['rest_route'], $matches );
+
+	return $valid_route && ! empty( $matches ) && ! empty( $body['action'] ) && 'create-image-subsizes' === $body['action'];
+}
+
+add_filter( 'wp_get_missing_image_subsizes', 'webp_uploads_wp_get_missing_image_subsizes', 10, 3 );
```
felixarntz, line 342: Please move this call right below the callback function so that they're colocated.

modules/images/webp-uploads/load.php
```diff
@@ -285,3 +284,53 @@ function webp_uploads_remove_sources_files( $attachment_id ) {
...
+ * @return bool `true` if the current request is a valid REST request to process
+ */
+function _webp_uploads_is_valid_rest_for_post_process( $body = '' ) {
+	$matches     = array();
+	$body        = json_decode( $body, true );
+	$valid_rest  = defined( 'REST_REQUEST' ) && REST_REQUEST;
+	$valid_route = $valid_rest && isset( $GLOBALS['wp'], $GLOBALS['wp']->query_vars['rest_route'] ) && preg_match( '/media\/\d+\/post-process/', $GLOBALS['wp']->query_vars['rest_route'], $matches );
+
+	return $valid_route && ! empty( $matches ) && ! empty( $body['action'] ) && 'create-image-subsizes' === $body['action'];
+}
```
felixarntz: Based on what we discussed yesterday, there is a cleaner solution that having these two functions - we shouldn't need to specifically check whether it's an AJAX or REST request. What about if any other logic generates image sub-sizes via `wp_update_image_subsizes()`? Your current code wouldn't support that.

Let's remove these functions and instead implement a more stable solution based on what we talked about yesterday:
* In `webp_uploads_create_sources_property()`, after the initial checks but before starting the `foreach` loop, we can set a custom meta flag to indicate that this attachment is awaiting additional MIME type generation. We could use e.g. `_webp_uploads_awaiting_additional_mime_subsizes` as name.
* We should set this meta key on the attachment (e.g. via `update_post_meta()`), only if it is not set yet, so that it is not set over and over again on subsequent calls.
* Right after the foreach loop (i.e. if the whole process was successful), we should unset that flag again (e.g. via `delete_post_meta()`).
* In your above `webp_uploads_wp_get_missing_image_subsizes()` function, we can then check if `empty( $missing_sizes ) && get_post_meta( $attachment_id, '_webp_uploads_awaiting_additional_mime_subsizes', true )` instead of the less reliable check for whether this is an AJAX or REST request.

modules/images/webp-uploads/load.php
```diff
@@ -285,3 +284,53 @@ function webp_uploads_remove_sources_files( $attachment_id ) {
...
+ * @param array[] $missing_sizes Associative array of arrays of image sub-size.
+ * @param array   $image_meta    The metadata from the image.
+ * @param int     $attachment_id The ID of the attachment.
+ *
+ * @return array[] $missing_sizes Associative array of arrays of image sub-size.
+ */
+function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta, $attachment_id ) {
+	if ( _webp_uploads_is_valid_ajax_for_image_sizes() || _webp_uploads_is_valid_rest_for_post_process( file_get_contents( 'php://input' ) ) ) {
+		webp_uploads_create_sources_property( $image_meta, $attachment_id );
+	}
```
felixarntz: We should only run any additional checks here if `$missing_sizes` is empty. If it's not empty, our other hook will trigger anyway, and then we shouldn't use the hack here.

More importantly though, please review my other comment, I don't think we should be checking for whether this is an AJAX or REST request here since it's unreliable.

felixarntz (commented):
@mitogh This looks great. The main thing that needs iteration though is using a more reliable approach for the "hack" here - we shouldn't be checking for whether this is an AJAX or REST request as that would mean the code doesn't support any other calls to `wp_update_image_subsizes()`. While this is a hack, it should still be reliable.

modules/images/webp-uploads/load.php
mitogh: Good point.

modules/images/webp-uploads/load.php
mitogh: Actually, we are doing 2 operations here read and write, while doing the read we can use properties but for write operations, we need to make sure the `metadata` is updated as it removes the need to create an auxiliary variable and we can call `wp_update_attachment_metadata` with the`metadata` variable instead of the `properties` one.

mitogh: I see there are some places that can be updated to make this simpler tho, so will update accordingly.

modules/images/webp-uploads/load.php
mitogh: Good point about `missing_sizes` being empty.

modules/images/webp-uploads/load.php
mitogh: Actually, I was looking at the code, and having the meta just does not provide much value add is a meta that won't be used after the merge in core so there's no real benefit in keeping it together as if the function is executed and all sizes are looped the image sizes would be updated only if they do not exist which is exactly what we aim with the meta but without the need to define a new meta key and keep track of it in multiple places.

On the other hand, we know from the code the retry mechanism would be executed only from 2 places from ajax and REST, so the conditionals are in place to prevent exactly that execute the logic to create missing images in a context where was not requested, having the conditionals ensure that even if this code is executed from outside of those 2 contexts we are not executing that logic.

However, if the aim is to review this logic anytime this function is called regardless of the context we can just remove the conditionals and execute the function `webp_uploads_create_sources_property`as there's no real need to have an additional meta if the logic requires image generation would do so if not it would just loop into every image size skipping over every single one of them.

felixarntz: @mitogh 
> is a meta that won't be used after the merge in core so there's no real benefit

Indeed this is a workaround which we only need because there's no appropriate action for this in core. But the fact that we won't need it later in core IMO isn't a good argument against it. The current logic you have here checking for AJAX or REST request won't be needed in WP core either.

Also, the meta key will always be set on an attachment only temporarily and be removed as soon as all the image formats generation has completed, so it won't even stay around in the future, so we're not "polluting" the database in any way.

> On the other hand, we know from the code the retry mechanism would be executed only from 2 places from ajax and REST

I think this is a problematic conclusion: We know that that is the case from WP core itself, but we have to keep in mind the existence of plugins. Any plugin may integrate with this functionality, a plugin could potentially call `wp_update_image_subsizes()` directly. For example, what if there was a GraphQL plugin facilitating image uploads and using this function? The only way to have the implementation here reliable is to find a solution that works only based on the `wp_update_image_subsizes()` function, we can't really make an assumption where this is called or not.

> we can just remove the conditionals and execute the function `webp_uploads_create_sources_property`as there's no real need to have an additional meta

That won't work, because then the code may run unexpectedly and accidentally create image sizes even when the `wp_get_missing_image_subsizes()` function is called _just_ to list the missing sub-sizes (which happens e.g. in one of the REST endpoints). Setting an explicit flag (e.g. using the meta key) in our logic to generate the additional format sub-sizes, at a point when we _really_ know it should generate them, makes this happen on-demand.

Without such a flag, even when somebody calls the REST API to get information about the image 3 months after the upload, our logic would still try to run. Yes, it wouldn't actually generate image sizes then since they will all exist already, but even attempting to run the logic in this context is wasteful - and potentially even problematic from a security perspective, since that would mean a write action is happening during a REST API `GET` (read) request.

felixarntz: @mitogh Actually, I just tried another solution using `debug_backtrace()` that should mitigate both your and my concerns above. Really, what we need to check is whether `wp_get_missing_image_subsizes()` was called within `wp_update_image_subsizes()` - that's our condition. It's still hacky, but that's really the most reliable solution I think. Here's some demo code I used to try this out:
```php
add_filter(
	'wp_get_missing_image_subsizes',
	function( $missing_sizes, $image_meta, $attachment_id ) {
		if ( ! empty( $missing_sizes ) ) {
			return $missing_sizes;
		}
		// Should be at position 4, but look into the first 10 just to be safe.
		// Ignore 'args' to speed up the function execution.
		$trace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS, 10 );
		foreach ( $trace as $element ) {
			if ( isset( $element['function'] ) && 'wp_update_image_subsizes' === $element['function'] ) {
				// Call our function to generate more image formats here.
				break;
			}
		}
		return $missing_sizes;
	},
	10,
	3
);
```

It's a bit weird, but I actually think this is less hacky than the two other solutions we've looked at so far. What do you think?

mitogh: > I think this is a problematic conclusion: We know that that is the case from WP core itself, but we have to keep in mind the existence of plugins. Any plugin may integrate with this functionality, a plugin could potentially call wp_update_image_subsizes() directly. For example, what if there was a GraphQL plugin facilitating image uploads and using this function? The only way to have the implementation here reliable is to find a solution that works only based on the wp_update_image_subsizes() function, we can't really make an assumption where this is called or not.

Correct however we are under the assumption that the retry mechanism would be executed from within the core mechanism, if an upload is processed via a GraphQL plugin and the request fails there's no retry mechanism there that would be used so is a safer assumption to introduce the retry logic only in 2 places where we know for a fact is happening Ajax and REST

From my perspective, the way it should work is if a plugin fails to upload it should rely on WP core to add the missing images using the same APIs from WP Core, not the other way around fixing the plugins from the WP Core side. In this case, a plugin failing to process images should use the REST endpoints created for this reason, Which then would trigger the logic that we have in place.

> Without such a flag, even when somebody calls the REST API to get information about the image 3 months after the upload, our logic would still try to run. Yes, it wouldn't actually generate image sizes then since they will all exist already, but even attempting to run the logic in this context is wasteful - and potentially even problematic from a security perspective, since that would mean a write action is happening during a REST API GET (read) request.

Right, however even with a meta key in place this also can happen as if a meta exists and the maximum number of retries are executed the meta won't be removed, next time this function is executed from a GET context the meta would execute the logic to regenerate the images even when that's not the expected behavior.

From my perspective instead of limiting the scope from where this can be executed, we allow for safer usage of the function in other contexts like listing all missing sizes from an image and not actually triggering the regeneration of image sizes.

mitogh: Fantastic, the `debug_backtrace`  seems like the way to go :100: agree on that one.

adamsilverstein:
Looking good, nice work @mitogh & @felixarntz! I'll give this a test in my constrained environment.

modules/images/webp-uploads/load.php
```diff
@@ -285,3 +288,36 @@ function webp_uploads_remove_sources_files( $attachment_id ) {
...
+ * @param array   $image_meta    The metadata from the image.
+ * @param int     $attachment_id The ID of the attachment.
+ *
+ * @return array[] $missing_sizes Associative array of arrays of image sub-size.
+ */
+function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta, $attachment_id ) {
+	$trace = array();
+	// Only setup the trace if we no longer have more sizes.
+	if ( empty( $missing_sizes ) ) {
+		$trace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS, 10 );
```
adamsilverstein: 10 seems a little high here :)

adamsilverstein: Also, might be good to add an inline doc explaining why we use `debug_backtrace` since it is an odd thing to see in code, also that it would not be required for core.

mitogh: Great idea, just added some comments.

adamsilverstein:
@mitogh & @felixarntz I tested this locally with a 5 second timeout and a large image upload. The retries continued until all webp images were generated, even though it took a few tries (a total of 4 retries in my case). 

I tested uploading in both the media library and directly in the block editor and the retry mechanism worked correctly in both.

One difference I noticed is that in the editor, the image shows up greyed out immediately when the upload starts with a spinner over it:
image 
the spinner and greyed out image are a good indicator to the user that something is still happening.

The media editor on the other hand shows this sad grey box with a "progress" bar that stays full all throughout the regeneration process (I think it shows upload percent as the file uploads, but then remains full until subsize gerneration completes):
image

We may have a trac ticket for it already, if not I will create one: two small improvements for the media library uploader would be:
1. switch from the spinner to a progress bar
2. immediately show a greyed out thumbnail when the user drops the image to upload

I also tested what happens if you navigate away during the file regeneration. Neither the media library or the post screen warn about navigating away while images are still being processed (after you have hit publish or save draft in the editor), so if you start an upload with a constrained system, then navigate away, the retries will **not occur** and you get left with an incomplete set of images. 

For now I think we should at least warn users who try to navigate away that image processing isn't complete (similar to how we warn users they have unsaved changes on a post if they try to navigate away) - I can create a follow up trac ticket for that if there isn't one already. In addition, when we build out image regeneration, we can make sure any missing images get filled in.

adamsilverstein (approved):

modules/images/webp-uploads/load.php
```diff
@@ -42,60 +46,59 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
-			if ( empty( $sources[ $mime ] ) ) {
+			if ( empty( $properties['sources'][ $mime ] ) ) {
 				$source = webp_uploads_generate_image_size( $attachment_id, $size_name, $mime );
 				if ( is_array( $source ) ) {
-					$sources[ $mime ] = $source;
+					$properties['sources'][ $mime ]  = $source;
+					$metadata['sizes'][ $size_name ] = $properties;
+					wp_update_attachment_metadata( $attachment_id, $metadata );
 				}
 			}
```
felixarntz: Nit-pick, but code tends to be easier to read when avoiding excessive nesting. For readability I'd suggest changing this around, so that the check in the beginning is the opposite:
```php
if ( ! empty( $properties['sources'][ $mime ] ) ) {
	continue;
}
```

Then the other code below doesn't need to be nested/indented. The same applies to the `is_array` check, which I furthermore think we should change to use `is_wp_error`, which is more clear. We don't need to check whether the result is an array then, because we know it will always be either a `WP_Error` or an array. So here I'd suggest to do the following:
```php
if ( is_wp_error( $source ) ) {
	continue;
}
```

This way we have even less nesting/indentation.

modules/images/webp-uploads/load.php
```diff
@@ -285,3 +288,47 @@ function webp_uploads_remove_sources_files( $attachment_id ) {
...
 
 add_action( 'delete_attachment', 'webp_uploads_remove_sources_files', 10, 1 );
+
+/**
+ * Filter on `wp_get_missing_image_subsizes` acting as an action for the logic of the plugin
+ * to determine if additional mime types still need to be created.
+ *
+ * @since n.e.x.t
+ *
+ * @see   wp_get_missing_image_subsizes()
```
```suggestion
 * @see wp_get_missing_image_subsizes()
```

modules/images/webp-uploads/load.php
```diff
@@ -285,3 +288,47 @@ function webp_uploads_remove_sources_files( $attachment_id ) {
...
+ *
+ * @param array[] $missing_sizes Associative array of arrays of image sub-size.
+ * @param array   $image_meta    The metadata from the image.
+ * @param int     $attachment_id The ID of the attachment.
+ *
+ * @return array[] $missing_sizes Associative array of arrays of image sub-size.
```
felixarntz: Return docs shouldn't have any variable name, and there should be no blank line in between `@param` and `@return` per WP docs standards. Also, `array[]` isn't a valid type, WP core tends to always use just `array`. If anything, it could be `[]array` which would be the correct version for "array of arrays".
```suggestion
 * @param array   $missing_sizes Associative array of arrays of image sub-sizes.
 * @param array   $image_meta    The metadata from the image.
 * @param int     $attachment_id The ID of the attachment.
 * @return array Associative array of arrays of image sub-sizes.
```

modules/images/webp-uploads/load.php
```diff
@@ -285,3 +288,47 @@ function webp_uploads_remove_sources_files( $attachment_id ) {
...
+	$trace = array();
+	// Only setup the trace array if we no longer have more sizes.
+	if ( empty( $missing_sizes ) ) {
+		/**
+		 * The usage of `debug_backtrace` in this particular case is mainly to ensure the call to this
+		 * filter was originated from `wp_update_image_subsizes()` due this function will call
+		 * `wp_get_missing_image_subsizes` at some point but this function can be called from other places
+		 * as well without having to pass from `wp_update_image_subsizes` in an ideal world an action would
+		 * exist in `wp_update_image_subsizes`, in the meantime, this is a workaround for that scenario. The
+		 * limit to 10 is to have a buffer as in an ideal scenario the function would be index 5 on the array.
```
felixarntz: Nit-pick, trying to clarify this comment a bit:
```suggestion
		 * The usage of `debug_backtrace` in this particular case is mainly to ensure the call to
		 * `wp_get_missing_image_subsizes()` originated from `wp_update_image_subsizes()`, since only then the
		 * additional image sizes should be generated. `wp_get_missing_image_subsizes()` could also be called
		 * from other places in which case the custom logic should not trigger. In an ideal world an action
		 * would exist in `wp_update_image_subsizes` that runs any time, but the current
		 * `wp_generate_attachment_metadata` filter is skipped when all core sub-sizes have been generated.
		 * An eventual core implementation will not require this workaround. The limit of 10 is used to allow
		 * for some flexibility. While by default the function would be on index 5, other custom code may
		 * cause the index to be slightly higher.

File: modules/images/webp-uploads/load.php (hunk 11 above)
Comment (felixarntz) [reply modules/images/webp-uploads/load.php]: 10 is used to allow for some flexibility, since other plugins may cause the function to be at a slightly higher index than the expected 5 in the stack. E.g. if you use a profiling plugin that hooks into every hook, the indexes here may shift.

+1 on explaining this in a detailed comment. I also suggested a few iterations on that comment in my review.

File: modules/images/webp-uploads/load.php [hunk 16]
```diff
@@ -285,3 +288,47 @@ function webp_uploads_remove_sources_files( $attachment_id ) {
...
+ * @param array[] $missing_sizes Associative array of arrays of image sub-size.
+ * @param array   $image_meta    The metadata from the image.
+ * @param int     $attachment_id The ID of the attachment.
+ *
+ * @return array[] $missing_sizes Associative array of arrays of image sub-size.
+ */
+function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta, $attachment_id ) {
+	$trace = array();
+	// Only setup the trace array if we no longer have more sizes.
+	if ( empty( $missing_sizes ) ) {
```
felixarntz: No need to define the `$trace` variable here, it's only needed when we also use the `debug_backtrace()` workaround. Let's also reverse the condition here (similar to my other comments above) to use an early return instead, to avoid the nesting, which makes the code easier to read.
```php
if ( ! empty( $missing_sizes ) ) {
	return $missing_sizes;
}
// Only run all custom logic here, without nesting.
```

felixarntz (commented):
@mitogh This is pretty much good to go, all my comments are about documentation and for a few details worth improving IMO.

felixarntz:
@adamsilverstein 
> The media editor on the other hand shows this sad grey box with a "progress" bar that stays full all throughout the regeneration process (I think it shows upload percent as the file uploads, but then remains full until subsize gerneration completes):
> image

That's a great catch! I think it's definitely okay for now, but especially for the WP core patch we should investigate a solution. I'm curious how the individual "steps" of the progress bar are being determined in core today. Right now it makes total sense that it sticks around at 100% until our process is completed, since basically our logic happens at the end, and the client-side has no idea that these additional sizes are being generated, so I assume it doesn't take them into account when "computing" the progress bar steps.

felixarntz (approved):
@mitogh Awesome stuff! 🎉

adamsilverstein:
> That's a great catch! I think it's definitely okay for now, but especially for the WP core patch we should investigate a solution. I'm curious how the individual "steps" of the progress bar are being determined in core today. Right now it makes total sense that it sticks around at 100% until our process is completed, since basically our logic happens at the end, and the client-side has no idea that these additional sizes are being generated, so I assume it doesn't take them into account when "computing" the progress bar steps.

@felixarntz 

I'm pretty sure the progress bar is showing the upload status & progress, then shows full while image processing competes, finally the thumbnail is shown on completion.

in my local testing without the plugin, it behaves the same way, the progress bar goes to "full" but the box still shows gray until all of the image processing is complete ([screencast](https://share.getcloudapp.com/kpu4Xo5P))

Dragging multiple files, the "progress" bar is partially full for each image as the processing and uploading happens for each image ([screencast](https://share.getcloudapp.com/L1uW0PBE)). Also, each upload causes the async upload callback to be called (so essentially images are uploaded one at a time)

adamsilverstein:
testing on a remote site I can see the progress bar at work, showing the upload. then it stays 100% while the processing happens ([screencast](https://share.getcloudapp.com/DOumbN42)).

PR #191, opened 2022-02-28
Enable all non-experimental modules on plugin activation

### LINKED ISSUE #61: Define default module activation (onboarding)
Opening this issue to discuss if modules should be activated by default or not, maybe all modules other than the experimental for instance. When someone install the plugin, the expectation is generally that there is some preset for features right away.

If it is justified not to have any modules activated, we should at least guide plugin users to the module activation page, for instance redirecting to settings page upon plugin activation with a notice informing them to activate modules (and/or wp tooltip maybe).

kirtangajjar:
## Summary

Fixes #61. Enable all non-experimental modules on plugin activation.

Fix is quiet simple so there isn't much to explain here.

### DIFF
--- a/admin/load.php
+++ b/admin/load.php
@@ -344,7 +344,7 @@ function perflab_get_module_data( $module_file ) {
  */
 function perflab_activation_hook() {
 	// Bail if option is already set with any value.
-	if ( false !== get_option( PERFLAB_MODULES_SETTING, false ) ) {
+	if ( get_option( PERFLAB_MODULES_SETTING, false ) ) {
 		return;
 	}
 
--- a/admin/load.php
+++ b/admin/load.php
@@ -336,3 +336,23 @@ function perflab_get_module_data( $module_file ) {
 
 	return $module_data;
 }
+
+/**
+ * Enable all non-experimental modules on plugin activation.
+ *
+ * @since 1.0.0
+ */
+function perflab_activation_hook() {
+	$modules          = perflab_get_modules();
+	$modules_settings = perflab_get_module_settings();
+
+	foreach ( $modules as $module_name => $module_data ) {
+		if ( ! $module_data['experimental'] ) {
+			$modules_settings[ $module_name ]['enabled'] = true;
+		}
+	}
+
+	update_option( PERFLAB_MODULES_SETTING, $modules_settings );
+}
+
+register_activation_hook( __FILE__, 'perflab_activation_hook' );
--- a/admin/load.php
+++ b/admin/load.php
@@ -344,7 +344,7 @@ function perflab_get_module_data( $module_file ) {
  */
 function perflab_activation_hook() {
 	// Bail if option is already set with any value.
-	if ( get_option( PERFLAB_MODULES_SETTING, false ) ) {
+	if ( false !== get_option( PERFLAB_MODULES_SETTING, false ) ) {
 		return;
 	}
 
--- a/admin/load.php
+++ b/admin/load.php
@@ -343,6 +343,11 @@ function perflab_get_module_data( $module_file ) {
  * @since 1.0.0
  */
 function perflab_activation_hook() {
+	// Bail if option is already set with any value.
+	if ( false !== get_option( PERFLAB_MODULES_SETTING, false ) ) {
+		return;
+	}
+
 	$modules          = perflab_get_modules();
 	$modules_settings = perflab_get_module_settings();
 
--- a/admin/load.php
+++ b/admin/load.php
@@ -348,11 +348,11 @@ function perflab_activation_hook() {
 
 	foreach ( $modules as $module_name => $module_data ) {
 		if ( ! $module_data['experimental'] ) {
-			$modules_settings[ $module_name ]['enabled'] = true;
+			$modules_settings[ $module_name ] = array( 'enabled' => true );
 		}
 	}
 
 	update_option( PERFLAB_MODULES_SETTING, $modules_settings );
 }
 
-register_activation_hook( __FILE__, 'perflab_activation_hook' );
+register_activation_hook( dirname( __DIR__ ) . '/load.php', 'perflab_activation_hook' );


kirtangajjar:
@eclarke1 Can you please add "Needs Review" and other relevant labels as well?

admin/load.php
```diff
@@ -336,3 +336,23 @@ function perflab_get_module_data( $module_file ) {
...
+	foreach ( $modules as $module_name => $module_data ) {
+		if ( ! $module_data['experimental'] ) {
+			$modules_settings[ $module_name ]['enabled'] = true;
+		}
+	}
+
+	update_option( PERFLAB_MODULES_SETTING, $modules_settings );
+}
+
+register_activation_hook( __FILE__, 'perflab_activation_hook' );
```
felixarntz: I believe this needs to not use this file (`admin/load.php`), but the plugin main file (`load.php`).
```suggestion
register_activation_hook( dirname( __DIR__ ) . '/load.php', 'perflab_activation_hook' );
```

admin/load.php
```diff
@@ -336,3 +336,23 @@ function perflab_get_module_data( $module_file ) {
...
 }
+
+/**
+ * Enable all non-experimental modules on plugin activation.
+ *
+ * @since 1.0.0
+ */
+function perflab_activation_hook() {
+	$modules          = perflab_get_modules();
+	$modules_settings = perflab_get_module_settings();
```
felixarntz: One crucial thing missing here is that we should only set the option if it has no value yet. For example, if you already used the plugin and later reactivated it, it shouldn't modify any option value previously defined by the user.
```suggestion
function perflab_activation_hook() {
	// Bail if option is already set with any value.
	if ( false !== get_option( PERFLAB_MODULES_SETTING, false ) ) {
		return;
	}

	$modules          = perflab_get_modules();
	$modules_settings = perflab_get_module_settings();
```

admin/load.php
```diff
@@ -336,3 +336,23 @@ function perflab_get_module_data( $module_file ) {
...
+
+	foreach ( $modules as $module_name => $module_data ) {
+		if ( ! $module_data['experimental'] ) {
+			$modules_settings[ $module_name ]['enabled'] = true;
```
felixarntz: `$modules_settings[ $module_name ]` may not be set, we need to avoid potential PHP notices here. Since this should only run if the option is empty (see my other comment above), nothing will be set here, so we can do:
```suggestion
			$modules_settings[ $module_name ] = array( 'enabled' => true );
```

felixarntz (commented):
@kirtangajjar This looks like a solid start. I left a few comments and suggestions for parts that need iteration.

felixarntz (approved):
@kirtangajjar I've added those few requested changes and with that this looks good to me. Thank you for the PR! 🎉

admin/load.php
```diff
@@ -336,3 +336,28 @@ function perflab_get_module_data( $module_file ) {
...
+	foreach ( $modules as $module_name => $module_data ) {
+		if ( ! $module_data['experimental'] ) {
+			$modules_settings[ $module_name ] = array( 'enabled' => true );
+		}
+	}
+
+	update_option( PERFLAB_MODULES_SETTING, $modules_settings );
+}
+
+register_activation_hook( dirname( __DIR__ ) . '/load.php', 'perflab_activation_hook' );
```
mitogh, line 363: Would be great to add a deactivation hook, removing the option when the plugin is removed.

admin/load.php
```diff
@@ -336,3 +336,28 @@ function perflab_get_module_data( $module_file ) {
...
+/**
+ * Enable all non-experimental modules on plugin activation.
+ *
+ * @since 1.0.0
+ */
+function perflab_activation_hook() {
+	// Bail if option is already set with any value.
+	if ( false !== get_option( PERFLAB_MODULES_SETTING, false ) ) {
+		return;
+	}
```
mitogh, line 349:
```suggestion
	if ( get_option( PERFLAB_MODULES_SETTING, false ) ) {
		return;
	}
```

Looks like the value being stored is never false but it can be an empty array instead checking for `falsy` value should be enough

mitogh (commented):
Overall looks great, few minor comments.

admin/load.php
felixarntz: That may make sense, but should be a separate enhancement. This PR/issue is only to include the non-experimental modules as active by default.

admin/load.php
felixarntz: @mitogh The code suggestion here seems incorrect since now the code won't set the option if it's not set. Did you mean the reverse here?

The intention here is to check whether no value is set for the option. If the option is set to an empty array, we shouldn't do anything here. Only if the option isn't set at all, the logic here should be run. So in this case, getting an empty array from the database is different than getting an empty array from the option's default value.

mitogh: Good catch. I was intended to suggest the opposite, _let me update my suggestion._

kirtangajjar:
Thank you @felixarntz @mitogh for your reviews. I have committed the suggestion made by @mitogh. Please do let me know incase this needs something else from my end :)

felixarntz:
@kirtangajjar Sorry, see my comment in https://github.com/WordPress/performance/pull/191#discussion_r816339305, I'm not sure that change actually makes sense. From what I can tell it was better the way you originally had it:
* We want to only run our logic if the option is not set.
* But now after the change suggested by @mitogh the logic is also run if the option is set to e.g. an empty array.
* The latter is problematic since it would mean we may override the user-defined configuration just because they disabled all modules, which is unexpected UX.

admin/load.php
```diff
@@ -336,3 +336,28 @@ function perflab_get_module_data( $module_file ) {
...
+/**
+ * Enable all non-experimental modules on plugin activation.
+ *
+ * @since 1.0.0
+ */
+function perflab_activation_hook() {
+	// Bail if option is already set with any value.
+	if ( get_option( PERFLAB_MODULES_SETTING, false ) ) {
+		return;
+	}
```
felixarntz: I think this needs to be reverted to what it previously was: We should not modify the option if it's already set. Without the `false !==` bit, the code here would still modify the option if it was set to an empty array, for example if the user had deactivated all modules. In that case, they intentionally did that, so we shouldn't change it.

The goal here is to only set the option to all available (non-experimental) modules initially, e.g. when the plugin is _first_ activated and nothing is configured yet.
```suggestion
	// Bail if option is already set with any value.
	if ( false !== get_option( PERFLAB_MODULES_SETTING, false ) ) {
		return;
	}
```

felixarntz (changes requested):
@kirtangajjar One thing remaining here per the above.

cc @mitogh

kirtangajjar:
@felixarntz Fixed. Sorry didn't think that one through fully.

mitogh (approved):

felixarntz (approved):
Thank you @kirtangajjar for the PR - awesome work! 🙌

Issue #75, opened 2022-01-01
Ensure post meta reflects image mime type


adamsilverstein:
When uploading jpeg images with the `webp-uploads` module, we use WebP as the default file type for sub sized images. The post meta that lists the image sizes includes a mime-type value; we should ensure that matches the sub sized image mime type.

As reported in https://github.com/WordPress/performance/pull/32#issuecomment-992314408

mitogh:
Related with 

- #69
- #67

erikyo:
hi @adamsilverstein, 
[This ](https://gist.github.com/erikyo/f0aca6c34fd1d6bc5f73f953bb342a65)is an experiment, I did would try to solve these two issues (the other one is the #122), but after all I guess this definitively isn't in a great way because it adds an additional filter in order to work. Maybe it would be better to add a filter to media_handle_upload (wp-admin/includes/media.php) because without this, I can be wrong but as far as I know, the uploaded image cannot be touched before this hook.

I also don't understand why after this filter the "cloned" webp isn't deleted when I delete the image from WordPress media library, even if the attachment metadata array seems correct. I suspect this issue is related to wp_unique_filename() but not sure.

https://gist.github.com/erikyo/f0aca6c34fd1d6bc5f73f953bb342a65
(needs to be used alongside Module WebP Uploads)

adamsilverstein:
> I also don't understand why after this filter the "cloned" webp isn't deleted when I delete the image from WordPress media library, even if the attachment metadata array seems correct. I suspect this issue is related to wp_unique_filename() but not sure.

I think you need to add the new file data to the returned `new_sizes`? if you look at the meta data after uploading, does it reference both the jpeg and webp images, or just jpeg? or is your idea to add the webp type on the fly?

erikyo:
yes, my idea is to replace also the "full" size image on the fly with the copy in webp format (so replacing the $meta['file']), since the original jpg was already stored (or can be stored) as "original_image" and since I don't want to show jpg images when I `wp_get_attachment_image()`.

At the time the filter is fired, the image has already been saved (but the resizes are not and the `sizes` array is empty), my idea is to use this filter to hack the image_meta before the subsizes image were generated.  This to avoid https://github.com/WordPress/performance/issues/69 because sometimes (eg. small images without the -scaled) the image meta are like below:
```
Array
(
    [width] => 945
    [height] => 1007
    [file] => 2022/01/small-image.jpeg
    [sizes] => Array
    (...)

[...]

PR #194, opened 2022-02-28
Create additional MIME types for the full size image.

### LINKED ISSUE #174: When generating WebPs, ensure full size is generated(`-scaled` for large images)
## Bug Description

After #152 "Full"" sized images inserted by users will still use the original jpeg image or (`-scaled` version for larger images) because the WebP version is not being generated. See https://github.com/WordPress/performance/pull/152#issuecomment-1051118813.

Proposal to fix:
* Always generate a full sized WebP version when WebP output is active
* For larger images (over `big_image_size_threshold`), generate a `-scaled.webp` file

## Steps to reproduce

1. Enable webp-uploads
2. Upload and insert an image, select full size in the image properties sidebar
3. view the post and check the image source

#### Expected:
* All image references are to `.webp` files

#### Actual 
* some references are to a `.jpg` file.

## Screenshots

image
image
image

mitogh:
## Summary

Fixes #174 

## Relevant technical choices

The following PR introduces a `sources` property at the root of the metadata with all the information for every single mime type, in this case `file` and `filesize`.

The PR introduces a mechanism to remove the additional files created by the plugin

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -111,6 +111,102 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+ * Creates a `sources` property at the root of the metadata to store the information about
+ * the created full sized images. It creates an image with the size of the full image size.
+ *
+ * @since n.e.x.t
+ *
+ * @param array $metadata The current metadta of the modified image.
+ * @param int   $attachment_id The ID of the attachment being modified.
+ * @return array The metadata modified of the image.
+ */
+function webp_uploads_create_sources_property_for_full_image( $metadata, $attachment_id ) {
```
adamsilverstein: this creates the image and updates the sources meta data, right? function name doesn't indicate that, maybe `webp_uploads_create_full_image` or `webp_uploads_create_full_image_and_sources`?

modules/images/webp-uploads/load.php
```diff
@@ -111,6 +111,102 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+			'filesize' => isset( $image['path'] ) ? filesize( $image['path'] ) : 0,
+		);
+
+		wp_update_attachment_metadata( $attachment_id, $metadata );
+	}
+
+	return $metadata;
+}
+
+add_filter( 'wp_generate_attachment_metadata', 'webp_uploads_create_sources_property_for_full_image', 10, 2 );
```
adamsilverstein: Can we add to our existing callback rather than hooking again? that way our fix for constrained systems will work with the full size as well.

modules/images/webp-uploads/load.php
```diff
@@ -284,11 +378,47 @@ function webp_uploads_remove_sources_files( $attachment_id ) {
...
+	}
+
+	if ( ! isset( $metadata['sources'] ) || ! is_array( $metadata['sources'] ) ) {
+		return;
+	}
+
+	$original_mime_from_post = get_post_mime_type( $attachment_id );
+	$original_mime_from_file = wp_check_filetype( $file )['type'];
+
+	// Delete of full sizes mime types.
```
adamsilverstein: extra word?
```suggestion
	// Delete full sizes mime types.
```

modules/images/webp-uploads/load.php
mitogh: Sure, just wanted the avoid a large function but I see the benefit in keeping everything in a single place.

modules/images/webp-uploads/load.php
mitogh: :100: I think that's a good idea.

adamsilverstein:
🎉  This worked well in my testing, even with a resource constrained system where multiple retries were required.

modules/images/webp-uploads/load.php
```diff
@@ -11,42 +11,106 @@
 /**
  * Hook called by `wp_generate_attachment_metadata` to create the `sources` property for every image
  * size, the sources' property would create a new image size with all the mime types specified in
- * `webp_uploads_valid_image_mime_types`. If the original image is one of the mimes from
- * `webp_uploads_valid_image_mime_types` the image is just added to the `sources` property and  not
- * created again. If the uploaded attachment is not a valid image this function does not alter the
- * metadata of the attachment, on the other hand a `sources` property is added.
+ * `webp_uploads_get_supported_image_mime_transforms`. If the original image is one of the mimes from
+ * `webp_uploads_get_supported_image_mime_transforms` the image is just added to the `sources` property and  not
```
```suggestion
 * `webp_uploads_get_supported_image_mime_transforms` the image is just added to the `sources` property and not
```

adamsilverstein (approved):

modules/images/webp-uploads/load.php
```diff
@@ -11,42 +11,106 @@
...
+			continue;
+		}
+
+		$metadata['sources'][ $targeted_mime ] = array(
+			'file'     => $image['file'],
+			'filesize' => isset( $image['path'] ) ? filesize( $image['path'] ) : 0,
+		);
+
+		wp_update_attachment_metadata( $attachment_id, $metadata );
+	}
```
felixarntz: There is _a lot_ of duplicate code in this loop which is very similar to the same `foreach ( $formats ...)` loop further below for generating the additional formats for the sub-sizes. We should simplify this and reuse some common functionality. Specifically, we should rework the `webp_uploads_generate_image_size()` function so that it can be reused for both scenarios, both generating the full image as well as the sub-sized images.

I suggest the following:
* Introduce a new `webp_uploads_generate_additional_image_source( $attachment_id, $src_file, $dst_file, $mime, $size_data )`, where:
    * `$attachment_id` should be the attachment ID. That may not even be used by the function, but it should be specified for consistency with other functions and most importantly, to have that context available e.g. for potential filters introduced in the function later.
    * `$src_file` should be the full file name (including path) to the source file to generate the image from (i.e. the original image file). This should still be a parameter though to allow for flexibility and to control this from outside the function, which e.g. will also help with tests.
    * `$dst_file` should be the full file name (including path) where to save the generated file.
    * `$mime` should be the MIME type to generate the new image.
    * `$size_data` should be an associative array with `width`, `height` (both int), and `crop` (bool), as typically specified for WordPress image sizes).
* This function should mostly take over code that is currently present in `webp_uploads_generate_image_size()`, and it should return the same data, an associative array with `file` and `filesize` on success, or a `WP_Error` on failure.
* The function should receive an `@access private` annotation, to indicate that it shouldn't be used by external developers.
* The `webp_uploads_generate_image_size()` should be simplified to use the above new function, so it will become a very simple function that just looks up the correct `$src_file`, `$dst_file`, and `$size_data` and then passes those to the other new function.
* The above `foreach` loop` should then also be simplified so that it just calls the above new function as well. This will result in code like the following:
```suggestion
	foreach ( $valid_mime_transforms[ $mime_type ] as $targeted_mime ) {
		// If this property exists no need to create the image again.
		if ( ! empty( $metadata['sources'][ $targeted_mime ] ) ) {
			continue;
		}

		if ( ! isset( $allowed_mimes[ $targeted_mime ] ) || ! is_string( $allowed_mimes[ $targeted_mime ] ) ) {
			continue;
		}

		if ( ! isset( $metadata['width'], $metadata['height'] ) ) {
			continue;
		}

		$extension        = explode( '|', $allowed_mimes[ $targeted_mime ] );
		$extension        = $extension[0];
		$filename         = wp_basename( $file, ".{$original_extension}" );
		$target_file_name = trailingslashit( $original_directory ) . "{$filename}.{$extension}";

		$size_data = array(
			'width'  => $metadata['width'],
			'height' => $metadata['width'],
			'crop'   => false,
		);

		$source = webp_uploads_generate_additional_image_source( $attachment_id, $file, $target_file_name, $targeted_mime, $size_data );
		if ( is_wp_error( $source ) ) {
			continue;
		}

		$metadata['sources'][ $targeted_mime ] = $source;
		wp_update_attachment_metadata( $attachment_id, $metadata );
	}
```

With the above, we will drastically reduce the duplicate code here and make it easier to maintain.

modules/images/webp-uploads/load.php
```diff
@@ -11,42 +11,106 @@
...
+			continue;
+		}
+
+		if ( ! isset( $metadata['width'], $metadata['height'] ) ) {
+			continue;
+		}
+
+		$editor->resize( $metadata['width'], $metadata['height'], false );
+		$extension = explode( '|', $allowed_mimes[ $targeted_mime ] );
+		$extension = reset( $extension );
```
felixarntz: It would be more straightforward to do the following:
```suggestion
		$extension = $extension[0];
```

modules/images/webp-uploads/load.php
```diff
@@ -11,42 +11,106 @@
...
+
+		if ( empty( $extension ) || empty( $original_extension ) ) {
+			continue;
+		}
```
felixarntz: This is overkill and should be removed. While we should code defensively, we shouldn't need to check for conditions over-aggressively in cases where it's basically impossible for them to happen: `explode( '|', $allowed_mimes[ $targeted_mime ] )` will always be a non-empty array with one or more extensions, so the check for `empty( $extension )` here is unnecessary.

The check for `empty( $original_extension )` arguably makes a bit more sense, but since that value is defined outside of the `foreach` loop, it should be checked before even getting to this loop.

modules/images/webp-uploads/load.php
```diff
@@ -284,11 +346,47 @@ function webp_uploads_remove_sources_files( $attachment_id ) {
...
+
+	$original_mime_from_post = get_post_mime_type( $attachment_id );
+	$original_mime_from_file = wp_check_filetype( $file )['type'];
+
+	// Delete full sizes mime types.
+	foreach ( $metadata['sources'] as $mime => $properties ) {
+		// Don't remove the image with the same mime type as the original image as this would be removed by WordPress.
+		if ( $mime === $original_mime_from_post || $mime === $original_mime_from_file ) {
+			continue;
+		}
```
felixarntz, line 395: This is overkill, checking the MIME type from the post will be sufficient. In the existing code for the sub-sizes we do the same thing, only checking the metadata, but not the file itself.
```suggestion
	$original_mime = get_post_mime_type( $attachment_id );

	// Delete full sizes mime types.
	foreach ( $metadata['sources'] as $mime => $properties ) {
		// Don't remove the image with the same mime type as the original image as this would be removed by WordPress.
		if ( $mime === $original_mime ) {
			continue;
		}
```

felixarntz (commented):
@mitogh The implementation functionally looks solid to me, however there is a lot of duplicate code around the image generation, that we should reduce by introducing a low-level function with the common parts.

modules/images/webp-uploads/load.php
mitogh: That's a fair point, in the subsizes is determined during the upload however exists a use case when the mime type of the file and the `post->mime_type` won't be the same, specifically when using the filter that exists en core and the images are uploaded as WebP the `post->mime_type` would be set as JPEG when the actual mime type is WebP if that's the case duplicated WebP images would be generated.

Details can be found here: https://github.com/WordPress/performance/issues/75

This is not the same for the subsizes as at the time the mime property for the subsizes is set we can rely in the fact that is the correct one as it creates the mime after the file was uploaded as WebP. Agree this is an edge case but worth consider it as this an edge case to consider.

modules/images/webp-uploads/load.php
mitogh: Agree with that, originally this was a separate filter only to handle the `full size` but in order to keep things in a single place was merged into a single method.

---- 

This is the rationale behind not creating a new function or why not introducing additional parameters into the existing function.

The current logic on the method handles images sizes, and because the `full size` is a virtual size we can as you pointed out separate any depending logic from the sizes as parameters.

- The main problem with that was the fact that makes things more complex due number of parameters.
- The current logic to handle the sizes removes the suffix `scaled` or `rotated` if present, we can add and if block here to prevent this from happening where the image size is `full`. This won't be required anymore after #204 due the `scaled` was coming because we were using the `scaled` version in some cases, so that condition and replacement seem unnecessary after that PR is merged.

Possible problems when using the same function with additional parameters/conditionals inside: 

- Complexity increases due to the number of parameters
- Knowledge of the outside world by working differently in some cases.
- Problem when "full" size is defined by the user.

Agree that we can go deeper in breaking it down into smaller functions, but due we are using a global namespace this seems overkill, Lastly, I would argue that having duplicate code is better than having the wrong abstraction in place.

Happy to go down the route for more abstract modules but from my perspective it seems overkill for an additional image size, which effectively is not a size so that's why the abstraction does not fit.

felixarntz: @mitogh I'm not sure I understand some of your points. There may be a misunderstanding related to my original comment.

> originally this was a separate filter

I'm not talking about a separate filter, the way you have it all in one function makes sense. I'm just saying we should introduce another low-level function to remove duplicated logic in the two different `foreach` loops which is similar for creating an image based on another image. That part is the same no matter if it's for the original image or for a sub-sized image.

> The current logic on the method handles images sizes

I'm not suggesting to change the parameters of that existing function, I'm suggesting to introduce a new _low-level_ function that can be used in both relevant plaes.

> The current logic to handle the sizes removes the suffix `scaled` or `rotated` if present, we can add and if block here to prevent this from happening where the image size is `full`.

With my suggestion from above nothing would change here. That logic would be kept in the `webp_uploads_generate_image_size()` function just like it is today. Again, I'm talking about a new function, not expanding the behavior of the existing `webp_uploads_generate_image_size()` function.

> Possible problems when using the same function with additional parameters/conditionals inside:

See above, I'm not saying we should use the same function with additional parameters.

> Agree that we can go deeper in breaking it down into smaller functions, but due we are using a global namespace this seems overkill, Lastly, I would argue that having duplicate code is better than having the wrong abstraction in place.

What about my suggested abstraction is wrong in your opinion, can you clarify? My proposed new function `webp_uploads_generate_additional_image_source()` has nothing to do with sub-sized images or the original image, it is 100% agnostic of that. It has a single clear purpose: Create an image based on a given source file in a given MIME type for given size arguments. It does nothing else than that, and it won't have any if clauses that would be different depending on whether it's a sub-size or the original image (or the `full` image). All the dynamic parts relating to that would still be handled outside of that function, because they are different, I agree with you. The common behavior that should be put into its own function to remove duplicate code is the context-unaware creation of a new image based on whatever parameters you pass into it.

modules/images/webp-uploads/load.php
felixarntz: Thanks for clarifying, that makes sense!

felixarntz (commented):
@mitogh Left another reply on https://github.com/WordPress/performance/pull/194#discussion_r821259733, I think that point is still valid.

The other changes lgtm, however note that linting is currently failing, so something will need to be fixed there.

modules/images/webp-uploads/load.php
mitogh: > I'm not suggesting to change the parameters of that existing function, I'm suggesting to introduce a new low-level function that can be used in both relevant places.

Mostly wanted to avoid adding new global functions but I do see the benefit in this approach to simplify and cleanup duplication, so let me go with this route instead.

> With my suggestion from above nothing would change here. That logic would be kept in the webp_uploads_generate_image_size() function just like it is today. Again, I'm talking about a new function, not expanding the behavior of the existing webp_uploads_generate_image_size() function.

:+1: 

> What about my suggested abstraction is wrong in your opinion, can you clarify? My proposed new function webp_uploads_generate_additional_image_source() has nothing to do with sub-sized images or the original image, it is 100% agnostic of that.

I think this one is on me as I confused the new function `webp_uploads_generate_additional_image_source` with `webp_uploads_generate_image_size` instead so I agree that this function makes more sense to keep it private. One thing to add is besides marking the function as private we also should follow the convention to prefix such functions with a dash like `_webp_uploads_generate_image_size` instead like: `_wp_admin_bar_init` or `_get_cron_lock`

felixarntz: Thanks for clarifying, I had the impression this was a misunderstanding :)

> One thing to add is besides marking the function as private we also should follow the convention to prefix such functions with a dash

Note that this convention is not standardized in WP core. I have had multiple WP core committers tell me over the years that we shouldn't use the underscore prefix anymore, which makes sense to me. The `@private` is sufficient as an indicator, the underscore prefix doesn't _really_ mean anything. So I'd advise to leave the function name without underscore prefix.

mitogh: Thanks, just applied this feedback let me know in case you have any additional feedback on this one.

modules/images/webp-uploads/load.php
```diff
@@ -54,51 +55,40 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+		'crop'   => false,
+	);
+
 	$original_directory = pathinfo( $file, PATHINFO_DIRNAME );
-	$original_extension = pathinfo( $file, PATHINFO_EXTENSION );
+	if ( isset( $metadata['file'] ) ) {
+		$filename = pathinfo( $metadata['file'], PATHINFO_FILENAME );
+	} else {
+		$filename = pathinfo( $file, PATHINFO_FILENAME );
+	}
```
felixarntz: This condition is new here. Can you clarify why you added it? Isn't looking at the attached file in `$file` sufficient, like it was before?

modules/images/webp-uploads/load.php
```diff
@@ -286,6 +238,68 @@ function webp_uploads_get_supported_image_mime_transforms() {
...
+ *
+ * @since n.e.xt
+ * @param int         $attachment_id The ID of the attachment from where this image would be created.
+ * @param array       $size_data An array with the dimensions of the image: height, width and crop.
+ * @param string      $mime The target mime in which the image should be created.
+ * @param string|null $destination_file_name The path where the file would be stored, it should include the extension, when null would use `generate_filename` to create a destination.
+ * @return array|WP_Error An array with the file and filesize if the image was created correctly otherwise a WP_Error
+ */
+function webp_uploads_generate_additional_image_source( $attachment_id, array $size_data = array(), $mime = '', $destination_file_name = null ) {
+	$image_path = wp_get_original_image_path( $attachment_id );
```
felixarntz: Can you please revert this back to the attached image? I understand we want to eventually use `wp_get_original_image_path()`, but that is part of #204. We shouldn't mix up responsibilities between PRs. After the PR here is merged, we can update the PR for #204.

modules/images/webp-uploads/load.php
```diff
@@ -286,6 +238,68 @@ function webp_uploads_get_supported_image_mime_transforms() {
...
+ * a WP_Error would be returned otherwise an array with the file and filesize properties.
+ *
+ * @since n.e.xt
+ * @param int         $attachment_id The ID of the attachment from where this image would be created.
+ * @param array       $size_data An array with the dimensions of the image: height, width and crop.
+ * @param string      $mime The target mime in which the image should be created.
+ * @param string|null $destination_file_name The path where the file would be stored, it should include the extension, when null would use `generate_filename` to create a destination.
+ * @return array|WP_Error An array with the file and filesize if the image was created correctly otherwise a WP_Error
+ */
+function webp_uploads_generate_additional_image_source( $attachment_id, array $size_data = array(), $mime = '', $destination_file_name = null ) {
```
felixarntz: Some of these parameters shouldn't have defaults but be required instead, for the following reasons:
* Not providing any value for `$size_data` is pointless, it would result in a 0x0 image, so this parameter needs to be required one way or the other.
* Same for `$mime`, an empty string makes no sense here, so this also needs to be a required parameter.
* The `$destination_file_name` is different, you're providing a reasonable default below, so that works. Can you please change the default to be an empty string though? Better to work in a single type for a variable unless it's impossible - an empty string works better as a default for a string value.

modules/images/webp-uploads/load.php
```diff
@@ -286,6 +238,68 @@ function webp_uploads_get_supported_image_mime_transforms() {
...
+/**
+ * Creates a resized image with the provided dimensions out of an original attachment, the created image
+ * would be saved in the specified mime and stored in the destination file. If the image can't be saved correctly
+ * a WP_Error would be returned otherwise an array with the file and filesize properties.
+ *
+ * @since n.e.xt
+ * @param int         $attachment_id The ID of the attachment from where this image would be created.
+ * @param array       $size_data An array with the dimensions of the image: height, width and crop.
+ * @param string      $mime The target mime in which the image should be created.
+ * @param string|null $destination_file_name The path where the file would be stored, it should include the extension, when null would use `generate_filename` to create a destination.
```
felixarntz: See my other comment, let's make this always a string. The default should be an empty string. Please change that bit in the parameter description as well.

modules/images/webp-uploads/load.php
```diff
@@ -286,6 +238,68 @@ function webp_uploads_get_supported_image_mime_transforms() {
...
+		return new WP_Error( 'image_mime_type_not_supported', __( 'The provided mime type is not supported.', 'performance-lab' ) );
+	}
+
+	$height = isset( $size_data['height'] ) ? (int) $size_data['height'] : 0;
+	$width  = isset( $size_data['width'] ) ? (int) $size_data['width'] : 0;
+	$crop   = isset( $size_data['crop'] ) && $size_data['crop'];
+
+	$editor->resize( $width, $height, $crop );
+
+	if ( null === $destination_file_name ) {
```
felixarntz, line 293: See above, this should then become a `empty( $destination_file_name )` check.

felixarntz (commented):
Left a bit more feedback here. Mostly lgtm, however there are a few inconsistencies. Most importantly, we shouldn't get this PR mixed up with what the separate issue #204 is intended for.

modules/images/webp-uploads/load.php
```diff
@@ -227,6 +238,69 @@ function webp_uploads_get_supported_image_mime_transforms() {
...
+ * Creates a resized image with the provided dimensions out of an original attachment, the created image
+ * would be saved in the specified mime and stored in the destination file. If the image can't be saved correctly
+ * a WP_Error would be returned otherwise an array with the file and filesize properties.
+ *
+ * @since n.e.xt
+ * @private
+ * @param int         $attachment_id The ID of the attachment from where this image would be created.
+ * @param array       $size_data An array with the dimensions of the image: height, width and crop.
+ * @param string      $mime The target mime in which the image should be created.
+ * @param string|null $destination_file_name The path where the file would be stored, it should include the extension, when null would use `generate_filename` to create a destination.
```
felixarntz: According to WP docs standards:
```suggestion
 * @since n.e.xt
 * @private
 *
 * @param int    $attachment_id         The ID of the attachment from where this image would be created.
 * @param array  $size_data             An array with the dimensions of the image: height, width and crop.
 * @param string $mime                  The target mime in which the image should be created.
 * @param string $destination_file_name The path where the file would be stored, including the extension. If empty, `generate_filename` is used to create the destination file name.
```

The change above also includes my other suggestion to always use a `string` for the last parameter.

modules/images/webp-uploads/load.php
mitogh: Sure, this condition simplifies a section of code that was used before to create the `filename`. The condition ensures that the `file` exists otherwise it fallbacks to the full `$file` path.

This was before:

```diff
---	$original_extension = pathinfo( $file, PATHINFO_EXTENSION );
---	$filename         = wp_basename( $file, ".{$original_extension}" );
+++	if ( isset( $metadata['file'] ) ) {
+++		$filename = pathinfo( $metadata['file'], PATHINFO_FILENAME );
+++	} else {
+++		$filename = pathinfo( $file, PATHINFO_FILENAME );
+++	}
 
```

However, both lines can be simplified by just using `PATHINFO_FILENAME` from the `file` if present otherwise tries with `$file`

modules/images/webp-uploads/load.php
mitogh: Good point, let me do that :+1:

modules/images/webp-uploads/load.php
mitogh: > Some of these parameters shouldn't have defaults but be required instead, for the following reasons:

Good point, in some cases the default makes sense, so let me expand on the reasoning. 

> Not providing any value for $size_data is pointless, it would result in a 0x0 image, so this parameter needs to be required one way or the other.

In some cases, we do need to allow for images with a width of `0` or height of `0` specifically for wide or large images but agree we need to make sure this parameter is required. 

> Same for $mime, an empty string makes no sense here, so this also needs to be a required parameter.

:100: 

> The $destination_file_name is different, you're providing a reasonable default below, so that works. Can you please change the default to be an empty string though? Better to work in a single type for a variable unless it's impossible - an empty string works better as a default for a string value.

Agree on the consistency of type, however, WordPress core uses a similar approach using `null` for string values as defaults example: 

-> https://github.com/WordPress/wordpress-develop/blob/ea1fe31f028156a56eb97e8b7cce3d0f0e6e242f/src/wp-includes/class-wp-image-editor.php#L427

The other benefit of using `null` is identify between an actual default value and a desired value, for example `$destination_file_name` if using an empty string as a default value, there's no way for an external caller to set the `$destination_file_name` value to empty string instead when using `null` we can identify that a default behavior is expected and allows the door for empty string.

I do see the fact that `$editor->save( '', $mime );` might not be a wide use case but having a `null` allows for that flexibility.

Let me know what do you think about that, will update the remaining pieces as you suggested, thanks.

modules/images/webp-uploads/load.php
felixarntz: @mitogh Hmm, the previous version was not using an `if` conditional, so it was actually simpler than how you changed it now.

My original question though was more related to why did you even add `$metadata['file']` here? The code before didn't pay attention to `$metadata['file']` and instead only relied on the attached file `$file`, so why did you add this? The question is less about the code complexity, but more about what the reasoning was to even consider `$metadata['file']` when it previously relied on `$file` only. Hope that makes sense.

modules/images/webp-uploads/load.php
felixarntz: That's a fair point, aligning with this closely related method makes sense. 👍

modules/images/webp-uploads/load.php
```diff
@@ -11,42 +11,92 @@
...
-	if ( empty( $metadata['sizes'] ) || ! is_array( $metadata['sizes'] ) ) {
-		return $metadata;
-	}
 	// This should take place only on the JPEG image.
 	$valid_mime_transforms = webp_uploads_get_supported_image_mime_transforms();
-
 	// Not a supported mime type to create the sources property.
 	$mime_type = get_post_mime_type( $attachment_id );
-	if ( ! isset( $valid_mime_transforms[ $mime_type ] ) ) {
+	if ( ! isset( $valid_mime_transforms[ $mime_type ] ) || ! is_array( $valid_mime_transforms[ $mime_type ] ) ) {
```
felixarntz: The `! is_array` check is a bit over-aggressive to add. If it wasn't an array that would be invalid, and in that case the developer using the filter would have clearly done something wrong. If you prefer to validate this, it should happen in a centralized way in `webp_uploads_get_supported_image_mime_transforms()`, so that it doesn't need to be checked after every call to the function.

modules/images/webp-uploads/load.php
felixarntz: Another related question: Why do you use this condition to set the `$filename` but not the `$original_directory`? It seems inconsistent to use one file for one thing and potentially another file for the other.

modules/images/webp-uploads/load.php
```diff
@@ -11,42 +11,92 @@
...
+		$metadata['sources'][ $mime_type ] = array(
+			'file'     => wp_basename( $file ),
+			'filesize' => filesize( $file ),
+		);
+		wp_update_attachment_metadata( $attachment_id, $metadata );
+	}
+
+	$original_size_data = array(
+		'width'  => isset( $metadata['width'] ) ? (int) $metadata['width'] : 0,
+		'height' => isset( $metadata['height'] ) ? (int) $metadata['height'] : 0,
```
felixarntz, line 60: If `$metadata['width']` or `$metadata['height']` are not set, this entire block of code wouldn't really work, as dimensions of 0x0 don't make sense. I think it would make more sense to wrap the entire block of code to generate additional formats for the full-size image into a `isset( $metadata['width'] ) && isset( $metadata['height'] )` conditional instead, so that the rest of the dependant code isn't run if one of these is not set.

felixarntz (commented):
@mitogh A few additional nit-picks, but it's close. My main open question is still https://github.com/WordPress/performance/pull/194/files#r823007963 though, that would be great to clarify.

modules/images/webp-uploads/load.php
mitogh: > Hmm, the previous version was not using an if conditional, so it was actually simpler than how you changed it now.

Correct due now I'm accessing `file` property I first make sure this property exists. 

> My original question though was more related to why did you even add $metadata['file'] here? The code before didn't pay attention to $metadata['file'] and instead only relied on the attached file $file, so why did you add this? T

The problem was due `file` was using the original file, so `$file` and `file` property where not quite the same, looking into this it does not makes sense to keep using the property as `$file` would be always the attached file instead.

mitogh: > Another related question: Why do you use this condition to set the $filename but not the $original_directory? It seems inconsistent to use one file for one thing and potentially another file for the other.

Mostly because the directory can be determined by just using the file, after a closer look at this, the conditional is not required and we can keep on using the `$file` variable instead, thanks for flagging this one.

modules/images/webp-uploads/load.php
mitogh: Great point, let me add a condition to prevent moving forward if both dimensions are not defined.

modules/images/webp-uploads/load.php
mitogh: > The ! is_array check is a bit over-aggressive to add. If it wasn't an array that would be invalid, and in that case the developer using the filter would have clearly done something wrong. 

Right, however, if we let them break it, the `foreach` loop won't be executed and triggered an error we can let the error happen in that case or just ignore and not process any invalid value to the filter. I'm more inclined towards the second but let me know if you think otherwise.

> If you prefer to validate this, it should happen in a centralized way in webp_uploads_get_supported_image_mime_transforms(), so that it doesn't need to be checked after every call to the function.

We do the validation in order to allow running in this function as well specifically here:

- https://github.com/WordPress/performance/blob/223137d6190de58762ec49dba720a4737a44bb99/modules/images/webp-uploads/load.php#L72


However I do see an inconsistency over here: 

- https://github.com/WordPress/performance/blob/223137d6190de58762ec49dba720a4737a44bb99/modules/images/webp-uploads/load.php#L138

Where we are assuming the value is an `array` if is defined, so I agree we should standardized that condition, still I'm more inclined towards making sure the value is a valid array but let me know if you think otherwise.

felixarntz: > Right, however, if we let them break it, the `foreach` loop won't be executed and triggered an error we can let the error happen in that case or just ignore and not process any invalid value to the filter. I'm more inclined towards the second but let me know if you think otherwise.

I agree we should ignore invalid values and not cause an error. I'd say that should happen by sanitizing the value right after the filter though, within `webp_uploads_get_supported_image_mime_transforms()`. That ensures the return value of the function is reliable, without requiring any calling code to check itself. In other words, `webp_uploads_get_supported_image_mime_transforms()` should strip any array keys where the value is `! is_array`.

modules/images/webp-uploads/load.php
```diff
@@ -160,33 +271,22 @@ function webp_uploads_generate_image_size( $attachment_id, $size, $mime ) {
...
-	$height = null;
-	$crop   = false;
-
-	if ( isset( $metadata['sizes'][ $size ]['width'] ) ) {
-		$width = $metadata['sizes'][ $size ]['width'];
-	} elseif ( isset( $sizes[ $size ]['width'] ) ) {
-		$width = $sizes[ $size ]['width'];
+	if ( $width <= 0 && $height <= 0 ) {
+		return new WP_Error( 'image_wrong_dimensions', __( 'At least one of the dimensions must be a positive number.', 'performance-lab' ) );
 	}
```
felixarntz, line 289: I think this is not fully correct, both dimensions must be positive numbers, not just one. Something like `300x0` or `0x300` wouldn't make sense either.
```suggestion
	if ( $width <= 0 || $height <= 0 ) {
		return new WP_Error( 'image_wrong_dimensions', __( 'Both image dimensions must be positive numbers.', 'performance-lab' ) );
	}
```

felixarntz (commented):
@mitogh Just one more comment based on the newly added code, and I also left a reply to https://github.com/WordPress/performance/pull/194#discussion_r823169674 where I think we still should centralize the validation instead of checking outside of the filter function.

modules/images/webp-uploads/load.php
mitogh: Actually having an image with only one dimension is a valid image size like: 

-`300x0`

Specifically, this is used in cases when you want to preserve the aspect ratio of an image by just setting the width in most cases so the height would be 0 or can be specified as 0.

https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/media.php#L535

modules/images/webp-uploads/load.php
mitogh: > I agree we should ignore invalid values and not cause an error. I'd say that should happen by sanitizing the value right after the filter though, within webp_uploads_get_supported_image_mime_transforms(). That ensures the return value of the function is reliable, without requiring any calling code to check itself. In other words, webp_uploads_get_supported_image_mime_transforms() should strip any array keys where the value is ! is_array.

Great point, let me adjust that one.

modules/images/webp-uploads/load.php
felixarntz: Ahh I wasn't aware that works, I thought it would try to make it actually 300x0 then. In that case, all good!

felixarntz (approved):
Great!

PR #195, opened 2022-02-28
Replace the full size image in `the_content` with additional MIME type if available

### LINKED ISSUE #174: When generating WebPs, ensure full size is generated(`-scaled` for large images)

mitogh:
## Summary

Fixes #174

This PR requires:

- #152 
- #194 

## Relevant technical choices

This PR updates the changes introduced by #152 and #194 in order to replace the `full` size image if present in the new format, (in this case WebP) making sure all images served to the client are in the same format when a new image is uploaded into WordPress. 

## Checklist


adamsilverstein (approved):
Excellent!

modules/images/webp-uploads/load.php
```diff
@@ -529,17 +529,18 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
 
 	$basename = wp_basename( $metadata['file'] );
```
felixarntz, line 545: I'm realizing now that the `isset` check below actually needs to be around this, otherwise this code may raise a PHP notice if `file` is not set.

It's not a problem of this PR but was already introduced in the previous implementation, I just discovered it now.
```suggestion
	$basename = isset( $metadata['file'] ) ? wp_basename( $metadata['file'] ) : null;
```

modules/images/webp-uploads/load.php
```diff
@@ -529,17 +529,18 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
 	foreach ( $urls as $url ) {
-		if ( isset( $metadata['file'] ) && strpos( $url, $basename ) !== false ) {
-			// TODO: we don't have a replacement for full image yet, issue. See: https://github.com/WordPress/performance/issues/174.
-			continue;
+		$src_filename = wp_basename( $url );
+
+		// Replace the full size image if present.
+		if ( isset( $metadata['file'], $metadata['sources'][ $target_mime ]['file'] ) && strpos( $url, $basename ) !== false ) {
```
felixarntz: Based on the above comment:
```suggestion
		if ( isset( $basename, $metadata['sources'][ $target_mime ]['file'] ) && strpos( $url, $basename ) !== false ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -529,17 +529,18 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+			$image = str_replace( $src_filename, $metadata['sources'][ $target_mime ]['file'], $image );
```
felixarntz, line 551: I think this is missing a `continue` statement? If we replaced the full image, we already took care of this URL, so we should skip the rest of this code block to proceed with the next URL.
```suggestion
			$image = str_replace( $src_filename, $metadata['sources'][ $target_mime ]['file'], $image );
			continue;
```

Also, please review the related line 565 below (I can't comment there since it wasn't changed here): I believe we should put a `break;` right after the image replacement, otherwise we'll always run through all sizes even if we already made a successful replacement, which would be pointless. This is also something which we should have caught before, but better now than never :)

felixarntz (commented):
@mitogh The new code overall looks solid, I found a few small issues that mostly relate to the already existing code, but we had missed them before.

modules/images/webp-uploads/load.php
mitogh: Actually, there's a conditional at the top of this function: 

https://github.com/WordPress/performance/blob/76f9392cb7dad591be273cc541ae90137297baf4/modules/images/webp-uploads/load.php#L510

This would prevent the execution of the code if the key is not present.

modules/images/webp-uploads/load.php
mitogh: Same as above as well.

modules/images/webp-uploads/load.php
mitogh: Good catch :+1:

modules/images/webp-uploads/load.php
felixarntz: Ah right, in that case the code below still needs to be updated though, see other comment.

modules/images/webp-uploads/load.php
felixarntz: Per your comment in https://github.com/WordPress/performance/pull/195#discussion_r822179813 this is already checked before though, so the check here should be removed entirely as it's redundant.
```suggestion
		if ( isset( $metadata['sources'][ $target_mime ]['file'] ) && strpos( $url, $basename ) !== false ) {
```

modules/images/webp-uploads/load.php
felixarntz: Note the second part of my comment as well. Your change https://github.com/WordPress/performance/pull/195/commits/7e1b5298b22f784feabde43b66e8191bafa5d6a1 adds the `continue` here, but not the `break` further down in the code.

mitogh: Right, let me update that one.

felixarntz (approved):
LGTM!

Let's wait with merging this though until #194 is ready, since technically the implementation here depends on how we implement #194.

felixarntz:
Ah nevermind the above, I see you've created this against your branch for #194. I think for the future it would be better to create both PRs based on `trunk`, since if we merge this now, it would be harder to review #194 as it would include the changes from here which were already approved.

So I'd still say it's better to wait until #194 has been approved before we merge this.

mitogh:
Thanks, mostly I've created this based on #194 just to have smaller diffs I agree this should is currently being blocked by #194 but having everything based on `trunk` will create merge conflicts after approval which usually triggers another review, so when having this approved and PR against #194 allows for simpler flow, however, I do see the problem specifically around dependencies between each PR.

PR #199, opened 2022-03-01
Added admin pointer

### LINKED ISSUE #193: Provide admin pointer informing about the plugin's admin screen
For WordPress plugins that don't add their own top-level menu item to the WordPress admin menu it has been notoriously hard to be noticeable after plugin activation. The "Performance Lab" plugin adds a new submenu item "Settings > Performance", but that's obviously hard to find since it's under the existing "Settings" top-level menu item.

Displaying a WP admin pointer was suggested in https://github.com/WordPress/performance/issues/61#issuecomment-1012844968, and this issue is focused on implementing that.

Related is #192 which also adds an alternative way to discover the plugin's admin screen.

My suggestion for the admin pointer copy would be a simple text like this:
* Title: _Performance Lab_
* Content: _You can now test upcoming WordPress performance features. Open Settings > Performance to individually toggle the performance features included in the plugin._
* The _Settings > Performance_ part could be a link to the admin screen.

cc @JustinyAhin

dainemawer:
## Summary

Fixes #193
Screenshot 2022-03-01 at 11 01 05

## Checklist

### DIFF
--- a/admin/load.php
+++ b/admin/load.php
@@ -416,7 +416,7 @@ function perflab_render_pointer() {
 
 			jQuery( '#menu-settings' ).pointer( options ).pointer( 'open' );
 
-			} );
+		} );
 	</script>
 	<?php
 }
--- a/admin/load.php
+++ b/admin/load.php
@@ -407,7 +407,7 @@ function perflab_render_pointer() {
 					jQuery.post(
 						window.ajaxurl,
 						{
-							pointer: '<?php echo esc_js( PERLAB_ADMIN_POINTER ); ?>',
+							pointer: 'perflab-admin-pointer',
 							action: 'dismiss-wp-pointer',
 						}
 					);
--- a/admin/load.php
+++ b/admin/load.php
@@ -4,6 +4,7 @@
  *
  * @package performance-lab
  */
+
 /**
  * Adds the modules page to the Settings menu.
  *
--- a/admin/load.php
+++ b/admin/load.php
@@ -392,7 +392,7 @@ function perflab_render_pointer() {
 		jQuery( function() {
 			// Pointer Options
 			var options = {
-				content: '<h3><?php echo esc_js( $heading ); ?></h3>' + '<p><?php echo wp_kses( $content, $wp_kses_options ); ?></p>',
+				content: '<h3><?php echo esc_js( $heading ); ?></h3><p><?php echo wp_kses( $content, $wp_kses_options ); ?></p>',
 				position: {
 					edge:  'left',
 					align: 'right',
--- a/admin/load.php
+++ b/admin/load.php
@@ -405,7 +405,7 @@ function perflab_render_pointer() {
 						window.ajaxurl,
 						{
 							pointer: 'perflab-admin-pointer',
-							action: 'dismiss-wp-pointer',
+							action:  'dismiss-wp-pointer',
 						}
 					);
 				}
--- a/admin/load.php
+++ b/admin/load.php
@@ -342,34 +342,6 @@ function perflab_get_module_data( $module_file ) {
 }
 
 /**
- * Initialise Admin Pointer
- *
- * Handles the bootstrapping of the admin pointer.
- * Mainly jQuery code that is self-initialising.
- *
- * @param string $hook_suffix The current admin page.
- * @since 1.0.0
- */
-function perflab_admin_pointer( $hook_suffix ) {
-
-	if ( ! in_array( $hook_suffix, array( 'index.php', 'plugins.php' ), true ) ) {
-		return;
-	}
-
-	$current_user = get_current_user_id();
-	$dismissed    = explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) );
-
-	if ( in_array( PERLAB_ADMIN_POINTER, $dismissed, true ) ) {
-		return;
-	}
-
-	// Enqueue pointer CSS and JS.
-	wp_enqueue_style( 'wp-pointer' );
-	wp_enqueue_script( 'wp-pointer' );
-}
-add_action( 'admin_enqueue_scripts', 'perflab_admin_pointer' );
-
-/**
  * Renders the Admin Pointer
  *
  * Handles the rendering of the admin pointer.
@@ -377,7 +349,6 @@ add_action( 'admin_enqueue_scripts', 'perflab_admin_pointer' );
  * @since 1.0.0
  */
 function perflab_render_pointer() {
-
 	$heading         = __( 'Performance Lab', 'performance-lab' );
 	$wp_kses_options = array(
 		'a' => array(
@@ -415,12 +386,39 @@ function perflab_render_pointer() {
 			};
 
 			jQuery( '#menu-settings' ).pointer( options ).pointer( 'open' );
-
 		} );
 	</script>
 	<?php
 }
-add_action( 'admin_print_footer_scripts', 'perflab_render_pointer' );
+
+/**
+ * Initialise Admin Pointer
+ *
+ * Handles the bootstrapping of the admin pointer.
+ * Mainly jQuery code that is self-initialising.
+ *
+ * @param string $hook_suffix The current admin page.
+ * @since 1.0.0
+ */
+function perflab_admin_pointer( $hook_suffix ) {
+
+	if ( ! in_array( $hook_suffix, array( 'index.php', 'plugins.php' ), true ) ) {
+		return;
+	}
+
+	$current_user = get_current_user_id();
+	$dismissed    = explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) );
+
+	if ( in_array( PERLAB_ADMIN_POINTER, $dismissed, true ) ) {
+		return;
+	}
+
+	// Enqueue pointer CSS and JS.
+	wp_enqueue_style( 'wp-pointer' );
+	wp_enqueue_script( 'wp-pointer' );
+	add_action( 'admin_print_footer_scripts', 'perflab_render_pointer' );
+}
+add_action( 'admin_enqueue_scripts', 'perflab_admin_pointer' );
 
 /**
  * Adds a link to the modules page to the plugin's entry in the plugins list table.
--- a/admin/load.php
+++ b/admin/load.php
@@ -4,9 +4,6 @@
  *
  * @package performance-lab
  */
-
-define( 'PERLAB_ADMIN_POINTER', 'perflab-admin-pointer' );
-
 /**
  * Adds the modules page to the Settings menu.
  *
--- a/admin/load.php
+++ b/admin/load.php
@@ -337,18 +337,29 @@ function perflab_get_module_data( $module_file ) {
 	return $module_data;
 }
 
+/**
+ * Initialise Admin Pointer
+ *
+ * Handles the bootstrapping of the admin pointer.
+ * Mainly jQuery code that is self-initialising.
+ *
+ * @since 1.0.0
+ * @return void
+ */
 function perflab_admin_pointer() {
 
-    wp_enqueue_style( 'wp-pointer' );
-    wp_enqueue_script( 'wp-pointer', false, array( 'jquery' ) );
+	wp_enqueue_style( 'wp-pointer' );
+	wp_enqueue_script( 'wp-pointer', false, array( 'jquery' ) );
 
 	$current_user = get_current_user_id();
-	$meta_key = 'my-pointer-slug-dismissed';
+	$meta_key     = 'my-pointer-slug-dismissed';
 
 	$heading = __( 'Performance Lab', 'performance-lab' );
-	$content = sprintf(__( 'You can now test upcoming WordPress performance features. Open %s to individually toggle the performance features included in the plugin.', 'performance-lab' ), '<a href="'. esc_url( admin_url( '/options-general.php?page=perflab-modules' ) ) .'">'. __( 'Settings > Performance', 'performance-lab' ) .'</a>');
+	/* translators: %s: settings page link */
+	$content = sprintf( __( 'You can now test upcoming WordPress performance features. Open %s to individually toggle the performance features included in the plugin.', 'performance-lab' ), '<a href="' . esc_url( admin_url( '/options-general.php?page=perflab-modules' ) ) . '">' . __( 'Settings > Performance', 'performance-lab' ) . '</a>' );
 
-	if ( ! get_user_meta( $current_user, $meta_key, true ) ) : ?>
+	if ( ! get_user_meta( $current_user, $meta_key, true ) ) :
+		?>
 		<script type="text/javascript">
 
 			jQuery(function() {
@@ -378,17 +389,27 @@ function perflab_admin_pointer() {
 
 			});
 		</script>
-	<?php endif;
+		<?php
+	endif;
 }
 
 add_action( 'in_admin_header', 'perflab_admin_pointer' );
 
-function  perflab_update_pointer_meta() {
+/**
+ * Update Pointer User Meta
+ *
+ * Checks if the action $_POST index is set and if it is, it will update the user meta.
+ * This prevents the pointer from showing on refresh.
+ *
+ * @since 1.0.0
+ * @return void
+ */
+function perflab_update_pointer_meta() {
 
 	$current_user = get_current_user_id();
-	$meta_key = 'my-pointer-slug-dismissed';
+	$meta_key     = 'my-pointer-slug-dismissed';
 
-	if ( isset( $_POST['action'] ) && 'dismiss-wp-pointer' == $_POST['action'] ) {
+	if ( isset( $_POST['action'] ) && 'dismiss-wp-pointer' === $_POST['action'] ) {
 		update_user_meta(
 			$current_user,
 			$meta_key,
--- a/admin/load.php
+++ b/admin/load.php
@@ -348,7 +348,6 @@ function perflab_get_module_data( $module_file ) {
  * @since 1.0.0
  */
 function perflab_admin_pointer( $hook_suffix ) {
-
 	if ( ! in_array( $hook_suffix, array( 'index.php', 'plugins.php' ), true ) ) {
 		return;
 	}
--- a/admin/load.php
+++ b/admin/load.php
@@ -359,7 +359,7 @@ function perflab_admin_pointer( $hook_suffix ) {
 	$current_user = get_current_user_id();
 	$dismissed    = explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) );
 
-	if ( in_array( PERLAB_ADMIN_POINTER, $dismissed, true ) ) {
+	if ( in_array( 'perflab-admin-pointer', $dismissed, true ) ) {
 		return;
 	}
 
--- a/admin/load.php
+++ b/admin/load.php
@@ -5,6 +5,8 @@
  * @package performance-lab
  */
 
+define( 'PERLAB_ADMIN_POINTER', 'perflab-admin-pointer' );
+
 /**
  * Adds the modules page to the Settings menu.
  *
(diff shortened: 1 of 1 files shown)


admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+ * @since 1.0.0
+ * @return void
+ */
+function perflab_admin_pointer() {
+
+	wp_enqueue_style( 'wp-pointer' );
+	wp_enqueue_script( 'wp-pointer', false, array( 'jquery' ) );
+
+	$current_user = get_current_user_id();
+	$meta_key     = 'my-pointer-slug-dismissed';
```
mitogh: I think this one might need a tweak. Something along the lines of:

```suggestion
	$meta_key     = 'performance-initial-pointer';
```

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+add_action( 'in_admin_header', 'perflab_admin_pointer' );
+
+/**
+ * Update Pointer User Meta
+ *
+ * Checks if the action $_POST index is set and if it is, it will update the user meta.
+ * This prevents the pointer from showing on refresh.
+ *
+ * @since 1.0.0
+ * @return void
```
mitogh: Even this returns `void` currently this is not supported by Core doc block.

```suggestion
```

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+ */
+function perflab_update_pointer_meta() {
+
+	$current_user = get_current_user_id();
+	$meta_key     = 'my-pointer-slug-dismissed';
```
mitogh: Same note as above.

admin/load.php
jjgrainger: Can we update the meta key and pointer slug to something more specific?

```suggestion
	$meta_key     = 'perflab-intro-pointer-dismissed';
```

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+
+	if ( isset( $_POST['action'] ) && 'dismiss-wp-pointer' === $_POST['action'] ) {
```
mitogh: An early return wold allow for an early return removing an indentation level in the code, example: 

```suggestion
	if ( ! isset( $_POST['action'] ) || 'dismiss-wp-pointer' !== $_POST['action'] ) {
	    return;
	}
```

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+		update_user_meta(
+			$current_user,
+			$meta_key,
+			$_POST['pointer'],
```
mitogh: Instead of writing directly the value passed by the FE, and because the value stored here is a boolean, we can assume at this point that the value should be `true`.

```suggestion
			true,
```

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+						edge:  'left',
+						align: 'right',
+					},
+					pointerClass: 'wp-pointer arrow-top',
+					pointerWidth: 420,
+					close: function() {
+						jQuery.post(
+							ajaxURL,
+							{
+								pointer: 'my-pointer-slug-dismissed',
```
jjgrainger: Same again, update the slug to something specific.

```suggestion
								pointer: 'perflab-intro-pointer-dismissed',
```

admin/load.php
jjgrainger:
```suggestion
	$meta_key     = 'perflab-intro-pointer-dismissed';
```

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+ * Handles the bootstrapping of the admin pointer.
+ * Mainly jQuery code that is self-initialising.
+ *
+ * @since 1.0.0
+ * @return void
+ */
+function perflab_admin_pointer() {
+
+	wp_enqueue_style( 'wp-pointer' );
+	wp_enqueue_script( 'wp-pointer', false, array( 'jquery' ) );
```
mitogh: Any particular reason not to load this in the footer instead?

```suggestion
	wp_enqueue_script( 'wp-pointer', false, array( 'jquery' ), true );
```

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+
+	$current_user = get_current_user_id();
+	$meta_key     = 'my-pointer-slug-dismissed';
+
+	$heading = __( 'Performance Lab', 'performance-lab' );
+	/* translators: %s: settings page link */
+	$content = sprintf( __( 'You can now test upcoming WordPress performance features. Open %s to individually toggle the performance features included in the plugin.', 'performance-lab' ), '<a href="' . esc_url( admin_url( '/options-general.php?page=perflab-modules' ) ) . '">' . __( 'Settings > Performance', 'performance-lab' ) . '</a>' );
```
mitogh: Would be great to split this sentence into multiple lines to make it easier to follow.

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+
+	if ( ! get_user_meta( $current_user, $meta_key, true ) ) :
```
mitogh: Similar to the other comment on the conditional reversing the conditional an early return would allow removing the indentation on this block.

```suggestion
	if ( get_user_meta( $current_user, $meta_key, true ) ) {
		return;
	} 
```

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+		?>
+		<script type="text/javascript">
+
+			jQuery(function() {
+
+				var ajaxURL = '<?php echo admin_url( 'admin-ajax.php' ); ?>';
+				// Pointer Options
+				var options = {
+					content: '<h3><?php echo $heading; ?></h3>' + '<p><?php echo $content; ?></p>',
```
jjgrainger: Does `$heading` and `$content` need to be escaped with [`esc_js`](https://developer.wordpress.org/reference/functions/esc_js/)? I'm not sure how this will impact the link formatting in the content.

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+	/* translators: %s: settings page link */
+	$content = sprintf( __( 'You can now test upcoming WordPress performance features. Open %s to individually toggle the performance features included in the plugin.', 'performance-lab' ), '<a href="' . esc_url( admin_url( '/options-general.php?page=perflab-modules' ) ) . '">' . __( 'Settings > Performance', 'performance-lab' ) . '</a>' );
+
+	if ( ! get_user_meta( $current_user, $meta_key, true ) ) :
+		?>
+		<script type="text/javascript">
+
+			jQuery(function() {
+
+				var ajaxURL = '<?php echo admin_url( 'admin-ajax.php' ); ?>';
```
mitogh: I believe `ajaxurl` is a global on the admin so you might not need to define this one, you can check for `window.ajaxurl`

jjgrainger (dismissed):
Looks good @dainemawer , just a couple of minor things to address.

admin/load.php
mitogh: Actually looking at the logic a bit more it seems like the main reason for this section is to handle the dismiss of the meta, right? if that's the case it would be better to remove the meta instead if the pointer is dismissed?

admin/load.php
felixarntz: I think the _sentence_ can probably remain one, but I agree we should break this code into multiple lines - maybe that's what you also meant.

admin/load.php
felixarntz: Actually, none of these additional parameters should be present - the script is registered by WP core, so we shouldn't tamper with it, other than enqueuing it. I don't think providing parameters here even does anything, for example the script depends on `jquery` anyway so it's unnecessary to provide that dependency here again.
```suggestion
	wp_enqueue_script( 'wp-pointer' );
```

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
 }
+
+/**
+ * Initialise Admin Pointer
+ *
+ * Handles the bootstrapping of the admin pointer.
+ * Mainly jQuery code that is self-initialising.
+ *
+ * @since 1.0.0
+ * @return void
```
felixarntz: Please remove this, void return types are supposed to be omitted per core docs standards.
```suggestion
```

admin/load.php
felixarntz: I think it would be better to use WordPress core's intended storage for this instead of introducing a new meta key altogether. We can use code like this:
```php
$dismissed = explode( ',', (string) get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) );
```

What you have here should then rather be the "pointer ID" used to put into that array if the user dismisses it.

admin/load.php
felixarntz: We must be careful here not to unconditionally enqueue anything. If the pointer is dismissed, we need to bail early. Based on my other comments, I'd suggest the following:
```suggestion
function perflab_admin_pointer() {
	$pointer_id = 'perflab-intro-pointer';
	$dismissed  = explode( ',', (string) get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) );
	if ( in_array( $pointer_id, $dismissed, true ) ) {
		return;
	}
	
	wp_enqueue_style( 'wp-pointer' );
	wp_enqueue_script( 'wp-pointer', false, array( 'jquery' ) );
```

admin/load.php
felixarntz: See my above comments:
* We need to bail early before we enqueue anything.
* We should probably not introduce a new meta key.

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+								action: 'dismiss-wp-pointer',
+							}
+						);
+					}
+				};
+
+				jQuery('#menu-settings').pointer(options).pointer('open');
+
+			});
+		</script>
```
felixarntz: This whole `<script...` should be printed later, so that we know for sure the `wp-pointer` script has already been included in the page. A good hook is `admin_print_footer_scripts`.

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+							ajaxURL,
+							{
+								pointer: 'my-pointer-slug-dismissed',
+								action: 'dismiss-wp-pointer',
+							}
+						);
+					}
+				};
+
+				jQuery('#menu-settings').pointer(options).pointer('open');
```
felixarntz: For the entire JS code in here (not just this line), please follow the current WP core JS coding standards - basically there need to be spaces around almost everything. Here for example:
```suggestion
				jQuery( '#menu-settings' ).pointer( options ).pointer( 'open' );
```

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+/**
+ * Update Pointer User Meta
+ *
+ * Checks if the action $_POST index is set and if it is, it will update the user meta.
+ * This prevents the pointer from showing on refresh.
+ *
+ * @since 1.0.0
+ * @return void
+ */
+function perflab_update_pointer_meta() {
```
felixarntz: This function can be entirely removed, since WP core already has its own logic to handle dismissed pointers. See my related comments about not introducing a custom meta key for this.

admin/load.php
felixarntz: I don't think that should include a `-dismissed` suffix, it can simply be a pointer name. The storage itself is only for dismissed pointers, so no point in including `-dismissed` in the name.
```suggestion
								pointer: 'perflab-intro-pointer',
```

Also, we should use a variable for this, this needs to use the same pointer ID we use further above in the code.

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+			$current_user,
+			$meta_key,
+			$_POST['pointer'],
+			true
+		);
+	}
+
+}
+
+add_action( 'admin_init', 'perflab_update_pointer_meta' );
```
felixarntz: See my comment above, this hook usage shouldn't be necessary.

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+
+				jQuery('#menu-settings').pointer(options).pointer('open');
+
+			});
+		</script>
+		<?php
+	endif;
+}
+
+add_action( 'in_admin_header', 'perflab_admin_pointer' );
```
felixarntz: I think `admin_enqueue_scripts` would be a more suitable hook since this enqueues something.

admin/load.php
```diff
@@ -336,3 +336,88 @@ function perflab_get_module_data( $module_file ) {
...
+/**
+ * Initialise Admin Pointer
+ *
+ * Handles the bootstrapping of the admin pointer.
+ * Mainly jQuery code that is self-initialising.
+ *
+ * @since 1.0.0
+ * @return void
+ */
+function perflab_admin_pointer() {
```
felixarntz: We should limit where we show this pointer, to not be too aggressive. WordPress core does that with its own admin pointers too.

Based on my above comment, a more suitable action for this callback would be `admin_enqueue_scripts`. Through that hook, we can also access the current `$hook_suffix` in this function. I'd suggest we use that to only show our pointer on the WP admin dashboard (`index.php`) and the plugins page from where the plugin gets activated (`plugins.php`). Right now we show the pointer anywhere, throughout the entire admin, which is not only a bit disrupting, but also increases the chance for conflicts with other plugins operating on certain screens.
```suggestion
function perflab_admin_pointer( $hook_suffix ) {
	if ( ! in_array( $hook_suffix, array( 'index.php', 'plugins.php' ), true ) ) {
		return;
	}

```

felixarntz (changes requested):
@dainemawer Left a couple of comments. This looks on the right track, but there's a lot of code here that can be simplified, and overall we need to be more careful of not enqueue things too aggressively.

admin/load.php
dainemawer: Updated, thanks @jjgrainger

admin/load.php
dainemawer: Agreed, I have removed this @felixarntz

admin/load.php
dainemawer: Updated!

admin/load.php
dainemawer: I have updated this to use `esc_js` for `$heading` and `wp_kses` for `$content`

admin/load.php
dainemawer: Ran with @felixarntz suggestion here.

admin/load.php
dainemawer: Removed, thanks @mitogh !

admin/load.php
dainemawer: Updated!

admin/load.php
dainemawer: I have amended this based on @felixarntz feedback, thanks @mitogh

admin/load.php
dainemawer: This function has been removed now :)

admin/load.php
dainemawer: This has been updated, args removed.

admin/load.php
dainemawer: I've split this up over multiple lines now, should be a bit easier to read.

admin/load.php
dainemawer: Wen with @felixarntz suggestion here!

admin/load.php
dainemawer: Right you are, updated!

admin/load.php
dainemawer: Removed!

admin/load.php
dainemawer: I've updated this, thanks @felixarntz !

admin/load.php
dainemawer: Removed 👍🏻

admin/load.php
dainemawer: Removed!

admin/load.php
dainemawer: Agreed, updated

admin/load.php
dainemawer: @felixarntz I've updated this, though I am running into a bit of a strange problem. This works great - and I tested / confirmed that the pointer only shows up on the Dashboard and Plugins page. However, I've added a new function to render the `script` in the footer called: `perflab_render_pointer` - I get a jQuery error on any other page that is not the Dashboard / Plugins page. The reason for this is that we are not calling the `wp_enqueue_script` function on other pages, but the script is still getting rendered. I tried duplicated your code above but it renders nothing if I do. I may need to take a further look into it.

felixarntz:
@dainemawer FYI Since we've branched off from `trunk` into a `release/1.0.0-beta.1` branch, I've updated the base branch of this PR to that new branch. So going forward, please make sure to work off `release/1.0.0-beta.1`, and specifically **not merge `trunk`** into this again.

dainemawer:
Roger that @felixarntz will fix the merge conflict here!

admin/load.php
```diff
@@ -339,6 +341,87 @@ function perflab_get_module_data( $module_file ) {
...
+				}
+			};
+
+			jQuery( '#menu-settings' ).pointer( options ).pointer( 'open' );
+
+			} );
+	</script>
+	<?php
+}
+add_action( 'admin_print_footer_scripts', 'perflab_render_pointer' );
```
felixarntz: Adding this action unconditionally is problematic since it'll mean that it always adds this script. Let's add this action from within the above `perflab_admin_pointer()` function instead so that it's only run if necessary.

admin/load.php
```diff
@@ -339,6 +341,87 @@ function perflab_get_module_data( $module_file ) {
...
+	$current_user = get_current_user_id();
+	$dismissed    = explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) );
+
+	if ( in_array( PERLAB_ADMIN_POINTER, $dismissed, true ) ) {
+		return;
+	}
+
+	// Enqueue pointer CSS and JS.
+	wp_enqueue_style( 'wp-pointer' );
+	wp_enqueue_script( 'wp-pointer' );
```
felixarntz, line 365: See my other comment:
```suggestion
	wp_enqueue_script( 'wp-pointer' );
	add_action( 'admin_print_footer_scripts', 'perflab_render_pointer' );
```

admin/load.php
```diff
@@ -5,6 +5,8 @@
  * @package performance-lab
  */
 
+define( 'PERLAB_ADMIN_POINTER', 'perflab-admin-pointer' );
```
felixarntz: There's a typo in the constant name here. More importantly though, this is a global constant just for a very specific purpose. I think we should remove it and instead use the pointer name through a local variable that only is known to the functions that need it.

admin/load.php
```diff
@@ -339,6 +341,87 @@ function perflab_get_module_data( $module_file ) {
...
+function perflab_admin_pointer( $hook_suffix ) {
+
+	if ( ! in_array( $hook_suffix, array( 'index.php', 'plugins.php' ), true ) ) {
+		return;
+	}
+
+	$current_user = get_current_user_id();
+	$dismissed    = explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) );
+
+	if ( in_array( PERLAB_ADMIN_POINTER, $dismissed, true ) ) {
```
felixarntz: See above, I think it's okay to duplicate the name here. While that's code duplication, it's only used twice in the entire code, and introducing a global constant for it is worse IMO.
```suggestion
	if ( in_array( 'perflab-admin-pointer', $dismissed, true ) ) {
```

If we wanted to also avoid the duplication, an alternative would be to pass the pointer in a variable to the `perflab_render_pointer( $pointer_id )` method. But that would also come with drawbacks since then we'd need to make the filter callback an anonymous function, which isn't really "WP core-friendly".

admin/load.php
```diff
@@ -339,6 +341,87 @@ function perflab_get_module_data( $module_file ) {
...
+					edge:  'left',
+					align: 'right',
+				},
+				pointerClass: 'wp-pointer arrow-top',
+				pointerWidth: 420,
+				close: function() {
+					jQuery.post(
+						window.ajaxurl,
+						{
+							pointer: '<?php echo esc_js( PERLAB_ADMIN_POINTER ); ?>',
```
felixarntz: See above:
```suggestion
							pointer: 'perflab-admin-pointer',
```

admin/load.php
```diff
@@ -339,6 +341,87 @@ function perflab_get_module_data( $module_file ) {
...
+		__( 'You can now test upcoming WordPress performance features. Open %s to individually toggle the performance features included in the plugin.', 'performance-lab' ),
+		'<a href="' . esc_url( admin_url( '/options-general.php?page=perflab-modules' ) ) . '">' . __( 'Settings > Performance', 'performance-lab' ) . '</a>'
+	);
+
+	?>
+	<script id="perflab-admin-pointer" type="text/javascript">
+		jQuery( function() {
+			// Pointer Options
+			var options = {
+				content: '<h3><?php echo esc_js( $heading ); ?></h3>' + '<p><?php echo wp_kses( $content, $wp_kses_options ); ?></p>',
```
felixarntz: The concatenation here is unnecessary.
```suggestion
				content: '<h3><?php echo esc_js( $heading ); ?></h3><p><?php echo wp_kses( $content, $wp_kses_options ); ?></p>',
```

admin/load.php
```diff
@@ -339,6 +341,87 @@ function perflab_get_module_data( $module_file ) {
...
+	$wp_kses_options = array(
+		'a' => array(
+			'href' => array(),
+		),
+	);
+
+	$content = sprintf(
+		/* translators: %s: settings page link */
+		__( 'You can now test upcoming WordPress performance features. Open %s to individually toggle the performance features included in the plugin.', 'performance-lab' ),
+		'<a href="' . esc_url( admin_url( '/options-general.php?page=perflab-modules' ) ) . '">' . __( 'Settings > Performance', 'performance-lab' ) . '</a>'
```
felixarntz: Bit of a nit-pick, but it may be preferable to use `add_query_arg()` here:
```suggestion
		'<a href="' . esc_url( add_query_arg( 'page', 'perflab-modules', admin_url( 'options-general.php' ) ) ) . '">' . __( 'Settings > Performance', 'performance-lab' ) . '</a>'
```

felixarntz (commented):
@dainemawer Left a bit more feedback here, this is looking close now.

admin/load.php
```diff
@@ -339,6 +341,87 @@ function perflab_get_module_data( $module_file ) {
...
+							pointer: '<?php echo esc_js( PERLAB_ADMIN_POINTER ); ?>',
+							action: 'dismiss-wp-pointer',
+						}
+					);
+				}
+			};
+
+			jQuery( '#menu-settings' ).pointer( options ).pointer( 'open' );
+
+			} );
```
mitogh: Removing the indentation level:

```suggestion
		} );
```

admin/load.php
```diff
@@ -339,6 +341,87 @@ function perflab_get_module_data( $module_file ) {
...
+					align: 'right',
+				},
+				pointerClass: 'wp-pointer arrow-top',
+				pointerWidth: 420,
+				close: function() {
+					jQuery.post(
+						window.ajaxurl,
+						{
+							pointer: '<?php echo esc_js( PERLAB_ADMIN_POINTER ); ?>',
+							action: 'dismiss-wp-pointer',
```
mitogh: Based on the code when this section of the code is executed this will trigger an Ajax call, but I was not able to find where the logic for this is handled in this PR.

Maybe I'm missing something?

felixarntz: @mitogh This is a WP core AJAX action, so it doesn't need to be implemented here. The original iteration of this PR had a custom implementation for dismissal, but it's better to reuse what core has for that.

mitogh: Oh I wasn't aware of that one, if that's the case it makes total sense to go with that approach.

admin/load.php
mitogh: Thanks @dainemawer I've moved this hook as suggested by @felixarntz to make sure the script is only executed for this page.

felixarntz (approved):
Awesome work @dainemawer!

Thanks @mitogh for the final iteration.

admin/load.php
```diff
@@ -339,6 +339,84 @@ function perflab_get_module_data( $module_file ) {
...
+					align: 'right',
+				},
+				pointerClass: 'wp-pointer arrow-top',
+				pointerWidth: 420,
+				close: function() {
+					jQuery.post(
+						window.ajaxurl,
+						{
+							pointer: 'perflab-admin-pointer',
+							action: 'dismiss-wp-pointer',
```
adamsilverstein:
```suggestion
							action:  'dismiss-wp-pointer',
```

adamsilverstein (approved):
one tiny spacing change to align the values, otherwise 👍🏼

PR #200, opened 2022-03-01
Always recommend object cache on multisite

### LINKED ISSUE #161: Promote object caching in Site Health for multisites
## Feature Description

Follow up issue from https://github.com/WordPress/performance/issues/35. 

For sites that are running multisite without object caching, a warning should be displayed to users in the site health screen.

tillkruss:
## Summary

This is a followup to #111 and resolves #161.

A @spacedmonkey suggested, multisite greatly benefits from object caching, due to the extra queries to bootstrap WordPress.

## Checklist

### DIFF
--- a/modules/object-cache/persistent-object-cache-health-check/load.php
+++ b/modules/object-cache/persistent-object-cache-health-check/load.php
@@ -138,6 +138,10 @@ function perflab_oc_health_persistent_object_cache() {
 function perflab_oc_health_should_persistent_object_cache() {
 	global $wpdb;
 
+	if ( is_multisite() ) {
+		return true;
+	}
+
 	/**
 	 * Filter to force suggestion to use a persistent object cache and bypass threshold checks.
 	 *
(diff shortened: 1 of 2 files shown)


modules/object-cache/persistent-object-cache-health-check/load.php
```diff
@@ -138,6 +138,10 @@ function perflab_oc_health_persistent_object_cache() {
 function perflab_oc_health_should_persistent_object_cache() {
```
felixarntz: This is technically an oversight from the previous PR, but this function name is missing a word - let's fix that here :)

It should probably be `perflab_oc_health_should_suggest_persistent_object_cache()`.

felixarntz (commented):
@tillkruss LGTM, except for one thing that is more of an oversight from #111.

modules/object-cache/persistent-object-cache-health-check/load.php
tillkruss: Yikes

felixarntz (approved):
🎉

mitogh (approved):

PR #205, opened 2022-03-03
Mark Site Health Audit Enqueued Assets module as experimental for now.

felixarntz:
## Summary

Follow-up for #25

## Relevant technical choices

* Mark module as experimental based on latest comments.

## Checklist

### DIFF
--- a/modules/site-health/audit-enqueued-assets/load.php
+++ b/modules/site-health/audit-enqueued-assets/load.php
@@ -2,7 +2,7 @@
 /**
  * Module Name: Audit Enqueued Assets
  * Description: Adds a CSS and JS resource check in Site Health status.
- * Experimental: No
+ * Experimental: Yes
  *
  * @package performance-lab
  * @since 1.0.0


mitogh (approved):

ThierryA (approved):
LGTM

Making this experimental makes sense since:

1. it still has open follow up issue
2. some concern about the threshold was raised lately
3. it is much easier to start as experimental and then upgrade to stable than the other way around

felixarntz:
@manuelRod Would be great to get your feedback on the issue / this PR as well since you've done most of the work on the module. To clarify, it's not that there is a problem with the module, it's more about that it doesn't feel as polished yet, hence marking as "experimental" for this first release - as it sees more refinement over time, we could then mark it as non-experimental.

manuelRod:
@felixarntz I left some feedback on the previous issue [25](https://github.com/WordPress/performance/issues/25).

To sum up:
Said that I'm not totally sure if marking this as experimental ( I'm not against either ) would make a difference. To me, experimental sounds more like a change that may break your site or cause unexpected behaviors, this is just an inoffensive warning in the site health screen. And at this stage, this whole plugin is experimental itself, so people using it are kinda aware of it ( or at least, that's how I see it ).

To sum up, I see "experimental" as a different thing, but if we all agree, I'm in too.

audrasjb (approved):
LGTM 👍

felixarntz:
@manuelRod 
> experimental sounds more like a change that may break your site or cause unexpected behaviors

Indeed this wouldn't break anyone's site, but IMO the main point for marking this as experimental is that it's still in an earlier stage of development compared to the other modules. There are still foundational iterations happening, e.g. we haven't fully defined what the thresholds should be, and the approach to gathering the assets is known to be not yet reliable for certain environments. We also need to think about how to make this warning more actionable, e.g. at a minimum tell users which plugins or themes are responsible for enqueuing the majority of assets etc.

As we work towards these enhancements and define the intended scope of the module further, eventually we could then remove the experimental label from it.

PR #207, opened 2022-03-07
Use `original` image to generate all additional image format sub-sizes

### LINKED ISSUE #204: When creating additional mime types use the original image instead of attached image.
## Bug Description

The current implementation to create images in additional mime types, the image uses the attached image, with `get_attached_file` when the image is larger a `scaled` version would be created, however, WordPress creates all the subsizes using the original image instead of the `scaled` version.
 
The main problem with the current approach is that in some cases won't create an image of the exact size as the original image, one thing to mention is that the created image on the additional mime types would be of the same aspect ratio, usually, this happens when the scaled version is smaller in dimensions that the created image. 

One of the major problems when using the original image is that the resulting WebP images would result bigger in size than the same version in a JPEG, at least specifically with this image and the default quality setting for WebP, this is mostly because the current approach creates a WebP version out of an already compressed version. So would be something to consider either adjust the default WebP from within the plugin to make sure the generated WebP files result in smaller versions of the same JPEG image.

## Steps to reproduce

Given the following image subsizes:

```php
> array(6) {
  ["thumbnail"]=>
  array(3) {
    ["width"]=>
    int(150)
    ["height"]=>
    int(150)
    ["crop"]=>
    bool(true)
  }
  ["medium"]=>
  array(3) {
    ["width"]=>
    int(300)
    ["height"]=>
    int(300)
    ["crop"]=>
    bool(false)
  }
  ["medium_large"]=>
  array(3) {
    ["width"]=>
    int(768)
    ["height"]=>
    int(0)
    ["crop"]=>
    bool(false)
  }
  ["large"]=>
  array(3) {
    ["width"]=>
    int(1024)
    ["height"]=>
    int(1024)
    ["crop"]=>
    bool(false)
  }
  ["1536x1536"]=>
  array(3) {
    ["width"]=>
    int(1536)
    ["height"]=>
    int(1536)
    ["crop"]=>
    bool(false)
  }
  ["2048x2048"]=>
  array(3) {
    ["width"]=>
    int(2048)
    ["height"]=>
    int(2048)
    ["crop"]=>
    bool(false)
  }
}
```

1. Upload the following image

9VQGY3hyMKs

2. Add the following filter on your site:

```php
		// Use a 1500 threshold
		add_filter(
			'big_image_size_threshold',
			function () {
				return 1500;
			}
		);
```
 
3. And observe the metadata of the image and observe the size `1536x1536` the 2 images have different sizes:

Original JPEG image dimensions are: 1024x1536
Additional WebP dimensions are: 1000x1500

## Additional Context

 - PHP Version: 7
 - OS: Linux 5.15.12-1-MANJARO x86_64
 - Browser: Chrome
 - Plugin Version: `release/1.0.0-beta.1`
 - Device: Desktop.

mitogh:
## Summary

Fixes #204 

## Relevant technical choices

When creating all the sub-sizes for additional mime types, using the original image instead of the attached image would ensure that all additional sub-sizes have the same size as the original images. 

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -39,7 +39,7 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 		return $metadata;
 	}
 
-	$file = get_attached_file( $attachment_id, true );
+	$file = wp_get_original_image_path( $attachment_id, true );
```
felixarntz: What's the reasoning behind setting the `$unfiltered` parameter here? WP core doesn't set that either.

modules/images/webp-uploads/load.php
```diff
@@ -167,16 +167,16 @@ function webp_uploads_generate_image_size( $attachment_id, $size, $mime ) {
...
 	}
 
-	if ( isset( $metadata['sizes'][ $size ]['height'] ) ) {
-		$height = $metadata['sizes'][ $size ]['height'];
-	} elseif ( isset( $sizes[ $size ]['height'] ) ) {
+	if ( isset( $sizes[ $size ]['height'] ) ) {
 		$height = $sizes[ $size ]['height'];
+	} elseif ( isset( $metadata['sizes'][ $size ]['height'] ) ) {
+		$height = $metadata['sizes'][ $size ]['height'];
 	}
```
felixarntz: Why are these conditions reversed? I'm not saying it's a problem, but would like to understand the reasoning behind it.

modules/images/webp-uploads/load.php
```diff
@@ -185,7 +185,6 @@ function webp_uploads_generate_image_size( $attachment_id, $size, $mime ) {
 
 	$editor->resize( $width, $height, $crop );
 	$filename = $editor->generate_filename( null, null, $extension );
-	$filename = preg_replace( '/-(scaled|rotated|imagifyresized)/', '', $filename );
```
felixarntz: Nice that we can get rid of this now 👍

modules/images/webp-uploads/load.php
```diff
@@ -185,7 +185,6 @@ function webp_uploads_generate_image_size( $attachment_id, $size, $mime ) {
 
 	$editor->resize( $width, $height, $crop );
```
felixarntz: Now that we use the original image, we need to make sure to reapply any rotation based on the `orientation` EXIF data as needed, using the `maybe_exif_rotate()` method. This wasn't needed before when using the scaled or rotated image which already included that, but now we need it.

See the [`_wp_make_subsizes()` function](https://developer.wordpress.org/reference/functions/_wp_make_subsizes/) which also does that.

It's potentially not great to just add that here since it would mean we do this for every sub-size, which would be wasteful, for example the above WP core function does it only once and then generates all subsizes from it.

But for now let's just add it here so that it works, we can separately optimize the performance of that - restructuring the code that much would be a bit too intense for the scope of this PR>

felixarntz (commented):
@mitogh Mostly looks good, I left a few comments.

The one critical thing we need to add here is about considering the orientation and applying the rotation as needed, similar to how WP core does.

modules/images/webp-uploads/load.php
mitogh: Mostly to avoid the need to pass in filters when accessing the file. As in case of wanted to filter here there's always `wp_get_original_image_path` let me know if you think that makes sense.

felixarntz: > Mostly to avoid the need to pass in filters when accessing the file.

But why? I'd say we can let the filters run as usual here.

modules/images/webp-uploads/load.php
mitogh: Previously when using the image sizes in some scenarios additional WebP images were created in different dimensions see #204 

The idea was to use the dimensions from the image in order to match closely the original image, however, this was a wrong assumption as the problem was the scaled version was used instead, as in those scenarios the scaled version was smaller than the added images sizes.

Now using the original image size there's no reason to go with that approach and know the size used is the image size to accurately reflect the subsize, and just in a scenario where the dimensions are not part of the image size fallbacks on the image dimensions on the created subsize.

modules/images/webp-uploads/load.php
mitogh: Great point, let me add that one.

modules/images/webp-uploads/load.php
mitogh: Well to avoid the need for an additional call to a filter in a place where we are not really required due we need to access to the actual image file. 

I see the benefit in keeping the filters to maximize compatibility tho.

felixarntz (approved):
Great work @mitogh!

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -722,4 +722,33 @@ public function it_should_create_a_file_in_the_specified_location_with_the_speci
...
+
+		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/paint.jpeg' );
+		$metadata      = wp_get_attachment_metadata( $attachment_id );
+
+		$this->assertArrayHasKey( '1536x1536', $metadata['sizes'] );
+		foreach ( $metadata['sizes'] as $size ) {
+			$this->assertStringContainsString( $size['width'], $size['sources']['image/webp']['file'] );
+			$this->assertStringContainsString( $size['height'], $size['sources']['image/webp']['file'] );
+			$this->assertStringContainsString(
+				// Remove the extension ouf of the file.
```
adamsilverstein:
```suggestion
				// Remove the extension from the file.
```

modules/images/webp-uploads/load.php
adamsilverstein: thanks for explaining @mitogh, I had the same question!

adamsilverstein (changes requested):
Excellent! Left one tiny typo correction.

adamsilverstein (approved):

PR #221, opened 2022-03-08
Update release instructions to include proper branching strategy and protect release branches

### LINKED ISSUE #214: Update release documentation to separate release branch creation from release day process
As the initial release has shown (see #213), there were a few problems with the release process originally defined in #131. Most importantly, we need to change the branching strategy a bit, to properly allow for creating the release branch ahead of time, which is necessary to not block merging of other PRs ready for `trunk`.

That release branch was already created early for the initial release, which made sense due to the above, but then following the release instructions created some problems. Also, it shouldn't be allowed to just directly commit against the release branches, they should be protected in the same way as `trunk`. In other words, PRs should also have to go against the `release/**` branches.

felixarntz:
## Summary

Fixes #214

## Checklist

### DIFF
--- a/docs/Releasing-the-plugin.md
+++ b/docs/Releasing-the-plugin.md
@@ -76,7 +76,9 @@ After successful testing, inform the other maintainers on [Slack](https://wordpr
 
 ### Update `trunk`
 
-In case any additional code has landed in the release branch after it was branched off the main branch `trunk`, the latter needs to be updated to include these changes. To do that, open a pull request from the release branch to `trunk` (can be done via GitHub UI) and ask two maintainers to approve it. The code in this pull request does not need to be closely reviewed since it was already approved prior, so this is mostly a formal approval.
+Since the release branch was branched off the main branch `trunk`, the latter now needs to be updated to include any additional changes made to the release branch since.
+
+To do that, open a pull request from the release branch to `trunk` (can be done via GitHub UI) and ask two maintainers to approve it. The code in this pull request does not need to be closely reviewed since it was already approved prior, so this is mostly a formal approval.
 
 ### Close the release
 
--- a/docs/Releasing-the-plugin.md
+++ b/docs/Releasing-the-plugin.md
@@ -47,7 +47,7 @@ After running the command, check the `readme.txt` file to ensure the new changel
 
 ### Open a pull request
 
-Push your local branch to the GitHub repository and open a pull request against the release branch `release/{milestoneName}`. Make sure to tag at least 2 plugin maintainers to request a review, as usual. Once the pull request is approved by at least 2 plugin maintainers, the pull request can be merged.
+Push your local branch to the GitHub repository and open a pull request against the release branch `release/{milestoneName}`. As usual, make sure to request a review from at least 2 plugin maintainers. Once approved, the pull request can be merged.
 
 ## Publishing the release
 
--- a/docs/Releasing-the-plugin.md
+++ b/docs/Releasing-the-plugin.md
@@ -4,17 +4,32 @@
 
 This document describes the steps to release the Performance plugin.
 
+## Branching off `trunk`
+
+The cutoff point for a release specifies the point in time where no new features or enhancements should go into the release, only critical fixes. There is no hard rule for when this point should be reached, but the loose guidance is around a week before the tentative release date.
+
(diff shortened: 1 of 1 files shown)


docs/Releasing-the-plugin.md
```diff
@@ -24,24 +39,49 @@ In addition to those locations, run the `npm run since -- -r {version}` command
...
+After running the command, check the `readme.txt` file to ensure the new changelog for the release has been added. Also review its changelog entries for whether they make sense and are understandable. Make sure that no unexpected changelog entries are present.
 
 ### Open a pull request
 
-Push your local release branch to the GitHub repository and open a pull request against `trunk`. Make sure to tag at least 2 plugin maintainers to request a review. Once the pull request is approved by at least 2 plugin maintainers, the pull request can be merged.
-
-If this is a major or minor release, please keep the release branch around since that branch can be used to later create patch releases from the same state of the codebase.
+Push your local branch to the GitHub repository and open a pull request against the release branch `release/{milestoneName}`. Make sure to tag at least 2 plugin maintainers to request a review, as usual. Once the pull request is approved by at least 2 plugin maintainers, the pull request can be merged.
```
dainemawer: "As usual, make sure to request a review from 2 plugin maintainers. Once approved, the pull request can be merged."

docs/Releasing-the-plugin.md
```diff
@@ -24,24 +39,49 @@ In addition to those locations, run the `npm run since -- -r {version}` command
...
+After that, [create a new release tag for the plugin on GitHub](https://github.com/WordPress/performance/releases/new). The release tag should have the same name as the corresponding milestone used earlier, and it should be created from the release branch `release/{milestoneName}`. Finally, add the changelog (it can be found in the `readme.txt` file) to the release description and create the release.
+
+### Review and test the deployment
+
+Once a new version is released on GitHub, the plugin will be deployed to the [WordPress.org repository](https://wordpress.org/plugins/performance-lab/). [The deployment workflow progress can be tracked here.](https://github.com/WordPress/performance/actions/workflows/deploy-dotorg.yml)
+
+At this point, test the update process from your WordPress site, getting the plugin from wordpress.org. The main aspects to test at this point are:
+
+* Install or update the plugin via wordpress.org, ZIP upload, WP-CLI, etc.
+* Test the plugin on different WordPress core versions supported.
```
dainemawer, line 69: @felixarntz - considering the number of WP core versions - do we want to scope this? Maybe the previous 3 versions or something along those line? Just a thought!

docs/Releasing-the-plugin.md
```diff
@@ -24,24 +39,49 @@ In addition to those locations, run the `npm run since -- -r {version}` command
...
+* Test the plugin on a multisite environment as well.
```
dainemawer, line 70: Should we add some direction with regards to local environments to test this on? Like `wp-env`?

dainemawer (approved):
@felixarntz left a few comments, approving regardless :)

docs/Releasing-the-plugin.md
felixarntz: @dainemawer Note the "different WordPress core versions _supported_" which is very limited. The plugin will always have one of the latest WP core versions as a minimum requirement, so I don't foresee this to ever be more than 2 or 3 WP core versions.

For example, right now we would only need to test with WP 5.8 and WP 5.9.

docs/Releasing-the-plugin.md
felixarntz: @dainemawer I think that would be a bit overkill here, it can be any environment where these points can be tested. It can be a local site, somebody's production site, a quickly spun up test site using one of those throwaway site providers - so IMO specifying more details here wouldn't be necessary, but if it was, it should go elsewhere since it would go too much into detail and thus make this doc too long - it's already fairly long based on the current more crucial requirements defined.

docs/Releasing-the-plugin.md
dainemawer: Awesome :)

docs/Releasing-the-plugin.md
dainemawer: Cool beans!

JustinyAhin (approved):
Great additions @felixarntz.

PR #222, opened 2022-03-08
Replace unreliable activation hook with default value for enabled modules

### LINKED ISSUE #218: Default settings not working as expected in multisite
## Bug Description

When we install the plugin in a multisite environment and "Network Activate" the plugin, the default settings are only saved for the main site and not all the sites in the network.

## Steps to reproduce

1. Install Performance Lab plugin in a WordPress multisite environment
2. "Network Activate" the plugin
3. Check the default settings for the Performance Lab plugin in any site other than main site

## Screenshots

As can be seen none of the options are selected.
image

## Additional Context

 - PHP Version: 7.4
 - Plugin Version: [1.0.0-beta.1](https://github.com/WordPress/performance/releases/tag/1.0.0-beta.1)

### LINKED ISSUE #219: Default settings not working with WP-CLI
## Bug Description

When the plugin is activated using `wp-cli` the default options are not saved.

## Steps to reproduce

1. Install the plugin either via GUI `wp-admin/plugins.php` or via CLI `wp plugin install performance-lab`
2. Activate the plugin using `wp plugin activate performance-lab`
3. Check Performance Lab plugin settings

## Screenshots

image

## Additional Context

 - PHP Version: 7.4
 - Plugin Version: [1.0.0-beta.1](https://github.com/WordPress/performance/releases/tag/1.0.0-beta.1)

felixarntz:
## Summary

Fixes #218
Fixes #219

## Relevant technical choices

* The usage of plugin activation hooks is universally problematic when it comes to environments like multisite or tools like WP-CLI, since the hook may not get triggered, or in multisite it would only be triggered on the main site when network-activating the plugin.
* Therefore this PR replaces it with using a default value for the option.
    * The straightforward implementation here would be to either hard-code the list of non-experimental modules in there or to dynamically read them from the filesystem, however both of that is not desirable: Hard-coding would result in manual maintenance which would be easily overlooked, and dynamically reading from the filesystem would be bad for performance since this would then need to run on every page load, even in the frontend of the site.
    * Because of the above, this PR uses a separate PHP file `default-enabled-modules.php` which, similarly to the existing `module-i18n.php` file, can be automatically generated and maintained.
* For the above, the PR implements a new CLI script similar to some of the existing ones, which allows to refresh the new auto-generated PHP file with a single command.
* The release instructions are being expanded accordingly to include running this new command before every plugin release.

## Checklist

### DIFF
--- a/bin/plugin/commands/common.js
+++ b/bin/plugin/commands/common.js
@@ -62,11 +62,8 @@ exports.getModuleData = async ( modulesDir ) => {
 
 			// Parse experimental field into a boolean.
 			if ( typeof moduleData.experimental === 'string' ) {
-				if ( moduleData.experimental.toLowerCase() === 'yes' ) {
-					moduleData.experimental = true;
-				} else {
-					moduleData.experimental = false;
-				}
+				moduleData.experimental =
+					moduleData.experimental.toLowerCase() === 'yes';
 			}
 
 			return moduleData;
--- a/bin/plugin/commands/common.js
+++ b/bin/plugin/commands/common.js
@@ -6,32 +6,43 @@ const glob = require( 'fast-glob' );
 const fs = require( 'fs' );
 
 /**
- * @typedef WPModuleDescription
+ * @typedef WPModuleData
  *
- * @property {string} name        Module name.
- * @property {string} description Module description.
+ * @property {string}  slug         Module slug.
+ * @property {string}  focus        Module focus.
+ * @property {string}  name         Module name.
+ * @property {string}  description  Module description.
+ * @property {boolean} experimental Whether the module is experimental.
  */
 
 /**
- * Returns a promise resolving to the module description list string for the `readme.txt` file.
+ * Returns a promise resolving to the list of data for all modules.
  *
  * @param {string} modulesDir Modules directory.
  *
- * @return {Promise<[]WPModuleDescription>} Promise resolving to module description list.
+ * @return {Promise<[]WPModuleData>} Promise resolving to module data list.
  */
-exports.getModuleDescriptions = async ( modulesDir ) => {
+exports.getModuleData = async ( modulesDir ) => {
 	const moduleFilePattern = path.join( modulesDir, '*/*/load.php' );
 	const moduleFiles = await glob( path.resolve( '.', moduleFilePattern ) );
 
 	return moduleFiles
 		.map( ( moduleFile ) => {
+			// Populate slug and focus based on file path.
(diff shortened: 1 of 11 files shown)


gagan0123:
Tested scenarios, 
1. Multisite
   1. Activate on single site
   2. Network Activate
3. WP-CLI install and activate in single and multisite
4. WP-CLI activate in single site

All scenarios now working as expected with default values as they should be.

gagan0123 (approved):

bin/plugin/commands/common.js
```diff
@@ -49,6 +60,15 @@ exports.getModuleDescriptions = async ( modulesDir ) => {
 				match = regex.exec( fileContent );
 			}
 
+			// Parse experimental field into a boolean.
+			if ( typeof moduleData.experimental === 'string' ) {
+				if ( moduleData.experimental.toLowerCase() === 'yes' ) {
+					moduleData.experimental = true;
+				} else {
+					moduleData.experimental = false;
+				}
```
mitogh: This one can be expressed as a single statement without introducing complexity as the names of properties are expressive enough.

```suggestion
				moduleData.experimental = moduleData.experimental.toLowerCase() === 'yes';
```

load.php
```diff
@@ -76,7 +87,7 @@ function( $module_settings ) {
  * @return array Associative array of module settings keyed by module slug.
  */
 function perflab_get_module_settings() {
-	return (array) get_option( PERFLAB_MODULES_SETTING, array() );
+	return (array) get_option( PERFLAB_MODULES_SETTING );
```
mitogh, line 90: When this one is called for the first time this would return 1 element array:

```php
array( false )
```

Which is handled in other parts of the code by using `array_filter` but I would suggest to make sure this returns consistent results instead.

```suggestion
	return array_filter( (array) get_option( PERFLAB_MODULES_SETTING ) );
```

So when this is called when the setting does not exists this would return an empty array instead.

felixarntz: @mitogh It wouldn't return `array( false )`, since this setting has a registered default now (see above). The point of removing the second parameter here was to ensure the registered default is used. So the `get_option` call here can actually never result in `false` based on that.

mitogh: Good point :+1:

mitogh (approved):

bin/plugin/commands/common.js
felixarntz: Great point, updated!

JustinyAhin (approved):

PR #223, opened 2022-03-09
Use `.gitattributes` instead of `.distignore` to better support ZIP creation

### LINKED ISSUE #217: Introduce `.gitattributes` in favor of `.distignore`
## Feature Description

The current release zips created by GH everytime a new tag is created includes files that are not part of the release, for example:

- https://github.com/WordPress/performance/archive/refs/tags/1.0.0-beta.1.zip

Contains tests and configuration files, non required for testing or usage of the plugin.

The deploy action would use either:

- `.disignore` 
- `.gitattributes`

To exclude files that are going to be deployed so we can safely move `.disignore` to `.gitattributes` one of the things to consider is when moving the content into `.gitattributes` is the fact that the setting `export-ignore` should be added after each file for example:

```
/.git export-ignore
/node_modules export-ignore
```

`export-ignore`

> Files and directories with the attribute export-ignore won’t be added to archive files.

Additional [parameters](https://git-scm.com/docs/gitattributes).

thelovekesh:
## Summary

Fixes #217 

## Relevant technical choices

- Move `.disignore` to `.gitattributes`
- Add `README` file as `export-ignore` in `.gitattributes`

## Checklist


mitogh (approved):
Thanks, this looks good and works as expected, after running the `git archive` command.

thelovekesh:
Thanks, @mitogh for looking into this. I think this PR needs one more approval so can you please add reviewers?

mitogh:
That's correct @thelovekesh let's wait on someone else to take a look at this and give us some feedback before we can move forward with the merge, otherwise, feel free to mention in our next call next Tuesday (if you are unable to make it) I would make sure to mention about this.

Great work and thank you for your collaboration to the project.

felixarntz (approved):
Great work @thelovekesh - thank you for the PR!

PR #224, opened 2022-03-09
Adds sources information to the attachment media details of the REST response

### LINKED ISSUE #167: Expose `source_url` for all additional mime types via REST API
## Feature Description

WordPress already has a defined endpoint for each media item located at `wp-json/wp/v2/media/{ATTACHMENT_ID}` this ticket is to suggest the inclusion of additional mime types (if present) for that particular image beside the default mime type for an image, this would allow API consumers to have access to both image mime types using the REST API.

Another suggested approach could be to add a new query parameter to allow developers to request a specific mime type instead so the API would remain the same but with the difference that internally would query for the requested mime type if present and deliver those results back via the REST API.

This can be accomplished from within the plugin filtering `rest_prepare_attachment` to insert or update the response as need it. 

### Related issues

- #142 
- #150 
- #155

eugene-manuilov:
## Summary

Fixes #167

## Checklist


tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -531,4 +531,48 @@ public function data_provider_not_supported_webp_images() {
...
+	 *
+	 * @test
+	 */
+	public function it_should_add_sources_to_rest_response() {
+		$file_location = TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg';
+		$attachment_id = $this->factory->attachment->create_upload_object( $file_location );
+		$metadata      = wp_get_attachment_metadata( $attachment_id );
+
+		$request       = new WP_REST_Request();
+		$request['id'] = $attachment_id;
```
mitogh, line 633: Personally due `$request` is an object I would go with the object approach:

```suggestion
		$request->set_param( 'id', $attachment_id );
```

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -531,4 +531,48 @@ public function data_provider_not_supported_webp_images() {
...
+			$this->assertIsArray( $properties['sources'] );
+
+			foreach ( $mime_types as $mie_type ) {
+				$this->assertArrayHasKey( $mie_type, $properties['sources'] );
+
+				$this->assertArrayHasKey( 'filesize', $properties['sources'][ $mie_type ] );
+				$this->assertArrayHasKey( 'file', $properties['sources'][ $mie_type ] );
+				$this->assertArrayHasKey( 'source_url', $properties['sources'][ $mie_type ] );
+
+				$this->assertNotFalse( filter_var( $properties['sources'][ $mie_type ]['source_url'], FILTER_VALIDATE_URL ) );
```
mitogh: Looks like a typo?

```suggestion
			foreach ( $mime_types as $mime_type ) {
				$this->assertArrayHasKey( $mime_type, $properties['sources'] );

				$this->assertArrayHasKey( 'filesize', $properties['sources'][ $mime_type ] );
				$this->assertArrayHasKey( 'file', $properties['sources'][ $mime_type ] );
				$this->assertArrayHasKey( 'source_url', $properties['sources'][ $mime_type ] );

				$this->assertNotFalse( filter_var( $properties['sources'][ $mime_type ]['source_url'], FILTER_VALIDATE_URL ) );
```

modules/images/webp-uploads/load.php
```diff
@@ -475,3 +475,43 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+ * @param WP_REST_Response $response The original response object.
+ * @param WP_Post          $post     The post object.
+ * @param WP_REST_Request  $request  The request object.
+ * @return WP_REST_Response A new response object for the attachment with additional sources.
+ */
+function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Post $post, WP_REST_Request $request ) {
+	$metadata = wp_get_attachment_metadata( $post->ID );
+	if ( empty( $metadata['file'] ) ) {
+		return $response;
+	}
```
mitogh: Curios about the need for this conditional if the `file` key is not accessed in this function.

modules/images/webp-uploads/load.php
```diff
@@ -475,3 +475,43 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+	foreach ( $data['media_details']['sizes'] as $size => $details ) {
+		if ( empty( $metadata['sizes'][ $size ]['sources'] ) || ! is_array( $metadata['sizes'][ $size ]['sources'] ) ) {
+			continue;
+		}
+
+		$sources = array();
+		foreach ( $metadata['sizes'][ $size ]['sources'] as $mime => $mime_details ) {
+			$source_url       = dirname( $data['media_details']['sizes'][ $size ]['source_url'] ) . '/' . $mime_details['file'];
+			$sources[ $mime ] = array_merge( $mime_details, array( 'source_url' => $source_url ) );
+		}
```
mitogh: Small optimization:

```suggestion
		$directory = dirname( $data['media_details']['sizes'][ $size ]['source_url'] );
		foreach ( $metadata['sizes'][ $size ]['sources'] as $mime => $mime_details ) {
			$source_url       = "{$directory}/{$mime_details['file']}";
			$sources[ $mime ] = array_merge( $mime_details, array( 'source_url' => $source_url ) );
		}
```

modules/images/webp-uploads/load.php
```diff
@@ -475,3 +475,43 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+
+	foreach ( $data['media_details']['sizes'] as $size => $details ) {
+		if ( empty( $metadata['sizes'][ $size ]['sources'] ) || ! is_array( $metadata['sizes'][ $size ]['sources'] ) ) {
+			continue;
+		}
+
+		$sources = array();
+		foreach ( $metadata['sizes'][ $size ]['sources'] as $mime => $mime_details ) {
+			$source_url       = dirname( $data['media_details']['sizes'][ $size ]['source_url'] ) . '/' . $mime_details['file'];
+			$sources[ $mime ] = array_merge( $mime_details, array( 'source_url' => $source_url ) );
```
mitogh: In addition to that avoiding the need to create a full new array traversing all the elements of the array just appending the new value seems good enough.
```suggestion
			$mime_details['source_url'] = $source_url;
			$sources[ $mime ] = $mime_details;
```

mitogh (commented):
Great work on this one, few minor comments/questions.

felixarntz (commented):
@eugene-manuilov Nothing to add from my end other than what @mitogh has already commented on. Other than that, this looks solid 👍

tests/modules/images/webp-uploads/webp-uploads-test.php
eugene-manuilov: The `WP_REST_Request` implements the `ArrayAccess` interface and `$request[...] = ...` assignments are valid as well as direct `set_param` method calls. REST controllers in the core also use array access approach, see here: https://github.com/WordPress/wordpress-develop/blob/5.9/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L555

tests/modules/images/webp-uploads/webp-uploads-test.php
eugene-manuilov: Yep, thanks.

modules/images/webp-uploads/load.php
eugene-manuilov: Just a concise way to ensure that the metadata array contains data that belongs to an image.

modules/images/webp-uploads/load.php
eugene-manuilov: Makes sense. Thanks.

modules/images/webp-uploads/load.php
eugene-manuilov: Yep, also works! Thanks.

modules/images/webp-uploads/load.php
mitogh: I'm not sure if it follow that part, I think this sectionof code is not required due you already have this section in place few lines below:

```
if ( ! isset( $data['media_details']['sizes'] ) || ! is_array( $data['media_details']['sizes'] ) ) {
```

Which already covers that scenario.

tests/modules/images/webp-uploads/webp-uploads-test.php
mitogh: Good point. :heavy_check_mark:

modules/images/webp-uploads/load.php
felixarntz: I agree that check here looks unnecessary. A more appropriate check if we want to avoid running the following logic unnecessarily may be `empty( $metadata['sizes'] )` for example.

felixarntz (approved):
One minor comment, but overall lgtm!

modules/images/webp-uploads/load.php
eugene-manuilov: @mitogh @felixarntz ok, this check is removed. 👍

mitogh (approved):
:+1:

PR #225, opened 2022-03-10
Switch deploy workflow (released & prereleased → published)

### LINKED ISSUE #216: Update Workflow to deploy to WordPress.org
## Feature Description

The current deployment would be executed when a tag is `released` or `prereleased` 

- https://github.com/WordPress/performance/blob/trunk/.github/workflows/deploy-dotorg.yml#L6-L7

From the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release):

>  Note: The prereleased type will not trigger for pre-releases published from draft releases, but the published type will trigger

Same from the docs as well:

>  If you want a workflow to run when stable and pre-releases publish, subscribe to published instead of released and prereleased.

Based on the current workflow to deploy the `published` event would fit better

>  If you want a workflow to run when stable and pre-releases publish, subscribe to published instead of released and prereleased.

### Actionable items

Replace `released` and `prereleased` with `published` in the current workflow used to deploy to WordPress.org, the workflow can be found here:

- https://github.com/WordPress/performance/blob/trunk/.github/workflows/deploy-dotorg.yml

### Related issues

- #214

TechSolomon:
## Summary

Resolves #216 

## Relevant technical choices

This pull request builds on https://github.com/WordPress/performance/issues/214, reverting the release infrastructure decision made in https://github.com/WordPress/performance/pull/89#discussion_r787848618.

## Checklist

### DIFF
--- a/.github/workflows/deploy-dotorg.yml
+++ b/.github/workflows/deploy-dotorg.yml
@@ -2,7 +2,8 @@ name: Deploy to WordPress.org
 
 on:
   release:
-    types: [published]
+    types:
+      - published
 
 jobs:
   release:
--- a/.github/workflows/deploy-dotorg.yml
+++ b/.github/workflows/deploy-dotorg.yml
@@ -2,9 +2,7 @@ name: Deploy to WordPress.org
 
 on:
   release:
-    types:
-      - released
-      - prereleased
+    types: [published]
 
 jobs:
   release:


.github/workflows/deploy-dotorg.yml
```diff
@@ -2,9 +2,7 @@ name: Deploy to WordPress.org
 
 on:
   release:
-    types:
-      - released
-      - prereleased
+    types: [published]
```
felixarntz: Nit-pick, but maybe better to use the expanded syntax for consistency with the rest of our workflows.
```suggestion
    types:
      - published
```

felixarntz: Nit-pick, but maybe better to use the expanded syntax for consistency with the rest of our workflows.
```suggestion
    types:
      - published
```

felixarntz (approved):
@TechSolomon Thank you for the pull request! Looks good, just one tiny point of feedback. If you could address it, that would be great.

.github/workflows/deploy-dotorg.yml
TechSolomon: Fixed. ✅

felixarntz (approved):
@TechSolomon Thanks again! 🙌

mitogh (approved):

JustinyAhin (approved):

JustinyAhin (approved):

PR #32, opened 2021-12-02
Add WebP for uploads module 

adamsilverstein:
Fixes https://github.com/WordPress/performance/issues/4 
Also addresses https://github.com/WordPress/performance/issues/22

With this module enabled, uploaded JPEG images are saved as WebP sub-sizes.

* Adds test that: 
	* Validate the addition of the jpeg -> webp conversion mapping to the `image_editor_output_format` filter
	* Verify the filter is only applied when system supports WebP.

PR #227, opened 2022-03-11
Allow developers to tweak which image formats to generate on upload

### LINKED ISSUE #187: Allow developers to tweak which image formats should be used
#142 changes the default implementation of the `webp-uploads` module to generate WebP versions _in addition to_ JPEG versions for all sub-sizes of an image. This was decided in #96, replacing the original implementation from #4 where only WebP versions would be generated.

It has already been established that we would still like to provide developers with the flexibility to make their own decision though: For example, if they want to only generate WebP versions, or even only JPEG versions, they should be free to do so.

This issue is about defining the developer API and integration to control that behavior.

eugene-manuilov:
## Summary

Fixes #187 

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -30,9 +30,18 @@
 function webp_uploads_create_sources_property( array $metadata, $attachment_id ) {
 	// This should take place only on the JPEG image.
 	$valid_mime_transforms = webp_uploads_get_supported_image_mime_transforms();
+
 	// Not a supported mime type to create the sources property.
 	$mime_type = get_post_mime_type( $attachment_id );
-	if ( ! isset( $valid_mime_transforms[ $mime_type ] ) ) {
+	if ( ! isset( $valid_mime_transforms[ $mime_type ] ) || ! is_array( $valid_mime_transforms[ $mime_type ] ) ) {
```
mitogh: We decided to remove the conditional to verify if a valid mime transform is an array.

- https://github.com/WordPress/performance/pull/194#discussion_r823010943

modules/images/webp-uploads/load.php
```diff
@@ -157,6 +166,56 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+	// Skip conversion when creating the `-scaled` image (for large image uploads).
+	if ( preg_match( '/-scaled\..{3}.?$/', $filename ) ) {
+		return $output_format;
+	}
+
+	// Use the original mime type if this type is allowed.
+	$valid_mime_transforms = webp_uploads_get_supported_image_mime_transforms();
+	if (
+		! isset( $valid_mime_transforms[ $mime_type ] ) ||
+		! is_array( $valid_mime_transforms[ $mime_type ] ) ||
```
mitogh: Same comment as above.

modules/images/webp-uploads/load.php
```diff
@@ -157,6 +166,56 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+ * @param string $output_format The image editor default output format mapping.
+ * @param string $filename      Path to the image.
+ * @param string $mime_type     The source image mime type.
+ * @return string The new output format mapping.
+ */
+function webp_uploads_filter_image_editor_output_format( $output_format, $filename, $mime_type ) {
+	// Skip conversion when creating the `-scaled` image (for large image uploads).
+	if ( preg_match( '/-scaled\..{3}.?$/', $filename ) ) {
+		return $output_format;
+	}
```
mitogh: We actually want to do this we change this behavior in this issue:

- https://github.com/WordPress/performance/issues/174

modules/images/webp-uploads/load.php
```diff
@@ -157,6 +166,56 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+	// Find the first supported mime type by the image editor to use it as the default one.
+	foreach ( $valid_mime_transforms[ $mime_type ] as $target_mime ) {
+		if ( wp_image_editor_supports( array( 'mime_type' => $target_mime ) ) ) {
+			$new_mime_type = $target_mime;
+			break;
+		}
+	}
+
+	// Leave the original mime type if there is no other mime type supported by the image editor.
+	if ( empty( $new_mime_type ) ) {
```
```suggestion
	if ( null === $new_mime_type ) {
```

Makes more sense, due we are checking if no mime type was set above.

modules/images/webp-uploads/load.php
```diff
@@ -157,6 +166,56 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+
+	// Use the original mime type if this type is allowed.
+	$valid_mime_transforms = webp_uploads_get_supported_image_mime_transforms();
+	if (
+		! isset( $valid_mime_transforms[ $mime_type ] ) ||
+		! is_array( $valid_mime_transforms[ $mime_type ] ) ||
+		in_array( $mime_type, $valid_mime_transforms[ $mime_type ], true )
+	) {
+		return $output_format;
+	}
```
mitogh, line 182: I think the condition is not required as we actually need to return the first element from the array instead.

```suggestion
	) {
		return $valid_mime_transforms[ $mime_type ][0];
	}
```

That would effectively use the first element of the array as the primary mime type so for example if you have:

```php
'image/jpeg' => array( 'image/webp', 'image/jpeg' );
```

It should create by default the images as WebP and the sources as JPEG instead. However, I'm thinking that returning the first element prevents other mime types from the list to be returned so in that scenario the filter should be removed from where the sources is created. 

Let me know what you think or maybe I may be missing something.

modules/images/webp-uploads/load.php
```diff
@@ -157,6 +166,56 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+
+	$new_mime_type = null;
+
+	// Find the first supported mime type by the image editor to use it as the default one.
+	foreach ( $valid_mime_transforms[ $mime_type ] as $target_mime ) {
+		if ( wp_image_editor_supports( array( 'mime_type' => $target_mime ) ) ) {
```
mitogh, line 186: This condition does not seem required as the editor will check if the mime is supported. I agree this helps avoid passing incorrect values but that also won't work in creating the `sources` attributes anyways so the condition can be ignored.

modules/images/webp-uploads/load.php
eugene-manuilov: Your suggestion will not work if the returned value from the `webp_uploads_get_supported_image_mime_transforms` function is an empty array or an array with modified mime types.

modules/images/webp-uploads/load.php
eugene-manuilov: This condition is required to make sure that we don't use image formats that are not supported by the server.

modules/images/webp-uploads/load.php
eugene-manuilov: Ok, updated.

modules/images/webp-uploads/load.php

modules/images/webp-uploads/load.php
eugene-manuilov: ok, removed this condition.

modules/images/webp-uploads/load.php
mitogh: Right, however this is already checked by the editor image right after this is called:

- https://github.com/WordPress/wordpress-develop/blob/f0dfa682a5cf2004f468c18ba8d5c87c463cdeea/src/wp-includes/class-wp-image-editor.php#L370

modules/images/webp-uploads/load.php
mitogh: - If `webp_uploads_get_supported_image_mime_transforms` return `array()` this would not create the `sources` property which is expected in this case.
- If the provided mime type is invalid the code already handles this scenario here: https://github.com/WordPress/wordpress-develop/blob/f0dfa682a5cf2004f468c18ba8d5c87c463cdeea/src/wp-includes/class-wp-image-editor.php#L370 https://github.com/WordPress/performance/blob/87cc7f2dcc88d6f446a3a86a49ba22d65e85ce17/modules/images/webp-uploads/load.php#L335

modules/images/webp-uploads/load.php
eugene-manuilov: And? Not sure what you suggest... Not checking whether the mime type is supported and blindly return the first one? My understanding is that we need to return the first **supported** mime type to make sure everything works correctly and the image editor doesn't need to fallback to the [default format](https://github.com/WordPress/wordpress-develop/blob/f0dfa682a5cf2004f468c18ba8d5c87c463cdeea/src/wp-includes/class-wp-image-editor.php#L376-L378).

modules/images/webp-uploads/load.php
eugene-manuilov: > If webp_uploads_get_supported_image_mime_transforms return array() this would not create the sources property which is expected in this case.

This function is not about `sources`, this function tells the image editor which format to use for the default subsizes. If the `webp_uploads_get_supported_image_mime_transforms` function returns an empty array then your suggested changes to return the first item from the mime transforms array will cause an error/warning because there is no elements with `0` key in that array.

The same will happen if someone unassigns the first mime type in a hook for our filter:

```php
add_filter( 'webp_uploads_upload_image_mime_transforms', function( $transorms ) {
    unset( $transforms[ 'image/webp' ][ array_search( 'image/webp', $transforms[ 'image/webp' ] ) );
    return $transforms;
} );

function webp_uploads_get_supported_image_mime_transforms() {
	$image_mime_transforms = array(
		'image/jpeg' => array( 'image/jpeg', 'image/webp' ),
	);

	$transforms = (array) apply_filters( 'webp_uploads_upload_image_mime_transforms', $image_mime_transforms );

    ...
}

function webp_uploads_filter_image_editor_output_format( $output_format, $filename, $mime_type ) {
	// Use the original mime type if this type is allowed.
	$valid_mime_transforms = webp_uploads_get_supported_image_mime_transforms();
	if (
		! isset( $valid_mime_transforms[ $mime_type ] ) ||
		in_array( $mime_type, $valid_mime_transforms[ $mime_type ], true )
	) {
		// This will cause php warning for all images uploaded in webp format because it is unset in the hook:
		// ---> PHP Warning:  Undefined array key 0 ...
		return $valid_mime_transforms[ $mime_type ][0]; 
	}

	...
}
```

> If the provided mime type is invalid the code already handles this scenario here:

Once again the condition that you refer to will not be executed if we return the incorrect mime type here. Thus the image editor will have to fallback to the default mime type.

mitogh: Makes sense I was confusing it to a higher level before this point, thanks for the feedback.

modules/images/webp-uploads/load.php
mitogh: Sorry for the confusion, my suggestion here is around the need for the conditional as if you provide a nonsupported mime type from the filter the editor won't work anyway due which I think should be the expected behavior instead of changing let's say `image/avif` to `image/jpeg` if `image/avif` is not supported no image should be generated instead as currently, it would create images in a non requested mime type.

eugene-manuilov: > let's say `image/avif` to `image/jpeg` if `image/avif` is not supported no image should be generated instead as currently, it would create images in a non requested mime type.

If `image/avif` is not supported and no images are generated, what is the point of our plugin then? In this case, the plugin only makes worse by breaking subsizes functionality in WordPress. Our plugin should not make harm if a new format is not supported.

modules/images/webp-uploads/load.php
```diff
@@ -30,12 +30,21 @@
...
 	}
 
+	// Don't do anything if only the original mime type is needed.
+	if (
+		1 === count( $valid_mime_transforms[ $mime_type ] ) &&
+		current( $valid_mime_transforms[ $mime_type ] ) === $mime_type
+	) {
+		return $metadata;
+	}
+
```
felixarntz: I think this needs to be removed. While I can see how it would be nice to not run the logic below, it also populates the `sources` data even for the image formats that already exist, and we need to make sure that runs for consistent behavior.

As far as I can tell, this function is already set up correctly to not double-create any image in a MIME type that already exists, so this check here isn't necessary from that perspective either.

Maybe the first sub-bullet point in https://github.com/WordPress/performance/issues/187#issuecomment-1061277586 was confusing in that regard, my intention there was just to clarify this needs to be the case, not that we actually need to add code for this if it's already working as expected (which I think it does).

modules/images/webp-uploads/load.php
felixarntz: I agree with @eugene-manuilov that this check is needed to avoid trying to create a MIME type that isn't supported and therefore will fail. This check was also part of the original module implementation from #4 and should be kept for the same purpose.

modules/images/webp-uploads/load.php
```diff
@@ -157,6 +166,50 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+		return $output_format;
+	}
+
+	$new_mime_type = null;
+
+	// Find the first supported mime type by the image editor to use it as the default one.
+	foreach ( $valid_mime_transforms[ $mime_type ] as $target_mime ) {
+		if ( wp_image_editor_supports( array( 'mime_type' => $target_mime ) ) ) {
+			$new_mime_type = $target_mime;
+			break;
```
felixarntz: We can simplify the implementation here by returning immediately instead of having a separate `$new_mime_type` variable and extra conditions for it below.
```suggestion
			$output_format[ $mime_type ] = $target_mime;
			return $output_format;
```

modules/images/webp-uploads/load.php
```diff
@@ -157,6 +166,50 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+		}
+	}
+
+	// Leave the original mime type if there is no other mime type supported by the image editor.
+	if ( null === $new_mime_type ) {
+		return $output_format;
+	}
+
+	$output_format[ $mime_type ] = $new_mime_type;
+
```
felixarntz: See my above comment, this can be removed then.

modules/images/webp-uploads/load.php
```diff
@@ -219,28 +272,30 @@ function webp_uploads_generate_image_size( $attachment_id, $size, $mime ) {
...
-	foreach ( $transforms as $mime => $list_transforms ) {
-		if ( ! is_array( $list_transforms ) ) {
-			continue;
-		}
-		$valid_transforms[ $mime ] = $list_transforms;
+	$transforms = (array) apply_filters( 'webp_uploads_upload_image_mime_transforms', $image_mime_transforms );
+
+	// Return an empty array if non-array result is returned from the filter.
+	if ( ! is_array( $transforms ) ) {
+		return array();
```
felixarntz: Instead of returning an array, for an invalid filter usage like this we should return the default value specified above. We can put that into a separate `$default_image_mime_transforms` to access it again here.

modules/images/webp-uploads/load.php
```diff
@@ -219,28 +272,30 @@ function webp_uploads_generate_image_size( $attachment_id, $size, $mime ) {
  */
 function webp_uploads_get_supported_image_mime_transforms() {
```
felixarntz: Since we're changing the filter name, let's change the function name accordingly as well. For example to `webp_uploads_get_upload_image_mime_transforms()`.

felixarntz (commented):
@eugene-manuilov Mostly looks good, I left a few comments for things that can be improved. One of them is crucial, we still need to run through the overall `webp_uploads_create_sources_property()` function as before.

modules/images/webp-uploads/load.php
eugene-manuilov: Ok. Updated.

modules/images/webp-uploads/load.php
eugene-manuilov: Yep, makes sense. Updated.

modules/images/webp-uploads/load.php

modules/images/webp-uploads/load.php
eugene-manuilov: Yep. Updated.

modules/images/webp-uploads/load.php

modules/images/webp-uploads/load.php
```diff
@@ -157,6 +158,41 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+		if ( wp_image_editor_supports( array( 'mime_type' => $target_mime ) ) ) {
+			$output_format[ $mime_type ] = $target_mime;
+			break;
+		}
+	}
+
+	return $output_format;
+}
+
+add_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
```
felixarntz: Could be considered a personal preference, but I think it's more common with functional WP plugins to have the `add_filter` / `add_action` call right below the callback function.
```suggestion
}
add_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
```

modules/images/webp-uploads/load.php
```diff
@@ -157,6 +158,41 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 
 add_filter( 'wp_generate_attachment_metadata', 'webp_uploads_create_sources_property', 10, 2 );
```
felixarntz: See above, I guess I missed that before. Not critical in any case.
```suggestion
add_filter( 'wp_generate_attachment_metadata', 'webp_uploads_create_sources_property', 10, 2 );
```

felixarntz (approved):
@eugene-manuilov Awesome!

adamsilverstein:
Overall this looks great, I will give this a test specifically to verify that the filter works to set output to a single mime type (eg transform jpeg->webp) and also test with and without WebP support. 

Can we add some additional tests that validates new functionality as well?

modules/images/webp-uploads/load.php

modules/images/webp-uploads/load.php

eugene-manuilov:
> Overall this looks great, I will give this a test specifically to verify that the filter works to set output to a single mime type (eg transform jpeg->webp) and also test with and without WebP support.
> 
> Can we add some additional tests that validates new functionality as well?

@adamsilverstein @felixarntz @mitogh I have found and fixed a few issues while working on phpunit tests. Could you, please, take another look at the implementation?

Also I have refactored a bit our phpunit tests for the webp-uploads module to make it easier to check sources. Let me know what you think.

modules/images/webp-uploads/load.php
```diff
@@ -276,7 +278,15 @@ function webp_uploads_get_upload_image_mime_transforms() {
 		return $default_transforms;
 	}
 
-	return array_filter( $transforms, 'is_array' );
+	// Ensure that all mime types have correct transforms. If a mime type has invalid transforms array,
```
adamsilverstein, line 281: great

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -280,31 +257,18 @@ public function it_should_create_a_webp_version_with_all_the_required_properties
...
-		$this->assertArrayHasKey( 'image/jpeg', $metadata['sizes']['thumbnail']['sources'] );
-		$this->assertArrayHasKey( 'filesize', $metadata['sizes']['thumbnail']['sources']['image/jpeg'] );
-		$this->assertArrayHasKey( 'file', $metadata['sizes']['thumbnail']['sources']['image/jpeg'] );
-		$this->assertArrayHasKey( 'image/webp', $metadata['sizes']['thumbnail']['sources'] );
-		$this->assertArrayHasKey( 'filesize', $metadata['sizes']['thumbnail']['sources']['image/webp'] );
-		$this->assertArrayHasKey( 'file', $metadata['sizes']['thumbnail']['sources']['image/webp'] );
-		$this->assertStringEndsNotWith( '.jpeg', $metadata['sizes']['thumbnail']['sources']['image/webp']['file'] );
-		$this->assertStringEndsWith( '.webp', $metadata['sizes']['thumbnail']['sources']['image/webp']['file'] );
+		$this->assertImageHasSizeSource( 'image/jpeg', 'thumbnail', $attachment_id );
+		$this->assertImageHasSizeSource( 'image/webp', 'thumbnail', $attachment_id );
```
mitogh: I think the abstraction works great to ensure the presence of the key this 2 assertions would ensure that the correct file extensions are used and can be used after `assertImageHasSizeSource` as at that point if the previous assertions pass it can safely be used at this stage.

```suggestion
		$this->assertImageHasSizeSource( 'image/webp', 'thumbnail', $attachment_id );
		$this->assertStringEndsNotWith( '.jpeg', $metadata['sizes']['thumbnail']['sources']['image/webp']['file'] );
		$this->assertStringEndsWith( '.webp', $metadata['sizes']['thumbnail']['sources']['image/webp']['file'] );		
```

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -795,4 +755,33 @@ function () {
 			);
 		}
 	}
+
+	/**
+	 * Tests that we can force transformation from jpeg to webp by using the webp_uploads_upload_image_mime_transforms filter.
+	 *
+	 * @test
+	 */
+	public function it_should_transofrm_jpeg_to_webp_subsizes_using_transform_filter() {
```
mitogh:
```suggestion
	public function it_should_transform_jpeg_to_webp_subsizes_using_transform_filter() {
```

tests/utils/Constraint/ImageHasSizeSource.php
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	/**
+	 * Returns a string representation of the constraint.
+	 *
+	 * @return string String representation of the constraint.
+	 */
+	public function toString(): string {
+		return sprintf(
+			'%s the source with the "%s" mime type for the "%s" size',
```
mitogh, line 35:
```suggestion
			'It %s the source with the "%s" mime type for the "%s" size',
```

tests/utils/Constraint/ImageHasSource.php
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	/**
+	 * Returns a string representation of the constraint.
+	 *
+	 * @return string String representation of the constraint.
+	 */
+	public function toString(): string {
+		return sprintf(
+			'%s the source with the "%s" mime type',
```
mitogh, line 50:
```suggestion
			'It %s the source with the "%s" mime type',
```

tests/utils/Constraint/ImageHasSizeSource.php
```diff
@@ -0,0 +1,67 @@
...
+			$this->mime_type,
+			$this->size
+		);
+	}
+
+	/**
+	 * Evaluates the constraint for the provided attachment ID.
+	 *
+	 * @param int $attachment_id Attachment ID.
+	 * @return bool TRUE if the attachment has a source with the requested mime type for the subsize, otherwise FALSE.
```
mitogh, line 46:
```suggestion
	 * @return bool `true` if the attachment has a source with the requested mime type for the subsize, otherwise `false`.
```

tests/utils/TestCase/ImagesTestCase.php
```diff
@@ -0,0 +1,79 @@
...
+	/**
+	 * Asserts that an image has a source with the specific mime type.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $mime_type The mime type of the source.
+	 * @param int    $attachment_id The attachment ID.
+	 * @param string $message An optional message to show on failure.
+	 */
+	public static function assertImageHasSource( $mime_type, $attachment_id, $message = '' ) {
```
mitogh: I would update the parameters as the mime type depends on the `$attachment_id` it makes more sense to set first the `attachment_id`

```suggestion
	public static function assertImageHasSource( $attachment_id, $expected_mime_type, $message = '' ) {
```

tests/utils/TestCase/ImagesTestCase.php
```diff
@@ -0,0 +1,79 @@
...
+	/**
+	 * Asserts that an image doesn't have a source with the specific mime type.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $mime_type The mime type of the source.
+	 * @param int    $attachment_id The attachment ID.
+	 * @param string $message An optional message to show on failure.
+	 */
+	public static function assertImageNotHasSource( $mime_type, $attachment_id, $message = '' ) {
```
mitogh: Same as above.

```suggestion
	public static function assertImageNotHasSource( $attachment_id, $expected_mime_type, $message = '' ) {
```

tests/utils/TestCase/ImagesTestCase.php
```diff
@@ -0,0 +1,79 @@
...
+	 * Asserts that an image has a source with the specific mime type for a subsize.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $mime_type The mime type of the source.
+	 * @param string $size The subsize name.
+	 * @param int    $attachment_id The attachment ID.
+	 * @param string $message An optional message to show on failure.
+	 */
+	public static function assertImageHasSizeSource( $mime_type, $size, $attachment_id, $message = '' ) {
```
mitogh:
```suggestion
	public static function assertImageHasSizeSource( $attachment_id, $size, $expected_mime_type, $message = '' ) {
```

Since you are already using types would be great to specify the types as well for the parameters.

```suggestion
	public static function assertImageHasSizeSource( int $attachment_id, string $size, string $expected_mime_type, string $message = '' ) {
```

tests/utils/TestCase/ImagesTestCase.php
```diff
@@ -0,0 +1,79 @@
...
+	 * Asserts that an image doesn't have a source with the specific mime type for a subsize.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $mime_type The mime type of the source.
+	 * @param string $size The subsize name.
+	 * @param int    $attachment_id The attachment ID.
+	 * @param string $message An optional message to show on failure.
+	 */
+	public static function assertImageNotHasSizeSource( $mime_type, $size, $attachment_id, $message = '' ) {
```
mitogh: Similar to the above.

```suggestion
	public static function assertImageNotHasSizeSource( $attachment_id, $size, $expected_mime_type, $message = '' ) {
```

mitogh (commented):
Minor comments, great work on refactoring the tests :)

tests/utils/Constraint/ImageHasSizeSource.php
eugene-manuilov: @mitogh this string is used by phpunit to compose success/failure messages and it shouldn't start from `It` because it will cause the composed message  to look quite strange.

tests/utils/Constraint/ImageHasSource.php
eugene-manuilov: Same here. We can't use `It` here because this message is used in the phpunit to compose success/failure messages.

tests/utils/TestCase/ImagesTestCase.php
eugene-manuilov: The current order is used to mimic how the `ArrayHasKey` constraint work. The first param/params are needly and then goes the target that we want to check.

tests/utils/TestCase/ImagesTestCase.php
eugene-manuilov: Same reason. Just mimic how other constraints in phpunit work.

tests/utils/TestCase/ImagesTestCase.php
eugene-manuilov: I have added return types to custom constraints because it was required by php to make sure that the signature is compatible with the parent class. I am not sure about adding these types for params. Seems less important to me, but if we agree to add it, I'll update.

tests/utils/TestCase/ImagesTestCase.php
mitogh: Right, however by using this abstraction the second argument is no longer an array and if you are unfamiliar with the abstraction itself is difficult to know which is the base of the abstraction like if I want to use the assertion I would expect the first parameter to be the attachment ID similarly as how WP functions work due the expected_mime_type and additional parameters are based out of the attachment ID.

The logic comes from importance order:

- Attachment ID -> This is required and would always be present 
- Size -> This is also present as is a WP API the only difference is the size name in some cases might not be defined
- Mime Type -> This is a newly introduced value so it might not be present in most cases. 

Based on that order it makes more sense to pass the parameters in that order instead of the other way around.

tests/utils/Constraint/ImageHasSizeSource.php
eugene-manuilov: Is it a standard? I am not sure what is the right approach for mentioning the boolean value in phpdoc descriptions. Does "`" make any sense in phpdoc?

tests/utils/TestCase/ImagesTestCase.php
felixarntz: I don't have a strong opinion on either, but slightly leaning towards @mitogh's point here, since that order is what's more common in WordPress image functions.

The argument around `assertArrayHasKey` makes sense, but here we're passing the attachment ID, and the assertion is not looking for the size "in" the attachment ID. It's looking for it in the metadata array. If the assertion was `assertMetadataHasSource`, I think the `assertArrayHasKey` comparison would make more sense, but because we're passing an attachment ID here, IMO sticking to the order WordPress developers are used to makes more sense.

tests/utils/TestCase/ImagesTestCase.php
felixarntz: See my reply above, slightly leaning towards @mitogh's point. But I think either way has its benefits here, I don't have a strong opinion.

composer.json
```diff
@@ -34,5 +34,10 @@
       "dealerdirect/phpcodesniffer-composer-installer": true,
       "composer/installers": true
     }
+  },
+  "autoload-dev": {
+    "psr-4": {
+      "PerformanceLab\\Tests\\": "tests/utils"
+    }
   }
```
felixarntz, line 42: Not sure Composer usage like this would play well together with WordPress core, but since it's only tests IMO it should be okay at least in terms of this plugin. The classes would need to be renamed for core anyway.

felixarntz (approved):
@eugene-manuilov Great work refactoring the tests with the extra assertions! I have some minor comments, but nothing blocking.

adamsilverstein:
I did some testing to verify that the filter mapping works as expected, using the filter to test the following scenarios when uploading both jpegs and webp images:

* generate only WebP sub-sizes
* generate only jpeg sub-sizes
* generate both WebP and jpeg sub-sizes
* generate both, reversing the default

In code:
```php
add_filter( 'webp_uploads_upload_image_mime_transforms', function() {
	$test = 'only-webp'; // Try each

	switch ( $test ) {
		case 'only-webp':
			return [
				'image/jpeg' => [ 'image/webp' ],
				'image/webp' => [ 'image/webp' ],
			];
		case 'only-jpg':
			return [
				'image/jpeg' => [ 'image/jpeg' ],
				'image/webp' => [ 'image/jpeg' ],
			];
		case 'both':
			return [
				'image/jpeg' =>  [ 'image/jpeg', 'image/webp' ],
				'image/webp' =>  [ 'image/webp', 'image/jpeg' ],
			];
		case 'reverse':
			return [
				'image/jpeg' =>  [ 'image/webp', 'image/jpeg' ],
				'image/webp' =>  [ 'image/jpeg', 'image/webp' ],
			];
		case 'empty':
		default:
			return [];
	}

});
```
I reviewed the files generated as well as the meta data. Comments below:

✅  only-webp: worked as expected, only webp images generated for either webp or jpeg uploads
⚠️ only-jpeg: worked for jpegs, only jpeg was generated. _when I uploaded a WebP it generated webp, I was expecting jpeg_ - not sure of the use case, but was expecting this to work
✅  both: worked as expected, both formats created including -scaled version, worked for both jpeg and webp uploads
❌  reverse: didn't work as expected - behavior was the same as 'both', I was expecting the alternate mime type to be stored as the "primary" image, meaning stored in the sizes array, and first in the sources arrays. Not sure this is an issue or if others will expect this behavior as well. Not sure this is completely needed here, this is a little trickier to achieve in the plugin, for the core patch it will be much easier to just start with the first mime type.
✅  'empty': returning an empty array worked as expected, essentially disabling any additional mime type generation. When I uploaded a WebP only WebP sub-sizes were generated, when I uploaded a JPEG, only JPEG sub-sizes were created.

adamsilverstein (commented):
I noticed some issues in testing that might be worth fixing, please review comment

tests/utils/TestCase/ImagesTestCase.php
eugene-manuilov: Ok, sounds good. Updated.

tests/utils/TestCase/ImagesTestCase.php
eugene-manuilov: Ok, updated.

tests/utils/TestCase/ImagesTestCase.php
eugene-manuilov: Updated.

tests/modules/images/webp-uploads/webp-uploads-test.php
eugene-manuilov: I have added extensions check to image source constraints.

composer.json
eugene-manuilov: Do you mean I need to rename util classes to follow `class-...` pattern? If so, can we make an exception for phpunit classes only?

eugene-manuilov:
> ⚠️ only-jpeg: worked for jpegs, only jpeg was generated. _when I uploaded a WebP it generated webp, I was expecting jpeg_ - not sure of the use case, but was expecting this to work

Hm... it works correctly on my end. When I upload a webp image, all subsizes have been created in the jpeg format. The original image is still in webp though. Is it what you mean, @adamsilverstein?

eugene-manuilov:
> ❌ reverse: didn't work as expected - behavior was the same as 'both', I was expecting the alternate mime type to be stored as the "primary" image, meaning stored in the sizes array, and first in the sources arrays. Not sure this is an issue or if others will expect this behavior as well. Not sure this is completely needed here, this is a little trickier to achieve in the plugin, for the core patch it will be much easier to just start with the first mime type.

@adamsilverstein see https://github.com/WordPress/performance/issues/187#issuecomment-1061277586:

> If the filter is being used in a way so that for a given upload's original MIME type that same MIME type is not part of the transformation array (e.g. a developer decides that no JPEGs should be generated even if a JPEG is uploaded), the first format in the transformation list should be used as the default format to generate the image sizes for. In order to force WordPress core to not generate the original MIME type, the image_editor_output_format needs to be used. A callback should be added to that which calls the webp_uploads_get_supported_image_mime_transforms() function firing the filter and if the original format is not in the list, the filter value should be modified to generate the appropriate other image format. For that purpose and this particular scenario, the original implementation of the WebP Uploads module (see https://github.com/WordPress/performance/pull/32) should essentially be reintroduced.

Basically, what it says is that we need to select the alternate mime type for the "primary" image **only if the original mime type is not included into the transformation list** which is not the case in your situation. Does it make sense to you or I miss something?

composer.json
felixarntz: WordPress core still doesn't use _any_ autoloading, there's always some sort of "includes" file that `require_once`s all files. Obviously not great, but that's how I would do it, for maximum "compatibility" with the core paradigms.

However, since this is purely about PHPUnit classes and we're in this plugin, I'm not opposed to this (obviously better) approach :)

felixarntz (approved):

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -280,31 +257,17 @@ public function it_should_create_a_webp_version_with_all_the_required_properties
...
 		$this->assertStringEndsWith( $metadata['sources']['image/jpeg']['file'], $file );
 		$this->assertFileExists( path_join( $dirname, $metadata['sources']['image/jpeg']['file'] ) );
 		$this->assertSame( $metadata['sources']['image/jpeg']['filesize'], filesize( path_join( $dirname, $metadata['sources']['image/jpeg']['file'] ) ) );
 
-		$this->assertArrayHasKey( 'image/webp', $metadata['sources'] );
-		$this->assertIsArray( $metadata['sources']['image/webp'] );
-		$this->assertArrayHasKey( 'file', $metadata['sources']['image/webp'] );
-		$this->assertArrayHasKey( 'filesize', $metadata['sources']['image/webp'] );
-		$this->assertStringEndsWith( '.webp', $metadata['sources']['image/webp']['file'] );
+		$this->assertImageHasSource( $attachment_id, 'image/webp' );
```
mitogh, line 265:
```suggestion
		$this->assertImageHasSource( $attachment_id, 'image/webp' );
		$this->assertStringEndsWith( '.webp', $metadata['sources']['image/webp']['file'] );
```

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -280,31 +257,17 @@ public function it_should_create_a_webp_version_with_all_the_required_properties
...
-		$this->assertArrayHasKey( 'image/jpeg', $metadata['sizes']['thumbnail']['sources'] );
-		$this->assertArrayHasKey( 'filesize', $metadata['sizes']['thumbnail']['sources']['image/jpeg'] );
-		$this->assertArrayHasKey( 'file', $metadata['sizes']['thumbnail']['sources']['image/jpeg'] );
-		$this->assertArrayHasKey( 'image/webp', $metadata['sizes']['thumbnail']['sources'] );
-		$this->assertArrayHasKey( 'filesize', $metadata['sizes']['thumbnail']['sources']['image/webp'] );
-		$this->assertArrayHasKey( 'file', $metadata['sizes']['thumbnail']['sources']['image/webp'] );
-		$this->assertStringEndsNotWith( '.jpeg', $metadata['sizes']['thumbnail']['sources']['image/webp']['file'] );
-		$this->assertStringEndsWith( '.webp', $metadata['sizes']['thumbnail']['sources']['image/webp']['file'] );
+		$this->assertImageHasSizeSource( $attachment_id, 'thumbnail', 'image/jpeg' );
+		$this->assertImageHasSizeSource( $attachment_id, 'thumbnail', 'image/webp' );
```
mitogh, line 270:
```suggestion
		$this->assertImageHasSizeSource( $attachment_id, 'thumbnail', 'image/webp' );
		$this->assertStringEndsNotWith( '.jpeg', $metadata['sizes']['thumbnail']['sources']['image/webp']['file'] );
		$this->assertStringEndsWith( '.webp', $metadata['sizes']['thumbnail']['sources']['image/webp']['file'] );		
```

mitogh (commented):
Few minor comments, other than that great job :+1:

composer.json
eugene-manuilov: Awesome! Thanks 😄

tests/modules/images/webp-uploads/webp-uploads-test.php
eugene-manuilov: @mitogh the extension check is added to the image source constraint, no need to duplicate it here: https://github.com/WordPress/performance/blob/1babcc91febdd0da517e506e82154268f057e942/tests/utils/Constraint/ImageHasSource.php#L109-L117

tests/modules/images/webp-uploads/webp-uploads-test.php
eugene-manuilov: Same here. The extension is checked in the image source constraint: https://github.com/WordPress/performance/blob/1babcc91febdd0da517e506e82154268f057e942/tests/utils/Constraint/ImageHasSource.php#L109-L117

tests/modules/images/webp-uploads/webp-uploads-test.php
mitogh: Great, thanks for clarify.

mitogh (approved):

Issue #186, opened 2022-02-22
Select the smaller image size from the list of available sources


mitogh:
## Context

Every image metadata would contain a `sources` property for every single size available. The property is an associative array where each key on the array represents a mime type and each value on the array would be an associative array with 2 properties on it, the `file` and `filesize` of each mime type. For instance, this would like the following data for the `medium` size 

```php
// slice of a metadata from an image  stored in `_wp_attachment_metadata`
array(
  // ...additional image sizes
  'medium'  => array (
    'file' => '7rMFIAVGlVg-240x300.jpeg'
    'width' => 240,
    'height' => 300,
    'mime-type' => 'image/jpeg',
    'sources' => array(
      'image/jpeg' => array(
        'file' => '7rMFIAVGlVg-240x300.jpeg',
        'filesize' => 3831,
      ),
      'image/webp' => array(
        'file' => '7rMFIAVGlVg-240x300.webp',
        'filesize' => 2016
      )
    ),
  )
/// ...additional image sizes
)
```

The size of each image is expressed in `bytes` as returned by the [`filesize`](https://www.php.net/manual/en/function.filesize.php) function.

## Feature Description

Based on the work from:

- #142 
- #149 
- #8 

And considering the scenario where multiple mime-types coexists for the same image, load the image with the smaller size instead of a specific format. As there might be a case where a specific image compresses better in JPEG than WebP (or if another format is available like AVIF in those cases make sure that the image that is delivered to the client is the image with the smaller file size instead of defined image size. 

### Logic to deliver an image to the client

- If sources are available 
- order sources by filesize
- deliver the first one in the list if is supported otherwise move to the next one on the list until a supported format is found.
- otherwise deliver the former image for the specified size.

mitogh:
## Background

The information can be found on the metadata of an image using `wp_get_attachment_metadata` or directly reading the meta key `_wp_attachment_metadata`.

All image sizes contain a new property, named `sources`, this property is an array where each key on the array represents a mime for the image created for that particular size and mime. Each mime is stored as an array with 2 keys:

- file
- filesize

For the `full` image size the property can be found at the root level of the metadata of the file:

See: 

- #147 
- #194 
- #142 
- #174 

## Suggested approach

[...]

PR #230, opened 2022-03-15
Allow developers to select which image format to use for images in the content

### LINKED ISSUE #187: Allow developers to tweak which image formats should be used

eugene-manuilov:
## Summary

Fixes #187 

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -554,11 +554,31 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array The list of mime types that can be used to update images in the content.
+	 */
+	$target_mimes = apply_filters( 'webp_uploads_content_image_mimes', array( 'image/jpeg', 'image/webp' ) );
+
+	$target_mime  = null;
+	$target_mimes = array_reverse( $target_mimes );
+	foreach ( $target_mimes as $mime ) {
```
mitogh, line 617: Introducnig a `for` loop instead a `foreach`  helps to avoid the need for a reverse operation, and provides a more natural flow.

```suggestion
    for ( $i = count( $target_mimes ) - 1; $i >= 0; $i-- ) {
      $mime = $target_mimes[ $i ];
```

mitogh (commented):
Minor comment about the code.

Would be great to introduce some tests to test cases like:

- Default behavior
- Reverse behavior
- Empty value
- Unsupported mime types

modules/images/webp-uploads/load.php
```diff
@@ -554,11 +554,31 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+		if ( isset( $metadata['sources'][ $mime ] ) ) {
+			$target_mime = $mime;
+			break;
+		}
+	}
+
+	if ( null === $target_mime ) {
+		return $image;
+	}
```
felixarntz, line 626: We also need to add a condition to short-circuit somehow if the target MIME is the same like the extensions we're replacing (i.e. JPEG).

We can later add another filter to adjust those extensions as well.

felixarntz (commented):
@eugene-manuilov Looks great, just one comment.

modules/images/webp-uploads/load.php
adamsilverstein: I would still prefer to see this work in the reverse order: the first available mime type image in the list would be used instead of the last available type.

adamsilverstein:
Overall looks good although I still prefer to have the filter operate in the opposite order. Also, it would be good to add some tests verifying that the filter works as expected.

felixarntz:
@adamsilverstein 
> Overall looks good although I still prefer to have the filter operate in the opposite order.

The benefit of how it currently behaves is that the order of how items have to be sorted in the filter itself is arguably more aligned with the other filter, with the "foundational" format coming first (typically JPEG for our purposes currently). The reason we then iterate through the array backwards here is because the "priority order" of the formats would be actually the opposite, which especially makes sense when thinking about it in the context of future `picture` element support. Basically, the most advanced modern format should come first, but it's also the most likely format to not be supported by the browser. So there have to be fallbacks, all the way back to the "foundational" format which is expected to work everywhere.

> Also, it would be good to add some tests verifying that the filter works as expected.

+1 to that, @eugene-manuilov could you add a few tests for the filter usage?

adamsilverstein:
> which especially makes sense when thinking about it in the context of future picture element support. 

It actually makes **less** sense in the picture context where the `sources` array is stored in the order you want the browser to use the image type. For example, if you want to use AVIF and fall back to WEBP and then JPEG, you would put the AVIF sources array first, the WEBP array second and finally put the jpeg in the contained img element.

The way we have it now, we will need to reverse the order before building the picture element. The fact that we have to reverse the array every time we use it is surely a sign we have the data in the wrong order?

The image formats should be listed in the order in which you prefer to use the formats, not the reverse! 

>  is arguably more aligned with the other filter

I see these as two separate filters - one is a map of arrays and one is an array.

Since we can't seem to agree on this point, can we put it to a vote? I'll accept the majority view.

modules/images/webp-uploads/load.php
```diff
@@ -554,11 +554,31 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
 		return $image;
 	}
 
-	$urls = $matches[0];
-	// TODO: Add a filterable option to change the selected mime type. See https://github.com/WordPress/performance/issues/187.
-	$target_mime = 'image/webp';
+	/**
+	 * Filters mime types that should be used to update all images in the content. The order of
+	 * mime types matters. The last mime type in the list will be used if it is supported by an image.
```
adamsilverstein, line 604: If we keep the reverse order this doc block is confusing. maybe
"The last mime type in the list that is supported by the system will be used"

felixarntz:
@adamsilverstein I see your point around reversing the array, I agree that is not optimal. At the same time, my personal take is that it makes sense to have the orders for the two filters behave similarly so that it's easier to remember for developers (if we get rid of the `array_reverse`, it would mean e.g. that the other filter from #227 would need to have JPEG first and this one here would need to have JPEG last).

> Since we can't seem to agree on this point, can we put it to a vote? I'll accept the majority view.

Definitely, that may make sense, maybe even with some further discussion. I'm also happy to be outvoted. :) However, I'd suggest we do this in a follow-up issue to not block this PR which would be crucial to get into the next beta release. Would it be okay with you to for now go with the current approach, and open an issue to follow up on the discussion here? We would then put that into the next beta milestone, which would give us some time to discuss and have a vote.

modules/images/webp-uploads/load.php
mitogh: > I would still prefer to see this work in the reverse order: the first available mime type image in the list would be used instead of the last available type.

I see the benefit of that approach as well as is clear and more obvious by just looking at the order of the provided array.

eugene-manuilov: @mitogh, unfortunately, your `for` loop will stop working as soon as someone unsets a mime type in a filter hook:

```php
add_filter( 'webp_uploads_content_image_mimes', function ( $types ) {
    $types[] = 'image/avif';
    return $types;
}, 10 );

add_filter( 'webp_uploads_content_image_mimes', function ( $types ) {
    unset( $types[ array_search( 'image/webp', $types, true ) ] );
    return $types;
}, 15 );

...

for ( $i = count( $target_mimes ) - 1; $i >= 0; $i-- ) {
    $mime = $target_mimes[ $i ]; // --> "PHP Warning:  Undefined array key 1 ..." - because the $target_mimes is "array( 0 => 'image/jpeg', 2 => 'image/avif' )
}
```

modules/images/webp-uploads/load.php
eugene-manuilov: @felixarntz we already have it down the road: https://github.com/WordPress/performance/blob/f5465574d9640cdb5694f61ecf75a3d97c15b7f1/modules/images/webp-uploads/load.php#L575-L579

felixarntz:
Let's follow up on the filter order discussion in https://github.com/WordPress/performance/issues/187#issuecomment-1069976158, it's easier to follow on the issue than on this PR.

modules/images/webp-uploads/load.php
felixarntz: Ah great, I missed that.

modules/images/webp-uploads/load.php
felixarntz: Yeah, we would need to call `array_values` to make that reliable, so then it's also an extra array function call, so probably not better than `array_reverse`.

modules/images/webp-uploads/load.php
```diff
@@ -554,11 +554,31 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array The list of mime types that can be used to update images in the content.
+	 */
+	$target_mimes = apply_filters( 'webp_uploads_content_image_mimes', array( 'image/jpeg', 'image/webp' ) );
+
+	$target_mime  = null;
+	$target_mimes = array_reverse( $target_mimes );
```
felixarntz, line 616: Maybe we add a comment here for now to explain this:
```suggestion
	// Look for the most progressive image format first.
	$target_mimes = array_reverse( $target_mimes );
```

felixarntz (approved):
Just one tiny recommendation, but nothing blocking.

modules/images/webp-uploads/load.php
mitogh: > unfortunately, your for loop will stop working as soon as someone unsets a mime type in a filter hook:

If an element is unset from an an array the `count` value would still return the number:

```php
// Count => 3
$types = array( 'image/jpeg', 'image/webp', 'image/avif');

unset( $types[ array_search( 'image/webp', $types, true ) ] );
// After unset
count( $types ) // Returns 2
// So the for loop will start from N - 1 in this case 1 and 0 indexes 
// so even when using the unset the for loop will still work as expected.
```

Not sure if I am missing something, thanks.

felixarntz: @mitogh The count is correct still, but the indexes aren't. The `for` loop would for your example go from index `1` to `0`. However, because index `1` (`image/webp`) was removed from the array, `$target_mimes[1]` is undefined, causing a problem here. And it would skip `$target_mimes[2]` which is still there as `image/avif`.

mitogh: Woops, my bad!

In that case using iterators is another option.

```php
end( $target_mimes );
for ( $i = count( $target_mimes ) - 1; $i >= 0; $i-- ) {
    $mime = current( $target_mimes );
    prev($target_mimes)
}
```

modules/images/webp-uploads/load.php
```diff
@@ -554,11 +554,31 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
-	$target_mime = 'image/webp';
+	/**
+	 * Filters mime types that should be used to update all images in the content. The order of
+	 * mime types matters. The last mime type in the list will be used if it is supported by an image.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array The list of mime types that can be used to update images in the content.
+	 */
+	$target_mimes = apply_filters( 'webp_uploads_content_image_mimes', array( 'image/jpeg', 'image/webp' ) );
```
mitogh: This looks great but including the `$attachment_id` would allow to include custom logic like:

- #186  

And the `$context` to be aligned with other content functions.

```suggestion
	$target_mimes = apply_filters( 'webp_uploads_content_image_mimes', array( 'image/jpeg', 'image/webp' ), $attachment_id, $context );
```

cc @jjgrainger

modules/images/webp-uploads/load.php
eugene-manuilov: Updated.

modules/images/webp-uploads/load.php
eugene-manuilov: @mitogh why do you think that using the `for` loop and a bunch of iteration function calls are better than just the `foreach` loop and a single `array_reverse` call?

eugene-manuilov:
@adamsilverstein @felixarntz an additional test is added. Not sure if we need more tests because the majority of cases are already covered by existing tests.

modules/images/webp-uploads/load.php
mitogh: > why do you think that using the for loop and a bunch of iteration function calls are better than just the foreach loop and a single array_reverse call?

Not necessarily better but most performant due the avoiding the need to call `array_reverse` which is:

- N more in time to reverse
- N more in space to store the reverse of operation.

When the `for` is always `N` due no additional space or operation is required, I know this might be just 3 at the most elements but any cumulative gain would help in the long run.

@eugene-manuilov

eugene-manuilov: @mitogh, this performance optimisation doesn't make any sense for an array with a few items because even if there a performance gain (which is arguable) it will be scarce.

If you want to improve performance, I would encourage you to offer a replacement for all those preg_match calls to find images in the content which has much more significant impact on performance, especially for big posts with much content.

As to complexity of the algorithm, you need to take into account that function calls are expensive in PHP (you can read about it [here](https://stackoverflow.com/questions/3691625/why-are-php-function-calls-so-expensive)). So, sometimes it is easier and better to call one function with N complexity to avoid running an algorithm that has 2N+2 function calls.

felixarntz:
@eugene-manuilov Updated this PR to be against the `1.0.0-beta.2` branch.

modules/images/webp-uploads/load.php
mitogh: > this performance optimisation doesn't make any sense for an array with a few items because even if there a performance gain (which is arguable) it will be scarce.

Still hard to see why does not make sense having a performance gain.

> If you want to improve performance, I would encourage you to offer a replacement for all those preg_match calls to find images in the content which has much more significant impact on performance, especially for big posts with much content.

Under that premise we might need to go back into the WP Core and update many of those used to insert the `srcset` and other attributes, I think we need to take a look into a more iterative approach rather than a more drastic one.

Overall my main point is if we can optimize by just replacing `foreach` and an `array_reverse` would go for it but happy to move the code as it is.

eugene-manuilov: > Overall my main point is if we can optimize by just replacing foreach and an array_reverse would go for it but happy to move the code as it is.

IMO readability here worth more than a performance optimisation that makes the script run faster by ~0.0001ms. I don't mind changing it if we have a consensus here. @felixarntz @adamsilverstein what do you think?

felixarntz: I'd be in favor for keeping this as is - we're speaking about a micro optimization here, and it's unclear whether the iterator approach is actually faster than `array_reverse`. The code is more familiar to read the way it is now with the `foreach` loop on the reversed array.

We shouldn't block this issue because of this minor point, so I'd say let's go with what we have now.

mitogh (approved):

adamsilverstein (approved):

PR #231, opened 2022-03-15
Add plugin banner and icon assets

### LINKED ISSUE #144: Performance Lab: Branding
## Performance Lab: Branding
As briefly discussed in our [Weekly Performance Meeting from 1st February 2022](https://make.wordpress.org/core/2022/02/01/performance-team-meeting-summary-1-february-2022/)

@felixarntz brought up an item with regards to design some assets for the "Performance Lab" plugin.
As we're no stranger to opinion in the WP community, it would be great to get everyones ideas / thoughts / opinions in terms of what they would like to see in terms of creative options.

Some initial ideas / inspiration from my side:

1. Jetpack Boost has a nice feel to it: https://wordpress.org/plugins/jetpack-boost/ - I also like the fact that it has a catch phrase.
2. Coincidently, this Dribbble Shot: https://dribbble.com/shots/16485339-Axelerated-Software kinda incorporates the JetPack logo, with WordPress.org colours which is actually really nice.
3. We could also go with some a bit more 3D: https://dribbble.com/shots/17317377-Athens-Medica - 3D is trending quite wildly across the web at the moment, with more skeumorphic design making a come back. 

Considering this plugin will eventually land up in core, I think retain the default WordPress.org colours is a must, with a nice logon / monogram could be quite effective.

Lets get the discussion going :)

felixarntz:
## Summary

Fixes #144

## Relevant technical choices

* Assets in the new `.wordpress-org` folder will automatically deployed with every release tag via our GitHub workflow.

## Checklist


jeffpaul (changes requested):
you'll also want to add the `.wordpress-org` directory to the `.gitattributes` file

jeffpaul:
Also, please ensure that you prop @rwlands and any others from the source issue so that we're properly crediting folks for this effort

mitogh (commented):
My only concern is the SVG icon which is almost `8 MB` in size, it seems like the SVG is including nonrequired stuff. It would be great if we can have a smaller size created out of geometric shapes instead of actual images.

felixarntz:
@jeffpaul 
> you'll also want to add the `.wordpress-org` directory to the `.gitattributes` file

Ah right, that makes sense from a GitHub perspective. Just double-checking, will the 10up deploy action still consider the directory for the asset deployment though if it's "ignored" via `.gitattributes`?

> Also, please ensure that you prop @rwlands and any others from the source issue so that we're properly crediting folks for this effort

Absolutely, however right now we have no process at all for any props for this plugin. This would be mostly relevant for eventual WP core merges for sure, but how about during plugin development? How do other feature plugins handle that? If there's something we can do here I'd say let's open an issue to put the necessary guidelines / tooling in place.

@mitogh 
> My only concern is the SVG icon which is almost `8 MB` in size, it seems like the SVG is including nonrequired stuff. It would be great if we can have a smaller size created out of geometric shapes instead of actual images.

Good catch! Can you potentially look into that @rwlands? Indeed the SVG file is extremely large at the moment.

jeffpaul:
> Ah right, that makes sense from a GitHub perspective. Just double-checking, will the 10up deploy action still consider the directory for the asset deployment though if it's "ignored" via .gitattributes?

from the action readme, which also matches my experience with the action on 10up's plugins:
> This Action commits the contents of your Git tag to the WordPress.org plugin repository using the same tag name. It can exclude files as defined in either .distignore or .gitattributes, and moves anything from a .wordpress-org subdirectory to the top-level assets directory in Subversion (plugin banners, icons, and screenshots).

> Absolutely, however right now we have no process at all for any props for this plugin. This would be mostly relevant for eventual WP core merges for sure, but how about during plugin development? How do other feature plugins handle that? If there's something we can do here I'd say let's open an issue to put the necessary guidelines / tooling in place.

Ah right, yeah I suppose we should get something in place soon before it becomes unwieldy to determine who to prop if (🤞🏼 when 🤞🏼) the plugin merges to core.  10up tends to utilize a CREDITS.md file ([example](https://github.com/10up/ElasticPress/blob/develop/CREDITS.md)) to keep track of those folks across versions as well as proping them on changelog items.  I could help go back through things already merged and gather those props into the changelog and/or credits file, but curious which approach (or both) you would prefer to help track things ahead of a core merge?

felixarntz:
@jeffpaul 
> from the action readme, which also matches my experience with the action on 10up's plugins

Sounds good, updated!

> Ah right, yeah I suppose we should get something in place soon before it becomes unwieldy to determine who to prop if (🤞🏼 when 🤞🏼) the plugin merges to core. 10up tends to utilize a CREDITS.md file ([example](https://github.com/10up/ElasticPress/blob/develop/CREDITS.md)) to keep track of those folks across versions as well as proping them on changelog items. I could help go back through things already merged and gather those props into the changelog and/or credits file, but curious which approach (or both) you would prefer to help track things ahead of a core merge?

A credits file would help, however I think ours may need to be more granular than the one you've linked. There won't be a moment where "the plugin merges to core", since this is multiple WP core performance features in different stages of development, and they would merge individually when they're ready respectively. So to best support this workflow we would probably need some sort of credits file that is broken down by the different modules in this plugin - which is still possible, but a bit more involved.

Would you mind creating an issue so we can explore this further?

jeffpaul:
@felixarntz to further clarify my understanding before opening the issue, perhaps a credits file within the sub-directories in https://github.com/WordPress/performance/tree/trunk/modules would suffice for your concern?

Otherwise this PR looks ✅ to me.

jeffpaul (approved):
Hooray, beautiful branding!

felixarntz:
@jeffpaul 
> @felixarntz to further clarify my understanding before opening the issue, perhaps a credits file within the sub-directories in https://github.com/WordPress/performance/tree/trunk/modules would suffice for your concern?

That's an alternative option, yes. I think we should discuss that in the eventual issue, whether it's preferable to have
* either a single `CREDITS.md` file in the repository root, with contributors broken down per module
* or a `CREDITS.md` file for every module in the module's directory

felixarntz:
@mitogh I just removed the `icon.svg` for now since it's too large as you've pointed out. I'd say we can already proceed with the other banner and icon assets since they are the essentials of what's needed really. Many plugins do not provide any `icon.svg`, so it's okay to add it later I'd say. This way we can still release with the new assets next week.

rwlands:
Hi there, I can look into the svg size issue, that's certainly very overweight!

rwlands:
Icon_Black
Icon_Blue
Icon_White

mitogh (approved):

mitogh:
Sounds good @felixarntz looks like a new SVG was added tho in case we want to add those as well.

Thank you @rwlands for getting these SVG icons for us, great work :tada:

mitogh (approved):

felixarntz:
Thanks so much for fixing the SVG @rwlands! This is now good to be merged 🎉

PR #233, opened 2022-03-17
Select the smaller image size from the list of available sources

### LINKED ISSUE #186: Select the smaller image size from the list of available sources

jjgrainger:
## Summary

Fixes #186 Select the smaller image size from the list of available sources

## Checklist

### DIFF
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -603,3 +603,66 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
 }
 
 add_filter( 'the_content', 'webp_uploads_update_image_references', 10 );
+
+/**
+ * Returns the attachment sources array ordered by filesize.
+ *
+ * @since n.e.xt
+ * @private
+ *
+ * @param int    $attachment_id The attachment ID.
+ * @param string $size          The attachment size.
+ * 
+ * @return array The attachment sources array.
+ */
+function webp_uploads_get_attachment_sources( $attachment_id, $size = 'thumbnail' ) {
+	// Check for the sources attribute in attachment metadata.
+	$metadata = wp_get_attachment_metadata( $attachment_id );
+
+	// Return full image size sources.
+	if ( 'full' === $size && ! empty( $metadata['sources'] ) ) {
+		return $metadata['sources'];
+	}
+
+	// Return the resized image sources.
+	if ( isset( $metadata['sizes'][ $size ]['sources'] ) && ! empty( $metadata['sizes'][ $size ]['sources'] ) ) {
+		return $metadata['sizes'][ $size ]['sources'];	
+	}
+
+	// Return an empty array if no sources found.
+	return [];
+}
+
+/**
+* Returns the attachment smallest filesize source.
+ *
+ * @since n.e.xt
+ * @private
+ *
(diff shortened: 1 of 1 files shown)


modules/images/webp-uploads/load.php
```diff
@@ -603,3 +603,78 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+	// Check for the sources attribute in attachment metadata.
+	$metadata = wp_get_attachment_metadata( $attachment_id );
+
+	// Return full image size sources.
+	if ( 'full' === $size && ! empty( $metadata['sources'] ) ) {
+		return $metadata['sources'];
+	}
+
+	// Return the resized image sources.
+	if ( isset( $metadata['sizes'][ $size ]['sources'] ) && ! empty( $metadata['sizes'][ $size ]['sources'] ) ) {
```
mitogh, line 628: In this case `empty` works simiarly to `isset` so you can avoid the `isset` call.

```suggestion
	if ( ! empty( $metadata['sizes'][ $size ]['sources'] ) ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -603,3 +603,78 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+		return $metadata['sizes'][ $size ]['sources'];
+	}
+
+	// Return an empty array if no sources found.
+	return [];
```
mitogh, line 633: We are using the long syntax array for arrays due to WP styleguide.

```suggestion
	return array();
```

jjgrainger:
Thanks @mitogh 

As you suggested I've created a new PR that branches from other work this expands on.

I've addressed the feedback in the new PR #243 

Closing this.

PR #235, opened 2022-03-18
Generate secondary image MIME types when editing original image

### LINKED ISSUE #158: Perform updates on secondary image mime types when original image is edited
## Scenario

When an image is uploaded into WordPress at least 2 mime types would be created according to with:

- #142 

After the image completes the update into the media library, the image would be stored using at least 2 mime types, for this scenario let's use: 

- `image/webp`
- `image/jpeg` 

_For simplicity, we are going to call `webp` the primary image format and `jpeg` the secondary image format._

After an image is uploaded edits can be performed against this image, when using the image editor (see attached image) or from within the Gutenberg editor. 

- rotation
- flip
- crop

After an edit is performed in the original image the original sizes are stored inside of  `_wp_attachment_backup_sizes` and the files preserved to allow a restore of the files in case required. Subsequent edits (if more than a single update is performed) would be stored in the `_wp_attachment_backup_sizes` keeping `_wp_attachment_metadata` data with the last edit so the image always reflect the last action performed by the user. 

### Recreate image in secondary image format for every edit

For every edit performed on the original image, a primary image format would be stored (WordPress default behavior).

So the questions come into place: Should the secondary image format recreate the same edit format and keep the original secondary format the same way it happens with the primary image format. 

This behavior is consistent with the primary image format. In case an image is requested with the secondary format the image would be displayed in the same way as the primary image format.(with the edits applied to the original image)

## Notes

- Image edits should include the suffix `e-$id` where `$id` is a `13` digit number specifically `time() . rand( 100, 999 )` [created by WP ](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/includes/image-edit.php#L708) to reference any additional edit that is no longer relevant, the file name is important so WP can identify images that reference to edits and can be removed accordingly based on the definition and value of `IMAGE_EDIT_OVERWRITE` This suffix is also included as part of the image size name.

kirtangajjar:
## Summary
Fixes #158. 

This PR will create images of secondary mime type when original image is edited.

Logic to apply change to image of all sizes is inspired from [wp_save_image](https://github.com/WordPress/wordpress-develop/blob/22167a2d805254067162d3e5c3b168a028a924b5/src/wp-admin/includes/image-edit.php#L939).

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,84 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ *
+ * Updates webp image when original image is edited
+ *
+ * @since n.e.x.t
+ *
+ * @param bool|null       $override  Value to return instead of saving. Default null.
+ * @param string          $file      Name of the file to be saved.
+ * @param WP_Image_Editor $image     The image editor instance.
+ * @param string          $mime_type The mime type of the image.
+ * @param int             $post_id   Attachment post ID.
```
mitogh, line 744: Looks like we are missing a `@return` doc for it.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,84 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	$target                     = ! empty( $_REQUEST['target'] ) ? preg_replace( '/[^a-z0-9_-]+/i', '', $_REQUEST['target'] ) : '';
+	$nocrop                     = false;
+	$current_mime_type          = get_post_mime_type( $post_id );
+
+	foreach ( $valid_mime_transforms[ $mime_type ] as $targeted_mime ) {
+		if ( $targeted_mime === $current_mime_type ) {
+			continue;
+		}
+
+		$allowed_mimes = array_flip( wp_get_mime_types() );
```
mitogh: This operation can be placed outside of the `foreach` loop in order to avoid calling both functions multiple times with no additional reason as the value of `allowed_mimes` would be the same.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,84 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+		if ( ! wp_image_editor_supports( array( 'mime_type' => $targeted_mime ) ) ) {
+			return new WP_Error( 'image_mime_type_not_supported', __( 'The provided mime type is not supported.', 'performance-lab' ) );
+		}
+
+		$extension   = explode( '|', $allowed_mimes[ $targeted_mime ] );
+		$destination = trailingslashit( $original_directory ) . "{$filename}.{$extension[0]}";
+
+		$image->save( $destination, $targeted_mime );
+
+		$new_image = wp_get_image_editor( $destination );
```
mitogh: Make sure `$new_image` is not an instance of `WP_Error`

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,84 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+		if ( 'thumbnail' === $target ) {
+			$nocrop = true;
+		}
+
+		if ( isset( $sizes ) ) {
```
mitogh: An early return avoids an additional nesting in the code.

```suggestion
		if ( ! isset( $sizes ) ) {
		  continue;
		}
```

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,84 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+			foreach ( $sizes as $size ) {
+
+				if ( isset( $_wp_additional_image_sizes[ $size ] ) ) {
+					$width  = (int) $_wp_additional_image_sizes[ $size ]['width'];
+					$height = (int) $_wp_additional_image_sizes[ $size ]['height'];
+					$crop   = ( $nocrop ) ? false : $_wp_additional_image_sizes[ $size ]['crop'];
+				} else {
+					$height = get_option( "{$size}_size_h" );
+					$width  = get_option( "{$size}_size_w" );
```
mitogh: Make sure to cast those values as integers as well as an option is a string by default.

```suggestion
					$height = (int) get_option( "{$size}_size_h" );
					$width  = (int) get_option( "{$size}_size_w" );
```

mitogh (commented):
Hey, @kirtangajjar thank you for getting this initial PR done, one thing I'm able to identify by looking at the code is that this function is creating the additional mime type images but is not storing those images back into the metadata of the image, we need to make sure the `sources` property is populated back similarly as how it happens when an image is uploaded. 

The `sources` property is stored inside of each file size containing all the available mime types. 

Let me know if that makes sense or if you have any additional questions about that one.

mxbclang:
@kirtangajjar FYI I've removed the 1.0.0-beta.2 milestone from this PR since it won't be ready for Monday's release. Thanks!

mitogh:
Hey @kirtangajjar let me know in case you have any additional questions regarding the initial feedback or suggested approach, happy to provide some direction or give you a hand if you need it, just let me know. 

Great work so far :tada:

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
 	return rest_ensure_response( $data );
 }
 add_filter( 'rest_prepare_attachment', 'webp_uploads_update_rest_attachment', 10, 3 );
+
+/**
+ *
+ * Updates webp image when original image is edited
```
felixarntz: This isn't specific to WebP, it will generate image files for any formats specified in the filter from `webp_uploads_get_upload_image_mime_transforms()`.
```suggestion
 * Creates additional image formats when original image is edited.
```

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	}
+
+	$valid_mime_transforms      = webp_uploads_get_upload_image_mime_transforms();
+	$original_directory         = pathinfo( $file, PATHINFO_DIRNAME );
+	$filename                   = pathinfo( $file, PATHINFO_FILENAME );
+	$sizes                      = get_intermediate_image_sizes();
+	$_wp_additional_image_sizes = wp_get_additional_image_sizes();
+	$target                     = ! empty( $_REQUEST['target'] ) ? preg_replace( '/[^a-z0-9_-]+/i', '', $_REQUEST['target'] ) : '';
+	$nocrop                     = false;
+	$current_mime_type          = get_post_mime_type( $post_id );
```
felixarntz: This isn't necessary, the MIME type is already passed from the filter (see the `$mime_type` parameter of the function).

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+function webp_uploads_update_image_onchange( $override, $file, $image, $mime_type, $post_id ) {
+	if ( $override !== null ) {
+		return $override;
+	}
+
+	$valid_mime_transforms      = webp_uploads_get_upload_image_mime_transforms();
+	$original_directory         = pathinfo( $file, PATHINFO_DIRNAME );
+	$filename                   = pathinfo( $file, PATHINFO_FILENAME );
+	$sizes                      = get_intermediate_image_sizes();
+	$_wp_additional_image_sizes = wp_get_additional_image_sizes();
```
felixarntz: This can be simplified by reusing WP core's [`wp_get_registered_image_subsizes()` function](https://developer.wordpress.org/reference/functions/wp_get_registered_image_subsizes/), which already has both of these combined.

This also allows removing some of the logic in the below `foreach` loop.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	$target                     = ! empty( $_REQUEST['target'] ) ? preg_replace( '/[^a-z0-9_-]+/i', '', $_REQUEST['target'] ) : '';
+	$nocrop                     = false;
+	$current_mime_type          = get_post_mime_type( $post_id );
+	$allowed_mimes              = array_flip( wp_get_mime_types() );
+
+	foreach ( $valid_mime_transforms[ $mime_type ] as $targeted_mime ) {
+		if ( $targeted_mime === $current_mime_type ) {
```
felixarntz: See above.
```suggestion
		if ( $targeted_mime === $mime_type ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+		return $override;
+	}
+
+	$valid_mime_transforms      = webp_uploads_get_upload_image_mime_transforms();
+	$original_directory         = pathinfo( $file, PATHINFO_DIRNAME );
+	$filename                   = pathinfo( $file, PATHINFO_FILENAME );
+	$sizes                      = get_intermediate_image_sizes();
+	$_wp_additional_image_sizes = wp_get_additional_image_sizes();
+	$target                     = ! empty( $_REQUEST['target'] ) ? preg_replace( '/[^a-z0-9_-]+/i', '', $_REQUEST['target'] ) : '';
+	$nocrop                     = false;
```
felixarntz: These two variables aren't necessary anymore when using `wp_get_registered_image_subsizes()`, see above.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+		$new_image = wp_get_image_editor( $destination );
+
+		if ( is_wp_error( $new_image ) ) {
+			return $new_image;
+		}
+
+		if ( 'thumbnail' === $target ) {
+			$nocrop = true;
+		}
```
felixarntz: See above, this can be removed when using `wp_get_registered_image_subsizes()`.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+		if ( ! isset( $sizes ) ) {
+		  continue;
+		}
```
felixarntz: This is unnecessary and can be removed. `$sizes` will always be set.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+				$width  = (int) get_option( "{$size}_size_w" );
+				$crop   = ( $nocrop ) ? false : get_option( "{$size}_crop" );
+			}
+
+			$_sizes[ $size ] = array(
+				'width'  => $width,
+				'height' => $height,
+				'crop'   => $crop,
+			);
+		}
```
felixarntz: See above, this whole block of code isn't needed when using `wp_get_registered_image_subsizes()`. We can use the `$sizes` from `wp_get_registered_image_subsizes()` as is.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+			continue;
+		}
+
+		if ( ! isset( $allowed_mimes[ $targeted_mime ] ) || ! is_string( $allowed_mimes[ $targeted_mime ] ) ) {
+			return new WP_Error( 'image_mime_type_invalid', __( 'The provided mime type is not allowed.', 'performance-lab' ) );
+		}
+
+		if ( ! wp_image_editor_supports( array( 'mime_type' => $targeted_mime ) ) ) {
+			return new WP_Error( 'image_mime_type_not_supported', __( 'The provided mime type is not supported.', 'performance-lab' ) );
+		}
```
felixarntz: We cannot return `WP_Error`s from here, we have to pass through the original `$override` value in any case. `WP_Error` as return values must only be used in separate helper functions we would call from here.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+		$extension   = explode( '|', $allowed_mimes[ $targeted_mime ] );
+		$destination = trailingslashit( $original_directory ) . "{$filename}.{$extension[0]}";
+
+		$image->save( $destination, $targeted_mime );
```
felixarntz: We need to check that this is not a `WP_Error` before proceeding.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+		$new_image = wp_get_image_editor( $destination );
+
+		if ( is_wp_error( $new_image ) ) {
+			return $new_image;
+		}
```
felixarntz: Is this really needed? Why can't we reuse the previous editor instance in `$image`? The `save()` method called above already sets the file and MIME type to that new file anyway, so recreating is unnecessary.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+			}
+
+			$_sizes[ $size ] = array(
+				'width'  => $width,
+				'height' => $height,
+				'crop'   => $crop,
+			);
+		}
+
+		$new_image->multi_resize( $_sizes );
```
felixarntz: I thought at first this won't work, but looking at the WP core code in more depth it indeed will. Even without passing the MIME type, the method will reuse the MIME type previously set, which will happen on editor instantiation _and_ when calling the `save()` method.

That is a great find @kirtangajjar! 🎉 

cc @adamsilverstein @mitogh We can probably simplify _a lot_ of our implementation so far by using this approach.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	}
+}
+
+add_filter( 'wp_save_image_editor_file', 'webp_uploads_update_image_onchange', 10, 5 );
```
felixarntz: Let's put the filter calls right below their corresponding function.
```suggestion
add_filter( 'wp_save_image_editor_file', 'webp_uploads_update_image_onchange', 10, 5 );
```

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,91 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+			$_sizes[ $size ] = array(
+				'width'  => $width,
+				'height' => $height,
+				'crop'   => $crop,
+			);
+		}
+
+		$new_image->multi_resize( $_sizes );
+	}
```
felixarntz: As @mitogh mentioned, the one critical thing that is missing here is to update the metadata with the new sources.

You will need to store the return value from the above `save()` method call under the root `$metadata['sources']`, and you will need to store the return values from the above `multi_resize()` method call under each respective `$metadata[ $size ]['sources']`.

Since this hook cannot modify those, I'd say it would be best to add a separate `add_filter` for `wp_update_attachment_metadata` _within_ this method, so that it only gets hooked in once we've created these images. For example:
```php
add_filter(
	'wp_update_attachment_metadata',
	function( $metadata, $meta_post_id } use ( $post_id, $orig_source, $size_sources ) { // Reuse the same values for the new images generated above.
		if ( $meta_post_id !== $post_id ) { // Check to ensure we're working with the right post.
			return $metadata:
		}
		// New function to add, which expands/updates the given metadata with the additional sources passed.
		return webp_uploads_update_sources( $metadata, $orig_source, $size_sources );
	},
	10,
	2
);
```

Then, in the separate hook callback, update the metadata accordingly before it is saved.

felixarntz (commented):
@kirtangajjar This looks like a good start, and kudos for finding out we can just use `multi_resize()` without even passing the MIME type to it!

There are a couple of things to simplify though, and most importantly this is missing the part of storing metadata about the new image, as mentioned by @mitogh before.

modules/images/webp-uploads/load.php
kirtangajjar: Thanks for this. The code has become much simpler because of this 🥳

modules/images/webp-uploads/load.php
kirtangajjar: Oh yes, I didn't know `save()` method could change mime type. Thanks a lot!

modules/images/webp-uploads/load.php
kirtangajjar: @felixarntz You mean before saving? I have added one condition after it.

modules/images/webp-uploads/load.php
kirtangajjar: @felixarntz Actually I was able to figure out the adding `wp_update_attachment_metadata` filter part myself, but I was still looking for an optimal way to update metadata. Thankfully your in this PR made the code much simpler and so I was able to finish the rest of updating metadata part too!

kirtangajjar:
@mitogh @felixarntz Thank you very much for your support on this PR. Learnt a ton from this one. I have addressed the feedback here. LMK if it looks good to you so I can mark this PR as "Ready to review"

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,99 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ * @param int             $post_id   Attachment post ID.
+ *
+ * @return bool|null
+ */
+function webp_uploads_update_image_onchange( $override, $file, $image, $mime_type, $post_id ) {
+	if ( null !== $override ) {
+		return $override;
+	}
+
+	$valid_mime_transforms = webp_uploads_get_upload_image_mime_transforms()[ $mime_type ];
```
mitogh: We need to make sure that the `$mime_type` exists in the array before accessing it.

```suggestion
    if ( empty( webp_uploads_get_upload_image_mime_transforms()[ $mime_type ] ) {
      return $overrride;
    }

	$valid_mime_transforms = webp_uploads_get_upload_image_mime_transforms()[ $mime_type ];
```

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,99 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	$original_directory = pathinfo( $file, PATHINFO_DIRNAME );
+
+	foreach ( $metadata['sizes'] as $size => $size_details ) {
+		foreach ( $valid_mime_transforms as $targeted_mime ) {
+			$extension            = explode( '|', $allowed_mimes[ $targeted_mime ] );
+			$filename_without_ext = implode( explode( '.', $size_details['file'], -1 ) );
+			$newfile              = trailingslashit( $original_directory ) . $filename_without_ext . ".{$extension[0]}";
+
+			$metadata['sizes'][ $size ]['sources'][ $targeted_mime ] = array(
+				'file'     => $newfile,
```
```suggestion
				'file'     => wp_basename( $newfile ),
```

Only the basename is stored similar to how WordPress stores the sizes.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,99 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+		function ( $metadata, $post_meta_id ) use ( $post_id, $valid_mime_transforms, $allowed_mimes, $file ) {
+			if ( $post_meta_id !== $post_id ) {
+				return $metadata;
+			}
+
+			return webp_uploads_update_sources( $metadata, $valid_mime_transforms, $allowed_mimes, $file );
+		},
+		10,
+		2
+	);
```
mitogh: Looks like we are missing the filterable value.

```suggestion
      return $override;
	);
```

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,99 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+		}
+
+		if ( ! wp_image_editor_supports( array( 'mime_type' => $targeted_mime ) ) ) {
+			return $override;
+		}
+
+		$extension   = explode( '|', $allowed_mimes[ $targeted_mime ] );
+		$destination = trailingslashit( $original_directory ) . "{$filename}.{$extension[0]}";
+
+		if ( is_wp_error( $image->save( $destination, $targeted_mime ) ) ) {
```
mitogh: We need to store the result of this operation at the top level `sources` property for the full-size image as currently, we are missing that one. The `sources` at the top level for the full-size image are not created.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,99 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ * @return array                        Metadata with sources added.
+ */
+function webp_uploads_update_sources( $metadata, $valid_mime_transforms, $allowed_mimes, $file ) {
+	$original_directory = pathinfo( $file, PATHINFO_DIRNAME );
+
+	foreach ( $metadata['sizes'] as $size => $size_details ) {
+		foreach ( $valid_mime_transforms as $targeted_mime ) {
+			$extension            = explode( '|', $allowed_mimes[ $targeted_mime ] );
+			$filename_without_ext = implode( explode( '.', $size_details['file'], -1 ) );
+			$newfile              = trailingslashit( $original_directory ) . $filename_without_ext . ".{$extension[0]}";
```
mitogh: We need to make sure the file exists when this section of code is executed and to use functions like `filesize`.

```suggestion
			$newfile              = trailingslashit( $original_directory ) . $filename_without_ext . ".{$extension[0]}";
			if ( ! file_exists( $newfile ) {
			    continue;
			}
```

mitogh: Also my 2 cents on the variable name would be something more meaningful, like.

```suggestion
			$image_file           = trailingslashit( $original_directory ) . $filename_without_ext . ".{$extension[0]}";
```

mitogh (commented):
Great work on this PR @kirtangajjar  this is a very cleaver solution in how to go about the problem, it took me a bit of time to wrap my head around it but I love the simplicity of this approach. 

Added some comments / feedback but overall things are working on my local after running some tests.

Speaking of tests would be great if we can include some as well, let me know if you need a hand with that, I've created a PR for a related feature that you can use as en example in case that helps for your testing strategy:

- https://github.com/WordPress/performance/pull/242/files

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,120 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ * @param string $file                  Path to original file.
+ * @return array                        Metadata with sources added.
+ */
+function webp_uploads_update_sources( $metadata, $valid_mime_transforms, $allowed_mimes, $file ) {
+	$original_directory = pathinfo( $file, PATHINFO_DIRNAME );
+
+	foreach ( $valid_mime_transforms as $targeted_mime ) {
+		// Add sources to original image metadata.
+		$extension            = explode( '|', $allowed_mimes[ $targeted_mime ] );
+		$filename_without_ext = implode( explode( '.', $file, -1 ) );
```
mitogh: Would be ideal to rely on `pathinfo` instead.

```suggestion
		$filename_without_ext = pathinfo( $file, PATHINFO_FILENAME );
```

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,120 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+		$metadata['sources'][ $targeted_mime ] = array(
+			'file'     => wp_basename( $image_file ),
+			'filesize' => filesize( $image_file ),
+		);
+
+		foreach ( $metadata['sizes'] as $size => $size_details ) {
+			// Add sources to resized image metadata.
+			$filename_without_ext = implode( explode( '.', $size_details['file'], -1 ) );
+			$image_file           = trailingslashit( $original_directory ) . $filename_without_ext . ".{$extension[0]}";
```
```suggestion
			$image_file           = path_join( $original_directory, "$filename_without_ex.{$extension[0]}" )
```

modules/images/webp-uploads/load.php
mitogh: I would advise to rename this variable to make it explicit. 

```suggestion
		$filename_without_extension = implode( explode( '.', $file, -1 ) );
```

mitogh (commented):
Great work @kirtangajjar with the updates I've added some final comments besides that and the tests I think we would be good from my end.

One thing I was able to identify while testing your changes locally is the fact that the `sources` for the full size image is not updated correctly it always references the original image.

kirtangajjar:
@mitogh I've fixed updating the top-level `sources` key in metadata.

> Speaking of tests would be great if we can include some as well, let me know if you need a hand with that, I've created a PR for a related feature that you can use as an example in the case that helps for your testing strategy:

Thank you for this. That definitely made my work easy. 

I just have a question - What would be the best way to use the functionality of the `WP_Image_Edit` class in my PR?

I wanted to simulate the crop feature. If I try to copy what the function does in my unit test, the test looks dirty. Copying over the class would mean duplicate code in both the PRs 😅. 

I think it would be great if you could open and merge a separate PR just adding that helper class. Once done, then I can make use of it in my test and we would be good to go.

What do you think about it? Or do we have a better solution at hand here?

mitogh:
Great question, the way you can do it here is by doing a checkout of the file from the branch I pointed out so you have only the file. For example:

```
git checkout feature/228-restore-and-backup-image-sources -- tests/testdata/modules/images/webp-uploads/class-wp-image-edit.php
```

Make sure you have the branch `feature/228-restore-and-backup-image-sources` locally in your machine by pulling the latest from `origin` (in this case this repository).

That would pull the full file into this PR let me know in case you need additional help, the class already has a way to handle the crop. 

Let me know in case you have any additional questions @kirtangajjar

kirtangajjar:
Thanks a lot for your help @mitogh. I have added one test case to ensure that 'sources' properties are updated after the image update. 

Please share if you have more things that could be tested here. 

cc @felixarntz

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -809,6 +809,28 @@ function( $transforms ) {
...
+		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg' );
+
+		$editor = new WP_Image_Edit( $attachment_id );
+		$editor->crop( 10, 10, 0, 0 )->save();
+		$this->assertTrue( $editor->success() );
+
+		$metadata = wp_get_attachment_metadata( $attachment_id );
+
+		$this->assertArrayHasKey( 'sources', $metadata );
+		$this->assertArrayHasKey( 'sizes', $metadata );
```
mitogh, line 1028: This look good it would be great to include the custom assertion to verify the sources like

```
		$this->assertImageHasSource( $attachment_id, 'image/webp' );
		$this->assertImageNotHasSource( $attachment_id, 'image/jpeg' );
```

You can use the same assertions for the size names as well.

```
https://github.com/WordPress/performance/blob/5790b8027e9a4216de2709304b251dbba413bcab/tests/modules/images/webp-uploads/webp-uploads-test.php#L807-L808
```

Also, be aware that because of the dimensions of your edit not all subsizes would be affected so you can use a higher crop that affects all subsizes or you can perform a different edit like a flip.

Once assertion that would be wort checking is making sure the actual values from the sources are edited files you can use the following assertion:

```php
$this->assertRegExp( '/e\d{13}/', $file_name );
```

Due edited images have a `e{13_DIGIT_NUMBER}` to ensure the values on the meta are actual edited versions of the file and not the original image.

mitogh (commented):
Great work @kirtangajjar some minor comments on the testing part.

tests/modules/images/webp-uploads/webp-uploads-test.php
kirtangajjar: @mitogh Thanks a lot for these pointers. I have updated the code to include all the newer assertions as well.

mitogh (approved):

tests/modules/images/webp-uploads/webp-uploads-test.php
mitogh: Thanks, great work on getting this one ready.

modules/images/webp-uploads/load.php
felixarntz: We should still store the result of this operation here to reuse it for `sources` instead of "guessing" the changed image file name in `webp_uploads_update_sources()`.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,121 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+	foreach ( $valid_mime_transforms as $targeted_mime ) {
+		// Add sources to original image metadata.
+		$extension                  = explode( '|', $allowed_mimes[ $targeted_mime ] );
+		$filename_without_extension = pathinfo( $file, PATHINFO_FILENAME );
+		$image_file                 = path_join( $original_directory, "$filename_without_extension.{$extension[0]}" );
+
+		if ( ! file_exists( $image_file ) ) {
+			continue;
+		}
```
felixarntz: See my other comment, we should rather have the _actual_ file name (which is returned by the `WP_Image_Editor::save()` call) passed to this function via parameter.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,121 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+			return $override;
+		}
+
+		$_sizes = array();
+
+		foreach ( $sizes as $size => $size_details ) {
+			$_sizes[ $size ] = $size_details;
+		}
+
+		$image->multi_resize( $_sizes );
```
felixarntz: There is a problem with generating the sub-sizes here: In this code we generate new images for all sizes, however that may not be intended. We need to instead only generate the same sizes that were also generated for the original MIME by WordPress core.

To do that, I suggest we remove the process of generating additional sizes from _this_ function and instead handle it within the `wp_update_attachment_metadata` hook below. That would be better for the following reason:
* The function here is hooked into `wp_save_image_editor_file`. At this point, WordPress core has saved the edited (full) image, but it hasn't generated any sub-sizes for it yet, so generating the additional sub-sizes here is too early.
* By moving sub-sizes generation into the `wp_update_attachment_metadata` hook callback, we can compare the updated sizes (`$metadata['sizes']`) with the previous sizes (`wp_get_attachment_metadata( $post_meta_id )['sizes']`). Then, we should create the additional MIME type versions for _only_ the sizes that have changed.
* We can compare the file name for each size; if it hasn't changed in the new metadata compared to the other metadata, that sub-size wasn't generated, so we shouldn't do it either.

Side note: The `foreach` loop right now does nothing useful anyway, as it just copies over `$sizes` into `$_sizes`, without change.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,121 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	}
+	add_filter(
+		'wp_update_attachment_metadata',
+		function ( $metadata, $post_meta_id ) use ( $post_id, $valid_mime_transforms, $allowed_mimes, $file ) {
+			if ( $post_meta_id !== $post_id ) {
+				return $metadata;
+			}
+
+			return webp_uploads_update_sources( $metadata, $valid_mime_transforms, $allowed_mimes, $file );
```
felixarntz: See my comment above, it would be more appropriate to generate the additional image sub-sizes right here, before calling `webp_uploads_update_sources()`. We need to only generate sub-sizes for those in `$metadata['sizes']` that differ from those in `wp_get_attachment_metadata( $post_meta_id )['sizes']`.

Moving the logic itself should be fairly trivial, we can simply pass the `$image` variable (the `WP_Image_Editor` instance) through to the callback here as well.

Also, per my other comment, we should pass the additional full image's new filename to this callback so that we can use it to populate the main `$metadata['sources']`.

I think at that point I'm not sure how useful `webp_uploads_update_sources()` to just update the sources still is. Again, the primary purpose of that function right now is to "guess" the new image file names, but it would be much more reliable to directly use the return value of the `WP_Image_Editor::multi_resize()` call.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,121 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+		);
+
+		foreach ( $metadata['sizes'] as $size => $size_details ) {
+			// Add sources to resized image metadata.
+			$filename_without_extension = pathinfo( $size_details['file'], PATHINFO_FILENAME );
+			$image_file                 = path_join( $original_directory, "$filename_without_extension.{$extension[0]}" );
+
+			if ( ! file_exists( $image_file ) ) {
+				continue;
+			}
```
felixarntz: See my other comments, we shouldn't do this. `WP_Image_Editor::multi_resize()` returns the real file names generated, so let's reuse that instead of assuming the filename here, which is error-prone.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,121 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
 	return rest_ensure_response( $data );
 }
 add_filter( 'rest_prepare_attachment', 'webp_uploads_update_rest_attachment', 10, 3 );
+
+/**
+ * Adds sources to metadata for an attachment.
+ *
```
felixarntz, line 689: This is missing a `@since` annotation.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,121 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+		}
+
+		$_sizes = array();
+
+		foreach ( $sizes as $size => $size_details ) {
+			$_sizes[ $size ] = $size_details;
+		}
+
+		$image->multi_resize( $_sizes );
+	}
```
felixarntz: The final `return $override;` here is missing.

felixarntz (commented):
@mitogh @kirtangajjar There are a few problems still with this one. My main concern here is that we try to assemble the file names for the newly generated images instead of reliably using the real image file names which are returned via `WP_Image_Editor::save()` and `WP_Image_Editor::multi_resize()`.

We also ignore currently which sizes were generated by core and instead generate all of them, which may not always be correct.

kirtangajjar:
@felixarntz Thank you so much for the feedback. I think I have addressed all of your feedback, and now the code feels even more polished than earlier! Please have a look at it again and let me know if it needs anything else.

P.S. I will check on the failing unit tests tomorrow.

mitogh:
Thanks for your work @kirtangajjar I will take a look at the failing tests and move forward with the rest of the changes.

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+					$main_images[ $targeted_mime ]     = array( 'path' => $file );
+					$subsized_images[ $targeted_mime ] = $metadata['sizes'];
+					continue;
+				}
+
+				if ( ! isset( $allowed_mimes[ $targeted_mime ] ) || ! is_string( $allowed_mimes[ $targeted_mime ] ) ) {
+					return null;
+				}
+
+				if ( ! wp_image_editor_supports( array( 'mime_type' => $targeted_mime ) ) ) {
```
felixarntz: Technically, this check is not fully reliable, since it checks if _any_ of the available image editors supports the MIME type. But we're already using a specific image editor here (within the `$image` variable), so we need to check if _that_ image editor supports the target MIME type.

You can see how we can do that by checking the internals of `wp_image_editor_supports()` and `_wp_image_editor_choose()` - basically we have to rely on (whether there is) a `supports_mime_type()` method on `$image`.

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+			foreach ( $valid_mime_transforms as $targeted_mime ) {
+				if ( $targeted_mime === $mime_type ) {
+					$main_images[ $targeted_mime ]     = array( 'path' => $file );
+					$subsized_images[ $targeted_mime ] = $metadata['sizes'];
+					continue;
+				}
+
+				if ( ! isset( $allowed_mimes[ $targeted_mime ] ) || ! is_string( $allowed_mimes[ $targeted_mime ] ) ) {
+					return null;
```
felixarntz: Why return? Shouldn't this `continue` instead? We still need to make sure to process other MIME types.

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+				}
+
+				if ( ! wp_image_editor_supports( array( 'mime_type' => $targeted_mime ) ) ) {
+					return null;
```
felixarntz: See above, this needs to `continue` instead.

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+				}
+
+				$extension   = explode( '|', $allowed_mimes[ $targeted_mime ] );
+				$destination = trailingslashit( $original_directory ) . "{$filename}.{$extension[0]}";
+
+				$main_images[ $targeted_mime ] = $image->save( $destination, $targeted_mime );
+
+				if ( is_wp_error( $main_images[ $targeted_mime ] ) ) {
+					return null;
```
felixarntz: See above.

Also, we need to make sure to not set this in `$main_images[ $targeted_mime ]` until we know that it's _not_ a `WP_Error`, otherwise we may then pass a `WP_Error` within the data to `webp_uploads_update_sources()` below.

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+function webp_uploads_update_sources( $metadata, $valid_mime_transforms, $file, $main_images, $subsized_images ) {
+	$original_directory = pathinfo( $file, PATHINFO_DIRNAME );
+
+	foreach ( $valid_mime_transforms as $targeted_mime ) {
+		// Add sources to original image metadata.
+		$image_file = $main_images[ $targeted_mime ]['path'];
+
+		if ( ! file_exists( $image_file ) ) {
+			continue;
+		}
```
felixarntz: We don't really need to check here whether the file exists I'd say, since if it didn't exist, the previous function would have errored. What we _do_ need to check more importantly though is whether `$main_images[ $targeted_mime ]` is set, which it may not be if generating that image failed below. See my related comment on not setting a `WP_Error` in there.

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+		$metadata['sources'][ $targeted_mime ] = array(
+			'file'     => wp_basename( $image_file ),
```
felixarntz: This is an unnecessary function call to `wp_basename()`, we should use `$main_images[ $targeted_mime ]['file']` instead, which is already provided.

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+		foreach ( $metadata['sizes'] as $size => $size_details ) {
+			// Add sources to resized image metadata.
+			$image_file = $original_directory . '/' . $subsized_images[ $targeted_mime ][ $size ]['file'];
+
+			if ( ! file_exists( $image_file ) ) {
+				continue;
+			}
+
+			$metadata['sizes'][ $size ]['sources'][ $targeted_mime ] = array(
+				'file'     => wp_basename( $image_file ),
```
felixarntz: See above, we don't even need to care about the directory here and can avoid the function call. This is already provided to us.
```suggestion
				'file'     => $subsized_images[ $targeted_mime ][ $size ]['file'],
```

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+		}
+
+		$metadata['sources'][ $targeted_mime ] = array(
+			'file'     => wp_basename( $image_file ),
+			'filesize' => filesize( $image_file ),
+		);
+
+		foreach ( $metadata['sizes'] as $size => $size_details ) {
+			// Add sources to resized image metadata.
+			$image_file = $original_directory . '/' . $subsized_images[ $targeted_mime ][ $size ]['file'];
```
felixarntz: Similar to above, before accessing it we need to check whether `$subsized_images[ $targeted_mime ][ $size ]` is set.

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ *
+ * @since n.e.x.t
+ *
+ * @param array  $metadata              Metadata of the attachment.
+ * @param array  $valid_mime_transforms List of valid mime transforms for current image mime type.
+ * @param string $file                  Path to original file.
+ * @param array  $main_images           Path of all main image files of all mime types.
+ * @param array  $subsized_images       Path of all subsized image file of all mime types.
+ *
+ * @return array                        Metadata with sources added.
```
felixarntz: There doesn't need to be a blank line here and the description doesn't need to be indented similar to the `@param` docs.
```suggestion
 * @return array Metadata with sources added.
```

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ *
+ * @since n.e.x.t
+ *
+ * @param bool|null       $override  Value to return instead of saving. Default null.
+ * @param string          $file      Name of the file to be saved.
+ * @param WP_Image_Editor $image     The image editor instance.
+ * @param string          $mime_type The mime type of the image.
+ * @param int             $post_id   Attachment post ID.
+ *
+ * @return bool|null
```
felixarntz: See above, also this should have a description, e.g.:
```suggestion
 * @return bool|null Potentially modified $override value.
```

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ */
+function webp_uploads_update_image_onchange( $override, $file, $image, $mime_type, $post_id ) {
+	if ( null !== $override ) {
+		return $override;
+	}
+
+	if ( empty( webp_uploads_get_upload_image_mime_transforms()[ $mime_type ] ) ) {
+		return null;
```
```suggestion
		return $override;
```

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+				$subsized_images[ $targeted_mime ] = $image->multi_resize( $resize_sizes );
+			}
+
+			return webp_uploads_update_sources( $metadata, $valid_mime_transforms, $file, $main_images, $subsized_images );
+		},
+		10,
+		2
+	);
+
+	return null;
```
```suggestion
	return $override;
```

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ * @param int             $post_id   Attachment post ID.
+ *
+ * @return bool|null
+ */
+function webp_uploads_update_image_onchange( $override, $file, $image, $mime_type, $post_id ) {
+	if ( null !== $override ) {
+		return $override;
+	}
+
+	if ( empty( webp_uploads_get_upload_image_mime_transforms()[ $mime_type ] ) ) {
```
felixarntz: Nit-pick, but WP core standards suggest to avoid array access directly on a function return value, so let's assign this to a variable first.

I'd also suggest we pass this to the closure below via `use`, so that we don't have to look up the exact same thing again in that closure.

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	}
+
+	add_filter(
+		'wp_update_attachment_metadata',
+		function ( $metadata, $post_meta_id ) use ( $post_id, $file, $mime_type, $image ) {
+			if ( $post_meta_id !== $post_id ) {
+				return $metadata;
+			}
+
+			$valid_mime_transforms = webp_uploads_get_upload_image_mime_transforms()[ $mime_type ];
```
felixarntz: See above, we can remove this in favor of getting it from above where this is already looked up.

modules/images/webp-uploads/load.php
```diff
@@ -684,6 +684,135 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+			$original_directory    = pathinfo( $file, PATHINFO_DIRNAME );
+			$filename              = pathinfo( $file, PATHINFO_FILENAME );
+			$allowed_mimes         = array_flip( wp_get_mime_types() );
+			$old_metadata          = wp_get_attachment_metadata( $post_id );
+			$resize_sizes          = array();
+			$main_images           = array();
+			$subsized_images       = array();
```
felixarntz: Nit-pick, but having all these variables defined right here even though they're not used right away makes the code a bit hard to read. These are all local variables, so we don't need to put them all together, since they aren't used anywhere else anyway.

`$allowed_mimes`, `$main_images`, and `$subsized_images` are only used in the second `foreach` loop below, so I think it would be better for code readability and clarity on what these variables are for if we moved those three from here directly above the second `foreach`.

felixarntz (commented):
@mitogh @kirtangajjar Functionally, this mostly looks good, though there are still a few caveats to address, and some nitpicks. But this is really close now.

modules/images/webp-uploads/load.php
mitogh: At this point `$override` is `null` because of this:

- https://github.com/WordPress/performance/blob/2f47f0e99e24f0fbdf7c924fddea520586c535a6/modules/images/webp-uploads/load.php#L748

felixarntz:
@mitogh This was close to approval in my last review, but now you've added https://github.com/WordPress/performance/pull/235/commits/6d30ccb869938c0f3fae9f9a06b3906c5952557a which basically changes a ton of things around again, so we're "back at 0".

Could you please explain this and the other 2 most recent commits you've added? I don't think it's a good idea to adjust the approach here significantly once again - this is a significant change which you didn't explain for consideration before.

Unless there is a very good reason for this major update to the PR, my suggestion would be to revert the latest 3 commits. We can always iterate later in a follow-up PR, but this has been open for too long already to go back to the drawing board.

mitogh:
Good question @felixarntz 

The main problem with the existing approach was the fact that due the previous structure it was impossible to unhook due the usage of the closure internally, and we need it to unhook in order to remove the logic in a scenario when the tests where failing and in order to prevent an infinite loop in the admin area.

So we had some options:

1. Move the logic from the hooks into another function and pass the `use` variables as arguments, which would lead to another problem with a very similar code to the create function.
2. Adjust the code from the original create sources in order to behave in the way the filters `wp_generate_attachment_metadata` and `wp_update_attachment_metadata` works. 
3. Mark the failing tests as incomplete and implement the logic as it was with the provided feedback. However the main problem with this approach is the fact that won't be a final solution for #158 

I faced the problem until I was fixing the tests due this would lead to an unstable scenario. Happy to rollback the 3 last commits and adjust the code as need it to bypass the failing tests.

The last 2 commits includes 2 additional cases from the edit media screen that were not covered but are part of the edit media.

- When an image is edited the changes can be applied to All images, Thumbnail Only, All Images except the thumbnail.
- When an image is edited the attached image is used the original image is not used during the subsequent edits to an image.

Let me know what do you think, but I think if we go with approach 1 we eventually have to go back to this approach, and if we go with approach 3 I don't think we should include this change as it is into the release. @felixarntz

mitogh:
Hey @felixarntz this is ready for another review based on your feedback, let me know in case you have any additional comments.

I've removed the part of this PR that focuses in handling the case when only the thumbnail or all images except the thumbnail are edited, will try to wrap a PR shortly for that in case I'm not able to get it done before tomorrow EOD I'm going to open a ticket with a draft PR for this remaining feature.

Let me know in case you have any additional feedback about this PR, thanks.

modules/images/webp-uploads/load.php
felixarntz: Fair point. I'd still say always returning `$override` here is a more error-proof approach. There's no benefit in returning `null` over `$override`, and returning `$override` will ensure this still works, should that first condition be removed at any point in the future.

modules/images/webp-uploads/load.php
```diff
@@ -700,6 +700,11 @@ function webp_uploads_update_sources( $metadata, $valid_mime_transforms, $file,
 	$original_directory = pathinfo( $file, PATHINFO_DIRNAME );
 
 	foreach ( $valid_mime_transforms as $targeted_mime ) {
+		// Make sure the path and file exists as those values are being accessed.
+		if ( empty( $main_images[ $targeted_mime ]['path'] ) || empty( $main_images[ $targeted_mime ]['file'] ) ) {
+			continue;
+		}
+
 		// Add sources to original image metadata.
 		$image_file = $main_images[ $targeted_mime ]['path'];
```
felixarntz: The current implementation works, but it might be safer to rely on the directory from the saved image path, rather than the original file's directory.

They should be the same, but the one where the saved main image was stored in is more likely to be the one where all the other files are stored in. Basically, instead of assigning `$original_directory` outside of this loop, we could do it here:
```php
$original_directory = pathinfo( $main_images[ $targeted_mime ]['path'], PATHINFO_DIRNAME );
```

felixarntz (commented):
@mitogh This looks great!

Left two minor comments, but nothing really blocking.

felixarntz (approved):
Awesome work @kirtangajjar @mitogh - this was quite complex, but is good to merge now!

Let's create a follow-up issue and PR to enhance the target-specific logic.

modules/images/webp-uploads/load.php
mitogh: Agree, just wanted to avoid the need to call `pathinfo` with the assumption that all files including full size and subsizes are stored in the same directory.

mitogh:
Thanks @felixarntz  the issue has been created:

- #299

PR #240, opened 2022-03-18
Define the plugin's version support and backward compatibility policy

### LINKED ISSUE #239: Define support policy for WordPress core versions and plugin modules
As stated in https://make.wordpress.org/core/2022/03/07/the-performance-lab-plugin-has-been-released/:

> Because the Performance Lab plugin is a collection of potential WordPress core feature modules, the list of modules included may drastically change over time. New modules may be added regularly, while other modules may be removed in a future plugin version once they have landed in a WordPress core release.

Related to the above, specifically the point on module removal, it is essential that we define a policy to ensure users can rely on the plugin's functionality without having to worry that it is suddenly no longer there. In addition, we should formalize the policy for which minimum WordPress core version the plugin will require.

felixarntz:
## Summary

Fixes #239

## Relevant technical choices

* This is a draft with a proposed version support policy.
* Please review this PR not just for grammar and copy, but more importantly for the actual policy defined.

## Checklist


ThierryA (approved):

dainemawer (approved):

docs/Version-support-policy.md
```diff
@@ -0,0 +1,32 @@
...
+
+# Version support policy
+
+As outlined in the [plugin announcement post](https://make.wordpress.org/core/2022/03/07/the-performance-lab-plugin-has-been-released/), the Performance Lab plugin is a collection of performance-related "feature projects" for WordPress core. These feature projects are represented in this plugin as individual standalone [modules](./Writing-a-module.md).
+
+All modules bundled in the Performance Lab plugin are at different stages of development and may be merged into WordPress core at different points. This document describes the version support policy for the different modules, with a special focus on backward compatibility.
+
+## WordPress core versions
+
+**The Performance Lab plugin commits to supporting the latest stable version of WordPress core.** With that said, the minimum WordPress version requirement will not be bumped unnecessarily, so realistically the plugin will usually support the latest _two_ stable versions of WordPress core. If or once a feature that is only available in the latest WordPress core stable release is required for a plugin module, the minimum requirement of the plugin will be bumped to that version.
```
sgomes, line 11: I'm not sure if there's much value in specifically talking about two versions. Would it perhaps suffice to say we support the latest stable version, but may support more, in that we won't unnecessarily bump the requirement unless a feature demands it?

docs/Version-support-policy.md
```diff
@@ -0,0 +1,32 @@
...
+* If module development is considered completed and continuing development within WordPress core is preferred, the module should be removed from the plugin eventually.
+* If module development can continue in the plugin based on the parts that have already been merged into WordPress core, only the parts of the module that were merged should be removed from the plugin eventually.
+
+Whichever of the two options applies to a module, (critical functionality of) standalone modules in the Performance Lab plugin must never be removed in a way that would break sites which rely on the module's current feature set.
+
+Therefore, **the Performance Lab plugin commits to only removing a module in combination with bumping the minimum WordPress version requirement to the same stable core release that the merged module code was published with**. This is best clarified with an example:
+* If a module gets merged during development of WordPress 6.0, the module will not be removed from the plugin immediately.
+* Only after WordPress 6.0 has been released, the module can be removed from the plugin.
```
sgomes:
```suggestion
* Only after WordPress 6.0 has been released, can the module be removed from the plugin.
```

docs/Version-support-policy.md
sgomes: I'm not sure if there's much value in specifically talking about two versions. Would it perhaps suffice to say we support the latest stable version, but may support more, in that we won't unnecessarily bump the requirement unless a feature demands it?

To be clear, the approach seems good, I'm only trying to simplify the wording.

sgomes (approved):
LGTM, minor comments. Thanks, @felixarntz! 👍

docs/Version-support-policy.md
```diff
@@ -0,0 +1,32 @@
+[Back to overview](./README.md)
+
+# Version support policy
+
+As outlined in the [plugin announcement post](https://make.wordpress.org/core/2022/03/07/the-performance-lab-plugin-has-been-released/), the Performance Lab plugin is a collection of performance-related "feature projects" for WordPress core. These feature projects are represented in this plugin as individual standalone [modules](./Writing-a-module.md).
+
+All modules bundled in the Performance Lab plugin are at different stages of development and may be merged into WordPress core at different points. This document describes the version support policy for the different modules, with a special focus on backward compatibility.
```
mitogh, line 7: When we say `backward compatibility` do we mean with the previous version of the plugin or the previous version of WordPress, would be ideal if we can clarify it.

docs/Version-support-policy.md
```diff
@@ -0,0 +1,32 @@
...
+
+**The Performance Lab plugin commits to supporting the latest stable version of WordPress core.** With that said, the minimum WordPress version requirement will not be bumped unnecessarily, so realistically the plugin will usually support the latest _two_ stable versions of WordPress core. If or once a feature that is only available in the latest WordPress core stable release is required for a plugin module, the minimum requirement of the plugin will be bumped to that version.
+
+Supporting a greater array of older WordPress core versions would be detrimental for the plugin, as it is primarily intended for beta testing and often requires using features that have only been introduced in a recent WordPress core version.
+
+It needs to be clarified though that **the Performance Lab plugin will never require an alpha or beta version of a future WordPress core release**. This policy ensures that sites using the latest stable version of WordPress will always be able to use the plugin as well.
+
+## Removal of modules
+
+Once a module has been merged into WordPress core, there are two potential options for how to proceed with the module:
```
mitogh: I think we should standardize the way the modules are enabled currently is based on the settings but I think each module might be different so we should define a filterable function available in all modules like `function webp_uploads_should_module_load` with the specific logic for that module.

adamsilverstein: maybe "merged into WordPress and shipped in a release" here, we wouldn't remove the module code until the feature is available in stable WordPress

adamsilverstein:
```suggestion
Once a module has been merged and shipped in a WordPress release, there are two potential options for how to proceed with the module:
```

felixarntz: Fair point. I'm already clarifying the exact point _when_ to proceed in the section below, but probably including this notion here already can help.

docs/Version-support-policy.md
felixarntz: It's neither nor really, or both, depending how we wanna look at it. It's backward compatibility in the sense of "a site that has this feature should continue to be able to have that feature", which is a combination between WordPress core and the plugin playing well together.

I'm outlining this in depth below, so I'm thinking that's probably sufficient.

docs/Version-support-policy.md
felixarntz: Sounds good, I'll simplify the wording accordingly.

docs/Version-support-policy.md
felixarntz: This is a great point, we eventually need some way to not load a module in case the site is on a WordPress core version that already includes this feature (e.g. after a merge), and potentially we may even want to inform the admin about this when they access the plugin's settings screen.

That's unrelated to this documentation though, we can open a separate issue for it.

mitogh: :100: agree with that.

mitogh (approved):

PR #242, opened 2022-03-19
Restore and backup image sizes alongside the sources properties

### LINKED ISSUE #228: Restore and backup Image sizes with sources when an image is edited
## Description

- When an image is edited make sure that existing images are stored in the `_wp_attachment_backup_sizes` meta before applying the changes into the image. 
- If the image is restored update `_wp_attachment_backup_sizes` array with the appropriate values.
- At all times the value stored in the metadata of the image should reflect the same state as the images being edited.

When the constant `IMAGE_EDIT_OVERWRITE` is defined and is set to `true` edits on the file would be removed and only the previous edit would be stored in `_wp_attachment_backup_sizes` still allowing to rollback to the original image state, but subsequent edits would effectively remove any previous edit on the files.

## Respect the constant `IMAGE_EDIT_OVERWRITE`

This behavior should follow the same principles for the primary image format in the secondary image format meaning that:

- if `IMAGE_EDIT_OVERWRITE` is defined
  - if the value of the constant is true only the last copy of the secondary image would be kept in the filesystem and referenced in the database.
  - if the value of the constant is false multiple copies of the same edit would be kept in the filesystem and in the database in the same way it occurs with the primary image format.
- If the constant is not defined, we should follow the same scenario as if the constant was defined as false

## Restoring an image

When the original image is restored, the secondary image format would also be restored in the same way the process is applied to the primary image format. This restore would follow the rules from the constant definition `IMAGE_EDIT_OVERWRITE` to make sure files are removed or kept in place depending on the value of this constant.

## Notes

- In all scenarios, the values in `_wp_attachment_backup_sizes` should accurately reflect the file locations and properties of each edited image.
- Image edits should include the suffix `e-$id` where `$id` is a 13 digit number specifically `time() . rand( 100, 999 )` [created by WP ](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/includes/image-edit.php#L708)to reference any additional edit that is no longer relevant, the file name is important so WP can identify images that reference to edits and can be removed accordingly based on the definition and value of `IMAGE_EDIT_OVERWRITE` This suffix is also included as part of the image size name.
- Last update or the restore point in `_wp_attachment_backup_sizes` uses the suffix `-orig` on the file name and image size in order to quickly identify which image is the last one edited or the one we need to restore. [Reference](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/includes/image-edit.php#L739)

### Related issues:

The work in this ticket is part of:

- #158 

In order to keep things in a more manageable way.

mitogh:
## Summary

Fixes #228 

## Relevant technical choices

**Usage of `_sources` metadata key**

A `_sources` property is stored to hold the top-level `sources` for the full image size when an image is edited, after the backup is updated the `_sources` property would be moved into the appropriate `full` size inside of the backup metadata and the temporarily `_sources` property would be removed from the attachment image.

This key is used so when the metadata is updated we still have a reference to the previous sources values to the full size image when the image is edited. This is not the case for the rest of real sizes, this is only the case for full sizes.

When the image is restored the `sources` property is populated again from the backup metadata if the `full-orig` has the `sources` property.

**Removal of edited images**

When an image is edited and restored any edited version of the images are removed if `IMAGE_EDIT_OVERWRITE` is defined as a `truthy` value.

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,190 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ * @return array The updated metadata for the attachment.
+ */
+function webp_uploads_backup_sources( $attachment_id, $data ) {
+	if ( ! isset( $data['sources'] ) ) {
+		return $data;
+	}
+
+	$data['_sources'] = $data['sources'];
+	// Remove the current sources as at this point the current values are no longer accurate.
+	unset( $data['sources'] );
```
felixarntz: This is a bit problematic in terms of core, since core supports multiple revisions for edited images, which this code here doesn't. How about we follow the WP core implementation from `wp_save_image()` and we use a new `_wp_attachment_backup_sources` meta key to store the values? To elaborate a bit further:
1. Get the current value of `_wp_attachment_backup_sources` into `$backup_sources`. Fall back to an empty array if not set.
2. Define a suffix variable similar like in core, e.g. `$suffix = time() . rand( 100, 999 )`.
3. If `$backup_sources['sources-orig']` already exists, set the current `sources` in `$backup_sources["sources-$suffix"]`. Otherwise (no backup exists yet), use `$backup_sources['sources-orig']`.
4. Unset `sources` from the original metadata entirely (since we don't yet support generating sources for image edits, which will be part of #158).
5. Set `_wp_attachment_backup_sources` with the updated `$backup_sources`.

Doesn't have to be exactly like the above, but we need to support history, and I think a separate meta key makes sense here to follow core.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,190 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+		return $data;
+	}
+
+	// If `IMAGE_EDIT_OVERWRITE` is defined and is truthy remove any edited images if present before replacing the metadata.
+	if ( defined( 'IMAGE_EDIT_OVERWRITE' ) && IMAGE_EDIT_OVERWRITE ) {
+		$file     = get_attached_file( $attachment_id );
+		$dirname  = pathinfo( $file, PATHINFO_DIRNAME );
+		$metadata = wp_get_attachment_metadata( $attachment_id );
+
+		$sources = isset( $metadata['sources'] ) && is_array( $metadata['sources'] ) ? $metadata['sources'] : array();
```
felixarntz: This is incorrect I believe. You're hooking into `wp_update_attachment_metadata`, so the `sources` here will be the _original_ ones - which we must not delete under any circumstances.

When restoring, what should be deleted is the backed up sources. See my above suggestion, we could do that with the `_wp_attachment_backup_sources` meta key (or based on your current implementation `_sources`).

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,190 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+					continue;
+				}
+
+				// Delete only if it's an edited image.
+				if ( preg_match( '/-e\d{13}/', $source_properties['file'] ) ) {
+					$delete_file = path_join( $dirname, $source_properties['file'] );
+					wp_delete_file( $delete_file );
+				}
+			}
+		}
```
felixarntz: I actually think this entire block of code can be removed. We don't support generating additional MIME types for image edits yet, so none of the edited images (neither the original nor any `sizes`) will have any `sources` that we'd need to delete.

This will all be part of #158.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,190 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	}
+
+	if ( isset( $backup_sizes['full-orig']['sources'] ) ) {
+		$data['sources'] = $backup_sizes['full-orig']['sources'];
+	}
```
felixarntz: This is the only part of this function which we actually need. In `webp_uploads_backup_sources()` we (rightfully) only back up the main `sources`, so here we have to restore them. The `sources` for every image size should work out of the box with WP core, since it just copies/moves over the complete size data, which would include `sources`. Again, we don't generate additional sources for image edits, so no need to delete any files here.

On another note though, we can't rely on `$backup_sizes['full-orig']['sources']` as that won't be set. When you review the code from WP core's `wp_save_image()` and `wp_restore_image()` closely, you will see that they manually assemble the `full-orig` array, which means it'll never include the `sources`. Only for the sub-sizes it copies/moves over the original data, so the `sources` there are maintained.

In other words, what we need to do here is rely on our custom storage (e.g. the `_wp_attachment_backup_sources` meta key I proposed above, or your current implementation `_sources`).

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +683,190 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ * metadata back into the full size similar as how it's done on the rest of the sizes.
+ *
+ * @since n.e.x.t
+ *
+ * @param int    $meta_id The ID of the meta value stored in the DB.
+ * @param int    $attachment_id The ID of the post used for this meta, in our case the attachment ID.
+ * @param string $meta_name The name of the metadata to be updated.
+ * @param array  $backup_sizes An array with the metadata value in this case the backup sizes.
+ */
+function webp_updated_postmeta( $meta_id, $attachment_id, $meta_name, $backup_sizes ) {
```
felixarntz: I don't think this function is necessary, and it makes the code overly complicated. We don't have to put the `sources` into `_wp_attachment_backup_sizes`. I see WP core doesn't do this by default, since it manually assembles the array for the `full-orig` size. However, a much simpler and less hacky implementation would be to rely on the separate `_wp_attachment_backup_sources` meta key I suggested above. What you have here may work, but it has some crazy filter logic that would be better to avoid, the approach of a separate meta key will be much simpler.

This way we don't need to bother with WP core's `_wp_attachment_backup_sizes` and can get rid of this entire function.

felixarntz (commented):
@mitogh Left a few comments. High-level summary:
* We can simplify this implementation by introducing our own `_wp_attachment_backup_sources` meta key.
* This meta key will be used to store the `sources` **only** for the original image (since WP core otherwise wouldn't store it).
* The other `sources` are automatically backed up by WP core, since it just copies everything over for the sub-sizes. That actually already works today, which is nice to see.
* When restoring, we then only need to include logic to restore the original image `sources`.
* We don't need to delete any image files because edited images will not have any `sources` added (that'll be part of #158). In other words, there will not be anything to delete other than what WP core already handles.

modules/images/webp-uploads/load.php
mitogh: Agree with this approach, the main idea behind this initial approach was to maximize what we already have and avoid introducing more meta data keys. Will update the code with the suggested approach instead.

modules/images/webp-uploads/load.php
mitogh: > This is incorrect I believe. You're hooking into wp_update_attachment_metadata, so the sources here will be the original ones - which we must not delete under any circumstances.

This is correct, this function is fired by `wp_update_attachment_metadata` when this function is executed the passed values are still not written into the database, so the current state on the database: `$metadata = wp_get_attachment_metadata( $attachment_id );` is the state before we overwrite the values.  `$data` is the values we are going to write into the database.

So based on the above we look into the data that is going to be replaced before is overwritten so any edited version of the files should be removed if `IMAGE_EDIT_OVERWRITE` in order to make sure we follow the same pattern as WordPress already does with the subsizes. 

So after that, we restore returning the `$data` variable that's going to be used instead to write on the meta data of the image.

modules/images/webp-uploads/load.php
mitogh: > I actually think this entire block of code can be removed. We don't support generating additional MIME types for image edits yet, so none of the edited images (neither the original nor any sizes) will have any sources that we'd need to delete.

That's correct we currently don't have support for that but the existing issue #158 only handles the creation of those edited version images, this logic already covers the fact that you edited an image and then you restored, I agree that this logic right now won't be executed since we need #158 completed, however, this already works we can either move this code into a separate PR that's going to be related with #228 and #158 or we can create one to keep track of that, let me know what do you prefer but to me it seems reasonable to include it here already as it's part of the restore logic from #228.

Here's a test case covering that scenario: 

- https://github.com/WordPress/performance/blob/4820501b979d4e710baed19bd75d4597b628562b/tests/modules/images/webp-uploads/webp-uploads-test.php#L988

modules/images/webp-uploads/load.php
mitogh: > On another note though, we can't rely on $backup_sizes['full-orig']['sources'] as that won't be set. When you review the code from WP core's wp_save_image() and wp_restore_image() closely, you will see that they manually assemble the full-orig array, which means it'll never include the sources. Only for the sub-sizes it copies/moves over the original data, so the sources there are maintained.

Right this logic is based on the fact that we ensure there's always a `sources` attributes for `full-orig` image. 

> In other words, what we need to do here is rely on our custom storage (e.g. the _wp_attachment_backup_sources meta key I proposed above, or your current implementation _sources).

Great let me take a look at this approach instead which would mean this section of code need to be updated.

modules/images/webp-uploads/load.php
mitogh: > This is a bit problematic in terms of core, since core supports multiple revisions for edited images, which this code here doesn't.

Agree, however, this is already covered by this conditional.

- https://github.com/WordPress/performance/blob/4820501b979d4e710baed19bd75d4597b628562b/modules/images/webp-uploads/load.php#L838

> Get the current value of _wp_attachment_backup_sources into $backup_sources. Fall back to an empty array if not set.

I would argue that we should fallback to not set the `sources` attributes if we don't have any data to process rather than an empty array, however I do see the benefit for consistency, so let me do that.

> Define a suffix variable similar like in core, e.g. $suffix = time() . rand( 100, 999 ).

I don't think we should do that, instead we should apply this logic instead to determine the next available slot for full size images.

https://github.com/WordPress/performance/blob/4820501b979d4e710baed19bd75d4597b628562b/modules/images/webp-uploads/load.php#L842-L855

The main reason is that would give us the next available slot for full size images instead, if we create a custom hash we are not going to be able to determine which sources belong to which image sizes as if you perform multiple images multiple edited versions exists so in the case you performed and saved 5 edited operations and `IMAGE_EDIT_OVERWRITE` is not defined you would end up with 5 different `full-{hash or orig for the fist time}` so creating a custom hash from our end would not be beneficial and prevent us from relate the correct full versions of the edited images instead we should use the hash already stored on the backup image sizes.

> If $backup_sources['sources-orig'] already exists, set the current sources in $backup_sources["sources-$suffix"]. Otherwise (no backup exists yet), use $backup_sources['sources-orig'].

This logic is covered and answered on the previous item.

> Unset sources from the original metadata entirely (since we don't yet support generating sources for image edits, which will be part of https://github.com/WordPress/performance/issues/158).
Set _wp_attachment_backup_sources with the updated $backup_sources.

Right this is already covered here: https://github.com/WordPress/performance/blob/4820501b979d4e710baed19bd75d4597b628562b/modules/images/webp-uploads/load.php#L743

modules/images/webp-uploads/load.php
felixarntz: Ahh, yes you're right, sorry. I thought you were acting on the `$metadata` passed into the function, but when using the original metadata from before the update that's okay.

In any case, I'm not sure we still need this, per my other comments.

modules/images/webp-uploads/load.php
felixarntz: I think it would be better to move this out of here, to keep it simple. It makes more sense to have another PR against #158 for that IMO, since it is only needed if we also implement the other bit of #158. Including it here is a bit preliminary.

modules/images/webp-uploads/load.php
felixarntz: > I would argue that we should fallback to not set the `sources` attributes if we don't have any data to process rather than an empty array, however I do see the benefit for consistency, so let me do that.

The backup sources would be what's already backed up. If they're not set/empty, that just means there is no backup yet, but we still have to do something on that data, populate it the new backup sources.

> I don't think we should do that, instead we should apply this logic instead to determine the next available slot for full size images.

Good point, that makes sense to me.

modules/images/webp-uploads/load.php
mitogh: Sounds good just removed this section of code and added a `TODO` so we can address it once #158 is completed.

mitogh:
Thanks for this review @felixarntz just updated this PR based on your feedback, let me know in case you have any additional comments or feedback.

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,159 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+			case 'wp_restore_image':
+				// When an image has been restored.
+				return webp_uploads_restore_image( $attachment_id, $data );
+		}
+	}
+
+	return $data;
+}
+
+add_filter( 'wp_update_attachment_metadata', 'webp_wp_update_attachment_metadata', 10, 2 );
```
```suggestion
add_filter( 'wp_update_attachment_metadata', 'webp_wp_update_attachment_metadata', 10, 2 );
```

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,159 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ *
+ * @since n.e.x.t
+ *
+ * @see wp_update_attachment_metadata()
+ *
+ * @param array $data The current metadata of the attachment.
+ * @param int   $attachment_id The ID of the current attachment.
+ * @return array The updated metadata for the attachment to be stored in the meta table.
+ */
+function webp_wp_update_attachment_metadata( $data, $attachment_id ) {
```
felixarntz: Let's keep the `webp_uploads` prefix for all functions throughout.
```suggestion
function webp_uploads_update_attachment_metadata( $data, $attachment_id ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,159 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ * metadata back into the full size similar as how it's done on the rest of the sizes.
+ *
+ * @since n.e.x.t
+ *
+ * @param int    $meta_id The ID of the meta value stored in the DB.
+ * @param int    $attachment_id The ID of the post used for this meta, in our case the attachment ID.
+ * @param string $meta_name The name of the metadata to be updated.
+ * @param array  $backup_sizes An array with the metadata value in this case the backup sizes.
+ */
+function webp_updated_postmeta( $meta_id, $attachment_id, $meta_name, $backup_sizes ) {
```
```suggestion
function webp_uploads_updated_postmeta( $meta_id, $attachment_id, $meta_name, $backup_sizes ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,159 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+	$backup_sources[ $target ] = $backup_sources['_sources'];
+	// Remove the temporary used space.
+	unset( $backup_sources['_sources'] );
+	// Store the `sources` property into the full size if present.
+	update_post_meta( $attachment_id, '_wp_attachment_backup_sources', $backup_sources );
+}
+
+add_action( 'added_post_meta', 'webp_updated_postmeta', 10, 4 );
+add_action( 'updated_post_meta', 'webp_updated_postmeta', 10, 4 );
```
```suggestion
}
add_action( 'added_post_meta', 'webp_updated_postmeta', 10, 4 );
add_action( 'updated_post_meta', 'webp_updated_postmeta', 10, 4 );
```

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,159 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+	// When an edit to an image is only applied to a thumbnail there's nothing we need to backu up.
+	if ( 'thumbnail' === $target ) {
+		return $data;
+	}
+
+	// Store the current sources before the current metadata is overwritten.
+	$sources             = get_post_meta( $attachment_id, '_wp_attachment_backup_sources', true );
+	$sources             = is_array( $sources ) ? $sources : array();
+	$sources['_sources'] = $data['sources'];
```
felixarntz: Again, I'm not sure I follow how this supports history. By storing the sources unconditionally in a hard-coded `_sources`, there will always ever be one `_sources` stored. WordPress core supports multiple history iterations in its backup sizes, so we should follow that here. Basically, we should store something like `sources-orig`, `sources-{randomSuffix}`, etc.

Also, why are we looking at `$data` here?
* This function is called when the image has been edited, within `wp_save_image()`. The `$data` will be the _new_ attachment metadata after the edit.
* This is not what we need to back up. We need to back up the sources _prior to_ the edit. In other words, you'll need to get the previous attachment metadata here and back up the sources from there.

modules/images/webp-uploads/load.php
felixarntz: I still don't understand the need for all of this. We already back up the full image's sources in `webp_uploads_backup_sources()`, so why set that post meta value again here? Why do we need to do _anything_ when `_wp_attachment_backup_sizes` is updated like you're doing here?

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,159 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+			continue;
+		}
+		// If the target already has the sources attributes find the next one.
+		if ( isset( $backup_sources[ $size_name ] ) ) {
+			continue;
+		}
+
+		$target = $size_name;
+		break;
+	}
```
felixarntz: Are you sure this code is correct? It looks like backup sizes and backup sources got somehow mixed up in here. If this is indeed like you intended, can you please explain your thinking process? I don't understand why this function is needed.

I think the above backup and restore functions hooked into `wp_update_attachment_metadata` should be sufficient.

felixarntz (commented):
@mitogh Left a few follow-up comments, I still have some open questions, and a few small nit-picks.

modules/images/webp-uploads/load.php
mitogh: > Again, I'm not sure I follow how this supports history. By storing the sources unconditionally in a hard-coded _sources, there will always ever be one _sources stored.

Basically, the way this works in WP is it updates the metadata of the image first and last it would update the backup metadata.

https://github.com/WordPress/wordpress-develop/blob/2648a5f984b8abf06872151898e3a61d3458a628/src/wp-admin/includes/image-edit.php#L981-L982

So at this point, we don't know if we have any backup already, so first we store the current sources for the full image before is updated then after the backup sizes is populated we retrieve those values and store it into the right meta.

The reason for this is that at this point (before saving the backup sizes) we don't know if we already have a `full-orig` or a `full-hash` value and we can't do that after either because at that point the metadata was already overwritten so instead we first store the values we need to preserve. 

Then after the values are stored in the backup we move it to the correct location here:

- https://github.com/WordPress/performance/blob/ba0cbdf45808d8844fbd371da66e70b57ff5cdd3/modules/images/webp-uploads/load.php#L854

> Also, why are we looking at $data here?

Great point here we should back up the current metadata as you described at this point the `$data` would be the new values. So let me update that one.

modules/images/webp-uploads/load.php
mitogh: This is triggered once we know we have sizes on the backup which would allow us to find the right target for the `sources` of the full sizes images. As there can be several versions like:

- `full-orig`
- `full-{hash}`

By hooking after the backup size was created we can determine the correct full target to be allocated for the previous stored `sources` in the `_sources` key of the metadata we are using.

modules/images/webp-uploads/load.php
mitogh: Sure.

As the functions are fired in the following order:

- updated of the metadata of images
- updated of the backup sizes

- https://github.com/WordPress/wordpress-develop/blob/2648a5f984b8abf06872151898e3a61d3458a628/src/wp-admin/includes/image-edit.php#L981-L982


So before the metadata of images is updated we need to store the current `sources` value at the top level of the metadata which at this point is stored in `_sources` key so we can reference it later.

Then after the backup sizes are updated we then update the `_wp_attachment_backup_sources` key, this is done in this way because at this stage we know that all the previous metadata has been stored in the backup sizes.

The first time that this happens it would be stored in the `size-name-orig` key the subsequent backup of images would be stored in `size-name-{hash}`. So for `_wp_attachment_backup_sources` we need to reference only the `sources` of the full image size, in order for us to do that we need to find the appropriate key so for example the first time we are going to store only on `full-orig` the next time we are going to use the next `full-{hash}` value.

I see the point of just storing on `wp_update_attachment_metadata` and if `full-orig` is already set we just store it on a next hash and move on, however by doing that we miss the history point in time where an edit took place specifically for the full image size, by us keeping track of the right hash key we can determine with certain where an edit is related with the sources.

Having this relationship does not have any value yet, because by default we either restore everything or delete everything which means we just pull the `full-orig` or remove everything so the order as it is currently in WordPress does not matter much. 

Let me know if you are comfortable with that approach were there is no clear relationship between edit version and image sources. I see the value that would simplify this logic but would introduce the lack of relationship, which at this point is not used in either this plugin or WP Core.

felixarntz: @mitogh Ah now I see what you mean, it's about picking the right hash/suffix. That makes sense to me. I agree it may not be super necessary, but on the other hand, for consistency it indeed makes sense to store it.

I think we can potentially solve this in a less confusing way code-wise, by adding the hooks for the `_wp_attachment_backup_sizes` meta update _within_ the `webp_uploads_backup_sources()` function. We can also avoid storing data in `_wp_attachment_backup_sources` twice and instead pass it to the new function via PHP's `use`.

mitogh: Thanks for taking a look, absolutely let me take a look at this approach I agree it would simplify things as well the only downside of that approach tho is the fact that would limit the compatibility of the plugin as if we are using `use` PHP 7 would be required but let me verify that.

mitogh: Just take a look and this one and that's right we do need to bump the support in order to make this change to PHP 7.1 I don't think it makes sense as the benefit is only in the structure of the code while cutting support for installations running previous versions of PHP.


But let me know what do you think about that, thanks.

felixarntz: PHP's `use` for functions has been supported for quite a long time. I'm not exactly sure when but for sure that is supported in PHP 5.6, maybe even older versions than that. Where did you see it requires PHP 7.1?

mitogh: Great, Looks like is available since PHP 5.3 I was confusing it with the changelog on PHP docs

- https://www.php.net/functions.anonymous

mitogh: Thanks, just updated the code to match your suggested approach instead.

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,187 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+	$sources = $metadata['sources'];
+	// Prevent execution of the callbacks more than once if the callback was already executed.
+	$executed = false;
+
+	add_action(
+		'added_post_meta',
+		function ( $meta_id, $attachment_id, $meta_name, $backup_sizes ) use ( $sources, &$executed ) {
+			// The backup sources array.
+			if ( '_wp_attachment_backup_sizes' !== $meta_name || $executed ) {
```
felixarntz: Let's also add a check here to ensure the attachment ID here is the same as the outer `$attachment_id`. Otherwise we may accidentally adjust data for the wrong image. It's unlikely that this will be called for another image in the same request, but it's certainly possible. We have to write resilient code to prevent any issues like that.

Please rename the `$attachment_id` parameter of this closure to something else so that you can then compare it to the outer `$attachment_id`.

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,187 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+		return $data;
+	}
+
+	$sources = $metadata['sources'];
+	// Prevent execution of the callbacks more than once if the callback was already executed.
+	$executed = false;
+
+	add_action(
+		'added_post_meta',
+		function ( $meta_id, $attachment_id, $meta_name, $backup_sizes ) use ( $sources, &$executed ) {
```
felixarntz: We don't use the `$backup_sizes` here, so let's not pass it.

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,187 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+			// The backup sources array.
+			if ( '_wp_attachment_backup_sizes' !== $meta_name || $executed ) {
+				return;
+			}
+
+			$executed = true;
+			webp_uploads_backup_full_image_sources( $attachment_id, $sources );
+		},
+		10,
+		4
```
felixarntz: See above, let's then change this to `3`.

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,187 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	add_action(
+		'updated_post_meta',
+		function ( $meta_id, $attachment_id, $meta_name, $backup_sizes ) use ( $sources, &$executed ) {
+			// The backup sources array.
+			if ( '_wp_attachment_backup_sizes' !== $meta_name || $executed ) {
+				return;
+			}
+			$executed = true;
+			webp_uploads_backup_full_image_sources( $attachment_id, $sources );
+		},
```
felixarntz: Let's put this whole closure into a variable so that we can reuse it between both `add_action` calls. This is unnecessary duplicate code right now.

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,187 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+		10,
+		4
```
felixarntz: See above, this should then also become `3`.

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,187 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	// TODO: Requires to be updated from https://github.com/WordPress/performance/issues/158.
+	unset( $data['sources'] );
+
+	return $data;
+}
+
+/**
+ * Stores the provided sources for the attachment ID in the `_wp_attachment_backup_sources`  with
+ * the next available target if target is `null` no source would be stored.
+ *
```
felixarntz, line 805: Remember to provide `@since` annotations.
```suggestion
 * the next available target if target is `null` no source would be stored.
 *
 * @since n.e.x.t
 *
```

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,187 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	$backup_sources[ $target ] = $sources;
+	// Store the `sources` property into the full size if present.
+	update_post_meta( $attachment_id, '_wp_attachment_backup_sources', $backup_sources );
+}
+
+/**
+ * It finds the next available `full-{orig or hash}` key on the images if the name
+ * has not been used as part of the backup sources it would be used if no size is
+ * found or backup exists `null` would be returned instead.
+ *
```
felixarntz, line 832: See above.

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,187 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	foreach ( array_keys( $backup_sizes ) as $size_name ) {
+		// We are only interested in the `full-` sizes.
+		if ( strpos( $size_name, 'full-' ) === false ) {
+			continue;
+		}
+
+		// If the target already has the sources attributes find the next one.
+		if ( isset( $backup_sources[ $size_name ] ) ) {
+			continue;
+		}
```
felixarntz: Nice catch! This ensures we always get the target that is still missing for our data 💯

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,187 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ * Inspect if the current call to `wp_update_attachment_metadata()` was done from within the context
+ * of an edit to an attachment either restore or other type of edit, in that case we perform operations
+ * to save the sources properties, specifically for the `full` size image due this is a virtual image size.
+ *
+ * @since n.e.x.t
+ *
+ * @see wp_update_attachment_metadata()
+ *
+ * @param array $data The current metadata of the attachment.
+ * @param int   $attachment_id The ID of the current attachment.
```
felixarntz: Please indent parameter docs correctly, so that they are aligned:
```suggestion
 * @param array $data          The current metadata of the attachment.
 * @param int   $attachment_id The ID of the current attachment.
```

Please also do this for the other function doc comments in this PR.

felixarntz (commented):
@mitogh This looks great now from a functionality perspective! I left a few comments on cleaning up a few things, we can simplify some code and reuse the function instead of rewriting it twice.

The main point of feedback here is that we also should do an extra security check to ensure we're updating the correct attachment.

modules/images/webp-uploads/load.php
mitogh: Good point, thanks.

felixarntz (approved):
@mitogh Awesome work!

cc @adamsilverstein for incorporating this into the core patch (in there we can do it without that hook weirdness present here)

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,184 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ *
+ * @see wp_update_attachment_metadata()
+ *
+ * @param array $data          The current metadata of the attachment.
+ * @param int   $attachment_id The ID of the current attachment.
+ * @return array The updated metadata for the attachment to be stored in the meta table.
+ */
+function webp_uploads_update_attachment_metadata( $data, $attachment_id ) {
+
+	$trace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS, 10 );
```
eugene-manuilov, line 722: Since this function is intended only for ajax requests with the `image-editor` action, we can return early if we are not doing the `wp_ajax_image-editor` action:
```suggestion
	if ( ! doing_action( 'wp_ajax_image-editor' ) ) {
		return;
	}

	$trace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS, 10 );
```

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,184 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	if ( ! isset( $metadata['sources'] ) ) {
+		return $data;
+	}
+
+	$target = isset( $_REQUEST['target'] ) ? sanitize_text_field( $_REQUEST['target'] ) : 'all';
+
+	// When an edit to an image is only applied to a thumbnail there's nothing we need to back up.
+	if ( 'thumbnail' === $target ) {
+		return $data;
+	}
```
eugene-manuilov: The `target` value check is less expensive than metadata sources check. Let's swap these conditions to make sure we don't check metadata sources for thumbnail targets:
```php
	$target = isset( $_REQUEST['target'] ) ? sanitize_text_field( $_REQUEST['target'] ) : 'all';

	// When an edit to an image is only applied to a thumbnail there's nothing we need to back up.
	if ( 'thumbnail' === $target ) {
		return $data;
	}

	$metadata = wp_get_attachment_metadata( $attachment_id );
	// Nothing to back up.
	if ( ! isset( $metadata['sources'] ) ) {
		return $data;
	}
```

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,184 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ * @return array The updated metadata for the attachment.
+ */
+function webp_uploads_backup_sources( $attachment_id, $data ) {
+	$metadata = wp_get_attachment_metadata( $attachment_id );
+	// Nothing to back up.
+	if ( ! isset( $metadata['sources'] ) ) {
+		return $data;
+	}
+
+	$target = isset( $_REQUEST['target'] ) ? sanitize_text_field( $_REQUEST['target'] ) : 'all';
```
eugene-manuilov: The `sanitize_text_field` function call is redundant here. Let's get rid of it:
```suggestion
	$target = isset( $_REQUEST['target'] ) ? $_REQUEST['target'] : 'all';
```

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,184 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ *
+ * @param int   $attachment_id The ID of the attachment.
+ * @param array $sources       An array with the full sources to be stored on the next available key.
+ */
+function webp_uploads_backup_full_image_sources( $attachment_id, $sources ) {
+	$target = webp_uploads_get_next_full_size_key_from_backup( $attachment_id );
+
+	if ( null === $target || empty( $sources ) ) {
+		return;
+	}
```
eugene-manuilov: If sources are empty, we don't need to call the `webp_uploads_get_next_full_size_key_from_backup` function. Let's move the sources check to be done before calling the function:
```php
	if ( empty( $sources ) ) {
		return;
	}

	$target = webp_uploads_get_next_full_size_key_from_backup( $attachment_id );
	if ( null === $target ) {
		return;
	}
```

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,184 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+	$backup_sources            = get_post_meta( $attachment_id, '_wp_attachment_backup_sources', true );
+	$backup_sources            = is_array( $backup_sources ) ? $backup_sources : array();
```
eugene-manuilov, line 822: Technically saying we don't need this line because if we get here, it means that backup sources is already a non-empty array because otherwise the `webp_uploads_get_next_full_size_key_from_backup` would return null.

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,184 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	foreach ( array_keys( $backup_sizes ) as $size_name ) {
+		// We are only interested in the `full-` sizes.
+		if ( strpos( $size_name, 'full-' ) === false ) {
+			continue;
+		}
+
+		// If the target already has the sources attributes find the next one.
+		if ( isset( $backup_sources[ $size_name ] ) ) {
+			continue;
+		}
```
eugene-manuilov: The `isset` check is less expensive than the `strpos` function call. Let's change the order to run the less expensive check first:
```php
		// If the target already has the sources attributes find the next one.
		if ( isset( $backup_sources[ $size_name ] ) ) {
			continue;
		}

		// We are only interested in the `full-` sizes.
		if ( strpos( $size_name, 'full-' ) === false ) {
			continue;
		}
```

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,184 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ * @param int   $attachment_id The ID of the attachment.
+ * @param array $data          The current metadata to be stored in the attachment.
+ * @return array The updated metadata of the attachment.
+ */
+function webp_uploads_restore_image( $attachment_id, $data ) {
+	$backup_sources = get_post_meta( $attachment_id, '_wp_attachment_backup_sources', true );
+
+	if ( ! is_array( $backup_sources ) || ! isset( $backup_sources['full-orig'] ) || ! is_array( $backup_sources['full-orig'] ) ) {
+		return $data;
+	}
```
eugene-manuilov: Maybe let's split it into 3 checks to make it more readable?

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -809,6 +811,166 @@ function( $transforms ) {
...
+	public function it_should_restore_the_sources_array_from_the_backup_when_an_image_is_edited() {
+		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg' );
+		$metadata      = wp_get_attachment_metadata( $attachment_id );
+		$editor        = new WP_Image_Edit( $attachment_id );
+		$editor->rotate_right()->save();
+		$this->assertTrue( $editor->success() );
+		$backup_sources = get_post_meta( $attachment_id, '_wp_attachment_backup_sources', true );
+		$this->assertArrayHasKey( 'full-orig', $backup_sources );
+		$this->assertIsArray( $backup_sources['full-orig'] );
+		$this->assertSame( $metadata['sources'], $backup_sources['full-orig'] );
```
eugene-manuilov: Let's add some whitespace here to make it easier to read and follow:
```suggestion
		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg' );
		$metadata      = wp_get_attachment_metadata( $attachment_id );

		$editor = new WP_Image_Edit( $attachment_id );
		$editor->rotate_right()->save();
		$this->assertTrue( $editor->success() );

		$backup_sources = get_post_meta( $attachment_id, '_wp_attachment_backup_sources', true );
		$this->assertArrayHasKey( 'full-orig', $backup_sources );
		$this->assertIsArray( $backup_sources['full-orig'] );
		$this->assertSame( $metadata['sources'], $backup_sources['full-orig'] );
```

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -836,4 +998,130 @@ function () {
...
+	 */
+	public function it_should_store_the_metadata_on_the_next_available_hash() {
+		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg' );
+		$editor        = new WP_Image_Edit( $attachment_id );
+		$editor->rotate_right()->save();
+		$this->assertTrue( $editor->success() );
+		// Remove the filter to prevent the usage of the next target.
+		remove_filter( 'wp_update_attachment_metadata', 'webp_uploads_update_attachment_metadata' );
+		$editor->rotate_right()->save();
+		$this->assertTrue( $editor->success() );
```
eugene-manuilov: Same here:
```suggestion
		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg' );

		$editor = new WP_Image_Edit( $attachment_id );
		$editor->rotate_right()->save();
		$this->assertTrue( $editor->success() );

		// Remove the filter to prevent the usage of the next target.
		remove_filter( 'wp_update_attachment_metadata', 'webp_uploads_update_attachment_metadata' );
		$editor->rotate_right()->save();
		$this->assertTrue( $editor->success() );
```

eugene-manuilov (changes requested):
Nice work, @mitogh! Added some comments with minor improvements. Please, let me know whether it makes sense to you.

modules/images/webp-uploads/load.php
mitogh: Agree on that one :+1:

modules/images/webp-uploads/load.php
mitogh: Actually, this would be `true` the first time this is executed as the meta key won't exist and in that case, the returned value would be an empty string, in that case, we initialize it to an array the first time to setup the data structure.

modules/images/webp-uploads/load.php
mitogh: I think we might need to add a skip of the codesniffer but agree as this value is never stored in the DB is not required.

modules/images/webp-uploads/load.php
mitogh: Right we might need to update the test to work in that scenario but I agree that this can be safely introduced, let me verify on my end as well.

modules/images/webp-uploads/load.php
mitogh: I've split this in 2 conditionals as the last conditional uses the same key.

modules/images/webp-uploads/load.php
felixarntz: @eugene-manuilov @mitogh I'd prefer we avoid adding that condition. Yes, it's only called by the AJAX function in WordPress right now, but our code should remain agnostic of that if at all possible. If someone calls e.g. `wp_save_image()` in whatever other context, our code here should still do its job.

mitogh: Good point, this would help with tests too as I was spending some time already to pass the tests under this scenario.

eugene-manuilov: > If someone calls e.g. `wp_save_image()` in whatever other context, our code here should still do its job.

@felixarntz, I think that it's very unlikely that someone will use this function for their own purposes because that function is tough coupled with the logic implemented in the image editor and heavily relies on various request variables using a lot of `$_REQUEST[...]` calls.

The matter here is about consistency, I think... In some places, we fight for minor improvements to get a super tiny performance gain, in other places we are generous enough to leave a redundant function call that parses the stack trace and generates useless information.

eugene-manuilov (approved):
LGTM! Thanks, @mitogh.

modules/images/webp-uploads/load.php
felixarntz: @eugene-manuilov I agree in a few PRs so far some discussions were overly nit-picking, let's try to avoid "do we use `array_map` or `foreach`" type of discussions.

But in terms of implementation scope I'd always advocate for maximum compatibility when working with core. Because of the (lack of) core architecture in many places anything can happen, and it's important we don't only consider how core uses a function but also that a plugin could use it in a different context. For example just because a function is only called by core during an AJAX or REST request doesn't mean we should only support that case - the only exception to that would be of course functions that are explicitly named/marked as such (like the `wp_ajax_*()` functions).

PR #243, opened 2022-03-21
Select MIME type to use in frontend content based on file size

### LINKED ISSUE #186: Select the smaller image size from the list of available sources

jjgrainger:
## Summary

Fixes #186 

## Relevant technical choices

- Branched from `feature/187-content-images-format` PR #230 

## Checklist


tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -513,6 +513,31 @@ public function it_should_replace_the_references_to_a_jpg_image_to_a_webp_versio
...
+				return $mime_types;
+			}
+		);
+
+		$attachment_id = $this->factory->attachment->create_upload_object( $image_path );
+		$tag           = wp_get_attachment_image( $attachment_id, 'medium', false, array( 'class' => "wp-image-{$attachment_id}" ) );
+
+		$this->assertSame( $tag, webp_uploads_img_tag_update_mime_type( $tag, 'the_content', $attachment_id ) );
+	}
+
```
mitogh: Can we add some tests for filesize, you can use:

- https://github.com/WordPress/performance/blob/trunk/tests/testdata/modules/images/paint.jpeg 

On the tests that one creates smaller JPEGS than WebP images. To verify the selection process works, I can suggest other tests where you override the `metadata` information and set the filesize to 0 as well to ensure this works as expected.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +705,79 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+			}
+
+			return ( $a['filesize'] < $b['filesize'] ) ? -1 : 1;
+		}
+	);
+
+	// Create an array of available mime types ordered by smallest filesize.
+	$mime_types = array_values( array_keys( $sources ) );
+
+	return $mime_types;
```
```suggestion
	// Create an array of available mime types ordered by smallest filesize.
	return array_values( array_keys( $sources ) );
```

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +705,79 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+	if ( empty( $sources ) ) {
+		return $mime_types;
+	}
+
+	// Remove mime types with filesize of 0.
+	$sources = array_filter(
+		$sources,
+		function( $source ) {
+			return $source['filesize'] > 0;
```
mitogh: Make sure the key is present.

```suggestion
			return isset( $source['filesize'] ) && $source['filesize'] > 0;
```

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +705,79 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+	// Return an empty array if no sources found.
+	return array();
+}
+
+/**
+ * Filters on `webp_uploads_content_image_mimes` to generate the available mime types for an attachment
+ * and orders them by the smallest filesize first.
+ *
+ * @since n.e.xt
+ * @private
```
mitogh: I think this one does not qualify as `private` due is hooking into a `public` filter.

```suggestion
```

mitogh, line 743: I think this one does not qualify as `private` due is hooking into a `public` filter.

```suggestion
```

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +705,79 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+
+	// Order sources by filesize in ascending order.
+	uasort(
+		$sources,
+		function( $a, $b ) {
+			if ( $a['filesize'] === $b['filesize'] ) {
+				return 0;
+			}
+
+			return ( $a['filesize'] < $b['filesize'] ) ? -1 : 1;
```
eugene-manuilov:
```suggestion
			return $a['filesize'] - $b['filesize'];
```

modules/images/webp-uploads/load.php
eugene-manuilov: No need to use the `array_values` function here since the `array_keys` function already returns an array of values with sequential numeric keys.

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +705,79 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
 	return rest_ensure_response( $data );
 }
 add_filter( 'rest_prepare_attachment', 'webp_uploads_update_rest_attachment', 10, 3 );
+
+/**
+ * Returns the attachment sources array ordered by filesize.
+ *
+ * @since n.e.xt
```
eugene-manuilov:
```suggestion
 * @since n.e.x.t
```

modules/images/webp-uploads/load.php
```diff
@@ -683,3 +705,79 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ * @private
```
eugene-manuilov: No need to use the `@private` tag since this function is public. Private functions in WordPress start with `_`, so if you want to make it private then it should be `_webp_uploads_get_attachment_sources` (not sure if it is really needed though).

eugene-manuilov (changes requested):
Mostly looks good to me. Added a few comments with a few nitpicks, nothing major. Also, do we need to target this PR against the `release/1.0.0-beta.x` branch?

modules/images/webp-uploads/load.php
eugene-manuilov: Same for the `webp_uploads_get_mime_types_by_filesize` function.

modules/images/webp-uploads/load.php
jjgrainger: Updated return and removed `array_values`

tests/modules/images/webp-uploads/webp-uploads-test.php
jjgrainger: @mitogh I found the paint.jpeg file still returned a smaller webp version in my testing.

I've opted for using the `wp_get_attachment_metadata` filter to mock the `sources` array and set the `image/jpeg` mime type to be a smaller filesize.

I've also used the same filter to mock a mime type with a filesize of 0.

modules/images/webp-uploads/load.php
jjgrainger: I've removed the `@private` tag as this function could be useful publicly.

jjgrainger:
> Mostly looks good to me. Added a few comments with a few nitpicks, nothing major. Also, do we need to target this PR against the `release/1.0.0-beta.x` branch?

I agree, though not sure which release branch it should target? There are currently 3.

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,71 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
 add_filter( 'rest_prepare_attachment', 'webp_uploads_update_rest_attachment', 10, 3 );
+
+/**
+ * Returns the attachment sources array ordered by filesize.
+ *
+ * @since n.e.x.t
+ *
+ * @param int    $attachment_id The attachment ID.
+ * @param string $size          The attachment size.
+ *
```
eugene-manuilov: We don't use a blank line between params and the return tag, so let's remove it to be consistent with other functions.
```suggestion
```

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,71 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+/**
+ * Filters on `webp_uploads_content_image_mimes` to generate the available mime types for an attachment
+ * and orders them by the smallest filesize first.
+ *
+ * @since n.e.x.t
+ *
+ * @param array  $mime_types    The list of mime types that can be used to update images in the content.
+ * @param int    $attachment_id The attachment ID.
+ * @param string $context       The current context.
+ *
```
eugene-manuilov: Same here:
```suggestion
```

modules/images/webp-uploads/load.php
```diff
@@ -704,3 +704,71 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+ * @return array Array of available mime types ordered by filesize.
+ */
+function webp_uploads_get_mime_types_by_filesize( $mime_types, $attachment_id, $context ) {
```
eugene-manuilov: The context parameter seems to be redundant for this function. Let's not use it then:

```php
function webp_uploads_get_mime_types_by_filesize( $mime_types, $attachment_id ) {
    ...
}
add_filter( 'webp_uploads_content_image_mimes', 'webp_uploads_get_mime_types_by_filesize', 10, 2 );
```

eugene-manuilov (changes requested):
Almost looks good to me. Just added a few nitpicks.

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -808,4 +808,95 @@ function( $transforms ) {
...
+	 */
+	public function it_should_return_smaller_webp_mime_type() {
+		// File should generate smallest webp image size.
+		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/car.jpeg' );
+
+		$mime_types = webp_uploads_get_mime_types_by_filesize( array( 'image/jpeg', 'image/webp' ), $attachment_id, 'the_content' );
+
+		$this->assertIsArray( $mime_types );
+		$this->assertSame( 'image/webp', $mime_types[0] );
+		$this->assertSame( 'image/jpeg', $mime_types[1] );
```
mitogh:
```suggestion
		$this->assertSame( array( 'image/webp', 'image/jpeg' ), $mime_types );
```

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -808,4 +808,95 @@ function( $transforms ) {
...
+	 * @test
+	 */
+	public function it_should_return_smaller_jpeg_mime_type() {
+		// File should generate smallest jpeg image size.
+		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/paint.jpeg' );
+
+		// Mock attachment meta data to test when jpeg image is smaller.
+		add_filter(
+			'wp_get_attachment_metadata',
+			function( $data, $attachment_id ) {
```
mitogh, line 867: Looks like this function is not returning anything it can be safely removed?

But in case is required should we return the `$data` from this callback. ?

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -808,4 +808,95 @@ function( $transforms ) {
...
+			},
+			10,
+			2
+		);
+
+		$mime_types = webp_uploads_get_mime_types_by_filesize( array( 'image/jpeg', 'image/webp' ), $attachment_id, 'the_content' );
+
+		$this->assertIsArray( $mime_types );
+		$this->assertSame( 'image/jpeg', $mime_types[0] );
+		$this->assertSame( 'image/webp', $mime_types[1] );
```
mitogh:
```suggestion
		$this->assertSame( array( 'image/jpeg', 'image/webp' ), $mime_types );
```

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -808,4 +808,95 @@ function( $transforms ) {
...
+	 */
+	public function it_should_remove_mime_types_with_zero_filesize() {
+		// File should generate smallest jpeg image size.
+		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/paint.jpeg' );
+
+		// Mock attachment meta data to test mime type with zero filesize.
+		add_filter(
+			'wp_get_attachment_metadata',
+			function( $data, $attachment_id ) {
+				$data['sources'] = array(
```
mitogh, line 904: Same questions as above.

tests/modules/images/webp-uploads/webp-uploads-test.php
```diff
@@ -808,4 +808,95 @@ function( $transforms ) {
...
+			10,
+			2
+		);
+
+		$mime_types = webp_uploads_get_mime_types_by_filesize( array( 'image/jpeg', 'image/webp' ), $attachment_id, 'the_content' );
+
+		$this->assertIsArray( $mime_types );
+		$this->assertNotContains( 'image/invalid', $mime_types );
+		$this->assertSame( 'image/jpeg', $mime_types[0] );
+		$this->assertSame( 'image/webp', $mime_types[1] );
```
mitogh:
```suggestion
		$this->assertSame( array( 'image/jpeg', 'image/webp' ), $mime_types );
```

mitogh (commented):
Great work, just some minor suggestions / questions.

modules/images/webp-uploads/load.php
jjgrainger: Updated

modules/images/webp-uploads/load.php
jjgrainger: Updated

modules/images/webp-uploads/load.php
jjgrainger: Updated and removed the `$context` parameter

tests/modules/images/webp-uploads/webp-uploads-test.php
jjgrainger: Updated

tests/modules/images/webp-uploads/webp-uploads-test.php
jjgrainger: Yes, it should return the `$data`, I've updated this now

tests/modules/images/webp-uploads/webp-uploads-test.php
jjgrainger: Updated

tests/modules/images/webp-uploads/webp-uploads-test.php
jjgrainger: Yes, updated to return the `$data`

tests/modules/images/webp-uploads/webp-uploads-test.php
jjgrainger: Updated

eugene-manuilov (approved):

tests/modules/images/webp-uploads/webp-uploads-test.php
mitogh: Thanks for the update.

mitogh (approved):

mitogh (approved):

PR #250, opened 2022-03-22
Change expected order of items in the `webp_uploads_content_image_mimes` filter

### LINKED ISSUE #237: Decide on order of `webp_uploads_content_image_mimes` filter items
Follow-up on #187: The `webp_uploads_content_image_mimes` filter that was introduced as part of that issue controls which image formats should be preferred as replacements in the frontend content (all conditional based on whether the current attachment has a version in the given image format).

Currently the items in the filter are ordered so that:
* The most backward-compatible image format (i.e. with most widespread browser support) should come first.
* The most progressive image format (i.e. most efficient compression and file sizes) should come last.

Based on various comments (e.g. starting in https://github.com/WordPress/performance/pull/230#issuecomment-1068405922, and also in https://github.com/WordPress/performance/issues/187#issuecomment-1069976158), we still need to discuss what is the most suitable expected order for the filter.

Right now, since we want to pick the most progressive format in the current implementation, we go through the array in reverse, which arguably is confusing. However, a counter-argument is that the most backward-compatible format should come first which may be "safer" and aligns it better with the `webp_uploads_upload_image_mime_transforms` filter for which it definitely makes more sense to have the backward-compatible format first, as it is more essential to have.

We also have to consider how this filter could be leveraged in a potential future where there is `picture` element support and even more progressive image formats like AVIF could be supported. At that point, we would include all of the formats in the frontend based on the order of items in this filter, and JPEG would always be the fallback.

cc @adamsilverstein @getsource

eugene-manuilov:
## Summary

Fixes #237 

## Checklist

### DIFF
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -601,7 +601,7 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
 
 	/**
 	 * Filters mime types that should be used to update all images in the content. The order of
-	 * mime types matters. The last mime type in the list will be used if it is supported by an image.
+	 * mime types matters. The first mime type in the list will be used if it is supported by an image.
 	 *
 	 * @since 1.0.0
 	 *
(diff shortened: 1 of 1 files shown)


felixarntz (approved):
@eugene-manuilov Great, thanks!

mitogh (approved):

PR #251, opened 2022-03-22
Prepare 1.0.0-beta.3 release

### LINKED ISSUE #249: Prepare 1.0.0-beta.3 release
During the `1.0.0-beta.2` release (see #213), we identified a rather critical bug where the default setting value for active modules wouldn't trigger in time for when the plugin actually loads the active modules. This means that on plugin activation, all modules behave as if they're disabled, even though the UI in the settings screen shows they are enabled. See #247.

A temporary workaround for this is to save the settings on the settings screen, but this is a critical flaw that should be addressed promptly, hence a `1.0.0-beta.3` release should be created in a timely manner, with just this bug fix.

* [x] Define scope and tentative release date based on the scope
* [x] Create `release/1.0.0-beta.3` branch closer to the release date
* [x] Prepare the release (**Tuesday, March 22, 2022**)

felixarntz:
## Summary

Fixes #249

## Relevant technical choices

See https://github.com/WordPress/performance/blob/trunk/docs/Releasing-the-plugin.md#preparing-the-release

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "performance",
-  "version": "1.0.0-beta.2",
+  "version": "1.0.0-beta.3",
   "private": true,
   "description": "Performance plugin from the WordPress Performance Group, which is a collection of standalone performance modules.",
   "author": "The WordPress Contributors",
--- a/load.php
+++ b/load.php
@@ -41,7 +41,7 @@ add_action( 'init', 'perflab_register_modules_setting' );
 /**
  * Gets the default value for the performance modules setting.
  *
- * @since n.e.x.t
+ * @since 1.0.0
  */
 function perflab_get_modules_setting_default() {
 	// Since the default relies on some minimal logic that includes requiring an additional file,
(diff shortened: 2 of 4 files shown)


mitogh (approved):

JustinyAhin (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "performance",
-  "version": "1.0.0-beta.2",
+  "version": "1.0.0-beta.3",
```
eugene-manuilov, line 3: Do we really need to maintain the version here? All properties in the package.json file except `devDependencies` and `scripts` are redundant and can be deleted as not relevant to the plugin. 🤔 

```js
{
  "devDependencies": {
    ...
  },
  "scripts": {
    ...
  }
}
```

eugene-manuilov (approved):

package.json
felixarntz: Potentially, yes. Would you mind opening an issue with that proposal?

eugene-manuilov: Added #252

PR #254, opened 2022-03-22
Updated the package.json file to contain only required data

### LINKED ISSUE #252: Clean up the package.json file
## Feature Description

Currently, the `package.json` file contains information that is usually required for publishing npm packages on npmjs.org which is not the case for our plugin. The only needed information in the package.json file is `devDependencies` and `scripts`, everything else is redundant and can be removed to avoid a need to maintain that information in the future. Let's get rid of anything that is not needed for the plugin development and keep our `package.json` file in a good fit 😄

eugene-manuilov:
## Summary

Fixes #252

## Checklist


mitogh (commented):
This looks good, one last thing is the conflict resolution due to updates in `trunk`.

package.json
```diff
@@ -1,23 +1,4 @@
 {
-  "name": "performance",
-  "version": "1.0.0-beta.2",
-  "private": true,
-  "description": "Performance plugin from the WordPress Performance Group, which is a collection of standalone performance modules.",
-  "author": "The WordPress Contributors",
-  "license": "GPL-2.0-or-later",
```
felixarntz, line 7: When running `npm install`, there's a warning on the `license` field missing. There's no harm in keeping that as it will never change, so I'd be in favor of keeping it.

package.json
```diff
@@ -1,23 +1,4 @@
...
-  "keywords": [
-    "performance",
-    "wordpress"
-  ],
-  "homepage": "https://wordpress.org/plugins/performance-lab/",
-  "repository": "git+https://github.com/WordPress/performance.git",
```
felixarntz, line 13: Same here, `npm install` warns about the lack of `repository`. How about we keep it? It'll also never change.

package.json
```diff
@@ -1,23 +1,4 @@
...
-  "bugs": {
-    "url": "https://github.com/WordPress/performance/issues"
-  },
-  "engines": {
-    "node": ">=10.0.0",
-    "npm": ">=6.9.0"
-  },
```
felixarntz, line 20: Is it okay to remove this? I'd say it's useful information to specific which Node and NPM versions are required?

felixarntz (commented):
@eugene-manuilov Left a few comments.

package.json
thelovekesh: @felixarntz IMO the exact idea of Node and NPM version can be extracted from configuration files i.e. `.npmrc` and `.nvmrc`. AFAIK this is useful when you are reading metadata from `package.json` itself. 

If you have come around Heroku, then you would have noticed that, when you deploy a Heroku app, it chooses the Node buildpack according to the `engines` metadata in `package.json`.

In our use case `.nvmrc` and `.npmrc` should be enough.

felixarntz: Thanks @thelovekesh, makes sense.

package.json
eugene-manuilov: Ok, added it back although it is not needed and we can ignore those warnings.

package.json
eugene-manuilov: Yep, also added it back.

mitogh (approved):

felixarntz (approved):

PR #259, opened 2022-03-24
Select image editor based on WebP support instead of always using the default one

### LINKED ISSUE #258: WebP are not created when the default editor does not support WebP
## Bug Description

Given 2 image editors: `WP_Image_Editor_Imagick` and `WP_Image_Editor_GD` and the default is `WP_Image_Editor_Imagick` if the default editor does not supports WebP but `WP_Image_Editor_GD` it does support WebP the tests on Site Health would return that the site supports WebP, however the current implementation does handle this case instead it uses the default image editor regardless of the support.

## Support tickets:

- https://wordpress.org/support/topic/webp-image-not-created-2/
- https://wordpress.org/support/topic/webp-image-not-created/

mitogh:
## Summary

Fixes #258

## Relevant technical choices

The update uses WP internals to allow for a more appropriate editor instead of the default one based only on the image file when the provided mime type would make a difference as the provided mime type might be supported for other editors besides the default one.

## Checklist

### DIFF
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -304,19 +304,6 @@ function webp_uploads_get_upload_image_mime_transforms() {
  * @return array|WP_Error An array with the file and filesize if the image was created correctly otherwise a WP_Error
  */
 function webp_uploads_generate_additional_image_source( $attachment_id, array $size_data, $mime, $destination_file_name = null ) {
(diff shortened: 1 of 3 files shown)


felixarntz (approved):
@mitogh Awesome work catching and fixing this!

adamsilverstein (approved):

PR #262, opened 2022-03-24
Replace images in frontend content without using an additional regular expression

### LINKED ISSUE #238: Allow controlling which image format sources in the frontend can be replaced
Follow-up to #187: The `webp_uploads_content_image_mimes` currently controls which image formats can be used as replacements in the frontend. However, the formats that can currently _be replaced_ in the frontend are still hard-coded - at the moment it is just JPEG (specifically its extensions `jpg` and `jpeg`).

We need to implement a way so that these source image formats can also be controlled through a filter. For example, if somebody implemented support to create WebP versions for PNG uploads, it should be possible to use the WebP Uploads module's logic to replace PNGs as well.

Related is #237, however let's keep these two decoupled. Depending on the decision in #237, it may simply be that the expected order of the filtered items changes, which would be trivial to update here.

eugene-manuilov:
## Summary

Fixes #238

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -623,45 +611,23 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
-			}
-
-			// This is the same as the file we want to replace nothing to do here.
-			if ( $size_data['sources'][ $target_mime ]['file'] === $src_filename ) {
-				continue;
-			}
+	// Replace the full size image if present.
+	if ( isset( $metadata['sources'][ $target_mime ]['file'] ) ) {
+		$basename = wp_basename( $metadata['file'] );
+		$image    = str_replace( $basename, $metadata['sources'][ $target_mime ]['file'], $image );
```
felixarntz: These values could be the same. Just thinking out loud here, in terms of PHP performance (I know it's very minor but still worth thinking about) not sure whether that's okay or whether it would be better to use an if check around them.

modules/images/webp-uploads/load.php
```diff
@@ -623,45 +611,23 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+	foreach ( $metadata['sizes'] as $name => $size_data ) {
+		if (
+			! empty( $size_data['file'] ) &&
+			! empty( $size_data['sources'][ $target_mime ]['file'] )
+		) {
+			$image = str_replace(
+				$size_data['file'],
+				$size_data['sources'][ $target_mime ]['file'],
+				$image
+			);
```
felixarntz: Same here, the search and replace values could be the same. I wonder whether `str_replace` internally does this comparison already and bails early if so, or if we should wrap it 🤔

felixarntz (approved):
@eugene-manuilov Great work! One minor thing, but nothing blocking.

modules/images/webp-uploads/load.php
eugene-manuilov: Added a check to avoid replacing the same images.

modules/images/webp-uploads/load.php
eugene-manuilov: Same here. Updated condition to check whether images are different before replacing them.

felixarntz (approved):
Great!

modules/images/webp-uploads/load.php
```diff
@@ -623,45 +611,30 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
-
-			$image = str_replace( $src_filename, $size_data['sources'][ $target_mime ]['file'], $image );
-			break;
+	// Replace sub sizes for the image if present.
+	foreach ( $metadata['sizes'] as $name => $size_data ) {
+		if (
+			! empty( $size_data['file'] ) &&
+			! empty( $size_data['sources'][ $target_mime ]['file'] ) &&
+			$size_data['file'] !== $size_data['sources'][ $target_mime ]['file']
+		) {
```
mitogh: While technically this is the same group of conditionals as before I would argue that this way it's harder to follow, instead 1 line condition moving on makes for a more natural and simple flow.

mitogh (commented):
Just one comment besides that great refactor.

modules/images/webp-uploads/load.php
eugene-manuilov: I would agree with it if this condition is written on a single line. IMO having every check on its own line within one condition has the same effect as if we would add three separate conditions for every check. However, if more people think that it's hard to follow, I don't mind to update it to have separate conditions.

felixarntz: I think this could be considered a personal preference, but my take here is that I would agree with @mitogh the previous way was easier to follow. I also prefer the `if not x, then bail early` (i.e. `continue`) approach better than `if x, run the logic`, as it avoids excessive nesting. Of course it's just a single if clause here, but I think it's easier to read if the replacement remains directly within the `foreach` loop and the checks happen before as safe guards.

So to summarize my personal take:
* I prefer the way it was before.
* However, my main point would be to reverse the condition and bail early based on that, rather than only running the logic within the if clause.
* I don't have a strong preference on multiple if clauses or a single one, but I tend to agree when split in multiple it's easier to read.

eugene-manuilov: Sounds good to me then. Updated. Thanks, @mitogh and @felixarntz.

mitogh (approved):

felixarntz (approved):

PR #263, opened 2022-03-24
Add Site Health check for Full Page Cache usage

### LINKED ISSUE #220: Add Site Health test for full page caching (advanced cache)
## Feature Description

I previously reported this in trac ([#54423](https://core.trac.wordpress.org/ticket/54423)), but I'm copying it here for visibility:

> A [best practice](https://make.wordpress.org/hosting/handbook/performance/#full-page-cache) for performance is to enable full page caching for a WordPress site. Doing so reduces the time to first byte (TTFB) and thus directly impacts Core Web Vitals. This is especially important on shared hosts which have limited compute power, while also protecting a site from going down due to traffic spikes. 
>
> As with XML Sitemaps, it may make sense for WordPress core to include a basic full page caching solution out of the box which can be extended by the many plugins that provide page caching today. In the meantime, there should at least be a Site Health test with a low-severity `recommended` status to point users to where to go to enable page caching. Since hosts often have built-in caching solutions in place, the Site Health test should be easily filterable so they can provide a link to their specific dashboard or documentation to enable caching which may not be a plugin at all but rather a reverse proxy like Varnish.
>
> In the context of the AMP plugin for WordPress, we've already developed such a test which could be adapted for core. See https://github.com/ampproject/amp-wp/pull/6456. It's an async test which issues three loopback requests to the homepage, and the test will pass if responses include either of the following headers:
> 
> 1. `Cache-Control` with non-zero `max-age`.
> 2. `Expires` with a future time.
> 3. `Age` with a positive number (as is set via a proxy like Varnish).

Since the above was written and, after releasing the page caching site health test in the AMP plugin v2.2, it came to light that the detection was not robust enough. A major improvement followed (https://github.com/ampproject/amp-wp/pull/6849) which accounts for many more scenarios (https://github.com/ampproject/amp-wp/pull/6849#issuecomment-1023762369). These were uncovered with help from @piotrbak in a WP Rocket issue (https://github.com/wp-media/wp-rocket/issues/4638#issuecomment-1011429544).

The page caching code in question: https://github.com/ampproject/amp-wp/blob/f7c7230a27fb23932606f9847bfcf04c83b8ff06/src/Admin/SiteHealth.php#L333-L661

## Scenarios

Via https://github.com/ampproject/amp-wp/pull/6849#issuecomment-1023762369:

<blockquote cite="https://github.com/ampproject/amp-wp/pull/6849#issuecomment-1023762369">

### Failure to Do Loopback Request

With this plugin active:

```php
add_action('template_redirect', function () {
	status_header( 500 );
	exit;
});
```

image

### No Page Caching with Slow Response Time

With this plugin active:

```php
add_action( 'wp_footer', function () {
	sleep(1);
} );
```

image

### No Page Caching with Good Response Time

No special plugins active.

image

### Page Caching Header(s) with Bad Response Time

Test plugin:

```php
add_filter(
	'wp_headers',
	function ( $headers ) {
		$headers['etag'] = '"foo"';
		$headers['last-modified'] = gmdate('r');
		return $headers;
	}
);

add_action( 'wp_footer', function () {
	sleep(1);
} );
```

Note that since page caching response headers are present, the lack of a page caching plugin is not mentioned (since it's likely an external page cache is being used):

image

### Page Caching Header with Good Response Time

Plugin:

```php
add_filter(
	'wp_headers',
	function ( $headers ) {
		$headers['etag'] = '"foo"';
		$headers['last-modified'] = gmdate('r');
		return $headers;
	}
);
```

Note that since page caching response headers are present, the lack of a page caching plugin is not mentioned (since it's likely an external page cache is being used):

image

### Page Caching Plugin with Good Response Time

With WP Super Cache active and with the following plugin active:

```php
add_action( 'wp_footer', function () {
	sleep(1);
} );
```

Note that since the median time is used, the 1-second response time is disregarded:

image

</blockquote>

manuelRod:
## Summary

Fixes [#220 ](https://github.com/WordPress/performance/issues/220)
Trac [#54423](https://core.trac.wordpress.org/ticket/54423)

## Relevant technical choices
As @westonruter proposed, I've adapted and migrated the full page caching Site Health test from [AMP](https://github.com/ampproject/amp-wp/blob/f7c7230a27fb23932606f9847bfcf04c83b8ff06/src/Admin/SiteHealth.php#L333-L661) to the performance plugin.

## Checklist


ffPjrZUGXfcxuAj:
please add
`'x-aruba-cache' => $cache_hit_callback,`
for new aruba hispeed cache service

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * Callback for perflab_page_cache test.
+ *
+ * @since 1.0.0
+ *
+ * @return array
+ */
+function perflab_afpc_page_cache_test() {
+	$description  = '<p>' . esc_html__( 'WordPress performs at its best when page caching is enabled. This is because the additional optimizations performed require additional server processing time, and page caching ensures that responses are served quickly.', 'performance-lab' ) . '</p>';
```
mxbclang: @manuelRod This is a bit hard for me to parse. What do you think of something like:

"Page caching enhances the speed and performance of your site by saving and serving static pages instead of calling for a page every time a user visits."

cc @felixarntz

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,198 @@
...
+ * @since 1.0.0
+ */
+function perflab_afpc_page_cache_detected_test( $default_response, $page_cache_detail ) {
+
+	$default_response['status'] = $page_cache_detail['status'];
+
+	switch ( $page_cache_detail['status'] ) {
+		case 'recommended':
+			$default_response['badge']['color'] = 'orange';
+			$default_response['label']          = esc_html__( 'Page caching is not detected but the server response time is OK', 'performance-lab' );
```
mxbclang: Should this be "Page caching _was_ not detected," since the check has been run?

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,198 @@
...
+			break;
+		case 'good':
+			$default_response['badge']['color'] = 'green';
+			$default_response['label']          = esc_html__( 'Page caching is detected and the server response time is good', 'performance-lab' );
```
mxbclang: Same as above – should these all change to "was"?

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,198 @@
...
+
+	if ( empty( $page_cache_detail['response_time'] ) ) {
+		$page_cache_test_summary[] = '<span class="dashicons dashicons-dismiss"></span> ' . esc_html__( 'Server response time could not be determined. Verify that loopback requests are working.', 'performance-lab' );
+	} else {
+
+		$threshold = perflab_afpc_get_good_response_time_threshold();
+		if ( $page_cache_detail['response_time'] < $threshold ) {
+			$page_cache_test_summary[] = '<span class="dashicons dashicons-yes-alt"></span> ' . sprintf(
+				/* translators: %d is the response time in milliseconds */
+				esc_html__( 'Median server response time was %1$s milliseconds. This is less than the %2$s millisecond threshold.', 'performance-lab' ),
```
mxbclang: I'm guessing the threshold here is our own defined threshold – does it make sense to change this to something like "the _recommended_ [#] millisecond threshold"?

mxbclang:
@manuelRod Since I'm working on updating the language on the other Site Health reports, I took a look at the copy for this one as well and left some thoughts. Let me know if you have any questions!

bethanylang:
@manuelRod Since I'm working on updating the language on the other Site Health reports, I took a look at the copy for this one as well and left some thoughts. Let me know if you have any questions!

manuelRod:
Thanks for that @bethanylang, where I can find those thoughts?

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,198 @@
...
+ * Adds tests to site health.
+ *
+ * @since 1.0.0
+ *
+ * @param array $tests Site Health Tests.
+ * @return array
+ */
+function perflab_afpc_add_full_page_cache_test( $tests ) {
+	$tests['async']['perflab_page_cache'] = array(
+		'label'             => esc_html__( 'Page caching', 'performance-lab' ),
```
westonruter: I believe `label` is exported to JSON and is then injected with JS into the DOM in a way that does not deal properly with HTML tags and HTML entities. Therefore, I don't believe any escaping should be included here.

```suggestion
		'label'             => __( 'Page caching', 'performance-lab' ),
```

westonruter: Or even I don't know where this label is displayed, as there is another `badge.label` below which is probably what I have in mind here.

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,198 @@
...
+ * @return array
+ */
+function perflab_afpc_page_cache_test() {
+	$description  = '<p>' . esc_html__( 'WordPress performs at its best when page caching is enabled. This is because the additional optimizations performed require additional server processing time, and page caching ensures that responses are served quickly.', 'performance-lab' ) . '</p>';
+	$description .= '<p>' . esc_html__( 'Page caching is detected by looking for an active page caching plugin as well as making three requests to the homepage and looking for one or more of the following HTTP client caching response headers:', 'performance-lab' )
+					. ' <code>' . implode( '</code>, <code>', array_keys( perflab_afpc_get_page_cache_headers() ) ) . '.</code>';
+
+	$result = array(
+		'badge'       => array(
+			'label' => esc_html__( 'Performance', 'performance-lab' ),
```
westonruter: See note above.

```suggestion
			'label' => __( 'Performance', 'performance-lab' ),
```

modules/site-health/audit-full-page-cache/load.php
westonruter:
```suggestion
			$default_response['label']          = __( 'Page caching is not detected but the server response time is OK', 'performance-lab' );
```

modules/site-health/audit-full-page-cache/load.php
westonruter:
```suggestion
			$default_response['label']          = __( 'Page caching is detected and the server response time is good', 'performance-lab' );
```

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,198 @@
...
+			$default_response['badge']['color'] = 'orange';
+			$default_response['label']          = esc_html__( 'Page caching is not detected but the server response time is OK', 'performance-lab' );
+			break;
+		case 'good':
+			$default_response['badge']['color'] = 'green';
+			$default_response['label']          = esc_html__( 'Page caching is detected and the server response time is good', 'performance-lab' );
+			break;
+		default:
+			$default_response['badge']['color'] = 'red';
+			$default_response['label']          = esc_html__( 'Page caching is detected but the server response time is still slow', 'performance-lab' );
```
westonruter:
```suggestion
			$default_response['label']          = __( 'Page caching is detected but the server response time is still slow', 'performance-lab' );
```

westonruter: Should this line be moved to an `else` statement for the following `if` statement?

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,198 @@
...
+			if ( empty( $page_cache_detail['headers'] ) && ! $page_cache_detail['advanced_cache_present'] ) {
+				$default_response['label'] = esc_html__( 'Page caching is not detected and the server response time is slow', 'performance-lab' );
```
westonruter:
```suggestion
				$default_response['label'] = __( 'Page caching is not detected and the server response time is slow', 'performance-lab' );
```

westonruter (approved):
Overall looks good to me, but it looked good to me in the AMP plugin when I merged it, so I'm probably not the best to review for merging here.

modules/site-health/audit-full-page-cache/load.php
manuelRod: i don't quite understand what you mean here

modules/site-health/audit-full-page-cache/load.php
manuelRod: you are right, I double-checked how they do in the core.

manuelRod:
> please add `'x-aruba-cache' => $cache_hit_callback,` for new aruba hispeed cache service

added

modules/site-health/audit-full-page-cache/load.php
westonruter: I mean, instead of:

```php
			$default_response['label']          = esc_html__( 'Page caching is detected but the server response time is still slow', 'performance-lab' );
			if ( empty( $page_cache_detail['headers'] ) && ! $page_cache_detail['advanced_cache_present'] ) {
				$default_response['label'] = esc_html__( 'Page caching is not detected and the server response time is slow', 'performance-lab' );
			}
```

It can be:

```php
			if ( empty( $page_cache_detail['headers'] ) && ! $page_cache_detail['advanced_cache_present'] ) {
				$default_response['label'] = esc_html__( 'Page caching is not detected and the server response time is slow', 'performance-lab' );
			} else {
				$default_response['label'] = esc_html__( 'Page caching is detected but the server response time is still slow', 'performance-lab' );
			}
```

modules/site-health/audit-full-page-cache/helper.php
```diff
@@ -0,0 +1,206 @@
+<?php
+/**
+ * Helper functions used by module.
+ *
+ * @package performance-lab
+ * @since 1.0.0
+ */
+
+/**
+ * List of header ,and it's verification callback to verify if page cache is enabled or not.
```
felixarntz:
```suggestion
 * Returns a list of header and it's verification callback to verify if page cache is enabled or not.
```

modules/site-health/audit-full-page-cache/helper.php
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * Note: key is header name and value could be callable function to verify header value.
+ * Empty value mean existence of header detect page cache is enable.
```
felixarntz:
```suggestion
 * Empty value mean existence of header detect page cache is enabled.
```

modules/site-health/audit-full-page-cache/helper.php
```diff
@@ -0,0 +1,206 @@
...
+		'x-srcache-store-status' => $cache_hit_callback,
+		'x-srcache-fetch-status' => $cache_hit_callback,
+		'cf-edge-cache'          => static function ( $header_value ) {
+			return false !== strpos( strtolower( $header_value ), 'cache' );
+		},
+	);
+}
+
+/**
+ * Check if site has page cache enable or not.
```
felixarntz:
```suggestion
 * Checks if site has page cache enabled or not.
```

modules/site-health/audit-full-page-cache/helper.php
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * @since 1.0.0
+ * @return WP_Error|array {
```
felixarntz:
```suggestion
 * @since n.e.x.t
 *
 * @return WP_Error|array {
```

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,198 @@
...
+ * Callback for perflab_page_cache test.
+ *
+ * @since 1.0.0
+ *
+ * @return array
+ */
+function perflab_afpc_page_cache_test() {
+	$description  = '<p>' . __( 'WordPress performs at its best when page caching is enabled. This is because the additional optimizations performed require additional server processing time, and page caching ensures that responses are served quickly.', 'performance-lab' ) . '</p>';
+	$description .= '<p>' . __( 'Page caching is detected by looking for an active page caching plugin as well as making three requests to the homepage and looking for one or more of the following HTTP client caching response headers:', 'performance-lab' )
+					. ' <code>' . implode( '</code>, <code>', array_keys( perflab_afpc_get_page_cache_headers() ) ) . '.</code>';
```
felixarntz: You're missing the closing `</p>` tag here I think. Also, there probably doesn't need to be a period after the list of cache headers as that is not a sentence.

Last but not least, I think for code readability it would be better to break the second line here out into a separate assignment like `$description .= ' <code>' ...`.

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,198 @@
...
+			__( 'Learn more about page caching', 'performance-lab' ),
+			/* translators: The accessibility text. */
+			__( '(opens in a new tab)', 'performance-lab' )
+		),
+	);
+
+	$page_cache_detail = perflab_afpc_get_page_cache_detail();
+
+	if ( is_wp_error( $page_cache_detail ) ) {
+		return perflab_afpc_page_cache_unable_detect_cache_test( $result, $page_cache_detail );
```
felixarntz: Can you remove this function and put its code right here instead? Breaking this bit out into a new function is unnecessarily complicated since it touches the same `$result` variable, so having that within the same function where its initial values are defined makes it easier to follow.

Also this function isn't used anywhere else, so it shouldn't be necessary.

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	return perflab_afpc_page_cache_detected_test( $result, $page_cache_detail );
```
felixarntz: Same here, can you please remove this function and put the code right here instead? For the same reason I mentioned it above.

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,198 @@
...
+										sprintf(
+										/* translators: Placeholder is number of caching headers */
+											_n(
+												'There was %d client caching response header detected:',
+												'There were %d client caching response headers detected:',
+												count( $page_cache_detail['headers'] ),
+												'performance-lab'
+											),
+											count( $page_cache_detail['headers'] )
+										) . ' <code>' . implode( '</code>, <code>', $page_cache_detail['headers'] ) . '</code>.';
```
felixarntz: This code looks quite weird. Can we break it down into two assignments to a temporary string variable, and then afterwards assign that string to `$page_cache_test_summary[]`?

tests/modules/site-health/audit-full-page-cache/audit-full-page-cache-test.php
```diff
@@ -0,0 +1,367 @@
...
+		);
+	}
+
+	/**
+	 * Tests all possible scenarios given by dataProvider.
+	 *
+	 * @dataProvider provider_perflab_afpc_page_cache_test
+	 * @covers ::perflab_afpc_page_cache_test()
+	 * @covers ::perflab_afpc_get_page_cache_detail()
+	 * @covers ::perflab_afpc_page_cache_detected_test()
```
felixarntz: See above, can you remove this line based on my other comment to remove this function?

modules/site-health/audit-full-page-cache/helper.php
```diff
@@ -0,0 +1,206 @@
...
+		'page_caching_response_headers' => $page_caching_response_headers,
+		'response_timing'               => $response_timing,
+	);
+}
+
+/**
+ * Get page cache details.
+ *
+ * @since 1.0.0
+ * @return WP_Error|array {
```
felixarntz:
```suggestion
 * @since 1.0.0
 *
 * @return WP_Error|array {
```

tests/modules/site-health/audit-full-page-cache/audit-full-page-cache-test.php
```diff
@@ -0,0 +1,367 @@
+<?php
+/**
+ * Tests for audit-full-page-cache module.
+ *
+ * @package performance-lab
```
felixarntz, line 5: Can you add a group here with the module slug?
```suggestion
 * @package performance-lab
 * @group audit-full-page-cache
```

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,198 @@
+<?php
+/**
+ * Module Name: Audit Full Page Cache
+ * Description: Adds a Full page Cache checker in Site Health checks.
```
felixarntz: Let's align this more with the other Site Health module descriptions.
```suggestion
 * Description: Adds a check for full page cache in Site Health status.
```

felixarntz (commented):
@manuelRod Overall the code here looks solid. I left a few minor things for some cleanup, particularly I'd suggest to remove two of the functions that are only used in one place and actually make the code more complex to follow rather than less because of splitting how the `$result` is generated.

felixarntz:
@bethanylang Would be great to get your review here specifically on the copy in the Site Health check.

mxbclang (changes requested):

modules/site-health/audit-full-page-cache/helper.php
```diff
@@ -0,0 +1,205 @@
+<?php
+/**
+ * Helper functions used by module.
+ *
+ * @package performance-lab
+ * @since 1.0.0
+ */
+
+/**
+ * List of header ,and it's verification callback to verify if page cache is enabled or not.
```
mxbclang:
```suggestion
 * Returns a list of headers and its verification callback to verify if page cache is enabled or not.
```

felixarntz:
@manuelRod Moving this to the following milestone since it still needs a bit of iteration. We can probably get it into the next release though.

modules/site-health/audit-full-page-cache/load.php
manuelRod: understood :)

modules/site-health/audit-full-page-cache/load.php
manuelRod: @bethanylang the check is run in the bg once you enter into the Site Health screen, so checks are "live" and inmediate.

modules/site-health/audit-full-page-cache/load.php
manuelRod: right!

modules/site-health/audit-full-page-cache/load.php
manuelRod: makes sense

modules/site-health/audit-full-page-cache/load.php
manuelRod: makes sense

modules/site-health/audit-full-page-cache/load.php
manuelRod: makes sense

modules/site-health/audit-full-page-cache/load.php
manuelRod: @bethanylang makes sense, updated.

modules/site-health/audit-full-page-cache/load.php
manuelRod: your guess is right, I've updated.

modules/site-health/audit-full-page-cache/load.php
manuelRod: I think both ways can work.

mxbclang: @manuelRod Got it, fine to leave as is then.

modules/site-health/audit-full-page-cache/helper.php
```diff
@@ -0,0 +1,207 @@
+<?php
+/**
+ * Helper functions used by module.
+ *
+ * @package performance-lab
+ * @since 1.0.0
```
felixarntz:
```suggestion
 * @since n.e.x.t
```

modules/site-health/audit-full-page-cache/helper.php
```diff
@@ -0,0 +1,207 @@
...
+ */
+
+/**
+ * Returns a list of headers and its verification callback to verify if page cache is enabled or not.
+ *
+ * Note: key is header name and value could be callable function to verify header value.
+ * Empty value mean existence of header detect page cache is enabled.
+ *
+ * @return array List of client caching headers and their (optional) verification callbacks.
+ * @since 1.0.0
```
felixarntz: `@since` should come in its own block, _before_ `@param` and `@return`. Also, we need to use `n.e.x.t`.
```suggestion
 * @since n.e.x.t
 *
 * @return array List of client caching headers and their (optional) verification callbacks.
```

modules/site-health/audit-full-page-cache/helper.php
felixarntz: It's a nit-pick, but please address it :)

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,174 @@
...
+		$result['status']         = 'critical';
+		$error_info               = sprintf(
+		/* translators: 1 is error message, 2 is error code */
+			__( 'Unable to detect page caching due to possible loopback request problem. Please verify that the loopback request test is passing. Error: %1$s (Code: %2$s)', 'performance-lab' ),
+			$page_cache_detail->get_error_message(),
+			$page_cache_detail->get_error_code()
+		);
+		$result['description'] = wp_kses_post( "<p>$error_info</p>" ) . $result['description'];
+		return $result;
+	}
```
felixarntz: Should we even show the test when the loopback request is failing? It's a bit odd to not pass this test just because some other technical requirement for our own ability to conduct the test is missing.

My main point is: Having a page cache is crucial for performance. Loopback requests not working may not be that much of a problem depending on what you do with a site, at least it's not a performance problem. If loopback doesn't work, that means _we_ could not conduct this test - it doesn't mean that the site doesn't have page caching. So I think we should either not show the test in that situation at all or at least not mark it as critical and instead clarify that it is "N/A" because of the lack of loopback request support it couldn't be conducted.

tests/modules/site-health/audit-full-page-cache/audit-full-page-cache-test.php
```diff
@@ -0,0 +1,367 @@
...
+		}
+		$this->assertTrue( call_user_func( $route['permission_callback'] ) );
+	}
+
+	/**
+	 * Gets response data for test_perflab_afpc_page_cache_test().
+	 *
+	 * @return array[]
+	 */
+	public function provider_perflab_afpc_page_cache_test() {
```
felixarntz, line 138: A bit of a nit-pick, but could you move this provider method right below the test that it belongs to? That's how core structures it, and it helps for a better code overview when test and provider are colocated.

modules/site-health/audit-full-page-cache/helper.php
```diff
@@ -0,0 +1,207 @@
...
+		),
+		'page_caching_response_headers' => $page_caching_response_headers,
+		'response_timing'               => $response_timing,
+	);
+}
+
+/**
+ * Get page cache details.
+ *
+ * @since 1.0.0
```
felixarntz:
```suggestion
 * @since n.e.x.t
```

felixarntz (changes requested):
@manuelRod This looks solid and should be close to merge, I mostly have some nit-picks around documentation. Please make sure to update all version references for new code in this PR to `n.e.x.t`, per our guidelines.

Last but not least, based on #310, please refresh this PR against latest `trunk` and add the corresponding entries for this new module to the `CODEOWNERS` file, with at least yourself as the module owner.

modules/site-health/audit-full-page-cache/load.php
manuelRod: @felixarntz this was borrowed from the original code @westonruter can maybe elaborate further.

As far as I understand, if this error is triggered means the site is not accessible for whatever reason, since the loopback is just a wp_remote_get to the site.
This situation may not be realistic/reachable, or if that happens, the problem indeed may be somewhere else.

westonruter: Since there is another Site Health test for loopback requests, it would be redundant to also show an error in the page caching test for a loopback request failure. In that case, I think it would make sense for this to be downgraded from `critical` to either `recommended` with advice to look at the loopback request test to fix that issue since it is blocking this one, which the text of this is already doing.

manuelRod: Thanks, @westonruter for your input, I agree. I've downgraded the warning to "recommended".

.github/CODEOWNERS
```diff
@@ -54,3 +54,7 @@
 /modules/object-cache/persistent-object-cache-health-check                  @tillkruss @spacedmonkey
 /tests/modules/object-cache/persistent-object-cache-health-check            @tillkruss @spacedmonkey
 /tests/testdata/modules/object-cache/persistent-object-cache-health-check   @tillkruss @spacedmonkey
+
+# Module: Audit Enqueued Assets
+/modules/site-health/audit-full-page-cache                                  @manuelRod
```
adamsilverstein: should we add @westonruter as a code owner?

felixarntz:
@manuelRod @westonruter Given that this is an entirely new module and the `1.1.0` release is about to be finalized with the release coming next week, I think with the remaining iterations to make this won't give us enough time for testing prior to the release, so I think this would be better to include in the following `1.2.0` release. Let's make sure we finalize the PR soon so that it can be merged into `trunk` shortly after `1.1.0`, to have time for testing and potential further iterations.

.github/CODEOWNERS
manuelRod: let me know @westonruter

westonruter: Sure. Go for it.

manuelRod: done

akshitsethi:
I tested the module and so far it looks good to me. Multiple scenarios (looback request failed, without cache, with cache plugin) were tested and I've attached the screenshot for each below:

### Loopback failure
Screenshot 2022-05-25 at 4 43 40 PM

### Without cache
Screenshot 2022-05-25 at 5 00 36 PM

### Cache enabled
Screenshot 2022-05-25 at 5 05 32 PM

I'll test some more scenarios and update it over here.

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,174 @@
...
+
+	$page_cache_detail = perflab_afpc_get_page_cache_detail();
+
+	if ( is_wp_error( $page_cache_detail ) ) {
+		$result['badge']['color'] = 'orange';
+		$result['label']          = __( 'Unable to detect the presence of page caching', 'performance-lab' );
+		$result['status']         = 'recommended';
+		$error_info               = sprintf(
+		/* translators: 1 is error message, 2 is error code */
+			__( 'Unable to detect page caching due to possible loopback request problem. Please verify that the loopback request test is passing. Error: %1$s (Code: %2$s)', 'performance-lab' ),
```
felixarntz: Indentation is off here:
```suggestion
			/* translators: 1 is error message, 2 is error code */
			__( 'Unable to detect page caching due to possible loopback request problem. Please verify that the loopback request test is passing. Error: %1$s (Code: %2$s)', 'performance-lab' ),
```

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,174 @@
...
+ */
+require_once __DIR__ . '/helper.php';
+
+/**
+ * Adds tests to site health.
+ *
+ * @since n.e.x.t
+ *
+ * @param array $tests Site Health Tests.
+ * @return array
```
felixarntz: Please always provide a description for any `@param` or `@return`.
```suggestion
 * @return array Modified Site Health tests including Page caching test.
```

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,174 @@
...
+	return $tests;
+}
+add_filter( 'site_status_tests', 'perflab_afpc_add_full_page_cache_test' );
+
+/**
+ * Callback for perflab_page_cache test.
+ *
+ * @since n.e.x.t
+ *
+ * @return array
```
felixarntz: See above.
```suggestion
 * @return array The test results.
```

modules/site-health/audit-full-page-cache/load.php
```diff
@@ -0,0 +1,174 @@
...
+			);
+		}
+
+		if ( empty( $page_cache_detail['headers'] ) ) {
+			$page_cache_test_summary[] = '<span class="dashicons dashicons-warning"></span> ' . __( 'No client caching response headers were detected.', 'performance-lab' );
+		} else {
+			$headers_summary  = '<span class="dashicons dashicons-yes-alt"></span>';
+			$headers_summary .= sprintf(
+			/* translators: Placeholder is number of caching headers */
+				_n(
```
felixarntz: See above:
```suggestion
				/* translators: Placeholder is number of caching headers */
				_n(
```

felixarntz (approved):
@manuelRod Looks good to me, great stuff!

I left a few minor points on documentation and indentation, please address these. None of these are too critical though, so I'm already going to mark this as approved from my end.

felixarntz (approved):
Great, thanks @manuelRod!

@bethanylang You previously requested changes regarding the copy. Can you please give this another pass and approve once it's good from your end?

mxbclang (approved):
Language LGTM!

mxbclang:
@felixarntz @manuelRod Apologies if I was a blocker here; just approved on my side. Looks like this is good to go for 1.2.0! 🥳

felixarntz:
Thanks @bethanylang!

manuelRod:
awesome, thanks guys!

PR #269, opened 2022-03-28
Update Site Health reports copy

### LINKED ISSUE #229: Standardize structure for Site Health reports
## Feature Description

Currently, there's no clear definition in how some things are helpful for performance and why others are being measured, setting a standard by answering a set of predetermined questions would help in making sure the metrics are meaningful for the majority of users and not just technical users.

- **Why this measure is being tracked?** Answer this question in a clear way would help to understand the need to pass this test for example:

The metric above does not indicate why I need to keep WordPress up to date.

- **Why I need to pass that metric?** or **How the metric would help improve my performance?** Example:

What's the benefit of passing this test or keeping up to date with WordPress as a site owner what are the risk and benefits of going down this route? Is there any tradeoff If I decided not to? 

Avoid technical wording if possible

This example is missing some answers for the questions above, but also includes additional wording that might not be explicit for the vast majority of users.

mxbclang:
## Summary

Fixes #229.

## Checklist

### DIFF
--- a/modules/site-health/audit-autoloaded-options/load.php
+++ b/modules/site-health/audit-autoloaded-options/load.php
@@ -38,7 +38,7 @@ function perflab_aao_autoloaded_options_test() {
 	$autoloaded_options_count = count( wp_load_alloptions() );
 
 	$result = array(
-		'label'       => esc_html__( 'Autoloaded options', 'performance-lab' ),
+		'label'       => esc_html__( 'Autoloaded options are acceptable', 'performance-lab' ),
 		'status'      => 'good',
 		'badge'       => array(
 			'label' => esc_html__( 'Performance', 'performance-lab' ),
@@ -46,7 +46,7 @@ function perflab_aao_autoloaded_options_test() {
 		),
 		'description' => sprintf(
 		/* translators: 1: Number of autoloaded options. 2.Autoloaded options size. */
-			'<p>' . esc_html__( 'The amount of %1$s autoloaded options (size: %2$s) in options table is acceptable.', 'performance-lab' ) . '</p>',
+			'<p>' . esc_html__( 'Autoloaded options are configuration settings for plugins and themes that are automatically loaded with every page load in WordPress. Having too many autoloaded options can slow down your site. Your site has %1$s autoloaded options (size: %2$s) in the options table, which is acceptable.', 'performance-lab' ) . '</p>',
 			$autoloaded_options_count,
 			size_format( $autoloaded_options_size )
 		),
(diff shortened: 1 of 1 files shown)


modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -38,15 +38,15 @@ function perflab_aao_autoloaded_options_test() {
...
+		'label'       => esc_html__( 'Autoloaded options are acceptable', 'performance-lab' ),
 		'status'      => 'good',
 		'badge'       => array(
 			'label' => esc_html__( 'Performance', 'performance-lab' ),
 			'color' => 'blue',
 		),
 		'description' => sprintf(
 		/* translators: 1: Number of autoloaded options. 2.Autoloaded options size. */
-			'<p>' . esc_html__( 'The amount of %1$s autoloaded options (size: %2$s) in options table is acceptable.', 'performance-lab' ) . '</p>',
+			'<p>' . esc_html__( 'Autoloaded options are configuration settings for plugins and themes that are automatically loaded with every page load in WordPress. Having too many autoloaded options can slow down your site. Your site has %1$s autoloaded options (size: %2$s) in the options table, which is acceptable.', 'performance-lab' ) . '</p>',
```
felixarntz, line 49: I think it would be good to split this string into two, one with the sentence(s) that is the same as the one for the other case (line 75), the other one with what is specific to this case.

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -69,9 +69,10 @@ function perflab_aao_autoloaded_options_test() {
 
 	$result['status']         = 'critical';
 	$result['badge']['color'] = 'red';
+	$result['label']          = esc_html__( 'Autoloaded options could affect performance', 'performance-lab' );
 	$result['description']    = sprintf(
 	/* translators: 1: Number of autoloaded options. 2.Autoloaded options size. */
-		'<p>' . esc_html__( 'Your website uses %1$s autoloaded options (size: %2$s). Try to reduce the number of autoloaded options or performance will be affected.', 'performance-lab' ) . '</p>',
+		'<p>' . esc_html__( 'Autoloaded options are configuration settings for plugins and themes that are automatically loaded with every page load in WordPress. Your site has %1$s autoloaded options (size: %2$s) in the options table, which could cause your site to be slow. You can reduce the number of autoloaded options by cleaning up your site\'s options table.', 'performance-lab' ) . '</p>',
```
felixarntz, line 75: See above, the first sentence should be broken out since it can be reused. This will make it easier for translators. Also, I wonder about the second sentence you have above (_Having too many autoloaded options can slow down your site._) - that seems it would make sense to have here as well? Since it explains why to even care about autoloaded options.

felixarntz (commented):
@bethanylang Left two comments, mostly this looks good so far.

Would you like to incorporate your suggestions on the WebP Support and Object Cache checks from https://github.com/WordPress/performance/issues/229#issuecomment-1076490842 into this PR as well? I'd say that would work, and this way it could all be reviewed together. I agree the Enqueued Assets bit should remain separate as it needs more higher-level discussion.

mxbclang:
Requesting to close as I've updated the branch name and submitted a new PR: https://github.com/WordPress/performance/pull/272. Sorry for the trouble!

PR #271, opened 2022-03-28
Add partytown support to run scripts in Worker Thread

### LINKED ISSUE #176: Partytown support
# Description

## What is Partytown?

## How does Partytown work?

## Partytown tradeoffs

## Partytown Status

# Goals

  

# Proposals

```php
wp_enqueue_script(
  'non-critical-script',
  array(),
  '1.0.0',
  'worker-partytown'
);
```

Web Worker option in plugins

```php
  $config["debug"] = true;
  return $config;
}
```

```html
<script>
  partytown = {
    debug: true,
  };
</script>
```

## 2. Partytown module/plugin

```html
    /* ... */
  })(
    window,
    document,
    "script",
    "ga"
  );
  ga("create", "", "auto");
  ga("send", "pageview");
</script>
```

## Service Worker integration

thelovekesh:
## Summary

Fixes #176

## Relevant technical choices
- Add Web Worker module which allows running scripts within worker thread with the help of Partytown library.
- Add the Partytown package in dependencies
- Add npm script to copy Partytown lib files to module `assets/js` folder.
    - Currently, Partytown lib files are committed to the module.
    - We can also set up webpack to copy partytown js files in the module build folder when creating the release of the plugin.
    - Partytown library needs to be updated when a new version is released.
- Add tests for the Web Worker Module.
- Add `partytown_configuration` filter which can be used to update partytown [configurations](https://partytown.builder.io/configuration).

## Screencast
https://user-images.githubusercontent.com/54371619/159474078-d88fa7f7-293f-4347-90f4-4ab0d4881240.mp4

## Checklist

### DIFF
--- a/modules/images/webp-uploads/helper.php
+++ b/modules/images/webp-uploads/helper.php
@@ -20,6 +20,14 @@ function webp_uploads_get_upload_image_mime_transforms() {
 		'image/webp' => array( 'image/webp' ),
 	);
 
+	// Check setting for whether to generate both JPEG and WebP.
+	if ( true === (bool) get_option( 'perflab_generate_webp_and_jpeg' ) ) {
+		$default_transforms = array(
+			'image/jpeg' => array( 'image/jpeg', 'image/webp' ),
+			'image/webp' => array( 'image/webp', 'image/jpeg' ),
+		);
+	}
+
 	/**
 	 * Filter to allow the definition of a custom mime types, in which a defined mime type
 	 * can be transformed and provide a wide range of mime types.
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -14,6 +14,7 @@
 require_once __DIR__ . '/helper.php';
 require_once __DIR__ . '/rest-api.php';
 require_once __DIR__ . '/image-edit.php';
+require_once __DIR__ . '/settings.php';
 
 /**
  * Hook called by `wp_generate_attachment_metadata` to create the `sources` property for every image
--- a/modules/images/webp-uploads/settings.php
+++ b/modules/images/webp-uploads/settings.php
@@ -1,0 +1,77 @@
+<?php
+/**
+ * Settings for the WebP Uploads module.
+ *
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+/**
+ * Registers setting for generating both JPEG and WebP versions for image uploads.
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_register_media_settings_field() {
+	register_setting(
+		'media',
+		'perflab_generate_webp_and_jpeg',
+		array(
+			'type'         => 'boolean',
+			'default'      => false,
+			'show_in_rest' => false,
+		)
+	);
+}
+add_action( 'init', 'webp_uploads_register_media_settings_field' );
+
+/**
+ * Adds media settings field for the 'perflab_generate_webp_and_jpeg' setting.
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_add_media_settings_field() {
+	// Add settings field.
+	add_settings_field(
+		'perflab_generate_webp_and_jpeg',
+		__( 'Generate WebP and JPEG', 'performance-lab' ),
+		'webp_uploads_generate_webp_jpeg_setting_callback',
+		'media',
+		'uploads',
+		array( 'class' => 'perflab-generate-webp-and-jpeg' )
+	);
+}
+add_action( 'admin_init', 'webp_uploads_add_media_settings_field' );
+
+/**
+ * Renders the settings field for the 'perflab_generate_webp_and_jpeg' setting.
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_generate_webp_jpeg_setting_callback() {
+	?>
+	</td>
+	<td class="td-full">
+		<label for="perflab_generate_webp_and_jpeg">
+			<input name="perflab_generate_webp_and_jpeg" type="checkbox" id="perflab_generate_webp_and_jpeg" aria-describedby="perflab_generate_webp_and_jpeg_description" value="1"<?php checked( '1', get_option( 'perflab_generate_webp_and_jpeg' ) ); ?> />
+			<?php esc_html_e( 'Generate JPEG files in addition to WebP', 'performance-lab' ); ?>
+		</label>
+		<p class="description" id="perflab_generate_webp_and_jpeg_description"><?php esc_html_e( 'Enabling JPEG in addition to WebP can improve compatibility, but will effectively double the filesystem storage use of your images.', 'performance-lab' ); ?></p>
+	<?php
+}
+
+/**
+ * Adds custom style for media settings.
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_media_setting_style() {
+	?>
+	<style>
+	.form-table .perflab-generate-webp-and-jpeg th,
+	.form-table .perflab-generate-webp-and-jpeg td:not(.td-full) {
+		display: none;
+	}
+	</style>
+	<?php
+}
+add_action( 'admin_head-options-media.php', 'webp_uploads_media_setting_style' );
--- a/tests/modules/images/webp-uploads/helper-tests.php
+++ b/tests/modules/images/webp-uploads/helper-tests.php
@@ -421,6 +421,28 @@ class WebP_Uploads_Helper_Tests extends ImagesTestCase {
 	}
 
 	/**
+	 * Returns JPG and WebP transforms array when perflab_generate_webp_and_jpeg option is true.
+	 *
+	 * @test
+	 */
+	public function it_should_return_jpeg_and_webp_transforms_when_option_generate_webp_and_jpeg_set() {
+		remove_all_filters( 'webp_uploads_get_upload_image_mime_transforms' );
+
+		update_option( 'perflab_generate_webp_and_jpeg', true );
+
+		$transforms = webp_uploads_get_upload_image_mime_transforms();
+
+		$this->assertIsArray( $transforms );
+		$this->assertSame(
+			array(
+				'image/jpeg' => array( 'image/jpeg', 'image/webp' ),
+				'image/webp' => array( 'image/webp', 'image/jpeg' ),
+			),
+			$transforms
+		);
+	}
+
+	/**
 	 * @dataProvider data_provider_image_filesize
 	 *
 	 * @test
--- a/tests/modules/images/webp-uploads/load-tests.php
+++ b/tests/modules/images/webp-uploads/load-tests.php
@@ -109,6 +109,37 @@ class WebP_Uploads_Load_Tests extends ImagesTestCase {
 	}
 
 	/**
+	 * Create JPEG and WebP for JPEG images, if perflab_generate_webp_and_jpeg option set.
+	 *
+	 * @test
+	 */
+	public function it_should_create_jpeg_and_webp_for_jpeg_images_if_generate_webp_and_jpeg_set() {
+		update_option( 'perflab_generate_webp_and_jpeg', true );
+
+		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg' );
+
+		// There should be JPEG and WebP sources for the full image.
+		$this->assertImageHasSource( $attachment_id, 'image/jpeg' );
+		$this->assertImageHasSource( $attachment_id, 'image/webp' );
+
+		$metadata = wp_get_attachment_metadata( $attachment_id );
+
+		// The full image should be a JPEG.
+		$this->assertArrayHasKey( 'file', $metadata );
+		$this->assertStringEndsWith( $metadata['sources']['image/jpeg']['file'], $metadata['file'] );
+		$this->assertStringEndsWith( $metadata['sources']['image/jpeg']['file'], get_attached_file( $attachment_id ) );
+
+		// The post MIME type should be JPEG.
+		$this->assertSame( 'image/jpeg', get_post_mime_type( $attachment_id ) );
+
+		// There should be JPEG and WebP sources for all sizes.
+		foreach ( array_keys( $metadata['sizes'] ) as $size_name ) {
+			$this->assertImageHasSizeSource( $attachment_id, $size_name, 'image/jpeg' );
+			$this->assertImageHasSizeSource( $attachment_id, $size_name, 'image/webp' );
+		}
+	}
+
+	/**
 	 * Don't create the sources property if no transform is provided.
 	 *
 	 * @test
--- a/uninstall.php
+++ b/uninstall.php
@@ -38,4 +38,5 @@ perflab_delete_plugin_option();
  */
 function perflab_delete_plugin_option() {
 	delete_option( 'perflab_modules_settings' );
+	delete_option( 'perflab_generate_webp_and_jpeg' );
 }


mxbclang:
@thelovekesh Can you please add [Focus] and [Type] labels, as well as a `no milestone` label? These are required for all PRs. Thanks!

thelovekesh:
@bethanylang I have updated the PR description but I don't have access to add the label in PR. Can you please do that for me?

modules/javascript/web-worker/load.php
```diff
@@ -0,0 +1,105 @@
...
+					$tag               = $create_script_tag;
+				}
+				return $tag;
+			},
+			10,
+			3
+		);
+	}
+}
+add_action( 'wp_print_scripts', 'web_worker_partytown_worker_scripts' );
```
westonruter: Instead of `wp_print_scripts`, this should perhaps do `wp_head`. Otherwise, It can run multiple times as found in https://github.com/ampproject/amp-wp/issues/6643.

westonruter: Or maybe this should be refactored to have a `script_loader_tag` filter added at the top level.

modules/javascript/web-worker/assets/js/partytown/debug/partytown.js
```diff
@@ -0,0 +1,72 @@
...
+                scripts = doc.querySelectorAll('script[type="text/partytown"]');
+                if (top != win) {
+                    top.dispatchEvent(new CustomEvent("pt1", {
+                        detail: win
+                    }));
+                } else {
+                    timeout = setTimeout(fallback, 1e4);
+                    doc.addEventListener("pt0", clearFallback);
+                    useAtomics ? loadSandbox(1) : nav.serviceWorker ? nav.serviceWorker.register(libPath + (config.swPath || "partytown-sw.js"), {
+                        scope: libPath
```
westonruter: This will be compatible with the PWA plugin as long as the `scope` is defined as the actual library path and not `/`.

thelovekesh:
@westonruter What if we consider this feature to be integrated into the PWA plugin instead of this one? By both means, this feature can be suggested to be merged in Core but from the PWA plugin, it will be more tested with _PWA use cases_ as well. What do you think?

westonruter:
I think it makes sense as part of the Performance plugin, especially since the scope doesn't conflict.

aristath:
The goal of the performance plugin is to serve as a feature plugin for things that will - eventually (hopefully) - land in WP Core.
I love the idea of partytown, but at the same time, I'm not sure others would be as enthusiastic to add it to WP-Core... It's a new project, still in beta, relatively unproven. WordPress on the other hand avoids adding 3rd-party scripts unless there's no way around it, and usually prefers more "established" solutions because of its popularity & wide reach.

If the chances of something like this landing in core are slim because of WP's established attitude towards implementations like this, then perhaps there are alternatives we can explore?

luisherranz:
I don't think that is something we need to decide now.

How Partytown will mature is hard to predict. Right now, it's just promising, but it is already recommended in [Google's optimizing third-party scripts](https://web.dev/script-component/#using-web-workers) guide.

Having it here as a module means it will be exposed to more people than with a standalone plugin, and it will help us understand the best way to integrate it with WordPress. And one could even argue that the other way around is also true: people looking for Partytown in WordPress will find and install this plugin, and the rest of the modules will get more exposure and testing.

Over time, we can decide if the best approach is to use an external plugin or a pre-registered dependency in core.

modules/javascript/web-worker/assets/js/partytown/partytown.js
```diff
@@ -0,0 +1,2 @@
...
...type="text/partytown"]'),i!=t?i.dispatchEvent(new CustomEvent("pt1",{detail:t})):(d=setTimeout(w,1e4),e.addEventListener("pt0",f),r?h(1):n.serviceWorker?n.serviceWorker.register(a+(o.swPath||"partytown-sw.js"),{scope:a}).then((function(t){t.active?h():t.installing&&t.installing.addEventListener("statechange",(function(t){"activated"==t.target.state&&h()}))}),console.error):w())))}function h(t){c=e.createElement(t?"script":"iframe"),t||(c.setAttribute("style","display:block;width:0;height:0;border:0;visibility:hidden"),c.setAttribute("aria-hidden",!0)),c.src=a+"partytown-"+(t?"atomics.js?v=0.5.1":"sandbox-sw.html?"+Date.now()),e.body.appendChild(c)}function w(t,n){for(f(),t=0;t<s.length;t++)(n=e.createElement("script")).innerHTML=s[t].innerHTML,e.head.appendChild(n);c&&c.parentNode.removeChild(c)}function f(){clearTimeout(d)}o=t.partytown||{},i==t&&(o.forward||[]).map((function(e){p=t,e.split(".").map((function(e,n,i){p=p[i[n]]=n+1<i.length?"push"==i[n+1]?[]:p[i[n]]||{}:function(){(t._ptf=t._ptf||[]).push(i,arguments)}}))})),"complete"==e.readyState?u():(t.addEventListener("DOMContentLoaded",u),t.addEventListener("load",u))}(window,document,navigator,top,window.crossOriginIsolated);
```
mitogh: Is there any particular reason not to include the partytown dependency via NPM instead?this dependency via

modules/javascript/web-worker/load.php
```diff
@@ -0,0 +1,105 @@
...
+add_action( 'wp_head', 'web_worker_partytown_configuration', 1 );
+
+/**
+ * Initialize PartyTown
+ *
+ * @since 1.0.0
+ * @return void
+ */
+function web_worker_partytown_init() {
+	wp_enqueue_script(
```
mitogh: I think we should only register the script at this stage and only enqueue in the scenarios where is required instead of loading it all the time. Additionally is this required to be loaded at the top of the page instead of in the `<footer>` ?
```suggestion
	wp_register_script(
```

modules/javascript/web-worker/load.php
```diff
@@ -0,0 +1,105 @@
...
+	 */
+	$config = apply_filters( 'partytown_configuration', $config );
+
+	?>
+	<script>
+		window.partytown = <?php echo wp_json_encode( $config ); ?>;
+	</script>
+	<?php
+}
+add_action( 'wp_head', 'web_worker_partytown_configuration', 1 );
```
mitogh: Any particular reason why this was not localized instead with `wp_localize_script` associated with the script being enqueued instead of creating the variable all the time.

package.json
```diff
@@ -37,11 +37,15 @@
     "pretest-php-multisite": "wp-env run composer 'install --no-interaction'",
     "test-php-multisite": "wp-env run phpunit 'WP_MULTISITE=1 phpunit -c /var/www/html/wp-content/plugins/performance/tests/multisite.xml --verbose'",
     "wp-env": "wp-env",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "partytown": "partytown copylib './modules/javascript/web-worker/assets/js/partytown'"
```
mitogh: Curious why we need this moved into the repo and as a dependency?

tests/modules/javascript/web-worker/web-worker-test.php
```diff
@@ -0,0 +1,158 @@
...
+class Web_Worker_Test extends WP_UnitTestCase {
+
+	/**
+	 * @covers ::web_worker_partytown_configuration
+	 */
+	function test_web_worker_partytown_configuration() {
+		$this->assertEquals(
+			1,
+			has_action( 'wp_head', 'web_worker_partytown_configuration' )
+		);
```
mitogh:
```suggestion
		$this->assertNotFalse( has_action( 'wp_head', 'web_worker_partytown_configuration' ) );
```

tests/modules/javascript/web-worker/web-worker-test.php
```diff
@@ -0,0 +1,158 @@
...
+		foreach ( $desired_output_string_chunks as $chunk ) {
+			$this->assertStringContainsString( $chunk, $config_output );
+		}
+	}
+
+	/**
+	 * @covers ::web_worker_partytown_init
+	 */
+	function test_web_worker_partytown_init() {
+		$this->assertEquals(
```
mitogh: Same as above.

tests/modules/javascript/web-worker/web-worker-test.php
```diff
@@ -0,0 +1,158 @@
...
+		$this->assertTrue( wp_script_is( 'partytown', 'enqueued' ) );
+	}
+
+	/**
+	 * @covers ::web_worker_partytown_worker_scripts
+	 */
+	function test_web_worker_partytown_worker_scripts() {
+		global $wp_scripts;
+
+		$this->assertEquals(
```
mitogh: Same as above.

modules/javascript/web-worker/assets/js/partytown/partytown.js
luisherranz: Ideally, the script should be inlined.

> The partytown.js file must be in the `<head>` before all other third-party scripts, and should not have the async or defer attributes. Because the script must be executed immediately, and since it’s such a small file, it’s always preferred to inline the partytown.js script content, rather than it being an additional http request.

https://partytown.builder.io/distribution

thelovekesh: > Ideally, the script should be inlined.

Yes, it should be. I will send a commit for it 👍🏼

modules/javascript/web-worker/load.php
thelovekesh: > I think we should only register the script at this stage and only enqueue in the scenarios where is required instead of loading it all the time.

Partytown script should be enqueued before any other script in order to queue third-party script in a Worker thread so IMO it should be enqueued instead. Also, a check can be added to check if some script's dependency is partytown then only enqueue the Partytwon script. Make sense 👍🏼 

> Additionally is this required to be loaded at the top of the page instead of in the `<footer>` ?

It should be loaded in the head as most of the time third-party scripts like Google Analytics are enqueued in the head.

modules/javascript/web-worker/load.php
thelovekesh: Since partytown requires `partytown = {...}` as a window object, will it be the best case to localize it?

package.json
thelovekesh: Partytown `lib` files are required to be used in the browser so we need to include these in the module. The suggested way is to use `partytown copylib` to copy `lib` files in a directory from where you want to enqueue them.

Other possible ways:
- Require them from `node_modules`, but we can not ship `node_modules` with the plugin in production.
- Another possible way is to use webpack or some other budler to copy its content from `node_modules` to module assets.

modules/javascript/web-worker/load.php
luisherranz: EDIT: @thelovekesh already answered the same (I didn't get his answer, maybe a problem with my connection).

package.json
luisherranz: EDIT: @thelovekesh already answered the same (I didn't get his answer, maybe a problem with my connection).

modules/javascript/web-worker/load.php
thelovekesh: @luisherranz Any inputs on this one?

package.json
luisherranz: I think the current approach of using `partytown copylib` is fine.

modules/javascript/web-worker/load.php
luisherranz: To be honest, I have no idea. @mitogh, what is your suggestion? Could you add a code example?

modules/javascript/web-worker/load.php
luisherranz: > Also, a check can be added to check if some script's dependency is partytown then only enqueue the Partytwon script

I have a second thought about this: what happens if people add `scripts` manually without using `wp_enqueue_script`? [Something like this example](https://github.com/rtCamp/wp-partytown/blob/develop/examples/post-request.php#L21-L56). Is it something the plugin should support?

thelovekesh: @luisherranz You mentioned the nice use case. In this case, I think we need to enqueue it each time.

module-i18n.php
```diff
@@ -3,6 +3,8 @@
 $generated_i18n_strings = array(
 	_x( 'WebP Uploads', 'module name', 'performance-lab' ),
 	_x( 'Creates WebP versions for new JPEG image uploads if supported by the server.', 'module description', 'performance-lab' ),
+	_x( 'Web Worker', 'module name', 'performance-lab' ),
```
luisherranz: I would be more explicit on the naming and actually call it "Partytown Web Worker" because maybe there are other web worker implementations added in the future, like [Worker DOM](https://github.com/ampproject/worker-dom).

modules/javascript/web-worker/load.php
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * PartyTown Configuration
+ *
+ * @since 1.0.0
+ * @see https://partytown.builder.io/configuration
+ * @return void
+ */
+function web_worker_partytown_configuration() {
+	$config = array(
+		'lib' => str_replace( site_url(), '', plugin_dir_url( __FILE__ ) ) . 'assets/js/partytown/',
```
luisherranz: Maybe we can add an empty [forward](https://partytown.builder.io/forwarding-events) array by default, so plugins can add items to the array without accidentally deleting the configuration of other plugins:

```php
// Plugin 1
function partytown_add_forward ( $config ) {
	$config["forward"] = array("dataLayer.push");
	return $config;
}

// Plugin 2
function partytown_add_forward_2 ( $config ) {
	$config["forward"] = array("fbq"); // <- UPS!
	return $config;
}
```

They just need to add items to the existing array:

```php
// Plugin 1
function partytown_add_forward ( $config ) {
	$config["forward"][] = "dataLayer.push";
	return $config;
}

// Plugin 2
function partytown_add_forward_2 ( $config ) {
	$config["forward"][] = "fbq"; // <- Fine
	return $config;
}
```

luisherranz (approved):
I've been testing this and everything seems to work great. Great work @thelovekesh 🙂

Maybe in the future, we can add a `debug` toggle in the settings, but it can easily be activated using the configuration filter, so not a problem.

```php
function partytown_debug_mode ( $config ) {
	$config["debug"] = true;
	return $config;
}
add_filter( 'partytown_configuration', 'partytown_debug_mode' );
```

modules/javascript/web-worker/load.php
mitogh: > It should be loaded in the head as most of the time third-party scripts like Google Analytics are enqueued in the head.

Cool, this inline would impact the TTFB by having this inline executed at the top of the page.

My point by using a `wp_register_script` is the fact that we can deregister from an external plugin in the case where we don't want to execute Partytown so the plugin effectively first registers the script and then enqueue happens, by doing this as well it allows for other scripts that depend on this script to be enqueued when dependencies from party town are enqueued. 

Additionally this mentions that this should work every time a page loads similarly to Google Analytics, maybe I'm not understanding the use case, would you mind expanding a bit so I can have a better sense in that regard.

modules/javascript/web-worker/load.php
mitogh: This is linked to the previous comment by using `wp_register_script` we allow for a more WP way to include this type of global.

```php
wp_localize_script( 'partytown', 'partytown', $config ); 
```

package.json
mitogh: Great makes sense wasn't sure about the approach, I would suggest a follow-up issue once this is merged as an infrastructure task that can be included as part of the deployment with lower priority but just to keep track as a way to improve this mechanism.

Thanks.

modules/javascript/web-worker/load.php
luisherranz: The main partytown script needs to be a blocking `<head>` script. Not inlining it means that the browser needs to do an additional blocking round-trip to the server, which would delay the page load much more than adding those extra kbs to the HTML.

luisherranz: > would you mind expanding a bit so I can have a better sense in that regard

There are [some articles in their docs](https://partytown.builder.io/#articles) that you may find interesting, especially [the first one](https://dev.to/adamdbradley/introducing-partytown-run-third-party-scripts-from-a-web-worker-2cnp). If you want to see it in action, this [video with Adam Bradley](https://www.youtube.com/watch?v=btr1CiHMvzU) is a good resource.

luisherranz:
@thelovekesh, any progress here? Are you planning on finishing this PR? 🙂

thelovekesh:
@luisherranz yes and it's on my todo. Based on the above discussion I don't have much work to do on this PR. Once it got some reviews and ideas I will make changes accordingly.

luisherranz:
Awesome. Thanks @thelovekesh!

modules/javascript/web-worker/assets/js/partytown/partytown.js
adamsilverstein: could these partytown  files be created as part of a build process instead of copied manually into the plugin? that would make keeping everything up to date more straightforward.

thelovekesh: @adamsilverstein I was thinking the same. I think we can include a grunt file with required tasks or I can add a workflow for it. What is better in your opinion?

modules/javascript/web-worker/load.php
adamsilverstein: inlining is fine, however:
* `partytown` should only be added if used on the page (eg. some other script depends on it)

> the fact that we can deregister from an external plugin

I don't think this is an important concern for the plugin implementation (the module can already be disabled programmatically) , especially since plugins have to opt in so far.

modules/javascript/web-worker/load.php
```diff
@@ -0,0 +1,105 @@
...
+		plugin_dir_url( __FILE__ ) . 'assets/js/partytown/partytown.js',
+		array(),
+		PERFLAB_VERSION,
+		false
+	);
+}
+add_action( 'wp_enqueue_scripts', 'web_worker_partytown_init', 1 );
+
+/**
+ * Get all scripts tags which has `partytown` dependency.
```
adamsilverstein:
```suggestion
 * Get all scripts tags which have a `partytown` dependency.
```

modules/javascript/web-worker/load.php
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * @since 1.0.0
+ * @return void
+ */
+function web_worker_partytown_worker_scripts() {
```
adamsilverstein: we have switched to prefixing all functions with "perflab_"

modules/javascript/web-worker/load.php
```diff
@@ -0,0 +1,105 @@
...
+	global $wp_scripts;
+
+	$partytown_handles = array();
+
+	// Get all scripts which has `partytown` dependency.
```
adamsilverstein:
```suggestion
	// Get all scripts which have a `partytown` dependency.
```

modules/javascript/web-worker/load.php
```diff
@@ -0,0 +1,105 @@
...
+	foreach ( $wp_scripts->registered as $handle => $script ) {
+		if ( ! empty( $script->deps ) && in_array( 'partytown', $script->deps ) ) { // phpcs:ignore WordPress.PHP.StrictInArray.MissingTrueStrict
```
adamsilverstein: why not use a strict comparison here?

modules/javascript/web-worker/load.php
```diff
@@ -0,0 +1,105 @@
...
+			 * @param string $handle Script handle.
+			 * @param string $src Script source.
+			 * @param string $partytown_handle Script handle which have `partytown` dependency.
+			 *
+			 * @return string $tag Script tag with type="text/partytown".
+			 */
+			function( $tag, $handle, $src ) use ( $partytown_handle ) {
+				if ( $handle === $partytown_handle ) {
+					$create_script_tag = sprintf(
+						'<script type="text/partytown" src="%1s" id="%2s"></script>',
```
adamsilverstein: why create a script tag from scratch? this could loose other attributes added by other filters. Instead can we replace (or set) just the `type`? Is changing the ID required/expected?

tests/modules/javascript/web-worker/web-worker-test.php
```diff
@@ -0,0 +1,155 @@
...
+		 * Unset $wp_scripts->done as it get saved into transients of enqueued_scripts.
+		 * @see perflab_aea_audit_enqueued_scripts()
+		 */
+		$wp_scripts->done = array();
+	}
+
+	/**
+	 * Helper function to get all scripts tags which has `partytown` dependency.
+	 */
+	function get_partytown_handles() {
```
adamsilverstein: this helper function can be used in the main file as well.

modules/javascript/web-worker/load.php
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Get all scripts tags which has `partytown` dependency.
+ *
+ * @since 1.0.0
+ * @return void
+ */
+function web_worker_partytown_worker_scripts() {
+	global $wp_scripts;
+
+	$partytown_handles = array();
```
adamsilverstein: replace this section calling the helper function from the test file

adamsilverstein (changes requested):
Overall this looks really good! I left some code feedback for things that could be improved.

Some questions:
* are there any scripts loaded by core or core themes that are candidates for `partytown` loading?   
* how would this get used, do we expect plugin authors to add an integration when the Performance Lab plugin is active, or do we want to try to build some integrations for top plugins where we know this could be helpful?
* how does this affect performance? it would be great to run some lab tests with some common candidates like analytics to measure the impact of this change.

thelovekesh:
Closed accidently due to commits dropped from the patch branch. I have opened a new one here - #556

modules/javascript/web-worker/load.php
thelovekesh: 🤔 can't recall why but it seems there are no blockers to use strict comparison. So updated it to use strict comparison.

modules/javascript/web-worker/load.php
thelovekesh: Updated.

module-i18n.php
thelovekesh: Updated 👍🏼

modules/javascript/web-worker/load.php
thelovekesh: Ahh. Actually, it's needed else the plugins might confuse adding forwarding events, Great suggestion. Thanks 💯

PR #272, opened 2022-03-29
Update Site Health reports copy for more clarity and consistency

### LINKED ISSUE #229: Standardize structure for Site Health reports

mxbclang:
## Summary

Fixes #229.

## Checklist

### DIFF
--- a/modules/site-health/audit-autoloaded-options/load.php
+++ b/modules/site-health/audit-autoloaded-options/load.php
@@ -38,7 +38,7 @@ function perflab_aao_autoloaded_options_test() {
 	$autoloaded_options_count = count( wp_load_alloptions() );
 
 	$result = array(
-		'label'       => esc_html__( 'Autoloaded options', 'performance-lab' ),
+		'label'       => esc_html__( 'Autoloaded options are acceptable', 'performance-lab' ),
 		'status'      => 'good',
 		'badge'       => array(
 			'label' => esc_html__( 'Performance', 'performance-lab' ),
@@ -46,7 +46,7 @@ function perflab_aao_autoloaded_options_test() {
 		),
 		'description' => sprintf(
 		/* translators: 1: Number of autoloaded options. 2.Autoloaded options size. */
-			'<p>' . esc_html__( 'The amount of %1$s autoloaded options (size: %2$s) in options table is acceptable.', 'performance-lab' ) . '</p>',
+			'<p>' . esc_html__( 'Autoloaded options are configuration settings for plugins and themes that are automatically loaded with every page load in WordPress. Having too many autoloaded options can slow down your site. Your site has %1$s autoloaded options (size: %2$s) in the options table, which is acceptable.', 'performance-lab' ) . '</p>',
 			$autoloaded_options_count,
 			size_format( $autoloaded_options_size )
 		),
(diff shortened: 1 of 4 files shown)


mxbclang:
@felixarntz This is ready for review with all of the copy changes suggested in #229, as well as some modifications to the autloaded options copy based on your feedback in the previous PR. Apologies for the mess... it's been a very long time since I've worked with Git and I'm getting back into the swing of things!

tillkruss (approved):

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -69,9 +69,10 @@ function perflab_aao_autoloaded_options_test() {
 
 	$result['status']         = 'critical';
 	$result['badge']['color'] = 'red';
+	$result['label']          = esc_html__( 'Autoloaded options could affect performance', 'performance-lab' );
 	$result['description']    = sprintf(
-	/* translators: 1: Number of autoloaded options. 2.Autoloaded options size. */
-		'<p>' . esc_html__( 'Your website uses %1$s autoloaded options (size: %2$s). Try to reduce the number of autoloaded options or performance will be affected.', 'performance-lab' ) . '</p>',
+	/* translators: 1. Number of autoloaded options. 2. Autoloaded options size. */
+		'<p>' . esc_html__( 'Autoloaded options are configuration settings for plugins and themes that are automatically loaded with every page load in WordPress. Having too many autoloaded options can slow down your site. Your site has %1$s autoloaded options (size: %2$s) in the options table, which could cause your site to be slow. You can reduce the number of autoloaded options by cleaning up your site\'s options table.', 'performance-lab' ) . '</p>',
```
felixarntz: See my similar comment on #269, it would be better to break out the first sentence here into its own translation string because it's the exact same sentence between line 49 and here. This makes it easier for translators as it avoids duplicate translations on the same sentence.

Let me know if you prefer any help with that.

felixarntz (commented):
@bethanylang Looks great, just one comment.

manuelRod:
@bethanylang don't forget to upload the copies on the tests also, otherwise, unit testing will keep failing. Besides that, I really like the new wording :)

modules/site-health/audit-autoloaded-options/load.php
mxbclang: @felixarntz My PHP knowledge is extremely basic at best – the best way that I can find to do this is to define that first sentence ("Autoloaded options are configuration settings...") as a variable like the options count and sizes are, so there would be three within that sentence. That seems like it could still be confusing for translation, though. Is there an easier way?

felixarntz: @bethanylang I've pushed an additional commit that implements this. Let me know if this looks good to you 👍

felixarntz:
@manuelRod I've fixed the failing tests, and in the process I also simplified them a bit. Using data providers is helpful when passing different type of data to the same test, but all providers in the test class only had a single data set, which makes it not very practical and harder to follow for no real benefit. I also updated the tests to look for the most important indicating factor to see that the logic in the Site Health check function works, rather than comparing the whole result, which shouldn't really be needed e.g. for copy changes like this the tests should preferably not fail.

felixarntz (approved):

manuelRod:
looks good @felixarntz ! Thanks for that

modules/site-health/audit-autoloaded-options/load.php
mxbclang: @felixarntz Looks great, thank you for taking care of that!

mitogh (approved):

PR #287, opened 2022-04-08
Introduce `webp_uploads_prefer_smaller_image_file` filter allowing to opt in to preferring the smaller image file

### LINKED ISSUE #286: Only use smaller file size based on opt-in since it is not part of the core feature scope
Follow-up to #186: While this enhancement certainly makes sense here, we need to consider that the WebP by default implementation for WordPress core as of now is not going to select the smaller file size, as that implementation doesn't even store file sizes for the generated images.

Due to this, I would advise we make this enhancement opt-in at this point, since otherwise it would explicitly behave differently from the core implementation, which to a degree would defeat the primary purpose of this plugin which is testing future core features. It may be that we eventually decide to add this to the core implementation, at which point we could change it to be enabled by default here as well, but right now this is clearly different behavior, so I think we should keep this piece as opt-in.

I suggest we add a filter to enable it, e.g. `webp_uploads_prefer_smaller_file`.

cc @jjgrainger @mitogh @adamsilverstein

felixarntz:
## Summary

Fixes #286

## Relevant technical choices

* A simpler solution would have been to just remove the `add_filter` call to the new function, but that would make it much less intuitive for developers, and they would need to use this "generic" filter, and remember the name of our "internal" function.
* A dedicated filter is much more intuitive and easy to use.
* I chose `webp_uploads_prefer_smaller_image_file` for the filter name for now.

## Checklist

### DIFF
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -599,6 +599,23 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
 	 */
 	$target_mimes = apply_filters( 'webp_uploads_content_image_mimes', array( 'image/webp', 'image/jpeg' ), $attachment_id, $context );
 
+	/**
+	 * Filters whether the smaller image should be used regardless of which MIME type is preferred overall.
+	 *
+	 * This is disabled by default only because it is not part of the current WordPress core feature proposal, as it
+	 * requires storing the file sizes, which the plugin implementation does but the core implementation does not.
+	 *
+	 * By enabling this, the plugin will compare the image file sizes and prefer the smaller file regardless of MIME
+	 * type.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param bool $prefer_smaller_image_file Whether to prefer the smaller image file.
+	 */
+	if ( apply_filters( 'webp_uploads_prefer_smaller_image_file', false ) ) {
+		$target_mimes = webp_uploads_get_mime_types_by_filesize( $target_mimes, $attachment_id );
+	}
+
 	$target_mime = null;
 	foreach ( $target_mimes as $mime ) {
 		if ( isset( $metadata['sources'][ $mime ] ) ) {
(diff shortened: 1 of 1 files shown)


mitogh (approved):

modules/images/webp-uploads/load.php
```diff
@@ -599,6 +599,23 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
 	 */
 	$target_mimes = apply_filters( 'webp_uploads_content_image_mimes', array( 'image/webp', 'image/jpeg' ), $attachment_id, $context );
 
+	/**
+	 * Filters whether the smaller image should be used regardless of which MIME type is preferred overall.
+	 *
+	 * This is disabled by default only because it is not part of the current WordPress core feature proposal, as it
+	 * requires storing the file sizes, which the plugin implementation does but the core implementation does not.
```
adamsilverstein: noting the current core implementation does include storing file sizes

modules/images/webp-uploads/load.php
```diff
@@ -599,6 +599,23 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+	 *
+	 * By enabling this, the plugin will compare the image file sizes and prefer the smaller file regardless of MIME
+	 * type.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param bool $prefer_smaller_image_file Whether to prefer the smaller image file.
+	 */
+	if ( apply_filters( 'webp_uploads_prefer_smaller_image_file', false ) ) {
+		$target_mimes = webp_uploads_get_mime_types_by_filesize( $target_mimes, $attachment_id );
```
adamsilverstein, line 615: probably should pass context here as well?

adamsilverstein (changes requested):
left some small feedback points

modules/images/webp-uploads/load.php
felixarntz: @adamsilverstein That would certainly be possible, but I'm thinking that wouldn't make so much sense here since this filter here is to toggle that entire part of the behavior. I can't see why anyone would want to enable this in one place but not another. WDYT about that?

adamsilverstein: yea, I'm not sure why you would want to do this in one context and not another, jut noticed it was missing. fine to leave off for now.

adamsilverstein (approved):

PR #298, opened 2022-04-14
Break apart single massive WebP Uploads file into multiple more manageable files

### LINKED ISSUE #297: Break down WebP uploads module into more manageable files
The WebP Uploads module codebase has grown quite a bit, but all the code is still in a single `load.php` file, **now with over 1,000 lines**. We should split out the codebase a little to make the individual pieces more maintainable and simplify oversight. The tests should be broken out in the same way.

I suggest the following:
* `load.php` keeps the main logic to generate additional MIME types (WebP by default) and replace them in the content.
* `helper.php` should be introduced for some generally reused functions (most importantly `webp_uploads_get_upload_image_mime_transforms()`, to make one example). This file should not add any hooks.
* `rest-api.php` should be introduced to have the REST API integration logic.
* `image-edit.php` should be introduced for all the somewhat separate logic around generating additional MIME types when editing an image, including backup and restore, which is tied to that process in WordPress.

Let's make sure the PR for this does not include any actual changes, because reviewing them would be super challenging. The PR should exclusively move code around, nothing else.

felixarntz:
## Summary

Fixes #297

## Relevant technical choices

* This PR exclusively moves functions around (and their corresponding tests), nothing else.
* This PR does not change _anything_ in any of the module behavior, so it's not necessary to review every touched line of code.
* All tests still pass without adjustments, and the logic behaves exactly like before this PR.

## Checklist


tests/modules/images/webp-uploads/rest-api-tests.php
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Tests for webp-uploads module rest-api.php.
+ *
+ * @package performance-lab
+ * @group   webp-uploads
+ */
+
+use PerformanceLab\Tests\TestCase\ImagesTestCase;
+
+class WebP_Uploads_REST_API_Tests extends ImagesTestCase {
```
mitogh: I don't think extending for `ImagesTestCase` is either required or necessary here, since those assertions are not used in this class.


```suggestion
class WebP_Uploads_REST_API_Tests extends WP_UnitTestCase {
```

tests/modules/images/webp-uploads/helper-tests.php
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * Tests for webp-uploads module helper.php.
+ *
+ * @package performance-lab
+ * @group   webp-uploads
+ */
+
+use PerformanceLab\Tests\TestCase\ImagesTestCase;
+
+class WebP_Uploads_Helper_Tests extends ImagesTestCase {
```
mitogh: Same comment as the previous one.

```suggestion
class WebP_Uploads_Helper_Tests extends WP_UnitTestCase {
```

mitogh (approved):
Changes look good apart for a minor. 

Thanks for splitting this into separate modules, it's much more manageable in this way.

adamsilverstein (approved):
Great! suggestions from @mitogh seem valid, otherwise good to go, thanks for breaking this up.

felixarntz:
@mitogh Updated!

PR #301, opened 2022-04-16
Only update the specified target images when an image is edited.

### LINKED ISSUE #299: When an image is edited, perform updates only to the specified target
## Feature Description

This is a follow up issue from 

- #158 
- #235 

When an image is edited a user is able to select multiple targets from the edit screen (see attached screenshot).

The edits can happen to:

- All the images
- Thumbnail - Only the thumbnail image is updated
- All sizes except the thumbnail 

When either of those targets are selected, make sure that only the target is modified and subsequent edits uses the specified source.

**All images**

This is the current behavior and when an edit takes place all the subsizes as well with the full image source is modified applying the specified updates, all the available mime types should reflect this change as well besides the default mime type.

**Thumbnail**

Only the thumbnail size is modified, the rest of the image sizes remains the same, the thumbnail and sources for the thumbnail should reflect this change making sure all the specified mime types create a source with the same details as the primary mime type for the image.

**All sizes except the thumbnail**

This behavior is similar to all images, with the difference that the changes shouldn't be applied to  thumbnail image sizes and sources for that subsize.

All the available subsizes, including the full image size should be modified with the specified changes, and the sources property should be created for all the specified mime types.

mitogh:
## Summary

Fixes #299 

## Relevant technical choices

Allow to select specific target images when editing an image, if either: thumbnail, all except thumbnail or all is selected the edit behavior should only use the specified images instead of applying the edit to all the images regardless of the selected target.

## Checklist


modules/images/webp-uploads/image-edit.php
```diff
@@ -19,18 +19,41 @@
...
+		 * If no original image was provided the image_directory can't be determined, in that scenario try to
+		 * find it from the `file` property.
+		 *
+		 * @see get_attached_file()
+		 */
+		if (
+			null === $image_directory
+			&& isset( $metadata['file'] )
+			&& 0 !== strpos( $metadata['file'], '/' )
+			&& ! preg_match( '|^.:\\\|', $metadata['file'] )
```
eugene-manuilov: Does it check whether the file path is Windows-based? Like `C:\...`?  I think the regular expression is incorrect because it has three back slashes instead of two.
```suggestion
			&& ! preg_match( '|^.:\\|', $metadata['file'] )
```

Or even better to just check the substring of the file path:
```suggestion
			&& ':\' !== substr( $metadata['file'], 1, 2 )
```

eugene-manuilov (changes requested):

modules/images/webp-uploads/image-edit.php
mitogh: > Does it check whether the file path is Windows-based? Like C:\...? I think the regular expression is incorrect because it has three back slashes instead of two.

Looks like that's the case, however this might be an existing bug in WP core as well as this code comes from core `get_attached_file` will replace with the `substr` instead. 

https://github.com/WordPress/wordpress-develop/blob/5.9/src/wp-includes/post.php#L725

mitogh: Actually is because the `\` is escaped that's why it needs 3 if you need an expression with 2 back slashs.

eugene-manuilov: > Actually is because the `\` is escaped that's why it needs 3 if you need an expression with

Ah... okay, thanks for clarifying... makes sense. 👍

eugene-manuilov (approved):
LGTM. Thanks, @mitogh.

modules/images/webp-uploads/image-edit.php
```diff
@@ -129,16 +166,38 @@ function ( $metadata, $post_meta_id ) use ( $post_id, $file_path, $mime_type, $e
...
+					if ( is_wp_error( $editor ) ) {
+						continue;
+					}
+
+					$current_extension = pathinfo( $thumbnail_file, PATHINFO_EXTENSION );
+					// Create a file with then new extension out of the targeted file.
+					$target_file_name                  = preg_replace( "/\.$current_extension$/", ".$extension", $thumbnail_file );
+					$target_file_location              = path_join( $original_directory, $target_file_name );
+					$result                            = $editor->save( $target_file_location, $targeted_mime );
+					$subsized_images[ $targeted_mime ] = array( 'thumbnail' => $result );
```
felixarntz: `$result` can be a `WP_Error`, so we need to check that before using it, similar to how it's already in the `else` clause below.

modules/images/webp-uploads/image-edit.php
```diff
@@ -111,8 +138,18 @@ function ( $metadata, $post_meta_id ) use ( $post_id, $file_path, $mime_type, $e
 			$filename           = pathinfo( $file_path, PATHINFO_FILENAME );
 			$main_images        = array();
 			$subsized_images    = array();
+
 			foreach ( $mime_transforms as $targeted_mime ) {
-				if ( $targeted_mime === $mime_type ) {
+				if ( $targeted_mime === $original_mime_type ) {
+					// If the target is `thumbnail` make sure is the only selected size.
```
felixarntz:
```suggestion
					// If the target is `thumbnail` make sure it is the only selected size.
```

modules/images/webp-uploads/image-edit.php
```diff
@@ -129,16 +166,38 @@ function ( $metadata, $post_meta_id ) use ( $post_id, $file_path, $mime_type, $e
...
-				$extension   = explode( '|', $allowed_mimes[ $targeted_mime ] );
-				$destination = trailingslashit( $original_directory ) . "{$filename}.{$extension[0]}";
-				$result      = $editor->save( $destination, $targeted_mime );
-
-				if ( is_wp_error( $result ) ) {
-					continue;
+				$extension = explode( '|', $allowed_mimes[ $targeted_mime ] );
+				$extension = $extension[0];
+
+				if ( 'thumbnail' === $target ) {
```
felixarntz, line 174: Let's add a similar comment here to explain:
```suggestion
				// If the target is `thumbnail` make sure only that size is generated.
				if ( 'thumbnail' === $target ) {
```

modules/images/webp-uploads/image-edit.php
```diff
@@ -91,15 +113,20 @@ function ( $metadata, $post_meta_id ) use ( $post_id, $file_path, $mime_type, $e
...
 			if ( empty( $metadata['sizes'] ) ) {
 				return $metadata;
 			}
 
 			$old_metadata = wp_get_attachment_metadata( $post_id );
 			$resize_sizes = array();
+			$target       = isset( $_REQUEST['target'] ) ? $_REQUEST['target'] : 'all';
+
 			foreach ( $old_metadata['sizes'] as $size_name => $size_details ) {
+				if ( 'nothumb' === $target && 'thumbnail' === $size_name ) {
```
felixarntz, line 127: Let's add a comment to explain:
```suggestion
				// If the target is 'nothumb', skip generating the 'thumbnail' size.
				if ( 'nothumb' === $target && 'thumbnail' === $size_name ) {
```

felixarntz (commented):
@mitogh Mostly lgtm, a few minor comments. One more critical thing, the rest is around doc comments.

felixarntz:
@mitogh @eugene-manuilov I've updated the base branch to `release/1.1.0`, with the goal to get this into the upcoming release. Would one of you be able to address the final feedback? There's also a merge conflict to address.

mitogh:
Thanks I can take a look at the feedback later today and resolve any remaining conflicts.

mxbclang:
:wave: @mitogh We're in a bit of a rush with this one in prep for Monday's release, so @mehulkaklotar is going to finish this one up. Thank you for continuing to work on this to this point!

modules/images/webp-uploads/image-edit.php
mehulkaklotar: Fixed this!

mitogh:
Sure @bethanylang sorry I was not able to get it done in time.

Thanks, @mehulkaklotar for taking this one to the finish line.

jjgrainger (approved):
Thanks @mehulkaklotar Latest changes look good to me

felixarntz (approved):
Thanks @mitogh and @mehulkaklotar!

PR #302, opened 2022-04-17
Refine logic to select smaller image file in the frontend based on `webp_uploads_prefer_smaller_image_file` filter

### LINKED ISSUE #292: Refine logic to select smaller image size
As outlined in https://github.com/WordPress/performance/issues/186#issuecomment-1093133540, there is a flaw with the current approach of choosing the smaller image MIME file in frontend content:

> * The current implementation only looks at the `full` image and decides based on that which MIME type should be preferred.
> * But what if e.g. for most image sizes (including `full`) the WebP image is smaller, but for e.g. `thumbnail` the JPEG image is smaller? The current implementation would still always use WebP in that scenario, but it would be better to use JPEG for the thumbnail size.
> * So I'm thinking maybe we need to enhance this to more deeply integrate into the replacement logic, per image size, rather than the overall MIME types filter.

While attaching the `webp_uploads_get_mime_types_by_filesize()` function to the `webp_uploads_content_image_mimes` filter seems elegant, I think this is the wrong integration point here. It's not sufficient to universally choose a preferred MIME type based on the `full` image size. We have to instead modify `webp_uploads_img_tag_update_mime_type()` directly and make this check **for every image size individually**.

jjgrainger:
## Summary

Fixes #292 

## Relevant technical choices

- Refines logic to select the smaller image file, regardless of mime type, when the `webp_uploads_prefer_smaller_image_file` filter is enabled.

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -451,7 +451,9 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
 	 *
 	 * @param bool $prefer_smaller_image_file Whether to prefer the smaller image file.
 	 */
-	if ( apply_filters( 'webp_uploads_prefer_smaller_image_file', false ) ) {
+	$prefer_smaller_image_file = apply_filters( 'webp_uploads_prefer_smaller_image_file', false );
+
+	if ( true === $prefer_smaller_image_file ) {
```
mitogh: Personally I like this style of using strict comparison however I have seen a common pattern for WordPress developers to just pass any value that evaluates to `true` and such I think just with checking that the value is not falsy would be enough.

```suggestion
	if ( $prefer_smaller_image_file ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -493,6 +495,14 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
 			continue;
 		}
 
+		// If smaller image file is preferred, and the target mime type is a larger file size, do not update.
+		if (
+			true === $prefer_smaller_image_file &&
+			$size_data['sources'][ $size_data['mime-type'] ]['filesize'] < $size_data['sources'][ $target_mime ]['filesize']
+		) {
```
mitogh: Same as the above.

```suggestion
		if (
			$prefer_smaller_image_file &&
			$size_data['sources'][ $size_data['mime-type'] ]['filesize'] < $size_data['sources'][ $target_mime ]['filesize']
		) {
```

modules/images/webp-uploads/load.php
```diff
@@ -451,7 +451,9 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
 	 *
 	 * @param bool $prefer_smaller_image_file Whether to prefer the smaller image file.
 	 */
-	if ( apply_filters( 'webp_uploads_prefer_smaller_image_file', false ) ) {
+	$prefer_smaller_image_file = apply_filters( 'webp_uploads_prefer_smaller_image_file', false );
+
+	if ( true === $prefer_smaller_image_file ) {
 		$target_mimes = webp_uploads_get_mime_types_by_filesize( $target_mimes, $attachment_id );
```
felixarntz: I don't think we need this function anymore. We don't need to choose a preferred MIME type based on file size if below we look at every size individually anyway. So I think it's better to remove this function and purely rely on the new logic you added below.

modules/images/webp-uploads/load.php
```diff
@@ -493,6 +495,14 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
 			continue;
 		}
 
+		// If smaller image file is preferred, and the target mime type is a larger file size, do not update.
+		if (
+			true === $prefer_smaller_image_file &&
```
felixarntz: Nit-pick: We don't need to be that specific, any truthy value should be accepted, as there's no harm in doing so.
```suggestion
			$prefer_smaller_image_file &&
```

modules/images/webp-uploads/load.php
```diff
@@ -493,6 +495,14 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+			$size_data['sources'][ $size_data['mime-type'] ]['filesize'] < $size_data['sources'][ $target_mime ]['filesize']
```
felixarntz: One thing we need to fix here: The `filesize` may be `0` in cases where determining the file size for whatever reason didn't work. If one of the two sizes here is 0, we don't have the necessary data, so in that case we should keep using the `$target_mime` version as it is the default.

Also, it'd be safer to check whether `isset( $size_data['sources'][ $size_data['mime-type'] ]['filesize'] )` (or maybe `! empty`) before accessing it, as on older images that data may not be present.

felixarntz (commented):
@jjgrainger Mostly looks good, it needs a few small refinements. Most importantly though, I think we should remove the function previously introduced since it's not really helpful or accurate anyway. The new approach here works good just by itself.

modules/images/webp-uploads/load.php
```diff
@@ -493,6 +477,29 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+		 * This is disabled by default only because it is not part of the current WordPress core feature proposal.
+		 *
+		 * By enabling this, the plugin will compare the image file sizes and prefer the smaller file regardless of MIME
+		 * type.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param bool $prefer_smaller_image_file Whether to prefer the smaller image file.
+		 */
+		if ( apply_filters( 'webp_uploads_prefer_smaller_image_file', false ) ) {
```
felixarntz: A bit of a micro-optimization, but I think it would be cleaner to apply this filter outside of any `foreach` loop here and assign the value to a variable that is then checked here. It shouldn't be necessary to re-run the filter for every size, and that could even possibly result in weird behavior since in theory the value could be different in different calls of the filter.

This is even more helpful in regards to my other comment though: We need to also add a similar file size comparison for the `full` image.

modules/images/webp-uploads/load.php
```diff
@@ -493,6 +477,29 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
 			continue;
 		}
```
felixarntz, line 510: Commenting here since GitHub doesn't let me comment on lines 455-464: That is where the full image is replaced, so we should add a similar file size comparison for that as well and only do the replacement if needed. Basically compare `$metadata['sources'][ $target_mime ]['filesize']` with `$metadata['sources'][ $original_mime_type ]['filesize']`.

This also should only happen based on the filter though, so this relates to my other comment. We should only call this filter once somewhere in the beginning of this function and then check the value just by checking the value stored in a variable.

felixarntz (commented):
@jjgrainger This looks close, two follow-up comments though. One is actually critical, the PR currently doesn't cover the full sized image yet.

Last but not least, can you also add tests for the new logic?

mitogh (commented):
Besides the comment from @felixarntz this looks good to me @jjgrainger 

Thanks.

modules/images/webp-uploads/load.php
jjgrainger: Thanks @felixarntz this has been updated

modules/images/webp-uploads/load.php
jjgrainger: Thanks @felixarntz I've updated the PR to apply the logic to the full image as well

modules/images/webp-uploads/load.php
```diff
@@ -427,6 +427,20 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
 
+	/**
+	 * Filters whether the smaller image should be used regardless of which MIME type is preferred overall.
+	 *
+	 * This is disabled by default only because it is not part of the current WordPress core feature proposal.
+	 *
+	 * By enabling this, the plugin will compare the image file sizes and prefer the smaller file regardless of MIME
+	 * type.
+	 *
+	 * @since n.e.x.t
```
felixarntz: The filter already exists since 1.0.0, i.e. this shouldn't get `n.e.x.t`. You only moved the filter elsewhere, but it's not new.
```suggestion
	 * @since 1.0.0
```

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -572,94 +572,23 @@ function () {
...
+		$updated_tag   = webp_uploads_img_tag_update_mime_type( $tag, 'the_content', $attachment_id );
 
-		$mime_types = webp_uploads_get_mime_types_by_filesize( array( 'image/jpeg', 'image/webp' ), $attachment_id, 'the_content' );
+		// The full size WebP is smaller.
+		$expected_tag = str_replace( $metadata['sources']['image/jpeg']['file'], $metadata['sources']['image/webp']['file'], $tag );
 
-		$this->assertIsArray( $mime_types );
-		$this->assertNotContains( 'image/invalid', $mime_types );
-		$this->assertSame( array( 'image/jpeg', 'image/webp' ), $mime_types );
+		$this->assertSame( $expected_tag, $updated_tag );
```
felixarntz: Is this test in fact doing what the comment above says? It says this test covers the case where JPEG is smaller than WebP, but then for the `$expected_tag` it replaces the JPEG version with WebP, so that seems not right. If the JPEG version is smaller, the test should expect that the JPEG version is still used even though a WebP version exists.

felixarntz (commented):
@jjgrainger Production logic now looks good to me, just one nit-pick on the filter and a potential issue with the test.

tests/modules/images/webp-uploads/load-tests.php
jjgrainger: Thanks @felixarntz 

I've updated the comments to make things clearer and reviewed the logic and can confirm it is correct. This image produces a smaller WebP file size of the full image. All sub sizes have a smaller JPG.

I've taken an example of the image code when the `webp_uploads_prefer_smaller_image_file` filter is set to `true` here:

```php
<img loading="lazy" width="1080" height="720" src="http://localhost:8888/wp-content/uploads/2022/04/leafs.webp" alt="" class="wp-image-15" srcset="http://localhost:8888/wp-content/uploads/2022/04/leafs.webp 1080w, http://localhost:8888/wp-content/uploads/2022/04/leafs-300x200.jpg 300w, http://localhost:8888/wp-content/uploads/2022/04/leafs-1024x683.jpg 1024w, http://localhost:8888/wp-content/uploads/2022/04/leafs-768x512.jpg 768w" sizes="(max-width: 1080px) 100vw, 1080px">
```

From the `srcset` you can see all sub sizes use the JPG version as they are smaller than their WebP version.

The replace updates the original image tag, which uses only JPG, to replace the full size with the WebP version in order to match the output of `webp_uploads_img_tag_update_mime_type ` when the `webp_uploads_prefer_smaller_image_file` is set to `true`.

I've also created a shorthand version of the metadata array to compare filesizes and this is correct.

```php
Array
(
    [file] => 2022/04/leafs.jpg
    [full] => Array
        (
            [image/jpeg] => 336884
            [image/webp] => 328322
        )

    [medium] => Array
        (
            [image/jpeg] => 19820
            [image/webp] => 22332
        )

    [large] => Array
        (
            [image/jpeg] => 237115
            [image/webp] => 257974
        )

    [thumbnail] => Array
        (
            [image/jpeg] => 8082
            [image/webp] => 9008
        )

    [medium_large] => Array
        (
            [image/jpeg] => 135786
            [image/webp] => 153226
        )
)
```

modules/images/webp-uploads/load.php
jjgrainger: Thanks @felixarntz updated

tests/modules/images/webp-uploads/load-tests.php
felixarntz: Ah, thanks for clarifying. That makes sense, in that case indeed the updated comments look better.

felixarntz (approved):
Great work @jjgrainger

felixarntz:
@adamsilverstein @mitogh Can one of you please review this as well so we can get it merged?

adamsilverstein (approved):
Change looks good to me (although I think webp_uploads_prefer_smaller_image_file could start true here), I'll work on adding something similar to the core patch.

modules/images/webp-uploads/load.php
```diff
@@ -451,9 +439,22 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
 	 *
 	 * @param bool $prefer_smaller_image_file Whether to prefer the smaller image file.
 	 */
-	if ( apply_filters( 'webp_uploads_prefer_smaller_image_file', false ) ) {
-		$target_mimes = webp_uploads_get_mime_types_by_filesize( $target_mimes, $attachment_id );
-	}
+	$prefer_smaller_image_file = apply_filters( 'webp_uploads_prefer_smaller_image_file', false );
```
adamsilverstein, line 442: working on adding to core patch so this could be true

modules/images/webp-uploads/load.php
```diff
@@ -469,11 +470,25 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
 	// Replace the full size image if present.
 	if ( isset( $metadata['sources'][ $target_mime ]['file'] ) ) {
+		// Initially set the target mime as the replacement source.
+		$replacement_source = $metadata['sources'][ $target_mime ]['file'];
+
+		// Check for the smaller image file.
+		if (
+			$prefer_smaller_image_file &&
+			! empty( $metadata['sources'][ $target_mime ]['filesize'] ) &&
+			! empty( $metadata['sources'][ $original_mime ]['filesize'] ) &&
```
adamsilverstein, line 480: Looking at this part I realized that there _could potentially be several mime types_ in the sources array (eg. JPEG, WebP, AVIF), we should probably check all of them for the smallest size, rather than just the original and the mapped type. 

What do you think @felixarntz ? Fine to address in a follow up so we can get this merged since it covers our primary use case as is.

felixarntz: @adamsilverstein That's a good point, I think eventually that would make sense. It would require a bit more complex reworking though indeed, for example right now we only have a single `$target_mime`, and then there will need to be a list, all of which we would need to go through.

+1 to doing it in a separate issue.

modules/images/webp-uploads/load.php
felixarntz: @adamsilverstein Not sure we should add this to the initial core patch as it feels like an extra enhancement. The main patch has a ton of code already and this enhancement could easily be separate.

adamsilverstein: Ok, I'll add it as a separate patch

modules/images/webp-uploads/load.php
adamsilverstein: I'm taking the "look thru every mime type for the smallest" approach for the (follow up) core patch, basically go thru all sources to find the smallest and use that:
```php
foreach( $metadata['sources'] as $source ) {
		if ( empty( $source['filesize'] ) ) {
			continue;
		}
		if ( $source['filesize'] < $target_file['filesize'] ) {
			$target_file = $source;
		}
	}
```

PR #314, opened 2022-04-24
Back up edited `full` image sources when restoring the original image

### LINKED ISSUE #295: Ensure edited image sources are backed up when restoring the original image
## Bug Description

Follow-up to #228: When editing an image in WordPress core (related: #158), the original image's sizes are backed up. This is primarily to restore them later. If then the original image is restored again or another round of edits to the same image is saved, the edited image's sizes get backed up. There is no way to restore these via the WordPress UI, but WordPress intentionally does not delete any edits at any point, and it keeps their metadata backups as well.

Our WebP uploads module does not behave in the same way with our `_wp_attachment_backup_sources` storage: When restoring the original image, the edited image data is not included in that backup anymore, while it _is_ included in core's `_wp_attachment_backup_sizes`. Note that this problem only concerns the `full` size in any case, because all sub-size sources are included in WordPress's own storage by definition. But it is a bug we need to fix.

This is probably not critical because there's no way to restore them, but it's still a bug in the WebP uploads module based on how WordPress core works. We do keep the files around as expected, but we don't keep the meta data backup around, which makes it even more inconsistent even with our own approach.

## Steps to reproduce

1. Upload a JPEG image with the WebP Uploads module active. WebP versions are generated in addition to the JPEG versions.
2. Edit the image through the WordPress UI (e.g. crop, rotate, anything). Then save.
3. View the attachment's `_wp_attachment_backup_sizes` and `_wp_attachment_backup_sources` post meta values. Both have the necessary data stored, the first for all individual sizes (including `sources` for each size), the latter just the main `sources`, i.e. for the original image (under `full-orig`, for consistency with the backup sizes).
4. Now restore the original.
5. View the attachment's `_wp_attachment_backup_sizes` and `_wp_attachment_backup_sources` post meta values. You will now recognize that the previous edit's file paths etc are present in `_wp_attachment_backup_sizes`, but not in `_wp_attachment_backup_sources` - the latter still only has the `full-orig` data and is missing the hashed data with references to the previously edited image, which is the problem we need to fix here.

mitogh:
## Summary

Store the sources and update the `_wp_attachment_backup_sources` meta when the image is restored, only update the meta if the meta location has not been updated yet.

Fixes #295 

## Relevant technical choices

A new class was introduced when the constant `IMAGE_EDIT_OVERWITE` is set to true due this needs to happen before WordPress has been bootrstaped, this class can be used as time move forwards for any additional scenario when this constant is defined as `true` and allowing the rest of the other tests to behave in a scenario when the constant is not defined. `@runInIsolation` can't be used as part of the tests due we are using clousures and the clousures can't be serialized when running a test in isolaion.

The backup tries to find the edited image when the image is restored in order to store it in the same way as when the image is edited. 

## Checklist


modules/images/webp-uploads/image-edit.php
```diff
@@ -320,10 +320,39 @@ function webp_uploads_get_next_full_size_key_from_backup( $attachment_id ) {
...
+	}
 
+	$backup_sources = get_post_meta( $attachment_id, '_wp_attachment_backup_sources', true );
 	if ( ! is_array( $backup_sources ) ) {
-		return $data;
+		$backup_sources = array();
+	}
+
+	if ( ! defined( 'IMAGE_EDIT_OVERWRITE' ) || ! IMAGE_EDIT_OVERWRITE ) {
+		$target = null;
```
felixarntz: Maybe let's call this variable `$suffix`? `$target` is more commonly used for the size target (e.g. "thumbnail"). WordPress core calls this `$suffix`, so I think it's better to align with that here for more clarity. While unrelated, I suggest we also adjust that in the existing code for `webp_uploads_backup_full_image_sources()`.

modules/images/webp-uploads/image-edit.php
```diff
@@ -320,10 +320,39 @@ function webp_uploads_get_next_full_size_key_from_backup( $attachment_id ) {
  * @return array The updated metadata of the attachment.
  */
 function webp_uploads_restore_image( $attachment_id, $data ) {
-	$backup_sources = get_post_meta( $attachment_id, '_wp_attachment_backup_sources', true );
+	if ( empty( $data['sources'] ) ) {
+		return $data;
+	}
```
felixarntz: How can we expect `sources` to be set here? This hook is called from core's `wp_update_attachment_metadata()` call, and core doesn't set `sources` within the metadata. I think we need to look at the _previous_ attachment metadata here via `wp_get_attachment_metadata( $attachment_id )`, similar to how we do it already in our `webp_uploads_backup_sources()` implementation.

modules/images/webp-uploads/image-edit.php
```diff
@@ -320,10 +320,39 @@ function webp_uploads_get_next_full_size_key_from_backup( $attachment_id ) {
...
 
+	$backup_sources = get_post_meta( $attachment_id, '_wp_attachment_backup_sources', true );
 	if ( ! is_array( $backup_sources ) ) {
-		return $data;
+		$backup_sources = array();
+	}
+
+	if ( ! defined( 'IMAGE_EDIT_OVERWRITE' ) || ! IMAGE_EDIT_OVERWRITE ) {
```
felixarntz: Let's not include support for this constant in this PR. That creates inconsistency because we don't support it elsewhere so far. There is an explicit TODO comment in the codebase to add support for it, and when we do that we should add support for the constant holistically. Right now you fix it here, but not when backing up sources, and it's better for now if we just keep this out.

In other words, let's always backup the sources regardless of the constant. We can later implement support for the constant separately.

modules/images/webp-uploads/image-edit.php
```diff
@@ -320,10 +320,39 @@ function webp_uploads_get_next_full_size_key_from_backup( $attachment_id ) {
...
+		}
+
+		if ( null === $target ) {
+			$target = 'orig';
+		}
+
+		if ( empty( $backup_sources[ "full-{$target}" ] ) ) {
+			$backup_sources[ "full-{$target}" ] = $data['sources'];
+			update_post_meta( $attachment_id, '_wp_attachment_backup_sources', $backup_sources );
+		}
```
felixarntz: I think the implementation here is a bit unnecessarily complicated: Backing up the previously set sources in here can work exactly the same as when handling an image edit, i.e. we can reuse a lot of the existing logic we already have.

As far as I can tell, we should be able to re-use our `webp_uploads_backup_sources()` function just as is. Basically, we need to call that from `webp_uploads_update_attachment_metadata()` right before calling `webp_uploads_restore_image()`. And then we don't need to modify `webp_uploads_restore_image()` at all.

felixarntz (changes requested):
@mitogh This is a bit of a detail, but the implementation here has still some disparities for how WordPress core works. Try editing and restoring an image and compare the data in `_wp_attachment_backup_sizes` with the data in `_wp_attachment_backup_sources`.

Note how the hash used on the array keys in "sizes" is different from the hash used on the files themselves. However, in our own "sources" the hash on the entries is the same as the hash used on the files. So while that may feel correct, it is not the same as core's behavior.

So the custom implementation here is a bit off still. After a closer inspection, I think we can actually simplify this a lot _and_ fix it at the same time. Basically, we need to call our existing `webp_uploads_backup_sources()` function even when restoring an image. I've tested this locally, and with that change the result is as expected, with the hashes in `webp_uploads_backup_sizes` and `webp_uploads_backup_sources` matching correctly.

So this can be simplified quite a bit:
* The changes here in `webp_uploads_restore_image()` are not necessary.
* Instead, we need to add `$data = webp_uploads_backup_sources( $attachment_id, $data );` to the `case 'wp_restore_image'` condition in `webp_uploads_update_attachment_metadata()`, right before calling `webp_uploads_restore_image( $attachment_id, $data )`.
* The tests here need to be updated accordingly. We need to validate the hashes in `webp_uploads_backup_sizes` and `webp_uploads_backup_sources` match.

modules/images/webp-uploads/image-edit.php
```diff
@@ -320,10 +320,39 @@ function webp_uploads_get_next_full_size_key_from_backup( $attachment_id ) {
...
+		$target = null;
+		foreach ( $data['sources'] as $mime_type => $properties ) {
+			if ( empty( $properties['file'] ) ) {
+				continue;
+			}
+
+			preg_match( '/-e(\d{13})/', $properties['file'], $matches );
+			if ( empty( $matches ) || count( $matches ) < 2 || empty( $matches[1] ) ) {
+				continue;
+			}
```
eugene-manuilov: It's better to check the result of the `preg_match` function call rather than `$matches` array:
```suggestion
			$matches = array();
			if ( ! preg_match( '/-e(\d{13})/', $properties['file'], $matches ) ) {
				continue;
			}
```

mitogh: Good call.

modules/images/webp-uploads/image-edit.php
mitogh: Good question, the main point of this function is not to set up `sources` if `sources` is not present we can assume the image does not have any available sources. Since this function is being called when the `restore` image is executed. At that point, we are not creating sources but restoring them so is safe to assume here if a `sources` is not present there's nothing to be restored (at least nothing custom created by this plugin) since this plugin is the one creating `sources` property.

modules/images/webp-uploads/image-edit.php
mitogh: Sure, I can rename it, the main reason behind this was called `target` is because it's the target suffix we are going to use from the existing ones.

mxbclang:
@mitogh Just checking in on this!

mitogh:
Thanks for the ping @bethanylang  will wrap the work on this one over the weekend so is ready for a review on Monday :)

mitogh:
Just a heads up @bethanylang that this PR is ready for another review due the feedback from @felixarntz  has been addressed. 

Let me know in case you need additional feedback about this one, or if there's anything else I can help with on this one. 

Thanks.

modules/images/webp-uploads/image-edit.php
```diff
@@ -387,9 +388,8 @@ function webp_uploads_get_next_full_size_key_from_backup( $attachment_id ) {
  */
 function webp_uploads_restore_image( $attachment_id, $data ) {
 	$backup_sources = get_post_meta( $attachment_id, '_wp_attachment_backup_sources', true );
-
 	if ( ! is_array( $backup_sources ) ) {
-		return $data;
+		$backup_sources = array();
```
felixarntz, line 392: @mitogh Just double-checking, what was the particular reason you've added this change? I'm not saying it's wrong, but I would think this would also work without this change? Just wanna make sure I'm not missing something.

felixarntz (approved):
Thanks @mitogh, looks great! One follow-up question for you, would be great if you could have a quick look so that we can merge this in time for 1.2.0.

modules/images/webp-uploads/image-edit.php
mitogh: Sure, the main reason was that in case the backup source was not set up this would initialize it to a new empty array instead, in order to set up the references to the full-size images. 

Is an edge case, but wanted to make sure if the edge case happens the stored value is consistent, instead of having no value even if we have backup values available.

mukeshpanchal27, line 392: I have the same question as Felix and would like to explore that edge case.

I tried this edge case in which the `_wp_attachment_backup_sources` meta value is null or not an array. As per the below code snippet, it will initialize an empty array for `$backup_sources`.

	if ( ! is_array( $backup_sources ) ) {
		$backup_sources = array();
	}

After this snippet, we have another snippet that checks two things if `full-orig` key is not set in array `$backup_sources` or it's not an array.

	if ( ! isset( $backup_sources['full-orig'] ) || ! is_array( $backup_sources['full-orig'] ) ) {
		return $data;
	}

In this edge case the function return `$data` so as I understood it is worth initializing an empty array instead of returning the `$data` as pervious it does.

Please let me know if I'm getting anything wrong.

mukeshpanchal27 (approved):
PR looks good to me and is ready to marge.

modules/images/webp-uploads/image-edit.php
felixarntz: From what I can tell it doesn't really matter though which way to go about this line of code, for the following reason:
* If `$backup_sources` is not an array, the following condition `! isset( $backup_sources['full-orig'] ) || ! is_array( $backup_sources['full-orig'] )` will also fore sure evaluate to `false`.
* So we can either initialize `$backup_sources` as empty array, causing `$data` to be returned in the following clause, or we can return `$data` right away. I don't think it changes the function behavior at all honestly.

cc @mitogh

PR #315, opened 2022-04-24
Sort modules by the specified criteria on #306

### LINKED ISSUE #306: Fix command line utilities to sort modules correctly
The various JS-based command line utilities that access module files to parse out some information (`npm run translations` and `npm run enabled-modules`) currently do so in random order. While this is not a major problem, it's certainly inconvenient and undesirable, as it may introduce changes to the generated files without any of the actual data in there changing.

We should fix it, so that these commands always treat modules in the expected order, sorted as follows:
1. By the order of the focus areas (see how they are defined in [`perflab_get_focus_areas()`](https://github.com/WordPress/performance/blob/trunk/admin/load.php#L170).
2. By whether they are experimental or not (non-experimental modules come first).
3. Alphabetically by module name.

For the implementation, we should be able to fix it for both commands by making sure this order is respected in the `getModuleData( modulesDir )` JS function reused by both of them.

mitogh:
## Summary

The introduced changes creates a predictable outcome of modules by using the criteria specified in #306 to order the modules first by focus area, then by experimental and lastly by module name, in this particular case the slug was used as the name due the slug is normalized into a lowercase strings.

Fixes #306

## Checklist


bin/plugin/commands/common.js
```diff
@@ -68,5 +82,26 @@ exports.getModuleData = async ( modulesDir ) => {
...
+					? 1
+					: -1;
+			}
+
+			const bothModulesAreExperimental =
+				firstModule.experimental && secondModule.experimental;
+			if (
+				! bothModulesAreExperimental &&
+				( firstModule.experimental || secondModule.experimental )
+			) {
```
felixarntz: This looks overly complicated and potentially slightly off? We only need to compare based on experimental state if only one of the two modules to compare is experimental, otherwise this property isn't relevant as they are equal in that sense (similar to how you do it for the focus areas).
```suggestion
			if ( firstModule.experimental !== secondModule.experimental ) {
```

bin/plugin/commands/common.js
```diff
@@ -33,6 +45,8 @@ exports.getModuleData = async ( modulesDir ) => {
 			const moduleData = {
 				slug: path.basename( moduleDir ),
 				focus: path.basename( path.dirname( moduleDir ) ),
+				// Set default values, in order to allow to avoid errors when ordering, if set property would be override.
+				experimental: false,
```
felixarntz: Why is this needed? Every module needs to have an `Experimental` header, so I think we should remove this default and instead ensure that it is required below, when calling `Array.filter` to remove incomplete entries.

bin/plugin/commands/common.js
```diff
@@ -68,5 +82,26 @@ exports.getModuleData = async ( modulesDir ) => {
 
 			return moduleData;
 		} )
-		.filter( ( moduleData ) => moduleData.name && moduleData.description );
+		.filter( ( moduleData ) => moduleData.name && moduleData.description )
```
felixarntz: See my comment above, it would be better to remove the above default and instead do the following extra validation here:
```suggestion
		.filter( ( moduleData ) => moduleData.name && moduleData.description && typeof moduleData.experimental !== undefined )
```

felixarntz (commented):
@mitogh Looks great, thanks! Just a few minor things worth improving.

bin/plugin/commands/common.js
mitogh: Mainly due this property is set dynamic via the headers, in case a header is not set we ensure this property is initialized to a `false` value.

bin/plugin/commands/common.js
mitogh: Agree that this would work as well, however I see it like we have 2 options:

1. Initialize to a safe default in case the experimental value is missing in order to include a module.
2. Not include the file but create a warning that a file is missing a required property.

Removing a module by missing a header without an error message might introduce some silent errors if we are not being explicit either on the parameters or error message on the reason why a module is not included. 

Let me know what do you think about that :)

bin/plugin/commands/common.js
mitogh: Woops looks like I over complicated things on that one, agree just checking on different values would be enough at this level, we might need some additional level inside of the block but that is a different story. :+1:

mitogh: Actually looks like the body of the code should work as it is :)

bin/plugin/commands/common.js
felixarntz: @mitogh That's a good point. How about we expand the `filter` function to also trigger warnings? I think it's okay to not error, we can just ignore such modules technically, but at least there should be a corresponding console output with warnings.

In other words, I think we should go with my originally suggested idea in terms of conditions to check, but expand the function based on your idea that if any of the 3 conditions is not satisfied, there should be a warning triggered in the console. Probably most helpful would be to have a specific warning per which of the 3 checks went wrong.

bin/plugin/commands/common.js
felixarntz: @mitogh See my reply to the other comment. I like your idea there with a warning, let's rely primarily on validation like that. In other words, I still think the default here can be removed as this will be validated later anyway.

bin/plugin/commands/common.js
mitogh: Great I think that's a good call, let me update this filter to include those warnings.

felixarntz (approved):
Great work @mitogh!

felixarntz:
@eugene-manuilov Can you please review this so that we can merge it soon?

bin/plugin/commands/common.js
```diff
@@ -68,5 +81,52 @@ exports.getModuleData = async ( modulesDir ) => {
...
+				);
+			}
+
+			if ( typeof moduleData.experimental === 'undefined' ) {
+				log(
+					formats.warning(
+						`This module was not included due is missing a required property 'experimental', Details: ${ moduleDetails }`
+					)
+				);
+			}
```
eugene-manuilov: Please, return `false` early to avoid having multiple warning messages for the same module if it has multiple properties empty. Or maybe would be better to determine which properties are missed and then output a single unified warning message with all missing properties. Something like this:

```
This module was not included because it misses required properties: 'name', 'experimental'.
```

bin/plugin/commands/common.js
```diff
@@ -68,5 +81,52 @@ exports.getModuleData = async ( modulesDir ) => {
...
+				typeof moduleData.experimental !== 'undefined'
+			);
+		} )
+		.sort( ( firstModule, secondModule ) => {
+			// Not the same focus group.
+			if ( firstModule.focus !== secondModule.focus ) {
+				return FOCUS_AREAS[ firstModule.focus ] >
+					FOCUS_AREAS[ secondModule.focus ]
+					? 1
+					: -1;
```
eugene-manuilov:
```suggestion
				return FOCUS_AREAS[ firstModule.focus ] - FOCUS_AREAS[ secondModule.focus ];
```

eugene-manuilov (changes requested):
Almost looks good to me. Left a comment and a nit-pick suggestion. Cris, please, take a look.

mitogh:
Thanks @eugene-manuilov would address your PR feedback over the weekend.

bin/plugin/commands/common.js
```diff
@@ -68,5 +81,53 @@ exports.getModuleData = async ( modulesDir ) => {
...
+
+			if ( requiredProperties.length >= 1 ) {
+				const properties = requiredProperties
+					.map( ( property ) => `'${ property }'` )
+					.join( ', ' );
+				log(
+					formats.warning(
+						`This module was not included because it misses required properties: ${ properties }. \nDetails: ${ JSON.stringify(
+							moduleData
+						) }`
```
eugene-manuilov: Let's probably output formatted details:
```suggestion
						`This module was not included because it misses required properties: ${ properties }.\nDetails: ${ JSON.stringify( moduleData, null, 4 ) }`
```

eugene-manuilov (approved):
Looks good to me. Thanks, @mitogh. Added one more suggestion that is just a nitpick and shouldn't block this PR. Approved.

felixarntz:
@eugene-manuilov @mitogh Can one of you please address the JS lint error here?

**Update:** Fixed it below.

PR #318, opened 2022-04-27
Introduce filter `webp_uploads_pre_generate_additional_image_source` to short-circuit generating additional image sources on upload

### LINKED ISSUE #307: Introduce filter to use a custom mechanism to create additional mime images
The main objective of this filter is to allow third party plugins or developers to replace the current mechanism provided by this plugin in order to introduce his own logic or mechanism to create the additional mime types.

This is a sub-issue of #160 and related to #311.

## Filter

`webp_uploads_pre_generate_additional_image_source`

### Paramaters

- `image` - By default a `null` value, or either `WP_Error` or an array with the details of the created image, in the same way the [`save()`](https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-image-editor.php#L82-L84) method is used.
- `attachment_id`: The `ID` of the attachment being processed
- `image_size`: The name of the image size we are looking for. When creating the additional mime type for the full image this value would be `full` in the same way it occurs in WordPress core.
- `image_details`: An array containing the properties of image size.
- `mime_type`: The targeted mime type of the file we are looking for.

## Integration

This filter should be located when a new image is created in order to avoid using the performance plugin mechanism.

Suggested location for creation of images, as the first section of the function, if the filter return an array with the file and path property return that value.

- https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/helper.php#L67-L68

**Note:** This filter should cover only the first time an image is created rather than including the subsequent edited versions to an image, a new filter would be created for subsequent edits to an image. For more details see: https://github.com/WordPress/performance/issues/307#issuecomment-1109116528

~Suggested location for subsequent updates.~ 

- ~https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/image-edit.php#L134~

~When an image is updated from within the image editor, this location in order to be fired for every single image size.~

mehulkaklotar:
## Summary

Fixes #307 

## Checklist


modules/images/webp-uploads/helper.php
```diff
@@ -65,6 +65,28 @@ function webp_uploads_get_upload_image_mime_transforms() {
...
+	 *
+	 * @since n.e.xt
+	 *
+	 * @param array  $image         Image data {'path'=>string, 'file'=>string, 'width'=>int, 'height'=>int, 'mime-type'=>string} or empty array.
+	 * @param int    $attachment_id The ID of the attachment from where this image would be created.
+	 * @param string $size          The size name that would be used to create this image, out of the registered subsizes.
+	 * @param array  $size_data     An array with the dimensions of the image: height, width and crop {'height'=>int, 'width'=>int, 'crop'}.
+	 * @param string $mime          The target mime in which the image should be created.
+	 */
+	$image = (array) apply_filters( 'webp_uploads_pre_generate_additional_image_source', array(), $attachment_id, 'full', $size_data, $mime );
```
mitogh:
```suggestion
	$image = (array) apply_filters( 'webp_uploads_pre_generate_additional_image_source', null, $attachment_id, 'full', $size_data, $mime );
```

mitogh: Looks this function uses all the time `full` but that's not the accurate as the size changes depending on the parameters passed to this function.

mitogh: I don't think we should cast to `array`  due the filter can return `WP_Error` as well.

mitogh (commented):
Some comments, would be great to add some tests around the filter as well.

modules/images/webp-uploads/helper.php
mehulkaklotar: @mitogh I have removed cast to `array` now.

About the size default to full for all the sizes, I am confused on how do we get the the size name here as we don't have any size name related parameters in the function. so should we get the size name using `wp_get_registered_image_subsizes` and looping through the sizes and the get the matching size? Like this:

```
$sizes      = wp_get_registered_image_subsizes();
	$image_size = 'full';

	if ( ! empty( $sizes ) ) {
		foreach ( $sizes as $image_size_key => $size ) {
			if ( isset( $size['width'] )
				&& isset( $size['height'] )
				&& isset( $size['crop'] )
				&& (
					isset( $size_data['width'] )
					&& $size['width'] === $size_data['width']
					)
				&& (
					isset( $size_data['height'] )
					&& $size['height'] === $size_data['height']
				)
				&& (
					isset( $size_data['crop'] )
					&& $size['crop'] === $size_data['crop']
				)
			) {
				$image_size = $image_size_key;
				break;
			}
		}
	}

	/**
	 * Filter to allow the generation of additional image sources, in which a defined mime type
	 * can be transformed and create additional mime types for the file.
	 *
	 * @since n.e.xt
	 *
	 * @param array  $image         Image data {'path'=>string, 'file'=>string, 'width'=>int, 'height'=>int, 'mime-type'=>string} or null or WP_Error.
	 * @param int    $attachment_id The ID of the attachment from where this image would be created.
	 * @param string $image_size    The size name that would be used to create this image, out of the registered subsizes.
	 * @param array  $size_data     An array with the dimensions of the image: height, width and crop {'height'=>int, 'width'=>int, 'crop'}.
	 * @param string $mime          The target mime in which the image should be created.
	 *
	 * @return array|null|WP_Error An array with the file and filesize if the image was created correctly otherwise a WP_Error
	 */
	$image = apply_filters( 'webp_uploads_pre_generate_additional_image_source', null, $attachment_id, $image_size, $size_data, $mime );
```

mitogh: Good call I think one of the ways we can do it is by adding a new parameter to the top level function with the size name as the caller to `webp_uploads_generate_additional_image_source` has the information about the size name.

Let me know what do you think about that. 

I did a quick search and looks like this one is only used on the `full` size tho but I might be wrong here.

modules/images/webp-uploads/helper.php
```diff
@@ -65,6 +65,33 @@ function webp_uploads_get_upload_image_mime_transforms() {
...
+	 * @param string $mime          The target mime in which the image should be created.
+	 *
+	 * @return array|null|WP_Error An array with the file and filesize if the image was created correctly otherwise a WP_Error
+	 */
+	$image = apply_filters( 'webp_uploads_pre_generate_additional_image_source', null, $attachment_id, 'full', $size_data, $mime );
+
+	if ( ! is_wp_error( $image )
+		&& ! empty( $image['file'] )
+		&& ! empty( $image['path'] )
+	) {
```
eugene-manuilov: I think we can just check whether `$image` is an array and not check for wp_error at all:

```suggestion
	if (
		is_array( $image ) &&
		! empty( $image['file'] ) &&
		! empty( $image['path'] )
	) {
```

modules/images/webp-uploads/helper.php
```diff
@@ -65,6 +65,33 @@ function webp_uploads_get_upload_image_mime_transforms() {
  * @return array|WP_Error An array with the file and filesize if the image was created correctly otherwise a WP_Error
  */
 function webp_uploads_generate_additional_image_source( $attachment_id, array $size_data, $mime, $destination_file_name = null ) {
+
+	/**
+	 * Filter to allow the generation of additional image sources, in which a defined mime type
+	 * can be transformed and create additional mime types for the file.
+	 *
+	 * @since n.e.xt
```
eugene-manuilov:
```suggestion
	 * @since n.e.x.t
```

modules/images/webp-uploads/helper.php
```diff
@@ -65,6 +65,33 @@ function webp_uploads_get_upload_image_mime_transforms() {
...
+	 *
+	 * @param array  $image         Image data {'path'=>string, 'file'=>string, 'width'=>int, 'height'=>int, 'mime-type'=>string} or null or WP_Error.
+	 * @param int    $attachment_id The ID of the attachment from where this image would be created.
+	 * @param string $image_size    The size name that would be used to create this image, out of the registered subsizes.
+	 * @param array  $size_data     An array with the dimensions of the image: height, width and crop {'height'=>int, 'width'=>int, 'crop'}.
+	 * @param string $mime          The target mime in which the image should be created.
+	 *
```
eugene-manuilov, line 78:
```suggestion
```

tests/modules/images/webp-uploads/helper-tests.php
```diff
@@ -231,4 +231,37 @@ function () {
...
+
+	/**
+	 * Create an image with the filter webp_uploads_pre_generate_additional_image_source added.
+	 *
+	 * @test
+	 */
+	public function it_should_create_an_image_with_filter_webp_uploads_pre_generate_additional_image_source() {
+		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/car.jpeg' );
+
+		add_filter(
```
eugene-manuilov, line 251: We need to call `remove_all_filters( 'webp_uploads_pre_generate_additional_image_source' )` before adding a new filter to make sure that no other hooks are registered for this filter.

eugene-manuilov (changes requested):

tests/modules/images/webp-uploads/helper-tests.php
mehulkaklotar: @eugene-manuilov Thanks, I have updated the changes.

eugene-manuilov (approved):

mitogh (commented):
LGTM

However it looks like we might want to introduce a new filter in the `return` value as currently, we are assuming the image would be in the same directory as the WordPress installation and calling `filesize` for files stored in a bucket like S3 this is not the case, I think in those cases we can rely on a new filter being used so plugins can define the logic in how the `filesize` is calculated.

For example;

```php
return apply_filters( 'filter_name', array... 
```

Specifically here: 

- https://github.com/WordPress/performance/blob/2efa53adeef298f27117dddc8006bb3eed2b3e12/modules/images/webp-uploads/helper.php#L89-L92

mitogh (approved):

modules/images/webp-uploads/helper.php
```diff
@@ -65,6 +65,33 @@ function webp_uploads_get_upload_image_mime_transforms() {
...
  */
 function webp_uploads_generate_additional_image_source( $attachment_id, array $size_data, $mime, $destination_file_name = null ) {
+
+	/**
+	 * Filter to allow the generation of additional image sources, in which a defined mime type
+	 * can be transformed and create additional mime types for the file.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array  $image         Image data {'path'=>string, 'file'=>string, 'width'=>int, 'height'=>int, 'mime-type'=>string} or null or WP_Error.
```
felixarntz: This needs to be `array|WP_Error|null`. When updating, make sure to indent the following param docs correctly.

modules/images/webp-uploads/helper.php
```diff
@@ -65,6 +65,33 @@ function webp_uploads_get_upload_image_mime_transforms() {
...
+	if (
+		is_array( $image ) &&
+		! empty( $image['file'] ) &&
+		! empty( $image['path'] )
+	) {
+		return array(
+			'file'     => $image['file'],
+			'filesize' => filesize( $image['path'] ),
+		);
+	}
```
felixarntz: This logic is missing the `is_wp_error` check. A `WP_Error` is an explicitly documented possible type for `$image`, which should be used if the custom image generation process failed. We need to add that here. Please provide an additional test for that as well.

modules/images/webp-uploads/helper.php
```diff
@@ -65,6 +65,33 @@ function webp_uploads_get_upload_image_mime_transforms() {
  * @return array|WP_Error An array with the file and filesize if the image was created correctly otherwise a WP_Error
  */
 function webp_uploads_generate_additional_image_source( $attachment_id, array $size_data, $mime, $destination_file_name = null ) {
+
+	/**
+	 * Filter to allow the generation of additional image sources, in which a defined mime type
+	 * can be transformed and create additional mime types for the file.
+	 *
```
felixarntz, line 74:
```suggestion
	 * can be transformed and create additional mime types for the file.
	 *
	 * Returning an image data array or WP_Error here effectively short-circuits the default logic to generate the image source.
	 *
```

felixarntz (changes requested):
@mehulkaklotar The implementation is missing support for the possible `WP_Error` return type of the filter, which is critical and needs to be fixed. Other than that, just a documentation enhancement suggestion.

cc @eugene-manuilov @mitogh

felixarntz:
@mehulkaklotar I've updated the base branch to `release/1.1.0`, with the goal to get this into the upcoming release. It would be great if you could implement the requested changes by tomorrow so that we can include this filter in the release. 🙌

modules/images/webp-uploads/helper.php
mehulkaklotar: Fixed now!

modules/images/webp-uploads/helper.php
mehulkaklotar: Check is added now with few new tests.

modules/images/webp-uploads/helper.php
```diff
@@ -65,6 +65,45 @@ function webp_uploads_get_upload_image_mime_transforms() {
...
+			return $image;
+		}
+
+		if ( empty( $image['file'] ) ) {
+			return new WP_Error( 'image_file_not_present', __( 'The file key is not present on the image data', 'performance-lab' ) );
+		}
+
+		if ( empty( $image['path'] ) || ! file_exists( $image['path'] ) ) {
+			return new WP_Error( 'image_path_not_present', __( 'The file path is not present on the image data', 'performance-lab' ) );
+		}
```
felixarntz: These conditions shouldn't be there, we shouldn't return errors when the filter is being misused by a developer. Your original condition to only return the array below if `is_array( $image ) && ! empty( $image['file'] ) && ! empty( $image['path'] )` was more accurate here.

We should preferably fail silently if someone returns something from the filter that they shouldn't.

mehulkaklotar:
@felixarntz I am still not sure about the `$image_size` parameter in the filter here. https://github.com/WordPress/performance/pull/318#discussion_r861679822

so are we going to make bydefault `full` for all image size in filter? because it doesn't really make sense as I tested it and it turns out all image size data is being worked on in that.

modules/images/webp-uploads/helper.php
```diff
@@ -65,6 +65,45 @@ function webp_uploads_get_upload_image_mime_transforms() {
...
+	 * @param string               $image_size    The size name that would be used to create this image, out of the registered subsizes.
+	 * @param array                $size_data     An array with the dimensions of the image: height, width and crop {'height'=>int, 'width'=>int, 'crop'}.
+	 * @param string               $mime          The target mime in which the image should be created.
+	 *
+	 * @return array|null|WP_Error An array with the file and filesize if the image was created correctly otherwise a WP_Error
+	 */
+	$image = apply_filters( 'webp_uploads_pre_generate_additional_image_source', null, $attachment_id, 'full', $size_data, $mime );
+
+	// Try to check for not empty returned filtered image.
+	if ( ! empty( $image ) ) {
```
felixarntz: I don't think we need this actual wrapper, it's redundant I'd say.

Checking that either `is_wp_error` for the one use-case and then alternatively checking for `is_array( $image ) && ! empty( $image['file'] ) && ! empty( $image['path'] )` for the other use-case should be sufficient.

felixarntz (commented):
@mehulkaklotar Thank you for making the changes, I have two follow-up comments.

modules/images/webp-uploads/helper.php
```diff
@@ -65,6 +65,45 @@ function webp_uploads_get_upload_image_mime_transforms() {
...
+	 *
+	 * @param array|null|WP_Error  $image         Image data {'path'=>string, 'file'=>string, 'width'=>int, 'height'=>int, 'mime-type'=>string} or null or WP_Error.
+	 * @param int                  $attachment_id The ID of the attachment from where this image would be created.
+	 * @param string               $image_size    The size name that would be used to create this image, out of the registered subsizes.
+	 * @param array                $size_data     An array with the dimensions of the image: height, width and crop {'height'=>int, 'width'=>int, 'crop'}.
+	 * @param string               $mime          The target mime in which the image should be created.
+	 *
+	 * @return array|null|WP_Error An array with the file and filesize if the image was created correctly otherwise a WP_Error
+	 */
+	$image = apply_filters( 'webp_uploads_pre_generate_additional_image_source', null, $attachment_id, 'full', $size_data, $mime );
```
felixarntz: @mehulkaklotar 
> so are we going to make bydefault `full` for all image size in filter? because it doesn't really make sense as I tested it and it turns out all image size data is being worked on in that.

Ahh, sorry I had totally missed this. Yes, this absolutely shouldn't show `full` for any image size. We will need to provide the name of the image size where the `$size_data` comes from. We don't have access to that currently from this function, so let's add a new function parameter `$size_name` right before `$size_data`.

We then need to pass this new parameter accordingly everywhere that `webp_uploads_generate_additional_image_source()` is called. Can you please make that update here?

modules/images/webp-uploads/helper.php
mehulkaklotar: Fixed now!

modules/images/webp-uploads/helper.php
mehulkaklotar: I have added one check for `WP_Error` and return and one check for `$image` to return array of image data.

modules/images/webp-uploads/helper.php
mehulkaklotar: I have added changes for this in the latest commit.

modules/images/webp-uploads/helper.php
felixarntz: @mehulkaklotar This comment wasn't addressed prior, but it is valid: Per WP docs standards there should be no blank line between `@param` and `@return`.

modules/images/webp-uploads/load.php
```diff
@@ -72,6 +72,25 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+				&& $size['width'] === $original_size_data['width']
+				&& $size['height'] === $original_size_data['height']
+				&& $size['crop'] === $original_size_data['crop']
+			) {
+				$image_size_name = $image_size_key;
+				break;
+			}
+		}
+	}
+
```
felixarntz: This isn't really accurate, but more importantly it isn't needed: This function only deals with the `full` image size (for all other image sizes the `webp_uploads_generate_image_size()` function is used).

So this set of changes can be removed entirely, instead we can pass the hard-coded string `full` to `webp_uploads_generate_additional_image_source()` below.

felixarntz (commented):
@mehulkaklotar One more follow-up around the newly added change, which can be simplified.

modules/images/webp-uploads/helper.php
felixarntz: Thanks @mehulkaklotar. Please review my new feedback in https://github.com/WordPress/performance/pull/318#pullrequestreview-969948455 - part of your new implementation isn't needed, as the one function by definition is only used for the `full` image size, so we don't need to try to find the matching size, it's always `full` in that scenario.

modules/images/webp-uploads/load.php
mehulkaklotar: @felixarntz This changes has been done now.

mehulkaklotar:
@felixarntz We should look into [this comment](https://github.com/WordPress/performance/pull/318#pullrequestreview-965487438) from @mitogh about introducing another filter for returning value for function. 

Should we introduce that in this change or make new PR for it?

jjgrainger:
> @felixarntz We should look into [this comment](https://github.com/WordPress/performance/pull/318#pullrequestreview-965487438) from @mitogh about introducing another filter for returning value for function.
> 
> Should we introduce that in this change or make new PR for it?

@mehulkaklotar Is there a reason we couldn't expect the `webp_uploads_pre_generate_additional_image_source` filter to return the image array with the `file` and `filesize` keys? This would match the comment for the filter and also allow plugins to calculate the filesize appropriately depending on how the image is stored.

I would want @felixarntz and @mitogh to input here just incase I've overlooked something.

Thanks!

jjgrainger (approved):
Looks good to me, I've added a comment based on adding an additional filter to confirm https://github.com/WordPress/performance/pull/318#issuecomment-1125111322

felixarntz (approved):
Thank you @mehulkaklotar!

Regarding @mitogh's comment, that's an important point. I think a good solution would be to allow for an optional `path` key within the return array, which we would use instead of `file` in case it is set. Let's open a follow-up issue for that and work on it for the 1.2.0 plugin release.

PR #319, opened 2022-04-27
Introduce filter `webp_uploads_pre_replace_additional_image_source` to short-circuit replacing additional image sources in frontend content

### LINKED ISSUE #311: Introduce filter to use custom logic to find additional mime types images
The main objective of this filter is to allow third-party plugins or developers to change the current logic used for this plugin (the performance plugin) to locate additional mime images, for example, if a plugin stores the images in a different location and the file names are created out of a custom hash functions rather than in a predictable way, this filter would allow developers to provide the location and logic to replace images in the content of the site. 

This is a sub-issue of #160 and related to #307.

## Filter

`webp_uploads_pre_replace_additional_image_source`

## Parameters

- `image`: The value of the replaced image, if the image is the same as the value returned from the filter the replacement logic is used, meaning that the image was not modified.The HTML code of the image being modified.
- `attachment_id`: The ID of the attachment being processed.
- `image_size`: The name of the image size we are looking for. When creating the additional mime types for the original image this value would be `full`, in the same way it occurs in WordPress core.
- `mime_type`: The targeted mime type of the file we are looking for.
- `context`: Context is under where the image is being used.

## Integration

The suggestion location to add this filter are the following places:

- https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/load.php#L474
- https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/load.php#L496

### Suggested approach

```php
$filtered_image = apply_filters( 'webp_uploads_pre_replace_additional_image_source', $image, $attachment_id, $image_size, $mime_type, $context );
// Only plugin logic to replace the image if the `filtered_image` is the same as the image, meaning it was not modified.
if ( $filtered_image === $image ) {
      $filtered_image = str_replace( ... );
}
```

### Note

> We need to make sure the filter runs regardless of whether our own replacement mechanism finds a file. 

In other words, some of the conditions in our code need to be checked only after the new filter is run, since those conditions are only relevant for our own way to replace the URLs, but potentially not for how a plugin that stores image files elsewhere would need to replace them.

mehulkaklotar:
## Summary

Fixes #311 

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -498,6 +512,20 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+		 *
+		 * @since n.e.xt
+		 *
+		 * @param string $image         An <img> tag where the urls would be updated.
+		 * @param int    $attachment_id The ID of the attachment being modified.
+		 * @param string $size          The size name that would be used to create this image, out of the registered subsizes.
+		 * @param string $target_mime   The target mime in which the image should be created.
+		 * @param string $context       The context where this is function is being used.
+		 */
+		$image = (string) apply_filters( 'webp_uploads_pre_replace_additional_image_source', $image, $attachment_id, 'full', $target_mime, $context );
```
mitogh: This filter should be placed above the line 510 and the code from `str_replace` would be executed only if the image is different, take a look at `suggested approach` here: 

- https://github.com/WordPress/performance/issues/311

The `full` value in this instance should be the name of the size being modified at this point in time.

mitogh: A couple of tests would be great as well.

mehulkaklotar: @mitogh I have added changes as per your feedback in the latest commits.

modules/images/webp-uploads/load.php
```diff
@@ -471,11 +471,31 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
-				$basename,
-				$metadata['sources'][ $target_mime ]['file'],
-				$image
-			);
+
+			/**
+			 * Filter to replace additional image source file, by locating the original
+			 * mime types of the file and return correct file path in the end.
+			 *
+			 * @since n.e.xt
```
eugene-manuilov:
```suggestion
			 * @since n.e.x.t
```

modules/images/webp-uploads/load.php
```diff
@@ -493,11 +513,30 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
-		$image = str_replace(
-			$size_data['file'],
-			$size_data['sources'][ $target_mime ]['file'],
-			$image
-		);
+		/**
+		 * Filter to replace additional image source file, by locating the original
+		 * mime types of the file and return correct file path in the end.
+		 *
+		 * @since n.e.xt
```
eugene-manuilov:
```suggestion
		 * @since n.e.x.t
```

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -662,4 +662,25 @@ function( $data, $attachment_id ) {
...
+
+	/**
+	 * Prevent replacing an image if image was uploaded via external source or plugin.
+	 *
+	 * @group webp_uploads_update_image_references
+	 *
+	 * @test
+	 */
+	public function it_should_prevent_replacing_an_image_uploaded_via_external_source() {
+		add_filter(
```
eugene-manuilov, line 622: We need to call `remove_all_filters( 'webp_uploads_pre_replace_additional_image_source' )` before adding a new hook to make sure that no other hooks are registered for this filter.

eugene-manuilov (changes requested):

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: Thanks @eugene-manuilov I have added changes according to your feedback now.

eugene-manuilov (approved):

mukeshpanchal27:
Hi there!

Can we adjust the **webp_uploads_pre_replace_additional_image_source** filter code and doc block so it can be documented only one time and for other instances, we can add the below context as we use in WordPress?

`/** This filter is documented in modules/images/webp-uploads/load.php */`


Code changes that require for this.

Replace this line https://github.com/WordPress/performance/blob/enhancement/replace-additional-image-filter/modules/images/webp-uploads/load.php#L503 code with `foreach ( $metadata['sizes'] as $size => $size_data ) {`

Replace this whole doc block with https://github.com/WordPress/performance/blob/enhancement/replace-additional-image-filter/modules/images/webp-uploads/load.php#L516-L528 below context.

```
/** This filter is documented in modules/images/webp-uploads/load.php */
$filtered_image = (string) apply_filters( 'webp_uploads_pre_replace_additional_image_source', $image, $attachment_id, $size, $target_mime, $context );
```

Pull request: #324

Please let me it and let me know your thoughts.

mitogh (approved):

modules/images/webp-uploads/load.php
```diff
@@ -486,16 +486,36 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
-			$image = str_replace(
-				$basename,
-				$replacement_source,
-				$image
-			);
+
+			/**
+			 * Filter to replace additional image source file, by locating the original
+			 * mime types of the file and return correct file path in the end.
+			 *
```
felixarntz, line 493:
```suggestion
			 * mime types of the file and return correct file path in the end.
			 *
			 * Altering the $image tag through this filter effectively short-circuits the default replacement logic using the preferred MIME type.
			 *
```

modules/images/webp-uploads/load.php
```diff
@@ -486,16 +486,36 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
+			 * @param string $image         An <img> tag where the urls would be updated.
+			 * @param int    $attachment_id The ID of the attachment being modified.
+			 * @param string $size          The size name that would be used to create this image, out of the registered subsizes.
+			 * @param string $target_mime   The target mime in which the image should be created.
+			 * @param string $context       The context where this is function is being used.
+			 */
+			$filtered_image = (string) apply_filters( 'webp_uploads_pre_replace_additional_image_source', $image, $attachment_id, 'full', $target_mime, $context );
+
+			// Check if filtered image is same as the image, replace the image otherwise not.
```
```suggestion
			// If filtered image is same as the image, run our own replacement logic, otherwise rely on the filtered image.
```

modules/images/webp-uploads/load.php
```diff
@@ -518,11 +538,19 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
 
-		$image = str_replace(
-			$size_data['file'],
-			$size_data['sources'][ $target_mime ]['file'],
-			$image
-		);
+		/** This filter is documented in modules/images/webp-uploads/load.php */
+		$filtered_image = (string) apply_filters( 'webp_uploads_pre_replace_additional_image_source', $image, $attachment_id, $size, $target_mime, $context );
+
+		// Check if filtered image is same as the image, replace the image otherwise not.
```
```suggestion
		// If filtered image is same as the image, run our own replacement logic, otherwise rely on the filtered image.
```

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -608,4 +608,27 @@ public function it_should_replace_the_featured_image_to_web_p_when_requesting_th
...
+	 *
+	 * @test
+	 */
+	public function it_should_prevent_replacing_an_image_uploaded_via_external_source() {
+		remove_all_filters( 'webp_uploads_pre_replace_additional_image_source' );
+
+		add_filter(
+			'webp_uploads_pre_replace_additional_image_source',
+			function() {
+				return TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg';
```
felixarntz: While this is just a test, what the code here is doing is not correct based on what the filter is for: The filter receives the full image HTML tag, but here the code is replacing it with just a URL to an image. Let's make sure we use a proper example and instead return an **altered `img` tag** that e.g. includes this image URL as its `src`.

felixarntz (commented):
@mehulkaklotar Production code looks good to me. I have two minor documentation suggestions, more importantly though the test here should be updated to be a more appropriate usage: The filter is for an image **HTML tag**, so it shouldn't be used to replace the image tag with just a URL.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: yes, I have fixed it with `img` tag instead.

felixarntz (approved):
Great, thank you @mehulkaklotar!

PR #322, opened 2022-05-03
Add Performance Lab generator meta tag to `wp_head` output

### LINKED ISSUE #321: Add generator tag with plugin version and modules
## Feature Description

Given that the Performance Lab plugin is a beta testing plugin for future WordPress core performance features, it is crucial that through the plugin we can showcase how a certain feature/module improves performance as expected, and through that to prove and allow data-based reasoning around whether it should or should not be in WordPress core in its current state.

A transparent way to gather performance-related data is to rely on [HTTP Archive](https://httparchive.org/) and its public data sets. HTTP Archive uses [Wappalyzer](https://www.wappalyzer.com/) (see [its GitHub repo](https://github.com/wappalyzer/wappalyzer)) to identify web technologies, also used to identify many popular WordPress plugins.

This issue proposes the addition of a `meta[name="generator"]` tag to the Performance Lab plugin, which allows tools like Wappalyzer to identify...
* whether a site is using the Performance Lab plugin
* which version of the Performance Lab plugin it uses
* which modules of the Performance Lab plugin it has enabled

The proposed structure for the `content` value of the tag is: `Performance Lab <version>; modules: <active modules>`
For example: `<meta name="generator" content="Performance Lab 1.1.0; modules: images/webp-uploads, site-health/audit-autoloaded-options, object-cache/persistent-object-cache-health-check">`

While this tag is crucial for gathering performance-related metrics at scale, some sites prefer to remove any generator tags. While there is no real benefit in doing so, like with WordPress core's own generator tag, the proposed implementation would allow to easily unhook the logic, should a site prefer not to include it.

felixarntz:
## Summary

Fixes #321

## Relevant technical choices

* Includes a `perflab_get_generator_content()` function in addition to the main `perflab_render_generator()` function that renders the tag. The first function allows to use that information in another way (e.g. WP core supports generators in a variety of variants) and also to test it more appropriately.
* The `perflab_render_generator()` function is hooked into `wp_head`.

## Checklist


adamsilverstein:
+1 looks good and support adding this. Since you hooked on head, this would be straightforward to exclude if users want that (remove the hook).

adamsilverstein (approved):

jeffpaul:
We should be rather explicit in the changelog about this as it sounds like it automatically opts a site into this data gathering though I think some details from the related issue help spell out the rationale here (and hopefully anyone who's installed the plugin would be open to the generator tag if that helps prove that certain modules are worth bringing into core).

ThierryA (approved):
LGTM, very supportive of this (to the extent that I believe WP core should provide plugin information automatically without plugins needing to add their own tags 😄 ). Even more so for this plugin specifically since it will enable further measurement which is crucial to validate features which are aimed at being implemented in WordPress core.

eugene-manuilov (approved):
LGTM

felixarntz:
@jeffpaul 

> We should be rather explicit in the changelog about this as it sounds like it automatically opts a site into this data gathering though I think some details from the related issue help spell out the rationale here (and hopefully anyone who's installed the plugin would be open to the generator tag if that helps prove that certain modules are worth bringing into core).

Definitely, this should be highlighted in the changelog (which it will be given that this PR is marked as a feature). Any data gathering based on this though would be similar to WordPress's own generator tag, which is there by default. But I agree that it needs to be obvious in the changelog.

mitogh (commented):
LGTM 🌮

mitogh (approved):

load.php
```diff
@@ -137,6 +137,37 @@ function( $module_settings ) {
...
+ * Displays the HTML generator tag for the Performance Lab plugin.
+ *
+ * See {@see 'wp_head'}.
+ *
+ * @since n.e.x.t
+ */
+function perflab_render_generator() {
+	$content = perflab_get_generator_content();
+
+	echo '<meta name="generator" content="' . esc_attr( $content ) . '">';
```
mukeshpanchal27: For better readability can we append **\n** at the end of the tag? and the end tag for the meta. In WordPress core, we follow this format. https://github.com/WordPress/WordPress/blob/master/wp-includes/robots-template.php#L49

`echo '<meta name="generator" content="' . esc_attr( $content ) . '"  />' . "\n";`

felixarntz: @mukeshpanchal27 Great point, updated!

mitogh (approved):

mukeshpanchal27 (approved):

aristath (approved):

dajana91:
how to remove the unnecessary performance lab generator tag from the meta tags? i don't want to have it in the source code. i also removed the wordpress generator tag via functions.php so that you don't see the wordpress version in the source code. what exactly is the function for functions.php to remove the performance lab meta tag?

I found the solution myself.

Just put the following line in the functions.php:

remove_action('wp_head', 'perflab_render_generator');

And you're rid of the unnecessary entry.

mxbclang:
@dajana91 Thank you for your feedback here and in your forum post. If we do decide to remove this tag, we will open a separate issue.

felixarntz:
@dajana91 If you prefer to remove the generator tag, you can do by adding the following code to your site:

```php
remove_action( 'wp_head', 'perflab_render_generator' );
```

However, note that the generator tag exists to be able to have public performance metrics in [HTTP Archive](https://httparchive.org) to assess and prove how certain modules of the Performance Lab plugin improve performance, which is important in order to get them eventually merged into WordPress core. By removing the tag, the site will no longer contribute to these.

dajana91:
@felixarntz I don't care. I don't want unasked additional information in my source code that doesn't belong there.

PR #325, opened 2022-05-04
WIP: Generate images on the fly

eugene-manuilov:
## Summary

This PR adds functionality that allows us to generate images for missing image sizes on the fly using the standard `https://site.domain/wp-content/uploads/{year}/{month}/{image}-{width}x{height}.jpeg` URLs. These URLs are generated automatically for missing sub sizes and added to the `srcset` attribute, which means that new images will be generated only if they are really needed and requested by the browser.

You can test it on your local by switching your current theme to another one and checking generated `srcset` attributes which should contain URLs for new sub sizes used by the new theme even though actual files don't exist on the server yet.

Fixes #

## Checklist


mehulkaklotar:
@eugene-manuilov I tested this on local & it seems to be not working for some images even if WebP for those images exists.

Screenshot 2022-05-05 at 5 36 17 PM

Screenshot 2022-05-05 at 5 39 06 PM

eugene-manuilov:
@mehulkaklotar can you, please, provide more information as to which image is suppose to be generated and which steps i need to do on my end to reproduce it on my end?

mehulkaklotar:
@eugene-manuilov The image `image-alignment-300x200-1.jpg` is suppose to be `image-alignment-300x200-1.webp` as WebP image exists for it. If you need to replicate it, here is what I did: 

1. Import posts using theme unit test data https://codex.wordpress.org/Theme_Unit_Test
2. Go to Block Image post.
3. It will have various images on the post and those are different sizes for the images.

modules/images/webp-uploads/generation.php
```diff
@@ -0,0 +1,236 @@
...
+ *
+ * @since n.e.x.t
+ *
+ * @param string $img           An <img> tag.
+ * @param string $context       The context where this is function is being used.
+ * @param int    $attachment_id The ID of the attachment being modified.
+ * @return string The updated img tag.
+ */
+function webp_uploads_update_img_tag_srcset( $img, $context, $attachment_id ) {
+	list( $src, $width, $height ) = wp_get_attachment_image_src( $attachment_id, 'large', false );
```
mitogh: Do we want to review that the returned value from `wp_get_attachment_image_src` is not `false` or that the values from `$src`, `$width`, and `$height` are not `null`?

modules/images/webp-uploads/generation.php
```diff
@@ -0,0 +1,236 @@
...
+		}
+	}
+
+	// Bail if the subsize name is not found.
+	if ( empty( $size ) ) {
+		return;
+	}
+
+	// Bail if we can't find an attachment with requested file.
+	$attachment_id = attachment_url_to_postid( "{$uploads_dir['baseurl']}/{$matches[1]}.{$matches[4]}" );
```
mitogh, line 90: Would be great to use `path_join` here instead, that would help with creating a variable that can be used in this function a bit down.

```suggestion
    $path = path_join( $uploads_dir['baseurl'], "{$matches[1]}.{$matches[4]}" );
	$attachment_id = attachment_url_to_postid( $path );
```

modules/images/webp-uploads/generation.php
```diff
@@ -0,0 +1,236 @@
...
+	if ( empty( $metadata ) ) {
+		return;
+	}
+
+	require_once ABSPATH . '/wp-admin/includes/image.php';
+
+	// Generate missing sub size.
+	$metadata = _wp_make_subsizes(
+		array( $size => $sizes[ $size ] ),
+		"{$uploads_dir['basedir']}/{$matches[1]}.{$matches[4]}",
```
mitogh, line 106: Variable can be used here.

modules/images/webp-uploads/generation.php
```diff
@@ -0,0 +1,236 @@
...
+		$attachment_id
+	);
+
+	// Generate sources.
+	$metadata = webp_uploads_create_sources_property( $metadata, $attachment_id );
+
+	// Save updated metadata.
+	wp_update_attachment_metadata( $attachment_id, $metadata );
+
+	$image_file = $uploads_dir['basedir'] . $relpath;
```
mitogh, line 118: We should make sure the file exists.

```php
if ( ! file_exists( $image_file ) ) {
  return;
}
```

modules/images/webp-uploads/generation.php
```diff
@@ -0,0 +1,236 @@
...
+
+	// Send headers if headers haven't been sent yet.
+	if ( ! headers_sent() ) {
```
mitogh: How about flipping and adding an early return instead?

```suggestion
	if ( headers_sent() ) {
	  return
	}
```

modules/images/webp-uploads/generation.php
```diff
@@ -0,0 +1,236 @@
...
+		$filetype = wp_check_filetype( $image_file );
+		if ( ! empty( $filetype['type'] ) ) {
+			header( 'Content-type: ' . $filetype['type'] );
+		}
+
+		header( 'Content-Length: ' . filesize( $image_file ) );
+		header( 'Cache-Control: public, max-age=' . 0 );
```
mitogh: No need for an append :) 

```suggestion
		header( 'Cache-Control: public, max-age=0');
```

modules/images/webp-uploads/generation.php
```diff
@@ -0,0 +1,236 @@
...
+ *
+ * @param array  $sources       One or more arrays of source data to include in the 'srcset'.
+ * @param array  $size_array    An array of requested width and height values.
+ * @param string $image_src     The 'src' of the image.
+ * @param array  $image_meta    The image meta data as returned by 'wp_get_attachment_metadata()'.
+ * @param int    $attachment_id Image attachment ID or 0.
+ * @return array Srcset sources array.
+ */
+function webp_uploads_generate_image_srcset( $sources, $size_array, $image_src, $image_meta, $attachment_id ) {
+	$image_src_parts = wp_parse_url( $image_src );
```
mitogh, line 157: We should make sure the returned value from `wp_parse_url` is not `false` or make sure is a string.

modules/images/webp-uploads/generation.php
eugene-manuilov: It won't help because those are two different values, in the first one we have `baseurl` whilst in the second one we use `basedir`. So, we can't use one variable for both.
https://github.com/WordPress/performance/blob/3a2ad73e4b190aca24c46ca0485a46f540715825/modules/images/webp-uploads/generation.php#L90 https://github.com/WordPress/performance/blob/3a2ad73e4b190aca24c46ca0485a46f540715825/modules/images/webp-uploads/generation.php#L106

modules/images/webp-uploads/generation.php
eugene-manuilov: See my comment for the previous suggestion.

modules/images/webp-uploads/generation.php
eugene-manuilov: Yep, makes sense. Updated.

modules/images/webp-uploads/generation.php
eugene-manuilov: Yep, updated.

modules/images/webp-uploads/generation.php
eugene-manuilov: Yep, works for me. Updated.

modules/images/webp-uploads/generation.php
eugene-manuilov: Lol... very true... updated. 🙈

modules/images/webp-uploads/generation.php
eugene-manuilov: Agreed. Updated.

eugene-manuilov:
> @eugene-manuilov The image `image-alignment-300x200-1.jpg` is suppose to be `image-alignment-300x200-1.webp` as WebP image exists for it. If you need to replicate it, here is what I did:
> 
> 1. Import posts using theme unit test data https://codex.wordpress.org/Theme_Unit_Test
> 2. Go to Block Image post.
> 3. It will have various images on the post and those are different sizes for the images.

@mehulkaklotar if the only problem is that the image is not in the `webp` format, then that's okay because correct mime type selection isn't implemented yet in this PoC.

modules/images/webp-uploads/generation.php
```diff
@@ -0,0 +1,250 @@
...
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+/**
+ * Generates images on the fly for the registered image sizes and available transformations.
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_generate_missing_images() {
```
felixarntz, line 14: Since this is a POC, it's okay to have this as part of the WebP Uploads module, but if this became part of the plugin, it should be its own module instead, as it's not tied to WebP.

It would be great if you could clarify that in the PR description, to avoid confusion about that.

modules/images/webp-uploads/generation.php
```diff
@@ -0,0 +1,250 @@
...
+	if ( ! preg_match( '#/(.+)-(\d+)x(\d+)\.([a-zA-Z]+)$#', $relpath, $matches ) ) {
+		return;
+	}
+
+	$img_width  = intval( $matches[2] );
+	$img_height = intval( $matches[3] );
+
+	$size       = '';
+	$size_names = get_intermediate_image_sizes();
+	$sizes      = wp_get_additional_image_sizes();
```
felixarntz, line 40: Why not use [`wp_get_registered_image_subsizes()`](https://developer.wordpress.org/reference/functions/wp_get_registered_image_subsizes/) instead? You could reduce a lot of the code here, also the lookup of width and height you're doing below.

modules/images/webp-uploads/generation.php
```diff
@@ -0,0 +1,250 @@
...
+		"{$uploads_dir['basedir']}/{$matches[1]}.{$matches[4]}",
+		$metadata,
+		$attachment_id
+	);
+
+	// Generate sources.
+	$metadata = webp_uploads_create_sources_property( $metadata, $attachment_id );
+
+	// Save updated metadata.
+	wp_update_attachment_metadata( $attachment_id, $metadata );
```
felixarntz, line 115: As mentioned before, this is the biggest risk here I think. The code generates additional images in an unauthenticated request. How do we intend to protect this against abuse? You mentioned to use a scheduled task instead, but as far as I can tell that is not part of this.

I know it's a POC, but I think that is a critical concern that we have to find a way to address.

modules/images/webp-uploads/generation.php
```diff
@@ -0,0 +1,250 @@
...
+		return;
+	}
+
+	$uploads_dir = wp_upload_dir();
+	$baseurl     = wp_parse_url( $uploads_dir['baseurl'], PHP_URL_PATH );
+
+	// Bail if the current request is not to an uploaded image.
+	if ( substr( $_SERVER['REQUEST_URI'], 0, strlen( $baseurl ) ) !== $baseurl ) {
+		return;
+	}
```
felixarntz, line 25: As far as I can tell, you're using the "regular" URLs that look like they point to within the `uploads` directory. That won't work for a core implementation, as several environments are configured to never route "real-looking" direct file URLs to PHP. We need to use some sort of dedicated endpoint for this, with its own URL, to make it reliable for core.

felixarntz (commented):
@eugene-manuilov I've left a few comments.

This POC helps illustrating the idea, which is a good start. However, I currently don't see how it would address any of the previous concerns raised in our conversations about it, e.g. around security and reliability. I realize this is a POC so that may not have been the goal here, but it still has me question how realistic such a project would be to land in core.

To better support the idea, it would be great to think further about how to address the concerns raised.

mukeshpanchal27:
> @eugene-manuilov The image `image-alignment-300x200-1.jpg` is suppose to be `image-alignment-300x200-1.webp` as WebP image exists for it. If you need to replicate it, here is what I did:
> 
> 1. Import posts using theme unit test data https://codex.wordpress.org/Theme_Unit_Test
> 2. Go to Block Image post.
> 3. It will have various images on the post and those are different sizes for the images.

@mehulkaklotar I have tried one solution in PR #340  can you please check it?

spacedmonkey:
So the front end would generate these images? There maybe contexts, where front servers do not have access to write to the file system. It may also mean that a google index of a site, might result in 100s or 1000s of images being generated at once. This will file up file systems / may take down servers. This functionality may not scale.

eugene-manuilov:
> It may also mean that a google index of a site, might result in 100s or 1000s of images being generated at once. This will file up file systems / may take down servers. This functionality may not scale.

@spacedmonkey, yes, we know about this problem and have an idea how to deal with this problem (move images generation into a background tasks).

This PR is PoC that is intended to research possibilities to make sub sizes generation more "mindful" (instead generating dozens of subsizes for every image, generate only those subsizes that are really used). Nothing set in stone yet and any suggestions that help to make it better are welcome.

> So the front end would generate these images? There maybe contexts, where front servers do not have access to write to the file system.

Backend will generate those images, but requests to start the regeneration routing will come from frontend. So, there shouldn't be write permission issues, I think.

adamsilverstein (commented):
Generally think we should shift the approach to using the background process for this, rather than purely on demand.

PR #329, opened 2022-05-09
Correct label for persistent object cache Site Health check

### LINKED ISSUE #328: Correct persistent object cache label
## Bug Description

In persistent object cache module file - [modules/object-cache/persistent-object-cache-health-check/load.php](https://github.com/WordPress/performance/blob/trunk/modules/object-cache/persistent-object-cache-health-check/load.php#L22) it uses a key in the label it should be label like other site health module label https://github.com/WordPress/performance/blob/trunk/modules/site-health/webp-support/load.php#L21

mukeshpanchal27:
Fixes #328

## Checklist

### DIFF
--- a/modules/object-cache/persistent-object-cache-health-check/load.php
+++ b/modules/object-cache/persistent-object-cache-health-check/load.php
@@ -19,7 +19,7 @@
 function perflab_oc_health_add_tests( $tests ) {
 	if ( wp_get_environment_type() === 'production' ) {
 		$tests['direct']['persistent_object_cache'] = array(
-			'label' => 'persistent_object_cache',
+			'label' => esc_html__( 'Persistent object cache', 'performance-lab' ),
 			'test'  => 'perflab_oc_health_persistent_object_cache',
 		);
 	}
--- a/modules/object-cache/persistent-object-cache-health-check/load.php
+++ b/modules/object-cache/persistent-object-cache-health-check/load.php
@@ -19,7 +19,7 @@
 function perflab_oc_health_add_tests( $tests ) {
 	if ( wp_get_environment_type() === 'production' ) {
 		$tests['direct']['persistent_object_cache'] = array(
-			'label' => esc_html__( 'Persistent object cache', 'performance-lab' ),
+			'label' => __( 'Persistent object cache', 'performance-lab' ),
 			'test'  => 'perflab_oc_health_persistent_object_cache',
 		);
 	}


tillkruss (approved):

spacedmonkey (approved):

modules/object-cache/persistent-object-cache-health-check/load.php
```diff
@@ -19,7 +19,7 @@
 function perflab_oc_health_add_tests( $tests ) {
 	if ( wp_get_environment_type() === 'production' ) {
 		$tests['direct']['persistent_object_cache'] = array(
-			'label' => 'persistent_object_cache',
+			'label' => esc_html__( 'Persistent object cache', 'performance-lab' ),
```
felixarntz: No need for `esc_html__` here, that function should only be used when outputting content. See https://developer.wordpress.org/themes/theme-security/data-sanitization-escaping/#escaping-securing-output:
> Output escaping should occur as late as possible.
```suggestion
			'label' => __( 'Persistent object cache', 'performance-lab' ),
```

felixarntz (commented):
@mukeshpanchal27 One small thing to address here, other than that lgtm.

modules/object-cache/persistent-object-cache-health-check/load.php
mukeshpanchal27: @felixarntz Thank you for the feedback.

In this can do we need to change other labels that use the `esc_html__` function?

https://github.com/WordPress/performance/blob/trunk/modules/site-health/audit-autoloaded-options/load.php#L21
https://github.com/WordPress/performance/blob/trunk/modules/site-health/webp-support/load.php#L21
https://github.com/WordPress/performance/blob/trunk/modules/site-health/audit-enqueued-assets/load.php#L115
https://github.com/WordPress/performance/blob/trunk/modules/site-health/audit-enqueued-assets/load.php#L119

felixarntz: @mukeshpanchal27 Ah great point! I had missed these were using `esc_html__` as well. In that case, I'd say feel free to update them maybe in a separate pull request?

mukeshpanchal27: Ok. I will add new PR for that files.

felixarntz (approved):
Thanks @mukeshpanchal27!

PR #334, opened 2022-05-13
Update `webp_uploads_pre_generate_additional_image_source` filter to allow returning file size

### LINKED ISSUE #333: Enhance `webp_uploads_pre_generate_additional_image_source` filter to support returning `filesize` value
Follow-up to #307: As identified by @mitogh in https://github.com/WordPress/performance/pull/318#pullrequestreview-971118682 and then further discussed in https://github.com/WordPress/performance/issues/160#issuecomment-1125271238, we should allow developers to optionally return the file size of a custom generated additional image source when using the `webp_uploads_pre_generate_additional_image_source`. This is crucial for implementations where e.g. the image is hosted in a CDN, which would mean that calling `filesize()` on a path would not be feasible.

It is probably best to make the `filesize` field optional, for two reasons:
* To not break any existing filter usage, because the current version of the filter will already be released as part of plugin version 1.1.0.
* To keep things simpler for plugins that _do_ generate files within the WordPress environment, in which case WordPress can just call `filesize()` on the `path` returned.

Since the `path` returned is only relevant for calling `filesize()`, I suggest we implement the change as follows:
* Continue to require the `file` key to be included in the array.
* In addition to that, require _either_ `path` _or_ `filesize` to be included in the array.
* In case of `path`, keep calling `filesize()` on it like today.
* In case of `filesize`, use that value as is to return it within the result array.

eugene-manuilov:
## Summary

Fixes #333

## Checklist

### DIFF
--- a/modules/images/webp-uploads/helper.php
+++ b/modules/images/webp-uploads/helper.php
@@ -67,7 +67,6 @@ function webp_uploads_get_upload_image_mime_transforms() {
  * @return array|WP_Error An array with the file and filesize if the image was created correctly otherwise a WP_Error
  */
 function webp_uploads_generate_additional_image_source( $attachment_id, $image_size, array $size_data, $mime, $destination_file_name = null ) {
-
 	/**
 	 * Filter to allow the generation of additional image sources, in which a defined mime type
 	 * can be transformed and create additional mime types for the file.
@@ -84,19 +83,23 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 	 * @return array|null|WP_Error An array with the file and filesize if the image was created correctly otherwise a WP_Error
 	 */
 	$image = apply_filters( 'webp_uploads_pre_generate_additional_image_source', null, $attachment_id, $image_size, $size_data, $mime );
-
 	if ( is_wp_error( $image ) ) {
 		return $image;
 	}
 
 	if (
-		is_array( $image )
-		&& ! empty( $image['file'] )
-		&& ! empty( $image['path'] )
+		is_array( $image ) &&
+		! empty( $image['file'] ) &&
+		(
+			! empty( $image['path'] ) ||
+			! empty( $image['filesize'] )
+		)
 	) {
 		return array(
 			'file'     => $image['file'],
-			'filesize' => filesize( $image['path'] ),
+			'filesize' => ! empty( $image['filesize'] )
+				? $image['filesize']
+				: filesize( $image['path'] ),
 		);
 	}
 
@@ -110,14 +113,11 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 	}
 
 	$image_path = wp_get_original_image_path( $attachment_id );
-
-	// File does not exist.
 	if ( ! file_exists( $image_path ) ) {
 		return new WP_Error( 'original_image_file_not_found', __( 'The original image file does not exists, subsizes are created out of the original image.', 'performance-lab' ) );
 	}
 
 	$editor = wp_get_image_editor( $image_path, array( 'mime_type' => $mime ) );
-
 	if ( is_wp_error( $editor ) ) {
 		return $editor;
 	}
@@ -125,7 +125,6 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 	$height = isset( $size_data['height'] ) ? (int) $size_data['height'] : 0;
 	$width  = isset( $size_data['width'] ) ? (int) $size_data['width'] : 0;
 	$crop   = isset( $size_data['crop'] ) && $size_data['crop'];
-
 	if ( $width <= 0 && $height <= 0 ) {
 		return new WP_Error( 'image_wrong_dimensions', __( 'At least one of the dimensions must be a positive number.', 'performance-lab' ) );
 	}
@@ -144,7 +143,6 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 	}
 
 	$image = $editor->save( $destination_file_name, $mime );
-
 	if ( is_wp_error( $image ) ) {
 		return $image;
 	}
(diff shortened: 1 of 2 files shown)


modules/images/webp-uploads/helper.php
```diff
@@ -84,19 +83,23 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
...
+			! empty( $image['path'] ) ||
+			! empty( $image['filesize'] )
+		)
 	) {
 		return array(
 			'file'     => $image['file'],
-			'filesize' => filesize( $image['path'] ),
+			'filesize' => ! empty( $image['filesize'] )
+				? $image['filesize']
+				: filesize( $image['path'] ),
```
mitogh: Probably personal preference but I rather swap the condition here to remove the `!` operator.

```suggestion
			'filesize' => empty( $image['filesize'] )
				? filesize( $image['path'] )
				: $image['filesize'],
```

Another thing (unlikely) is the fact that `empty` would return `false` even when the key is present for `filesize` but the value is `0` or not truthy. I think would be safer to use either `isset` or `array_key_exists` instead.

```suggestion
			'filesize' => isset( $image['filesize'] )
				? $image['filesize']
				: filesize( $image['path'] ),
```

Or

```suggestion
			'filesize' => array_key_exists( $image,'filesize' )
				? $image['filesize']
				: filesize( $image['path'] ),
```

modules/images/webp-uploads/helper.php
```diff
@@ -84,19 +83,23 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
...
 
 	if (
-		is_array( $image )
-		&& ! empty( $image['file'] )
-		&& ! empty( $image['path'] )
+		is_array( $image ) &&
+		! empty( $image['file'] ) &&
+		(
+			! empty( $image['path'] ) ||
+			! empty( $image['filesize'] )
```
mitogh: Same comments apply here for the scenario where the key is present but contains a falsy value, would be safer to go with `isset` or `array_key_exists`

modules/images/webp-uploads/helper.php
eugene-manuilov: Yep. Makes sense. Updated.

modules/images/webp-uploads/helper.php
eugene-manuilov: Updated.

modules/images/webp-uploads/helper.php
```diff
@@ -84,19 +83,23 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
...
+			! empty( $image['path'] ) ||
+			array_key_exists( 'filesize', $image )
+		)
 	) {
 		return array(
 			'file'     => $image['file'],
-			'filesize' => filesize( $image['path'] ),
+			'filesize' => array_key_exists( 'filesize', $image )
+				? $image['filesize']
+				: filesize( $image['path'] ),
```
mehulkaklotar, line 102: We are also returning same array in later in this function. https://github.com/WordPress/performance/pull/334/files#diff-1f653bd37f60a6476485a6794f7003a91b3713868c7414588d5835332fe68642R156

so can we put checks there as well?

mehulkaklotar (changes requested):

modules/images/webp-uploads/helper.php
eugene-manuilov: That is different. The `$image` variable is returned from the image editor which doesn't return file size information, thus we don't need to check for the `filesize` property there.

tests/modules/images/webp-uploads/helper-tests.php
```diff
@@ -270,7 +270,33 @@ function () {
...
 		$this->assertStringEndsWith( 'image.webp', $result['file'] );
-		$this->assertFileExists( '/tmp/image.webp' );
+	}
+
+	/**
+	 * Tests the webp_uploads_pre_generate_additional_image_source filter returning filesize property.
+	 *
+	 * @test
+	 */
+	public function it_should_return_an_error_when_filter_webp_uploads_pre_generate_additional_image_source_returns_filesize() {
```
felixarntz: This name is not quite right as far as I can tell? Where in this test is there any error to be returned?

felixarntz (commented):
@eugene-manuilov Looks pretty much good to go, just one tiny comment.

tests/modules/images/webp-uploads/helper-tests.php
eugene-manuilov: Good catch. Forgot to change that part of the name after copying the test with an error. Now it should be good, do you mind checking it again?

tests/modules/images/webp-uploads/helper-tests.php
```diff
@@ -270,7 +270,33 @@ function () {
...
+			'webp_uploads_pre_generate_additional_image_source',
+			function () {
+				return array(
+					'file'     => 'image.webp',
+					'filesize' => 777,
+				);
+			}
+		);
+
+		$result = webp_uploads_generate_additional_image_source( 0, 'medium', array(), 'image/webp', '/tmp/image.jpg' );
```
felixarntz: Not a blocker, but I'd be in favor of using a real `$attachment_id` and `$size_data` here, similar to the test method right above. Yes, it is strictly not necessary, but I think it's better to use data that is appropriate for what the function expects. Unless we are testing explicitly how it behaves with "weird" data, which is not the case here though.

felixarntz (approved):
@eugene-manuilov Looks good!

One remaining non-blocking comment.

jjgrainger (approved):

tests/modules/images/webp-uploads/helper-tests.php
eugene-manuilov: Ok, updated.

felixarntz (approved):

mehulkaklotar (approved):
LGTM 🚀

PR #340, opened 2022-05-21
Update old content image to webp

### LINKED ISSUE #339: Old content images not show in webp format
## Bug Description

As discussed on #325. 

Some old content images already have WebP format images but they will not load because the new image has a different attachment ID than the old attachment ID.

## Steps to reproduce
1. Activate the Performance Lab plugin
2. Import posts and attachments using theme unit test data https://codex.wordpress.org/Theme_Unit_Test
3. Go to Block Image post.
4. It will have various images on the post and those are different sizes for the images.

mukeshpanchal27:
## Summary

Fixes #339

## Checklist

### DIFF
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -416,11 +416,11 @@ function webp_uploads_update_image_references( $content ) {
 		}
 		
 		// Get attachment id by image src
-		$updated_id = attachment_url_to_postid( $image_src[1] );
+		$updated_attachment_id = attachment_url_to_postid( $image_src[1] );
 		
-		if ( $attachment_id !== $updated_id ) {
-			$updated_img = preg_replace( '/wp-image-([\d]+)/i', sprintf( 'wp-image-%d', $updated_id ), $img );
-			$content = str_replace( $img, webp_uploads_img_tag_update_mime_type( $updated_img, 'the_content', $updated_id ), $content );
+		if ( $attachment_id !== $updated_attachment_id ) {
+			$update_class_name = preg_replace( '/wp-image-([\d]+)/i', sprintf( 'wp-image-%d', $updated_attachment_id ), $img );
+			$content = str_replace( $img, webp_uploads_img_tag_update_mime_type( $update_class_name, 'the_content', $updated_attachment_id ), $content );
 		} else {
 			$content = str_replace( $img, webp_uploads_img_tag_update_mime_type( $img, 'the_content', $attachment_id ), $content );
 		}
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -405,19 +405,21 @@ function webp_uploads_update_image_references( $content ) {
 	foreach ( $images as $img => $attachment_id ) {
 
 		/**
-		 * Check for old uploaded images
+		 * Check for old uploaded images.
 		 */
-		if( ! preg_match( '~src*=*["\'](.*?)["\']~', $img, $image_src ) ) {
+		if ( ! preg_match( '~src*=*["\'](.*?)["\']~', $img, $image_src ) ) {
 			continue;
 		}
 
 		if ( empty( $image_src ) ) {
 			continue;
 		}
-		
-		// Get attachment id by image src
+
+		/**
+		 * Get attachment id by the attachment url.
+		 */
 		$updated_attachment_id = attachment_url_to_postid( $image_src[1] );
-		
+
 		if ( $attachment_id !== $updated_attachment_id ) {
 			$update_class_name = preg_replace( '/wp-image-([\d]+)/i', sprintf( 'wp-image-%d', $updated_attachment_id ), $img );
 			$content = str_replace( $img, webp_uploads_img_tag_update_mime_type( $update_class_name, 'the_content', $updated_attachment_id ), $content );
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -418,7 +418,7 @@ function webp_uploads_update_image_references( $content ) {
 		// Get attachment id by image src
 		$updated_id = attachment_url_to_postid( $image_src[1] );
 		
-		if ( $attachment_id != $updated_id ) {
+		if ( $attachment_id !== $updated_id ) {
 			$updated_img = preg_replace( '/wp-image-([\d]+)/i', 'wp-image-'.$updated_id, $img );
 			$content = str_replace( $img, webp_uploads_img_tag_update_mime_type( $updated_img, 'the_content', $updated_id ), $content );
 		} else {
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -419,7 +419,7 @@ function webp_uploads_update_image_references( $content ) {
 		$updated_id = attachment_url_to_postid( $image_src[1] );
 		
 		if ( $attachment_id !== $updated_id ) {
-			$updated_img = preg_replace( '/wp-image-([\d]+)/i', 'wp-image-'.$updated_id, $img );
+			$updated_img = preg_replace( '/wp-image-([\d]+)/i', sprintf( 'wp-image-%d', $updated_id ), $img );
 			$content = str_replace( $img, webp_uploads_img_tag_update_mime_type( $updated_img, 'the_content', $updated_id ), $content );
 		} else {
 			$content = str_replace( $img, webp_uploads_img_tag_update_mime_type( $img, 'the_content', $attachment_id ), $content );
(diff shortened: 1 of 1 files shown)


modules/images/webp-uploads/load.php
```diff
@@ -403,7 +403,27 @@ function webp_uploads_update_image_references( $content ) {
...
+		}
+
+		if ( empty( $image_src ) ) {
+			continue;
+		}
+		
+		// Get attachment id by image src
+		$updated_id = attachment_url_to_postid( $image_src[1] );
+		
+		if ( $attachment_id != $updated_id ) {
```
akshitsethi:
```suggestion
		if ( $attachment_id !== $updated_id ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -403,7 +403,27 @@ function webp_uploads_update_image_references( $content ) {
...
+			$updated_img = preg_replace( '/wp-image-([\d]+)/i', 'wp-image-'.$updated_id, $img );
```
akshitsethi:
```suggestion
			$updated_img = preg_replace( '/wp-image-([\d]+)/i', sprintf( 'wp-image-%d', $updated_id ), $img );
```

akshitsethi (changes requested):
@mukeshpanchal27 Thank you for the PR.

There are few linter issues that needs to go away. Also, we can simplify the below code by removing the `else` statement.
```
if ( $attachment_id != $updated_id ) {
    $updated_img = preg_replace( '/wp-image-([\d]+)/i', 'wp-image-'.$updated_id, $img );
    $content = str_replace( $img, webp_uploads_img_tag_update_mime_type( $updated_img, 'the_content', $updated_id ), $content );
} else {
    $content = str_replace( $img, webp_uploads_img_tag_update_mime_type( $img, 'the_content', $attachment_id ), $content );
}
```

modules/images/webp-uploads/load.php
akshitsethi: Also, we are updating `img` class here. Updating the variable name will make it easier to understand :)

modules/images/webp-uploads/load.php
mukeshpanchal27: Thanks for the review @akshitsethi 

I update the code for if condition. In the else condition we need two variables for the update so I have not removed it do you have any idea how we remove it?

modules/images/webp-uploads/load.php
```diff
@@ -403,7 +403,29 @@ function webp_uploads_update_image_references( $content ) {
...
+		 * Get attachment id by the attachment url.
+		 */
+		$updated_attachment_id = attachment_url_to_postid( $image_src[1] );
+
+		if ( $attachment_id !== $updated_attachment_id ) {
+			$update_class_name = preg_replace( '/wp-image-([\d]+)/i', sprintf( 'wp-image-%d', $updated_attachment_id ), $img );
+			$content           = str_replace( $img, webp_uploads_img_tag_update_mime_type( $update_class_name, 'the_content', $updated_attachment_id ), $content );
+		} else {
+			$content = str_replace( $img, webp_uploads_img_tag_update_mime_type( $img, 'the_content', $attachment_id ), $content );
+		}
```
akshitsethi, line 428:
```suggestion
		if ( $attachment_id !== $updated_attachment_id ) {
			// Updating class name in the <img> tag.
			$img           = preg_replace( '/wp-image-([\d]+)/i', sprintf( 'wp-image-%d', $attachment_id ), $img );
			$attachment_id = $updated_attachment_id;
		}

		$content = str_replace( $img, webp_uploads_img_tag_update_mime_type( $img, 'the_content', $attachment_id ), $content );
```

akshitsethi: @mukeshpanchal27 We can do something like this to get rid of the `else` statement.

akshitsethi (changes requested):
@mukeshpanchal27 Thank you for making the changes. I've left some additional feedback. 

Also, on testing this locally, only the first image tends to switch to `webp` format. Other images on the page stay the same. Can you please have a look and verify this at your end?

modules/images/webp-uploads/load.php
mukeshpanchal27: In my setup it changes all old images to webp please check the attached screenshot.

We can't use $img variable `$img = preg_replace( '/wp-image-([\d]+)/i', sprintf( 'wp-image-%d', $attachment_id ), $img );` here because it will not replace anything in below statement. We must use different variable so it get replace.

image

akshitsethi: @mukeshpanchal27 Thank you for the clarification. It happens sporadically for me. It works well for some posts and on one of the post, I see the `$updated_attachment_id` to be `0` for all images except for one. And because of this, the attachment metadata is empty which returns the image as it is.

Screenshot 2022-05-25 at 10 46 02 AM

Also, can you please tell me why we are updating `id` for the `img` class name? The reason I'm asking is that if we use an image block, the same class is used for the block wrapper `div` as well. So, is updating image class necessary?

mukeshpanchal27: @akshitsethi Can you please send me steps to reproduce that?

I replace the id because when a new image is imported into WordPress it gets a new ID for the attachment and it was not updated in the IMG tag class.

mukeshpanchal27: This code only replace the_content IMG tag only not other dynamic blocks.

akshitsethi: @mukeshpanchal27 The class added to the wrapper for the image block was done manually, so my comment above can be ignored. Still, below is the screenshot to explain what I meant. This is from the image block post of the Theme_Unit_Test data.

Screenshot 2022-05-25 at 1 33 54 PM


Overall, this looks good to me. Can we address the edge case like the one I'm experiencing where the `updated_attachment_id` is coming out to be 0. Something like below:

```
if ( ! $updated_attachment_id ) {
	$updated_attachment_id = $attachment_id;
}
```

What are your thoughts on this?

mukeshpanchal27: I don't think we have to check it because if the current attachment ID and the static content image are not the same ID then it will not affect anything.

`if ( $attachment_id !== $updated_attachment_id ) {`

akshitsethi: This is for the case which I'm experiencing. Please see the two screenshots below:

**On this branch**
Screenshot 2022-05-25 at 2 25 57 PM


**On trunk**
Screenshot 2022-05-25 at 2 26 17 PM

These are the same images but when on this branch it only updates the first image to `webp` and the rest of the other images on page are not updated. When I checked the value of `updated_attachment_id` in the loop, it only existed for the first image and rest others were `0`.

This is weird because the `webp` version does exist but still this needs to be investigated further as to why this is happending because on trunk it is working correctly. I'll take out some time today to check this out and update you.

modules/images/webp-uploads/load.php
```diff
@@ -403,7 +403,29 @@ function webp_uploads_update_image_references( $content ) {
...
+
+		if ( empty( $image_src ) ) {
+			continue;
+		}
+
+		/**
+		 * Get attachment id by the attachment url.
+		 */
+		$updated_attachment_id = attachment_url_to_postid( $image_src[1] );
```
adamsilverstein, line 421: this can be an expensive query to run on every page load

modules/images/webp-uploads/load.php
```diff
@@ -403,7 +403,29 @@ function webp_uploads_update_image_references( $content ) {
...
+
+		if ( $attachment_id !== $updated_attachment_id ) {
+			$update_class_name = preg_replace( '/wp-image-([\d]+)/i', sprintf( 'wp-image-%d', $updated_attachment_id ), $img );
+			$content           = str_replace( $img, webp_uploads_img_tag_update_mime_type( $update_class_name, 'the_content', $updated_attachment_id ), $content );
```
adamsilverstein, line 425: `updated_attachment_id` might be zero here if image is not found

modules/images/webp-uploads/load.php
felixarntz: +1, I think unfortunately calling that function here, potentially numerous times on page load, is not feasible from a performance perspective.

felixarntz (changes requested):
@mukeshpanchal27 While this implementation would work, I have several concerns with this PR. Most importantly, it adds a quite expensive uncached query that would happen for every image, on every page load - that's a no-go from a performance perspective.

In addition, it also adds 2 more regular expression checks, which is certainly not _as_ significant, but still a potential performance concern particularly because this runs on every content image on every page load.

Of course we can try to iterate on this PR, but frankly I am not sure that there is any straightforward solution that would not result in major performance problems. Let me know your thoughts - I think we may want to re-evaluate our options with the issue here as a whole.

modules/images/webp-uploads/load.php
mukeshpanchal27: Yes. Thanks for the review.

felixarntz:
Closing this per the conversation on #339 - fixing this here creates performance problems and is not really the right place. The problem comes from the WordPress importer not behaving correctly.

PR #343, opened 2022-05-31
Fix accepted_args for add_action

### LINKED ISSUE #342: `accepted_args` parameter issue in `add_action`
## Bug Description

In the admin_menu action we use `add_action( "load-{$hook_suffix}", 'perflab_load_modules_page', 10, 0 );` action which tell user that it's the number of arguments this function accepts and we use 0 so this function will not take any arguments. But on function we use two arguments https://github.com/WordPress/performance/blob/trunk/admin/load.php#L49, https://github.com/WordPress/performance/blob/trunk/tests/admin/load-tests.php#L100 and https://github.com/WordPress/performance/blob/trunk/tests/admin/load-tests.php#L109

It will miss guide users so if I'm not wrong do we need to add 2 in the accepted_args(4th parameter) of the action?

Additionally, we have to add a return doc for the `perflab_add_modules_page` https://github.com/WordPress/performance/blob/trunk/admin/load.php#L13-L35 something like the below context.

`* @return string|false The resulting page's hook_suffix, or false if the user does not have the capability required.`

mukeshpanchal27:
## Summary

Fixes #342

## Checklist

### DIFF
--- a/admin/load.php
+++ b/admin/load.php
@@ -26,7 +26,7 @@ function perflab_add_modules_page() {
 
 	// Add the following hooks only if the screen was successfully added.
 	if ( false !== $hook_suffix ) {
-		add_action( "load-{$hook_suffix}", 'perflab_load_modules_page', 10, 0 );
+		add_action( "load-{$hook_suffix}", 'perflab_load_modules_page', 10, 2 );
 		add_action( 'plugin_action_links_' . plugin_basename( PERFLAB_MAIN_FILE ), 'perflab_plugin_action_links_add_settings' );
 	}
 
--- a/admin/load.php
+++ b/admin/load.php
@@ -9,6 +9,8 @@
  * Adds the modules page to the Settings menu.
  *
  * @since 1.0.0
+ *
+ * @return string|false The resulting page's hook_suffix, or false if the user does not have the capability required.
  */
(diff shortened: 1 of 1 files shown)


admin/load.php
```diff
@@ -26,7 +28,7 @@ function perflab_add_modules_page() {
 
 	// Add the following hooks only if the screen was successfully added.
 	if ( false !== $hook_suffix ) {
-		add_action( "load-{$hook_suffix}", 'perflab_load_modules_page', 10, 0 );
+		add_action( "load-{$hook_suffix}", 'perflab_load_modules_page', 10, 2 );
```
felixarntz, line 31: This isn't correct, it needs to remain at `0`. The intention here is exactly that the action hook does _not_ pass any parameters. The parameters of that function are intended for other calls to that function, and if the action passes its default parameters to the function, it would break.

felixarntz (commented):
@mukeshpanchal27 Thanks for this PR, but I think this is working as expected. I think we can close this one.

mxbclang:
Closing per @felixarntz's comment above.

admin/load.php
mukeshpanchal27: Got it. Thanks for the information.

PR #345, opened 2022-06-02
Introduce plugin uninstaller routine

merkys7:
## Summary

Cleanup `wp_options` table after plugin uninstall.

Currently, if we uninstall `performance-lab` plugin - it leaves `wp_options` entries with AUTOLOAD = yes values.
download (6)

## Relevant technical choices

Introduce Uninstall method referenced at https://developer.wordpress.org/plugins/plugin-basics/uninstall-methods/

## Checklist

### DIFF
--- a/uninstall.php
+++ b/uninstall.php
@@ -1,0 +1,14 @@
+<?php
+/**
+ * Plugin uninstaller logic.
+ *
+ * @package performance-lab
+ * @since 1.2.0
+ */
+
+// If uninstall.php is not called by WordPress, bail.
+if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
+	exit;
+}
+
+delete_option( 'perflab_modules_settings' );


uninstall.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * performance-lab uninstall method
```
felixarntz: Let's add the `@package` here like in the other files.
```suggestion
 * Plugin ninstaller logic.
 *
 * @package performance-lab
 * @since n.e.x.t
```

uninstall.php
```diff
@@ -0,0 +1,11 @@
...
+ */
+
+// if uninstall.php is not called by WordPress, die
+if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
+	die;
```
felixarntz: Super-nitpick, but I think most plugins use `exit` :)
```suggestion
	exit;
```

uninstall.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * performance-lab uninstall method
+ */
+
+// if uninstall.php is not called by WordPress, die
```
felixarntz: Comments should start with a capital letter and end in a full stop.
```suggestion
// If uninstall.php is not called by WordPress, bail.
```

felixarntz (commented):
@merkys7 Thanks for the PR! Looks good to me, just a few tiny code suggestions for consistency.

uninstall.php
merkys7: Resolved, thank you

uninstall.php
merkys7: Resolved, thank you

uninstall.php
merkys7: Resolved, thank you

uninstall.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
+ * Plugin uninstaller logic.
+ *
+ * @package performance-lab
+ * @since 1.2.0
```
felixarntz, line 6: This should technically be `n.e.x.t` per convention, but since the next release is going to be `1.2.0`, it's probably fine this time.

felixarntz (approved):
Great, thanks!

uninstall.php
merkys7: Thank you

mukeshpanchal27:
Do we need to remove options from the Multisite using `delete_site_option( 'perflab_modules_settings' );` function as WordPress Beta Tester plugin does [here](https://github.com/afragen/wordpress-beta-tester/blob/develop/uninstall.php#L31)

merkys7:
@mukeshpanchal27 I think there is no such need as multi-site options are not being created inside this plugin because `update_site_option` function is not being called (only `update_option`)

mukeshpanchal27:
> @mukeshpanchal27 I think there is no such need as multi-site options are not being created inside this plugin because `update_site_option` function is not being called (only `update_option`)

Thanks for the reply.

If we setup multisite then it will save the option in their sub-site DB. Like if you have two sub-site then the plugin option will save the option in the below tables.

**wp_options** - Main site option table
**wp_2_options** - 1st Sub-site option table
**wp_3_options** - 2nd Sub-site option table

I check the `update_option` in multisite and it doesn't delete the sub-sites options in all wp_SITE-ID_options tables. in my further investigation of multisite, `delete_site_option` also does not remove the sub-site options.

We should check in multisite for deleting the options from the multisite something like the [Google Web Stories](https://github.com/GoogleForCreators/web-stories-wp/blob/main/uninstall.php#L46-L65) plugin did for uninstall method.

I have prepared a code snippet that removes plugin options from the multisite.

     if ( is_multisite() ) {
            $site_ids = get_sites(
                [
                    'fields'                 => 'ids',
                    'number'                 => '',
                    'update_site_cache'      => false,
                    'update_site_meta_cache' => false,
                ]
            );

            foreach ( $site_ids as $site_id ) {
                switch_to_blog( $site_id );
                delete_option( 'perflab_modules_settings' );	
            }
            restore_current_blog();
     } else {
            delete_option( 'perflab_modules_settings' );
     }

Additional enhancement for introducing a filter for whether data should be erased when uninstalling the plugin so if anyone doesn't want to delete their plugin setting then use this filter.

    /**
    * Filters whether data should be erased when uninstalling the plugin.
    *
    * @since n.e.x.t
    *
    * @param bool $erase Whether to erase data. Default false.
    */
    $erase = (bool) apply_filters( 'perflab_erase_data_on_uninstall', false );

    if ( false === $erase ) {
        return;
    }

mitogh (approved):
LGTM 👍

felixarntz:
Thanks @mukeshpanchal27 for the additional feedback.

> We should check in multisite for deleting the options from the multisite something like the [Google Web Stories](https://github.com/GoogleForCreators/web-stories-wp/blob/main/uninstall.php#L46-L65) plugin did for uninstall method.
> 
> I have prepared a code snippet that removes plugin options from the multisite.
> 
> ```
>  if ( is_multisite() ) {
>         $site_ids = get_sites(
>             [
>                 'fields'                 => 'ids',
>                 'number'                 => '',
>                 'update_site_cache'      => false,
>                 'update_site_meta_cache' => false,
>             ]
>         );
> 
>         foreach ( $site_ids as $site_id ) {
>             switch_to_blog( $site_id );
>             delete_option( 'perflab_modules_settings' );	
>         }
>         restore_current_blog();
>  } else {
>         delete_option( 'perflab_modules_settings' );
>  }
> ```

That's a fair point. The complexity here with multisite is that there could be a ton of sites on certain environments, and then your code wouldn't be scalable. This is a known problem with WordPress core.

What we could do is set a sensible limit, e.g. set `number` to `50` or something. Then it would least support the majority of multisites that don't have a ton of sites.

Because of that extra complexity though, I would prefer if we continued thinking and iterating on that in a separate follow-up issue. Would you mind opening one?

> Additional enhancement for introducing a filter for whether data should be erased when uninstalling the plugin so if anyone doesn't want to delete their plugin setting then use this filter.
> 
> ```
> /**
> * Filters whether data should be erased when uninstalling the plugin.
> *
> * @since n.e.x.t
> *
> * @param bool $erase Whether to erase data. Default false.
> */
> $erase = (bool) apply_filters( 'perflab_erase_data_on_uninstall', false );
> 
> if ( false === $erase ) {
>     return;
> }
> ```

Normally I would agree with this, but I think for this plugin it's not really necessary since the options are so basic and easy to reconfigure. If we at some point want to allow control on _not_ deleting the data on uninstall, I think we should rather go a separate approach and include a checkbox somewhere so that the site administrator can control that without writing code.

felixarntz:
Thanks again @merkys7 for the PR! Would you be interested in working on a follow-up enhancement to also support multisite in the uninstaller? See especially the above comments https://github.com/WordPress/performance/pull/345#issuecomment-1154705327 and https://github.com/WordPress/performance/pull/345#issuecomment-1155608644

merkys7:
Thank you so much! Yes, I will be trying to improve it based on the mentioned comments

PR #359, opened 2022-06-10
Avoid overwriting existing WebP files when creating WebP images

### LINKED ISSUE #358: WebP file overwritten if filename has the same name with extension of jpeg and jpg
## Bug Description

This may be a misconfiguration issue, but at the moment if a file is uploaded with the same filename but the extension is jpeg or jpg, the corresponding webp file generated is overwritten by the last file uploaded.

## Steps to reproduce

1. Create 2 files, for example, `Paul-Test.jpeg` and `Paul-Test.jpg`
2. Go to Media Gallery
3. Click on Add New
4. Upload `Paul-Test.jpeg`
5. Check `~/wp-content/uploads/2022/06/Paul-Test.jpeg`
6. Check `~/wp-content/uploads/2022/06/Paul-Test.webp`
7. Upload `Paul-Test.jpg`
8. Check `~/wp-content/uploads/2022/06/Paul-Test.jpg`
9. Check `~/wp-content/uploads/2022/06/Paul-Test.webp`
10. The file `Paul-Test.webp` created from `Paul-Test.jpeg` is overwritten with the webp generated from `Paul-Test.jpg`

## Additional Context

I have added this to functions.php:
```
/* Image conversion attributes */

    return [
        'image/jpeg' =>  [ 'image/jpeg', 'image/webp' ],
        'image/png' =>  [ 'image/png', 'image/webp' ],
        'image/webp' =>  [ 'image/webp', 'image/jpeg' ],
    ];
});
```

 - PHP Version: 7.4
 - OS: RHEL 8.6
 - Browser: Firefox
 - Plugin Version: 1.1.0

To note, PNGs with the same filename do not overwrite the webp image (an image is not created at all).

mukeshpanchal27:
## Summary

This PR skips the creation of duplicate WebP images if anyone uses an image file with the same name with an extension of jpg, jpe, or jpeg.

Please check more detail about the issue and the options for the solution [here](https://github.com/WordPress/performance/issues/358#issuecomment-1150695492)

As @felixarntz suggested we should go with 1st approach.

If you want to generate a WebP image for `.png` images then please add the below code snippet in your current themes functions.php

        return [
            'image/jpeg' =>  [ 'image/jpeg', 'image/webp' ],
            'image/png' =>  [ 'image/png', 'image/webp' ],
            'image/webp' =>  [ 'image/webp', 'image/jpeg' ],
        ];
    });

Fixes #358

## Images for testing

If you want to test the different images then please download the zip in which I added all image formats( jpg, jpe, jpeg and png ) - [image.zip](https://github.com/WordPress/performance/files/8876261/image.zip)

## Checklist

### DIFF
--- a/modules/images/webp-uploads/helper.php
+++ b/modules/images/webp-uploads/helper.php
@@ -142,6 +142,11 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 		$destination_file_name = $editor->generate_filename( null, null, $extension[0] );
 	}
 
+	// Skip creation of duplicate WebP image if an image file already exists in the directory.
+	if ( file_exists( $destination_file_name ) ) {
+		return new WP_Error( 'webp_image_file_present', __( 'The webP image already exists.', 'performance-lab' ) );
+	}
+
 	$image = $editor->save( $destination_file_name, $mime );
 	if ( is_wp_error( $image ) ) {
 		return $image;
--- a/modules/images/webp-uploads/helper.php
+++ b/modules/images/webp-uploads/helper.php
@@ -144,7 +144,7 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 
 	// Skip creation of duplicate WebP image if an image file already exists in the directory.
 	if ( file_exists( $destination_file_name ) ) {
-		return new WP_Error( 'webp_image_file_present', __( 'The webP image already exists.', 'performance-lab' ) );
+		return new WP_Error( 'webp_image_file_present', __( 'The WebP image already exists.', 'performance-lab' ) );
 	}
 
 	$image = $editor->save( $destination_file_name, $mime );
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -90,6 +90,12 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 
 		$extension   = explode( '|', $allowed_mimes[ $targeted_mime ] );
 		$destination = trailingslashit( $original_directory ) . "{$filename}.{$extension[0]}";
+		
+		// Skip creation of duplicate WebP image if a image file has the same name with an extension of jpe, jpg, or jpeg.
+		if ( file_exists( $destination ) ) {
+			continue;
+		}
+
 		$image       = webp_uploads_generate_additional_image_source( $attachment_id, 'full', $original_size_data, $targeted_mime, $destination );
 
 		if ( is_wp_error( $image ) ) {
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -90,13 +90,13 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 
 		$extension   = explode( '|', $allowed_mimes[ $targeted_mime ] );
 		$destination = trailingslashit( $original_directory ) . "{$filename}.{$extension[0]}";
-		
+
 		// Skip creation of duplicate WebP image if a image file has the same name with an extension of jpe, jpg, or jpeg.
 		if ( file_exists( $destination ) ) {
 			continue;
 		}
 
-		$image       = webp_uploads_generate_additional_image_source( $attachment_id, 'full', $original_size_data, $targeted_mime, $destination );
+		$image = webp_uploads_generate_additional_image_source( $attachment_id, 'full', $original_size_data, $targeted_mime, $destination );
 
 		if ( is_wp_error( $image ) ) {
 			continue;
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -90,13 +90,7 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 
 		$extension   = explode( '|', $allowed_mimes[ $targeted_mime ] );
 		$destination = trailingslashit( $original_directory ) . "{$filename}.{$extension[0]}";
-
-		// Skip creation of duplicate WebP image if a image file has the same name with an extension of jpe, jpg, or jpeg.
-		if ( file_exists( $destination ) ) {
-			continue;
-		}
-
-		$image = webp_uploads_generate_additional_image_source( $attachment_id, 'full', $original_size_data, $targeted_mime, $destination );
+		$image       = webp_uploads_generate_additional_image_source( $attachment_id, 'full', $original_size_data, $targeted_mime, $destination );
 
 		if ( is_wp_error( $image ) ) {
 			continue;


jjgrainger (approved):
Looks goods

modules/images/webp-uploads/load.php
```diff
@@ -90,7 +90,13 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 
 		$extension   = explode( '|', $allowed_mimes[ $targeted_mime ] );
 		$destination = trailingslashit( $original_directory ) . "{$filename}.{$extension[0]}";
-		$image       = webp_uploads_generate_additional_image_source( $attachment_id, 'full', $original_size_data, $targeted_mime, $destination );
+
+		// Skip creation of duplicate WebP image if a image file has the same name with an extension of jpe, jpg, or jpeg.
+		if ( file_exists( $destination ) ) {
+			continue;
```
mitogh: Do we want to try to create a different destination name instead? Like appending a number or a UUID instead, in the same way, WordPress handles it when the file already exists?

mitogh (commented):
Just one question, besides that things look good on my end.

I would encourage to add a test case replicating the current issue BTW.

modules/images/webp-uploads/load.php
felixarntz: @mitogh Yes, this is the eventual plan, but for now adding this stop-gap to not overwrite/confuse images should be sufficient.

Let's work on a proper solution where both images can be kept in a separate PR.

modules/images/webp-uploads/load.php
```diff
@@ -90,7 +90,13 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+		}
```
felixarntz: Instead of adding this check here, I think a cleaner solution would be to handle it within the `webp_uploads_generate_additional_image_source()` function which is the function where the file is actually generated. That would fix the problem at its root and it would ensure you don't have to add a condition around it everywhere you call that function.

The current implementation here is actually missing the scenario when a new image size is generated where that function is also called - right now there would still be a chance for the image to be overwritten. By adding the condition to `webp_uploads_generate_additional_image_source()` and returning a `WP_Error` in that case, we only need to make that change there and it will work everywhere without any further changes.

felixarntz (changes requested):
@mukeshpanchal27 This PR only fixes one half of the bug - for when the full size is generated, but not for when the other sizes are generated. See my comment below, I think this should rather be fixed at the root of the problem, in `webp_uploads_generate_additional_image_source()`. Otherwise, it's easy to miss certain situations, one of them already missing here.

Could you then also add a test for the new condition in `webp_uploads_generate_additional_image_source()`, to cover the new clause that an existing image file would cause a `WP_Error`?

mukeshpanchal27:
If we consider checking the `file_exists` in function `webp_uploads_generate_additional_image_source()` then we can check it after the editor `generate_filename` function because we will get all additional thumbnail size URLs after this code snippet.

	if ( null === $destination_file_name ) {
		$extension             = explode( '|', $allowed_mimes[ $mime ] );
		$destination_file_name = $editor->generate_filename( null, null, $extension[0] );
	}

As I troubleshoot the issue, we have two options here.

1. We can check `file_exists` after the code snippet so we get full and other thumbnail images.

		if ( null === $destination_file_name ) {
			$extension             = explode( '|', $allowed_mimes[ $mime ] );
			$destination_file_name = $editor->generate_filename( null, null, $extension[0] );
		}

		if ( file_exists( $destination_file_name ) ) {
			return new WP_Error( 'webp_image_file_exist', __( 'The webP image already exists.', 'performance-lab' ) );
		}

In this approach there may be some unnecessary checking happens for the duplicate image in function `webp_uploads_generate_additional_image_source()`

2. We pass `$destination_file_name` in function `webp_uploads_generate_image_size()` for consistancy for thumbnail images so we can check `file_exists` at top in function `webp_uploads_generate_additional_image_source()`

		$filename    = pathinfo( $properties['file'], PATHINFO_FILENAME );
		$extension   = explode( '|', $allowed_mimes[ $targeted_mime ] );
		$destination = trailingslashit( $original_directory ) . "{$filename}.{$extension[0]}";
		$source = webp_uploads_generate_image_size( $attachment_id, $size_name, $mime, $destination );

If we consider this option then we don't need to call the `generate_filename` function to get the image name.

	if ( null === $destination_file_name ) {
		$extension             = explode( '|', $allowed_mimes[ $mime ] );
		$destination_file_name = $editor->generate_filename( null, null, $extension[0] );
	}

In this approach we have to update unit test, we use `webp_uploads_generate_image_size()` function in unit test.

If I miss anything then please let me know.

felixarntz:
@mukeshpanchal27 I think we should go with your approach 1, always run the `file_exists` check right after we have the `$destination_file_name` set for sure. That way we only need to have the check in one place.

> In this approach there may be some unnecessary checking happens for the duplicate image in function `webp_uploads_generate_additional_image_source()`

Can you clarify what you mean? I think there would only be a single `file_exists` check in each different situation, only in `webp_uploads_generate_additional_image_source()`.

mukeshpanchal27:
> Can you clarify what you mean?

For duplicate image, `webp_uploads_generate_additional_image_source()` function check this https://github.com/mukeshpanchal27/performance/blob/fix/358-skip-duplicate-webp-image/modules/images/webp-uploads/helper.php#L85-L143 code that may be not require for already exist images.

PR updated with approach 1 and requested the review.

modules/images/webp-uploads/helper.php
```diff
@@ -142,6 +142,11 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 		$destination_file_name = $editor->generate_filename( null, null, $extension[0] );
 	}
 
+	// Skip creation of duplicate WebP image if an image file already exists in the directory.
+	if ( file_exists( $destination_file_name ) ) {
+		return new WP_Error( 'webp_image_file_present', __( 'The webP image already exists.', 'performance-lab' ) );
```
felixarntz:
```suggestion
		return new WP_Error( 'webp_image_file_present', __( 'The WebP image already exists.', 'performance-lab' ) );
```

felixarntz (approved):
@mukeshpanchal27 
> For duplicate image, `webp_uploads_generate_additional_image_source()` function check this https://github.com/mukeshpanchal27/performance/blob/fix/358-skip-duplicate-webp-image/modules/images/webp-uploads/helper.php#L85-L143 code that may be not require for already exist images.

Right, I agree it's a bit wasteful that we have to run all that logic before the `file_exists` check, but unless `$destination_file_name` is already provided, some of that logic is critical to run before since it affects the file name (for example the `resize()` call).

I think the current implementation is good for now since it solves the problem. In WordPress core this will eventually be implemented differently anyway. I left one more tiny comment, but already approving this since it's good to go.

modules/images/webp-uploads/helper.php
mukesh-10up: @felixarntz thanks for the feedback. I have updated the text.

mukeshpanchal27: @felixarntz thanks for the feedback. I have updated the text.

akshitsethi (approved):

PR #360, opened 2022-06-10
Provide fallback JPEG images in frontend when WebP is not supported by the browser

### LINKED ISSUE #341: Provide polyfill to serve JPEG or PNG if a page contains WebP images
## Feature Description

In order to provide support for webp on older browser we need to load webp-hero polyfill on pages that have webp images. The polyfill needs to be loaded only if webp isn't supported by the browser.

https://github.com/chase-moskal/webp-hero

eugene-manuilov:
## Summary

Fixes #341

## Checklist


mukeshpanchal27:
@eugene-manuilov Is it possible that we can check the dynamic array key value based on `webp_uploads_get_upload_image_mime_transforms()` function data because that function use one filter `webp_uploads_upload_image_mime_transforms` to alter that mime type transform and and edge case user change it then it create fatal error in checking.

For filter edge cases please check #361 in which I added some cases for the filter.

eugene-manuilov:
@mukeshpanchal27 i have added a check to make sure that the mime types transforms array is an array.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,106 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+ */
+function webp_uploads_wepb_fallback() {
+	// Get mime type transofrms for the site.
+	$transforms = webp_uploads_get_upload_image_mime_transforms();
+	if ( ! is_array( $transforms ) ) {
+		return;
+	}
+
+	// We need to add fallback only if jpeg alternatives for the webp images are enabled for the server.
+	$jpegs = in_array( 'image/jpeg', $transforms['image/jpeg'], true ) && in_array( 'image/webp', $transforms['image/jpeg'], true );
```
mitogh: Personal preference but I would rename those variables to something like:

```php
$supports_jpeg
$supports_jpeg_transform
```

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,106 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+	$webp  = in_array( 'image/jpeg', $transforms['image/webp'], true );
+	if ( ! $jpegs && ! $webp ) {
+		return;
+	}
+
+	?>
+	<script>
```
```suggestion
	<script type="application/javascript">
```

mitogh: Any reason why not to enqueue this one via a JS file instead?

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,106 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+		( function() {
+			window.fallbackWebpImages = function( media ) {
```
mitogh: Probably we want to use 
```suggestion
			window._fallbackWebpImages = function( media ) {
```

To avoid any possible conflict, another thing is  to make sure the object has not been defined already such as:

```suggestion
	if ( '_ fallbackWebpImages' in window && typeof window['fallbackWebpImages'] === 'function') { 
	return;
	}
```

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,106 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+
+						var images = document.querySelectorAll( 'img.wp-image-' + media[i].id );
+						for ( var j = 0; j < images.length; j++ ) {
+							images[j].src = images[j].src.replace( /\.webp$/i, ext[0] );
+							var srcset = images[j].getAttribute( 'srcset' );
+							if ( srcset ) {
+								images[j].setAttribute( 'srcset', srcset.replace( /\.webp(\s)/ig, ext[0] + '$1' ) );
+							}
+						}
+					} catch ( e ) {
```
mitogh: Maybe worth sending this to the browser.

```suggestion
					} catch ( e ) {
					console.error(e);
```

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,106 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+		return;
+	}
+
+	?>
+	<script>
+		( function() {
+			window.fallbackWebpImages = function( media ) {
+				for ( var i = 0; i < media.length; i++ ) {
+					try {
+						var ext = media[i].media_details.sources['image/jpeg'].file.match( /\.\w+$/i );
```
mitogh: Do we want to verify that every media item has a `sources` array and each of those has the `file` property on it ?

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,106 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+
+			var loadMediaDetails = function( nodes ) {
+				var ids = [];
+				for ( var i = 0; i < nodes.length; i++ ) {
+					if ( nodes[i].nodeName !== "IMG" || ! nodes[i].src.match( /\.webp$/i ) ) {
+						continue;
+					}
+
+					var attachment = nodes[i].className.match( /wp-image-(\d+)/i );
+					if ( attachment && attachment[1] && ids.indexOf( attachment[1] ) === -1 ) {
```
mitogh: Small suggestion would be to replace the array with an object structure in that sense you don't need to check if the item exists as you can set it regardless like:

```js
ids[ attachment[1] ] = true;
```

Then you can just return the keys from it, if you only need the values: 

```js
return Object.keys( ids );
```

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,106 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+						ids.push( attachment[1] );
+					}
+				}
+
+				if ( ! ids.length ) {
```
mitogh: Not required since if the length is zero, the following loop would set `pages = 0` which will prevent the for loop to be executed at all.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,106 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+					return;
+				}
+
+				for ( var page = 0, pages = Math.ceil( ids.length / 100 ); page < pages; page++ ) {
+					var pageIds = [];
+					for ( var i = 0; i < 100 && i + page * 100 < ids.length; i++ ) {
```
mitogh: Not sure what's the reason behind the (i+page * 100), and the nested loop. Just curious about the reason.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,106 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+				loadMediaDetails( document.querySelectorAll( 'img' ) );
+
+				// Start the mutation observer to update images added dynamically.
+				var observer = new MutationObserver( function( mutationList ) {
+					for ( var i = 0; i < mutationList.length; i++ ) {
+						loadMediaDetails( mutationList[i].addedNodes );
+					}
+				} );
+
+				observer.observe( document.documentElement, {
```
mitogh: I think we are only interested in the body right ?

```suggestion
				observer.observe( document.body, {
```

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,106 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+					subtree: true,
+					childList: true,
+				} );
+			};
+		} )();
+	</script>
+	<?php
+}
+add_action( 'wp_head', 'webp_uploads_wepb_fallback' );
```
mitogh: Have we considered running this in the footer instead?

modules/images/webp-uploads/load.php
eugene-manuilov: 👍 renamed.

modules/images/webp-uploads/load.php
eugene-manuilov: Yep, good catch! Removed.

modules/images/webp-uploads/load.php
eugene-manuilov: 100 is the max value for `per_page`.

modules/images/webp-uploads/load.php
eugene-manuilov: Yep, updated.

modules/images/webp-uploads/load.php

modules/images/webp-uploads/load.php
felixarntz: +1, instead of a large inline script like this, we should reduce the inline JavaScript added on all pages (even those without need for the shim) to a minimum. Most of this script should be within its own JS file.

Btw, when creating the JS file, feel free to implement it without any build process. Adding a build process now, especially for a little script file like that, would go beyond the scope of this issue.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,102 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+
+	// We need to add fallback only if jpeg alternatives for the webp images are enabled for the server.
+	$preserve_jpegs_for_jpeg_transforms = in_array( 'image/jpeg', $transforms['image/jpeg'], true ) && in_array( 'image/webp', $transforms['image/jpeg'], true );
+	$preserve_jpegs_for_webp_transforms = in_array( 'image/jpeg', $transforms['image/webp'], true );
+	if ( ! $preserve_jpegs_for_jpeg_transforms && ! $preserve_jpegs_for_webp_transforms ) {
+		return;
+	}
+
+	?>
+	<script>
```
felixarntz: Instead of printing a `<script>` tag, we should be using [`wp_print_inline_script_tag()`](https://developer.wordpress.org/reference/functions/wp_print_inline_script_tag/).

Together with making the majority of the script a separate file, the JS code in here should be minimal enough to "easily" inline it using that function.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,102 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+			var img = document.createElement( 'img' );
+
+			// Verify two webp images.
+			img.src = "data:image/webp;base64,UklGRjIAAABXRUJQVlA4ICYAAACyAgCdASoCAAEALmk0mk0iIiIiIgBoSygABc6zbAAA/v56QAAAAA==";			
+			img.onload = function() {
+				img.src = "data:image/webp;base64,UklGRh4AAABXRUJQVlA4TBEAAAAvAQAAAAfQ//73v/+BiOh/AAA=";
+			};
+
+			// Error handler will be executed if the browser doesn't support webp.
+			img.onerror = function() {
```
felixarntz: Following up on the other comment thread above started by @mitogh, I think for the inline script itself we should only keep this code, starting with the `var img` assignment.

Within the `onerror` function here, we could then inject a script tag with the rest of the fallback approach so that it isn't loaded/parsed by browsers that don't need it.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,102 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+					var pageIds = [];
+					for ( var i = 0; i < 100 && i + page * 100 < ids.length; i++ ) {
+						pageIds.push( ids[ i + page * 100 ] );
+					}
+
+					var jsonp = document.createElement( 'script' );
+					jsonp.src = '<?php echo esc_js( get_rest_url() ); ?>wp/v2/media/?_fields=id,media_details&_jsonp=_fallbackWebpImages&per_page=100&include=' + pageIds.join( ',' );
+					document.body.appendChild( jsonp );
+				}
+			};
```
felixarntz: This is currently implementing approach 3 from https://github.com/WordPress/performance/issues/341#issuecomment-1143842022, which requires potentially a bunch of extra REST requests.

Given that all WebP images should always have JPEG available unless disabled on the entire setup (in which case this script isn't loaded anyway), I think the approach 2 from that implementation may be a more performant approach - just change file extensions from `.webp` to `.jpeg`. Otherwise, the REST requests may add a considerable amount of loading time, which would also result in the `img` tags not updated as quickly, thus causing the WebP file requests to not be cancelled.

The approach here is certainly a more reliable one - but it comes at the cost of performance I think, so we may still want to think about which of the two to go with further.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,102 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+function webp_uploads_wepb_fallback() {
+	// Get mime type transofrms for the site.
+	$transforms = webp_uploads_get_upload_image_mime_transforms();
+	if ( ! is_array( $transforms ) ) {
+		return;
+	}
+
+	// We need to add fallback only if jpeg alternatives for the webp images are enabled for the server.
+	$preserve_jpegs_for_jpeg_transforms = in_array( 'image/jpeg', $transforms['image/jpeg'], true ) && in_array( 'image/webp', $transforms['image/jpeg'], true );
+	$preserve_jpegs_for_webp_transforms = in_array( 'image/jpeg', $transforms['image/webp'], true );
```
felixarntz: We should probably also check whether `image/webp` is in the transforms list here, similar to the line above. Otherwise, all WebP uploads will exclusively rely on JPEG anyway, so there would be no need for the fallback script to load.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,102 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+	if ( ! $preserve_jpegs_for_jpeg_transforms && ! $preserve_jpegs_for_webp_transforms ) {
```
felixarntz, line 666: I think this is okay as long as we go with approach 3 - but if we decided to go with approach 2, I would make this more strict and say that we should only use the script if _both_ transforms lists include both formats. If any of the two does not include JPEG, we can assume the site owner doesn't care much about backward compatibility.

In that case it should be:
```php
	if ( ! $preserve_jpegs_for_jpeg_transforms || ! $preserve_jpegs_for_webp_transforms ) {
```

felixarntz (commented):
@eugene-manuilov Great work so far!

I still think we should think through the differences in approach 2 vs 3 a bit further. You went with the latter which is more reliable, but it also comes with more negative performance implications, e.g. the REST requests, and additional `document.querySelectorAll` calls - with approach 2 we could just once iterate through `img` tags and update all their `src` and `srcset` files.

With that said, indeed approach 3 is more reliable, and maybe that is the better option here. Curious what you think.

modules/images/webp-uploads/load.php
felixarntz: Let's prefix this global function with something specific to this module, e.g. `webpUploadsFallbackWebpImages`.

mukeshpanchal27:
> @mukeshpanchal27 i have added a check to make sure that the mime types transforms array is an array.

Thanks for the update @eugene-manuilov

`$transforms = webp_uploads_get_upload_image_mime_transforms();` always return array value as per filter code snippets.

    /**
	 * Filter to allow the definition of a custom mime types, in which a defined mime type
	 * can be transformed and provide a wide range of mime types.
	 *
	 * The order of supported mime types matters. If the original mime type of the uploaded image
	 * is not needed, then the first mime type in the list supported by the image editor will be
	 * selected for the default subsizes.
	 *
	 * @since 1.0.0
	 *
	 * @param array $default_transforms A map with the valid mime transforms.
	 */
	$transforms = (array) apply_filters( 'webp_uploads_upload_image_mime_transforms', $default_transforms );

It's worth checking the mime types transforms array is an array.

For the below edge case of `webp_uploads_upload_image_mime_transforms` filter it creates an error in current PR.

1. If the filter returns an empty value.

			return;
		});

2. If the filter returns PNG mime transforms in the array.

			return [
				'image/png' =>  [ 'image/png', 'image/webp' ]
			];
		});

To tackle this we should check `isset` value in the array respectively. Will add a review on the PR soon.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,102 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+ * Adds a fallback mechanism to replace webp images with jpeg alternatives on older browsers.
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_wepb_fallback() {
+	// Get mime type transofrms for the site.
+	$transforms = webp_uploads_get_upload_image_mime_transforms();
+	if ( ! is_array( $transforms ) ) {
+		return;
+	}
```
mukeshpanchal27: Please remove this checking as the function always return array values.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,102 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+
+	// We need to add fallback only if jpeg alternatives for the webp images are enabled for the server.
+	$preserve_jpegs_for_jpeg_transforms = in_array( 'image/jpeg', $transforms['image/jpeg'], true ) && in_array( 'image/webp', $transforms['image/jpeg'], true );
```
mukeshpanchal27: We should check isset of the array key to check if it exists or not.

    $preserve_jpegs_for_jpeg_transforms = isset( $transforms['image/jpeg'] ) && in_array( 'image/jpeg', $transforms['image/jpeg'], true ) && in_array( 'image/webp', $transforms['image/jpeg'], true );
	$preserve_jpegs_for_webp_transforms = isset( $transforms['image/webp'] ) && in_array( 'image/jpeg', $transforms['image/webp'], true );

modules/images/webp-uploads/load.php
eugene-manuilov: Sounds good. Updated.

modules/images/webp-uploads/load.php

modules/images/webp-uploads/load.php
eugene-manuilov: 👍 Updated.

modules/images/webp-uploads/load.php
eugene-manuilov: Yep, updated.

modules/images/webp-uploads/load.php
eugene-manuilov: Sounds good. Updated.

modules/images/webp-uploads/load.php
eugene-manuilov: Removed.

modules/images/webp-uploads/load.php
eugene-manuilov: Added.

modules/images/webp-uploads/fallback.js
```diff
@@ -0,0 +1,67 @@
...
+					if ( srcset ) {
+						images[j].setAttribute( 'srcset', srcset.replace( /\.webp(\s)/ig, ext[0] + '$1' ) );
+					}
+				}
+			} catch ( e ) {
+			}
+		}
+	};
+
+	var restApi = d.getElementById( 'webpUploadsFallbackWebpImages' ).getAttribute( 'data-rest-api' );
```
mitogh: Just curios on the reason why passing `document` as a variable and not `window` as well.

mitogh (commented):
Small question besides that looking good.

modules/images/webp-uploads/fallback.js
eugene-manuilov: `window` is used only once in the `fallback` script and passing it via arguments won't help to make the script code smaller. Instead, it will make the script code bigger because it will add additional variable declaration (in the function params) and variable call (to define a new function).

modules/images/webp-uploads/load.php
eugene-manuilov: No need to verify that because if the property is missing this will throw an exception that will be caught by the try/catch block and the image will be ignored.

modules/images/webp-uploads/load.php
eugene-manuilov: I don't think we really need it.

modules/images/webp-uploads/load.php
eugene-manuilov: > Given that all WebP images should always have JPEG available unless disabled on the entire setup (...), I think the approach 2 from that implementation may be a more performant approach - just change file extensions from .webp to .jpeg. 

@felixarntz, I went with the 2nd approach at the beginning but then it turned out that it wouldn't work well because JPEG images can have either `jpg` or `jpeg` extension and there is no way to know which one is used for an image without fetching this information from the server using REST APIs.

> This is currently implementing approach 3 from https://github.com/WordPress/performance/issues/341#issuecomment-1143842022, which requires potentially a bunch of extra REST requests.

Every request contains 100 images per page. I believe it's very unlikely to encounter a page that contains more than 100 images.

> Otherwise, the REST requests may add a considerable amount of loading time, which would also result in the img tags not updated as quickly, thus causing the WebP file requests to not be cancelled.

There is definitely a delay caused by the REST request and in some cases it causes slightly weird UI issues (images shows up as broken for a moment before a JPEG version is loaded). Unfortunately, it's a drawback of this approach, but I think that's fine because it will appear only for very old browser where such problem is not the worst one.

> The approach here is certainly a more reliable one - but it comes at the cost of performance I think, so we may still want to think about which of the two to go with further.

Unfortunately, JPEG extensions leave us only the third approach.

modules/images/webp-uploads/load.php
eugene-manuilov: Unfortunately, we can go only with the 3rd approach. See my comment for your previous message.

modules/images/webp-uploads/fallback.js
mitogh: Thanks for your answer.

I don't see any difference tho by not passing document and just using it directly as an argument.

adamsilverstein:
I did some testing with Browserstack across several older Safari versions on Mac OS Sierra, El Capitan and old iOS devices like iPhone 6. The approach worked well in my testing in all cases and the image loaded correctly after a brief delay.  Still reviewing code.

https://user-images.githubusercontent.com/2676022/174409883-95acb6b5-a02d-4bea-a199-396d6a07a451.mp4

modules/images/webp-uploads/load.php
felixarntz: @eugene-manuilov 
> > Given that all WebP images should always have JPEG available unless disabled on the entire setup (...), I think the approach 2 from that implementation may be a more performant approach - just change file extensions from .webp to .jpeg.
>
> @felixarntz, I went with the 2nd approach at the beginning but then it turned out that it wouldn't work well because JPEG images can have either `jpg` or `jpeg` extension and there is no way to know which one is used for an image without fetching this information from the server using REST APIs.

That's a fair point, thanks for clarifying. I didn't think about that, and I agree that the more reliable approach makes more sense here. Guessing the extension and then trying yet another time (e.g. first `jpg`, then `jpeg`) would be even more wasteful than the originally envisioned retry approach.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,48 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+		d.body.appendChild( s );
+	};
+} )( document, 'img', 'script', 'data:image/webp;base64,' );
+EOL;
+
+	wp_print_inline_script_tag(
+		preg_replace( '/\s+/', '', $script ),
+		array(
+			'id'            => 'webpUploadsFallbackWebpImages',
+			'data-rest-api' => get_rest_url(),
```
felixarntz: This is probably the case by default, but since it could be filtered, I'd prefer if we really ensured that this has a trailing slash by wrapping it with `trailingslashit()` - otherwise it can lead to errors in https://github.com/WordPress/performance/pull/360/files#diff-1955d3204dc651eeab9fb7cd46f410cccbc3bfd518f6d438ea84f5df6b57226dR45.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +573,48 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+		)
+	);
+}
+add_action( 'wp_footer', 'webp_uploads_wepb_fallback' );
```
felixarntz: Since this is now unconditionally hooked into `wp_footer`, should we maybe add another conditional to the beginning of this function to not operate if no JPEG image in the content has been replaced with WebP (e.g. relevant on old content where no WebP versions are present)?

We could e.g. set a flag in our function that performs the replacement, or we could even add this action only once from there, if needed. That way we don't unnecessarily load this script on URLs that don't have any WebP images due to our replacement.

We could even check specifically for the replacement from `image/jpeg` to `image/webp` since that is the only replacement for which this JS fallback mechanism applies.

modules/images/webp-uploads/fallback.js
```diff
@@ -0,0 +1,67 @@
...
+
+	var loadMediaDetails = function( nodes ) {
+		var ids = [];
+		for ( var i = 0; i < nodes.length; i++ ) {
+			if ( nodes[i].nodeName !== "IMG" || ! nodes[i].src.match( /\.webp$/i ) ) {
```
felixarntz: We probably should also check if the `srcset` includes any `.webp` files? Especially in the context of #372, it is certainly possible that the main `src` would be a JPEG while some images in `srcset` would be WebP and thus qualify for the fallback.

felixarntz (commented):
@eugene-manuilov Thank you for the updates, this is looking great. I also agree with you now that we should use the REST based approach - it will work better for any WebP images that don't have JPEG versions available.

I left a few more comments with follow-up feedback. Let's also add at least some basic PHPUnit tests later, to ensure that the script is only loaded when needed.

modules/images/webp-uploads/load.php
eugene-manuilov: Ok, updated.

modules/images/webp-uploads/fallback.js
eugene-manuilov: Updated.

modules/images/webp-uploads/load.php
eugene-manuilov: Good idea. Updated PR to load the hook only if we have updated at least one image.

felixarntz (commented):
@eugene-manuilov Excellent, I think this is good to evolve beyond a PoC now :)

Can you add some unit test coverage for the relevant PHP bits? We should also test this a bit more on browsers without WebP support, to at least be a bit more confident that the JS logic all works cross-browser without errors or other problems. It looks solid from a code review perspective though.

eugene-manuilov:
@felixarntz, I have added 2 phpunit tests. Please, take a look.

felixarntz (approved):
@eugene-manuilov Excellent work, LGTM!

adamsilverstein (approved):
Great work!

felixarntz:
@peterwilsoncc Would you mind reviewing this one as well?

modules/images/webp-uploads/fallback.js
```diff
@@ -0,0 +1,73 @@
+( function( d ) {
+	window.webpUploadsFallbackWebpImages = function( media ) {
```
peterwilsoncc:
```suggestion
window.wpPerfLab = wpPerfLab || {};
( function( d ) {
	window.wpPerfLab.webpUploadsFallbackWebpImages = function( media ) {
```

I suggest putting anything exposed like this in a namespace. 

If you intent to merge to core, you might want to use the `wp` namespace.

modules/images/webp-uploads/fallback.js
```diff
@@ -0,0 +1,73 @@
+( function( d ) {
```
peterwilsoncc:
```suggestion
( function( document ) {
```

As this is the full version of the script, I'd pass the document as `document` and the advantage of compression can be handled by the build process. 

Using `d` like this often results in future developers adding code with `document` and then the advantage of passing the variable is lost. 

:1234: would require changes throughout

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +584,47 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+	// We need to add fallback only if jpeg alternatives for the webp images are enabled for the server.
+	$preserve_jpegs_for_jpeg_transforms = isset( $transforms['image/jpeg'] ) && in_array( 'image/jpeg', $transforms['image/jpeg'], true ) && in_array( 'image/webp', $transforms['image/jpeg'], true );
+	$preserve_jpegs_for_webp_transforms = isset( $transforms['image/webp'] ) && in_array( 'image/jpeg', $transforms['image/webp'], true ) && in_array( 'image/webp', $transforms['image/webp'], true );
+	if ( ! $preserve_jpegs_for_jpeg_transforms && ! $preserve_jpegs_for_webp_transforms ) {
+		return;
+	}
+
+	$fallback_url = plugins_url( '/fallback.js', __FILE__ );
+
+	$script = <<<EOL
```
peterwilsoncc: As above, it would be good to use build tools to compress this and keep it readable for humans in the source. 

See how the emoji script does it in core for an example.

modules/images/webp-uploads/fallback.js
```diff
@@ -0,0 +1,73 @@
...
+		for ( var i = 0; i < media.length; i++ ) {
+			try {
+				var ext = media[i].media_details.sources['image/jpeg'].file.match( /\.\w+$/i );
+				if ( ! ext || ! ext[0] ) {
+					continue;
+				}
+
+				var images = d.querySelectorAll( 'img.wp-image-' + media[i].id );
+				for ( var j = 0; j < images.length; j++ ) {
+					images[j].src = images[j].src.replace( /\.webp$/i, ext[0] );
```
peterwilsoncc, line 18: Will the file name always match? 

The thought behind the question: Someone uploads the image `hamilton-the-musical.jpg`, this will generate fallback image `hamilton-the-musical.webp`. Later the image `hamilton-the-musical.jpeg` is uploaded (note the `e`), this will generate images of a different name for the fallback -- probably `hamilton-the-musical-2.webp`.

:1234: The same question applies for `srcset`.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +584,47 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+( function( d, i, s, p ) {
+	s = d.createElement( s );
+	s.src = '{$fallback_url}';
```
peterwilsoncc: `$fallback_url` needs escaping. 

```php
s.src = <?php wp_json_encode( sanitize_url( $fallback_url ) ); ?>
```

It's somewhat complicated by this being in a HEREDOC so you might find it easier to pass it around another way but try to avoid escaping early. 

`sanitize_url()` was undeprecated in WP 5.9 -- if you are supporting earlier versions of WP then use `esc_url_raw()` in it's place.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +584,47 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+		d.body.appendChild( s );
+	};
+} )( document, 'img', 'script', 'data:image/webp;base64,UklGR' );
+EOL;
+
+	wp_print_inline_script_tag(
+		preg_replace( '/\s+/', '', $script ),
+		array(
+			'id'            => 'webpUploadsFallbackWebpImages',
+			'data-rest-api' => trailingslashit( get_rest_url() ),
```
peterwilsoncc:
```suggestion
			'data-rest-api' => sanitizie_url( trailingslashit( get_rest_url() ) ),
```

`wp_print_inline_script_tag()` does escape attributes with `esc_attr()` -- this ensures it's a URL (basically `esc_attr( sanitize_url() ) === esc_url()`.

peterwilsoncc (commented):
I've added some notes inline.

The only things I'd consider blockers are the escaping of URLs.

modules/images/webp-uploads/fallback.js
eugene-manuilov: Sounds good. Updated.

modules/images/webp-uploads/fallback.js
eugene-manuilov: Good point. Updated.

modules/images/webp-uploads/load.php

modules/images/webp-uploads/load.php

modules/images/webp-uploads/load.php
eugene-manuilov: A build tool is out of scope for this PR. I think we will use something similar to the emoji script when we submit this fallback mechanic into the core.

modules/images/webp-uploads/fallback.js
eugene-manuilov: This is actually a good point but I think it is out of scope for this issue. The problem is bigger that just properly replacing webp files here. I tried to replicate your edge case and it looks like in this case webp images for the second image will override webp images for the first image 🤔. 

I uploaded `20180607_144302.jpg` image first, webp images have been created correctly. Then I uploaded the same image but with another extension `20180607_144302.jpeg` and in this time webp images overriden webp images generated for the first image. Here is what I got on the server:

```
-rw-r--r-- 1 www-data www-data 161891 Jul  8 18:15 20180607_144302-1024x576.jpeg
-rw-r--r-- 1 www-data www-data 161891 Jul  8 18:13 20180607_144302-1024x576.jpg
-rw-r--r-- 1 www-data www-data 170276 Jul  8 18:14 20180607_144302-1024x576.webp
-rw-r--r-- 1 www-data www-data   8220 Jul  8 18:15 20180607_144302-150x150.jpeg
-rw-r--r-- 1 www-data www-data   8220 Jul  8 18:13 20180607_144302-150x150.jpg
-rw-r--r-- 1 www-data www-data   8724 Jul  8 18:14 20180607_144302-150x150.webp
-rw-r--r-- 1 www-data www-data 330858 Jul  8 18:15 20180607_144302-1536x864.jpeg
-rw-r--r-- 1 www-data www-data 330858 Jul  8 18:14 20180607_144302-1536x864.jpg
-rw-r--r-- 1 www-data www-data 331106 Jul  8 18:14 20180607_144302-1536x864.webp
-rw-r--r-- 1 www-data www-data 529186 Jul  8 18:15 20180607_144302-2048x1152.jpeg
-rw-r--r-- 1 www-data www-data 529186 Jul  8 18:14 20180607_144302-2048x1152.jpg
-rw-r--r-- 1 www-data www-data 506556 Jul  8 18:14 20180607_144302-2048x1152.webp
-rw-r--r-- 1 www-data www-data  17533 Jul  8 18:15 20180607_144302-300x169.jpeg
-rw-r--r-- 1 www-data www-data  17533 Jul  8 18:13 20180607_144302-300x169.jpg
-rw-r--r-- 1 www-data www-data  18692 Jul  8 18:14 20180607_144302-300x169.webp
-rw-r--r-- 1 www-data www-data  96605 Jul  8 18:15 20180607_144302-768x432.jpeg
-rw-r--r-- 1 www-data www-data  96605 Jul  8 18:14 20180607_144302-768x432.jpg
-rw-r--r-- 1 www-data www-data 102924 Jul  8 18:14 20180607_144302-768x432.webp
-rw-r--r-- 1 www-data www-data 934789 Jul  8 18:15 20180607_144302.jpeg
-rw-r--r-- 1 www-data www-data 934789 Jul  8 18:13 20180607_144302.jpg
-rw-r--r-- 1 www-data www-data 826980 Jul  8 18:14 20180607_144302.webp
```

This doesn't look right although this is a very unlikely edge case. I think it is worth creating a separate issue for it 🤔.

cc @felixarntz @adamsilverstein

adamsilverstein: This was previously discussed and should have been fixed in the plugin - see https://github.com/WordPress/performance/issues/358 - not sure why you still see this happening. I'm not sure this fix is great though, it skips the duplicate generation; instead behavior should be like or use `wp_unique_filename`.

For core I am working on including this in the core patch as well, I tried one approach in [93542af (#2393)](https://github.com/WordPress/wordpress-develop/pull/2393/commits/93542af6127c77f0a4609ce77abf5f616f6984e8) - that isn't quite finished - note there can actually be three variations - cat.jpe, cat.jpeg, and cat.jpg - yes ".jpe" is also valid! 

Another option would be to incorporate the fix into `WP_Image_Editor::generate_filename`  or `save` - again the tricky part is making sure overwriting works when you want it to (ie. regenerate images).

peterwilsoncc: > This is actually a good point but I think it is out of scope for this issue. The problem is bigger that just properly replacing webp files here.

Thanks for testing this. I'm happy if y'all decide it is out of scope for this ticket. It seems like a much bigger problem than I thought when asking the question :)

modules/images/webp-uploads/load.php
```diff
@@ -637,3 +648,47 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+		};
+
+		i.onerror = function() {
+			d.body.appendChild( s );
+		};
+	} )( document, 'img', 'script', 'data:image/webp;base64,UklGR' );
+	<?php
+
+	wp_print_inline_script_tag(
+		preg_replace( '/\s+/', '', ob_get_clean() ),
```
peterwilsoncc:
```suggestion
	$javascript = ob_get_clean();

	wp_print_inline_script_tag(
		preg_replace( '/\s+/', '', $javascript ),
```

Defensive coding in case a later developer wishes to output something prior to printing the script tag.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -635,4 +635,48 @@ function() {
...
+	 * @test
+	 */
+	public function it_should_not_add_fallback_script_if_content_has_no_updated_images() {
+		apply_filters( 'the_content', '<p>no image</p>' );
+
+		$this->assertFalse( has_action( 'wp_footer', 'webp_uploads_wepb_fallback' ) );
+
+		ob_start();
+		wp_footer();
+		$footer = ob_get_clean();
```
peterwilsoncc:
```suggestion
		$footer = get_echo( 'wp_footer' );
```

The WordPress test suite includes a helper for doing this, see [source code](https://github.com/WordPress/wordpress-develop/blob/80380cd3742c623dbe8d7a7c8384e7ceb74ccf4b/tests/phpunit/includes/utils.php#L351-L355).

🔢 Applies to other tests in this file too.

peterwilsoncc (commented):
Couple more notes, nothing major.

modules/images/webp-uploads/fallback.js
```diff
@@ -0,0 +1,75 @@
+window.wpPerfLab = window.wpPerfLab || {};
+
+( function( document ) {
+	window.wpPerfLab.webpUploadsFallbackWebpImages = function( media ) {
+		for ( var i = 0; i < media.length; i++ ) {
+			try {
+				var ext = media[i].media_details.sources['image/jpeg'].file.match( /\.\w+$/i );
```
felixarntz, line 11: Let's be a bit more defensive here. There could be images that don't have the `image/jpeg` MIME type even available, or some images may not even have `sources` set - both cases would result in an error here right now.
So let's cater for that, e.g. like this:
```suggestion
				if ( ! media[i].media_details.sources || ! media[i].media_details.sources['image/jpeg'] ) {
					continue;
				}

				var ext = media[i].media_details.sources['image/jpeg'].file.match( /\.\w+$/i );
```

modules/images/webp-uploads/fallback.js
felixarntz: The problem of avoiding the override was indeed fixed in #358 - however that shouldn't have affected the PR here anyway, since realistically the problem is only on the JPEG (jpeg/jpg/jpe) side. Now that #358 is in place, no conflicting WebP image would be created so we should be good for now.

With that said, I agree this is a valid point, and I think a much more stable implementation for the replacement would be to replace the full file name for the `image/webp` version _per size_ with its corresponding full file name for the `image/jpeg` version. Let's open a follow-up issue to add that.

felixarntz (commented):
@eugene-manuilov This looks solid to me. I'll address the few remaining quirks.

@peterwilsoncc Thank you for the additional feedback! Regarding the bigger conversation on more accurate replacement of image file URLs, I think right now this is not a "real" problem because of the way that WebP files are conditionally generated, but it is certainly something we should address to avoid problems in the future. So my take would be to open a follow-up issue.

felixarntz (approved):
LGTM!

@peterwilsoncc Your feedback should be addressed, will pass back to you for a final review.

felixarntz:
Opened #426 as a follow-up issue to enhance the WebP file replacement in the frontend.

peterwilsoncc (approved):
LGTM, thanks folks.

PR #376, opened 2022-06-15
Use `wp_filesize` instead of `filesize` if available

### LINKED ISSUE #344: Use `wp_filesize` instead of `filesize` if available
WordPress 6.0 introduced the `wp_filesize()` function which should be used in favor of PHP's `filesize()` going forward. We should use it in this plugin where available.

Since we currently support WordPress 5.8 and 5.9 as well, it's probably not worth bumping up the requirement just for that. I suggest we instead create a backward compatibility layer. Maybe the easiest approach would be to implement a "polyfill", where we re-implement `wp_filesize()` in the main plugin file if the function does not exist (yet).

We can then update all module code that calls `filesize()` to instead unconditionally call `wp_filesize()`.

mehulkaklotar:
## Summary
WordPress 6.0 introduced the wp_filesize() function which should be used in favor of PHP's filesize() going forward.

Fixes #344 

## Relevant technical choices

This PR changes `filesize` function to `wp_filesize` everywhere it was used in the plugin. Code and test cases.

## Checklist

### DIFF
--- a/load.php
+++ b/load.php
@@ -197,3 +197,47 @@ perflab_load_active_modules();
 if ( is_admin() ) {
 	require_once plugin_dir_path( __FILE__ ) . 'admin/load.php';
 }
+
+// WP Filesize function.
+if ( ! function_exists( 'wp_filesize' ) ) {
+	/**
+	 * Wrapper for PHP filesize with filters and casting the result as an integer.
+	 *
+	 * This function is introduced in WP 6.0, to support backward compatibility
+	 * function is added as backup here.
+	 *
+	 * @param string $path Path to the file.
+	 *
+	 * @return int The size of the file in bytes, or 0 in the event of an error.
+	 * @since n.e.x.t
+	 *
+	 * @link https://www.php.net/manual/en/function.filesize.php
+	 */
+	function wp_filesize( $path ) {
+		/**
+		 * Filters the result of wp_filesize before the PHP function is run.
+		 *
+		 * @param null|int $size The unfiltered value. Returning an int from the callback bypasses the filesize call.
+		 * @param string $path Path to the file.
+		 *
+		 * @since n.e.x.t
+		 */
+		$size = apply_filters( 'pre_wp_filesize', null, $path );
+
+		if ( is_int( $size ) ) {
+			return $size;
+		}
+
+		$size = file_exists( $path ) ? (int) filesize( $path ) : 0;
+
+		/**
+		 * Filters the size of the file.
+		 *
+		 * @param int $size The result of PHP filesize on the file.
+		 * @param string $path Path to the file.
+		 *
+		 * @since n.e.x.t
+		 */
+		return (int) apply_filters( 'wp_filesize', $size, $path );
+	}
+}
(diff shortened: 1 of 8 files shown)


load.php
```diff
@@ -197,3 +197,47 @@ function perflab_load_active_modules() {
...
+	 * function is added as backup here.
+	 *
+	 * @param string $path Path to the file.
+	 *
+	 * @return int The size of the file in bytes, or 0 in the event of an error.
+	 * @since n.e.x.t
+	 *
+	 * @link https://www.php.net/manual/en/function.filesize.php
+	 */
+	function wp_filesize( $path ) {
```
akshitsethi: @mehulkaklotar Can we move this function to a separate file? Something like `polyfills.php` and include that file here. That will make it easier to add/remove polyfills as we bump versions.

modules/site-health/audit-enqueued-assets/helper.php
```diff
@@ -115,6 +115,6 @@ function perflab_aea_get_path_from_resource_url( $resource_url ) {
  * @return int Returns size if file exists, 0 if it doesn't.
  */
 function perflab_aea_get_resource_file_size( $file_src ) {
-	return file_exists( $file_src ) ? filesize( $file_src ) : 0;
+	return file_exists( $file_src ) ? wp_filesize( $file_src ) : 0;
```
mukeshpanchal27: We have to just `return wp_filesize( $file_src );` because `wp_filesize` function check `file_exists` condition so we don't need to check it twice.

mehulkaklotar: Fixed! thanks @mukeshpanchal27

load.php
mehulkaklotar: Fixed! thanks @akshitsethi

modules/site-health/audit-enqueued-assets/helper.php
mukeshpanchal27: Thank you @mehulkaklotar

polyfills.php
```diff
@@ -0,0 +1,53 @@
...
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+// WP Filesize function.
+if ( ! function_exists( 'wp_filesize' ) ) {
+	/**
+	 * Wrapper for PHP filesize with filters and casting the result as an integer.
+	 *
+	 * This function is introduced in WP 6.0, to support backward compatibility
```
mitogh:
```suggestion
	 * This function was introduced in WP 6.0, for backward compatibility the
```

mitogh (commented):
Small suggestion on the doc block since 6.0 was already released.

Looking good overall.

polyfills.php
felixarntz: Let's also split the sentence, since it's more like 2 separate sentences.
```suggestion
	 * This function was introduced in WP 6.0. To support backward compatibility, the
```

felixarntz (commented):
@mehulkaklotar Thank you, that looks solid. Just one tiny documentation follow-up comment.

akshitsethi (approved):
LGTM :100:

felixarntz (approved):
Great work, thank you @mehulkaklotar!

PR #377, opened 2022-06-15
Add New module proposal issue template

mxbclang:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,6 +1,9 @@
 ---
 name: Bug report
 about: Create a report to help us improve
+title: ''
+labels: ''
```
felixarntz: We can probably set `[Type] Bug` here out of the box.
```suggestion
labels: [ "[Type] Bug" ]
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,6 +1,9 @@
 ---
 name: Bug report
 about: Create a report to help us improve
+title: ''
```
felixarntz: Does this actually do anything? If it's empty, it's the default anyway I assume, so we shouldn't set it here?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,6 +1,9 @@
...
+labels: ''
+assignees: ''
```
felixarntz: Similar to the above, we probably shouldn't set this as it's just the default?

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -1,6 +1,9 @@
 ---
 name: Feature request
 about: Suggest an idea for this project
+title: ''
+labels: ''
+assignees: ''
```
felixarntz: See my comments above, I don't think we should include these if they are just the defaults.

.github/ISSUE_TEMPLATE/new-module-proposal.md
```diff
@@ -0,0 +1,32 @@
+---
+name: New module proposal
+about: Propose a new module for the Performance Lab plugin
+title: 'Module Proposal: '
+labels: Needs Discussion, no milestone, [Type] Module
```
felixarntz: I don't think we need the `no milestone` label on these. A bit unsure about `Needs Discussion` as well, maybe we should limit the defaults to just the label that this needs for sure (which is `[Type] Module`)?

Also I'm not sure the formatting here is correct, according to https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms this needs to be an array I think?
```suggestion
labels: [ "[Type] Module" ]
```

.github/ISSUE_TEMPLATE/new-module-proposal.md
```diff
@@ -0,0 +1,32 @@
...
+assignees: ''
```
felixarntz: See above, let's remove this since it's just the default.

felixarntz (commented):
Thanks @bethanylang! The content lgtm, I left a few comments on the fields definition, there may be a few small problems there.

.github/ISSUE_TEMPLATE/bug_report.md
mxbclang: @felixarntz I actually didn't touch the other templates, so it looks like the commit just pulled them all in by default instead of just the new one. I don't know if this can be removed; let me check.

.github/ISSUE_TEMPLATE/bug_report.md
mxbclang: Same as above

.github/ISSUE_TEMPLATE/feature_request.md
mxbclang: Same as above

.github/ISSUE_TEMPLATE/new-module-proposal.md
mxbclang: @felixarntz I didn't do anything with formatting, just did this in the UI, so again, not sure how to change this. I'll have to set up a local and try updating the actual files tomorrow instead and I can update the labels then, too.

.github/ISSUE_TEMPLATE/new-module-proposal.md
mxbclang: Same as above, will need to play wit this more.

mitogh (approved):
Great work 🎉

felixarntz (approved):
@bethanylang Awesome, thanks!

PR #381, opened 2022-06-16
Optimize computing dominant color and transparency for images by combining the two functions

### LINKED ISSUE #373: Avoid duplicate execution of setting up image editor when determining dominant color and transparency
Follow-up to #19 / #282: Currently, both the `dominant_color_get_dominant_color()` and the `dominant_color_has_transparency()` functions execute the same logic to get the attachment file and instantiate the correct editor based on it.

This is unnecessary and somewhat wasteful since the only place where the two functions are relevant is in immediately subsequent calls for when the image has just been uploaded.

In addition to that, the two functions suggest by their public nature and clear names that they are okay to be called by third-party developers - which is not the case. These two functions perform expensive actions, and they **must not** be called "lightly". Their current name and documentation may lead to plugin developers calling them in the frontend to determine the dominant color and transparency of an image, but that is definitely not the intended use-case.

If we want to provide an easy-to-use API to get the dominant color and transparency of an image, these functions should rely on the already computed and stored values in the attachment's metadata. The way that the current functions behave is a no go for a "public API" though, and per the initial point of this issue is also slightly wasteful for performance during the upload process of an image.

Here's what I propose:
* Combine the two current functions `dominant_color_get_dominant_color()` and `dominant_color_has_transparency()` into a single function marked as `@access private` that computes both values and returns them, or a `WP_Error`.
* Introduce two new functions `dominant_color_get_dominant_color()` and `dominant_color_has_transparency()` instead, which can truly serve as an "API" that is okay to be called from third-party developers. Both of these functions should simply look at the attachment's metadata and returned the stored value if set, or something like `null` otherwise.

cc @spacedmonkey @pbearne

pbearne:
…or_has_transparency() functions into single function

Updated tests and improved the images
Introduced two new functions dominant_color_get_dominant_color() and dominant_color_has_transparency() that fetchs the date from image meta

## Summary

Fixes #

## Checklist


modules/images/dominant-color/class-dominant-color-image-editor-gd.php
```diff
@@ -35,12 +35,7 @@ public function get_dominant_color() {
 		$shorted_image = imagecreatetruecolor( 1, 1 );
 		imagecopyresampled( $shorted_image, $this->image, 0, 0, 0, 0, 1, 1, imagesx( $this->image ), imagesy( $this->image ) );
 
-		$hex = dechex( imagecolorat( $shorted_image, 0, 0 ) );
-
-		if ( strlen( $hex ) < 6 ) {
-			return new WP_Error( 'image_editor_dominant_color_error', __( 'Dominant color detection failed.', 'performance-lab' ) );
-		}
-		return $hex;
+		return dechex( imagecolorat( $shorted_image, 0, 0 ) );
```
spacedmonkey: Can you explain this change a little more.

modules/images/dominant-color/class-dominant-color-image-editor-imagick.php
```diff
@@ -63,8 +61,25 @@ public function has_transparency() {
...
+			for ( $x = 0; $x < $w; $x++ ) {
+				for ( $y = 0; $y < $h; $y++ ) {
+					$pixel = $this->image->getImagePixelColor( $x, $y );
+					$color = $pixel->getColor();
+					if ( $color['a'] > 0 ) {
+						return true;
+					}
+				}
+			}
+			return false;
```
spacedmonkey, line 92: Why is this change needed?

modules/images/dominant-color/load.php
```diff
@@ -339,3 +320,45 @@ function wp_get_attachment_file_path( $attachment_id, $size = 'medium' ) {
...
+/**
+ * Gets the Dominant color for an image attachment.
+ *
+ * @since n.e.x.t
+ *
+ * @param int $attachment_id Attachment ID for image.
+ * @return string|null|WP_Error hex value of dominant color or null if not set.
+ */
+function dominant_color_get_dominant_color( $attachment_id ) {
+	$image_meta = wp_get_attachment_metadata( $attachment_id );
```
spacedmonkey, line 336:
```suggestion
	if ( ! wp_attachment_is_image( $attachment_id ) ) {
		return null;
	}
	$image_meta = wp_get_attachment_metadata( $attachment_id );
```

spacedmonkey: Should we not check if the attachent exists / is image.

modules/images/dominant-color/load.php
```diff
@@ -339,3 +320,45 @@ function wp_get_attachment_file_path( $attachment_id, $size = 'medium' ) {
...
+/**
+ * Gets the has_transparency value for an image attachment.
+ *
+ * @since n.e.x.t
+ *
+ * @param int $attachment_id Attachment ID for image.
+ * @return bool|null|WP_Error true if may have transparency or null if not set.
+ */
+function dominant_color_has_transparency( $attachment_id ) {
+	$image_meta = wp_get_attachment_metadata( $attachment_id );
```
spacedmonkey, line 357: Feedback - https://github.com/WordPress/performance/pull/381/files#r901628920

modules/images/dominant-color/class-dominant-color-image-editor-gd.php
pbearne: we are getting color with the shorten form #fff back from the underlining PHP code

modules/images/dominant-color/class-dominant-color-image-editor-imagick.php
pbearne: I believe that an image just having an alpha channel doesn't mean it has any transparency so I added a test to check to look actual transparent pixels

tests/utils/TestCase/DominantColorTestCase.php
```diff
@@ -13,107 +13,91 @@ abstract class DominantColorTestCase extends WP_UnitTestCase {
...
 				'image_path'     => TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/dominant-color/red.webp',
-				'expected_color' => array( 'ff0505', 'ff55', 'ff0506', 'ff56' ),
+				'expected_color' => array( 'ff0000', 'ff00' ),
 				'is_wp_error'    => false,
 			),
-			'green_webp'  => array(
+			'green_webp'    => array(
 				'image_path'     => TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/dominant-color/green.webp',
-				'expected_color' => array( '1de303', '1de33' ),
+				'expected_color' => array( '00ff00', '0ff0', 'ff01', 'ff00' ),
```
spacedmonkey: Can we ensure that we ALWAYS generate a 6 character hex color? I don't love the results are this unpredictable here.

modules/images/dominant-color/class-dominant-color-image-editor-gd.php
spacedmonkey: I think it better to always return a 6 character hex here.

modules/images/dominant-color/load.php
```diff
@@ -303,15 +274,25 @@ function dominant_color_has_transparency( $attachment_id ) {
...
+
 	if ( ! method_exists( $editor, 'has_transparency' ) ) {
 		return new WP_Error( 'unable_to_find_method', __( 'Unable to find has_transparency method', 'performance-lab' ) );
 	}
 	$has_transparency = $editor->has_transparency();
-	if ( is_wp_error( $has_transparency ) ) {
-		return $has_transparency;
+	if ( ! is_wp_error( $has_transparency ) ) {
+		$dominant_color_data['has_transparency'] = $has_transparency;
+	}
```
spacedmonkey, line 284: Why are we ignoring this error here? This also makes the code much harder to test.

modules/images/dominant-color/load.php
```diff
@@ -303,15 +274,25 @@ function dominant_color_has_transparency( $attachment_id ) {
...
+
+	if ( ! method_exists( $editor, 'get_dominant_color' ) ) {
+		return new WP_Error( 'unable_to_find_method', __( 'Unable to find get_dominant_color method', 'performance-lab' ) );
+	}
+
+	$dominant_color = $editor->get_dominant_color();
+	if ( ! is_wp_error( $dominant_color ) ) {
+		$dominant_color_data['dominant_color'] = $dominant_color;
```
spacedmonkey, line 292: Same question - https://github.com/WordPress/performance/pull/381/files#r902470598

modules/images/dominant-color/load.php
```diff
@@ -257,41 +258,11 @@ function dominant_color_set_image_editors() {
...
-
-/**
- * Computes whether the given attachment image has transparency.
+ * @return array|WP_Error The dominant color of the image, or WP_Error on error.
  *
- * @since n.e.x.t
- *
- * @param int $attachment_id The attachment ID.
- * @return bool|WP_Error True if the color has transparency or WP_Error on error.
+ * @access private
```
felixarntz: This should go right under the `@since` line.

Also, can you please update the function description accordingly? It should now say that it computes _both_ the dominant color and whether the image has transparency.

modules/images/dominant-color/load.php
felixarntz: No need to change this - it should still return a `WP_Error` if it can't determine whether the image has transparency.

modules/images/dominant-color/load.php
```diff
@@ -303,15 +274,25 @@ function dominant_color_has_transparency( $attachment_id ) {
...
+	}
+
+	if ( ! method_exists( $editor, 'get_dominant_color' ) ) {
+		return new WP_Error( 'unable_to_find_method', __( 'Unable to find get_dominant_color method', 'performance-lab' ) );
+	}
+
+	$dominant_color = $editor->get_dominant_color();
+	if ( ! is_wp_error( $dominant_color ) ) {
+		$dominant_color_data['dominant_color'] = $dominant_color;
 	}
```
felixarntz: Same here, this should return the error instead.
```suggestion
	if ( is_wp_error( $dominant_color ) ) {
		return $dominant_color;
	}
	$dominant_color_data['dominant_color'] = $dominant_color;
```

modules/images/dominant-color/load.php
```diff
@@ -339,3 +320,48 @@ function wp_get_attachment_file_path( $attachment_id, $size = 'medium' ) {
 
 	return $filepath;
 }
+
+/**
+ * Gets the Dominant color for an image attachment.
```
felixarntz:
```suggestion
 * Gets the dominant color for an image attachment.
```

modules/images/dominant-color/load.php
```diff
@@ -339,3 +320,48 @@ function wp_get_attachment_file_path( $attachment_id, $size = 'medium' ) {
...
+ * @param int $attachment_id Attachment ID for image.
+ * @return string|null|WP_Error hex value of dominant color or null if not set.
+ */
+function dominant_color_get_dominant_color( $attachment_id ) {
+	if ( ! wp_attachment_is_image( $attachment_id ) ) {
+		return null;
+	}
+	$image_meta = wp_get_attachment_metadata( $attachment_id );
+	if ( ! is_array( $image_meta ) ) {
+		return new WP_Error( 'unable_to_find_imagemeta_for_id', __( 'Unable to find metadata for image for given ID', 'performance-lab' ) );
```
felixarntz: Functions that just return existing data shouldn't return a `WP_Error`, but rather `null`. If there is no image metadata, this is not an error for this function, it just means no dominant color is set.
```suggestion
		return null;
```

modules/images/dominant-color/load.php
```diff
@@ -339,3 +320,48 @@ function wp_get_attachment_file_path( $attachment_id, $size = 'medium' ) {
...
 	return $filepath;
 }
+
+/**
+ * Gets the Dominant color for an image attachment.
+ *
+ * @since n.e.x.t
+ *
+ * @param int $attachment_id Attachment ID for image.
+ * @return string|null|WP_Error hex value of dominant color or null if not set.
```
felixarntz: See my other comment:
```suggestion
 * @return string|null Hex value of dominant color or null if not set.
```

modules/images/dominant-color/load.php
```diff
@@ -339,3 +320,48 @@ function wp_get_attachment_file_path( $attachment_id, $size = 'medium' ) {
...
+ *
+ * @since n.e.x.t
+ *
+ * @param int $attachment_id Attachment ID for image.
+ * @return bool|null|WP_Error true if may have transparency or null if not set.
+ */
+function dominant_color_has_transparency( $attachment_id ) {
+	$image_meta = wp_get_attachment_metadata( $attachment_id );
+	if ( ! is_array( $image_meta ) ) {
+		return new WP_Error( 'unable_to_find_imagemeta_for_id', __( 'Unable to find metadata for image for given ID', 'performance-lab' ) );
```
felixarntz: See my other comment on the function above.
```suggestion
		return null;
```

modules/images/dominant-color/load.php
```diff
@@ -339,3 +320,48 @@ function wp_get_attachment_file_path( $attachment_id, $size = 'medium' ) {
...
+	return $image_meta['dominant_color'];
+}
+
+/**
+ * Gets the has_transparency value for an image attachment.
+ *
+ * @since n.e.x.t
+ *
+ * @param int $attachment_id Attachment ID for image.
+ * @return bool|null|WP_Error true if may have transparency or null if not set.
```
felixarntz: See related comment:
```suggestion
 * @return bool|null Whether the image has transparency, or null if not set.
```

modules/images/dominant-color/load.php
```diff
@@ -339,3 +320,48 @@ function wp_get_attachment_file_path( $attachment_id, $size = 'medium' ) {
...
+
+	if ( ! isset( $image_meta['dominant_color'] ) ) {
+		return null;
+	}
+
+	return $image_meta['dominant_color'];
+}
+
+/**
+ * Gets the has_transparency value for an image attachment.
```
felixarntz:
```suggestion
 * Returns whether an image attachment has transparency.
```

felixarntz (commented):
@pbearne I think the changes related to the issue look good for the most part, but this PR includes a bunch of additional changes which I think are out of scope as they do not relate to the underlying issue.

I think the only changes that are needed here are those in the `load.php` file (plus related tests). What is the reasoning for modifying the implementation of the image classes here? That seems out of scope.

modules/images/dominant-color/class-dominant-color-image-editor-gd.php
spacedmonkey:
```suggestion
		$rgb = imagecolorat( $shorted_image, 0, 0 );
		$r = ($rgb >> 16) & 0xFF;
		$g = ($rgb >> 8) & 0xFF;
		$b = $rgb & 0xFF;
		return sprintf("%02x%02x%02x", $r, $g, $b);
```

Spent ages on this one. But finally got 6 digit hex values returning. Basically what this does, cover to rgb and then convert to hex from there.

spacedmonkey: Fixed.

modules/images/dominant-color/load.php
pbearne: changed

modules/images/dominant-color/load.php
pbearne: updated

modules/images/dominant-color/class-dominant-color-image-editor-gd.php
```diff
@@ -35,11 +35,7 @@ public function get_dominant_color() {
 		$shorted_image = imagecreatetruecolor( 1, 1 );
 		imagecopyresampled( $shorted_image, $this->image, 0, 0, 0, 0, 1, 1, imagesx( $this->image ), imagesy( $this->image ) );
 
-		$rgb = imagecolorat( $shorted_image, 0, 0 );
-		$r   = ( $rgb >> 16 ) & 0xFF;
-		$g   = ( $rgb >> 8 ) & 0xFF;
-		$b   = $rgb & 0xFF;
-		return sprintf( '%02x%02x%02x', $r, $g, $b );
+		return str_pad( dechex( imagecolorat( $shorted_image, 0, 0 ) ), 6, '0', STR_PAD_LEFT );
```
spacedmonkey: Let's go back to this. 

```
sprintf( '%02x%02x%02x', $color['r'], $color['g'], $color['b'] );
```

It is pretty clear what it does. It uses red, green and blue to generate a hex colour. This is used elsewhere and works just fine.

modules/images/dominant-color/class-dominant-color-image-editor-gd.php
```diff
@@ -35,11 +35,15 @@ public function get_dominant_color() {
 		$shorted_image = imagecreatetruecolor( 1, 1 );
 		imagecopyresampled( $shorted_image, $this->image, 0, 0, 0, 0, 1, 1, imagesx( $this->image ), imagesy( $this->image ) );
 
-		$hex = dechex( imagecolorat( $shorted_image, 0, 0 ) );
-
-		if ( strlen( $hex ) < 6 ) {
+		$rgb = imagecolorat( $shorted_image, 0, 0 );
+		$r   = ( $rgb >> 16 ) & 0xFF;
```
spacedmonkey, line 39: https://www.php.net/manual/en/function.imagecolorat.php#example-2657

spacedmonkey (approved):
I pushed up some changed to add some more unit test coverage.

modules/images/dominant-color/load.php
```diff
@@ -255,43 +252,15 @@ function dominant_color_set_image_editors() {
  * Computes the dominant color of the given attachment image.
```
felixarntz:
```suggestion
 * Computes the dominant color of the given attachment image and whether it has transparency.
```

modules/images/dominant-color/load.php
```diff
@@ -339,3 +320,69 @@ function wp_get_attachment_file_path( $attachment_id, $size = 'medium' ) {
...
+ *
+ * @since n.e.x.t
+ *
+ * @param int $red Red 0-255.
+ * @param int $green Green 0-255.
+ * @param int $blue Blue 0-255.
+ *
+ * @return string|null Hex color or null if error.
+ */
+function get_hex_from_rgb( $red, $green, $blue ) {
```
felixarntz: Let's prefix this function as usual. Also, I think it's more common for naming of this type of function to put the source format first, i.e. "rgb to hex". There are already some similarly named functions in WordPress core, e.g. `wp_tinycolor_hue_to_rgb()`. So I think we should rename this to:
```suggestion
function dominant_color_rgb_to_hex( $red, $green, $blue ) {
```

modules/images/dominant-color/load.php
```diff
@@ -339,3 +320,69 @@ function wp_get_attachment_file_path( $attachment_id, $size = 'medium' ) {
...
+	if ( ! isset( $image_meta['has_transparency'] ) ) {
+		return null;
+	}
+
+	return $image_meta['has_transparency'];
+}
+
+
+/**
+ * Get Hex color from RGB.
```
felixarntz:
```suggestion
 * Gets hex color from RGB.
```

felixarntz (commented):
@pbearne Pretty much LGTM now. The remaining comments are minor, would be great if you could address them.

felixarntz (approved):
@spacedmonkey Thanks for the quick iteration. LGTM! 🎉

PR #390, opened 2022-06-22
Implement mechanism to not load module if core version is available

### LINKED ISSUE #293: Implement mechanism to not load module if core version is available
As entire modules eventually get merged into core, the plugin should provide a mechanism to not load certain modules if their core counterpart is already available on the site.

We need some centralized infrastructure that each module can leverage to determine when the module is relevant to be loaded. If a module defines that based on the current situation it is no longer needed to be loaded, this should furthermore surface to the administrator on the Performance Lab modules admin screen.

Proposed implementation:
* Add support for an **optional** `can-load.php` file that can be present in each module directory.
    * If the file is present, it should be expected to **return a closure function that computes whether the module is relevant to be loaded in the current environment**. For example, if a module has already been merged into WordPress core and that core version is already running on the current site, the closure function should return `false`. This can be typically accomplished by checking for the existence of certain WordPress core functions related to the module's feature set.
    * If the file is not present, it should be assumed that the module can be loaded (similar to today).
    * The foundation for this should be implemented in a `perflab_can_load_module( $module ) : bool` function which expects the module directory relative to the modules root directory to be passed (e.g. `images/webp-uploads`).
* When loading the active modules, for each module the above should be checked via `perflab_can_load_module()`. In case the `can-load.php` file exists and its closure function returns `false`, that module should not be loaded.
* The "Performance" admin screen should be enhanced so that modules that cannot be loaded in the current environment (checked via `perflab_can_load_module()`) are marked as such:
    * Each module that cannot be loaded in the current environment should have its `input[type="checkbox"]` as `readonly`.
    * In addition, the checkbox label should in that case be overwritten to say: _%s is already part of your WordPress version and therefore cannot be loaded as part of the plugin._ (where `%s` is the module name)

As part of this issue, the new feature should only be implemented by the `images/webp-uploads` module:
* If the function `wp_image_use_alternate_mime_types()` exists (see WordPress core PR https://github.com/WordPress/wordpress-develop/pull/2393), the module's `can-load.php` closure should return `false`. Otherwise, it should return `true`.

mukeshpanchal27:
## Summary

This PR works when function `wp_image_use_alternate_mime_types()` (see WordPress core PR https://github.com/WordPress/wordpress-develop/pull/2393) implemented in WordPress core.

## How to test PR in WordPress 6.0
If you want to test PR then please  pull from the git change [this](https://github.com/WordPress/performance/blob/fix/293-mechanism-not-load-module/modules/images/webp-uploads/can-load.php#L17) line code.

`if ( function_exists( 'wp_image_use_alternate_mime_types' ) ) {`

Replace it with `wp_recursive_ksort()` which introduce in WordPress 6.0.

`if ( function_exists( 'wp_recursive_ksort' ) ) {`

If you want to test `can-load.php` for the other module then please copy the file from the `webp-uploads` module and paste it into the other modules.

For testing purposes, I have created a `can-load.php` file for `dominant-color` and `persistent-object-cache-health-check` modules.

- [`can-load.php`](https://github.com/mukeshpanchal27/performance/blob/fix/mechanism-not-load-module/modules/images/dominant-color/can-load.php) file for `dominant-color`
- [`can-load.php`](https://github.com/mukeshpanchal27/performance/blob/fix/mechanism-not-load-module/modules/object-cache/persistent-object-cache-health-check/can-load.php) file for `persistent-object-cache-health-check`

Fixes #293

## Checklist


load.php
```diff
@@ -168,6 +186,38 @@ function perflab_render_generator() {
...
 
+/**
+ * Don't load the performance modules if the core version of the module is available.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $module The name of the module.
+ * @return bool Whether to load module or not.
+ */
+function perflab_can_load_module( $module ) {
```
mitogh, line 203: Probably personal preference but `can load` the module might involve other criteria besides just having the method available in WP Core like if your installation supports or not WebP conversion (for WebP images).

I feel like something along the lines of `function_exists` (from PHP Core) makes more sense like:

- `perlab_module_exists`

load.php
```diff
@@ -134,7 +134,25 @@ function( $module_settings ) {
...
+	$active_modules = array();
+	foreach ( $modules as $module ) {
+
+		// Do not load module if it marge in WordPress.
+		$can_load_module = perflab_can_load_module( $module );
+		if ( ! $can_load_module ) {
+			continue;
+		}
+
+		// Do not load module if it no longer exists.
```
mitogh:
```suggestion
		// Do not load module if no longer exists.
```

load.php
```diff
@@ -134,7 +134,25 @@ function( $module_settings ) {
 	 */
 	$modules = apply_filters( 'perflab_active_modules', $modules );
 
-	return $modules;
+	$active_modules = array();
+	foreach ( $modules as $module ) {
+
+		// Do not load module if it marge in WordPress.
```
mitogh:
```suggestion
		// Do not load module if was merge in WordPress.
```

load.php
```diff
@@ -168,6 +186,38 @@ function perflab_render_generator() {
...
+function perflab_can_load_module( $module ) {
+	$module_load_file = plugin_dir_path( __FILE__ ) . 'modules/' . $module . '/can-load.php';
+
+	// If the load file does not exist, the feature is not in core and the module can be loaded.
+	if ( ! file_exists( $module_load_file ) ) {
+		return true;
+	}
+
+	// Require the file to include the add_filter function.
+	require_once $module_load_file;
```
mitogh: I feel like a filter should be in place to conditionally determine if the module should be loaded or not, before the actual file is included.

modules/images/webp-uploads/can-load.php
```diff
@@ -0,0 +1,30 @@
...
+
+	if ( 'images/webp-uploads' !== $module ) {
+		return $can_load;
+	}
+
+	if ( function_exists( 'wp_image_use_alternate_mime_types' ) ) {
+		return false;
+	}
+
+	return true;
```
mitogh:
```suggestion
	return ! function_exists( 'wp_image_use_alternate_mime_types' );
```

tests/load-tests.php
```diff
@@ -110,25 +110,53 @@ function( $module_settings ) {
...
 		$new_value = array(
 			'inactive-module' => array( 'enabled' => false ),
 			'active-module'   => array( 'enabled' => true ),
 		);
 		update_option( PERFLAB_MODULES_SETTING, $new_value );
 		$active_modules = perflab_get_active_modules();
-		$this->assertSame( array( 'active-module' ), $active_modules );
+		$this->assertSame( array(), $active_modules );
+
+		// Assert that it only allow existing modules.
```
mitogh:
```suggestion
		// Assert that only allows existing modules.
```

load.php
jjgrainger: Fix spelling
```suggestion
		// Do not load module if it was merged into WordPress.
```

load.php
```diff
@@ -168,6 +186,38 @@ function perflab_render_generator() {
...
+
+	/**
+	 * Filters whether the module can load or not.
+	 *
+	 * You can set this to false in order to disable the module.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param bool   $can_load_module Whether to load module. default true.
```
jjgrainger: Minor spelling
```suggestion
	 * @param bool   $can_load_module Whether to load module. Default true.
```

modules/images/webp-uploads/can-load.php
```diff
@@ -0,0 +1,30 @@
...
+ * @since   n.e.x.t
+ * @package performance-lab
+ */
+
+/**
+ * Filters whether the module can load or not.
+ *
+ * @since n.e.x.t
+ *
+ * @param bool   $can_load Whether to load module. default true.
```
jjgrainger:
```suggestion
 * @param bool   $can_load Whether to load module. Default true.
```

jjgrainger (changes requested):
Looks good to me based on our discussions previously, just a few spelling fixes I noticed.

admin/load.php
```diff
@@ -126,29 +126,38 @@ function perflab_render_modules_page() {
...
+	$enabled         = isset( $module_settings['enabled'] ) && $module_settings['enabled'];
+	$can_load_module = perflab_can_load_module( $module_slug );
 	?>
 	<fieldset>
 		<legend class="screen-reader-text">
 			<?php echo esc_html( $module_data['name'] ); ?>
 		</legend>
 		<label for="<?php echo esc_attr( "{$base_id}_enabled" ); ?>">
-			<input type="checkbox" id="<?php echo esc_attr( "{$base_id}_enabled" ); ?>" name="<?php echo esc_attr( "{$base_name}[enabled]" ); ?>" aria-describedby="<?php echo esc_attr( "{$base_id}_description" ); ?>" value="1"<?php checked( $enabled ); ?>>
+			<input type="checkbox" id="<?php echo esc_attr( "{$base_id}_enabled" ); ?>" name="<?php echo esc_attr( "{$base_name}[enabled]" ); ?>" aria-describedby="<?php echo esc_attr( "{$base_id}_description" ); ?>" value="1"<?php checked( $enabled ); ?><?php echo ( ! $can_load_module ? ' readonly onclick="return false;"' : '' ); ?>>
```
felixarntz: I don't think we should have the `onclick` attribute in here - what's the purpose of that? `readonly` should be sufficient?

admin/load.php
```diff
@@ -126,29 +126,38 @@ function perflab_render_modules_page() {
...
+						esc_html( $module_data['name'] )
+					);
+				} else {
+					printf(
+						/* translators: %s: module name */
+						__( 'Enable %s', 'performance-lab' ),
+						esc_html( $module_data['name'] )
+					);
+				}
 			}
```
felixarntz: Nit-pick, but it would be a bit more elegant here to avoid further nesting by using `elseif` and keep these alternate scenarios at one indentation level.

load.php
felixarntz:
```suggestion
		// Do not load module if it cannot be loaded, e.g. if it was already merged and is available in WordPress core.
```

load.php
```diff
@@ -134,7 +134,25 @@ function( $module_settings ) {
...
+		// Do not load module if it no longer exists.
+		$module_file = plugin_dir_path( __FILE__ ) . 'modules/' . $module . '/load.php';
+		if ( ! file_exists( $module_file ) ) {
+			continue;
+		}
+
+		$active_modules[] = $module;
+	}
+
+	return $active_modules;
```
felixarntz: It is worth discussing, but I think it would be more appropriate to put all of this logic into the `perflab_load_active_modules()` function. Which modules are technically active isn't affected by which modules can be loaded, and I think it makes more sense to handle the whole logic related to loading a module in that `perflab_load_active_modules()` function.

This keeps the two functions more clearly separated. By adding all this logic here, we basically change this function into a "get active and loadable modules" function, but it is only `perflab_get_active_modules()`.

I'd say we should either move all these checks (back) into `perflab_load_active_modules()`, or alternatively we could introduce a completely new function, e.g. `perflab_get_active_and_valid_modules()` for it. (That name would make sense for parity with the similar core function [`wp_get_active_and_valid_plugins()`](https://developer.wordpress.org/reference/functions/wp_get_active_and_valid_plugins/).)

load.php
```diff
@@ -168,6 +186,38 @@ function perflab_render_generator() {
 }
 add_action( 'wp_head', 'perflab_render_generator' );
 
+/**
+ * Don't load the performance modules if the core version of the module is available.
```
```suggestion
 * Checks whether the given module can be loaded in the current environment.
```

load.php
```diff
@@ -168,6 +186,38 @@ function perflab_render_generator() {
...
+ *
+ * @since n.e.x.t
+ *
+ * @param string $module The name of the module.
+ * @return bool Whether to load module or not.
```
felixarntz: We call this "module slug" (or sometimes "module directory") elsewhere.
```suggestion
 * @param string $module Slug of the module.
 * @return bool Whether the module can be loaded or not.
```

load.php
```diff
@@ -168,6 +186,38 @@ function perflab_render_generator() {
...
+ */
+function perflab_can_load_module( $module ) {
+	$module_load_file = plugin_dir_path( __FILE__ ) . 'modules/' . $module . '/can-load.php';
+
+	// If the load file does not exist, the feature is not in core and the module can be loaded.
```
felixarntz: Similar to my other feedback, I'd rephrase the comments since the module being in core is not necessarily the only way that affects this. We should rather frame this more generically just around whether the module can be loaded by the current environment, and the module being in core is just one potential reason for that not being possible.
```suggestion
	// If the `can-load.php` file does not exist, assume the module can be loaded.
```

load.php
felixarntz: There's already a filter to control the active modules, so if any site wanted to add control beyond that, they could already do so with that filter. I don't think a filter is needed here.

load.php
```diff
@@ -168,6 +186,38 @@ function perflab_render_generator() {
...
+	 * Filters whether the module can load or not.
+	 *
+	 * You can set this to false in order to disable the module.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param bool   $can_load_module Whether to load module. default true.
+	 * @param string $module          The name of the module.
+	 */
+	return apply_filters( 'perflab_can_load_module', true, $module );
```
felixarntz: This implementation is different from the proposed one in the issue. Could you explain why you went with that approach instead?

I think using a filter here makes this more complicated than it needs to be. To clarify, what was outlined in the issue would be the following:
```suggestion
	// Require the file to get the closure for whether the module can load.
	$can_load = require $module_load_file;

	// If the `can-load.php` file is invalid and does not return a closure, assume the module can be loaded.
	if ( ! is_callable( $can_load ) ) {
		return true;
	}

	// Call the closure to determine whether the module can be loaded.
	return $can_load();
```

load.php
```diff
@@ -181,13 +231,8 @@ function perflab_load_active_modules() {
...
 
 	foreach ( $active_modules as $module ) {
-		// Do not load module if it no longer exists.
-		$module_file = plugin_dir_path( __FILE__ ) . 'modules/' . $module . '/load.php';
-		if ( ! file_exists( $module_file ) ) {
-			continue;
-		}
 
-		require_once $module_file;
+		require_once plugin_dir_path( __FILE__ ) . 'modules/' . $module . '/load.php';
```
felixarntz: See my comment further above, maybe we should rather do the checks for whether the module can be loaded in here? Either directly within this function or via a separate `perflab_get_active_and_valid_modules()`?

modules/images/webp-uploads/can-load.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+function perflab_check_webp_uploads_core_functions( $can_load, $module ) {
+
+	if ( 'images/webp-uploads' !== $module ) {
+		return $can_load;
+	}
+
+	return ! function_exists( 'wp_image_use_alternate_mime_types' );
+}
+add_filter( 'perflab_can_load_module', 'perflab_check_webp_uploads_core_functions', 10, 2 );
```
felixarntz: See my other comment, here you can better tell how the proposed implementation from the issue would make this much simpler than it is right now. Based on the issue description, this would be a lot less code, a simple closure, and one less condition:
```suggestion
return function() {
	return ! function_exists( 'wp_image_use_alternate_mime_types' );
};
```

felixarntz (changes requested):
@mukeshpanchal27 This is mostly looking good so far, however the API approach you're taking here is different from the issue, so I'm curious why you decided to divert from that - IMO it's more complicated than needed.

That's my main feedback, other than that I mostly have minor comments.

load.php
mukeshpanchal27: Thanks for the feedback! Text changes updated.

load.php
mukeshpanchal27: Thanks for the feedback! Text changes updated.

tests/load-tests.php
mukeshpanchal27: Thanks for the feedback! Text changes updated.

load.php
mukeshpanchal27: Thanks for the feedback! Text changes updated.

load.php
mukeshpanchal27: Thanks for the feedback! Text changes updated.

modules/images/webp-uploads/can-load.php
mukeshpanchal27: Thanks for the feedback! Text changes updated.

admin/load.php
mukeshpanchal27: Thanks for the feedback! Condition changed

load.php
mukeshpanchal27: Thanks for the feedback! Text changes updated.

admin/load.php
mukeshpanchal27: I forget to add details related to this change in the original comment, Thanks for pointing it out.

If we use `readonly` than we don't block checkbox functinality because WordPress has only below CSS for the `readable` attribute and there is no other JS for it. In most cases for `readable`, they have used additional CSS property called `pointer-events: none;` but we don't have it so as per [this doc](https://stackoverflow.com/questions/12267242/how-can-i-make-a-checkbox-readonly-not-disabled) I use the return function and use `readonly` so it feels like it is disabled.

    input.readonly, input[readonly], textarea.readonly, textarea[readonly] {
        background-color: #f0f0f1;
    }

eugene-manuilov: Why not to just disable such checkboxes using the `disabled` attribute? What is the point of having a checkbox to be readonly? This `onclick="return false;"` approach looks like a hack.

admin/load.php
```diff
@@ -126,18 +126,25 @@ function perflab_render_modules_page() {
  * @param array  $module_settings Associative array of the module's current settings.
  */
 function perflab_render_modules_page_field( $module_slug, $module_data, $module_settings ) {
-	$base_id   = sprintf( 'module_%s', $module_slug );
-	$base_name = sprintf( '%1$s[%2$s]', PERFLAB_MODULES_SETTING, $module_slug );
-	$enabled   = isset( $module_settings['enabled'] ) && $module_settings['enabled'];
+	$base_id         = sprintf( 'module_%s', $module_slug );
+	$base_name       = sprintf( '%1$s[%2$s]', PERFLAB_MODULES_SETTING, $module_slug );
+	$enabled         = isset( $module_settings['enabled'] ) && $module_settings['enabled'];
+	$can_load_module = perflab_can_load_module( $module_slug );
```
eugene-manuilov: @felixarntz When I see `can load module` it always feels like there are might be some limitations why the module can't be loaded which is not the case in our situation because what we check here is whether the module is implemented in the core or not. Maybe we should rename this function to be `perflab_module_is_implemented_in_core` or something similar to make it clear what it does?

felixarntz: @eugene-manuilov That would limit the use-case more than needed. This closure could be used for other checks than just whether this exists in core. We could even have something like a module that only works in PHP 7 just to make an example. So we shouldn't limit ourselves artificially.

I'm certainly open to finding a better name, but it shouldn't be tied to whether the feature is available in core.

admin/load.php
felixarntz: @mukeshpanchal27 @eugene-manuilov I think technically `readonly` makes more sense here than `disabled` because otherwise we won't send the current value of the option with the form submission, which could cause it to change. That wouldn't be a major problem since the module can't be loaded anyway, but I'd prefer if we kept the actual option values unaffected by whether the module can load or not.

Since WordPress doesn't style the `readonly` attribute inputs correctly, maybe we should make it `disabled`, but then add an `input[type="hidden"]` right after just so that the value is still included in the form submission as usual?

load.php
mukeshpanchal27: Thanks for the review! That makes sense. We don't need a filter here to check if a module is loaded or not.

load.php
mukeshpanchal27: The docblock is updated. Thanks!

load.php
mukeshpanchal27: At first, I implement it with the function based solution [here](https://github.com/mukeshpanchal27/performance/pull/2/commits/cfd8b5f27732376f0c3edfbe06c9ad48e9d0bcec) but in some scenarios, it was given me the wrong return value for the module function when it calls so to scale the functionality I implement filter based function.

I agreed that this code snippet makes more sense than the filter and it has less complex. I'm going to update it and mark this conversation as resolved. Thanks!

modules/images/webp-uploads/can-load.php
mukeshpanchal27: Thanks for the review! The proposed code changes are updated in PR.

load.php
mukeshpanchal27: Thanks for the feedback. Yes, we can consider it, here we have one more option like `extension_loaded` (PHP Core),

- `perlab_module_loaded`

@felixarntz do we need to change the function name? among these two option?

admin/load.php
mukeshpanchal27: > Why not to just disable such checkboxes using the `disabled` attribute? What is the point of having a checkbox to be readonly? This `onclick="return false;"` approach looks like a hack.

Thanks, Eugene for the concern.  If we use `disabled` then as Felix said it will not send the option when we submit the form and it also changes the checked state of the option to unchecked. Yes, It's a hack to stop the checkbox click and it will not change the check box state on submission.

@felixarntz if we go with a two checkbox approach can we go with these code changes? In the second checkbox, I removed the ID so we don't get a duplicate ID on the page.

    <input type="checkbox" id="<?php echo esc_attr( "{$base_id}_enabled" ); ?>" name="<?php echo esc_attr( "{$base_name}[enabled]" ); ?>" aria-describedby="<?php echo esc_attr( "{$base_id}_description" ); ?>" value="1"<?php checked( $enabled ); ?><?php echo ( ! $can_load_module ? ' disabled' : '' ); ?>>
    <?php
    if ( ! $can_load_module ) {
        ?>
        <input type="checkbox" name="<?php echo esc_attr( "{$base_name}[enabled]" ); ?>" aria-describedby="<?php echo esc_attr( "{$base_id}_description" ); ?>" value="1"<?php checked( $enabled ); ?>  readonly >
        <?php
    }

admin/load.php
eugene-manuilov: Ah.. ok, thanks for explaining, @felixarntz.

load.php
mukeshpanchal27: Thanks for the review and explanation, Felix.

`perflab_get_active_modules()` called in two functions `perflab_get_generator_content()` and `perflab_load_active_modules()` in which we get active module lists, so I update the `perflab_get_active_modules()` function code so we don't need to add duplicate code for both functions.

If we add a new function perflab_load_active_modules(), still we have to use the if statement in both functions to determine whether a module is active or not?

    if ( ! perflab_load_active_modules( $active_modules ) ) {
        return;
    }

Or should we `perflab_get_active_modules()` to `rename perflab_get_active_modules()` with a document that contains both logic in a single function?

load.php
mukeshpanchal27: If we check whether the module can be loaded here then we have to check similar in the `perflab_render_generator()` function.

admin/load.php
felixarntz: @mukeshpanchal27 I was rather thinking the second `input` would be of type `hidden` - no need to show a second checkbox.

Good point on not duplicating the ID. In addition to that, we can probably remove the `name` attribute from the first checkbox though since it won't do anything, and I have some concerns on a duplicate name attribute when it's only relevant for the second `hidden` input. So I'd advise to keep the `id` only on the (visible) checkbox and the `name` only on the `hidden` input.

load.php
felixarntz: @mukeshpanchal27 What about my idea of adding a `perflab_get_active_and_valid_modules()` function? That function could call `perflab_get_active_modules()` and then run the additional logic.

Then, all code that needs to get all active modules that also can actually be loaded can simply call `perflab_get_active_and_valid_modules()` while other code that only cares about what's actually configured as active can call `perflab_get_active_modules()`. This way there shouldn't be any duplicate code.

In addition, as I mentioned before that new function would also nicely mirror the WordPress plugin API in core, where there is a corresponding `wp_get_active_and_valid_plugins()` function.

load.php
felixarntz: @mukeshpanchal27 I think the current `perflab_can_load_module` name makes most sense especially _because_ it should not be limited to only the module already being in core. This should be a flexible function that can check _anything_ related to the requirements to loading a module. So I think we should keep this as is.

load.php
felixarntz: @mukeshpanchal27 Good point. However, then let's go with my above suggestion of adding a reusable function `perflab_get_active_and_valid_modules()` and have that logic in there. This way we can leave `perflab_get_active_modules()` untouched and with a single purpose.

felixarntz (commented):
@mukeshpanchal27 Thanks for the updates, the `perflab_can_load_module()` behavior looks good now.

I left some replies on the relevant comment threads above around how to handle the additional logic about whether a module can be loaded. TLDR I think we should add a new `perflab_get_active_and_valid_modules()` function which calls `perflab_get_active_modules()` and then further filters out any modules that cannot be loaded. All the additional checks can happen in that function, and most code that currently calls `perflab_get_active_modules()` should then call `perflab_get_active_and_valid_modules()` instead.

admin/load.php
mukeshpanchal27: Thanks. Changes applied on PR.

load.php
mukeshpanchal27: Perflab_get_active_and_valid_modules() is introduced with changes.

load.php
mukeshpanchal27: Perflab_get_active_and_valid_modules() is introduced with changes. Can you please review it? Thanks!

mukeshpanchal27:
@felixarntz Thanks for the review. I have implemented the review changes. Can you please review it again and let me know if it needs any changes?

admin/load.php
```diff
@@ -126,31 +126,44 @@ function perflab_render_modules_page() {
...
-			} else {
-				printf(
-					/* translators: %s: module name */
-					__( 'Enable %s', 'performance-lab' ),
-					esc_html( $module_data['name'] )
-				);
-			}
-			?>
+			<?php if ( ! $can_load_module ) { ?>
+				<input type="checkbox" id="<?php echo esc_attr( "{$base_id}_enabled" ); ?>" disabled>
```
felixarntz: Nit-pick, but this should probably also have the `aria-describedby` attribute. The only thing that _has_ to be omitted is the `name` attribute (and probably the `value` attribute).

admin/load.php
```diff
@@ -126,31 +126,44 @@ function perflab_render_modules_page() {
...
+				<input type="hidden" name="<?php echo esc_attr( "{$base_name}[enabled]" ); ?>">
```
felixarntz: This needs the `value` attribute now I think, but only _if_ that module is currently active via the option.

We need to make sure that a module that is configured as active in the DB remains active when updating any other modules on this page, even when it can't be loaded. The same for when a module is inactive, it should remain inactive.

For example like this:
```suggestion
				<input type="hidden" name="<?php echo esc_attr( "{$base_name}[enabled]" ); ?>" value="<?php echo $enabled ? '1' : ''; ?>">
```

load.php
```diff
@@ -137,6 +137,41 @@ function( $module_settings ) {
...
+		$can_load_module = perflab_can_load_module( $module );
+		if ( ! $can_load_module ) {
+			continue;
+		}
+
+		// Do not load module if no longer exists.
+		$module_file = plugin_dir_path( __FILE__ ) . 'modules/' . $module . '/load.php';
+		if ( ! file_exists( $module_file ) ) {
+			continue;
+		}
```
felixarntz: Nit-pick, but maybe let's put this conditional first in the `foreach` because it is faster and also more "essential" to check: If the module no longer exists, we already know there is no way that it can be loaded.

load.php
```diff
@@ -169,29 +204,52 @@ function perflab_render_generator() {
...
+	}
+
+	// Call the closure to determine whether the module can be loaded.
+	return $can_load();
+}
+
+/**
+ * Loads the active and valid performance modules.
  *
  * @since 1.0.0
```
felixarntz, line 226: Maybe let's add a doc-block since the rename here is somewhat significant.
```suggestion
 * @since 1.0.0
 * @since n.e.x.t Renamed to perflab_load_active_and_valid_modules().
```

tests/load-tests.php
```diff
@@ -121,14 +121,32 @@ function( $module_settings ) {
 	}
 
 	public function test_perflab_get_generator_content() {
-		// Assert that it returns the current version and active modules.
+		// Assert that it doesn't returns dummy modules.
 		$dummy_active_modules = array( 'images/a-module', 'object-cache/another-module' );
```
felixarntz: That's not the point of this test, this comment shouldn't be changed and the test still should ensure that it covers the previous comment.

We need to make sure of course that this test works with the new changes, but in order to do that we need to replace the usage of dummy modules here with real modules (for which the new can-load logic will work).

For example like this:
```suggestion
		// Assert that it returns the current version and active modules.
		// For this test, set the active modules to all defaults but the last one.
		$dummy_active_modules = require plugin_dir_path( PERFLAB_MAIN_FILE ) . 'default-enabled-modules.php';
		array_pop( $dummy_active_modules );
```

We should also rename `$dummy_active_modules` simply to something like `$active_modules`.

tests/load-tests.php
```diff
@@ -121,14 +121,32 @@ function( $module_settings ) {
...
+			'object-cache/persistent-object-cache-health-check' => array( 'enabled' => true ),
+		);
+		update_option( PERFLAB_MODULES_SETTING, $new_value );
+		$dummy_active_modules = array( 'images/webp-uploads', 'object-cache/persistent-object-cache-health-check' );
+		add_filter(
+			'perflab_active_modules',
+			function() use ( $dummy_active_modules ) {
+				return $dummy_active_modules;
+			}
+		);
```
felixarntz: This should be removed again, see above. This test should keep its original purpose, and in addition we shouldn't hard-code real modules into these generic tests as it makes them unstable.

felixarntz (commented):
@mukeshpanchal27 The production code looks great now, I only left some minor feedback on that.

The test change needs to be fixed though, you have changed the purpose of the test overall which shouldn't happen, we need to make sure the point of the test remains as before. I left 2 comments with more details on that.

One additional feedback here would be to please add tests for the new functions. We should preferably have test coverage for both `perflab_get_active_and_valid_modules()` and `perflab_can_load_module( $module )`. In order to write tests for those, maybe you can rely on some of the demo modules. For example if you use `../tests/testdata/demo-modules/javascript/demo-module-1` or `../tests/testdata/demo-modules/images/demo-module-3`, you should be able to rely on the files from those modules to run the code. You could add a demo `can-load.php` file to one of them and that way you could test both types of situations.

admin/load.php
mukeshpanchal27: Thanks! `aria-describedby` attribute added in checkbox input.

admin/load.php
mukeshpanchal27: Thanks. It makes sense to me. The review was implemented in PR.

load.php
mukeshpanchal27: I agree. Before it can be loaded, we should check if the module exists or not. It's minor, but it helps for faster execution. It was applied in PR. Thanks!

load.php
mukeshpanchal27: Doc-block added for function rename.

tests/load-tests.php
mukeshpanchal27: It was removed now.

tests/load-tests.php
mukeshpanchal27: Changes for the tests were reverted, and changes related to the active module were applied on PR. Thanks!

tests/load-tests.php
```diff
@@ -148,6 +150,59 @@ public function test_perflab_render_generator() {
...
+			);
+			$output = perflab_get_active_and_valid_modules();
+
+			// Remove Image module as it will marge in 6.1.
+			$remove_marge_modules = array_shift( $active_modules );
+			$this->assertSame( $active_modules, $output );
+		}
+	}
+
+	public function test_perflab_can_load_module() {
```
jjgrainger: @mukeshpanchal27 This test and the following `test_if_can_load_file_available_but_module_does_not_marge_in_core` look okay and tests the logic within the `perflab_can_load_module` function, but feels odd that it doesn't call the function directly.

I can see why as the `plugin_dir_path( __FILE__ )` used to require in the files is not filterable in WordPress, so it's currently not possible to use the test modules.

I'm wondering if there is a way we can either introduce a filter for the modules path in the plugin or provide a constant that we can overwrite during tests to make use of the test modules? I'm not sure though if this would be beneficial outside of the unit tests we're writing.

I would be interested in getting @felixarntz thoughts on this though to discuss a way forward.

jjgrainger (changes requested):
Thanks @mukeshpanchal27 added a comment related to a couple of the tests. It would be good to get @felixarntz input on this.

tests/load-tests.php
felixarntz: @mukeshpanchal27 I agree with @jjgrainger here, I think these tests don't really help since they don't test any part of the production code.

Let's see if we can find a way to test the actual `perflab_can_load_module()` function. Otherwise, we should remove these tests.

Have you tried the approach I suggested in https://github.com/WordPress/performance/pull/390#pullrequestreview-1023696565? I'm not 100% sure this will work, but wondering, can we rely on `..` in those paths to get to the correct test modules?

mukeshpanchal27: @jjgrainger Thanks for the feedback.

@felixarntz Yes, before I missed something on it and I can't figure it out, but again I reviewed it and it works fine with the `perflab_can_load_module()` function test. The `..` approach is working.

I have update unit test for `perflab_can_load_module()` for review.

jjgrainger (approved):
Thanks @mukeshpanchal27 test look good now 👍

admin/load.php
```diff
@@ -126,31 +126,44 @@ function perflab_render_modules_page() {
...
-				);
-			} else {
-				printf(
-					/* translators: %s: module name */
-					__( 'Enable %s', 'performance-lab' ),
-					esc_html( $module_data['name'] )
-				);
-			}
-			?>
+			<?php if ( ! $can_load_module ) { ?>
```
mitogh: Personal preference but swaping the conditions removes the need to use `! $can_load_module`.

For instance: 


```php
if ( $can_load_module ) {
 // Swap with section B
} else {
 // Swap with section A
}

File: admin/load.php [hunk 27]
```diff
@@ -126,31 +126,44 @@ function perflab_render_modules_page() {
...
+				<input type="checkbox" id="<?php echo esc_attr( "{$base_id}_enabled" ); ?>" aria-describedby="<?php echo esc_attr( "{$base_id}_description" ); ?>" disabled>
+				<input type="hidden" name="<?php echo esc_attr( "{$base_name}[enabled]" ); ?>" value="<?php echo $enabled ? '1' : ''; ?>">
```
mitogh: Would be worth fallback to `0` instead of an empty string as the value?

load.php
```diff
@@ -137,6 +137,41 @@ function( $module_settings ) {
...
 
+/**
+ * Gets the active and valid performance modules.
+ *
+ * @since n.e.x.t
+ *
+ * @return array List of active valid module slugs.
+ */
+function perflab_get_active_and_valid_modules() {
+	$active_modules = perflab_get_active_modules();
```
mitogh: Suggestion accepting an array as parameters here removes the need that this function those 2 things get active modules and filter to valid ones instead my suggestion would be to accept an array of modules and focus this function on only returning the active ones based on the provided parameter, for instance:


```suggestion
function perflab_get_valid_modules( array $modules = array() ) {
```

That way this function is decoupled from the need to be linked to the `active` modules and you can pass any value in your test to verify the behavior is independent from the active modules.

load.php
```diff
@@ -137,6 +137,41 @@ function( $module_settings ) {
...
+
+	if ( ! $active_modules ) {
```
mitogh:
```suggestion
	if ( ! is_array( $active_modules ) ) {
```

load.php
```diff
@@ -137,6 +137,41 @@ function( $module_settings ) {
...
+		return array();
+	}
+
+	$active_and_valid_modules = array();
```
mitogh:
```suggestion
	$valid_modules = array();
```

load.php
```diff
@@ -137,6 +137,41 @@ function( $module_settings ) {
...
+
+		// Do not load module if no longer exists.
+		$module_file = plugin_dir_path( __FILE__ ) . 'modules/' . $module . '/load.php';
+		if ( ! file_exists( $module_file ) ) {
+			continue;
+		}
+
+		// Do not load module if it cannot be loaded, e.g. if it was already merged and is available in WordPress core.
+		$can_load_module = perflab_can_load_module( $module );
+		if ( ! $can_load_module ) {
```
mitogh:
```suggestion
		if ( ! $perflab_can_load_module( $module ) ) {
```

load.php
```diff
@@ -145,12 +180,12 @@ function( $module_settings ) {
  * @since 1.1.0
  */
 function perflab_get_generator_content() {
-	$active_modules = perflab_get_active_modules();
+	$active_and_valid_modules = perflab_get_active_and_valid_modules();
```
mitogh:
```suggestion
	$active_and_valid_modules = perflab_get_valid_modules( perflab_get_active_modules() );
```

load.php
```diff
@@ -169,29 +204,53 @@ function perflab_render_generator() {
...
+function perflab_can_load_module( $module ) {
+	$module_load_file = plugin_dir_path( __FILE__ ) . 'modules/' . $module . '/can-load.php';
+
+	// If the `can-load.php` file does not exist, assume the module can be loaded.
+	if ( ! file_exists( $module_load_file ) ) {
+		return true;
+	}
+
+	// Require the file to get the closure for whether the module can load.
+	$can_load = require $module_load_file;
```
mitogh:
```suggestion
	$module = require $module_load_file;
```

load.php
```diff
@@ -169,29 +204,53 @@ function perflab_render_generator() {
...
+
+	// If the `can-load.php` file is invalid and does not return a closure, assume the module can be loaded.
+	if ( ! is_callable( $can_load ) ) {
```
mitogh:
```suggestion
	if ( ! is_callable( $module ) ) {
```

load.php
```diff
@@ -169,29 +204,53 @@ function perflab_render_generator() {
...
+		return true;
+	}
+
+	// Call the closure to determine whether the module can be loaded.
+	return $can_load();
```
mitogh:
```suggestion
	return (bool) $module();
```

tests/testdata/demo-modules/javascript/demo-module-1/can-load.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * Can load function to determine if module already marge in WordPress core.
+ *
+ * @since   n.e.x.t
+ * @package performance-lab
+ */
+
+return function() {
```
mitogh: Looks like this file should be renamed to `can-not-load`

tests/testdata/demo-modules/images/demo-module-3/can-load.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * Can load function to determine if module already marge in WordPress core.
+ *
+ * @since   n.e.x.t
+ * @package performance-lab
+ */
+
+return function() {
```
mitogh: You can replace this with:

```suggestion
return __return_true;
```

tests/testdata/demo-modules/javascript/demo-module-1/can-load.php
mitogh: You can replace this with: 
```suggestion
return __return_false;
```

admin/load.php
mukeshpanchal27: These changes look good to me. PR updated according to it.

admin/load.php
mukeshpanchal27: agreed. For now, I have updated PR accordingly.

@felixarntz Do we need to set an empty value here?

load.php
mukeshpanchal27: LGTM.

load.php
mukeshpanchal27: Thanks for the feedback. Similar changes are made in the plugin's admin file `load.php`.

load.php
mukeshpanchal27: Thanks for the name suggestion; it makes sense.

load.php
mukeshpanchal27: Variables are updated according to changes.

tests/testdata/demo-modules/images/demo-module-3/can-load.php
mukeshpanchal27: Sorry, I missed it completely. Thanks for pointing it out.

tests/testdata/demo-modules/javascript/demo-module-1/can-load.php
mukeshpanchal27: Sorry, I missed it completely. Thanks for pointing it out. 

> Looks like this file should be renamed to `can-not-load`

According to https://github.com/WordPress/performance/blob/fix/293-mechanism-not-load-module/load.php#L214, if we change the file name, we can't test whether that module is marge or not in the core.

load.php
mukeshpanchal27: Thanks for the feedback.

Yes, this change will help with the unit tests. The PR was updated with the changes.

mitogh (approved):
Great work 👍

mukeshpanchal27:
The PR needs a final review from @felixarntz.

adamsilverstein (commented):
Nice work! LGTM pending final review from Felix.

load.php
```diff
@@ -137,6 +137,40 @@ function( $module_settings ) {
...
 
+/**
+ * Gets the active and valid performance modules.
+ *
+ * @since n.e.x.t
+ *
+ * @param array $modules List of active module slugs.
+ * @return array List of active valid module slugs.
+ */
+function perflab_get_valid_modules( array $modules = array() ) {
```
felixarntz: Two things about this function:
* I would say the `get` in the name is a bit confusing, since it _filters_ an existing input variable rather than getting a distinct value from somewhere. So I think a better name would be `perflab_filter_valid_modules()`. Or, we change this implementation to just be for a single module, e.g. `perflab_is_valid_module( $module )`, and then you could call it via `array_filter( perflab_get_active_modules(), 'perflab_is_valid_module' )`.
* Why is there a default? I would say this should rather be a required parameter, especially since calling it without any value is pointless since it would always just return the empty array then.

tests/load-tests.php
```diff
@@ -148,6 +150,52 @@ public function test_perflab_render_generator() {
 		$this->assertContains( $expected, $output );
 	}
 
+	public function test_perflab_get_valid_modules() {
+		global $wp_version;
+
+		// Assert that it returns the empty array for dummy modules.
```
felixarntz: This comment isn't accurate any more.

tests/load-tests.php
```diff
@@ -148,6 +150,52 @@ public function test_perflab_render_generator() {
...
+		$dummy_active_modules = array(
+			'../tests/testdata/demo-modules/javascript/demo-module-1',
+			'../tests/testdata/demo-modules/something/demo-module-2',
+			'../tests/testdata/demo-modules/images/demo-module-3',
+		);
+
+		$output = perflab_get_valid_modules( $dummy_active_modules );
+		$this->assertIsArray( $output );
```
felixarntz: This is redundant since right afterwards there is the `assertSame` anyway.

tests/load-tests.php
```diff
@@ -148,6 +150,52 @@ public function test_perflab_render_generator() {
...
+		$this->assertSame( array( '../tests/testdata/demo-modules/something/demo-module-2', '../tests/testdata/demo-modules/images/demo-module-3' ), $output );
```
felixarntz: Nit-pick, but can you put this array into a variable (e.g. `$expected`) so that it's easier to read, and put that next to the definition of the above `$dummy_active_modules`?

tests/load-tests.php
```diff
@@ -148,6 +150,52 @@ public function test_perflab_render_generator() {
...
+				function() use ( $active_modules ) {
+					return $active_modules;
+				}
+			);
+			$output = perflab_get_valid_modules( perflab_get_active_modules() );
+
+			// Remove Image module as it will marge in 6.1.
+			$remove_marge_modules = array_shift( $active_modules );
+			$this->assertSame( $active_modules, $output );
+		}
```
felixarntz: I think this would be better to remove. This is not a test for the infrastructure here, it's a test for whether the WebP Uploads module-specific logic in its `can-load.php` file behaves as expected.

So either we should migrate this test into a very simple test as part of the WebP Uploads module (only `require` the file and check that the closure function does the right thing) or remove it. It shouldn't be part of the infrastructure tests, since that file is not part of the infrastructure, but of the WebP Uploads module.

tests/load-tests.php
```diff
@@ -148,6 +150,52 @@ public function test_perflab_render_generator() {
...
+		$demo_modules = array(
+			'javascript/demo-module-1' => false,
+			'something/demo-module-2'  => true,
+			'images/demo-module-3'     => true,
+		);
+
+		foreach ( $demo_modules as $module => $can_load ) {
+			$output = perflab_can_load_module( '../tests/testdata/demo-modules/' . $module );
+			$this->assertSame( $can_load, $output );
+		}
```
felixarntz: Nit-pick, but this is essentially a `foreach` loop running the same test for different sets of data. For this purpose, a data provider method should be used, and then the test method body should be only what you have here within the `foreach`.

You can set that up via `@dataProvider` annotations. There are several existing data provider usages in the plugin tests or in WordPress core, so you can probably look at those for reference.

felixarntz (commented):
@mukeshpanchal27 Looks good now, mostly a few last nitpicks and minor things to improve around the tests.

load.php
mukeshpanchal27: Thanks for the feedback.

I like the `perflab_is_valid_module( $module )` approach instead of the function name change. Now we should have passed a valid module slug. I have updated unit tests according to these changes.

tests/load-tests.php
mukeshpanchal27: Comment removed.

tests/load-tests.php
mukeshpanchal27: Thanks for the review. I have updated the unit tests and removed this check.

tests/load-tests.php
mukeshpanchal27: I have updated the unit test according to the function change. For this, we use a single module slug in the function, so we don't need it, and we use `@dataProvider` annotations for the dummy module list.

tests/load-tests.php
mukeshpanchal27: I have removed these checks because we have already tested if the module is valid or not in the test.

tests/load-tests.php
mukeshpanchal27: Thanks for the suggestion for `@dataProvider` annotations. I have added it and updated the tests accordingly.

mukeshpanchal27:
Thanks for the feedback @felixarntz, I have addressed the feedback and updated the PR and unit tests accordingly.

load.php
```diff
@@ -137,6 +137,34 @@ function( $module_settings ) {
 	return $modules;
 }
 
+/**
+ * Gets the active and valid performance modules.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $module Slug of the module.
+ * @return null|string Module slug name on success, null on failure.
```
felixarntz: This is a bit odd of a return value, I think this should return a boolean (`false` or `true`) since it is an "is ..." check. Since the developer passes `$module` into the function anyway, no point to return that same value.

tests/load-tests.php
```diff
@@ -148,6 +151,42 @@ public function test_perflab_render_generator() {
...
+	}
+
+	/**
+	 * @dataProvider provider_dummy_valid_modules
+	 */
+	public function test_perflab_is_valid_module( $dummy_module ) {
+		$output = perflab_is_valid_module( $dummy_module );
+		$this->assertNotEmpty( $output );
+		$this->assertIsString( $output );
+	}
```
felixarntz: See above, these assertions should be using `assertFalse` or `assertTrue` based on the alternate return type.

Also, it would be more efficient to combine these two tests into one. Both could use the same `@dataProvider`, you could include the empty string in there, and then every data set could also include the expected `true` or `false`. You could use `assertSame` for that, similarly to how you have it below for the `perflab_can_load_module` tests.

Also it would be great if you could use the same method name for the data provider method, except replacing `test_` with `data_` (see other comment below).

tests/load-tests.php
```diff
@@ -148,6 +151,42 @@ public function test_perflab_render_generator() {
...
+
+	/**
+	 * @dataProvider provider_dummy_can_load_modules
+	 */
+	public function test_perflab_can_load_module( $dummy_can_load_modules, $module_status ) {
+		$this->assertSame( $module_status, perflab_can_load_module( $dummy_can_load_modules ) );
+	}
+
+	public function provider_dummy_can_load_modules() {
```
felixarntz: Nit-pick, but the WP core convention is to name the data provider method similarly to the test method it belongs to (unless it belongs to multiple test methods, which is not the case here though). The only difference should be to use `data_` instead of the `test_` prefix.

Also some of the variable names could be clarified.
```suggestion
	/**
	 * @dataProvider data_perflab_can_load_module
	 */
	public function test_perflab_can_load_module( $dummy_module, $expected_status ) {
		$this->assertSame( $expected_status, perflab_can_load_module( $dummy_module ) );
	}

	public function data_perflab_can_load_module() {
```

felixarntz (commented):
@mukeshpanchal27 This is looking good now, thank you for the updates!

I left a few last suggestions to further improve the tests, but this is pretty much good to go.

load.php
mukeshpanchal27: Thanks for the feedback. The PR was updated to reflect these changes.

tests/load-tests.php
mukeshpanchal27: Thanks for the feedback. I removed the empty module test and merged it into `test_perflab_is_valid_module` so it tests all scenarios. For this test, I use `@dataProvider data_perflab_is_valid_module` because we can't test an empty module in `test_perflab_can_load_module` test.

tests/load-tests.php
mukeshpanchal27: Thanks for the feedback. The PR was updated to reflect these changes.

load.php
```diff
@@ -137,6 +137,34 @@ function( $module_settings ) {
...
+	if ( ! file_exists( $module_file ) ) {
+		return false;
+	}
+
+	// Do not load module if it cannot be loaded, e.g. if it was already merged and is available in WordPress core.
+	if ( ! perflab_can_load_module( $module ) ) {
+		return false;
+	}
+
+	return true;
```
felixarntz: This could be shortened to the following:
```suggestion
	return perflab_can_load_module( $module );
```

Not a big deal though.

tests/load-tests.php
```diff
@@ -148,6 +151,37 @@ public function test_perflab_render_generator() {
...
+	public function test_perflab_is_valid_module( $dummy_module, $expected_status ) {
+		$this->assertSame( $expected_status, perflab_is_valid_module( $dummy_module ) );
+	}
+
+	public function data_perflab_is_valid_module() {
+		return array(
+			array( '', false ),
+			array( '../tests/testdata/demo-modules/javascript/demo-module-1', false ),
+			array( '../tests/testdata/demo-modules/something/demo-module-2', true ),
+			array( '../tests/testdata/demo-modules/images/demo-module-3', true ),
```
felixarntz, line 167: With the empty module you cover the first condition of `perflab_is_valid_module`, with the other three modules here you cover the last condition that checks `perflab_can_load_module`. It would be good to add one more test case here for a module which actually does not exist, to also cover the remaining condition that has the `file_exists` check, which currently isn't covered yet.

felixarntz (approved):
@mukeshpanchal27 Two super-minor nit-picks, but this is looking very solid now. Great work!

I'll just add those quickly, it's basically good to merge.

PR #393, opened 2022-06-22
Ensure incorrect usage of `webp_uploads_upload_image_mime_transforms` filter is treated correctly

### LINKED ISSUE #361: Optimize: `webp_uploads_upload_image_mime_transforms` filter
## Bug Description

As we have provided filter `webp_uploads_upload_image_mime_transforms` for the mime type transforms in any case user does not want to transform the mime type then they just return `empty`, `true/false` or `array()` then this file [helper.php](https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/helper.php#L38-L40) code is not require because `$transforms = (array) apply_filters( 'webp_uploads_upload_image_mime_transforms', $default_transforms );` filter always retunt array value.

**Return type empty**

		return;
	});

**Return type true/false**

		return false;
	});

**Return type empty array:**

		return array();
	});

Also, we have to check if an array is null and then return an empty array so it can't execute next [foreach](https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/helper.php#L44-L48) code.

## Additional Context

 - PHP Version: 8.0.15
 - OS: Windows
 - Browser: Chrome
 - Plugin Version: 1.1.0

jjgrainger:
## Summary

Fixes #361

## Relevant technical choices

* Removes casting the value returned by the `webp_uploads_upload_image_mime_transforms` to an array.
* Add tests to confirm behaviour works as expected based on `webp_uploads_get_upload_image_mime_transforms` comments.

## Checklist

### DIFF
--- a/modules/images/webp-uploads/helper.php
+++ b/modules/images/webp-uploads/helper.php
@@ -32,7 +32,7 @@ function webp_uploads_get_upload_image_mime_transforms() {
 	 *
 	 * @param array $default_transforms A map with the valid mime transforms.
 	 */
-	$transforms = (array) apply_filters( 'webp_uploads_upload_image_mime_transforms', $default_transforms );
+	$transforms = apply_filters( 'webp_uploads_upload_image_mime_transforms', $default_transforms );
 
 	// Return the default mime transforms if a non-array result is returned from the filter.
 	if ( ! is_array( $transforms ) ) {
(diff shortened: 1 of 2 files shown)


felixarntz (approved):
@jjgrainger Looks good, +1 for the extensive test coverage!

eugene-manuilov (approved):
LGTM 👍

mitogh (approved):

PR #401, opened 2022-06-28
Fix dominant color upload process to not override potential third-party editors

### LINKED ISSUE #374: Dominant color upload process ignores potential third-party editors
## Bug Description

Follow-up to #19 / #282: The current Dominant Color module enforces usage of its extended `WP_Image_Editor` classes. While this works with WordPress core alone, it is problematic in environments where e.g. a plugin has already added its own extended `WP_Image_Editor` classes.

We need to make sure that the Dominant Color implementation doesn't conflict with such plugins. The `dominant_color_set_image_editors()` filter callback function should be fixed to not override the entire array of editor class names, but rather only replace the specific WordPress core editors `WP_Image_Editor_GD` and `WP_Image_Editor_Imagick` with their customized variants from the module.

cc @pbearne @spacedmonkey

spacedmonkey:
## Summary

Replace editor in a cleaner way.

Fixes #374

## Checklist

### DIFF
--- a/modules/images/dominant-color/load.php
+++ b/modules/images/dominant-color/load.php
@@ -235,6 +235,8 @@ add_filter( 'wp_enqueue_scripts', 'dominant_color_add_inline_style' );
  *
  * @since 1.2.0
  *
+ * @param string[] $editors Array of available image editor class names. Defaults are 'WP_Image_Editor_Imagick', 'WP_Image_Editor_GD'.
+ *
  * @return string[] Registered image editors class names.
  */
 function dominant_color_set_image_editors( $editors ) {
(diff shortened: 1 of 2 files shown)


pbearne (approved):
nice

modules/images/dominant-color/load.php
```diff
@@ -235,17 +235,31 @@ function dominant_color_add_inline_style() {
  *
  * @since 1.2.0
  *
+ * @param string[] $editors Array of available image editor class names. Defaults are 'WP_Image_Editor_Imagick', 'WP_Image_Editor_GD'.
+ *
```
felixarntz:
```suggestion
 * @param string[] $editors Array of available image editor class names. Defaults are 'WP_Image_Editor_Imagick', 'WP_Image_Editor_GD'.
```

felixarntz (commented):
@spacedmonkey Thanks, production code looks good, just one tiny nit-pick.

There are unit test failures though, could you check and address those?

spacedmonkey:
> There are unit test failures though, could you check and address those?

Tests are failing because https://github.com/WordPress/performance/issues/405. This is because ImageMagik is flagging all gifs as transparent. I have put a workaround in for now and we can tackle that issue there.

adamsilverstein (approved):
Code looks good to me!

PR #411, opened 2022-07-01
Remove additional image backup sources & sizes files when attachment deleted

### LINKED ISSUE #375: WebP images not deleted when saved in backend after edit
## Bug Description

WebP images that are generated when uploaded and edited and saved. When deleted the image, the WebP versions are not being removed from files.

## Steps to reproduce

1. Go to Media, Upload the image
2. Open the image in backend image editor and rotate it and save the image.
3. Delete the image completely and check the files.
4. See that original WebP version of the files are not removed.

## Additional Context

 - PHP Version: 7.4
 - OS: MacOS
 - Browser: Chrome
 - Plugin Version: 1.1.0
 - Device: MacBook Pro

mehulkaklotar:
## Summary

Fixes #375 

## Checklist

### DIFF
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -387,6 +387,30 @@ function webp_uploads_remove_sources_files( $attachment_id ) {
 			wp_delete_file_from_directory( $backup_intermediate_file, $intermediate_dir );
 		}
 	}
+
+	$backup_sources = get_post_meta( $attachment_id, '_wp_attachment_backup_sources', true );
+	$backup_sources = is_array( $backup_sources ) ? $backup_sources : array();
+
+	// Delete full sizes backup mime types.
+	foreach ( $backup_sources as $backup_mimes ) {
+
+		foreach ( $backup_mimes as $backup_mime_properties ) {
+			if ( ! is_array( $backup_mime_properties ) || empty( $backup_mime_properties['file'] ) ) {
+				continue;
+			}
+
+			$full_size = str_replace( $basename, $backup_mime_properties['file'], $file );
+			if ( empty( $full_size ) ) {
+				continue;
+			}
+
+			$full_size_file = path_join( $upload_path['basedir'], $full_size );
+			if ( ! file_exists( $full_size_file ) ) {
+				continue;
+			}
+			wp_delete_file_from_directory( $full_size_file, $intermediate_dir );
+		}
+	}
 }
 add_action( 'delete_attachment', 'webp_uploads_remove_sources_files', 10, 1 );
 
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -347,6 +347,46 @@ function webp_uploads_remove_sources_files( $attachment_id ) {
 		}
 		wp_delete_file_from_directory( $full_size_file, $intermediate_dir );
 	}
+
+	$backup_sizes = get_post_meta( $attachment_id, '_wp_attachment_backup_sizes', true );
+	$backup_sizes = is_array( $backup_sizes ) ? $backup_sizes : array();
+
+	foreach ( $backup_sizes as $backup_size ) {
+		if ( ! isset( $backup_size['sources'] ) || ! is_array( $backup_size['sources'] ) ) {
+			continue;
+		}
+
+		$original_backup_size_mime = empty( $backup_size['mime-type'] ) ? '' : $backup_size['mime-type'];
+
+		foreach ( $backup_size['sources'] as $backup_mime => $backup_properties ) {
+			/**
+			 * When we face the same mime type as the original image, we ignore this file as this file
+			 * would be removed when the size is removed by WordPress itself. The meta information as well
+			 * would be deleted as soon as the image is removed.
+			 *
+			 * @see wp_delete_attachment
+			 */
+			if ( $original_backup_size_mime === $backup_mime ) {
+				continue;
+			}
+
+			if ( ! is_array( $backup_properties ) || empty( $backup_properties['file'] ) ) {
+				continue;
+			}
+
+			$backup_intermediate_file = str_replace( $basename, $backup_properties['file'], $file );
+			if ( empty( $backup_intermediate_file ) ) {
+				continue;
+			}
+
+			$backup_intermediate_file = path_join( $upload_path['basedir'], $backup_intermediate_file );
+			if ( ! file_exists( $backup_intermediate_file ) ) {
+				continue;
+			}
+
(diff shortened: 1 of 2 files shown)


mitogh (approved):
Changes look solid to me (great work). Question looking into these changes, are we expecting to delete the meta fields as well created by the plugin if the image is deleted such as: 

- `_wp_attachment_backup_sources `

And any other custom meta associated to an image, or those are deleted when the attachment ID is effectively deleted from within the DB?

mehulkaklotar:
@mitogh Meta fields are deleted when attachment ID is deleted. so we don't need to do anything to delete them.

mitogh:
Thanks @mehulkaklotar makes sense.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -311,6 +311,43 @@ public function it_should_remove_full_size_images_when_no_size_image_exists() {
...
+		wp_delete_attachment( $attachment_id, true );
+
+		$this->assertFileDoesNotExist( path_join( $dirname, $backup_sources['full-orig']['image/webp']['file'] ) );
+		$this->assertFileDoesNotExist( path_join( $dirname, $backup_sizes['thumbnail-orig']['sources']['image/webp']['file'] ) );
+
+		$backup_sources = get_post_meta( $attachment_id, '_wp_attachment_backup_sources', true );
+		$this->assertEmpty( $backup_sources );
+
+		$backup_sizes = get_post_meta( $attachment_id, '_wp_attachment_backup_sizes', true );
+		$this->assertEmpty( $backup_sizes );
```
mukeshpanchal27: The PR looks good to me.

One question for the unit test: When the attachment is deleted using the `wp_delete_attachment` function, it also removes the meta associated with it, so it's worthwhile to check the post meta after it. What do you think?

mehulkaklotar: Yes, @mukeshpanchal27 We are checking after deleting attachment here that post meta is not found by asserting it is empty.

modules/images/webp-uploads/load.php
```diff
@@ -347,6 +347,70 @@ function webp_uploads_remove_sources_files( $attachment_id ) {
...
+			}
+
+			wp_delete_file_from_directory( $backup_intermediate_file, $intermediate_dir );
+		}
+	}
+
+	$backup_sources = get_post_meta( $attachment_id, '_wp_attachment_backup_sources', true );
+	$backup_sources = is_array( $backup_sources ) ? $backup_sources : array();
+
+	// Delete full sizes backup mime types.
```
adamsilverstein, line 394: Have you verified this works correctly with the "-scaled" version created when uploading large images?

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -311,6 +311,43 @@ public function it_should_remove_full_size_images_when_no_size_image_exists() {
 		$this->assertFileDoesNotExist( path_join( $dirname, $metadata['sources']['image/jpeg']['file'] ) );
 	}
 
+	/**
+	 * Remove the attached WebP version if the attachment is force deleted but empty trash day is not defined
```
adamsilverstein: comment should end in a period

tests/modules/images/webp-uploads/load-tests.php
adamsilverstein: this currently is only checking that the backup meta is empty, not that the original attachment meta is deleted.

tests/modules/images/webp-uploads/load-tests.php
adamsilverstein: What do you mean by "empty trash day is not defined", would the test look different if that was not the case?

adamsilverstein (commented):
Overall looks good - left some questions. I'm going to test this locally to verify the data handling works as expected as well as verifying -scaled or -rotated image handling is correct

modules/images/webp-uploads/load.php
mehulkaklotar: Yes @adamsilverstein I tested and `-scaled` version of the large image is also removed.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: That was mistakenly added in the function comment. I have fixed the comment now.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: we only add these two backup meta from the plugin, so we just need to check if these two meta is deleted. To check them deleted or not, we need to assert that it is empty when we get them using `get_post_meta` function which returns an empty string if a valid but non-existing post ID is passed.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -311,6 +311,43 @@ public function it_should_remove_full_size_images_when_no_size_image_exists() {
...
+		$this->assertFileExists( $file );
+
+		$editor = new WP_Image_Edit( $attachment_id );
+		$editor->rotate_right()->save();
+
+		$backup_sources = get_post_meta( $attachment_id, '_wp_attachment_backup_sources', true );
+		$this->assertNotEmpty( $backup_sources );
+
+		$backup_sizes = get_post_meta( $attachment_id, '_wp_attachment_backup_sizes', true );
+		$this->assertIsArray( $backup_sizes );
```
felixarntz: Why not use the same assertions for both? Shouldn't they both be non-empty arrays? Not a big deal, but this just struck me as a bit odd.

tests/modules/images/webp-uploads/load-tests.php
felixarntz: I personally think we should remove the checks for the post meta being empty. The post metadata is being deleted by core, so what we're testing with those two assertions is core, not our own logic. I guess it doesn't hurt having them, but if we have them here, we may also add a ton more for core tests - I think that's a bit beyond the point of unit tests, they should be for our plugin code, not the core code we're integrating with.

felixarntz (commented):
@mehulkaklotar Mostly looks great! I left just a few minor comments.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: Yes, right, added both assertions for not empty and is array.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: Removed the post meta assertions for the plugin tests.

felixarntz (approved):
Thanks @mehulkaklotar, this looks great!

PR #415, opened 2022-07-03
Allow control for which image sizes to generate additional MIME type versions

### LINKED ISSUE #413: Allow control for which sizes to generate a WebP version
Related to #372, not all sizes should for sure have WebP / additional MIME type versions available. In addition to the checks from #372, developers should have control over which image sizes to generate additional MIME type versions for.

As outlined in https://make.wordpress.org/core/2022/06/30/plan-for-adding-webp-multiple-mime-support-for-images/, this should behave as follows eventually:
* There should be a filter like `wp_image_sizes_with_additional_mime_type_support`, that filters the array/list of image sizes that should receive additional MIME types.
* WP core's `add_image_size()` function should receive an additional optional parameter e.g. `$provide_additional_mime_types` which will be `false` by default.
* The above filter will rely on that new image size registration value for its default / filter input.
* The WP core sizes will have this value set to `true` so that they are enabled by default. Custom sizes will have to opt in if they are used in the frontend, or opt out if there are used in special scenarios where additional MIME types are irrelevant or not supported (e.g. Open Graph tags).

For the plugin, we shouldn't bother updating the `add_image_size()` function since we can't modify it from there. So the plugin implementation should be mostly focused on the filter:
* Implement a function like `webp_uploads_get_image_sizes_with_additional_mime_type_support()`.
* Hard-code an array of the default core sizes `thumbnail`, `medium`, `medium_large`, `large`, `post-thumbnail`.
* Call `wp_get_additional_image_sizes()` add check for each image size in there whether it has a key `provide_additional_mime_types` set, with a truthy value. If so, include the size in the array.
* Run the array through a new `webp_uploads_image_sizes_with_additional_mime_type_support` filter and then return it.
* In the code where additional MIME types are generated for the different image sizes, call the new function. Then ensure that only the sizes which are also in this list are covered with generating additional MIME type versions.

eugene-manuilov:
## Summary

Fixes #413

## Checklist

### DIFF
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -599,10 +599,7 @@ function webp_uploads_get_image_sizes_with_additional_mime_type_support() {
 	);
 
 	foreach ( $additional_sizes as $size => $size_details ) {
-		if (
-			isset( $size_details['provide_additional_mime_types'] ) &&
-			filter_var( $size_details['provide_additional_mime_types'], FILTER_VALIDATE_BOOLEAN )
-		) {
+		if ( ! empty( $size_details['provide_additional_mime_types'] ) ) {
 			$allowed_sizes[ $size ] = $size;
 		}
 	}
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -654,25 +654,11 @@ add_filter( 'post_thumbnail_html', 'webp_uploads_update_featured_image', 10, 3 )
 function webp_uploads_get_image_sizes_additional_mime_type_support() {
 	$additional_sizes = wp_get_additional_image_sizes();
 	$allowed_sizes    = array(
-		'thumbnail'                        => true,
-		'medium'                           => true,
-		'medium_large'                     => true,
-		'large'                            => true,
-		'post-thumbnail'                   => true,
-		// 2x medium_large size.
-		'1536x1536'                        => true,
-		// 2x large size.
-		'2048x2048'                        => true,
-		// Twentyeleven theme.
-		'large-feature'                    => true,
-		'small-feature'                    => true,
-		// Twentyfourteen theme.
-		'twentyfourteen-full-width'        => true,
-		// Twentyseventeen theme.
-		'twentyseventeen-featured-image'   => true,
-		'twentyseventeen-thumbnail-avatar' => true,
-		// Twentytwenty theme.
-		'twentytwenty-fullscreen'          => true,
+		'thumbnail'      => true,
+		'medium'         => true,
+		'medium_large'   => true,
+		'large'          => true,
+		'post-thumbnail' => true,
 	);
 
 	foreach ( $additional_sizes as $size => $size_details ) {
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -105,11 +105,11 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 		return $metadata;
 	}
 
-	$sizes_with_mime_type_support = webp_uploads_get_image_sizes_with_additional_mime_type_support();
+	$sizes_with_mime_type_support = webp_uploads_get_image_sizes_additional_mime_type_support();
 
 	foreach ( $metadata['sizes'] as $size_name => $properties ) {
 		// Do nothing if this image size is not an array or is not allowed to have additional mime types.
-		if ( ! is_array( $properties ) || ! isset( $sizes_with_mime_type_support[ $size_name ] ) ) {
+		if ( ! is_array( $properties ) || empty( $sizes_with_mime_type_support[ $size_name ] ) ) {
 			continue;
 		}
 
@@ -641,39 +641,50 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
 add_filter( 'post_thumbnail_html', 'webp_uploads_update_featured_image', 10, 3 );
 
 /**
- * Returns an array of image sizes that can have alternative mime types. By default, only core sizes
- * are allowed and will be returned.
+ * Returns an array of image size names that have secondary mime type output enabled. Core sizes and
+ * core theme sizes are enabled by default.
  *
- * If a custom size wants to have additional mime types support, they will need to have
- * the provide_additional_mime_types property in the size details or it should be added using the
+ * Developers can control the generation of additional mime images for all sizes using the
  * webp_uploads_image_sizes_with_additional_mime_type_support filter.
  *
  * @since n.e.x.t
  *
  * @return array An array of image sizes that can have additional mime types.
  */
-function webp_uploads_get_image_sizes_with_additional_mime_type_support() {
+function webp_uploads_get_image_sizes_additional_mime_type_support() {
 	$additional_sizes = wp_get_additional_image_sizes();
 	$allowed_sizes    = array(
-		'thumbnail'      => 'thumbnail',
-		'medium'         => 'medium',
-		'medium_large'   => 'medium_large',
-		'large'          => 'large',
-		'post-thumbnail' => 'post-thumbnail',
+		'thumbnail'                        => true,
+		'medium'                           => true,
+		'medium_large'                     => true,
+		'large'                            => true,
+		'post-thumbnail'                   => true,
+		// 2x medium_large size.
+		'1536x1536'                        => true,
+		// 2x large size.
+		'2048x2048'                        => true,
+		// Twentyeleven theme.
+		'large-feature'                    => true,
+		'small-feature'                    => true,
+		// Twentyfourteen theme.
+		'twentyfourteen-full-width'        => true,
+		// Twentyseventeen theme.
+		'twentyseventeen-featured-image'   => true,
+		'twentyseventeen-thumbnail-avatar' => true,
+		// Twentytwenty theme.
+		'twentytwenty-fullscreen'          => true,
 	);
 
 	foreach ( $additional_sizes as $size => $size_details ) {
-		if ( ! empty( $size_details['provide_additional_mime_types'] ) ) {
-			$allowed_sizes[ $size ] = $size;
-		}
+		$allowed_sizes[ $size ] = ! empty( $size_details['provide_additional_mime_types'] );
 	}
 
 	/**
-	 * Filters image sizes that allowed to have additional mime types.
+	 * Filters whether additional mime types are allowed for image sizes.
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @param array $allowed_image_sizes An array of image sizes that allowed to have additional mime types.
+	 * @param array $allowed_sizes A map of image size names and whether they are allowed to have additional mime types.
 	 */
 	$allowed_sizes = apply_filters( 'webp_uploads_image_sizes_with_additional_mime_type_support', $allowed_sizes );
 
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -573,3 +573,40 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
 	return webp_uploads_img_tag_update_mime_type( $html, 'post_thumbnail_html', $attachment_id );
 }
 add_filter( 'post_thumbnail_html', 'webp_uploads_update_featured_image', 10, 3 );
+
+/**
+ * Returns an array of image sizes that can have alternative mime types. By default, only core sizes
+ * are allowed and will be returned.
+ *
+ * If a custom size wants to have additional mime types support, they will need to have
+ * the provide_additional_mime_types property in the size details or it should be added using the
+ * webp_uploads_image_sizes_with_additional_mime_type_support filter.
+ *
+ * @since n.e.x.t
+ *
+ * @return array An array of image sizes that can have additional mime types.
+ */
+function webp_uploads_get_image_sizes_with_additional_mime_type_support() {
+	$allowed_sizes    = array( 'thumbnail', 'medium', 'medium_large', 'large', 'post-thumbnail' );
+	$additional_sizes = wp_get_additional_image_sizes();
+
+	foreach ( $additional_sizes as $size => $size_details ) {
+		if (
+			isset( $size_details['provide_additional_mime_types'] ) &&
+			filter_var( $size_details['provide_additional_mime_types'], FILTER_VALIDATE_BOOLEAN )
+		) {
+			$allowed_sizes[] = $size;
+		}
+	}
+
+	/**
+	 * Filters image sizes that allowed to have additional mime types.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $allowed_image_sizes An array of image sizes that allowed to have additional mime types.
+	 */
+	$allowed_sizes = apply_filters( 'webp_uploads_image_sizes_with_additional_mime_type_support', $allowed_sizes );
+
+	return $allowed_sizes;
+}
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -105,9 +105,11 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 		return $metadata;
 	}
 
+	$sizes_with_mime_type_support = webp_uploads_get_image_sizes_with_additional_mime_type_support();
+
 	foreach ( $metadata['sizes'] as $size_name => $properties ) {
-		// This image size is not defined or not an array.
-		if ( ! is_array( $properties ) ) {
+		// Do nothing if this image size is not an array or is not allowed to have additional mime types.
+		if ( ! is_array( $properties ) || ! isset( $sizes_with_mime_type_support[ $size_name ] ) ) {
 			continue;
 		}
 
@@ -587,15 +589,21 @@ add_filter( 'post_thumbnail_html', 'webp_uploads_update_featured_image', 10, 3 )
  * @return array An array of image sizes that can have additional mime types.
  */
 function webp_uploads_get_image_sizes_with_additional_mime_type_support() {
-	$allowed_sizes    = array( 'thumbnail', 'medium', 'medium_large', 'large', 'post-thumbnail' );
 	$additional_sizes = wp_get_additional_image_sizes();
+	$allowed_sizes    = array(
+		'thumbnail'      => 'thumbnail',
+		'medium'         => 'medium',
+		'medium_large'   => 'medium_large',
+		'large'          => 'large',
+		'post-thumbnail' => 'post-thumbnail',
+	);
 
 	foreach ( $additional_sizes as $size => $size_details ) {
 		if (
 			isset( $size_details['provide_additional_mime_types'] ) &&
 			filter_var( $size_details['provide_additional_mime_types'], FILTER_VALIDATE_BOOLEAN )
 		) {
-			$allowed_sizes[] = $size;
+			$allowed_sizes[ $size ] = $size;
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -631,4 +649,28 @@ function() {
...
+				$sizes['allowed_size_400x300'] = 'allowed_size_400x300';
+				return $sizes;
+			}
+		);
+
+		add_image_size( 'allowed_size_400x300', 400, 300, true );
+		add_image_size( 'not_allowed_size_200x150', 200, 150, true );
+
+		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/car.jpeg' );
+
```
peterwilsoncc, line 657:
```suggestion

	// Clean up.
	remove_image_size( 'allowed_size_400x300' );
	remove_image_size( 'not_allowed_size_200x150' );
```

The test suite doesn't clean up image sizes between tests, so best to so so manually to avoid subsequent tests generating these image sizes.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -498,6 +498,15 @@ function () {
 			}
 		);
 
+		remove_all_filters( 'webp_uploads_image_sizes_with_additional_mime_type_support' );
```
peterwilsoncc: I don't think these lines are needed.

The test suite resets all hooks during tear down so all previous filters are already gone.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +575,46 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+		'medium_large'   => 'medium_large',
+		'large'          => 'large',
+		'post-thumbnail' => 'post-thumbnail',
+	);
+
+	foreach ( $additional_sizes as $size => $size_details ) {
+		if (
+			isset( $size_details['provide_additional_mime_types'] ) &&
+			filter_var( $size_details['provide_additional_mime_types'], FILTER_VALIDATE_BOOLEAN )
+		) {
```
peterwilsoncc:
```suggestion
		if ( ! empty( $size_details['provide_additional_mime_types'] ) ) {
```

WP Core doesn't use `filter_var()` because it's not guaranteed to be available and/or accurate. If the intent of the performance lab plugin is to be merged in to core then it's best to avoid to reduce refactoring for the WP merge.

If this code isn't eventually intended for core then what you have is probably fien.

jjgrainger: @eugene-manuilov if there is a need to check that the value is boolean, could `is_bool` be used here instead?

jjgrainger (commented):
@eugene-manuilov Looks good, just added a comment following on from @peterwilsoncc review

tests/modules/images/webp-uploads/load-tests.php
eugene-manuilov: Thanks. Updated.

tests/modules/images/webp-uploads/load-tests.php
eugene-manuilov: Sounds good. Updated.

modules/images/webp-uploads/load.php
eugene-manuilov: The idea was to allow developers to use `on`/`off`, `yes`/`no`, `1`/`0`, and `true`/`false` as possible values for the `provide_additional_mime_types` argument. Updated to check whether that argument is not empty.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +575,43 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+ */
+function webp_uploads_get_image_sizes_with_additional_mime_type_support() {
+	$additional_sizes = wp_get_additional_image_sizes();
+	$allowed_sizes    = array(
+		'thumbnail'      => 'thumbnail',
+		'medium'         => 'medium',
+		'medium_large'   => 'medium_large',
+		'large'          => 'large',
+		'post-thumbnail' => 'post-thumbnail',
+	);
```
felixarntz: Using a map here doesn't really make sense I think. This is documented as a simple array (as in "indexed") of image sizes, and having it be a map makes using the filter more complicated.

Maybe you're doing this to be able to use `isset` instead of `array_search`? That seems reasonable to me, but we should handle that internally so that external developers don't have to deal with it.

I'd say either we should handle the transformation to this "map" in a way that the filter usage is not affected by it, or if we prefer to use a map it should avoid the need to define duplicate values like now. If we wanted this to be a map for more efficient PHP lookups, I think the better approach would be to include all sizes by default as keys and then use a boolean as value, which would only be `true` for the core sizes.

If we do the latter, we also will need to update all documentation for the return value and most importantly filter value to clarify that format.

modules/images/webp-uploads/load.php
```diff
@@ -105,9 +105,11 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 		return $metadata;
 	}
 
+	$sizes_with_mime_type_support = webp_uploads_get_image_sizes_with_additional_mime_type_support();
+
 	foreach ( $metadata['sizes'] as $size_name => $properties ) {
-		// This image size is not defined or not an array.
-		if ( ! is_array( $properties ) ) {
+		// Do nothing if this image size is not an array or is not allowed to have additional mime types.
+		if ( ! is_array( $properties ) || ! isset( $sizes_with_mime_type_support[ $size_name ] ) ) {
```
felixarntz: See my other comment. Depending what we decide there, this may need to be changed to
```suggestion
		if ( ! is_array( $properties ) || empty( $sizes_with_mime_type_support[ $size_name ] ) ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +575,43 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+ *
+ * If a custom size wants to have additional mime types support, they will need to have
+ * the provide_additional_mime_types property in the size details or it should be added using the
+ * webp_uploads_image_sizes_with_additional_mime_type_support filter.
+ *
+ * @since n.e.x.t
+ *
+ * @return array An array of image sizes that can have additional mime types.
+ */
+function webp_uploads_get_image_sizes_with_additional_mime_type_support() {
```
felixarntz: See my other comment: If we change this to return _all_ sizes with a boolean for their additional MIME type support, we should probably rename this to reflect that data format better, e.g.
```suggestion
function webp_uploads_get_image_sizes_additional_mime_type_support() {
```

Just because then this value is no longer _only_ the sizes _with_ additional MIME type support, I guess.

felixarntz (commented):
@eugene-manuilov This mostly looks solid. I have some thoughts on the data format chosen here which is overly complicated IMO. It makes sense to be able to use `isset` over `array_search`, but we should not make such very "minor" performance enhancements at the cost of developer experience (the duplication of data there right now is problematic to me). We can probably find a way to do it in a way that is both performant and more clear for developers.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +575,43 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+	$allowed_sizes    = array(
+		'thumbnail'      => 'thumbnail',
+		'medium'         => 'medium',
+		'medium_large'   => 'medium_large',
+		'large'          => 'large',
+		'post-thumbnail' => 'post-thumbnail',
+	);
+
+	foreach ( $additional_sizes as $size => $size_details ) {
+		if ( ! empty( $size_details['provide_additional_mime_types'] ) ) {
```
adamsilverstein: where does `provide_additional_mime_types` property come from or this something you were thinking we would add later?

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +575,43 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+ * the provide_additional_mime_types property in the size details or it should be added using the
+ * webp_uploads_image_sizes_with_additional_mime_type_support filter.
+ *
+ * @since n.e.x.t
+ *
+ * @return array An array of image sizes that can have additional mime types.
+ */
+function webp_uploads_get_image_sizes_with_additional_mime_type_support() {
+	$additional_sizes = wp_get_additional_image_sizes();
+	$allowed_sizes    = array(
```
adamsilverstein, line 656: This is the complete list of core sizes including core themes:

```php
$default_core_sizes = array (
		'thumbnail',
		'medium',
		'medium_large',
		'large',
		// 2x medium_large size.
		'1536x1536',
		// 2x large size.
		'2048x2048',
		// Twentyeleven theme.
		'large-feature',
		'small-feature',
		// Twentyfourteen theme.
		'twentyfourteen-full-width',
		// Twentyseventeen theme.
		'twentyseventeen-featured-image',
		'twentyseventeen-thumbnail-avatar',
		// Twentytwenty theme.
		'twentytwenty-fullscreen',
	);
```

modules/images/webp-uploads/load.php
```diff
@@ -105,9 +105,11 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 		return $metadata;
 	}
 
+	$sizes_with_mime_type_support = webp_uploads_get_image_sizes_with_additional_mime_type_support();
+
 	foreach ( $metadata['sizes'] as $size_name => $properties ) {
```
adamsilverstein, line 110: how about filtering the sizes you iterate through ahead of time, filtering out any where the key isn't in $sizes_with_mime_type_support?  

Q: This loop already only applies to non primary mime types, right?

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +575,43 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
 	return webp_uploads_img_tag_update_mime_type( $html, 'post_thumbnail_html', $attachment_id );
 }
 add_filter( 'post_thumbnail_html', 'webp_uploads_update_featured_image', 10, 3 );
+
+/**
+ * Returns an array of image sizes that can have alternative mime types. By default, only core sizes
```
adamsilverstein: Suggested wording changes:
"can have alternative mime types" -> "have secondary mime type output enabled"
"By default, only core sizes..." -> "Core sizes and core theme sizes are enabled by default"

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +575,43 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+ * are allowed and will be returned.
+ *
+ * If a custom size wants to have additional mime types support, they will need to have
```
adamsilverstein: Replace with
"Developers can control the generation of additional mime images for all sizes using this filter."

We can add the bit about adding via `add_image_size` later when we support that.

modules/images/webp-uploads/load.php
```suggestion
 * Returns an array of image size names that can have alternative mime types. By default, only core sizes
```

modules/images/webp-uploads/load.php
felixarntz: This would be the field that can later be populated via `add_image_size`'s additional parameter. So having this here prepares us for that.
Technically, one could already set it now by directly touching the `$_wp_additional_image_sizes` global - which of course is discouraged, but I think it makes sense supporting this here already since we'll need it anyway and it's technically possible.

modules/images/webp-uploads/load.php
eugene-manuilov: Thanks, @felixarntz. I have left the map, but updated it to have size names only as keys and used a boolean as values as you said. Does it look good now?

modules/images/webp-uploads/load.php
eugene-manuilov: Yep, updated.

modules/images/webp-uploads/load.php

modules/images/webp-uploads/load.php
eugene-manuilov: Ok, added image size names from core themes as well... however not sure if we really need them there 🤔

modules/images/webp-uploads/load.php

modules/images/webp-uploads/load.php
eugene-manuilov: Updated the text but decided not to add information about `add_image_size` yet.

modules/images/webp-uploads/load.php
eugene-manuilov: > how about filtering the sizes you iterate through ahead of time, filtering out any where the key isn't in $sizes_with_mime_type_support?

I think we don't need to do it because we return early anyway.

> Q: This loop already only applies to non primary mime types, right?

The `sizes` property contains all sizes that an image has including non primary sizes and mime types too.

modules/images/webp-uploads/load.php
felixarntz: @eugene-manuilov Let's remove those again. The core theme sizes for sure should not be included here, as the themes should be adjusted instead.

For the additional core sizes `1536x1536` and `2048x2048`, they also are added via `add_image_size` so the default here would be automatically overwritten by the following `foreach` loop anyway. So let's remove these as well. We should later pass the parameter as `true` there from within core.

This list should only include the core sizes which _do not_ use `add_image_size` and therefore are not part of `wp_get_additional_image_sizes()`.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -502,6 +502,14 @@ function () {
 			}
 		);
 
+		add_filter(
+			'webp_uploads_image_sizes_with_additional_mime_type_support',
+			function( $sizes ) {
+				$sizes['1536x1536'] = '1536x1536';
+				return $sizes;
+			}
+		);
```
felixarntz: Let's remove this, as this size does not have to exist for each image - that doesn't belong here. We shouldn't include checks for specific sizes here. Let's also remove the `1536x1536`-related assertion below.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -602,6 +610,14 @@ public function it_should_use_smaller_jpg_image_when_smaller_size_preferred() {
 	 * @test
 	 */
 	public function it_should_replace_the_featured_image_to_web_p_when_requesting_the_featured_image() {
+		add_filter(
+			'webp_uploads_image_sizes_with_additional_mime_type_support',
+			function( $sizes ) {
+				$sizes['1536x1536'] = '1536x1536';
+				return $sizes;
+			}
+		);
```
felixarntz: Same as above, let's remove this.

We should then either remove the assertion below that this doesn't include any `.jpeg`, or (maybe better), we should ensure that it only includes `.jpeg` for sizes that do not have additional MIME type support (e.g. for the above `1536x1536`).

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -635,4 +651,30 @@ function() {
...
+		add_filter(
+			'webp_uploads_image_sizes_with_additional_mime_type_support',
+			function( $sizes ) {
+				$sizes['allowed_size_400x300'] = 'allowed_size_400x300';
+				return $sizes;
+			}
+		);
+
+		add_image_size( 'allowed_size_400x300', 400, 300, true );
+		add_image_size( 'not_allowed_size_200x150', 200, 150, true );
```
felixarntz: I think for full coverage, we should add another test that does the same thing, but without the filter and instead manually setting that new field in the `$_wp_additional_image_sizes` global (which will later become available as a `add_image_size` parameter.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -635,4 +651,30 @@ function() {
...
+	/**
+	 * Tests whether additional mime types generated only for allowed image sizes or not.
+	 *
+	 * @test
+	 */
+	public function it_should_create_mime_types_for_allowed_sizes_only() {
+		add_filter(
+			'webp_uploads_image_sizes_with_additional_mime_type_support',
+			function( $sizes ) {
+				$sizes['allowed_size_400x300'] = 'allowed_size_400x300';
```
felixarntz: Let's use the new format here:
```suggestion
				$sizes['allowed_size_400x300'] = true;
```

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -635,4 +651,30 @@ function() {
 		$tag = wp_get_attachment_image( $attachment_id, 'medium', false, array( 'class' => "wp-image-{$attachment_id}" ) );
 		$this->assertNotSame( $tag, webp_uploads_img_tag_update_mime_type( $tag, 'the_content', $attachment_id ) );
 	}
+
+	/**
+	 * Tests whether additional mime types generated only for allowed image sizes or not.
+	 *
+	 * @test
+	 */
+	public function it_should_create_mime_types_for_allowed_sizes_only() {
```
felixarntz: See below: If we add another test case using the image sizes global, we may want to rename this to
```suggestion
	public function it_should_create_mime_types_for_allowed_sizes_only_via_filter() {
```

felixarntz (commented):
@eugene-manuilov Production code looks good, however we should remove all the additional sizes again. The remaining feedback is exclusively some minor things about tests.

modules/images/webp-uploads/load.php
eugene-manuilov: Sounds good. Removed other sizes.

tests/modules/images/webp-uploads/load-tests.php
eugene-manuilov: Yep, removed.

tests/modules/images/webp-uploads/load-tests.php
eugene-manuilov: Yes, removed and updated this test to ensure that the `src` attribute contains the webp version of the image which satisfy the requirements for this test - we confirm that the correct webp version has been used for the post thumbnail image. Does it look good to you?

tests/modules/images/webp-uploads/load-tests.php
eugene-manuilov: Good idea. Added one more test.

tests/modules/images/webp-uploads/load-tests.php
eugene-manuilov: Good catch. Updated.

tests/modules/images/webp-uploads/load-tests.php
eugene-manuilov: Updated.

tests/modules/images/webp-uploads/load-tests.php
felixarntz: LGTM.

felixarntz (approved):
@eugene-manuilov Looks great, thank you!

felixarntz:
@adamsilverstein @peterwilsoncc Can you please give this a pass too? We need at least a second approval :)

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -635,4 +635,50 @@ function() {
...
+	/**
+	 * Tests whether additional mime types generated only for allowed image sizes or not when the global variable is updated.
+	 *
+	 * @test
+	 */
+	public function it_should_create_mime_types_for_allowed_sizes_only_via_global_variable() {
+		add_image_size( 'allowed_size_400x300', 400, 300, true );
+		add_image_size( 'not_allowed_size_200x150', 200, 150, true );
+
+		$GLOBALS['_wp_additional_image_sizes']['allowed_size_400x300']['provide_additional_mime_types'] = true;
```
adamsilverstein, line 675: I assume this could be removed once we update `add_image_size` to accept the new "output_additional_mimes" parameter, right? Can you add a TODO comment to remove this once that is in place?

adamsilverstein:
+1 looks good to me, I left one request to comment on the `$GLOBAL` setting line in the tests which shouldn't be needed once `add_image_size` is updated.

felixarntz:
@adamsilverstein Added the `TODO` comment, back to you for approval.

adamsilverstein (approved):
Excellent!

felixarntz:
FYI: https://github.com/WordPress/wordpress-develop/pull/3166 is the new PR that implements a slightly modified version of this for WordPress core.

PR #418, opened 2022-07-06
Discard WebP image if it is larger than corresponding JPEG image

### LINKED ISSUE #372: Discard WebP image if it is larger than corresponding JPEG image
Following some brainstorming and discussion around the WebP efforts at WordCamp Europe, one of the key takeaways was that we should include the following enhancement in the implementation: **WebP versions of an image size should only be kept if they are in fact smaller than the corresponding JPEG version.**

This will certainly be the case most of the time, but in the special cases where it isn't, there is in fact no point in keeping the WebP version around. WebP versions for each size are created for their performance benefit, and if in any scenario the WebP file is larger than the JPEG file, this benefit is no longer present.

While this only has a minor impact on storage usage, the enhancement is also slightly related to that, as it helps not using any storage space _wastefully_. All other considerations that add to the necessary storage space are tied to a particular benefit (either performance or backward compatibility), but keeping larger WebP files around is simply wasteful and should be avoided.

The goal here is the following:
* Files of the primary MIME type (i.e. in our focused scenario usually JPEG) should always be kept.
* Files of additional MIME types (i.e. in our focused scenario usually WebP) should only be kept if they are in fact smaller than the corresponding file of the primary MIME type.

Here's the proposed implementation:
* Modify `webp_uploads_create_sources_property()`:
    * In both places where an additional image source file is created, its `filesize` should be immediately compared to the `filesize` of the primary MIME type file (i.e. the one that is stored already for each size and for the "full" attachment file).
    * If the created file has a larger `filesize`, it should be deleted again immediately (and of course not be added to the `sources` map).
* PHPUnit tests should be added to verify this behavior.

mukeshpanchal27:
## Summary

After these changes, it affects many current unit tests because those tests are expecting .webp extensions but the images are bigger, so they end up with .jpg extensions. To handle the tests I use the filter approach to pass tests. thanks, @peterwilsoncc for that review.

After implementing this change we no longer need the smaller file size filter and related code. The filter `$prefer_smaller_image_file = apply_filters('webp_uploads_prefer_smaller_image_file', false);` and code for full size image https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/load.php#L477-L485 and thumbnail size image https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/load.php#L533-L541 because now images have only files that have a smaller size.

### Introduce a new filter `webp_prioritise_filesize_over_mime_type` to handle old unit tests.
The new file size changes affect some old tests to handle it a new filter `webp_prioritise_filesize_over_mime_type` introduce, that allow users to disable to the new behavior. by default its value is `true` in code base. For the unit tests we don't need this filter so added `add tests_add_filter( 'webp_prioritise_filesize_over_mime_type', '__return_false' );` in the tests bootstrap file so old tests don;t affect with new changes. For new tests, we include `add_filter( 'webp_prioritise_filesize_over_mime_type', '__return_true' );`.

Fixes #372

## Checklist

### DIFF
--- a/modules/images/webp-uploads/helper.php
+++ b/modules/images/webp-uploads/helper.php
@@ -240,3 +240,36 @@ function webp_uploads_get_attachment_sources( $attachment_id, $size = 'thumbnail
 	// Return an empty array if no sources found.
 	return array();
 }
+
+/**
+ * Check whether the additional image is larger than the original image.
+ *
+ * @since n.e.x.t
+ *
+ * @param array $original An array with the metadata of the attachment.
+ * @param array $additional An array containing the filename and file size for additional mime.
+ * @return bool True if the additional image is larger than the original image, otherwise false.
+ */
+function webp_uploads_should_discard_additional_image_file( array $original, array $additional ) {
+	$original_image_filesize   = isset( $original['filesize'] ) ? (int) $original['filesize'] : 0;
+	$additional_image_filesize = isset( $additional['filesize'] ) ? (int) $additional['filesize'] : 0;
+	if ( $original_image_filesize > 0 && $additional_image_filesize > 0 ) {
+		/**
+		 * Filter whether WebP images that are larger than the matching JPEG should be discarded.
+		 *
+		 * By default the performance lab plugin will use the mime type with the smaller filesize
+		 * rather than defaulting to `webp`.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param bool $preferred_filesize Prioritize file size over mime type. Default true.
+		 */
+		if (
+			apply_filters( 'webp_uploads_discard_larger_generated_images', true )
+			&& $additional_image_filesize >= $original_image_filesize
+		) {
+			return true;
+		}
+	}
+	return false;
+}
--- a/modules/images/webp-uploads/helper.php
+++ b/modules/images/webp-uploads/helper.php
@@ -264,10 +264,9 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
 		 *
 		 * @param bool $preferred_filesize Prioritize file size over mime type. Default true.
 		 */
-		if (
-			apply_filters( 'webp_uploads_discard_larger_generated_images', true )
-			&& $additional_image_filesize >= $original_image_filesize
-		) {
+		$webp_discard_larger_images = apply_filters( 'webp_uploads_discard_larger_generated_images', true );
+
+		if ( $webp_discard_larger_images && $additional_image_filesize >= $original_image_filesize ) {
 			return true;
 		}
 	}
--- a/modules/images/webp-uploads/helper.php
+++ b/modules/images/webp-uploads/helper.php
@@ -246,7 +246,7 @@ function webp_uploads_get_attachment_sources( $attachment_id, $size = 'thumbnail
  *
  * @since n.e.x.t
  *
- * @param array $original An array with the metadata of the attachment.
+ * @param array $original   An array with the metadata of the attachment.
  * @param array $additional An array containing the filename and file size for additional mime.
  * @return bool True if the additional image is larger than the original image, otherwise false.
  */
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -96,30 +96,9 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 			continue;
 		}
 
-		$original_image_filesize = isset( $metadata['filesize'] ) ? (int) $metadata['filesize'] : 0;
-		$webp_image_filesize     = isset( $image['filesize'] ) ? (int) $image['filesize'] : 0;
-		if ( $original_image_filesize > 0 && $webp_image_filesize > 0 ) {
-			/**
-			 * Filter whether WebP images that are larger than the matching JPEG should be discarded.
-			 *
-			 * By default the performance lab plugin will use the mime type with the smaller filesize
-			 * rather than defaulting to `webp`.
-			 *
-			 * @since n.e.x.t
-			 *
-			 * @param bool $preferred_filesize Prioritize file size over mime type. Default true.
-			 */
-			if (
-				apply_filters( 'webp_uploads_discard_larger_generated_images', true )
-				&& $webp_image_filesize >= $original_image_filesize
-			) {
-				if ( ! file_exists( $destination ) ) {
-					continue;
-				}
-
-				wp_delete_file_from_directory( $destination, $original_directory );
-				continue;
-			}
+		if ( webp_uploads_should_discard_additional_image_file( $metadata, $image ) ) {
+			wp_delete_file_from_directory( $destination, $original_directory );
+			continue;
 		}
 
 		$metadata['sources'][ $targeted_mime ] = $image;
@@ -180,22 +159,10 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 				continue;
 			}
 
-			$original_thumbnail_image_filesize = isset( $properties['filesize'] ) ? (int) $properties['filesize'] : 0;
-			$webp_thumbnail_image_filesize     = isset( $source['filesize'] ) ? (int) $source['filesize'] : 0;
-			if ( $original_thumbnail_image_filesize > 0 && $webp_thumbnail_image_filesize > 0 ) {
-				/** This filter is documented in modules/images/webp-uploads/load.php */
-				if (
-					apply_filters( 'webp_uploads_discard_larger_generated_images', true )
-					&& $webp_thumbnail_image_filesize >= $original_thumbnail_image_filesize
-				) {
-					$destination = path_join( $original_directory, $source['file'] );
-					if ( ! file_exists( $destination ) ) {
-						continue;
-					}
-
-					wp_delete_file_from_directory( $destination, $original_directory );
-					continue;
-				}
+			if ( webp_uploads_should_discard_additional_image_file( $properties, $source ) ) {
+				$destination = path_join( $original_directory, $source['file'] );
+				wp_delete_file_from_directory( $destination, $original_directory );
+				continue;
 			}
 
 			$properties['sources'][ $mime ]  = $source;
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -597,14 +597,21 @@ function webp_uploads_img_tag_update_mime_type( $original_image, $context, $atta
 		}
 	}
 
-	if (
-		! has_action( 'wp_footer', 'webp_uploads_wepb_fallback' ) &&
-		$image !== $original_image &&
-		'the_content' === $context &&
-		'image/jpeg' === $original_mime &&
-		'image/webp' === $target_mime
-	) {
-		add_action( 'wp_footer', 'webp_uploads_wepb_fallback' );
+	foreach ( $target_mimes as $target_mime ) {
+
+		if ( $target_mime === $original_mime ) {
+			continue;
+		}
+
+		if (
+			! has_action( 'wp_footer', 'webp_uploads_wepb_fallback' ) &&
+			$image !== $original_image &&
+			'the_content' === $context &&
+			'image/jpeg' === $original_mime &&
+			'image/webp' === $target_mime
+		) {
+			add_action( 'wp_footer', 'webp_uploads_wepb_fallback' );
+		}
 	}
 
 	return $image;
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -103,7 +103,7 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 		}
 
 		/**
-		 * Filter whether to prioritize the file type over the mime type.
+		 * Filter whether WebP images that are larger than the matching JPEG should be discarded.
 		 *
 		 * By default the performance lab plugin will use the mime type with the smaller filesize
 		 * rather than defaulting to `webp`.
@@ -113,7 +113,7 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 		 * @param bool $preferred_filesize Prioritize file size over mime type. Default true.
 		 */
 		if (
-			apply_filters( 'webp_prioritise_filesize_over_mime_type', true )
+			apply_filters( 'webp_uploads_discard_larger_generated_images', true )
 			&& $webp_image_filesize >= $original_image_filesize
 		) {
 			if ( ! file_exists( $destination ) ) {
@@ -190,7 +190,7 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 
 			/** This filter is documented in modules/images/webp-uploads/load.php */
 			if (
-				apply_filters( 'webp_prioritise_filesize_over_mime_type', true )
+				apply_filters( 'webp_uploads_discard_larger_generated_images', true )
 				&& $webp_thumbnail_image_filesize >= $original_thumbnail_image_filesize
 			) {
 				$destination = trailingslashit( $original_directory ) . "{$source['file']}";
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -192,16 +192,7 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 			// Remove the WebP image if it is bigger than the original thumbnail image.
 			if (
 				$webp_thumbnail_image_filesize >= $original_thumbnail_image_filesize
-				/**
-				 * Filter whether to prioritize the file type over the mime type.
-				 *
-				 * By default the performance lab plugin will use the mime type with the smaller filesize
-				 * rather than defaulting to `webp`.
-				 *
-				 * @since n.e.x.t
-				 *
-				 * @param bool $perfer_filesize Prioritize file size over mime type. Default true.
-				 */
+				/** This filter is documented in modules/images/webp-uploads/load.php */
 				&& apply_filters( 'webp_prioritise_filesize_over_mime_type', true )
 			) {
 				$destination = trailingslashit( $original_directory ) . "{$source['file']}";
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -102,7 +102,7 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 			continue;
 		}
 
-		/* Remove WebP image if it bigger then original image */
+		// Remove WebP image if it is bigger than the original image.
 		if (
 			$webp_image_filesize >= $original_image_filesize
 			/**
@@ -113,10 +113,14 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 			 *
 			 * @since n.e.x.t
 			 *
-			 * @param bool $perfer_filesize Prioritize file size over mime type. Default true.
+			 * @param bool $preferred_filesize Prioritize file size over mime type. Default true.
 			 */
 			&& apply_filters( 'webp_prioritise_filesize_over_mime_type', true )
 		) {
+			if ( ! file_exists( $destination ) ) {
+				continue;
+			}
+
 			wp_delete_file_from_directory( $destination, $original_directory );
 			continue;
 		}
@@ -185,7 +189,7 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 				continue;
 			}
 
-			/* Remove WebP image if it bigger then original thumbnail image */
+			// Remove the WebP image if it is bigger than the original thumbnail image.
 			if (
 				$webp_thumbnail_image_filesize >= $original_thumbnail_image_filesize
 				/**
@@ -201,6 +205,10 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 				&& apply_filters( 'webp_prioritise_filesize_over_mime_type', true )
 			) {
 				$destination = trailingslashit( $original_directory ) . "{$source['file']}";
+				if ( ! file_exists( $destination ) ) {
+					continue;
+				}
+
 				wp_delete_file_from_directory( $destination, $original_directory );
 				continue;
 			}
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -96,6 +96,31 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 			continue;
 		}
 
+		$original_image_filesize = isset( $metadata['filesize'] ) ? (int) $metadata['filesize'] : 0;
+		$webp_image_filesize     = isset( $image['filesize'] ) ? (int) $image['filesize'] : 0;
+		if ( $original_image_filesize <= 0 && $webp_image_filesize <= 0 ) {
+			continue;
+		}
+
+		/* Remove WebP image if it bigger then original image */
+		if (
+			$webp_image_filesize >= $original_image_filesize
+			/**
+			 * Filter whether to prioritize the file type over the mime type.
+			 *
+			 * By default the performance lab plugin will use the mime type with the smaller filesize
+			 * rather than defaulting to `webp`.
+			 *
+			 * @since n.e.x.t
+			 *
+			 * @param bool $perfer_filesize Prioritize file size over mime type. Default true.
+			 */
+			&& apply_filters( 'webp_prioritise_filesize_over_mime_type', true )
+		) {
+			wp_delete_file_from_directory( $destination, $original_directory );
+			continue;
+		}
+
 		$metadata['sources'][ $targeted_mime ] = $image;
 		wp_update_attachment_metadata( $attachment_id, $metadata );
 	}
@@ -151,6 +176,32 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 
 			$source = webp_uploads_generate_image_size( $attachment_id, $size_name, $mime );
 			if ( is_wp_error( $source ) ) {
+				continue;
+			}
+
+			$original_thumbnail_image_filesize = isset( $properties['filesize'] ) ? (int) $properties['filesize'] : 0;
+			$webp_thumbnail_image_filesize     = isset( $source['filesize'] ) ? (int) $source['filesize'] : 0;
+			if ( $original_thumbnail_image_filesize <= 0 && $webp_thumbnail_image_filesize <= 0 ) {
+				continue;
+			}
+
+			/* Remove WebP image if it bigger then original thumbnail image */
+			if (
+				$webp_thumbnail_image_filesize >= $original_thumbnail_image_filesize
+				/**
+				 * Filter whether to prioritize the file type over the mime type.
+				 *
+				 * By default the performance lab plugin will use the mime type with the smaller filesize
+				 * rather than defaulting to `webp`.
+				 *
+				 * @since n.e.x.t
+				 *
+				 * @param bool $perfer_filesize Prioritize file size over mime type. Default true.
+				 */
+				&& apply_filters( 'webp_prioritise_filesize_over_mime_type', true )
+			) {
+				$destination = trailingslashit( $original_directory ) . "{$source['file']}";
+				wp_delete_file_from_directory( $destination, $original_directory );
 				continue;
 			}
 
@@ -415,6 +466,7 @@ add_filter( 'the_content', 'webp_uploads_update_image_references', 10 );
  * for the specified image sizes, the *.webp references are stored inside of each size.
  *
  * @since 1.0.0
+ * @since n.e.x.t Remove `webp_uploads_prefer_smaller_image_file` filter
  *
  * @param string $image         An <img> tag where the urls would be updated.
  * @param string $context       The context where this is function is being used.
@@ -423,23 +475,10 @@ add_filter( 'the_content', 'webp_uploads_update_image_references', 10 );
  */
 function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id ) {
 	$metadata = wp_get_attachment_metadata( $attachment_id );
+
 	if ( empty( $metadata['file'] ) ) {
 		return $image;
 	}
-
-	/**
-	 * Filters whether the smaller image should be used regardless of which MIME type is preferred overall.
-	 *
-	 * This is disabled by default only because it is not part of the current WordPress core feature proposal.
-	 *
-	 * By enabling this, the plugin will compare the image file sizes and prefer the smaller file regardless of MIME
-	 * type.
-	 *
-	 * @since 1.0.0
-	 *
-	 * @param bool $prefer_smaller_image_file Whether to prefer the smaller image file.
-	 */
-	$prefer_smaller_image_file = apply_filters( 'webp_uploads_prefer_smaller_image_file', false );
 
 	/**
 	 * Filters mime types that should be used to update all images in the content. The order of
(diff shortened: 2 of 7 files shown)


modules/images/webp-uploads/load.php
```diff
@@ -96,6 +96,31 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 			continue;
 		}
 
+		$original_image_filesize = isset( $metadata['filesize'] ) ? (int) $metadata['filesize'] : 0;
+		$webp_image_filesize     = isset( $image['filesize'] ) ? (int) $image['filesize'] : 0;
+		if ( $original_image_filesize <= 0 && $webp_image_filesize <= 0 ) {
```
mehulkaklotar: can you use yoda conditions for comparison here?

modules/images/webp-uploads/load.php
```diff
@@ -96,6 +96,31 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+			 * Filter whether to prioritize the file type over the mime type.
+			 *
+			 * By default the performance lab plugin will use the mime type with the smaller filesize
+			 * rather than defaulting to `webp`.
+			 *
+			 * @since n.e.x.t
+			 *
+			 * @param bool $perfer_filesize Prioritize file size over mime type. Default true.
+			 */
+			&& apply_filters( 'webp_prioritise_filesize_over_mime_type', true )
```
mehulkaklotar: Just a suggestion: can we put the filter first and then put the condition to check image file size? or we don't want to run the extra codes from the plugins/themes before we actually check for the file sizes?

modules/images/webp-uploads/load.php
```diff
@@ -96,6 +96,31 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 			continue;
 		}
 
+		$original_image_filesize = isset( $metadata['filesize'] ) ? (int) $metadata['filesize'] : 0;
+		$webp_image_filesize     = isset( $image['filesize'] ) ? (int) $image['filesize'] : 0;
+		if ( $original_image_filesize <= 0 && $webp_image_filesize <= 0 ) {
+			continue;
+		}
+
+		/* Remove WebP image if it bigger then original image */
```
mehulkaklotar:
```suggestion
		/* Remove WebP image if it is bigger than original image */
```

modules/images/webp-uploads/load.php
```diff
@@ -96,6 +96,31 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+			$webp_image_filesize >= $original_image_filesize
+			/**
+			 * Filter whether to prioritize the file type over the mime type.
+			 *
+			 * By default the performance lab plugin will use the mime type with the smaller filesize
+			 * rather than defaulting to `webp`.
+			 *
+			 * @since n.e.x.t
+			 *
+			 * @param bool $perfer_filesize Prioritize file size over mime type. Default true.
```
mehulkaklotar:
```suggestion
			 * @param bool $preferred_filesize Prioritize file size over mime type. Default true.
```

modules/images/webp-uploads/load.php
```diff
@@ -96,6 +96,31 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+			 */
+			&& apply_filters( 'webp_prioritise_filesize_over_mime_type', true )
+		) {
```
mehulkaklotar: can we also check the file using `file_exists` before we try to delete it?

modules/images/webp-uploads/load.php
```diff
@@ -154,6 +179,32 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 				continue;
 			}
 
+			$original_thumbnail_image_filesize = isset( $properties['filesize'] ) ? (int) $properties['filesize'] : 0;
+			$webp_thumbnail_image_filesize     = isset( $source['filesize'] ) ? (int) $source['filesize'] : 0;
+			if ( $original_thumbnail_image_filesize <= 0 && $webp_thumbnail_image_filesize <= 0 ) {
+				continue;
+			}
+
+			/* Remove WebP image if it bigger then original thumbnail image */
```
mehulkaklotar:
```suggestion
			/* Remove WebP image if it is bigger than original thumbnail image */
```

modules/images/webp-uploads/load.php
```diff
@@ -154,6 +179,32 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 				continue;
 			}
 
+			$original_thumbnail_image_filesize = isset( $properties['filesize'] ) ? (int) $properties['filesize'] : 0;
+			$webp_thumbnail_image_filesize     = isset( $source['filesize'] ) ? (int) $source['filesize'] : 0;
+			if ( $original_thumbnail_image_filesize <= 0 && $webp_thumbnail_image_filesize <= 0 ) {
```
mehulkaklotar: same yoda condition checks here?

modules/images/webp-uploads/load.php
```diff
@@ -154,6 +179,32 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+				 * rather than defaulting to `webp`.
+				 *
+				 * @since n.e.x.t
+				 *
+				 * @param bool $perfer_filesize Prioritize file size over mime type. Default true.
+				 */
+				&& apply_filters( 'webp_prioritise_filesize_over_mime_type', true )
+			) {
+				$destination = trailingslashit( $original_directory ) . "{$source['file']}";
+				wp_delete_file_from_directory( $destination, $original_directory );
```
mehulkaklotar, line 166: Check for file exists here as well?

tests/bootstrap.php
```diff
@@ -53,5 +53,8 @@ function() {
 	1
 );
 
+// Remove the filter so that other tests are not impacted.
+tests_add_filter( 'webp_prioritise_filesize_over_mime_type', '__return_false' );
```
mehulkaklotar: Nice!

mehulkaklotar (changes requested):
@mukeshpanchal27 Please have a look at my comments below.

modules/images/webp-uploads/load.php
spacedmonkey: https://make.wordpress.org/core/2022/06/14/upcoming-disallow-assignments-in-conditions-and-remove-the-yoda-condition-requirement-for-php/

mukeshpanchal27: Thanks for the feedback.

As per the WordPress proposal, the Yoda condition is not necessary. We already use a similar condition [helper.php](https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/helper.php#L125-L128)

modules/images/webp-uploads/load.php
mukeshpanchal27: Check my previous comment on a similar review.

tests/bootstrap.php
mukeshpanchal27: Thanks!

modules/images/webp-uploads/load.php
mukeshpanchal27: Text updated, thanks!

modules/images/webp-uploads/load.php
mukeshpanchal27: Text updated.

modules/images/webp-uploads/load.php
mukeshpanchal27: Good catch.

modules/images/webp-uploads/load.php
mehulkaklotar: Yes, that's fine I guess.

modules/images/webp-uploads/load.php
mukeshpanchal27: Thanks for pointing it out. If we check the filter before the file size, the old unit tests fail.

mehulkaklotar (approved):

modules/images/webp-uploads/load.php
```diff
@@ -96,6 +96,34 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+		 *
+		 * By default the performance lab plugin will use the mime type with the smaller filesize
+		 * rather than defaulting to `webp`.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param bool $preferred_filesize Prioritize file size over mime type. Default true.
+		 */
+		if (
+			apply_filters( 'webp_prioritise_filesize_over_mime_type', true )
```
adamsilverstein: for filter name, the plugin uses the module name prefix. Also, lets stick to the US spelling of "prioritize" with a "z". So maybe `webp_uploads_prioritize_filesize_over_mime_type`

modules/images/webp-uploads/load.php
```diff
@@ -96,6 +96,34 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 			continue;
 		}
 
+		$original_image_filesize = isset( $metadata['filesize'] ) ? (int) $metadata['filesize'] : 0;
+		$webp_image_filesize     = isset( $image['filesize'] ) ? (int) $image['filesize'] : 0;
+		if ( $original_image_filesize <= 0 && $webp_image_filesize <= 0 ) {
+			continue;
+		}
```
eugene-manuilov: If we don't know a filesize of either original or converted image, then we should not "continue" the loop because it will skip updating the `sources` property and the converted image will remain on the disk. What we need to do here is to skip checking whether we can delete the webp image:

```php
$original_image_filesize = isset( $metadata['filesize'] ) ? (int) $metadata['filesize'] : 0;
$webp_image_filesize     = isset( $image['filesize'] ) ? (int) $image['filesize'] : 0;
if ( $original_image_filesize > 0 && $webp_image_filesize > 0 ) {
    if (
        apply_filters( 'webp_prioritise_filesize_over_mime_type', true )
        && $webp_image_filesize >= $original_image_filesize
    ) {
        ...
        wp_delete_file_from_directory( $destination, $original_directory );
        continue;
    }
}

$metadata['sources'][ $targeted_mime ] = $image;
```

modules/images/webp-uploads/load.php
```diff
@@ -154,6 +182,26 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 				continue;
 			}
 
+			$original_thumbnail_image_filesize = isset( $properties['filesize'] ) ? (int) $properties['filesize'] : 0;
+			$webp_thumbnail_image_filesize     = isset( $source['filesize'] ) ? (int) $source['filesize'] : 0;
+			if ( $original_thumbnail_image_filesize <= 0 && $webp_thumbnail_image_filesize <= 0 ) {
+				continue;
+			}
```
eugene-manuilov: Same here. We should not skip the rest of the loop if we don't know sizes of files, we should skip deleting webp version instead.

modules/images/webp-uploads/load.php
```diff
@@ -96,6 +96,34 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+		$webp_image_filesize     = isset( $image['filesize'] ) ? (int) $image['filesize'] : 0;
+		if ( $original_image_filesize <= 0 && $webp_image_filesize <= 0 ) {
+			continue;
+		}
+
+		/**
+		 * Filter whether to prioritize the file type over the mime type.
+		 *
+		 * By default the performance lab plugin will use the mime type with the smaller filesize
+		 * rather than defaulting to `webp`.
```
adamsilverstein: Return false to disable this behavior?

modules/images/webp-uploads/load.php
```diff
@@ -96,6 +96,34 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param bool $preferred_filesize Prioritize file size over mime type. Default true.
+		 */
```
adamsilverstein: document `@return`

modules/images/webp-uploads/load.php
```diff
@@ -154,6 +182,26 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+			if ( $original_thumbnail_image_filesize <= 0 && $webp_thumbnail_image_filesize <= 0 ) {
+				continue;
+			}
+
+			/** This filter is documented in modules/images/webp-uploads/load.php */
+			if (
+				apply_filters( 'webp_prioritise_filesize_over_mime_type', true )
+				&& $webp_thumbnail_image_filesize >= $original_thumbnail_image_filesize
+			) {
+				$destination = trailingslashit( $original_directory ) . "{$source['file']}";
```
eugene-manuilov: Use `path_join` instead of `trailingslashit`:
```suggestion
				$destination = path_join( $original_directory, $source['file'] );
```

modules/images/webp-uploads/load.php
```diff
@@ -456,63 +492,42 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
...
-			if ( $filtered_image === $image ) {
-				$image = str_replace(
-					$basename,
-					$metadata['sources'][ $target_mime ]['file'],
-					$image
-				);
-			} else {
-				$image = $filtered_image;
-			}
+		$basename = wp_basename( $metadata['file'] );
```
eugene-manuilov: This variable is needed only if `$filtered_image === $image`. Let's move it into the condition to avoid calling the `wp_basename` function when it's not needed.

modules/images/webp-uploads/load.php
```diff
@@ -96,6 +96,34 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+		$original_image_filesize = isset( $metadata['filesize'] ) ? (int) $metadata['filesize'] : 0;
+		$webp_image_filesize     = isset( $image['filesize'] ) ? (int) $image['filesize'] : 0;
+		if ( $original_image_filesize <= 0 && $webp_image_filesize <= 0 ) {
+			continue;
+		}
+
+		/**
+		 * Filter whether to prioritize the file type over the mime type.
+		 *
+		 * By default the performance lab plugin will use the mime type with the smaller filesize
```
adamsilverstein: This documentation can be improved a bit - the filter controls whether to retain WebP images even if they are larger than the jpegs, right?

Maybe something like
"Filter whether WebP images that are larger than the matching JPEG should be discarded. Default is true"

Also makes me think the filter name should reflect this as well, maybe something like `webp_uploads_discard_larger_generated_images`, curious what @felixarntz thinks since he is great at naming things.

adamsilverstein (changes requested):
Overall approach seems solid, nice work! I left a few suggestions mostly around filter naming and documentation.

tests/bootstrap.php
eugene-manuilov: This filter should be added in the `WebP_Uploads_Load_Tests::set_up` method to avoid affecting tests for other modules.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -349,9 +349,18 @@ public function it_should_prevent_replacing_a_webp_image() {
 			TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/balloons.webp'
 		);
 
-		$tag = wp_get_attachment_image( $attachment_id, 'medium', false, array( 'class' => "wp-image-{$attachment_id}" ) );
+		$tag          = wp_get_attachment_image( $attachment_id, 'medium', false, array( 'class' => "wp-image-{$attachment_id}" ) );
+		$expected_tag = $tag;
+		$metadata     = wp_get_attachment_metadata( $attachment_id );
+		foreach ( $metadata['sizes'] as $size => $properties ) {
+			$expected_tag = str_replace( $properties['sources']['image/webp']['file'], $properties['sources']['image/jpeg']['file'], $expected_tag );
+		}
```
eugene-manuilov, line 368: Why this image should use jpeg versions of the image? The original image is already in the `webp` format and the intention of this PR is to make sure we don't generate alternative mime types if they are bigger than the original image. This test shouldn't change.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -631,4 +619,82 @@ function() {
...
+		add_filter( 'webp_prioritise_filesize_over_mime_type', '__return_true' );
+
+		// Look for an image that contains all of the additional mime type images.
+		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/car.jpeg' );
+		$tag           = wp_get_attachment_image( $attachment_id, 'full', false, array( 'class' => "wp-image-{$attachment_id}" ) );
+		$expected_tag  = $tag;
+		$metadata      = wp_get_attachment_metadata( $attachment_id );
+		$result        = webp_uploads_img_tag_update_mime_type( $tag, 'the_content', $attachment_id );
+		$this->assertImageHasSource( $attachment_id, 'image/webp' );
+		$this->assertImageHasSizeSource( $attachment_id, 'thumbnail', 'image/webp' );
```
eugene-manuilov, line 651: This test is intended to check that webp image is not going to be generated and used for an image when webp version is bigger than the original jpeg, right? Then why do we still have `image/webp` sources in the metadata?

eugene-manuilov: Hard to follow what this test does... If my original guess is correct and we expect that no webp version will be generated, then what we need to do here is the following:

1. upload a new jpeg image
2. make sure that there is no sources for the `image/webp` mime type
3. make sure that there is no webp files for this image on the disk
4. make sure that there is no `.webp` URLs when we generate the `` tag for this image.

In other words, we shouldn't use str_replace calls to cater the expected tag, but we need to add a check to make sure that webp files are deleted from the disk.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -631,4 +619,82 @@ function() {
...
+		$this->assertNotSame( $tag, $expected_tag );
+		$this->assertSame( $expected_tag, $result );
+	}
+
+	/**
+	 * The image with the smaller filesize should be used when webp_prioritise_filesize_over_mime_type is set to true.
+	 *
+	 * @test
+	 */
+	public function it_should_use_smaller_mime_type_image_from_specific_source() {
```
eugene-manuilov: I assume this test is intended for the case when webp version is smaller than the original jpeg image. In this case we need to check the same things that we check in the previous test but other way around.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -631,4 +619,82 @@ function() {
...
+		$this->assertNotSame( $tag, webp_uploads_img_tag_update_mime_type( $tag, 'the_content', $attachment_id ) );
+
+	}
+
+	/**
+	 * The image with the smaller filesize should be used when webp_prioritise_filesize_over_mime_type is set to true.
+	 *
+	 * @test
+	 */
+	public function it_should_use_smaller_mime_type_image_from_specific_source_for_webp_image() {
```
eugene-manuilov: If we already have two tests that check both cases (original is bigger than alternative, and alternative is bigger than original), then what does this test check and how this is different from the previous two tests?

eugene-manuilov (changes requested):
Thanks, @mukeshpanchal27. This is a good start. Left some comments for you to address. Please, take a look.

modules/images/webp-uploads/load.php
mukeshpanchal27: Yes.

modules/images/webp-uploads/load.php
mukeshpanchal27: Thanks for the feedback, Adam.

I think it should be a `param` for the filter, as I checked the similar filters in the Performance plugin and WordPress code, and there it uses `param` for the filters.

https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/load.php#L430-L442
https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/options.php#L198-L208
https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/rest-api/class-wp-rest-server.php#L267-L274

modules/images/webp-uploads/load.php
mukeshpanchal27: Thanks for the feedback. I have updated the filter name to `webp_uploads_discard_larger_generated_images` as you suggest in other comments. That makes more sense.

modules/images/webp-uploads/load.php
mukeshpanchal27: Thanks for the feedback. I have updated the documentation for the filter and updated the filter name to `webp_uploads_discard_larger_generated_images` for now. Will update it when Felix reviews this and suggests the changes regarding the filter name.

tests/bootstrap.php
mukeshpanchal27: @eugene-manuilov OK, However, if we take this approach, we must also include `WebP_Uploads_Image_Edit_Tests::set_up` and `WebP_Uploads_REST_API_Tests::set_up`, as these two classes are also affected by new functionality.

modules/images/webp-uploads/load.php
mukeshpanchal27: Thanks for the explanation of the changes, Eugene. It makes sense to me. I have updated PR and addressed these changes.

modules/images/webp-uploads/load.php
mukeshpanchal27: Addressed.

modules/images/webp-uploads/load.php
mukeshpanchal27: Nice catch. I followed a similar format that is used here: https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/load.php#L92 But this review makes more sense than the current approach.

modules/images/webp-uploads/load.php
mukeshpanchal27: LGTM

tests/bootstrap.php
mukeshpanchal27: I have added a filter in three files with `set_up` method.

tests/modules/images/webp-uploads/load-tests.php
mukeshpanchal27: Thanks for pointing it out. 

Now we use https://github.com/WordPress/performance/blob/fix/372-discard-larger-webp/modules/images/webp-uploads/load.php#L486 for getting additional mime types. When you upload `.webp` then it will also generate a smaller version of the `.jpeg` image and display it on the frontend side.

I think we have to either remove these tests or change them.

tests/modules/images/webp-uploads/load-tests.php
mukeshpanchal27: Thanks for the detailed review.

I have updated the test function name for clarification.

- In `it_should_use_smaller_mime_type_image_from_source_for_all_sizes` test I used `car.jpeg` image that has all smaller `.webp` images
- In `it_should_use_smaller_mime_type_image_for_full_size` test I used `leafs.jpg` image that has only one smaller `.webp` image for full size.
- In `it_should_use_smaller_mime_type_image_from_specific_source_for_webp_image` test I used `balloons.webp` image that has smaller `.jpeg` image for medium and thumbnail size.

Does this make sense? Or do I need to remove these tests?

tests/modules/images/webp-uploads/load-tests.php
mukeshpanchal27: Please check my comment on the similar review.

tests/modules/images/webp-uploads/load-tests.php
mukeshpanchal27: Please check my comment on the similar review.

modules/images/webp-uploads/load.php
```diff
@@ -154,6 +180,24 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
 			}
 
+			$original_thumbnail_image_filesize = isset( $properties['filesize'] ) ? (int) $properties['filesize'] : 0;
+			$webp_thumbnail_image_filesize     = isset( $source['filesize'] ) ? (int) $source['filesize'] : 0;
+			if ( $original_thumbnail_image_filesize > 0 && $webp_thumbnail_image_filesize > 0 ) {
+				/** This filter is documented in modules/images/webp-uploads/load.php */
+				if (
+					apply_filters( 'webp_uploads_discard_larger_generated_images', true )
+					&& $webp_thumbnail_image_filesize >= $original_thumbnail_image_filesize
+				) {
```
felixarntz: This is exactly the same logic as in https://github.com/WordPress/performance/pull/418/files#diff-71ed188a16b79d179868ee80e76f775462a451e871fc3d6214fcacca3edcea8cR99-R115 I think. Let's implement a separate function for this, especially for the filter.

I would suggest implementing a `webp_uploads_should_discard_additional_image_file( array $original, array $additional, string $size_name ) : bool` function. That function could include all this logic, comparing the file sizes if they are not `0` and then run the filter, and return the boolean result.

We could then call this function in both places instead of duplicating all this logic.

modules/images/webp-uploads/load.php
```diff
@@ -96,6 +96,32 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+			 *
+			 * @param bool $preferred_filesize Prioritize file size over mime type. Default true.
+			 */
+			if (
+				apply_filters( 'webp_uploads_discard_larger_generated_images', true )
+				&& $webp_image_filesize >= $original_image_filesize
+			) {
+				if ( ! file_exists( $destination ) ) {
+					continue;
+				}
```
felixarntz: Is this check really needed? When would the file not exist here? I would think this is impossible to get to because if generating the file failed, we would have already bailed earlier above in the `is_wp_error` check.

modules/images/webp-uploads/load.php
```diff
@@ -154,6 +180,24 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
...
+			if ( $original_thumbnail_image_filesize > 0 && $webp_thumbnail_image_filesize > 0 ) {
+				/** This filter is documented in modules/images/webp-uploads/load.php */
+				if (
+					apply_filters( 'webp_uploads_discard_larger_generated_images', true )
+					&& $webp_thumbnail_image_filesize >= $original_thumbnail_image_filesize
+				) {
+					$destination = path_join( $original_directory, $source['file'] );
+					if ( ! file_exists( $destination ) ) {
+						continue;
+					}
```
felixarntz: Same question as above.

felixarntz (commented):
@mukeshpanchal27 Mostly looks solid to me. I have one suggestion about avoiding duplicate code.

modules/images/webp-uploads/load.php
mukeshpanchal27: Thanks for the suggestion, Felix.

Do we need `string $size_name` parameter in the function as we don't check the image size name in our logic? or do we need to check any logic on the image size name?

modules/images/webp-uploads/load.php
mukeshpanchal27: Thanks for pointing it out. Initially, I didn't use it, but after feedback from someone, I included it for verification before deleting it. You are right, we don't need to check it. Changes were made to the PR.

modules/images/webp-uploads/load.php
mukeshpanchal27: Thanks for pointing it out. Changes were made to the PR.

modules/images/webp-uploads/load.php
mukeshpanchal27: I have updated the PR that did not include `string $size_name` in the function. Waiting for your reply to the previous comment on the function parameter.

modules/images/webp-uploads/helper.php
```diff
@@ -240,3 +240,36 @@ function webp_uploads_get_attachment_sources( $attachment_id, $size = 'thumbnail
...
+		 *
+		 * By default the performance lab plugin will use the mime type with the smaller filesize
+		 * rather than defaulting to `webp`.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param bool $preferred_filesize Prioritize file size over mime type. Default true.
+		 */
+		if (
+			apply_filters( 'webp_uploads_discard_larger_generated_images', true )
```
felixarntz: This is a bit hard to read like that, more importantly though because of the extra line in between the filter and its doc block, I don't think this will be parsed correctly.

Let's assign the filter value to a variable, this way the if clause will also be easier to read.

modules/images/webp-uploads/load.php
felixarntz: Thank you @mukeshpanchal27, that makes sense. I was probably thinking to include `$size_name` in the filter as an additional parameter, but that shouldn't be necessary for now. We can always add it later if it comes up as a request.

felixarntz (commented):
@mukeshpanchal27 Thank you for the updates. The production code looks good now, except for the small problem with the doc block for the filter.

Now that we have this function, can you add test coverage for it? Probably a good one to use a `@dataProvider` with different values for the `$original` and `$additional`, plus I think it would be good to also have one test to ensure that setting the filter to `false` disables that behavior.

tests/bootstrap.php
eugene-manuilov: Thanks, @mukeshpanchal27.

tests/modules/images/webp-uploads/load-tests.php
eugene-manuilov: Ok, thanks.

tests/modules/images/webp-uploads/load-tests.php
eugene-manuilov: Ok, in this case, I would call these tests a bit differently:

1. `it_should_not_create_webp_when_jpeg_image_is_smaller_than_webp`
2. `it_should_create_webp_when_webp_is_smaller_than_jpegs`
3. `it_should_create_webp_for_some_sizes_which_are_smaller_in_webp_format`

mukeshpanchal27: Thank you for the tests name suggestions.

I have updated the tests' names according to their work. Please have a look.

modules/images/webp-uploads/helper.php
mukeshpanchal27: Thank you for the feedback.

Changes were made to the PR.

mukeshpanchal27:
@felixarntz Thank you for the feedback. I have addressed the feedback and updated the PR and unit tests accordingly.

spacedmonkey:
Stupid question, what about pngs and gifs?

tests/modules/images/webp-uploads/helper-tests.php
```diff
@@ -410,4 +410,52 @@ function () {
...
+	 *
+	 * @dataProvider data_provider_image_filesize
+	 *
+	 * @test
+	 */
+	public function it_should_return_an_additional_image_if_it_is_bigger_than_original( $original_filesize, $additional_filesize ) {
+		add_filter( 'webp_uploads_discard_larger_generated_images', '__return_false' );
+
+		$output = webp_uploads_should_discard_additional_image_file( $original_filesize, $additional_filesize );
+		$this->assertSame( $output, false );
```
felixarntz: Why not `assertFalse`?

tests/modules/images/webp-uploads/helper-tests.php
```diff
@@ -410,4 +410,52 @@ function () {
...
+	}
+
+	/**
+	 * Check the disabled behaviour of the webp_uploads_discard_larger_generated_images filter.
+	 *
+	 * @dataProvider data_provider_image_filesize
+	 *
+	 * @test
+	 */
+	public function it_should_return_an_additional_image_if_it_is_bigger_than_original( $original_filesize, $additional_filesize ) {
```
felixarntz: That method name is not really accurate, this method is about testing that no images are discarded regardless of file size if the filter is set to `false`.

felixarntz (approved):
@mukeshpanchal27 Pretty much LGTM! Two minor nit-picks that aren't blockers though.

felixarntz:
@spacedmonkey 
> Stupid question, what about pngs and gifs?

The scope of the current project is only to generate WebP for JPEG. For PNG and even more so GIF, there are additional complexities to consider. If someone wanted to generate WebP from PNG (or even JPEG from PNG), they could easily do so using the available filters, and it all would be subject to the same behavior as here. This is in fact a very powerful feature of this "additional MIME type support" feature, that it can also satisfy lots of other potential use-cases with few lines of code. We can always think about changing the defaults and expanding support inside of core later.

felixarntz:
@adamsilverstein @eugene-manuilov Can you please re-review this PR since you previously requested changes?

modules/images/webp-uploads/helper.php
```diff
@@ -240,3 +240,35 @@ function webp_uploads_get_attachment_sources( $attachment_id, $size = 'thumbnail
 	// Return an empty array if no sources found.
 	return array();
 }
+
+/**
+ * Check whether the additional image is larger than the original image.
+ *
+ * @since n.e.x.t
+ *
+ * @param array $original An array with the metadata of the attachment.
```
adamsilverstein:
```suggestion
 * @param array $original   An array with the metadata of the attachment.
```

adamsilverstein: align comment clocks

adamsilverstein (approved):
Left one nit pick spacing feedback item on a doc block. Nice work!

modules/images/webp-uploads/helper.php
mukeshpanchal27: Thank you for pointing it out. The lint command does not show any errors for this method. PR is updated with these changes.

mukeshpanchal27:
Assign it to @felixarntz. It is now ready to merge.

tests/modules/images/webp-uploads/helper-tests.php
mukeshpanchal27: Yes. Thank you for fixing it.

tests/modules/images/webp-uploads/helper-tests.php
mukeshpanchal27: Thank you!

mxbclang:
@mukeshpanchal27 I still see unaddressed changes from @eugene-manuilov and it sounds like @felixarntz wanted him to do a final review, as well – can you take another look?

mukeshpanchal27:
@bethanylang All the changes are addressed, which were raised by @eugene-manuilov. and resolved all conversations.

eugene-manuilov (approved):
LGTM

felixarntz:
@mukeshpanchal27 Looks like there is a merge conflict here now that something else got merged. Can you please address this? Then we can merge it.

mukeshpanchal27:
@felixarntz Merge conflict resolved. 

The foreach loop added for the https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/load.php#L625-L633 to get `$target_mime`.

felixarntz:
Excellent, thanks @mukeshpanchal27!

PR #420, opened 2022-07-06
Introduce `webp_uploads_get_content_image_mimes()` to get content image MIME replacement rules

### LINKED ISSUE #414: Background images set with CSS do not uses the *.webp version if present
## Bug Description
We already got a fix at #288 by @mitogh for the featured images - thanks for that! Some time ago i developed a small simple website with gutenberg & generatepress. i tried to use as many default blocks as possible. Some elements are working with the webp replacement as expected - some not yet.

generatepress offers a container block. it allows to choose a image from the library and sets its url into the field
image

the background image is set over css and currently not replaced by a webp image - although one would be available
image

This currently seems to be the only issue left on my testpage where webp images dont get automatically choosen.

## Steps to reproduce

1. Upload jpeg images - implement them as background images over css 
2. install performance plugin, activate webp
3. use Regenerate Thumbnails plugin to generate the web
4. See on frontend if they get used as webp or still jpg

## Additional Context

 - PHP Version: 8.0
 - OS: Windows
 - Browser: MS Edge
 - Plugin Version: Version 1.2.0

eugene-manuilov:
## Summary

Originally intended as a fix to #414, but that issue isn't feasible to fix in the plugin, so now it's just a little API enhancement.

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -573,3 +562,52 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
 add_filter( 'post_thumbnail_html', 'webp_uploads_update_featured_image', 10, 3 );
+
+/**
+ * Updates the image src data to use the webp format if an image has webp version.
+ *
+ * @since n.e.x.t
+ *
+ * @param array|bool   $image         Image data.
+ * @param int          $attachment_id Image attachment ID.
+ * @param string|array $size          Requested image size.
```
mukeshpanchal27: We should match the doc block written for WordPress: https://developer.wordpress.org/reference/hooks/wp_get_attachment_image_src/

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +562,52 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+function webp_uploads_update_attachment_image_src( $image, $attachment_id, $size ) {
+	// Do nothing if image data is falsy or the size is not a string.
+	if ( empty( $image ) || ! is_string( $size ) ) {
+		return $image;
+	}
+
+	// Do nothing if not in the "the_content" context.
+	if ( ! doing_filter( 'the_content' ) ) {
+		return $image;
+	}
```
mukeshpanchal27: I think we have to check this condition first in the function as it only works if the image is in the `the_content` context.

mukeshpanchal27 (commented):
I have added some feedback regarding the doc block and code changes. [PR 418](https://github.com/WordPress/performance/pull/418) only allows a smaller version of an additional mime type image when this PR is merged. After that, this PR needs to be updated according to it.

modules/images/webp-uploads/load.php
```diff
@@ -573,3 +562,52 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+	if ( null === $target_mime ) {
+		return $image;
+	}
+
+	// Replace the image with the new mime type.
+	$image[0] = dirname( $image[0] ) . '/' . $sources[ $target_mime ]['file'];
+
+	return $image;
+}
+add_filter( 'wp_get_attachment_image_src', 'webp_uploads_update_attachment_image_src', 10, 3 );
```
felixarntz: I think this is out of scope for the plugin, as functions controlled by developers should leave that control to developers. See https://github.com/WordPress/performance/issues/414#issuecomment-1182366880

modules/images/webp-uploads/load.php
```diff
@@ -468,6 +454,9 @@ function webp_uploads_img_tag_update_mime_type( $image, $context, $attachment_id
 		return $image;
 	}
 
+	// Get the original mime type for comparison.
+	$original_mime = get_post_mime_type( $attachment_id );
```
felixarntz: Nice!

felixarntz (commented):
@eugene-manuilov I don't think we should address this in the plugin, as it isn't really an appropriate solution, since in core we should do this primarily based on a parameter - and not just replace the images but use the right images from the start. See https://github.com/WordPress/performance/issues/414#issuecomment-1182366880 for more details.

With that said, you already put some work into this and I think breaking out the filter into a separate function and moving the unnecessary function call below the `foreach` loop are nice little improvements. So I'd say we can still merge this if we remove the new integrations around `wp_get_image_attachment_src`.

modules/images/webp-uploads/load.php
eugene-manuilov: Thanks, but we reverted this function.

modules/images/webp-uploads/load.php
eugene-manuilov: Same here, we reverted this function.

modules/images/webp-uploads/load.php
eugene-manuilov: Sounds good to me. Reverted it.

felixarntz (approved):
Thanks @eugene-manuilov! Will remove this PR from the issue since it doesn't really address it anymore, but rather is a nice little code enhancement.

felixarntz:
@eugene-manuilov There are some merge conflicts here, can you look into those? This is not critical for the release since it's just a little code enhancement, so I'm going to move it to the next milestone.

eugene-manuilov:
> @eugene-manuilov There are some merge conflicts here, can you look into those? This is not critical for the release since it's just a little code enhancement, so I'm going to move it to the next milestone.

@felixarntz, merge conflicts are fixed.

adamsilverstein (approved):
Nice!

mukeshpanchal27 (commented):
@eugene-manuilov Just needs to correct the merge conflict.

felixarntz (approved):
@eugene-manuilov Great!

PR #422, opened 2022-07-08
Ensure regenerating an image also regenerates WebP while not conflicting with other attachments

### LINKED ISSUE #395: Thumbnails not converted to webp since update 1.2.0
Hello, since update 1.2.0, my different thumbnail sizes are not automatically converted to webp. The options are still checked in settings, and it worked before. 

I'm using Real Thumbnail Generator to generate several thumbnails on each upload, and it worked fine with performance.

mukeshpanchal27:
## Summary

This issue is related to #359. 

Fixes #395

## Checklist

### DIFF
--- a/modules/images/webp-uploads/helper.php
+++ b/modules/images/webp-uploads/helper.php
@@ -144,7 +144,13 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 
 	// Skip creation of duplicate WebP image if an image file already exists in the directory.
 	if ( file_exists( $destination_file_name ) ) {
-		return new WP_Error( 'webp_image_file_present', __( 'The WebP image already exists.', 'performance-lab' ) );
+		$extension = explode( '|', $allowed_mimes[ $mime ] );
+		$filename  = pathinfo( $destination_file_name, PATHINFO_FILENAME );
+		$filename  = "{$filename}.{$extension[0]}";
+		return array(
+			'file'     => $filename,
+			'filesize' => wp_filesize( $destination_file_name ),
+		);
 	}
 
 	$image = $editor->save( $destination_file_name, $mime );
--- a/modules/images/webp-uploads/helper.php
+++ b/modules/images/webp-uploads/helper.php
@@ -144,11 +144,8 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 
 	// Skip creation of duplicate WebP image if an image file already exists in the directory.
 	if ( file_exists( $destination_file_name ) ) {
-		$extension = explode( '|', $allowed_mimes[ $mime ] );
-		$filename  = pathinfo( $destination_file_name, PATHINFO_FILENAME );
-		$filename  = "{$filename}.{$extension[0]}";
 		return array(
-			'file'     => $filename,
+			'file'     => wp_basename( $destination_file_name ),
 			'filesize' => wp_filesize( $destination_file_name ),
 		);
 	}


modules/images/webp-uploads/helper.php
```diff
@@ -144,7 +144,13 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 
 	// Skip creation of duplicate WebP image if an image file already exists in the directory.
 	if ( file_exists( $destination_file_name ) ) {
-		return new WP_Error( 'webp_image_file_present', __( 'The WebP image already exists.', 'performance-lab' ) );
+		$extension = explode( '|', $allowed_mimes[ $mime ] );
+		$filename  = pathinfo( $destination_file_name, PATHINFO_FILENAME );
+		$filename  = "{$filename}.{$extension[0]}";
+		return array(
+			'file'     => $filename,
```
jjgrainger: @mukeshpanchal27 is it possible to use something like `wp_basename( $destination_file_name );` here?

It looks like we already run similar logic to create the `$destination_file_name` above?

jjgrainger (changes requested):
Thanks @mukeshpanchal27 left a question for you to review.

jjgrainger:
@mukeshpanchal27 Speaking with you earlier today it looked like you discovered another issue when two or more images share the same file name but different mime types.

For example, if there are two different image files uploaded to WordPress named `image.jpg` and `image.jpeg`, the second image uploaded will reference the same WebP image generated for the first in the `sources` metadata.

This might be better visualised like this. After both images are uploaded here are the results.
```
image.jpg (image of car) => image.webp (image of a car)
image.jpeg (image of a tree) => image.webp (image of a car)
```

This only occurs with the latest changes you've made in the PR. Currently, if the file exists we skip the WebP generation and do not update the `sources` attribute metadata and therefore the WebP image is only used once.

This seems to be an edge case which only impacts images with the same filename but different extensions.

One solution I can see would be to use the file extension in the WebP filename, something like `image-jpeg.webp`, but I'm not sure if this is the best solution?

I also think this might be better addressed in a different issue or PR?

@felixarntz and @adamsilverstein, I would be interested to hear your thoughts on this?

mukeshpanchal27:
Similar discussion here https://github.com/WordPress/wordpress-develop/pull/2393#issuecomment-1178213379

modules/images/webp-uploads/helper.php
mukeshpanchal27: Thank you Joe, Yes it makes sense to use the `wp_basename` function.

mukeshpanchal27:
@adamsilverstein tried the approach to solve this in WP https://github.com/WordPress/wordpress-develop/pull/2393/commits/93542af6127c77f0a4609ce77abf5f616f6984e8 based on that I use it for our plugin.

### Update code in helper.php [here](https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/helper.php#L146-L148)
```
if ( file_exists( $destination_file_name ) ) {
	$original_directory    = pathinfo( $destination_file_name, PATHINFO_DIRNAME );
	$ext                   = pathinfo( $image_meta['file'], PATHINFO_EXTENSION );
	$destination_file_name = webp_uploads_unique_media_filename( $original_directory, wp_basename( $destination_file_name ), $ext );
}
```

### New function for unique filename
```
function webp_uploads_unique_media_filename( $dir, $filename, $ext ) {
	$extension_sets = array(
		'image/jpeg' => array( 'jpe', 'jpeg', 'jpg' ),
	);
	foreach( $extension_sets as $mime => $extensions ) {
		if ( ! in_array( $ext, $extensions, true ) ) {
			continue;
		}
		foreach( $extensions as $extension ) {
			$new_filename = $filename;
			$new_filename = wp_unique_filename( $dir, $new_filename );
			if ( $new_filename !== $filename ) {
				$new_filename = str_replace( $extension, $ext, $new_filename );
				return path_join( $dir, $new_filename );
			}
		}
	}
	return $filename;
}
```

## Result

### Turned off WebP and uploaded `testimage.jpeg` and `testimage.jpg` twice.

wp

### Turned on WebP and uploaded `testimage.jpeg` and `testimage.jpg` twice.

webp

Also, it generates WebP images for all images smaller than the original and works fine if we regenerate them.

## If we use this script, we will encounter a bug.

If we regenerate an image twice or more, then this script will generate a WebP image again and again with a new filename.

mukeshpanchal27:
Additionally, I researched other plugins and how they generated WebP images and file names for them. I checked [EWWW Image Optimizer](https://wordpress.org/plugins/ewww-image-optimizer/) plugin which generates the WebP name with an image extension. `filename.jpeg.webp` and `filename.jpg.webp` are two examples.

modules/images/webp-uploads/helper.php
```diff
@@ -144,7 +144,10 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 
 	// Skip creation of duplicate WebP image if an image file already exists in the directory.
 	if ( file_exists( $destination_file_name ) ) {
-		return new WP_Error( 'webp_image_file_present', __( 'The WebP image already exists.', 'performance-lab' ) );
+		return array(
+			'file'     => wp_basename( $destination_file_name ),
+			'filesize' => wp_filesize( $destination_file_name ),
+		);
```
felixarntz, line 150: This fixes the problem for regenerating images, but it then reintroduces a bug similar to #395. Let me clarify:
* Before #395 was fixed, a file for an image upload could overwrite a file for a previous image upload.
* After #395 was fixed, no overwrite would happen anymore, but now regeneration will fail for every image.
* With what you have here, regeneration will work again, but then the scenario from #395 would result in the image for the previous upload being used for the new upload, which would be a new bug.

I think we need an if condition here:
* If the `$destination_file_name` belongs to the given `$attachment_id`, then we shouldn't even return anything here - neither the error nor the previous file - since in that case we should actually overwrite the image (e.g. for a regeneration process).
* Otherwise (i.e. `$destination_file_name` is for another attachment), then we should still return the `WP_Error` present here.

felixarntz (commented):
@mukeshpanchal27 This is really tricky. You're fixing the problem here, but at the same time introducing another bug similar to #395.

I left a suggestion for how we could address this below. Let me know your thoughts. This is a P0, so let's make sure to fix this for the upcoming 1.3.0 release.

modules/images/webp-uploads/helper.php
mukeshpanchal27: Yes, it's a tricky issue. If we fix one problem it arises another.

> If the $destination_file_name belongs to the given $attachment_id, then we shouldn't even return anything here - neither the error nor the previous file - since in that case we should actually overwrite the image (e.g. for a regeneration process).

We can't check `$destination_file_name` is from which `$attachment_id` because when we regenerate an existing image `car.jpeg`, we don't have `sources` in metadata. The Regenerate script tries to generate `sources` for the image, but if we don't return an array of file and filesize, then it will not generate a WebP image.

Please check the attachment metadata that I printed before the `file_exists` condition.

image

If we remove the `file_exists` condition, then only it will work fine for regeneration. We will check that edge case after this release.

mukeshpanchal27: @felixarntz, Please check and share your feedback on this?

felixarntz: @mukeshpanchal27 After giving this some further thought, I think the problem here points to a greater issue that we have to consider when designing the feature to regenerate existing images. We need a way to know what the original image files were that were present _before_ starting the regeneration process for an attachment. I think we could use a meta key similar to `_wp_attachment_backup_sizes` to store the image files for the different sizes from right before the regeneration temporarily, until the process has been completed for the image. This will allow us to know whether an existing file belongs to the same attachment or not. cc @jjgrainger 

With that said, current solutions to regenerate existing images do not have that. While that is not great, I would argue, returning an error here is definitely the safer path. We may end up with some content not having WebP files when regenerated, but we absolutely cannot risk mixing up image files from another attachment. So I would say that we should close this PR without merging it. Great job finding that problem though, since it will be crucial for the regenerating images implementation!

mukeshpanchal27: Yes agreed. Let me close the PR.

PR #423, opened 2022-07-08
Rename Site Health check modules for language and consistency

### LINKED ISSUE #326: Update Settings language for health checks
## Feature Description

Currently, the Persistent Object Cache health check is the only one of the four health checks in the plugin that includes that phrase on the Settings screen:

image

The other health checks should be updated to add the word "Health Check" to make them easier to parse without having to read their description text.

OllieJones:
## Summary

Updated the (English) language captions in the Settings panel.  

Changed the site-health slugs to, for example, to `autoloaded-options-health-check` to  `audit-autoloaded-options`.

Made corresponding changes in the unit-test directory and file names.   

(This very easy dev task will help me ensure I know how to drive the PR workflow.)

Fixes #326

## Relevant technical choices

Trivial editing of the strucutured comments at the top of four load.php files

## Checklist


mxbclang (approved):
LGTM!

mukeshpanchal27:
IMO plugin creates a folder structure based on the module name, so if we change the module name, then it does not match.

mxbclang:
@felixarntz Tagging you in to take a quick look at this one based on @mukeshpanchal27's comment above.

felixarntz (approved):
Thanks @OllieJones! Potentially we should rename the module slugs (i.e. move the directories) as well?

felixarntz (changes requested):
After further thought, I tend to agree with @mukeshpanchal27's feedback in https://github.com/WordPress/performance/pull/423#issuecomment-1180359734 that we should probably rename the modules as a whole.

@OllieJones Could you rename the module directories and references to it?

This is not extremely critical and the deadline for release PRs is today, so I'm going to remove this one for now, we can target it for 1.4.0.

mukeshpanchal27:
@OllieJones We are two weeks away from the `1.4.0` release. Could you follow up on that? It would be great. Thank you!

OllieJones:
I will do this. 

I need clarification.  I think you want, for example, the Autoloaded Options Health Check module to be placed in the directory  `.../modules/health-check/autoloaded-options` rather than  `.../modules/site-health/audit-autoloaded-options`. Is that correct?

mukeshpanchal27:
It should be `.../modules/site-health/autoloaded-options-health-check` rather than `.../modules/site-health/audit-autoloaded-options` similar for other modules. Does that make sense?

OllieJones:
I made the changes requested by @mukeshpanchal27 and @felixarntz .

adamsilverstein (approved):

mukeshpanchal27 (approved):
LGTM

felixarntz (changes requested):
@OllieJones Apologies, I am now seeing another problem here regarding the module renaming (even though I originally proposed it): Renaming the modules is problematic since it may already be stored in the sites' DB options, or developers may use the module slugs in the filter. So it would result in breakage currently where if a site had one of the remained modules active, it would no longer be active after the update, which we must avoid.

If we want to go through with the module slug/directory changes, at a minimum we would need to add logic to retrieving the option value so that it migrates the old values to the new ones. But even that would not be 100% proof.

So I think (sorry again about that) it would be best to go back to your previous implementation where you only adjust the module names themselves. One thing missing here then is to also update the `CODEOWNERS` file comments where these same names are referenced accordingly. Also see the related #465 PR - we need to decide which one of these naming conventions to go with. cc @bethanylang

mxbclang:
Thanks @felixarntz! My personal preference would be to go with @OllieJones' approach and add "Health Check" to the end of each appropriate module title in order to distinguish those from the other, non-Health Check modules, but let me know what you think.

felixarntz:
@bethanylang Thanks, SGTM.

@OllieJones Can you revert the directory name changes in this PR so that it's back to only changes to the actual module titles?

I'll move this out of the 1.4.0 milestone for now since it's not ready yet, but if it gets ready tomorrow, we could potentially still include it in this release.

OllieJones:
The module slugs (directory names) are now reverted. Should be ready to merge.

mxbclang:
Thanks, @OllieJones! There are still changes that need to be made per @felixarntz's requests here: https://github.com/WordPress/performance/pull/423/files/8cf50133dacc686c153dfb632bfb0feb6369a497. Can you make those updates when you get a moment, as well? Then this should be ready to merge and I can update #465. Thanks!

felixarntz:
@bethanylang I think actually everything has been addressed now. Since we're only changing the module names here vs the slugs, no migration of the option should happen.

I think we're good to merge this and then update #465.

felixarntz (approved):
Great, thank you @OllieJones!

PR #424, opened 2022-07-12
Add replacing of images only in frontend context

### LINKED ISSUE #379: Only replace content images with additional MIME type version when in a frontend request context by default
To help address concerns around WebP compatibility, we should ensure that WebP images are used by default for the content **only** in the front end (or "template") context. 

Critically, when content is filtered outside this context - for example in REST endpoints and RSS feeds, the original mime type should be used by default (ew can leverage the existing filter here to make this customizable). In addition, core functions like `wp_get_attachment_image_url` which would be used to generate Open Graph tags should always return the original MIME type unless called in the template context.

Some research is needed on the best approach here to ensure template context is recognized and image replacement works in places like widgets or images placed in FSE headers while feeds and REST endpoints remain unaffected. We can possibly leverage `template_redirect` to ensure we are in the "right" loop.

akshitsethi:
## Summary

Fixes #379

## Relevant technical choices

Bails early if the request is not for the frontend context or is coming from a feed url. We also check if the request is made inside the `wp_head` hook so that original images are returned for use within `<head>` tags.

In-progress:
- [x] Add tests to verify

## Checklist


modules/images/webp-uploads/helper.php
```diff
@@ -240,3 +240,19 @@ function webp_uploads_get_attachment_sources( $attachment_id, $size = 'thumbnail
 	// Return an empty array if no sources found.
 	return array();
 }
+
+/**
+ * Verify if the request if for the frontend context.
```
mukeshpanchal27: It should be `Verify if the request is for the frontend context.`

modules/images/webp-uploads/helper.php
```diff
@@ -240,3 +240,19 @@ function webp_uploads_get_attachment_sources( $attachment_id, $size = 'thumbnail
...
+ * The is_feed() function is to check if the query is for feeds (rdf, rss, rss2, atom).
+ *
+ * @since n.e.x.t
+ *
+ * @return bool
+ */
+function webp_uploads_is_frontend_context() {
```
mitogh: Curios on why not just use `!is_feed()` additionally we might want to find a different name since this might be confused with the existing `is_front_page()` function that already exists but just test the front page.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -484,6 +484,61 @@ public function data_provider_not_supported_webp_images() {
...
+			TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg'
+		);
+
+		$tag = wp_get_attachment_image( $attachment_id, 'medium', false, array( 'class' => "wp-image-{$attachment_id}" ) );
+
+		$this->assertSame( $tag, webp_uploads_update_image_references( $tag ) );
+	}
+
+	/**
+	 * Mock is_home in $wp_query.
```
mukeshpanchal27: Should `is_feed`?

modules/images/webp-uploads/helper.php
felixarntz: > Curios on why not just use `!is_feed()`

There are other non-frontend contexts, like REST API. And plugins can also do whatever they want outside of a frontend context, so we should really use the checks for the hooks which are the only reliable way.

> additionally we might want to find a different name since this might be confused with the existing `is_front_page()` function that already exists but just test the front page

I'm less concerned about confusion with `is_front_page()`, however I agree the name here could be refined, because it doesn't check just the context. It also checks whether we're in the part of the HTML frontend that is shown to the user, a.k.a. not the `head`.

How about
```suggestion
function webp_uploads_in_frontend_body() {
```

modules/images/webp-uploads/helper.php
```diff
@@ -240,3 +240,19 @@ function webp_uploads_get_attachment_sources( $attachment_id, $size = 'thumbnail
...
+	if ( ! did_action( 'template_redirect' ) || doing_action( 'wp_head' ) || doing_action( 'wp_footer' ) || is_feed() ) {
```
felixarntz: I don't think we should include `wp_footer` here. It can also be used to inject stuff that is shown to the user, and most importantly, it _is_ within the `body` tag which is generally what's used for user-facing content. You're missing the check for `! did_action( 'wp_head' )` though - without that any code that executes after `template_redirect` but before `wp_head` would get `true` here which wouldn't be correct.

`is_feed()` isn't absolutely necessary since the other conditions would already cater for that, however I like the idea of including some of those cheaper checks here to not have to call the action checks unless necessary. In that case though, I think we should also include something for a REST API request, and we should put those first, above the hook checks.

We also need to check whether the `$wp_query` global is set, otherwise `is_feed()` can cause PHP warnings. In addition, it is safe to say that if that object is not set, we're not in a frontend context.

```suggestion
	global $wp_query;

	// Check if this request is generally outside (or before) any frontend context.
	if ( ! isset( $wp_query ) || defined( 'XMLRPC_REQUEST' ) || defined( 'REST_REQUEST' ) || is_feed() ) {
		return false;
	}

	// Check if we're anywhere before the 'wp_head' action has completed.
	if ( ! did_action( 'template_redirect' ) || ! did_action( 'wp_head' ) || doing_action( 'wp_head' ) ) {
```

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -484,6 +484,61 @@ public function data_provider_not_supported_webp_images() {
...
 	}
 
+	/**
+	 * Ensure jpg image is replaced if the request is for the frontend context.
+	 *
+	 * @group webp_uploads_update_image_references
+	 *
+	 * @test
+	 */
+	public function it_should_not_replace_image_references_when_request_is_for_frontend_context() {
```
felixarntz: I think the "not" here is a typo? We can also shorten this a bit.
```suggestion
	public function it_should_replace_image_references_when_in_frontend_context() {
```

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -484,6 +484,61 @@ public function data_provider_not_supported_webp_images() {
...
+		$this->mock_action_template_redirect();
```
felixarntz: See my other comment on the missing `wp_head` check above, you should also add a `mock_action_wp_head()` helper method and call that here too.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -484,6 +484,61 @@ public function data_provider_not_supported_webp_images() {
...
+	 * Prevent replacing a jpg image if the request is not for the frontend context.
+	 * In this test case, the request if for the `feed` page.
+	 *
+	 * @group webp_uploads_update_image_references
+	 *
+	 * @test
+	 */
+	public function it_should_not_replace_image_references_when_request_is_not_for_frontend_context() {
+		$this->mock_action_template_redirect();
+		$this->mock_is_feed_page();
```
felixarntz: This isn't a real scenario so I don't think we should test it. `template_redirect` will never fire if it is a feed. Let's rather focus on purely the feed check in this test.
```suggestion
		$this->mock_is_feed_page();
```

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -484,6 +484,61 @@ public function data_provider_not_supported_webp_images() {
...
+
+	/**
+	 * Prevent replacing a jpg image if the request is not for the frontend context.
+	 * In this test case, the request if for the `feed` page.
+	 *
+	 * @group webp_uploads_update_image_references
+	 *
+	 * @test
+	 */
+	public function it_should_not_replace_image_references_when_request_is_not_for_frontend_context() {
```
felixarntz:
```suggestion
	public function it_should_not_replace_image_references_when_request_is_for_a_feed() {
```

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -484,6 +484,61 @@ public function data_provider_not_supported_webp_images() {
...
+	public function it_should_not_replace_image_references_when_request_is_for_frontend_context() {
+		$this->mock_action_template_redirect();
+
+		$attachment_id = $this->factory->attachment->create_upload_object(
+			TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg'
+		);
+
+		$tag = wp_get_attachment_image( $attachment_id, 'medium', false, array( 'class' => "wp-image-{$attachment_id}" ) );
+		$this->assertNotSame( $tag, webp_uploads_update_image_references( $tag ) );
+	}
```
felixarntz: Let's add one more test that covers the scenario where you're in a frontend context, but before the user facing content (i.e. `template_redirect` has run but `wp_head` hasn't). Maybe that could be called `it_should_not_replace_image_references_when_in_frontend_context_but_before_body`.

felixarntz (commented):
@akshitsethi This looks like a good start, however some of the conditions need to be refined a bit. In combination with that, some more accurate test coverage would be helpful too.

modules/images/webp-uploads/helper.php
akshitsethi: @felixarntz In my test results, with the default environment settings, images were not replaced in the content as well when I added the check for `! did_action( 'wp_head` )`

tests/modules/images/webp-uploads/load-tests.php
akshitsethi: @felixarntz While testing this, `template_redirect` did fire for `rss2` feeds. You can verify this by creating a test post with images and check the `rss2` page `http://localhost:8888/rss2`

modules/images/webp-uploads/helper.php
akshitsethi: @felixarntz I'm curious as to the usage of `XMLRPC_REQUEST` here since it `XMLRPC` only accepts `POST` requests. Also, I don't see `template_redirect` every running for `REST_REQUEST` but I'm +1 on having a double check in place.

tests/modules/images/webp-uploads/load-tests.php
mukeshpanchal27: Thank you. Fixed.

modules/images/webp-uploads/helper.php
felixarntz: @akshitsethi 
> @felixarntz In my test results, with the default environment settings, images were not replaced in the content as well when I added the check for `! did_action( 'wp_head` )`

Hmm, I can't imagine that - can you provide some more detail? For example, which theme/plugins did you use? `wp_head` runs in the `<head>` of the page where no user-facing content is rendered, so we have to make sure this hook has already run. For example if someone includes an image _before_ `wp_head`, it should not be replaced.

adamsilverstein: Nice addition here to check these constants for efficiency ahead of the filter checks. a few other conditionals to consider here: DOING_AJAX, WP_ADMIN (or is_admin), DOING_AUTOSAVE, DOING_CRON & IFRAME_REQUEST. The content filter could in theory run in any of these contexts (not sure about IFRAME_REQUEST).

Q: should we have a filter in place that will enable overriding this limitation?

felixarntz: @adamsilverstein Great points. @akshitsethi Let's add these constants as well, except for the `IFRAME_REQUEST` one (since an iframe is also a frontend request).

Regarding a filter, that may be useful, but I think we could start without one unless/until someone requests an actual need / use-case for it.

tests/modules/images/webp-uploads/load-tests.php
felixarntz: @akshitsethi Oh sorry, I didn't know that, you're right. In that case this is good to keep.

That is another good reminder though why we need to also check for `did_action( 'wp_head' )`.

felixarntz:
@akshitsethi @adamsilverstein I have made the additional changes here around the conditions and enhanced test coverage, made it specific to the new function.

@akshitsethi Your previous point on `did_action( 'wp_head' )` being problematic was correct, however it is only a problem for full site editing themes, because for them, all the page content is generated before the page is rendered, so `wp_head` will only run after all the content hooks. On traditional themes, it works fine. With that said, of course this is not an acceptable solution, so I decided it is okay to omit the check for now. Content in the head should always be rendered as part of `wp_head`, and that applies even to full site editing themes.

felixarntz (approved):
Thanks @akshitsethi!

adamsilverstein (approved):
Nice!

tests/modules/images/webp-uploads/helper-tests.php
```diff
@@ -454,4 +454,57 @@ public function it_should_never_discard_additional_image_if_filter_is_false( $or
...
+	}
+
+	public function test_webp_uploads_in_frontend_body_after_template_redirect() {
+		$this->mock_empty_action( 'template_redirect' );
+		$result = webp_uploads_in_frontend_body();
+
+		$this->assertTrue( $result );
+	}
+
+	public function test_webp_uploads_in_frontend_body_within_wp_head() {
```
mehulkaklotar, line 489: More of nit-pick here but the test function names are different than what we actually use the names here like `it_should` or `it_` like that. And these functions doesn't have function doc comment.

cc: @felixarntz @akshitsethi

felixarntz: @mehulkaklotar I didn't think it was too critical to point it out in earlier tests for the plugin, but the convention for WordPress core tests is to always use `test_` followed by the function name, then optionally anything else. So I stuck with that here.

I would advise to follow that convention for future tests to monitor the core plan.

Regarding doc comments, test methods do not require any doc comments per core specifications - only when helpful. Preferably, the name of the method should be clear enough to know what is going on, or when it makes sense to clarify specific code, inline comments should be used.

cc @akshitsethi @adamsilverstein @mukeshpanchal27 @jjgrainger @eugene-manuilov

PR #431, opened 2022-07-15
Fix typos and Coding standard

akkspros:
fix some typos and coding standards

## Checklist

### DIFF
--- a/admin/load.php
+++ b/admin/load.php
@@ -408,7 +408,7 @@ function perflab_render_pointer() {
 	?>
 	<script id="perflab-admin-pointer" type="text/javascript">
 		jQuery( function() {
-			// Pointer Options
+			// Pointer Options.
 			var options = {
 				content: '<h3><?php echo esc_js( $heading ); ?></h3><p><?php echo wp_kses( $content, $wp_kses_options ); ?></p>',
 				position: {
--- a/tests/testdata/modules/images/webp-uploads/class-wp-image-edit.php
+++ b/tests/testdata/modules/images/webp-uploads/class-wp-image-edit.php
@@ -136,7 +136,7 @@ class WP_Image_Edit {
 
 	/**
 	 * Setup the $_REQUEST global so `wp_save_image` can process the image with the same editions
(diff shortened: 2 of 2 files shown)


mukeshpanchal27:
Nice catch.

admin/load.php
```diff
@@ -408,7 +408,7 @@ function perflab_render_pointer() {
 	?>
 	<script id="perflab-admin-pointer" type="text/javascript">
 		jQuery( function() {
-			// Pointer Options
+			// Pointer Options.
```
mxbclang, line 411: This period is unnecessary, so let's remove.

mxbclang (changes requested):

admin/load.php
mitogh: I think part of the standard is that all one-line comments end with a period.

> Comments come before the code to which they refer, and should always be preceded by a blank line. Capitalize the first letter of the comment, and include a period at the end when writing full sentences

https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/#comments

mxbclang: Hmm, interesting @mitogh! I guess I'd argue that this is not a "full sentence," but of course want to follow the coding standards.

felixarntz: Yeah, it's correct that basically any doc comment needs to end with a period per WP doc standards.

felixarntz (approved):
Thank you @akkspros!

mitogh (approved):

admin/load.php
mitogh: > Hmm, interesting @mitogh! I guess I'd argue that this is not a "full sentence," but of course want to follow the coding standards.

Right, this is mostly because the way the standard assumes what "full sentence" means as is just any complete line. @bethanylang

mukeshpanchal27 (approved):

mxbclang (approved):

felixarntz:
Removing the milestone here since this doesn't affect end users or any external integration point.

PR #442, opened 2022-07-18
Update language throughout the plugin from Group to Team

### LINKED ISSUE #441: Update plugin description and plugin author from "Group" to "Team"
Now that we're an official core WP Team, we should update the language in the plugin description and the plugin author name to reflect this. Currently:

image

mxbclang:
## Summary

Fixes #441 

## Checklist


felixarntz (approved):
Thank @bethanylang!

We may also want to replace the entry point links for the Performance Team to be `https://make.wordpress.org/performance/` instead of `https://make.wordpress.org/core/tag/performance/` (the latter is still relevant, but no longer as the main entry point). We can probably wait a bit more with that though, until that site has some content.

jeffpaul (approved):
Looks to be all the spots I see to update as well, nice update!

docs/Proposing-a-new-module.md
```diff
@@ -18,10 +18,10 @@ Proposed or upcoming WordPress core features are also called “[feature project
...
 
 ## Requirements for a new module
-Based on the above lifecycle for a module, there are a few initial requirements for proposing a new module. The performance team would like to ensure that modules inside the Performance Lab plugin keep moving forward over time, and the requirements outlined here help set the foundation.
+Based on the above lifecycle for a module, there are a few initial requirements for proposing a new module. The Performance Team would like to ensure that modules inside the Performance Lab plugin keep moving forward over time, and the requirements outlined here help set the foundation.
```
mukeshpanchal27, line 24: Can we use lowercase text for `Performance Team` as we used before?

mukeshpanchal27 (approved):
LGTM. One question regarding the text case.

docs/Proposing-a-new-module.md
mxbclang: @mukeshpanchal27 I think having it lowercase before was actually an oversight; we should always use the capital case Performance Team (or previously, Performance Group).

mukeshpanchal27: OK, Thanks for the clerification.

mxbclang:
@felixarntz Circling back on this one as I noticed these language updates again with today's release.  :)

felixarntz:
@bethanylang There was still my feedback from https://github.com/WordPress/performance/pull/442#pullrequestreview-1042105714 which was missing, but I just added this now in https://github.com/WordPress/performance/pull/442/commits/7fedc96fdecd0865309ff536188c187817b0ec35. This way we point users to the correct place `https://make.wordpress.org/performance/` instead of `https://make.wordpress.org/core/tag/performance/` when referring to the Performance Team.

felixarntz (approved):

PR #443, opened 2022-07-20
Enhance JS replacement mechanism for WebP to JPEG to more reliably replace full file name

### LINKED ISSUE #426: Enhance JS replacement mechanism for WebP to JPEG to more reliably replace full file name
Follow-up to #341: The implementation from that PR #360 which should replace WebP file URLs with their corresponding JPEG file URLs is currently a bit simplified, replacing purely the `.webp` extension with the corresponding `image/jpeg` file's extension.

While this (correctly) _does_ take the JPEG extension into account (which could be `jpeg`, `jpg`, or `jpe`), there are potentially other quirks with file names that may not always match. #358 has ensured that a WebP file generated would never be for another JPEG file and thus would never have a different name by default, which makes this a non-urgent issue, but still the current implementation is not quite reliable and error-prone, should the behavior of additional MIME type generation change in the future.

A potential approach to fix this would be to change the JS implementation as follows:
* Instead of just replacing all `.webp` extensions with the corresponding `image/jpeg` file's extension, the replacement should happen individually for the full/original image size as well as for all sub-sizes.
* For every size, the full file URL to the `image/webp` version should be replaced with the full file URL to the `image/jpeg` version, or at least the full file name should be replaced instead of only the extension.
* The most important here though is to do it per size individually since there may be slight differences between the files for those as well. We shouldn't only rely on the full image to determine how to replace all URLs.

This originally was raised in https://github.com/WordPress/performance/pull/360#discussion_r915453955. cc @peterwilsoncc

mukeshpanchal27:
## Summary

In this PR I use `media[i].media_details.sizes` to check if the image size has an additional mime type or not. I also use `webp_uploads_get_image_sizes_additional_mime_type_support()` to get only allowed sizes, so we don't need to check for other sizes.

The `console.log` for the image data.

image

Fixes #426

## Checklist

### DIFF
--- a/modules/images/webp-uploads/fallback.js
+++ b/modules/images/webp-uploads/fallback.js
@@ -8,8 +8,14 @@ window.wpPerfLab = window.wpPerfLab || {};
 					media_details = image.media_details,
 					media_sources = media_details.sources,
 					sizes         = media_details.sizes,
-					sizes_keys    = Object.keys( sizes ),
-					images        = document.querySelectorAll( 'img.wp-image-' + image.id );
+					sizes_keys    = Object.keys( sizes );
+
+				// If the full image has no JPEG version available, no sub-size will have JPEG available either.
+				if ( sizes.full && ! sizes.full.sources['image/jpeg'] ) {
+					continue;
+				}
+
+				var images = document.querySelectorAll( 'img.wp-image-' + image.id );
 
 				for ( var j = 0; j < images.length; j++ ) {
 
@@ -20,11 +26,6 @@ window.wpPerfLab = window.wpPerfLab || {};
 						src = src.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
 						images[ j ].setAttribute( 'src', src );
 						break;
-					}
-
-					// If the full image has no JPEG version available, no sub-size will have JPEG available either.
-					if ( ! sizes.full.sources['image/jpeg'] ) {
-						continue;
 					}
 
 					var srcset = images[ j ].getAttribute( 'srcset' );
--- a/modules/images/webp-uploads/fallback.js
+++ b/modules/images/webp-uploads/fallback.js
@@ -21,11 +21,14 @@ window.wpPerfLab = window.wpPerfLab || {};
 
 					var src = images[ j ].src;
 
-					// If there are no sizes, then replace src through sources, there is nothing more to replace.
-					if ( media_sources && ! sizes.full ) {
-						src = src.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
-						images[ j ].setAttribute( 'src', src );
-						break;
+					// If there are sources but no sizes, then attempt to replace src through sources. In that case, there is nothing more to replace.
+					if ( media_sources && ! sizes_keys.length ) {
+						// Only modify src if available and the relevant sources are set.
+						if ( src && media_sources['image/webp'] && media_sources['image/jpeg'] ) {
+							src = src.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
+							images[ j ].setAttribute( 'src', src );
+						}
+						continue;
 					}
 
 					var srcset = images[ j ].getAttribute( 'srcset' );
--- a/modules/images/webp-uploads/fallback.js
+++ b/modules/images/webp-uploads/fallback.js
@@ -4,35 +4,44 @@ window.wpPerfLab = window.wpPerfLab || {};
 	window.wpPerfLab.webpUploadsFallbackWebpImages = function( media ) {
 		for ( var i = 0; i < media.length; i++ ) {
 			try {
+				var image         = media[ i ],
+					media_details = image.media_details,
+					media_sources = media_details.sources,
+					sizes         = media_details.sizes,
+					sizes_keys    = Object.keys( sizes ),
+					images        = document.querySelectorAll( 'img.wp-image-' + image.id );
 
-				var images = document.querySelectorAll( 'img.wp-image-' + media[ i ].id ),
-					media_sources = media[ i ].media_details.sources;
 				for ( var j = 0; j < images.length; j++ ) {
 
-					if ( ! media_sources || ! media_sources['image/jpeg'] ) {
+					var src    = images[ j ].src,
+						srcset = images[ j ].getAttribute( 'srcset' );
+
+					// If there are no sizes, then replace src through sources, there is nothing more to replace.
+					if ( media_sources ) {
+						src = src.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
+						images[ j ].setAttribute( 'src', src );
+						break;
+					}
+
+					// If the full image has no JPEG version available, no sub-size will have JPEG available either.
+					if ( ! sizes.full.sources['image/jpeg'] ) {
 						continue;
 					}
 
-					var srcset = images[ j ].getAttribute( 'srcset' ),
-						sizes = media[ i ].media_details.sizes,
-						sizes_keys = Object.keys( sizes ),
-						flag = true;
-
-					// If a full image is present in srcset, it should be updated.
-					if ( srcset && media_sources['image/webp'] ) {
-						srcset = srcset.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
-					}
-
 					for ( var k = 0; k < sizes_keys.length; k++ ) {
-						var media_sizes_sources = media[ i ].media_details.sizes[ sizes_keys[ k ] ].sources;
+						var media_sizes_sources = sizes[ sizes_keys[ k ] ].sources;
 						if ( ! media_sizes_sources || ! media_sizes_sources['image/webp'] || ! media_sizes_sources['image/jpeg'] ) {
 							continue;
 						}
 
 						// Check to see if the image src has any size set, then update it.
-						if ( flag && media_sizes_sources['image/webp'].source_url === images[ j ].src ) {
-							images[ j ].src = media_sizes_sources['image/jpeg'].source_url;
-							flag = false;
+						if ( media_sizes_sources['image/webp'].source_url === src ) {
+							src = media_sizes_sources['image/jpeg'].source_url;
+
+							// If there is no srcset and the src has been replaced, there is nothing more to replace.
+							if ( ! srcset ) {
+								break;
+							}
 						}
 
 						if ( srcset && media_sizes_sources['image/webp'] ) {
@@ -44,9 +53,8 @@ window.wpPerfLab = window.wpPerfLab || {};
 						images[ j ].setAttribute( 'srcset', srcset );
 					}
 
-					// If the src has not been updated, then update the image src with the sources.
-					if ( flag && media_sources['image/webp'] ) {
-						images[ j ].src = images[ j ].src.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
+					if ( src ) {
+						images[ j ].setAttribute( 'src', src );
 					}
 				}
 			} catch ( e ) {
--- a/modules/images/webp-uploads/fallback.js
+++ b/modules/images/webp-uploads/fallback.js
@@ -4,46 +4,49 @@ window.wpPerfLab = window.wpPerfLab || {};
 	window.wpPerfLab.webpUploadsFallbackWebpImages = function( media ) {
 		for ( var i = 0; i < media.length; i++ ) {
 			try {
-				var images = document.querySelectorAll( 'img.wp-image-' + media[i].id );
+
+				var images = document.querySelectorAll( 'img.wp-image-' + media[ i ].id ),
+					media_sources = media[ i ].media_details.sources;
 				for ( var j = 0; j < images.length; j++ ) {
 
-					if ( ! media[i].media_details.sources || ! media[i].media_details.sources['image/jpeg'] ) {
+					if ( ! media_sources || ! media_sources['image/jpeg'] ) {
 						continue;
 					}
 
-					var sizes = media[i].media_details.sizes;
-					var sizes_keys = Object.keys( sizes );
-					var srcset = images[j].getAttribute( 'srcset' );
+					var srcset = images[ j ].getAttribute( 'srcset' ),
+						sizes = media[ i ].media_details.sizes,
+						sizes_keys = Object.keys( sizes ),
+						flag = true;
 
 					// If a full image is present in srcset, it should be updated.
-					if( srcset && media[i].media_details.sources['image/webp'] ) {
-						srcset = srcset.replace( media[i].media_details.sources['image/webp'].file, media[i].media_details.sources['image/jpeg'].file );
+					if ( srcset && media_sources['image/webp'] ) {
+						srcset = srcset.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
 					}
 
-					var flag = true;
 					for ( var k = 0; k < sizes_keys.length; k++ ) {
-						if( ! media[i].media_details.sizes[sizes_keys[k]].sources || ! media[i].media_details.sizes[sizes_keys[k]].sources['image/webp'] || ! media[i].media_details.sizes[sizes_keys[k]].sources['image/jpeg'] ) {
+						var media_sizes_sources = media[ i ].media_details.sizes[ sizes_keys[ k ] ].sources;
+						if ( ! media_sizes_sources || ! media_sizes_sources['image/webp'] || ! media_sizes_sources['image/jpeg'] ) {
 							continue;
 						}
 
 						// Check to see if the image src has any size set, then update it.
-						if( flag && media[i].media_details.sizes[sizes_keys[k]].sources['image/webp'].source_url === images[j].src ) {
-							images[j].src = media[i].media_details.sizes[sizes_keys[k]].sources['image/jpeg'].source_url;
+						if ( flag && media_sizes_sources['image/webp'].source_url === images[ j ].src ) {
+							images[ j ].src = media_sizes_sources['image/jpeg'].source_url;
 							flag = false;
 						}
 
-						if( srcset && media[i].media_details.sizes[sizes_keys[k]].sources['image/webp'] ) {
-							srcset = srcset.replace( media[i].media_details.sizes[sizes_keys[k]].sources['image/webp'].source_url, media[i].media_details.sizes[sizes_keys[k]].sources['image/jpeg'].source_url );
+						if ( srcset && media_sizes_sources['image/webp'] ) {
+							srcset = srcset.replace( media_sizes_sources['image/webp'].source_url, media_sizes_sources['image/jpeg'].source_url );
 						}
 					}
 
-					if( srcset ) {
-						images[j].setAttribute( 'srcset', srcset );
+					if ( srcset ) {
+						images[ j ].setAttribute( 'srcset', srcset );
 					}
 
 					// If the src has not been updated, then update the image src with the sources.
-					if( flag && media[i].media_details.sources['image/webp'] ) {
-						images[j].src = images[j].src.replace( media[i].media_details.sources['image/webp'].file, media[i].media_details.sources['image/jpeg'].file );
+					if ( flag && media_sources['image/webp'] ) {
+						images[ j ].src = images[ j ].src.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
 					}
 				}
 			} catch ( e ) {
@@ -56,7 +59,7 @@ window.wpPerfLab = window.wpPerfLab || {};
 	var loadMediaDetails = function( nodes ) {
 		var ids = [];
 		for ( var i = 0; i < nodes.length; i++ ) {
-			var node = nodes[i];
+			var node = nodes[ i ];
 			var srcset = node.getAttribute( 'srcset' ) || '';
 
 			if (
@@ -91,7 +94,7 @@ window.wpPerfLab = window.wpPerfLab || {};
 		// Start the mutation observer to update images added dynamically.
 		var observer = new MutationObserver( function( mutationList ) {
 			for ( var i = 0; i < mutationList.length; i++ ) {
-				loadMediaDetails( mutationList[i].addedNodes );
+				loadMediaDetails( mutationList[ i ].addedNodes );
 			}
 		} );
 	
--- a/modules/images/webp-uploads/fallback.js
+++ b/modules/images/webp-uploads/fallback.js
@@ -9,7 +9,6 @@ window.wpPerfLab = window.wpPerfLab || {};
 
 				var images = document.querySelectorAll( 'img.wp-image-' + media[i].id );
 				for ( var j = 0; j < images.length; j++ ) {
-					console.log( media[i] );
 					var srcset = images[j].getAttribute( 'srcset' );
 					if ( srcset ) {
 						// Update full image in srcset.
--- a/modules/images/webp-uploads/fallback.js
+++ b/modules/images/webp-uploads/fallback.js
@@ -4,21 +4,32 @@ window.wpPerfLab = window.wpPerfLab || {};
 	window.wpPerfLab.webpUploadsFallbackWebpImages = function( media ) {
 		for ( var i = 0; i < media.length; i++ ) {
 			try {
-				if ( ! media[i].media_details.sources || ! media[i].media_details.sources['image/jpeg'] ) {
-					continue;
-				}
-
-				var ext = media[i].media_details.sources['image/jpeg'].file.match( /\.\w+$/i );
-				if ( ! ext || ! ext[0] ) {
-					continue;
-				}
+				var webp_sizes = window.wpPerfLabWebpSizes;
+				var size_keys = Object.keys( webp_sizes );
 
 				var images = document.querySelectorAll( 'img.wp-image-' + media[i].id );
 				for ( var j = 0; j < images.length; j++ ) {
-					images[j].src = images[j].src.replace( /\.webp$/i, ext[0] );
+					console.log( media[i] );
 					var srcset = images[j].getAttribute( 'srcset' );
 					if ( srcset ) {
-						images[j].setAttribute( 'srcset', srcset.replace( /\.webp(\s)/ig, ext[0] + '$1' ) );
+						// Update full image in srcset.
+						srcset = srcset.replace( images[j].src, media[i].media_details.sizes.full.source_url );
+						images[j].setAttribute( 'srcset', srcset );
+
+						// Update full image in src.
+						images[j].src = media[i].media_details.sizes.full.source_url;
+
+						// Update thumbnail images.
+						for ( var k = 0; k < size_keys.length; k++ ) {
+
+							if( ! media[i].media_details.sizes[size_keys[k]].sources['image/webp'] ) {
+								continue;
+							}
+
+							srcset = srcset.replace( media[i].media_details.sizes[size_keys[k]].sources['image/webp'].source_url, media[i].media_details.sizes[size_keys[k]].sources['image/jpeg'].source_url );
+							images[j].setAttribute( 'srcset', srcset );
+						}
+
 					}
 				}
 			} catch ( e ) {
--- a/modules/images/webp-uploads/fallback.js
+++ b/modules/images/webp-uploads/fallback.js
@@ -4,31 +4,46 @@ window.wpPerfLab = window.wpPerfLab || {};
 	window.wpPerfLab.webpUploadsFallbackWebpImages = function( media ) {
 		for ( var i = 0; i < media.length; i++ ) {
 			try {
-				var webp_sizes = window.wpPerfLabWebpSizes;
-				var size_keys = Object.keys( webp_sizes );
-
 				var images = document.querySelectorAll( 'img.wp-image-' + media[i].id );
 				for ( var j = 0; j < images.length; j++ ) {
+
+					if ( ! media[i].media_details.sources || ! media[i].media_details.sources['image/jpeg'] ) {
+						continue;
+					}
+
+					var sizes = media[i].media_details.sizes;
+					var sizes_keys = Object.keys( sizes );
 					var srcset = images[j].getAttribute( 'srcset' );
-					if ( srcset ) {
-						// Update full image in srcset.
-						srcset = srcset.replace( images[j].src, media[i].media_details.sizes.full.source_url );
-						images[j].setAttribute( 'srcset', srcset );
 
-						// Update full image in src.
-						images[j].src = media[i].media_details.sizes.full.source_url;
+					// If a full image is present in srcset, it should be updated.
+					if( srcset && media[i].media_details.sources['image/webp'] ) {
+						srcset = srcset.replace( media[i].media_details.sources['image/webp'].file, media[i].media_details.sources['image/jpeg'].file );
+					}
 
-						// Update thumbnail images.
-						for ( var k = 0; k < size_keys.length; k++ ) {
-
-							if( ! media[i].media_details.sizes[size_keys[k]].sources['image/webp'] ) {
-								continue;
-							}
-
-							srcset = srcset.replace( media[i].media_details.sizes[size_keys[k]].sources['image/webp'].source_url, media[i].media_details.sizes[size_keys[k]].sources['image/jpeg'].source_url );
-							images[j].setAttribute( 'srcset', srcset );
+					var flag = true;
+					for ( var k = 0; k < sizes_keys.length; k++ ) {
+						if( ! media[i].media_details.sizes[sizes_keys[k]].sources || ! media[i].media_details.sizes[sizes_keys[k]].sources['image/webp'] || ! media[i].media_details.sizes[sizes_keys[k]].sources['image/jpeg'] ) {
+							continue;
 						}
 
+						// Check to see if the image src has any size set, then update it.
+						if( flag && media[i].media_details.sizes[sizes_keys[k]].sources['image/webp'].source_url === images[j].src ) {
+							images[j].src = media[i].media_details.sizes[sizes_keys[k]].sources['image/jpeg'].source_url;
+							flag = false;
+						}
+
+						if( srcset && media[i].media_details.sizes[sizes_keys[k]].sources['image/webp'] ) {
+							srcset = srcset.replace( media[i].media_details.sizes[sizes_keys[k]].sources['image/webp'].source_url, media[i].media_details.sizes[sizes_keys[k]].sources['image/jpeg'].source_url );
+						}
+					}
+
+					if( srcset ) {
+						images[j].setAttribute( 'srcset', srcset );
+					}
+
+					// If the src has not been updated, then update the image src with the sources.
+					if( flag && media[i].media_details.sources['image/webp'] ) {
+						images[j].src = images[j].src.replace( media[i].media_details.sources['image/webp'].file, media[i].media_details.sources['image/jpeg'].file );
 					}
 				}
 			} catch ( e ) {
--- a/modules/images/webp-uploads/fallback.js
+++ b/modules/images/webp-uploads/fallback.js
@@ -13,11 +13,10 @@ window.wpPerfLab = window.wpPerfLab || {};
 
 				for ( var j = 0; j < images.length; j++ ) {
 
-					var src    = images[ j ].src,
-						srcset = images[ j ].getAttribute( 'srcset' );
+					var src = images[ j ].src;
 
 					// If there are no sizes, then replace src through sources, there is nothing more to replace.
-					if ( media_sources ) {
+					if ( media_sources && ! sizes.full ) {
 						src = src.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
 						images[ j ].setAttribute( 'src', src );
 						break;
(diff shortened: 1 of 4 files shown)


modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,32 @@ window.wpPerfLab = window.wpPerfLab || {};
...
-				if ( ! ext || ! ext[0] ) {
-					continue;
-				}
+				var webp_sizes = window.wpPerfLabWebpSizes;
+				var size_keys = Object.keys( webp_sizes );
 
 				var images = document.querySelectorAll( 'img.wp-image-' + media[i].id );
 				for ( var j = 0; j < images.length; j++ ) {
-					images[j].src = images[j].src.replace( /\.webp$/i, ext[0] );
+					console.log( media[i] );
```
mehulkaklotar: @mukeshpanchal27 you might want to remove `console.log` from here?

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,32 @@ window.wpPerfLab = window.wpPerfLab || {};
...
 					var srcset = images[j].getAttribute( 'srcset' );
 					if ( srcset ) {
-						images[j].setAttribute( 'srcset', srcset.replace( /\.webp(\s)/ig, ext[0] + '$1' ) );
+						// Update full image in srcset.
+						srcset = srcset.replace( images[j].src, media[i].media_details.sizes.full.source_url );
```
mehulkaklotar: Here the we should check the `media[i].media_details.sizes.full.source_url` is set or not to avoid undefined errors. And we should get the output source url into variable and then use them later in the code.

modules/images/webp-uploads/load.php
```diff
@@ -675,7 +675,10 @@ function webp_uploads_wepb_fallback() {
 	<?php
 	$javascript = ob_get_clean();
 
+	$sizes_with_mime_type_support = webp_uploads_get_image_sizes_additional_mime_type_support();
+
 	wp_print_inline_script_tag(
+		sprintf( 'window.wpPerfLabWebpSizes = %s;', wp_json_encode( array_filter( $sizes_with_mime_type_support ) ) ) . "\n" .
```
felixarntz: This shouldn't be needed. We don't need to worry for the frontend replacement for which sizes _should_ have additional MIME types. The only thing that matters in the frontend is whether the current attachment in fact has the MIME type versions available or not.

modules/images/webp-uploads/fallback.js
felixarntz: The `src` does not always have the "full" size, so this doesn't look correct to me. We always need to replace the `image/webp` version for each size with the `image/jpeg` version if it exists.

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,32 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+						images[j].setAttribute( 'srcset', srcset );
+
+						// Update full image in src.
+						images[j].src = media[i].media_details.sizes.full.source_url;
```
felixarntz: I think all of this shouldn't be needed. The REST API includes the "full" size in its response under `sizes`, so the following `for` loop should be able to cover all of it.

Also, see my above comment, we can't unconditionally assume that `src` always has the "full" size.

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,32 @@ window.wpPerfLab = window.wpPerfLab || {};
...
-				if ( ! media[i].media_details.sources || ! media[i].media_details.sources['image/jpeg'] ) {
-					continue;
-				}
-
-				var ext = media[i].media_details.sources['image/jpeg'].file.match( /\.\w+$/i );
-				if ( ! ext || ! ext[0] ) {
-					continue;
-				}
+				var webp_sizes = window.wpPerfLabWebpSizes;
+				var size_keys = Object.keys( webp_sizes );
```
felixarntz: See my other comment. We should iterate through the `sizes` from the REST API response and check for each one whether both a `image/webp` and `image/jpeg` version available, and if so we can do the replacement.

We don't need to know which sizes _should_ have a WebP version available here.

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,32 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+						srcset = srcset.replace( images[j].src, media[i].media_details.sizes.full.source_url );
+						images[j].setAttribute( 'srcset', srcset );
+
+						// Update full image in src.
+						images[j].src = media[i].media_details.sizes.full.source_url;
+
+						// Update thumbnail images.
+						for ( var k = 0; k < size_keys.length; k++ ) {
+
+							if( ! media[i].media_details.sizes[size_keys[k]].sources['image/webp'] ) {
```
felixarntz: We need to also check for whether that size has an `image/jpeg` version, which is not always the case.

In addition, `sources` may not be set at all for older images, so we need to check for that too to prevent JS errors.
```suggestion
							if( ! media[i].media_details.sizes[size_keys[k]].sources || ! media[i].media_details.sizes[size_keys[k]].sources['image/webp'] || ! media[i].media_details.sizes[size_keys[k]].sources['image/jpeg'] ) {
```

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Removed

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,32 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+								continue;
+							}
+
+							srcset = srcset.replace( media[i].media_details.sizes[size_keys[k]].sources['image/webp'].source_url, media[i].media_details.sizes[size_keys[k]].sources['image/jpeg'].source_url );
+							images[j].setAttribute( 'srcset', srcset );
```
felixarntz: In addition to replacing the URL in the `srcset`, we also need to see if the WebP file for this size is used for the `src`, and if so, we need to update the `src` with the JPEG version. See my related above comments, `src` is not always the "full" size.

felixarntz (changes requested):
@mukeshpanchal27 Left some feedback. Most importantly, we need to rely on the actually available MIME types for each size rather than on the filter value for which sizes _should_ have additional MIME types.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thank you for the review. I have updated it. Please re-review it.

modules/images/webp-uploads/load.php
mukeshpanchal27: Thank you, Felix, for the review and details. We only allow some sizes for additional mime types, so we should check only those sizes, not all, because other sizes do not have WebP images.

I have updated a PR that removes this code and checks whether the current attachment has the MIME type versions available or not.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thank you for pointing this out. Check all the different image sizes for the image src.

I set a flag to determine if the image has a different size than full. We update the image src in the sizes `for` loop, and we don't need to update the src again.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: makes sense. Removed check.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thank you for the review. The PR was updated with the logic.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: 👍

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thank you for the review. Please check my other comments for review.

mehulkaklotar (approved):

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,46 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+
+					var sizes = media[i].media_details.sizes;
+					var sizes_keys = Object.keys( sizes );
 					var srcset = images[j].getAttribute( 'srcset' );
-					if ( srcset ) {
-						images[j].setAttribute( 'srcset', srcset.replace( /\.webp(\s)/ig, ext[0] + '$1' ) );
+
+					// If a full image is present in srcset, it should be updated.
+					if( srcset && media[i].media_details.sources['image/webp'] ) {
+						srcset = srcset.replace( media[i].media_details.sources['image/webp'].file, media[i].media_details.sources['image/jpeg'].file );
```
adamsilverstein: nitpick - i believe core JS coding standards still specify spaces around variables in brackets, eg `media[ i ]`

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,46 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+					}
+
+					var flag = true;
+					for ( var k = 0; k < sizes_keys.length; k++ ) {
+						if( ! media[i].media_details.sizes[sizes_keys[k]].sources || ! media[i].media_details.sizes[sizes_keys[k]].sources['image/webp'] || ! media[i].media_details.sizes[sizes_keys[k]].sources['image/jpeg'] ) {
```
adamsilverstein: media[i].media_details.sizes[sizes_keys[k]].sources is used repeatedly in this block, I would set a variable to this value at the top and use that instead

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,46 @@ window.wpPerfLab = window.wpPerfLab || {};
...
-					if ( srcset ) {
-						images[j].setAttribute( 'srcset', srcset.replace( /\.webp(\s)/ig, ext[0] + '$1' ) );
+
+					// If a full image is present in srcset, it should be updated.
+					if( srcset && media[i].media_details.sources['image/webp'] ) {
+						srcset = srcset.replace( media[i].media_details.sources['image/webp'].file, media[i].media_details.sources['image/jpeg'].file );
+					}
+
+					var flag = true;
+					for ( var k = 0; k < sizes_keys.length; k++ ) {
```
adamsilverstein, line 36: if there is no srcset, we can just search for the img src and skip the rest of this loop

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,46 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+						continue;
+					}
+
+					var sizes = media[i].media_details.sizes;
+					var sizes_keys = Object.keys( sizes );
 					var srcset = images[j].getAttribute( 'srcset' );
-					if ( srcset ) {
-						images[j].setAttribute( 'srcset', srcset.replace( /\.webp(\s)/ig, ext[0] + '$1' ) );
+
+					// If a full image is present in srcset, it should be updated.
```
adamsilverstein: if the full image is the img src url and there is no srcset we should be done here

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thank you for pointing out these changes. The PR does not validate JS coding standards. Do we need to fix it? Can I raise the issue of tracking it?

The PR was updated with JS coding standards.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: It makes sense. Set variables for the main sources `media_sources = media[ i ].media_details.sources;` as well as size sources `media_sizes_sources = media[ i ].media_details.sizes[ sizes_keys[ k ] ].sources;`

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thank you for the review.

- If there is no `srcset` then the script will skip the search we check [the condition](https://github.com/WordPress/performance/blob/fix/426-enhance-js-replacement-mechanism/modules/images/webp-uploads/fallback.js#L38-L40) for all srcset data. 

- The `flag` is set to `false` once the img src is updated and it will skip the checking again for other sizes.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thank you for the review. Can you please share more information about this?

- Currently, we update IMG src after the sizes checking because if we update image src then we can't replace the full image src in srcset.
``````

File: modules/images/webp-uploads/fallback.js [hunk 12]
```diff
@@ -4,21 +4,49 @@ window.wpPerfLab = window.wpPerfLab || {};
...
-				var images = document.querySelectorAll( 'img.wp-image-' + media[i].id );
+				var images = document.querySelectorAll( 'img.wp-image-' + media[ i ].id ),
+					media_sources = media[ i ].media_details.sources;
 				for ( var j = 0; j < images.length; j++ ) {
-					images[j].src = images[j].src.replace( /\.webp$/i, ext[0] );
-					var srcset = images[j].getAttribute( 'srcset' );
+
+					if ( ! media_sources || ! media_sources['image/jpeg'] ) {
+						continue;
+					}
```
felixarntz: Great idea! If there is no JPEG for the `full` image, it means no image for that attachment will have a JPEG, as JPEGs are always created unless it is disabled for the entire attachment (e.g. if a site is configured to only generate WebP).

Would be good to add a comment though to explain this, otherwise it may not make sense to someone without the context.
```suggestion
					// If the full image has no JPEG version available, no sub-size will have JPEG available either.
					if ( ! media_sources || ! media_sources['image/jpeg'] ) {
						continue;
					}
```

**Update:** Based on my other feedback, once the REST API implementation has been updated, we should rather check `media[ i ].media_details.sizes.full.sources['image/jpeg']` here.

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,49 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+
+					var srcset = images[ j ].getAttribute( 'srcset' ),
+						sizes = media[ i ].media_details.sizes,
```
felixarntz: Let's assign this variable outside of this `for` loop for a bit more efficiency. It is the same for every value in `images`, so we can set this variable once, e.g. in the same place where you're setting `media_sources`.

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,49 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+						sizes_keys = Object.keys( sizes ),
```
felixarntz: Same here, this is based on `sizes`, so it can be set further above outside of this `for` loop for `images`.

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,49 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+						flag = true;
+
+					// If a full image is present in srcset, it should be updated.
+					if ( srcset && media_sources['image/webp'] ) {
+						srcset = srcset.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
+					}
```
felixarntz: This code makes sense the way our REST API response is currently structured, however I just checked that again and found out our REST API implementation is inconsistent with core's:
* WordPress core returns the data for the `full` image within `media_details.sizes` as opposed to a separate field. When you review the WordPress core implementation of the REST endpoint, you will see that it adds the `full` image data to `media_details.sizes`. This is different from how the metadata itself is structured, but we should stick to the convention that core has established here.
* So we should probably monitor this in our REST API implementation and only provide the `full` image's `sources` within that value too, instead of `media_details.sources`.
* Let's adjust our implementation in the `rest-api.php` file so that it also includes the correct `sources` for the `full` image size.

Once we have made that change, we will no longer need this extra call here. We can simply loop through `sizes_keys` like below, because that will then include the `full` size as expected. It should notably simplify the logic here.

modules/images/webp-uploads/fallback.js
felixarntz: I think what @adamsilverstein is saying is we could check if the `src` is the `full` size image and no `srcset` attribute exists, we don't even need the loop below.

However, please see my other comment on the inconsistency in our REST API implementation. `full` is treated just like any other image size by the REST API endpoint for attachments in core, so we should do that too. This then also means we shouldn't have any logic that is "special" for the full size anymore in this JS code, since it will be like any other size.

modules/images/webp-uploads/fallback.js
felixarntz: +1 to what @adamsilverstein said. Trying to clarify (the way I understand it):
* First of all, we should probably store `src` in a variable, similar to how we store `srcset` in a variable, before the `sizes_keys` loop.
* In the if clause where we adjust the `src` attribute, we can then check if there is no `srcset` attribute, and if so we can completely `break` out of the loop. Because if there is only `src` and no `srcset` and we have already replaced the `src`, we know for sure that there is no more image URL to replace.

Let me know if that makes more sense.

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,49 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+						var media_sizes_sources = media[ i ].media_details.sizes[ sizes_keys[ k ] ].sources;
+						if ( ! media_sizes_sources || ! media_sizes_sources['image/webp'] || ! media_sizes_sources['image/jpeg'] ) {
+							continue;
+						}
+
+						// Check to see if the image src has any size set, then update it.
+						if ( flag && media_sizes_sources['image/webp'].source_url === images[ j ].src ) {
+							images[ j ].src = media_sizes_sources['image/jpeg'].source_url;
+							flag = false;
+						}
```
felixarntz: I don't think we really need the `flag` variable at all. We can simply check if the URL per size is the same as `src` and if so replace it. Once we have replaced it, it'll never be the same like any other URL to go through anyway, and that's a cheap enough comparison to make.

Also see my above reply to @adamsilverstein's comment: We should break here if there's no `srcset` attribute to make the loop more efficient, and I think it would be most straightforward to handle `src` in a variable as well.

Something like this:
```suggestion
						if ( media_sizes_sources['image/webp'].source_url === src ) {
							src = media_sizes_sources['image/jpeg'].source_url;
							// If there is no srcset and the src has been replaced, there is nothing more to replace.
							if ( ! srcset ) {
								break;
							}
						}
```

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,49 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+
 					if ( srcset ) {
-						images[j].setAttribute( 'srcset', srcset.replace( /\.webp(\s)/ig, ext[0] + '$1' ) );
+						images[ j ].setAttribute( 'srcset', srcset );
+					}
+
+					// If the src has not been updated, then update the image src with the sources.
+					if ( flag && media_sources['image/webp'] ) {
+						images[ j ].src = images[ j ].src.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
 					}
```
felixarntz: See above, based on my suggested change to the REST API response to be consistent with core, we will afterwards no longer need any special treatment for the `full` size, so this code can be removed. We also don't need the flag when we simply store the `src` in a variable.

You could then have something here exactly like you have right above for `srcset`:
```js
if ( src ) {
	images[ j ].setAttribute( 'src', src );
}
```

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,49 @@ window.wpPerfLab = window.wpPerfLab || {};
...
-					var srcset = images[j].getAttribute( 'srcset' );
+
+					if ( ! media_sources || ! media_sources['image/jpeg'] ) {
+						continue;
+					}
+
+					var srcset = images[ j ].getAttribute( 'srcset' ),
+						sizes = media[ i ].media_details.sizes,
+						sizes_keys = Object.keys( sizes ),
+						flag = true;
```
felixarntz: See my comments below, based on those I don't think we even need a flag like this.

felixarntz (commented):
@mukeshpanchal27 Thanks for the updates. The approach here now looks good, however I think there's in fact another underlying problem that makes the implementation here overly complicated: The plugin's implementation of the REST API response integration is inconsistent with the core implementation of the endpoint - we need to also add the `sources` to the `full` size entry within `media_details.sizes`. See details in my comments below.

That is crucial for the core implementation too and should be straightforward, so it would be great if you could make this change in the `rest-api.php` file as part of this pull request. From there, we can simplify the JS code to no longer have any special treatment for the `full` size, as it should be handled just like any other size.

modules/images/webp-uploads/fallback.js
adamsilverstein: 👍🏼 thanks for clarifying that is exactly what I meant @felixarntz

mukeshpanchal27: Thank you for explaining 😄 Adam I get your point now.

This feedback is addressed.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thank you for more information, Felix

This feedback is addressed with the REST API endpoint changes.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thank you for the review. This feedback is addressed.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thank you for the review. This feedback is addressed.

Additional to this I found one case in which the REST API endpoint does not return `sizes`. If you upload an image with 100X100 size then we don't get `sizes` in `media_details` to tackle this case:

- Check if `sources` are present in `media_details` then we don't have `sizes` for that image in that case update image `src` only and break the loop https://github.com/WordPress/performance/blob/fix/426-enhance-js-replacement-mechanism/modules/images/webp-uploads/fallback.js#L19-L24
- We have unset `sources` if we have the full size in `sizes` https://github.com/WordPress/performance/blob/fix/426-enhance-js-replacement-mechanism/modules/images/webp-uploads/rest-api.php#L49

Please review it and share your feedback for this implementation.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thanks for the review and guidance. It makes sense to make these changes.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thank you. This feedback is addressed.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thank you. This feedback is addressed.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thank you. This feedback is addressed.

modules/images/webp-uploads/rest-api.php
```diff
@@ -23,20 +23,30 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
 	}
 
 	$metadata = wp_get_attachment_metadata( $post->ID );
```
felixarntz: Another minor improvement here: I think we don't even need to get the attachment metadata at all in this function, since `$data['media_details']` already has exactly that.

modules/images/webp-uploads/rest-api.php
```diff
@@ -23,20 +23,30 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
-	foreach ( $data['media_details']['sizes'] as $size => $details ) {
+	foreach ( $data['media_details']['sizes'] as $size => &$details ) {
 		if ( empty( $metadata['sizes'][ $size ]['sources'] ) || ! is_array( $metadata['sizes'][ $size ]['sources'] ) ) {
```
felixarntz: See above, this could then become:
```suggestion
		if ( empty( $details['sources'] ) || ! is_array( $details['sources'] ) ) {
```

modules/images/webp-uploads/rest-api.php
```diff
@@ -23,20 +23,30 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
 			continue;
 		}
-
 		$sources   = array();
```
felixarntz: This variable is unnecessary I think.
```suggestion
```

modules/images/webp-uploads/rest-api.php
```diff
@@ -23,20 +23,30 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
-		$directory = dirname( $data['media_details']['sizes'][ $size ]['source_url'] );
+		$directory = dirname( $details['source_url'] );
 		foreach ( $metadata['sizes'][ $size ]['sources'] as $mime => $mime_details ) {
```
felixarntz: See above, this could then become:
```suggestion
		foreach ( $details['sources'] as $mime => &$mime_details ) {
```

modules/images/webp-uploads/rest-api.php
```diff
@@ -23,20 +23,30 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+			$source_url                  = "{$directory}/{$mime_details['file']}";
+			$mime_details['source_url']  = $source_url;
+			$details['sources'][ $mime ] = $mime_details;
```
felixarntz: And this could be simplified accordingly:
```suggestion
			$mime_details['source_url']  = "{$directory}/{$mime_details['file']}";
```

modules/images/webp-uploads/rest-api.php
```diff
@@ -23,20 +23,30 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+		}
+	}
+
+	$full_src = wp_get_attachment_image_src( $post->ID, 'full' );
+
+	if ( ! empty( $full_src ) ) {
```
felixarntz: We should also check here whether all other necessary data is set, just to be safe:
```suggestion
	if ( ! empty( $full_src ) && ! empty( $data['media_details']['sources'] ) && ! empty( $data['media_details']['sizes']['full'] ) ) {
```
If any of this is not here for some reason, we also can't / shouldn't set it.

modules/images/webp-uploads/rest-api.php
```diff
@@ -23,20 +23,30 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
+		$directory = dirname( $full_src[0] );
+		foreach ( $data['media_details']['sources'] as $mime => &$mime_details ) {
 			$source_url                 = "{$directory}/{$mime_details['file']}";
 			$mime_details['source_url'] = $source_url;
```
felixarntz: One more unnecessary variable here:
```suggestion
			$mime_details['source_url'] = "{$directory}/{$mime_details['file']}";
```

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,57 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+					if ( media_sources ) {
+						src = src.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
+						images[ j ].setAttribute( 'src', src );
+						break;
+					}
+
+					// If the full image has no JPEG version available, no sub-size will have JPEG available either.
+					if ( ! sizes.full.sources['image/jpeg'] ) {
+						continue;
+					}
```
felixarntz: This check is not in the right place here: It isn't based on `images` at all, so we should check that right away, above the `images` loop: If the full size has no `image/jpeg` version available, then we don't even need to attempt replacement for this `image`.

Even better would be: Let's check this even before the `document.querySelectorAll` call, because if there is no JPEG, we don't even need to query the image elements for this attachment.

Last but not least: Note that the if check should first check that `sizes.full` is actually set.

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,57 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+
+					var src    = images[ j ].src,
+						srcset = images[ j ].getAttribute( 'srcset' );
+
+					// If there are no sizes, then replace src through sources, there is nothing more to replace.
+					if ( media_sources ) {
+						src = src.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
+						images[ j ].setAttribute( 'src', src );
+						break;
+					}
```
felixarntz: **Update:** Based on https://github.com/WordPress/performance/pull/443#discussion_r930959952, this is correct. However, we should really only use it if no sizes are set, so let's add `&& ! sizes` to the if clause around.

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,57 @@ window.wpPerfLab = window.wpPerfLab || {};
...
+						if ( media_sizes_sources['image/webp'].source_url === src ) {
+							src = media_sizes_sources['image/jpeg'].source_url;
+
+							// If there is no srcset and the src has been replaced, there is nothing more to replace.
+							if ( ! srcset ) {
+								break;
+							}
+						}
+
+						if ( srcset && media_sizes_sources['image/webp'] ) {
```
felixarntz: The one variable is already checked above, so the additional check here is unnecessary:
```suggestion
						if ( srcset ) {
```

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,57 @@ window.wpPerfLab = window.wpPerfLab || {};
...
-					continue;
-				}
-
-				var ext = media[i].media_details.sources['image/jpeg'].file.match( /\.\w+$/i );
-				if ( ! ext || ! ext[0] ) {
-					continue;
-				}
+				var image         = media[ i ],
+					media_details = image.media_details,
+					media_sources = media_details.sources,
```
felixarntz, line 9: See my other feedback: Now this variable shouldn't even be relevant anymore, so let's remove it. Based on the REST API response updates, it won't ever be part of the data.

felixarntz (commented):
@mukeshpanchal27 Thank you for the updates. The REST API logic looks right now, I only have some small comments on improving and simplifying the code there. The JS logic also looks good for the most part, there are just a few quirks left on removing references to the root `media_details.sources` and the part of assuming those sources apply to `src`.

modules/images/webp-uploads/fallback.js
felixarntz: @mukeshpanchal27 Ah, good point. Sorry, I only read this after my review. I will update my one comment below to reflect this, as I agree with your observation.

modules/images/webp-uploads/rest-api.php
mukeshpanchal27: Thank you for the review. Use `$data['media_details']` instead of `$metadata`.

modules/images/webp-uploads/rest-api.php
mukeshpanchal27: Thank you. This feedback is addressed.

modules/images/webp-uploads/rest-api.php
mukeshpanchal27: 👍

modules/images/webp-uploads/rest-api.php
mukeshpanchal27: Thank you. This feedback is addressed.

modules/images/webp-uploads/rest-api.php
mukeshpanchal27: Make sense 👍

modules/images/webp-uploads/rest-api.php
mukeshpanchal27: Here I use a similar approach that we used in Core Patch https://github.com/WordPress/wordpress-develop/pull/3025#issuecomment-1196901580.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: The REST API returns a `sizes` array without any sizes, so I have to use `&& ! sizes.full` Then it works.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: The https://github.com/WordPress/performance/pull/443/commits/c2a93ba3383381d3658ad28219b00a066639e892 address this feedback.

modules/images/webp-uploads/fallback.js
adamsilverstein: yes, it would be good to add linting for JS if that is missing

modules/images/webp-uploads/rest-api.php
```diff
@@ -22,21 +22,27 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
 		return $response;
 	}
 
-	$metadata = wp_get_attachment_metadata( $post->ID );
-	foreach ( $data['media_details']['sizes'] as $size => $details ) {
-		if ( empty( $metadata['sizes'][ $size ]['sources'] ) || ! is_array( $metadata['sizes'][ $size ]['sources'] ) ) {
+	foreach ( $data['media_details']['sizes'] as $size => &$details ) {
```
adamsilverstein, line 25: curious why you have the details value by reference here, is this changed below?

mukeshpanchal27: Yes, we are changing `source_url` here https://github.com/WordPress/performance/blob/fix/426-enhance-js-replacement-mechanism/modules/images/webp-uploads/rest-api.php#L33

adamsilverstein: that updates `$mime_details`, I don't see `$details` getting modified, it shouldn't need passing by reference?

adamsilverstein (commented):
left one more small question/point of feedback; this is ready for some more testing!

modules/images/webp-uploads/rest-api.php
mukeshpanchal27: Yes, it updates `$mime_details`. It is part of the `$details` value. Let me share a screenshot of the output with and without pass-by-reference.

With reference to `$details` - https://snipboard.io/CAabcf.jpg

Without reference to `$details` - https://snipboard.io/eGt2I5.jpg

Please correct me if I've made a mistake.

felixarntz: @adamsilverstein Both `$details` and `$mime_details` must be passed by reference, otherwise this logic breaks. Value variables in `foreach` loops will otherwise not get updated.

modules/images/webp-uploads/fallback.js
```diff
@@ -4,21 +4,59 @@ window.wpPerfLab = window.wpPerfLab || {};
...
-					var srcset = images[j].getAttribute( 'srcset' );
+
+					var src = images[ j ].src;
+
+					// If there are no sizes, then replace src through sources, there is nothing more to replace.
+					if ( media_sources && ! sizes.full ) {
+						src = src.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
+						images[ j ].setAttribute( 'src', src );
+						break;
+					}
```
felixarntz: There are a few issues here still:
* The `! sizes.full` check here doesn't look right: We don't care here whether there is no `full` size, it's about whether there are no sizes at all.
* We need to check whether the sources actually have `image/webp` and `image/jpeg` in them.
* We must not `break` here as we still need to process other potential `images`. So it needs to be `continue`.
* We must check here whether `src` is actually present before trying to modify it. (Even when it is not present though for whatever reason, we can still bail though, because `srcset` then definitely won't be present either.)

In summary:
```suggestion
					// If there are sources but no sizes, then attempt to replace src through sources. In that case, there is nothing more to replace.
					if ( media_sources && ! sizes_keys.length ) {
						// Only modify src if available and the relevant sources are set.
						if ( src && media_sources['image/webp'] && media_sources['image/jpeg'] ) {
							src = src.replace( media_sources['image/webp'].file, media_sources['image/jpeg'].file );
							images[ j ].setAttribute( 'src', src );
						}
						continue;
					}
```

modules/images/webp-uploads/rest-api.php
```diff
@@ -22,21 +22,27 @@ function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Pos
...
-		$sources   = array();
-		$directory = dirname( $data['media_details']['sizes'][ $size ]['source_url'] );
-		foreach ( $metadata['sizes'][ $size ]['sources'] as $mime => $mime_details ) {
-			$source_url                 = "{$directory}/{$mime_details['file']}";
-			$mime_details['source_url'] = $source_url;
-			$sources[ $mime ]           = $mime_details;
+		$directory = dirname( $details['source_url'] );
+		foreach ( $details['sources'] as $mime => &$mime_details ) {
+			$mime_details['source_url'] = "{$directory}/{$mime_details['file']}";
+		}
```
felixarntz: This code still needs to be changed to also rely on `wp_basename()` rather than `dirname()`, similar to https://core.trac.wordpress.org/changeset/53786.
```suggestion
		$image_url_basename = wp_basename( $details['source_url'] );
		foreach ( $details['sources'] as $mime => &$mime_details ) {
			$mime_details['source_url'] = str_replace( $image_url_basename, $mime_details['file'], $details['source_url'] );
		}
```

felixarntz (changes requested):
@mukeshpanchal27 This is almost good to go, I see two remaining problems here.

modules/images/webp-uploads/fallback.js
mukeshpanchal27: Thanks, Felix for the detailed review. This feedback is addressed.

modules/images/webp-uploads/rest-api.php
mukeshpanchal27: Make sense 👍

felixarntz (approved):
@mukeshpanchal27 Excellent work!

Now that this has been completed, we can add the remaining changes to https://github.com/WordPress/wordpress-develop/pull/3034 and finalize that core PR. 🎉

PR #451, opened 2022-07-25
Improve inline document

### LINKED ISSUE #450: Incorrect return type documented for site-health/audit-enqueued-assets module
In the doc block of site-health/audit-enqueued-assets module/helper.php file, we can read that it always returns an int|False. This is incorrect. As the return value is int|bool

dilipbheda:
Fixes https://github.com/WordPress/performance/issues/450

### DIFF
--- a/modules/site-health/audit-enqueued-assets/helper.php
+++ b/modules/site-health/audit-enqueued-assets/helper.php
@@ -11,7 +11,7 @@
  *
  * @since 1.0.0
  *
- * @return int|false Number of total scripts or false if transient hasn't been set.
+ * @return int|bool Number of total scripts or false if transient hasn't been set.
  */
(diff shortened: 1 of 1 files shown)


tillkruss:
`false` seems accurate. How are these methods returning `true`?

mukeshpanchal27:
The function never returns a `true` value so I don't think we need to change it. A similar function with a `False` return is available in the WordPress core.

Some examples.
https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/includes/class-custom-image-header.php#L1573-L1599
https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-customize-nav-menus.php#L1552-L1585

PR #457, opened 2022-07-28
Fix REST API support for plain permalinks.

### LINKED ISSUE #456: WebP Module fallback.js does not support REST API plain permalinks.
## Bug Description

On sites with plain permalinks (ie not using rewrite rules) the rest requests in fallback.js result in 404 File Not Found errors.

The cause of the error is that the rest_route always appends the fields portion of the querystring with a question mark, for plain permalinks an ampersand should be used.

## Steps to reproduce

1. Configure WordPress to use plain permalinks
2. Upload some a jpeg that will be converted to webp (ie, webp is the smaller file version)
3. Include the image in a post and publish
4. Visit the published post in a browser that does not support webp images
5. In network tools observe that the REST request to get the fallback images fails. The URL will look something like `/index.php?rest_route=/wp/v2/media/?_jsonp=wpPerfLab.webpUploadsFallbackWebpImages&per_page=100&include=38`
6. Observe the querystring has two question marks.

## Screenshots

Screen Shot 2022-07-28 at 1 16 51 pm

## Additional Context

 - PHP Version: 7.4
 - OS: Ubuntu?? (What ever exact version [Chassis/Chassis](https://github.com/chassis/chassis) uses)
 - Browser: macOS/Firefox Dev Edition 104b with `image.webp.enabled` set to false in `about:config`
 - Plugin Version: trunk at ab08abd377780c2f8bf72d8dc737ee7d32f34cda
 - Device: Laptop

peterwilsoncc:
## Summary

Ensure the WebP fallback JavaScript works with sites using plain permalinks.

Fixes #456.

## Relevant technical choices

I decided to copy the approach used in the `@wordpress/apiFetch` [`rootUrl` middleware](https://github.com/WordPress/gutenberg/blob/f84d26e41a6b253b853d591eb040219a11c042cb/packages/api-fetch/src/middlewares/root-url.js) that handles this exact issue. It seems to work. 

Ensures that the querystring does not have a double `?` if the site is configured using plain permalinks. 

## Checklist

### DIFF
--- a/modules/images/webp-uploads/fallback.js
+++ b/modules/images/webp-uploads/fallback.js
@@ -54,7 +54,11 @@ window.wpPerfLab = window.wpPerfLab || {};
 			}
 
 			var jsonp = document.createElement( 'script' );
-			jsonp.src = restApi + 'wp/v2/media/?_fields=id,media_details&_jsonp=wpPerfLab.webpUploadsFallbackWebpImages&per_page=100&include=' + pageIds.join( ',' );
+			var restPath = 'wp/v2/media/?_fields=id,media_details&_jsonp=wpPerfLab.webpUploadsFallbackWebpImages&per_page=100&include=' + pageIds.join( ',' );
+			if ( -1 !== restApi.indexOf( '?' ) ) {
+				restPath = restPath.replace( '?', '&' );
+			}
+			jsonp.src = restApi + restPath;
 			document.body.appendChild( jsonp );
 		}
 	};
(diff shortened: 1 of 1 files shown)


jjgrainger (approved):
Thanks @peterwilsoncc looks good

adamsilverstein (approved):
Nice!

mukeshpanchal27 (approved):
### 🚀

felixarntz (approved):
@peterwilsoncc Great catch, thanks!

PR #458, opened 2022-07-28
Remove plugin option network-wide for Multisite during uninstall

### LINKED ISSUE #382: Remove plugin option for Multisite
## Bug Description

Its follow-up enhancement for [#353](https://github.com/WordPress/performance/pull/345) and feedback https://github.com/WordPress/performance/pull/345#issuecomment-1155608644

If we set up multisite in WP then it will save the option in their sub-site DB. Like if you have two sub-site then the plugin option will save the option in the below tables.

wp_options - Main site option table
wp_2_options - 1st Sub-site option table
wp_3_options - 2nd Sub-site option table

The current uninstaller route only deletes options from the main site `perflab_modules_settings` but it will not remove the option for its subsite.

## Steps to reproduce

1. Setup [Multisite](https://wordpress.org/support/article/create-a-network/)
2. Create 2-3 subsite
3. Download and activate the Performance Lab plugin for all sites.
4. Delete plugin
5. Check the DB `perflab_modules_settings` option for sub-site.

## Screenshots
image

merkys7:
## Summary

Fixes https://github.com/WordPress/performance/issues/382

## Relevant technical choices

Remove plugin option for Multisite during uninstall execution

## Checklist

### DIFF
--- a/uninstall.php
+++ b/uninstall.php
@@ -13,7 +13,6 @@ if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
 
 // For a multisite, delete the option for all sites (however limited to 100 sites to avoid memory limit or timeout problems in large scale networks).
 if ( is_multisite() ) {
-
 	$site_ids = get_sites(
 		array(
 			'fields'                 => 'ids',
@@ -26,8 +25,8 @@ if ( is_multisite() ) {
 	foreach ( $site_ids as $site_id ) {
 		switch_to_blog( $site_id );
 		perflab_delete_plugin_option();
+		restore_current_blog();
 	}
-	restore_current_blog();
 }
 
 perflab_delete_plugin_option();
--- a/uninstall.php
+++ b/uninstall.php
@@ -11,4 +11,32 @@ if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
 	exit;
 }
 
-delete_option( 'perflab_modules_settings' );
+// For a multisite, delete the option for all sites (however limited to 100 sites to avoid memory limit or timeout problems in large scale networks).
+if ( is_multisite() ) {
+
+	$site_ids = get_sites(
+		array(
+			'fields'                 => 'ids',
+			'number'                 => 100,
+			'update_site_cache'      => false,
+			'update_site_meta_cache' => false,
+		)
+	);
+
+	foreach ( $site_ids as $site_id ) {
+		switch_to_blog( $site_id );
+		perflab_delete_plugin_option();
+	}
+	restore_current_blog();
+}
+
+perflab_delete_plugin_option();
+
+/**
+ * Delete the current site's option.
+ *
+ * @since n.e.x.t
+ */
+function perflab_delete_plugin_option() {
+	delete_option( 'perflab_modules_settings' );
+}


uninstall.php
```diff
@@ -11,4 +11,34 @@
...
+ */
+function delete_data() {
+
+	delete_option( 'perflab_modules_settings' );
+}
+
+/**
+ * Delete data for multisite blogs.
+ */
+function delete_data_multisite() {
```
mukeshpanchal27: I don't think we need a separate function for this. You can add this function code in the `if ( is_multisite() ) {` condition because it calls only for multisite.

uninstall.php
```diff
@@ -11,4 +11,34 @@
...
+	global $wpdb;
+
+	$blog_ids         = $wpdb->get_col( "SELECT blog_id FROM {$wpdb->blogs}" );
```
mukeshpanchal27: Why do you prefer the SQL query-based approach? In https://github.com/WordPress/performance/pull/345#issuecomment-1155608644 It used `get_sites()` to get the site ID. Many plugins use a similar approach for removing the options.

mukeshpanchal27: If we still go with SQL query approach then use `$wpdb->prepare` inside `$wpdb->get_col`. Example. https://developer.wordpress.org/reference/classes/wpdb/get_col/#highlighter_403812

uninstall.php
```diff
@@ -11,4 +11,34 @@
...
-delete_option( 'perflab_modules_settings' );
+if ( is_multisite() ) {
+	delete_data_multisite();
+} else {
+	delete_data();
+}
+
+/**
+ * Delete data for the current blog.
+ */
```
mukeshpanchal27: Use proper documentation for the function.
```
/**
 * Delete plugin option data for the current blog.
 *
 * @since n.e.x.t
 */
```

mukeshpanchal27 (changes requested):
@merkys7 Left some feedback.

uninstall.php
merkys7: Thank you, added

uninstall.php
merkys7: Thank you, corrected

uninstall.php
merkys7: Thank you, changed it to `get_sites()`

merkys7: Thank you, changed it to `get_sites()`

mukeshpanchal27: Look good now.

uninstall.php
```diff
@@ -11,4 +11,32 @@
 	exit;
 }
 
-delete_option( 'perflab_modules_settings' );
+if ( is_multisite() ) {
+	$site_ids = get_sites(
+		array(
+			'fields'                 => 'ids',
+			'number'                 => '',
```
mukeshpanchal27: As Felix comment https://github.com/WordPress/performance/pull/345#issuecomment-1155608644

> What we could do is set a sensible limit, e.g. set number to 50 or something. Then it would least support the majority of multisites that don't have a ton of sites.

uninstall.php
```diff
@@ -11,4 +11,32 @@
...
+	delete_data();
+}
+
+/**
+ * Delete plugin option data for the current blog.
+ *
+ * @since n.e.x.t
+ */
+function delete_data() {
+
```
mukeshpanchal27: Remove space.

uninstall.php
```diff
@@ -11,4 +11,32 @@
...
+} else {
+	delete_data();
+}
+
+/**
+ * Delete plugin option data for the current blog.
+ *
+ * @since n.e.x.t
+ */
+function delete_data() {
```
mukeshpanchal27: As per the plugin function name we should use the `perflab_` prefix in name of the function. Something like `perflab_delete_option` or `perflab_delete_plugin_option`.

cc. @felixarntz

mukeshpanchal27 (commented):
@merkys7 Thanks for the updates. The approach here now looks good, Left some more feedback.

uninstall.php
adamsilverstein: I feel like this should support a higher number of sites, maybe 100? Also, should we consider checking the number of sites in the network before the query and if there are over X just skip trying to delete entirely?

uninstall.php
adamsilverstein: do we really need a helper function for one line of code?

adamsilverstein (commented):
left some questions/feedback

uninstall.php
mukeshpanchal27: In my initial approach, I don't use a helper function for a single line: https://github.com/WordPress/performance/pull/345#issuecomment-1155608644 But I think it was used two times, so Merkys introduced a helper function. If it doesn't make sense, then can we remove it?

merkys7: Thank you for the feedback, removed it

uninstall.php
merkys7: Removed the function

uninstall.php
merkys7: I agree with Adam, we should check and skip

merkys7:
I really appreciate your feedback @mukeshpanchal27 @adamsilverstein :bowtie: thank you for taking the time.
I have adjusted the PR based on the comments

uninstall.php
```diff
@@ -11,4 +11,22 @@
 	exit;
 }
 
-delete_option( 'perflab_modules_settings' );
+if ( is_multisite() && 100 >= get_sites( array( 'count' => true ) ) ) {
+
+	$site_ids = get_sites(
+		array(
+			'fields'                 => 'ids',
+			'number'                 => '100',
```
mukeshpanchal27: The number parameter defaults to 100. https://developer.wordpress.org/reference/classes/wp_site_query/__construct/ We should remove it.

uninstall.php
```diff
@@ -11,4 +11,22 @@
 	exit;
 }
 
-delete_option( 'perflab_modules_settings' );
+if ( is_multisite() && 100 >= get_sites( array( 'count' => true ) ) ) {
```
mukeshpanchal27: I feel like we should check the count condition inside the loop. Otherwise, it will run else part for multisite with more than 100 sites.

mukeshpanchal27 (commented):
@merkys7 Thanks for the updates. It looks good. I left some final feedback.

uninstall.php
felixarntz: I would argue we don't need that check for the site count here at all. It's another query to run, which we can avoid. I think we should still handle multisites in the right way, and we can just limit the `get_sites()` query like you do below. That means:
* For networks with 100 or fewer sites, it will wipe all sites.
* For networks with more than 100 sites, it will at least wipe 100 sites.

The latter is not great, but better than what we currently have, while accounting for scalability issues. Also, deleting the option for 100 sites when there are more is still better than falling back to only deleting the option for the current site.

uninstall.php
felixarntz: I think we should rather keep this here, since we really want it to be 100 sites. WordPress core's default could possibly change, so I think explicitly passing the 100 here makes sense.

However, it should probably be an integer per the documented values.
```suggestion
			'number'                 => 100,
```

uninstall.php
```diff
@@ -11,4 +11,22 @@
...
+	);
+
+	foreach ( $site_ids as $site_id ) {
+		switch_to_blog( $site_id );
+		delete_option( 'perflab_modules_settings' );
+	}
+	restore_current_blog();
+} else {
+	delete_option( 'perflab_modules_settings' );
+}
```
felixarntz: I think it would be best to not have this in an `else`, but instead always run it. IMO it makes sense to at least always wipe the current site's (main site) option on uninstall. If a multisite has more than 100 sites, we should still ensure the main site's option is deleted.

So I would suggest the following:
```suggestion
}

// Delete the current site's option.
delete_option( 'perflab_modules_settings' );
```

uninstall.php
felixarntz: Also, let's add a comment here explaining this block.
```suggestion
// For a multisite, delete the option for all sites (however limited to 100 sites to avoid memory limit or timeout problems in large scale networks).
if ( is_multisite() ) {
```

felixarntz (commented):
@merkys7 Thank you for the PR! Left some comments with additional suggestions.

uninstall.php
felixarntz: See my feedback in https://github.com/WordPress/performance/pull/458#discussion_r940206315, IMO it's not worth checking for the number of sites before the main site query. What is the harm of deleting the option for 100 sites when there are more than 100 sites? IMO it's a better outcome to the option removed on e.g. 100 out of 130 sites rather than only on 1 out of 130 sites.

uninstall.php
felixarntz: Not a blocker, but I also prefer to have a function for this. Yes, it's one line, but IMO it makes sense even to have a function for this to clarify that this needs to be exactly the same logic for a single site as well as for every site in a multisite. Plus it avoids a (admittedly tiny) chance of e.g. a typo in the option name in one of the calls, which in a function call would immediately be flagged as fatal error, but in a string isn't.

uninstall.php
mukeshpanchal27: Make sense. Thanks for the clarification, Felix.

merkys7:
Thank you for the feedback guys :bowtie:  I have improved the PR based on the comments

uninstall.php
```diff
@@ -11,4 +11,33 @@
...
+
+perflab_delete_plugin_option();
+
+/**
+ * Delete the current site's option.
+ *
+ * @since n.e.x.t
+ */
+function perflab_delete_plugin_option() {
+
```
mukeshpanchal27: Not a blocker. Can we remove this space?

mukeshpanchal27 (approved):

uninstall.php
merkys7: Removed, thank you

uninstall.php
```diff
@@ -11,4 +11,32 @@
...
+			'update_site_cache'      => false,
+			'update_site_meta_cache' => false,
+		)
+	);
+
+	foreach ( $site_ids as $site_id ) {
+		switch_to_blog( $site_id );
+		perflab_delete_plugin_option();
+	}
+	restore_current_blog();
```
felixarntz: One remaining issue here which I had missed yesterday: `restore_current_blog()` must be called in every run of the `foreach` loop. This may sound confusing, and it certainly is. `restore_current_blog()` is a terrible name for the function, because it actually does not restore the current blog, but rather the previous blog, whether that was the current blog or another switched one.

You can see this also in a couple places in WordPress core: When looping through sites and switching to them, you always need to switch back ("restore") at the end of each loop cycle.
```suggestion
	foreach ( $site_ids as $site_id ) {
		switch_to_blog( $site_id );
		perflab_delete_plugin_option();
		restore_current_blog();
	}
```

uninstall.php
```diff
@@ -11,4 +11,32 @@
 	exit;
 }
 
-delete_option( 'perflab_modules_settings' );
+// For a multisite, delete the option for all sites (however limited to 100 sites to avoid memory limit or timeout problems in large scale networks).
+if ( is_multisite() ) {
+
```
felixarntz: Nit-pick, but +1 on removing this blank line.

felixarntz (commented):
@merkys7 Thank you for the updates, looks great now! I found one more issue that needs to be addressed.

Small aside: Please avoid force-pushing. Preferably you would simply add commits with your changes to this PR. Otherwise it messes up the git history.

felixarntz (approved):
@merkys7 Made the final mini updates here, so this is good to go into 1.4.0 now. Thanks again!

merkys7:
@felixarntz Thank you for the help and feedback!

PR #460, opened 2022-07-31
Fix PHPUnit versions in composer.json

### LINKED ISSUE #351: PHPUnit version in composer is too new
## Bug Description

The PHPUnit version is too new in composer:
` "phpunit/phpunit": "^8.5",`

When running `composer test` it returns:
```Looks like you're using PHPUnit 8.5.21. WordPress requires at least PHPUnit 5.7.21 and is currently only compatible with PHPUnit up to 7.x.```

## Steps to reproduce

1. Run `composer install` && `composer test`

## Screenshots
image

## Additional Context
This is an issue only for advance usage. There's a default way to test `npm run test-php`.

### PR DESCRIPTION
**Pull request by merkys7**
## Summary

Fixes https://github.com/WordPress/performance/issues/351
```
Looks like you're using PHPUnit 8.5.21. WordPress requires at least PHPUnit 5.7.21 and is currently only compatible with PHPUnit up to 7.x.
```
## Relevant technical choices

Mentioned PHPUnit error appears when using 8.1 PHP version and PHPUnit 8.5.21 version.
PHPUnit versions are chosen based on the following table: https://make.wordpress.org/core/handbook/references/phpunit-compatibility-and-wordpress-versions/
Different PHP versions require different PHPUnit versions.
The following approach is taken from `WordPress/WordPres` - `sodium_compat` 
https://github.com/WordPress/WordPress/blob/f2138f29a83594c60122e6c5416c7a8a4717c519/wp-includes/sodium_compat/composer.json#L57

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,14 +14,15 @@
   "require-dev": {
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7",
     "phpcompatibility/php-compatibility": "^9.3",
-    "phpunit/phpunit": "^8.5",
+    "phpunit/phpunit": "^4|^5|^6|^7|^8|^9",
     "squizlabs/php_codesniffer": "^3.5",
     "wp-coding-standards/wpcs": "^2.2",
     "wp-phpunit/wp-phpunit": "^5.8",
     "yoast/phpunit-polyfills": "^1.0"
   },
   "require": {
-    "composer/installers": "~1.0"
+    "composer/installers": "~1.0",
+    "php": ">=5.6|^7|^8"
   },
   "scripts": {
     "format": "phpcbf --standard=phpcs.xml.dist --report-summary --report-source",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: composer.json [hunk 1]
```diff
@@ -14,14 +14,15 @@
...
+    "phpunit/phpunit": "^3|^4|^5|^6|^7|^8|^9",
     "squizlabs/php_codesniffer": "^3.5",
     "wp-coding-standards/wpcs": "^2.2",
     "wp-phpunit/wp-phpunit": "^5.8",
     "yoast/phpunit-polyfills": "^1.0"
   },
   "require": {
-    "composer/installers": "~1.0"
+    "composer/installers": "~1.0",
+    "php": "^5.2.4|^5.3|^5.4|^5.5|^5.6|^7|^8"
```
mukeshpanchal27: I think we don't need to check below 5.6. So here it should be `>=5.6`

merkys7: Thank you for your feedback, I have adjusted the code now

mukeshpanchal27:
Thanks for the update. I have some issues with my composer setup. I will check it once it works. @bvdr Can you please review the PR and share your thoughts?

adamsilverstein (approved):
Looks good, fixed issue in my local testing

bvdr (approved):
👍 Thank you!

mukeshpanchal27 (approved):

PR #465, opened 2022-08-03
Update CODEOWNERS

mxbclang:
Removed "Health Check" from the Persistent Object Cache title, since we don't use that title elsewhere

## Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -50,7 +50,7 @@
 /tests/modules/site-health/audit-enqueued-assets                            @manuelRod
 /tests/testdata/modules/site-health/audit-enqueued-assets                   @manuelRod
 
-# Module: Persistent Object Cache Health Check
+# Module: Persistent Object Cache
 /modules/object-cache/persistent-object-cache-health-check                  @tillkruss @spacedmonkey
 /tests/modules/object-cache/persistent-object-cache-health-check            @tillkruss @spacedmonkey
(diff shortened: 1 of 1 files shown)


tillkruss (approved):

mukeshpanchal27 (approved):

merkys7 (approved):

.github/CODEOWNERS
```diff
@@ -50,7 +50,7 @@
 /tests/modules/site-health/audit-enqueued-assets                            @manuelRod
 /tests/testdata/modules/site-health/audit-enqueued-assets                   @manuelRod
 
-# Module: Persistent Object Cache Health Check
+# Module: Persistent Object Cache
```
felixarntz, line 53: The names here are all based on the actual module name. If we change this here, we should probably rename the module itself as well (in `modules/object-cache/persistent-object-cache-health-check/load.php`).

felixarntz (commented):
@bethanylang One point of feedback here.

felixarntz:
@bethanylang Actually, the change here is somewhat doing the opposite renaming from what #423 is doing. So we should align on going with one of these 2 naming approaches. cc @OllieJones

felixarntz:
@bethanylang Given that in #423 we have decided to use the "Health Check" suffix as the convention, I believe this PR is now obsolete. Is that correct?

mxbclang:
@felixarntz Yep, that's right! Closing this PR.

Issue #467, opened 2022-08-08
JavaScript bundling with WP-Scripts and Webpack


erikyo:
## Feature Description
We should use wp-script to postprocess js file so that all scripts will be minified and processed by babel... This gives countless benefits among them a smaller file and greater compatibility.
I Noticed also that there is no browserslist. if it is possible i would like to add that as well! 

What do you think? is it an easy addition that would improve the js scripts now and those that will be added in the future

mxbclang:
Thanks @erikyo! Tagging @felixarntz in here for his feedback.

mxbclang:
Oops @erikyo I just saw that you discussed this with him in #459, so it looks like you already have what you need here! Reassigning to you.

erikyo:
thanks to you! and sorry I hadn't linked the original conversation... 😓

swissspidy:
Duplicate of #226

PR #468, opened 2022-08-08
JavaScript bundling with WP-Scripts and Webpack #467

erikyo:
## Summary
This pull request enables the use of wp-scripts in order to compile ES to JS. This intrinsically brings many improvements, among them code minification, transpilation to the most used version of js and an increased compatibility with various devices thanks to browserlists. No stylesheet are available at the moment but in the future it might be useful to know that we can now use also scss

## Relevant technical choices
- 2 script were added in package.json, build-js and start-js and are used to do exactly what is documented [here](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-scripts/#setup)
- The file `webpack.config.js` was added with a funtion used to "enqueue" scripts, actually is used as suggested by @felixarntz to bundle fallback.js but any other script could be added in this way:

```javascript
const ref= addModule( 'myfile.js', 'modules/moduleType/moduleName/' );
```

and then simply add to module.exports an array with all the needed modules comma separated, like below
`module.exports = [first, second, etc]`

- Changed the `modules/images/webp-uploads/load.php` path since must now point to the minified version inside the dist folder, but I haven't published that one, so if you need it will be a good opportunity for a double check 😉 

## Checklist


mxbclang:
PR for #467

package.json
```diff
@@ -6,6 +6,7 @@
     "@octokit/rest": "16.26.0",
     "@wordpress/env": "^4.2.2",
     "@wordpress/scripts": "^19.0",
+    "@wordpress/browserslist-config": "^4.1.2",
```
ntwb: This is included in the `@wordpress/scripts` package, so no need to add it separately

https://github.com/WordPress/gutenberg/blob/83a5df739dc904bc106a36002674aa7c68130aee/packages/scripts/package.json#L39

erikyo: yep! Thank you, good catch! fixed!

erikyo:
closing because outdated!

PR #472, opened 2022-08-12
Fix incorrect usage of badge colors in all Site Health checks

### LINKED ISSUE #471: Fix incorrect usage of Site Health checks `color` field
## Bug Description

See https://github.com/WordPress/wordpress-develop/pull/2890#discussion_r939699786: Several of our current Site Health check modules set the `color` based on the status of the check. However, that is not how the color should be used. Instead, it should be based on what area it applies to (in our case probably always `Performance`, for which WordPress core uses `blue`).

cc @manuelRod @tillkruss

felixarntz:
## Summary

Fixes #471

## Relevant technical choices

* All badges should always be "Performance" and use the color `blue`.
* Setting the badge color based on check result status is incorrect.

## Checklist

### DIFF
--- a/modules/site-health/audit-autoloaded-options/load.php
+++ b/modules/site-health/audit-autoloaded-options/load.php
@@ -43,7 +43,7 @@ function perflab_aao_autoloaded_options_test() {
 		'label'       => esc_html__( 'Autoloaded options are acceptable', 'performance-lab' ),
 		'status'      => 'good',
 		'badge'       => array(
-			'label' => esc_html__( 'Performance', 'performance-lab' ),
+			'label' => __( 'Performance', 'performance-lab' ),
 			'color' => 'blue',
 		),
 		'description' => sprintf(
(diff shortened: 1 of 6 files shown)


modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -69,10 +69,9 @@ function perflab_aao_autoloaded_options_test() {
 		return $result;
 	}
 
-	$result['status']         = 'critical';
-	$result['badge']['color'] = 'red';
-	$result['label']          = esc_html__( 'Autoloaded options could affect performance', 'performance-lab' );
-	$result['description']    = sprintf(
+	$result['status']      = 'critical';
+	$result['label']       = esc_html__( 'Autoloaded options could affect performance', 'performance-lab' );
```
mukeshpanchal27:
```suggestion
	$result['label']       = __( 'Autoloaded options could affect performance', 'performance-lab' );
```

jjgrainger (approved):

mehulkaklotar (approved):
Looks good! 🚀

mukeshpanchal27 (approved):

ankitrox (approved):
LGTM

modules/site-health/audit-autoloaded-options/load.php
```diff
@@ -40,15 +40,15 @@ function perflab_aao_autoloaded_options_test() {
 	$base_description = __( 'Autoloaded options are configuration settings for plugins and themes that are automatically loaded with every page load in WordPress. Having too many autoloaded options can slow down your site.', 'performance-lab' );
 
 	$result = array(
-		'label'       => esc_html__( 'Autoloaded options are acceptable', 'performance-lab' ),
```
adamsilverstein, line 43: I noticed we are applying `esc_html__` inconsistently to translations and that you removed a bunch in this PR. 

For core, we don't escape translations, however for plugins I generally recommend escaping (as 10up [recommends](https://10up.github.io/Engineering-Best-Practices/php/#:~:text=Translated%20text%20also%20often%20needs%20to%20be%20escaped%20on%20output.)). 

Maybe rather than changing here we can open a follow issue to use a uniform approach and change throughout the plugin?

felixarntz: While escaping is recommended for plugins more than core, the rules also say that escaping should happen at the output level. In all the places here though, we're only writing data into an array, so per those guidelines escaping here would not follow the best practice.

All escaping for Site Health check content should happen in WordPress core per that definition, since that's where the data is being output.

PR #496, opened 2022-08-22
Add image regeneration module

### LINKED ISSUE #489: Create Regenerate existing images module
Create the initial module infrastructure for the Regenerate Existing Images module. This is required before any further development of the feature can take place. Additionally, as part of this work, the main feature branch `feature/regenerate-existing-images` should be created for all future work to be branched from and merged into.

## Acceptance criteria

- A new folder `modules/images/regenerate-existing-images` is added to the plugin
- Inside the new folder a `can-load.php` exists and return an anonymous function
    - The anonymous function returns `true` for now
- Inside the new folder a `load.php` file is created and has the correct docblock at the top of the file
- A new `feature/regenerate-existing-images` branch is created

ankitrox:
## Summary

This PR adds the `regenerate-existing-images` module in the plugin. This shall be the first thing to get merged for image regeneration work.

All of the subsequent work related to image regeneration would reside in this module.

Fixes #489 

## Checklist


modules/images/regenerate-existing-images/can-load.php
```diff
@@ -0,0 +1,12 @@
...
+ * Can load function to determine if image regenaration module is already merged in WordPress core.
+ *
+ * @todo Adjust the returned value with appropriate condition.
+ * @since   n.e.x.t
+ * @package performance-lab
+ */
+
+return function() {
+	return true;
+};
```
mukeshpanchal27: It should be like below?
```suggestion
return '__return_true';
```

mukeshpanchal27 (commented):
Left one minor review.

modules/images/regenerate-existing-images/can-load.php
ankitrox: @mukeshpanchal27 True value is temporary and will be replaced by appropriate condition eventually. Also, [this snippet](https://github.com/WordPress/performance/blob/trunk/load.php#L216-L218) will load the module if the returned value is not callable, so I think we will save calling an extra function which eventually returns `true` which we are already returning.

Please let me know your opinion.

felixarntz: Similar to my comment in #489, I'd argue we don't even need this file right now. We can add it later if we ever want to add custom logic to determine whether the module can load or not.

modules/images/regenerate-existing-images/load.php
```diff
@@ -0,0 +1,10 @@
+<?php
+/**
+ * Module Name: Regenarate Existing Images
+ * Description: Regenerate existing images when the image sizes changes during plugin/theme
+ * activation/deactivation or for any other reasons.
```
felixarntz: Please don't break this across multiple lines, as it doesn't work with the module header parsing. Also, suggesting a few tweaks to the description.
```suggestion
 * Description: Introduces background process infrastructure to regenerate existing image sizes and MIME types on demand.
```

felixarntz (commented):
@ankitrox Two minor comments below, and one more here: Can you please also add this module to the `CODEOWNERS` file, similar to the other image modules? For now maybe add @jjgrainger and yourself as owners if that works for you?

modules/images/regenerate-existing-images/load.php
```diff
@@ -0,0 +1,10 @@
...
+ * Experimental: No
+ *
+ * @since   n.e.x.t
```
mukeshpanchal27: Remove space
```suggestion
 * @since n.e.x.t
```

mukeshpanchal27 (commented):
Nit-pick change

felixarntz (approved):
Great, thank you @ankitrox!

Issue #491, opened 2022-08-19
Create Background Processing Infrastructure: Create Queue Admin Screen


jjgrainger:
Create a basic queue admin screen to display a list of all jobs. This will make use of the default taxonomy admin screen but made read only by removing the create/edit functionality. This screen will be used by developers for debugging and will not be made visible by default. A filter will be created to allow developers to enable the screen when required.

## Acceptance Criteria
- Utilise the existing taxonomy screen for the `background_job` taxonomy
- Disable user creating/editing and deleting of jobs by assigning custom capabilities to the `background_job` taxonomy
- Create a `perflab_enable_background_process_queue_screen` filter used to hide/show the admin screen
- The admin screen is disabled by default

PR #497, opened 2022-08-23
Background job taxonomy for queuing jobs

### LINKED ISSUE #478: Create Background Processing Infrastructure: Create background_job taxonomy
Create a new Background Jobs taxonomy to store each job to run in the background process infrastructure. The taxonomy will be registered in the `modules/regenerate-existing-images/load.php` file.

## Acceptance Criteria
- A new `background_job` taxonomy is registered
- The taxonomy has the following attributes
    - Assigned to all public post types
    - Set `public` to false
    - Set `show_in_rest` to false

ankitrox:
## Summary

Create the `background_job` taxonomy. This taxonomy is non-public and hidden in REST. This will by default be hidden in admin screen, but can be made visible using `perflab_enable_background_process_queue_screen` filter. ( #491 )

Fixes #478

## Relevant technical choices

Taxonomy has custom capabilities

```
manage_jobs,
edit_jobs,
delete_jobs,
assign_jobs
```

User must have these capabilities assigned for the taxonomy (job queue) management. By default admin screen for this taxonomy (job queue) will be hidden.

## Checklist


modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+	$caps = get_background_job_capabilities();
+
+	// Taxonomy arguments.
+	$args = array(
+		'hierarchical'       => false, // We do not need child job.
+		'labels'             => $labels,
+		'public'             => false,
+		'show_ui'            => true,
+		'show_in_menu'       => false,
+		'show_in_nav_menus'  => true,
```
felixarntz: This should be `false` I believe? We don't want this taxonomy to be in any nav menu.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+		'show_in_quick_edit' => false,
+		'show_admin_column'  => false,
+		'show_in_rest'       => false,
+		'query_var'          => true,
```
felixarntz: This should also be `false`, we don't ever want to query this in a frontend request.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+		'rewrite'            => array( 'slug' => 'background_job' ),
```
felixarntz: I would say let's remove this since we don't need this taxonomy in the frontend anyway, and even if so the default there should be fine.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+		'singular_name'     => _x( 'Background Job', 'taxonomy singular name', 'performance-lab' ),
+		'search_items'      => __( 'Search Background Jobs', 'performance-lab' ),
+		'all_items'         => __( 'All Background Jobs', 'performance-lab' ),
+		'parent_item'       => __( 'Parent Background Job', 'performance-lab' ),
+		'parent_item_colon' => __( 'Parent Background Job:', 'performance-lab' ),
+		'edit_item'         => __( 'Edit Background Job', 'performance-lab' ),
+		'update_item'       => __( 'Update Background Job', 'performance-lab' ),
+		'add_new_item'      => __( 'Add New Background Job', 'performance-lab' ),
+		'new_item_name'     => __( 'New Background Job Name', 'performance-lab' ),
+		'menu_name'         => __( 'Background Job', 'performance-lab' ),
```
felixarntz: This one should be the plural, as the menu items should always be plural. Alternatively, we could just remove this entry as it should automatically be set to the value of `name` in that case.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+	);
```
felixarntz: There are several labels missing here, see https://github.com/WordPress/wordpress-develop/blob/6.0.1/src/wp-includes/class-wp-taxonomy.php#L612 for the full list of labels.

We definitely don't need all of them, as some of these areas are really unlikely to ever apply to this taxonomy, but some of the more "critical" ones should be added, e.g.
* `view_item`
* `not_found`
* `no_terms`
* `items_list_navigation`
* `items_list`
* `back_to_items`

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+		'show_admin_column'  => false,
+		'show_in_rest'       => false,
+		'query_var'          => true,
+		'rewrite'            => array( 'slug' => 'background_job' ),
+		'capabilities'       => array(
+			'manage_terms' => 'manage_options',
+			'edit_terms'   => $caps['edit'],
+			'delete_terms' => $caps['delete'],
+			'assign_terms' => $caps['assign'],
+		),
```
felixarntz: It's a bit over-complicated to manually assign this here for each field. Why not change the capabilities function to simply return the data with the right keys? That way you can simply change it to something like
```suggestion
		'capabilities'       => perflab_get_background_job_capabilities(),
```

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+
+/**
+ * Register the background job taxonomy.
+ *
+ * Intentionally on lower priority, so that other post types can be
+ * registered successfully and taxonomy can be used for all the post types.
+ *
+ * This taxonomy will be used as a queue system where each term represents a job in the queue.
+ *
+ * @return void
```
felixarntz: `@return` docs should be omitted when it's just `void`, see WordPress core documentation standards.

Instead, we need to add a `@since n.e.x.t` tag here, as for every function/class/method etc.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+ */
+function perflab_background_job_taxonomy() {
```
felixarntz: Let's rename this to a more "descriptive" `perflab_register_background_job_taxonomy()`. Function names should typically begin with an action verb.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+ *
+ * @return array
+ */
+function get_background_job_capabilities() {
+	return array(
+		'manage' => 'manage_jobs', // To manage background_job terms.
+		'edit'   => 'edit_job', // To edit background_job terms.
+		'delete' => 'delete_job', // To delete background_job terms.
+		'assign' => 'assign_jobs', // To assign background_job terms to posts.
+	);
```
felixarntz: See my comment above, this should return the data with the right keys so that it can be simply assigned to the `capabilities` entry above.

Also, the capabilities should use plural always, i.e. `edit_jobs` and `delete_jobs`.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+	register_taxonomy( 'background_job', $object_types, $args );
+}
+add_action( 'init', 'perflab_background_job_taxonomy', 100 );
+
+/**
+ * Retrieve list of capabilities to manage background jobs.
+ *
+ * @return array
+ */
+function get_background_job_capabilities() {
```
felixarntz: This function is missing the required code prefix of `perflab_`.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+
+	// Register background_job taxonomy.
+	register_taxonomy( 'background_job', $object_types, $args );
+}
+add_action( 'init', 'perflab_background_job_taxonomy', 100 );
+
+/**
+ * Retrieve list of capabilities to manage background jobs.
+ *
+ * @return array
```
felixarntz: See above, please add a `@since` tag, and also a description for the return value.
```suggestion
 * @since n.e.x.t
 *
 * @return array Map of core term capabilities and the actual capability names to use.
```

tests/modules/images/regenerate-existing-images/load-tests.php
```diff
@@ -0,0 +1,36 @@
...
+	 * Ensure taxonomy is non public and would not show in REST.
+	 *
+	 * @return void
+	 */
+	public function test_background_job_taxonomy_is_non_public() {
+		$job_tax = get_taxonomy( 'background_job' );
+
+		$this->assertFalse( $job_tax->public );
+		$this->assertFalse( $job_tax->show_in_rest );
+		$this->assertFalse( $job_tax->hierarchical );
```
felixarntz, line 33: Let's also add assertions for the other fields here that should be `false`.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
  * @since n.e.x.t
  * @package performance-lab
  */
+
+/**
+ * Register the background job taxonomy.
```
felixarntz: Nit-pick, but function descriptions should begin in a third-person verb.
```suggestion
 * Registers the background job taxonomy.
```

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+		),
+	);
+
+	// Register background_job taxonomy.
+	register_taxonomy( 'background_job', $object_types, $args );
+}
+add_action( 'init', 'perflab_background_job_taxonomy', 100 );
+
+/**
+ * Retrieve list of capabilities to manage background jobs.
```
felixarntz: Nit-pick, see above.
```suggestion
 * Retrieves list of capabilities to manage background jobs.
```

felixarntz (commented):
@ankitrox This is a good start, I left some comments (mostly minor) for things to refine.

One small note, please avoid using the `feature/` prefix for your development branches as those branches will need branch protection rules. Only special branches (like the overall `feature/regenerate-existing-images` branch) should use that prefix. 👍

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,81 @@
...
+		'menu_name'             => __( 'Background Jobs', 'performance-lab' ),
+		'view_item'             => __( 'View Background Job', 'performance-lab' ),
+		'not_found'             => __( 'Background Job Not Found', 'performance-lab' ),
+		'no_terms'              => __( ' No Background Jobs', 'performance-lab' ),
+		'items_list'            => __( 'Background Jobs List', 'performance-lab' ),
+		'items_list_navigation' => __( 'Background Jobs List Navigation', 'performance-lab' ),
+		'back_to_items'         => __( 'Back to Background Jobs', 'performance-lab' ),
+	);
+
+	$caps = perflab_get_background_job_capabilities();
```
mukeshpanchal27: Do you need this?

tests/modules/images/regenerate-existing-images/load-tests.php
```diff
@@ -0,0 +1,41 @@
+<?php
+/**
+ * Tests for regenerate-existing-images module load.php.
+ *
+ * @package performance-lab
+ * @group   regenerate-existing-images
+ */
+
+
```
mukeshpanchal27: Remove space.

mukeshpanchal27 (commented):
Left some nit-pick reviews.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+ * @return array Map of core term capabilities and the actual capability names to use.
+ */
+function perflab_get_background_job_capabilities() {
+	return array(
+		'manage_terms' => 'manage_jobs', // To manage background_job terms.
+		'edit_terms'   => 'edit_jobs', // To edit background_job terms.
+		'delete_terms' => 'delete_jobs', // To delete background_job terms.
+		'assign_terms' => 'assign_jobs', // To assign background_job terms to posts.
+	);
+}
```
eugene-manuilov, line 79: Why do we need a separate function for capabilities? Are we going to use it somewhere else? If not, let's remove it and define capabilities as a nested array in the taxonomy args array.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+ * @since n.e.x.t
+ */
+function perflab_register_background_job_taxonomy() {
+	// Get all registered public post types.
+	$object_types = get_post_types(
+		array(
+			'public' => true,
+		),
+		'names'
+	);
```
eugene-manuilov: Do we really need to register this taxonomy for all public post types? I think this is not necessary and may cause problems if a 3rd party develop would like to use it to process posts of a hidden post type in their plugin/theme.

I think we should use an empty array for post types when we register the taxonomy:

```php
register_taxonomy( 'background_job', array(), $args );
```

We will still be able to assign a post to a taxonomy term programmatically even if that post type is not assigned with the taxonomy.

cc @jjgrainger @felixarntz @adamsilverstein

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+		'items_list_navigation' => __( 'Background Jobs List Navigation', 'performance-lab' ),
+		'back_to_items'         => __( 'Back to Background Jobs', 'performance-lab' ),
+	);
+
+	// Taxonomy arguments.
+	$args = array(
+		'hierarchical'       => false, // We do not need child job.
+		'labels'             => $labels,
+		'public'             => false,
+		'show_ui'            => true,
```
eugene-manuilov, line 46: I think this `show_ui` property should be `false` because the standard admin UI for this taxonomy won't make any sense. We will need a custom admin page that will display all active background jobs.

cc @jjgrainger

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+		'show_in_menu'       => false,
+		'show_in_nav_menus'  => false,
+		'show_in_quick_edit' => false,
+		'show_admin_column'  => false,
+		'show_in_rest'       => false,
+		'query_var'          => false,
+		'capabilities'       => perflab_get_background_job_capabilities(),
```
eugene-manuilov: * `hierarchical` and `show_admin_column` can be removed as their default value is already `false`.
* `show_ui`, `show_in_menu`, `show_in_nav_menus`, `show_in_quick_edit` inherit the value from the `public` property and can be removed since they don't change the value of the `public` property.
* `rewrite` has to be added with `false` to prevent rewrite rules generation.

```php
$args = array(
    'labels'       => $labels,
    'public'       => false,
    'show_in_rest' => false,
    'query_var'    => false,
    'rewrite'      => false,
    'capabilities' => array( ... ),
);
```

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+ * @since n.e.x.t
+ *
+ * @return array Map of core term capabilities and the actual capability names to use.
+ */
+function perflab_get_background_job_capabilities() {
+	return array(
+		'manage_terms' => 'manage_jobs', // To manage background_job terms.
+		'edit_terms'   => 'edit_jobs', // To edit background_job terms.
+		'delete_terms' => 'delete_jobs', // To delete background_job terms.
+		'assign_terms' => 'assign_jobs', // To assign background_job terms to posts.
```
eugene-manuilov, line 77: This is a system taxonomy which means that it will be processed in a background where there is no current user. Thus I am not sure that this capabilities will be really needed and will make any sense at all... 🤔 

cc @felixarntz @adamsilverstein @jjgrainger

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,79 @@
...
+		'show_admin_column'  => false,
+		'show_in_rest'       => false,
+		'query_var'          => false,
+		'capabilities'       => perflab_get_background_job_capabilities(),
+	);
+
+	// Register background_job taxonomy.
+	register_taxonomy( 'background_job', $object_types, $args );
+}
+add_action( 'init', 'perflab_register_background_job_taxonomy', 100 );
```
eugene-manuilov: The `100` priority isn't needed here if we remove post types.

eugene-manuilov (changes requested):
Thanks, @ankitrox. Added some feedback to your code that will probably will require additional discussion within our joint team.

Also, I think we need to remove the mention of the 491 ticket in the PR description because this PR doesn't implement the admin UI screen.

modules/images/regenerate-existing-images/load.php
jjgrainger: @eugene-manuilov The intention is to use custom capabilities to hide the admin screen initially but allow developers to access this by adding these capabilities to their own user. The screen is mean for debugging purposes for now and not mean to be available for all admins.

modules/images/regenerate-existing-images/load.php
jjgrainger: Thanks @eugene-manuilov 

As long as we can still assign terms to posts programatically without the need to register the taxonomy to the post type I'm happy to keep this as an empty array.

modules/images/regenerate-existing-images/load.php
ankitrox: @eugene-manuilov Post type/Object type is required parameter for `register_taxonomy`

modules/images/regenerate-existing-images/load.php
eugene-manuilov: The standard taxonomy screen does not make any sense for background jobs even if we hide the second for creating new terms there and leave just the table. We will need to customise it so deeply to provide useful details, so I think it would be better to create a new tools page for it from scratch.

modules/images/regenerate-existing-images/load.php
eugene-manuilov: Yes, it is required parameter, but as I mentioned in another comment we don't need to pass post types to it, we can use just an empty array which means we don't need to wait for other post types to be registered hence no need to use `100` priority here.

modules/images/regenerate-existing-images/load.php
ankitrox: I think we should keep it for convenience of other developers who want to customise admin screen. They can use this function and assign these capabilities to user.

modules/images/regenerate-existing-images/load.php
ankitrox: As of now, we have to keep it as `true`. If `show_ui` will be `false`, visiting `https://mysite.com/wp-admin/edit-tags.php?taxonomy=background_job` will give the error 
```
Sorry, you are not allowed to edit terms in this taxonomy.
```

modules/images/regenerate-existing-images/load.php
ankitrox: It has been addressed with empty array.

ankitrox:
@eugene-manuilov We should keep #491 also as we have taken the decision to manipulate the visibility based on user capabilities.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,74 @@
...
+		'items_list_navigation' => __( 'Background Jobs List Navigation', 'performance-lab' ),
+		'back_to_items'         => __( 'Back to Background Jobs', 'performance-lab' ),
+	);
+
+	// Taxonomy arguments.
+	$args = array(
+		'labels'       => $labels,
+		'public'       => false,
+		'show_ui'      => true,
+		'show_in_rest' => false,
```
felixarntz: Nit-pick, technically no need to pass `show_in_rest` since it's `false` anyway by default. But also perfectly fine to keep it.

modules/images/regenerate-existing-images/load.php
felixarntz: I agree that the capabilities still make sense in order to manage the jobs - regardless of whether we implement a new screen or modify the default one.

FWIW, since we're building this for WP core and since it's an advanced utility screen anyway that would be enabled by default, I think we should use the default screen and customize it as needed. This way we reuse the existing infrastructure and keep the work minimal, which given that this screen won't be ever used by most users is reasonable IMO.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,74 @@
...
+		'update_item'           => __( 'Update Background Job', 'performance-lab' ),
+		'add_new_item'          => __( 'Add New Background Job', 'performance-lab' ),
+		'new_item_name'         => __( 'New Background Job Name', 'performance-lab' ),
+		'menu_name'             => __( 'Background Jobs', 'performance-lab' ),
+		'view_item'             => __( 'View Background Job', 'performance-lab' ),
+		'not_found'             => __( 'Background Job Not Found', 'performance-lab' ),
+		'no_terms'              => __( ' No Background Jobs', 'performance-lab' ),
+		'items_list'            => __( 'Background Jobs List', 'performance-lab' ),
+		'items_list_navigation' => __( 'Background Jobs List Navigation', 'performance-lab' ),
+		'back_to_items'         => __( 'Back to Background Jobs', 'performance-lab' ),
```
felixarntz: Please align the casing here with core's taxonomies, not all words should start with an uppercase letter. In addition, some of the wording chosen here is different from core and partly inaccurate. We should align with one of the core taxonomies consisting of two words for the labels. For reference, check the strings used for e.g. `wp_block`.

modules/images/regenerate-existing-images/load.php
```diff
@@ -7,3 +7,74 @@
...
+		'parent_item'           => __( 'Parent Background Job', 'performance-lab' ),
+		'parent_item_colon'     => __( 'Parent Background Job:', 'performance-lab' ),
+		'edit_item'             => __( 'Edit Background Job', 'performance-lab' ),
+		'update_item'           => __( 'Update Background Job', 'performance-lab' ),
+		'add_new_item'          => __( 'Add New Background Job', 'performance-lab' ),
+		'new_item_name'         => __( 'New Background Job Name', 'performance-lab' ),
+		'menu_name'             => __( 'Background Jobs', 'performance-lab' ),
+		'view_item'             => __( 'View Background Job', 'performance-lab' ),
+		'not_found'             => __( 'Background Job Not Found', 'performance-lab' ),
+		'no_terms'              => __( ' No Background Jobs', 'performance-lab' ),
```
felixarntz: There's a space here that needs to be removed.

tests/modules/images/regenerate-existing-images/load-tests.php
```diff
@@ -0,0 +1,40 @@
...
+	 *
+	 * @return void
+	 */
+	public function test_background_job_taxonomy_is_non_public() {
+		$job_tax = get_taxonomy( 'background_job' );
+
+		$this->assertFalse( $job_tax->public );
+		$this->assertFalse( $job_tax->show_in_rest );
+		$this->assertFalse( $job_tax->hierarchical );
+		$this->assertFalse( $job_tax->show_in_menu );
```
felixarntz: I believe this is incorrect, `show_in_menu` is expected to be `true` since `show_ui` is `true`. On that note, please add the assertion for `show_ui` as well.

I just realized that tests aren't running, probably because of the new feature branch. I'll configure this shortly.

modules/images/regenerate-existing-images/load.php
felixarntz: See my other response, I think we should keep this for now, potentially make it filterable later. IMO we should try first to work with the existing screen rather than putting in all the work to create an entirely custom one for this UI.

Note that this is specifically about the jobs infrastructure screens. The more user-facing regenerate existing images screen _will_ be completely custom as it needs to be more intuitive and user-friendly.

felixarntz (commented):
@ankitrox Only a few nit-picks left.

As @eugene-manuilov  suggested, please remove #491 from the PR mentions. While this PR does part of introducing the default screen, it doesn't fully make it available, so let's work on that in a separate PR.

ankitrox:
@felixarntz Addressed the feedback. Also, removed #491 from this PR.

modules/images/regenerate-existing-images/load.php
eugene-manuilov: @felixarntz, if we need a screen where we can see all background jobs, then we can add a custom tools page where we can display all background jobs using the existing WP functionality. See my PoC in #506.

felixarntz: That may work, but let's evaluate that in a separate PR indeed. For now we can keep this as is. I left a few early comments on your experimental PR.

felixarntz (approved):
@ankitrox Great work!

@eugene-manuilov Could you please re-review? I left replies to your concerns. Let's address the admin screen details in a follow up.

eugene-manuilov (approved):
LGTM. Thanks, @ankitrox.

PR #498, opened 2022-08-26
Background process runner

### LINKED ISSUE #482: Create Background Processing Infrastructure: Perflab_Background_Process class
### Create Perflab_Background_Process::run method
Create a new a `Perflab_Background_Process` class to handle the background process runs.

This class will contain other helper methods for running jobs the background process built out in other issues.

- [x] #493 
- [x] #494 

## Acceptance Criteria

- Create a new `run( int $job_id )` method in the `Perflab_Background_Process` class.
- The method will run through the following steps sequentially.

1. Set the start time
A `$start_time` variable is set using the current timestamp as the value.

2. Lock the process
- The job process is locked by setting a term meta against the job term. The meta key will be `job_lock` and the value the current timestamp.
- The job status is also updated by setting the `job_status` meta to `running`

3.  Retrieve the job information
Using the `$job_id` passed to the `run` method get the following information.

- `$job` - Create a new Job object using the `Perflab_Background_Job` class
- `$job_name` this is the the job terms slug
- `$job_data` this is from the `job_data` term meta

4. Call the job action
Within a do/while loop call `do_action( 'perflab_job_' . $job_name, $job_id, $job_data );`

The while loop conditions will call the following like so:
`while( ! $this->memory_exceeded() && ! $this->time_exceeded() && ! $job->is_completed( $job_id ) )`

5. Unlock process
Outside of the while loop, to end the process, unlock the job by removing the `job_lock` term meta.

6. Start the next batch process
Start the next batch process by calling the function `perflab_background_process_next_batch()` passing the `$job_id` as the only parameter.

### LINKED ISSUE #494: Create Background Processing Infrastructure: Create Perflab_Background_Process::time_exceeded method
Create a new `Perflab_Background_Process::time_exceeded` to calculate the total time used within the background process. This function checks the total time taken within the background process so that it can break the while loop 10 seconds before the server timeout limit is reached. The boolean value returned by the function will be filterable to allow other developers to modify the timeout limit.

## Acceptance Criteria
- Create a new `time_exceeded` method in the `Perflab_Background_Process` class
- The method accepts no parameters
- The method returns a boolean
- The total timeout limit is calculated by taking the value of `ini_get( 'max_execution_time' )` or a fallback value of 30 seconds, minus 10 seconds. This becomes the `$limit` value.
- The timeout limit is tested by checking the `time() >= ( $start_time + $limit )`
- The boolean value returned is filtered using apply_filters( `perflab_background_process_time_exceeded`, $return );`

ankitrox:
Background process runner class.

Mainly responsible for handling and validating the heavy lifting tasks in background without hampering the user experience.

## Summary

Fixes #482  #494 

## Checklist

### DIFF
--- a/modules/images/regenerate-existing-images/background-process/class-perflab-background-job.php
+++ b/modules/images/regenerate-existing-images/background-process/class-perflab-background-job.php
@@ -1,0 +1,131 @@
+<?php
+/**
+ * Background job class.
+ *
+ * @since n.e.x.t
+ * @package performance-lab
+ */
+
+/**
+ * Class Perflab_Background_Job.
+ *
+ * Runs the heavy lifting tasks in background in separate process.
+ */
+class Perflab_Background_Job {
+
+	/**
+	 * Job ID.
+	 *
+	 * @var int
+	 */
+	public $job_id;
+
+	/**
+	 * Job name.
+	 *
+	 * @var string
+	 */
+	public $name;
+
+	/**
+	 * Job data.
+	 *
+	 * @var array
+	 */
+	public $data;
+
+	/**
+	 * Perflab_Background_Job constructor.
+	 *
+	 * @param int    $job_id job ID.
+	 * @param string $name   Name of the job to run/create.
+	 * @param array  $data   Data for the corresponding job.
+	 */
+	public function __construct( $job_id = 0, $name = '', array $data = array() ) {
+		$this->job_id = absint( $job_id );
+		$this->name   = (string) $name;
+	}
+
+	/**
+	 * Retrieves the job data.
+	 *
+	 * @return array|null
+	 */
+	public function data() {
+		/**
+		 * If we have cached data, return it.
+		 */
+		if ( ! is_null( $this->data ) ) {
+			return $this->data;
+		}
+
+		$this->data = get_term_meta( $this->job_id, 'job_data', true );
+
+		return $this->data;
+	}
+
+	/**
+	 * Create a new job in the queue.
+	 *
+	 * Job refers to the term and queue refers to the `background_job` taxonomy.
+	 *
+	 * @return int|WP_Error
+	 */
+	public function create() {
+        // Create job only when job ID for current instance is zero.
+        if ( $this->job_id > 0 ) {
+            return new WP_Error( __( 'Cannot create the job as it exists already.', 'performance-lab' ) );
+        }
+
+        $term_name = 'job_' . microtime();
+
+        // Insert the new job in queue.
+        $term_data = wp_insert_term(
+            $term_name,
+            'background_job',
+            array(
+                'slug' => $term_name,
+            )
+        );
+
+        if ( ! is_wp_error( $term_data ) ) {
+            $this->job_id = $term_data['term_id'];
+
+            update_term_meta( $this->job_id, $this->data, 'job_data' );
+
+            /**
+             * Fires when the job has been created successfully.
+             * 
+             * @since n.e.x.t
+             * 
+             * @param int    $job_id Job ID.
+             * @param string $name Job name.
+             * @param array  $data Job data.
+             */
+            do_action( 'perflab_job_created', $this->job_id, $this->name, $this->data );
+        }
+
+        return $term_data;
+	}
+
+	/**
+	 * Run this job.
+	 *
+	 * This needs to be implemented by concrete implementations.
+	 *
+	 * @return void
+	 */
+	public function run() {
+		/**
+		 * Hook to this action to run the job.
+		 *
+		 * Concrete implementations would add the logic to run this job.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param int   $job_id Job ID.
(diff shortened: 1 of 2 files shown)


felixarntz (commented):
@ankitrox I think this PR is a bit too much all at once. Let's work in smaller units, to be able to have focused reviews and not block multiple items because of having them in one PR.

Your PR currently tackles #482, #494, _and_ also #503. Can you please reduce this (or open a new PR with the code instead) which only focuses on #503? We should always try to have one PR for one issue. Of course, don't completely delete the other code, but let's split that into multiple PRs. I believe the first one should be the PR for #503 since that is the most foundational part of the three.

modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
```diff
@@ -0,0 +1,178 @@
...
+	/**
+	 * Perflab_Background_Process constructor.
+	 *
+	 * Adds the necessary hooks to trigger process handling.
+	 *
+	 * @return void
+	 */
+	public function __construct() {
+		add_action( 'wp_ajax_' . Perflab_Background_Process::BG_PROCESS_ACTION, array( $this, 'handle_request' ) );
+		add_action( 'wp_ajax_nopriv' . Perflab_Background_Process::BG_PROCESS_ACTION, array( $this, 'handle_request' ) );
```
mukeshpanchal27, line 39:
```suggestion
		add_action( 'wp_ajax_nopriv_' . Perflab_Background_Process::BG_PROCESS_ACTION, array( $this, 'handle_request' ) );
```

modules/images/regenerate-existing-images/background-process/class-perflab-background-job.php
```diff
@@ -0,0 +1,236 @@
...
+ * Background job class.
+ *
+ * @since n.e.x.t
+ * @package performance-lab
+ */
+
+/**
+ * Class Perflab_Background_Job.
+ *
+ * Runs the heavy lifting tasks in background in separate process.
```
mukeshpanchal27, line 12: Add `@since n.e.x.t`

modules/images/regenerate-existing-images/background-process/class-perflab-background-job.php
```diff
@@ -0,0 +1,236 @@
...
+ */
+final class Perflab_Background_Job {
+	/**
+	 * Meta key for storing job name/action.
+	 *
+	 * @const string
```
mukeshpanchal27, line 18: Add `@since n.e.x.t` for all constant and variables https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#2-1-class-members

modules/images/regenerate-existing-images/background-process/class-perflab-background-job.php
```diff
@@ -0,0 +1,236 @@
...
+	 * Job data.
+	 *
+	 * @var array
+	 */
+	public $data;
+
+	/**
+	 * Perflab_Background_Job constructor.
+	 *
+	 * @param int $job_id job ID.
```
mukeshpanchal27, line 53: Add `@since n.e.x.t` for all the functions.

modules/images/regenerate-existing-images/background-process/class-perflab-background-job.php
```diff
@@ -0,0 +1,236 @@
...
+		 *
+		 * By default this will be empty array. Consumer code will need to use this filter in
+		 * order to return the list of items which needs to be processed in current batch.
+		 *
+		 * @param array  $items  Items for the current batch.
+		 * @param int    $job_id Background job ID.
+		 * @param string $name   Job identifier or name.
+		 * @param array  $data   Job data.
+		 *
+		 * @since n.e.x.t
```
mukeshpanchal27, line 208: As per document standard https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters you should add `@since` before the parameter document.

modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
mukeshpanchal27, line 39: For the same file class file, we should use `self::`. It should be updated in other places as well. 
```suggestion
		add_action( 'wp_ajax_' . self::BG_PROCESS_ACTION, array( $this, 'handle_request' ) );
		add_action( 'wp_ajax_nopriv' . self::BG_PROCESS_ACTION, array( $this, 'handle_request' ) );
```

mukeshpanchal27 (commented):
Left some quick feedback.

ankitrox:
Thanks for the review @mukeshpanchal27 .

As per discussion with @felixarntz , I have created a separate (and smaller) PR background job class: https://github.com/WordPress/performance/pull/507

So, this PR will be closed and further work will be continued in those smaller chunks. Please consider to review those new PRs.

jjgrainger:
Closing to avoid confusion with another open PR

PR #500, opened 2022-08-26
Correct docblock for `wp_get_attachment_file_path()`

### LINKED ISSUE #499: Correct docblock for wp_get_attachment_file_path()
## Bug Description
The inline document specifies that `thumbnail` is the default size instead of `medium`
Ref: https://github.com/WordPress/performance/blob/trunk/modules/images/dominant-color/load.php#L322

dilipbheda:
Fixes #499
## Checklist

### DIFF
--- a/modules/images/dominant-color/load.php
+++ b/modules/images/dominant-color/load.php
@@ -319,7 +319,7 @@ function _dominant_color_get_dominant_color_data( $attachment_id ) {
  * @param string $size          Optional. Image size. Default 'thumbnail'.
  * @return false|string Path to an image or false if not found.
  */
-function wp_get_attachment_file_path( $attachment_id, $size = 'medium' ) {
+function wp_get_attachment_file_path( $attachment_id, $size = 'thumbnail' ) {
 	$imagedata = wp_get_attachment_metadata( $attachment_id );
 	if ( ! is_array( $imagedata ) ) {
 		return false;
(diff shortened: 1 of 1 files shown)


modules/images/dominant-color/load.php
```diff
@@ -319,7 +319,7 @@ function _dominant_color_get_dominant_color_data( $attachment_id ) {
  * @param string $size          Optional. Image size. Default 'thumbnail'.
  * @return false|string Path to an image or false if not found.
  */
-function wp_get_attachment_file_path( $attachment_id, $size = 'medium' ) {
+function wp_get_attachment_file_path( $attachment_id, $size = 'thumbnail' ) {
```
mukeshpanchal27: Why you want to update the size to `thumbnail`?

mukeshpanchal27 (commented):
@dilipbheda There are no size restrictions that we require in order to function.

The size changes break the unit tests, so if you have a strong response to this update, then you have to update the unit tests.

modules/images/dominant-color/load.php
dilipbheda: > `@param` string $size          Optional. Image size. Default 'thumbnail'.


@mukeshpanchal27 In inline document mentioned `thumbnail` default image size.
should I update the inline document instead of changing the image size?

mukeshpanchal27: Yes that make sense. Update the document only as there is no need to update size in function.

dilipbheda: @mukeshpanchal27 Done!

mukeshpanchal27 (approved):
@dilipbheda Thanks for the update. LGTM.

pbearne (approved):

mehulkaklotar (approved):

PR #502, opened 2022-08-26
Remove return type from apply_filters

### LINKED ISSUE #501: Improve inline document of webp_uploads_pre_generate_additional_image_source() filter.
## Bug Description
I think it should be better if we removed the `@return` type from apply_filters.
Ref: https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/helper.php#L83

dilipbheda:
Fixes https://github.com/WordPress/performance/issues/501


felixarntz (approved):
Thank @dilipbheda, great catch!

eugene-manuilov (approved):
LGTM

mukeshpanchal27 (approved):
Great catch @dilipbheda.

PR #504, opened 2022-08-26
Run GitHub workflows also for feature branches

felixarntz:
## Summary

As noted in e.g. #497, the new `feature/regenerate-existing-images` branch does not run our GitHub workflows. This PR adds that, generally for all `feature/*` branches, which should only be used for broader features (i.e. not PR branches in which you're actually coding).

Once this is merged into `trunk`, we can merge it back into `feature/regenerate-existing-images` to apply it there.

## Relevant technical choices

* Add `feature/**` to the allowlist of branches.

## Checklist

### DIFF
--- a/.github/workflows/js-lint.yml
+++ b/.github/workflows/js-lint.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - trunk
       - 'release/**'
-      - 'feature/**'
     # Only run if JS/JSON/Lint/NVM files changed.
     paths:
       - '.github/workflows/js-lint.yml'
--- a/.github/workflows/js-lint.yml
+++ b/.github/workflows/js-lint.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - trunk
       - 'release/**'
+      - 'feature/**'
     # Only run if JS/JSON/Lint/NVM files changed.
     paths:
       - '.github/workflows/js-lint.yml'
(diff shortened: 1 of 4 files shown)


.github/workflows/php-test.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - trunk
       - 'release/**'
+      - 'feature/**'
```
eugene-manuilov: There is probably no need in running php tests on every push to a feature branch because all tests will be executed when we create a PR for that feature against the `trunk` branch.
```suggestion
```

.github/workflows/php-lint.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - trunk
       - 'release/**'
+      - 'feature/**'
```
eugene-manuilov: Same here, we don't need to run this workflow on every push to a feature branch:
```suggestion
```

.github/workflows/js-lint.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - trunk
       - 'release/**'
+      - 'feature/**'
```
eugene-manuilov: And here:
```suggestion
```

felixarntz:
Thanks @eugene-manuilov, makes sense. Added your suggested changes.

adamsilverstein (approved):
Excellent!

eugene-manuilov (approved):
LGTM

Issue #480, opened 2022-08-16
Create Background Processing Infrastructure: Create perflab_start_background_job


mxbclang:
Create a new `perflab_start_background_job()` function to start the background process for a specific job ID by making a HTTP POST request to the `admin-ajax.php` endpoint.

## Acceptance Criteria
- Function accepts the following parameters
    - `$job_id` - A integer that is the ID for a job returned by the `perflab_create_background_job` function
- The function will make a HTTP request to the `admin-ajax.php` endpoint with the `wp_remote_post` along with the following parameters
    - action: `wp_ajax_background_process_handle_request`
    - nonce: `wp_create_nonce( 'wp_ajax_background_process_handle_request' )`
    - timeout: `0.1`
    - data: `array( $job_id )`
    - blocking: `false`
    - sslverify: `apply_filters( 'https_local_ssl_verify', false )`
- The function will return the following depending on success or failure
    - On failure the function returns the `WP_Error` from `wp_remote_post`
    - On success, the function returns `true`

PR #507, opened 2022-08-30
Background job class

ankitrox:
## Summary

Background job class to manage the life cycle of a background job.

This class is responsible for:

1. Creation of job in the queue (taxonomy).
2. Locking/unlocking a particular job.
3. Job status management (running, failed, complete etc.)
4. Logging the errors for job.
5. Data retrieval of a job.

Fixes #503 #479 

## Checklist

- [x] PR has either `[Focus]` or `Infrastructure` label.
- [x] PR has a `[Type]` label.
- [x] PR has a milestone or the `no milestone` label.

Issue #490, opened 2022-08-19
Create Background Processing Infrastructure: Create perflab_background_process_next_batch function


jjgrainger:
Create a new `perflab_background_process_next_batch` function to handle subsequent requests to process additional job batches.

This function is created in the module file called `modules/images/regenerate-existing-images/background-process.php`

## Acceptance Criteria
- Function accepts the following parameters
    - `$job_id` - A integer that is the ID for a job to run the next batch for.
- The function will create a new requests key to pass in the request.
    - Generate a random key with the `wp_generate_password` function
    - Save that key to the options table with the key `background_process_key_{$job_id}`
- The function will make a HTTP request to the `admin-ajax.php` endpoint with the `wp_remote_post` along with the following parameters
    - action: `wp_ajax_background_process_next_batch`
    - timeout: `0.1`
    - data: `array( 'job_id' => $job_id, 'key' => $key )`
    - blocking: `false`
    - sslverify: `apply_filters( 'https_local_ssl_verify', false )`
- The function will return the following depending on success or failure
    - On failure the function returns the `WP_Error` from `wp_remote_post`
    - On success, the function returns `true`

Issue #484, opened 2022-08-16
Create Background Processing Infrastructure: Create Status Cron Event


mxbclang:
Create a Status Cron Event to check jobs are running correctly and remove old completed jobs.

## Acceptance Criteria

**Cron Schedule**
- Scheduled a new event using the [wp_schedule_event](https://developer.wordpress.org/reference/functions/wp_schedule_event/) function.
- The event is scheduled with the following parameters
    - `timestamp`: `time()`
    - `recurrence`: `hourly`
    - `hook`: `perflab_background_process_status_check`

**Event hook**
Create a new function for the `perflab_background_process_status_check` hook that checks the following

**Check jobs are running correctly**
- Query jobs that are currently running with `get_terms`
- Retrieve running jobs by querying for terms with the `job_status` meta and the value `running`.
- If no terms are returned continue to next check.
- If terms are passed, loop over each term and check the following:
    - Retrieve the `job_lock` term meta value
    - Check the job_lock value against the `ini_get( 'max_execution_time' )`
    - If the `job_lock` meta exists and the value is less than the `max_execution_time` continue to the next term
    - If the `job_lock` meta does not exist or the value is greater the `max_execution_time` restart the job by calling `perflab_start_background_job` passing the `$job_id`

**Remove old completed jobs**
- Query jobs that are currently running with `get_terms`
- Retrieve running jobs by querying for terms with the `job_status` meta and the value `completed`.
- If no terms are returned exit early.
- If terms are passed, loop over each term and check the following:
    - Retrieve the `job_completed_at` term meta value
    - If the job_completed_at meta exists and the timestamp value is older than 7 days, the term is deleted.

PR #512, opened 2022-09-05
Background process runner

### LINKED ISSUE #482: Create Background Processing Infrastructure: Perflab_Background_Process class

### LINKED ISSUE #493: Create Background Processing Infrastructure: Create Perflab_Background_Process::memory_exceeded method
Create a new `Perflab_Background_Process::memory_exceeded` method to calculate and return the memory used within the background process. This functions checks the total memory used is 90% of the servers memory limit. The boolean returned is also filterable so other developers can set the memory limit.

## Acceptance Criteria
- Create a new `memory_exceeded` method the `Perflab_Background_Process` class
- The method accepts no parameters
- The method returns a boolean
- The total memory limit is calculated by taking `ini_get( 'memory_limit' )` or a fallback value of `128M`. This value is then converted to bytes using the `wp_convert_hr_to_bytes` function and multiplied by `0.9` to calculate 90% of the total memory limit.
- The calculated memory limit is tested by checking if it is `>=` to the value returned by `memory_get_usage()`
- The boolean value returned is filtered using apply_filters( `perflab_background_process_memory_exceeded`, $return );`

### LINKED ISSUE #494: Create Background Processing Infrastructure: Create Perflab_Background_Process::time_exceeded method

ankitrox:
## Summary

Create a background process runner class. This is mainly responsible for:

1. Handling the background job call via ajax action.
2. Validating the parameters.
3. Running the process.
4. Watching the process for time and memory exhaustion.
5. Terminating the process.
6. Error handling.

Fixes #482 #493 #494 

## Checklist

### DIFF
--- a/modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
+++ b/modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
@@ -24,6 +24,15 @@ class Perflab_Background_Process {
 	const BG_PROCESS_ACTION = 'perflab_background_process_handle_request';
 
 	/**
+	 * Name of the ajax action which will trigger next batch of background process to run the job.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @var string
+	 */
+	const BG_PROCESS_NEXT_BATCH_ACTION = 'perflab_background_process_next_batch';
+
+	/**
 	 * Job instance.
 	 *
 	 * @since n.e.x.t
@@ -41,30 +50,44 @@ class Perflab_Background_Process {
 	 */
 	public function __construct() {
 		// Handle job execution request.
-		add_action( 'wp_ajax_' . Perflab_Background_Process::BG_PROCESS_ACTION, array( $this, 'handle_request' ) );
+		add_action( 'wp_ajax_' . self::BG_PROCESS_ACTION, array( $this, 'handle_request' ) );
+		add_action( 'wp_ajax_' . self::BG_PROCESS_NEXT_BATCH_ACTION, array( $this, 'handle_request' ) );
 	}
 
 	/**
 	 * Handles incoming request to run the batch of job.
 	 *
 	 * @since n.e.x.t
-	 *
-	 * @throws Exception Invalid request for background process.
 	 */
 	public function handle_request() {
-		// Exit if nonce varification fails.
-		$nonce_check = check_ajax_referer( self::BG_PROCESS_ACTION, 'nonce', false );
-
-		// If nonce check fails, fallback to key checking.
-		if ( false === $nonce_check ) {
-			throw new Exception( __( 'Invalid nonce passed to process.', 'performance-lab' ) );
-		}
-
 		$job_id = isset( $_REQUEST['job_id'] ) ? absint( sanitize_text_field( $_REQUEST['job_id'] ) ) : 0;
 
 		// Job ID is mandatory to be specified.
 		if ( empty( $job_id ) ) {
-			throw new Exception( __( 'No job specified to execute.', 'performance-lab' ) );
+			wp_send_json_error( __( 'No job specified to execute.', 'performance-lab' ), 400 );
+		}
+
+		$current_action = current_action();
+
+		if ( 'wp_ajax_' . self::BG_PROCESS_ACTION === $current_action ) {
+			// Exit if nonce varification fails.
+			$nonce_check = check_ajax_referer( self::BG_PROCESS_ACTION, 'nonce', false );
+
+			// If nonce check fails, fallback to key checking.
+			if ( false === $nonce_check ) {
+				wp_send_json_error( __( 'Invalid nonce passed to process.', 'performance-lab' ), 400 );
+			}
+		} elseif ( 'wp_ajax_' . self::BG_PROCESS_NEXT_BATCH_ACTION === $current_action ) {
+			/**
+			 * Check if the key in request matches that of in the options.
+			 */
+			$request_key = isset( $_REQUEST['key'] ) ? urldecode_deep( sanitize_text_field( $_REQUEST['key'] ) ) : '';
+			$stored_key  = get_option( 'background_process_key_' . $job_id );
+
+			// Check if key is present in request.
+			if ( empty( $request_key ) || $stored_key !== $request_key ) {
+				wp_send_json_error( __( 'Key passed to the request is not valid.', 'performance-lab' ), 400 );
+			}
 		}
 
 		$this->job = perflab_get_background_job( $job_id );
@@ -120,11 +143,6 @@ class Perflab_Background_Process {
 	 * @param int $job_id Job ID. This is the term id from `background_job` taxonomy.
 	 */
 	private function next_batch( $job_id ) {
-		// Do not call the background process from within the script if the real cron has been setup to do so.
-		if ( defined( 'ENABLE_BG_PROCESS_CRON' ) ) {
-			return;
-		}
-
 		perflab_start_background_job( $job_id );
 	}
 
--- a/modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
+++ b/modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
@@ -81,7 +81,7 @@ class Perflab_Background_Process {
 			/**
 			 * Check if the key in request matches that of in the options.
 			 */
-			$request_key = isset( $_REQUEST['key'] ) ? urldecode_deep( sanitize_text_field( $_REQUEST['key'] ) ) : '';
+			$request_key = isset( $_REQUEST['key'] ) ? urldecode_deep( $_REQUEST['key'] ) : '';
 			$stored_key  = get_option( 'background_process_key_' . $job_id );
 
 			// Check if key is present in request.
--- a/modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
+++ b/modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
@@ -118,9 +118,6 @@ class Perflab_Background_Process {
 			 */
 			do_action( 'perflab_job_' . $this->job->get_name(), $this->job->get_data() );
 		} while ( ! $this->memory_exceeded() && ! $this->time_exceeded() && ! $this->job->is_completed() );
-
-		// Once job ran successfully, change its status to queued.
-		$this->job->set_status( Perflab_Background_Job::JOB_STATUS_PARTIAL );
 	}
 
 	/**
@@ -141,7 +138,7 @@ class Perflab_Background_Process {
 		}
 
 		$nonce  = wp_create_nonce( self::BG_PROCESS_ACTION );
-		$job_id = $this->job->job_id;
+		$job_id = $this->job->get_id();
 
 		$url    = admin_url( 'admin-ajax.php' );
 		$params = array(
--- a/modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
+++ b/modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
@@ -192,9 +192,11 @@ class Perflab_Background_Process {
 			/**
 			 * Consumer code will hook to this action to perform necessary tasks.
 			 *
+			 * @since n.e.x.t
+			 *
 			 * @param array $data Job data.
 			 */
-			do_action( 'perflab_job_' . $this->job->get_identifier(), $this->job->get_data() );
+			do_action( 'perflab_job_' . $this->job->get_action(), $this->job->get_data() );
 		} while ( ! $this->memory_exceeded() && ! $this->time_exceeded( $this->job->get_id() ) && ! $this->job->is_completed() );
 	}
 
--- a/modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
+++ b/modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
@@ -86,7 +86,7 @@ class Perflab_Background_Process {
 		} catch ( Exception $e ) {
 			if ( $this->job instanceof Perflab_Background_Job ) {
 				$error = new WP_Error( 'perflab_job_failure', $e->getMessage() );
-				$this->job->record_error( $error );
+				$this->job->set_error( $error );
 			}
 		} finally {
 			if ( $this->job instanceof Perflab_Background_Job ) {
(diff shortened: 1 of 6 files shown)


modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
```diff
@@ -0,0 +1,236 @@
...
+class Perflab_Background_Process {
+	/**
+	 * Name of the action which will trigger background
+	 * process to run the job.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @var string Background process action name.
+	 */
+	const BG_PROCESS_ACTION = 'background_process_handle_request';
```
jjgrainger: @ankitrox we should prefix this action name with `perflab_`
```suggestion
	const BG_PROCESS_ACTION = 'perflab_background_process_handle_request';
```

modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
```diff
@@ -0,0 +1,236 @@
...
+		$current_time       = time();
+		$run_start_time     = $this->job->get_start_time();
+		$min_execution_time = 20; // Default to 20 seconds. Almost, all servers will have this much of time.
+
+		if ( function_exists( 'ini_get' ) ) {
+			$time               = ini_get( 'max_execution_time' );
+			$max_execution_time = ( ! empty( $time ) && ( $time > $min_execution_time ) ) ? $time - 10 : $min_execution_time;
+		}
+
+		return ( $current_time >= ( $run_start_time + $max_execution_time ) );
```
jjgrainger: @ankitrox we should also use `apply_filters` on the value returned here as we have with the memory limit. The filter should be `perflab_background_process_time_exceeded`
```suggestion
		$time_exceeded = ( $current_time >= ( $run_start_time + $max_execution_time ) );
		return apply_filters( 'perflab_background_process_time_exceeded', $time_exceeded );
```

modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
```diff
@@ -0,0 +1,236 @@
...
+	 * Handle incoming request to run the batch of job.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @throws Exception Invalid request for background process.
+	 *
+	 * @return void
+	 */
+	public function handle_request() {
+		try {
```
jjgrainger: @ankitrox I feel like this method and try block does too much? I think we can split some of the logic between the `handle_request()` and `run()` methods. I think this because we have exceptions thrown that don't get caught as the job is not available.

The way I see it, `handle_request()` should verify and determine the request is valid. If not, the method should return early, or maybe even return a [`wp_send_json_error`](https://developer.wordpress.org/reference/functions/wp_send_json_error/) error message? If the request is valid it then calls the `run()` method,

Then the run method is responsible for running the job and I think this could contain everything from the `$this->job->lock()` line down.

modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
```diff
@@ -0,0 +1,236 @@
...
+	 */
+	private function run() {
+		do {
+			/**
+			 * Consumer code will hook to this action to perform necessary tasks.
+			 *
+			 * @param array $data Job data.
+			 */
+			do_action( 'perflab_job_' . $this->job->get_name(), $this->job->get_data() );
+		} while ( ! $this->memory_exceeded() && ! $this->time_exceeded() );
```
jjgrainger: @ankitrox we should also check if the job is marked as complete here too.

```suggestion
		} while ( ! $this->memory_exceeded() && ! $this->time_exceeded() && ! $this->job->is_completed() );
```

modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
```diff
@@ -0,0 +1,236 @@
...
+	 * process specific action to continue executing the job in a new process.
+	 *
+	 * @return void
+	 */
+	private function next_batch() {
+		/**
+		 * Do not call the background process from within the script if the
+		 * real cron has been setup to do so.
+		 */
+		if ( defined( 'ENABLE_BG_PROCESS_CRON' ) ) {
```
jjgrainger: @ankitrox I think the configuration we have defined so far is to disable this background process entirely with a `DISABLE_BACKGROUND_PROCESS` constant. I don't believe we have the infrastructure defined to have the WP Cron handle the additional batches.

I think we should remove this here and consider where the `DISABLE_BACKGROUND_PROCESS` check would need to be made.

jjgrainger (changes requested):
@ankitrox Looking good so far. I've left a couple of comments to address.

Additionally, there are a couple of other things. One is the failing tests on this PR the second is the inclusion of the Job class which is also in #507 and should not be present in this PR.

modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
ankitrox: We will also have this constant to let the real cron job call background process. In such cases, script would not call next batch of the job from within itself.
This is discussed with @felixarntz in alternate approaches to improve background process workflow.

ankitrox:
> @ankitrox Looking good so far. I've left a couple of comments to address.
> 
> Additionally, there are a couple of other things. One is the failing tests on this PR the second is the inclusion of the Job class which is also in #507 and should not be present in this PR.

@jjgrainger Let's aim to get [background class PR](https://github.com/WordPress/performance/pull/507) merge first. As process runner is using job instance, I have merged it in this PR. Once background job class is merged, we can get this PR ready for review to others.

jjgrainger:
@ankitrox I think this PR is becoming too large and tackling too many parts of the background process.

For starters, I would argue the work for status cron #484 can be moved out of the background process runner and into its own PR/file/class. I think this should definitely be moved out of the PR and into a new one focused on fixing that specific issue.

Additionally, this PR seems to also fix #480 #481 and even #490 #483, which were already being addressed in other PR's. We want to keep these PR's smaller to make the review process more focused and less complex.

Can you update this PR to focus on the acceptance criteria in #482 #493 #494.

Let me know if you have any questions.

ankitrox:
> @ankitrox I think this PR is becoming too large and tackling too many parts of the background process.
> 
> For starters, I would argue the work for status cron #484 can be moved out of the background process runner and into its own PR/file/class. I think this should definitely be moved out of the PR and into a new one focused on fixing that specific issue.
> 
> Additionally, this PR seems to also fix #480 #481 and even #490 #483, which were already being addressed in other PR's. We want to keep these PR's smaller to make the review process more focused and less complex.
> 
> Can you update this PR to focus on the acceptance criteria in #482 #493 #494.
> 
> Let me know if you have any questions.

Thanks @jjgrainger 
I have removed the cron related code from PR. I will create a new PR for that one.
Also, updated the dependent issue in PR description.

modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
```diff
@@ -0,0 +1,247 @@
...
+	 * Perflab_Background_Process constructor.
+	 *
+	 * Adds the necessary hooks to trigger process handling.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function __construct() {
+		// Handle job execution request.
+		add_action( 'wp_ajax_' . Perflab_Background_Process::BG_PROCESS_ACTION, array( $this, 'handle_request' ) );
+		add_action( 'wp_ajax_nopriv_' . Perflab_Background_Process::BG_PROCESS_ACTION, array( $this, 'handle_request' ) );
```
jjgrainger: @ankitrox we should remove this action as we do not want unauthenticated requests triggering the background process. We should stick with the `wp_ajax_perflab_background_process_handle_request` action only.

modules/images/regenerate-existing-images/helper.php
```diff
@@ -71,3 +71,66 @@ function perflab_delete_background_job( $job_id ) {
...
+}
+
+/**
+ * Dispatches the request to trigger the background job.
+ *
+ * @since n.e.x.t
+ *
+ * @param int $job_id Job ID.
+ */
+function perflab_dispatch_background_process_request( $job_id ) {
```
jjgrainger: @ankitrox based on the acceptance criteria in #480 this function should be named `perflab_start_background_job`

modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
```diff
@@ -0,0 +1,247 @@
...
+	 *
+	 * @param int $job_id Job ID. This is the term id from `background_job` taxonomy.
+	 */
+	private function next_batch( $job_id ) {
+		// Do not call the background process from within the script if the real cron has been setup to do so.
+		if ( defined( 'ENABLE_BG_PROCESS_CRON' ) ) {
+			return;
+		}
+
+		perflab_dispatch_background_process_request( $job_id );
```
jjgrainger: @ankitrox we are handling next batch requests differently to the start request. See #490 for the AC's here. The function name that should be called is `perflab_background_process_next_batch`.

modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
jjgrainger: Thanks @ankitrox I'm not sure how this works in the grand scheme of things as we haven't defined or built the part that handles subsequent batches. I think we can approach this change in 1 of 2 ways.

1. Keep the constant as part of this PR, but make a change as this is used twice in the code, once here and again inside the `perflab_dispatch_background_process_request`. I think we should aim to only have to check this constant once. 

2. Remove the constant from the PR and we can add it as part of a new PR and stream of work to introduce this functionality.

I think in both cases we should also create a new GitHub issue with acceptance criteria to outline the work.

cc @felixarntz

modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	/**
+	 * Handles incoming request to run the batch of job.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @throws Exception Invalid request for background process.
+	 */
+	public function handle_request() {
```
jjgrainger, line 62: @ankitrox I think this method does too much and should only be responsible for handling the HTTP request and calling the `run()` method. I would make the following changes:

- Remove the try/catch block and exceptions - as we're handling the HTTP request we can return early with a error response. This could be a combination of using `WP_Error` instead of `Exception` and returning it with [`wp_send_json_error`](https://developer.wordpress.org/reference/functions/wp_send_json_error/)
- The final part of the `handle_request()` method should be the `$this->run()` call.
- The `unlock()` and `next_batch()` lines should be moved to the end of the `run()` method. Reasoning here is that this logic is part of the job running processes. If we were to trigger running a job via a different method (WP_Cron) calling `$this->run()` would still work as expected.
- If you wanted you can use a try/catch block inside the `run()` method to catch any system errors while jobs are running and handle those properly.

Let me know if this makes sense, happy to discuss on a call

jjgrainger (changes requested):
Thanks @ankitrox left some comments but looking good so far

modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
ankitrox: There's only a difference of `$key` in `start_job` and `next_batch` functions. I think we can somehow combine/compose these functions to avoid rewriting entire thing twice.

ankitrox:
> Thanks @ankitrox I'm not sure how this works in the grand scheme of things as we haven't defined or built the part that handles subsequent batches. I think we can approach this change in 1 of 2 ways.
> 
> Keep the constant as part of this PR, but make a change as this is used twice in the code, once here and again inside the perflab_dispatch_background_process_request. I think we should aim to only have to check this constant once.
> 
>  Remove the constant from the PR and we can add it as part of a new PR and stream of work to introduce this functionality.
> 
>  I think in both cases we should also create a new GitHub issue with acceptance criteria to outline the work.
> 
>  cc @felixarntz
> 

@jjgrainger  Regarding this comment, I think option #2 is better; to handle this in separate issue and PR.

modules/images/regenerate-existing-images/background-process/class-perflab-background-process.php
ankitrox: Made the proposed changes.
I think adding `try catch` to run method would make more sense when we will implement error handling.

adamsilverstein:
@ankitrox looks like this PR has gone stale, can we go ahead and close it?

adamsilverstein (commented):
is this still needed?

westonruter:
@adamsilverstein looking back at old PRs from this time period, it seems the main push was to add image regeneration as a background process which this PR would serve as infrastructure for. It seems this topic hasn't been on the radar for quite some time, so potentially several such PRs should be closed. Not sure why that got deprioritized.

swissspidy:
Closing for now to clean up the repo since this plugin/module is not a priority right now.

PR #515, opened 2022-09-07
Add JavaScript coding standard workflow

### LINKED ISSUE #459: Add JavaScript coding standard workflow
## Bug Description
We should add a JavaScript coding standard because we have now JavaScript in the plugin. I got some JavaScript CS error when I use https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/fallback.js for the WebP core patch.

mukeshpanchal27:
## Summary
I'm referring WordPress JavaScript coding standard workflow: https://github.com/WordPress/wordpress-develop/blob/trunk/.github/workflows/coding-standards.yml

Fixes #459

## Checklist


mehulkaklotar (commented):
@mukeshpanchal27 We should migrate phpcs checks to this configuration, as there will be two different configurations for the `php-lint.yml` and `js-lint.yml` as well as this `conding-standards.yml`. I think it is better to have them in one file instead of separate files. Or keep them separate but these new jshint change should be in `js-lint.yml`. What do you say? 🤔

mukeshpanchal27:
> @mukeshpanchal27 We should migrate phpcs checks to this configuration, as there will be two different configurations for the `php-lint.yml` and `js-lint.yml` as well as this `conding-standards.yml`. I think it is better to have them in one file instead of separate files. Or keep them separate but these new jshint change should be in `js-lint.yml`. What do you say? 🤔

@mehulkaklotar I merged the workflow in `js-lint.yml`

.github/workflows/js-lint.yml
```diff
@@ -49,4 +53,49 @@ jobs:
...
+  # - Logs debug information about the GitHub Action runner.
+  # - Installs NodeJS.
+  # - Logs updated debug information.
+  # - Installs NPM dependencies.
+  # - Run the WordPress JSHint checks.
+  # - Ensures version-controlled files are not modified or deleted.
+  jshint:
+    name: coding standards
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
```
adamsilverstein, line 73: does the timeout need to be this high?

adamsilverstein:
Overall this looks good, however I had some trouble getting it working. 

Running `grunt jshint` here seemed to work, but I noticed it isn't picking up changes for me in `modules/images/webp-uploads/fallback.js`. I tried introducing errors there and the command didn't report them. What command did you run to catch the fixes you introduced?

I noticed these lines - https://github.com/WordPress/performance/blob/b9b62cc90f46a83e565290593f1accea3c12e1f9/package.json#L24-L25
seem to run the checks agains `./bin`, shouldn't those instead run against `modules/**/*.js`?

ntwb:
> I noticed these lines -
> 
> https://github.com/WordPress/performance/blob/b9b62cc90f46a83e565290593f1accea3c12e1f9/package.json#L24-L25

This comes from the @wordpress/scripts package, and I’d suggest using the ESLint linting included in this package rather than JSHint, more issues will be flagged along with additional coding standards that WP JSHint package doesn’t include, the Gutenberg repo uses this ESLint configuration and when merged back into the WordPress repo will also pass the JSHint linting

.github/workflows/js-lint.yml
mukeshpanchal27: It was previously used in workflow https://github.com/WordPress/performance/blob/trunk/.github/workflows/js-lint.yml#L40 so i put it here. I don't think we need it.

mukeshpanchal27:
Strange behaviour I run a similar JS lint on the WP repository, but it returns no similar JS errors. https://github.com/WordPress/wordpress-develop/pull/3207

westonruter:
@mukeshpanchal27 I guess this PR is obsolete?

mukeshpanchal27:
Yup. Closing now.

PR #517, opened 2022-09-08
Added support for attachment image html markup

maheshwaghmare:
## Summary

After enabling the SVG image support when we use the function `wp_get_attachment_image()` used as:

```php
$html = wp_get_attachment_image( $image_id, $size );
```

Then the HTML is generated as:

```html
<img width="150" height="150" src="http://example.com/wp-content/uploads/2022/09/dummy-150x150.jpg" class="alignright not-transparent" alt="" loading="lazy" data-has-transparency="false" data-dominant-color="a26fd1" style="--dominant-color: #a26fd1;">
```

It keep serve the `.jpg` image version instead of `.webp`

This PR add the support with filter `wp_get_attachment_image` same as added support for `post_thumbnail_html`.

After adding the support the HTML markup generated as:

```html
<img width="150" height="150" src="http://example.com/wp-content/uploads/2022/09/dummy-150x150-jpg.webp" class="alignright not-transparent" alt="" loading="lazy" data-has-transparency="false" data-dominant-color="a26fd1" style="--dominant-color: #a26fd1;">
```

## Checklist

### DIFF
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -628,6 +628,8 @@ add_filter( 'post_thumbnail_html', 'webp_uploads_update_featured_image', 10, 3 )
  * Updates the references of the attachment image to the a new image format if available, in the same way it
  * occurs in the_content of a post.
  *
+ * @since n.e.x.t
+ *
  * @param string $html          The current HTML markup of the image.
  * @param int    $attachment_id The ID of the attachment image.
  * @return string The updated HTML markup.
--- a/modules/images/webp-uploads/load.php
+++ b/modules/images/webp-uploads/load.php
@@ -628,8 +628,6 @@ add_filter( 'post_thumbnail_html', 'webp_uploads_update_featured_image', 10, 3 )
  * Updates the references of the attachment image to the a new image format if available, in the same way it
  * occurs in the_content of a post.
  *
- * @since n.e.x.t
- *
  * @param string $html          The current HTML markup of the image.
  * @param int    $attachment_id The ID of the attachment image.
  * @return string The updated HTML markup.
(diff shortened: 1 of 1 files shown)


modules/images/webp-uploads/load.php
```diff
@@ -624,6 +624,19 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
 }
 add_filter( 'post_thumbnail_html', 'webp_uploads_update_featured_image', 10, 3 );
 
+/**
+ * Updates the references of the attachment image to the a new image format if available, in the same way it
+ * occurs in the_content of a post.
+ *
```
mukeshpanchal27, line 630: Add `@since`.

```suggestion
 *
 * @since n.e.x.t
 *
```

modules/images/webp-uploads/load.php
```diff
@@ -624,6 +624,19 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+ * @param string $html          The current HTML markup of the featured image.
```
mukeshpanchal27:
```suggestion
 * @param string $html          The current HTML markup of the image.
```

mukeshpanchal27 (commented):
@maheshwaghmare Thanks for the PR. Can you please correct the unit tests?

@felixarntz I think it's good to update the markup for `wp_get_attachment_image`. Please share your thoughts on this.

felixarntz (changes requested):
@maheshwaghmare Thank you for the PR and for contributing to the performance plugin.

Unfortunately, at this point making this change is a bit preliminary and goes beyond what the current plan of the WebP module is, which is to automatically modify in-content images. Images retrieved via `wp_get_attachment_image()` are controlled by code (e.g. in a theme), and for now it was decided to leave control about that to the theme owner. For example, @jjgrainger has already been working on a core PR https://github.com/WordPress/wordpress-develop/pull/3074 for this (which is closed for now, since this infrastructure will currently not proceed).

I think it's great you are raising this problem again in this PR though, and I agree we should think through it again, so I opened #523 for this purpose. Let's discuss there first whether that makes sense and what the potential complications are, before potentially proceeding with this PR.

Only marking this as requesting changes for now to first have this conversation.

PR #527, opened 2022-09-13
Generate only WebP images by default for JPEG and WebP uploads

### LINKED ISSUE #524: Modify WebP default behavior to follow latest decision (generate only WebP for JPEG)
As part of https://core.trac.wordpress.org/ticket/55443, it was recently decided to change the default behavior of the WebP implementation back to generating only WebP versions for the sub-sizes of a JPEG upload, rather than both WebP and JPEG versions for every size.

We should align with this in the plugin, i.e. adjust the default value for the `webp_uploads_upload_image_mime_transforms` filter, so that both `image/jpeg` and `image/webp` result in only `image/webp`.

While the [short-term future of the WebP feature is unclear](https://make.wordpress.org/core/2022/09/11/webp-in-core-for-6-1/), we should definitely align the plugin implementation with the latest decision made for the core side, as that is the safer default.

eugene-manuilov:
## Summary

Fixes #524

## Relevant technical choices

- Changes in other files are required to fix issues that have been found after removing the `image/jpeg` mime type from the list of allowed transformations for jpeg images.

## Checklist


modules/images/webp-uploads/helper.php
```diff
@@ -16,7 +16,7 @@
  */
 function webp_uploads_get_upload_image_mime_transforms() {
 	$default_transforms = array(
-		'image/jpeg' => array( 'image/jpeg', 'image/webp' ),
+		'image/jpeg' => array( 'image/webp' ),
 		'image/webp' => array( 'image/webp', 'image/jpeg' ),
```
felixarntz: We should remove `image/jpeg` from here as well for the default (replying to your question from https://github.com/WordPress/performance/issues/524#issuecomment-1245114194).

modules/images/webp-uploads/helper.php
```diff
@@ -144,7 +144,10 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 		$destination_file_name = $editor->generate_filename( $suffix, null, $extension[0] );
 	}
 
+	remove_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
 	$image = $editor->save( $destination_file_name, $mime );
+	add_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
```
felixarntz, line 149: Can you clarify why you add this change? Would be great to add a one line comment on why the filter needs to be temporarily removed here.

I assume it is because here we output a WebP already if needed, so we shouldn't inside of it have the mapping of WebP? Which problems did you get during testing? However, I don't understand how exactly this causes a problem. If it's just about test failures, maybe it's our tests that need changes, but when thinking through the logic here, I can't see how having this filter run here would be problematic. Let me clarify by example:
1. User uploads JPEG.
2. Our `image_editor_output_format` callback fires and changes the default output for the JPEG to WebP, since that's the only one in the array.
3. Then our custom logic in `webp_uploads_create_sources_property()` runs. It shouldn't really do anything because the only MIME type to output is WebP which WordPress should have done already, so I think there may be a flaw in our existing logic. But even if it then went through the array of MIME types, it would only generate WebP (again).
4. In that case it would call `webp_uploads_generate_additional_image_source()` passing `image/webp` for the `$mime_type`, so that would then be passed to the `WP_Image_Editor::save()` method.
5. That would trigger the `image_editor_output_format` filter again, however passing `image/webp` to it. Our own callback would then (unnecessarily, to be fair) look at WebP output mappings and see that `image/webp` is an output mapping there, so it wouldn't tweak the list, so we would still get a WebP out of that.

In other words, while it may be okay to remove the filter here temporarily, I am questioning that that is really _needed_. We may have problems somewhere else though because this logic here should never run in the first place if WordPress only generates one MIME type.

I wonder if the problem is around https://github.com/WordPress/performance/blob/1.4.0/modules/images/webp-uploads/load.php#L42-L67: We look at the attachment MIME type, which will remain `image/jpeg` if the uploaded image is a JPEG. However, even the "full" image (`_wp_attached_file`) may already be a WebP image, if it was downscaled (`-scaled` file name suffix), because WP core then already ran the `image_editor_output_format` filter. So I think our assumption there that `get_post_mime_type()` is the MIME type of `get_attached_file()` is incorrect. We still need to call `get_post_mime_type()`, but only to know what was originally uploaded - that file may however be completely unused, if WordPress downscaled the image (which it would already generate a WebP for instead of JPEG). Because of that, we then set an incorrect source for `image/jpeg` which is actually a WebP image, and that then results in the `foreach` loop below to try to generate `image/webp` (again), since it doesn't find it in the sources. Did that maybe lead to the problem you were running into?

modules/images/webp-uploads/image-edit.php
```diff
@@ -187,7 +187,10 @@ function ( $metadata, $post_meta_id ) use ( $post_id, $file_path, $mime_type, $e
 					// Create a file with then new extension out of the targeted file.
 					$target_file_name     = preg_replace( "/\.$current_extension$/", ".$extension", $thumbnail_file );
 					$target_file_location = path_join( $original_directory, $target_file_name );
-					$result               = $editor->save( $target_file_location, $targeted_mime );
+
+					remove_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
+					$result = $editor->save( $target_file_location, $targeted_mime );
+					add_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
```
felixarntz, line 193: Probably makes sense to do this here, but still see my comment above. I think we have another problem that is more critical and we should focus on.

If the output configuration for `image/jpeg` does only include a single MIME type (`image/webp`), I don't even think this code needs to be run at all. WordPress core handles a single MIME type perfectly fine, so all of this here seems irrelevant. It is needed for multiple MIME types, but when the output filter says to only use one output MIME type, almost nothing of the module is needed in fact (so we could add early returns to a bunch of entry filter callbacks).

Maybe I'm oversimplifying this, but again, I don't think we should _need_ to temporarily remove the filter here and elsewhere in this PR, the problems seem to be somewhere else. If our tests are failing without this, maybe it is because our tests assume there are multiple MIME types in the output for each MIME type? Maybe for those tests, we'll need to manually add a filter to ensure those tests always run with multiple MIME types (e.g. `'image/jpeg' => array( 'image/webp', 'image/jpeg' )`).

modules/images/webp-uploads/image-edit.php
```diff
@@ -196,7 +199,10 @@ function ( $metadata, $post_meta_id ) use ( $post_id, $file_path, $mime_type, $e
 					$subsized_images[ $targeted_mime ] = array( 'thumbnail' => $result );
 				} else {
 					$destination = trailingslashit( $original_directory ) . "{$filename}.{$extension}";
-					$result      = $editor->save( $destination, $targeted_mime );
+
+					remove_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
+					$result = $editor->save( $destination, $targeted_mime );
+					add_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
```
felixarntz, line 205: See above.

modules/images/webp-uploads/load.php
```diff
@@ -97,7 +97,10 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 			continue;
 		}
 
-		if ( webp_uploads_should_discard_additional_image_file( $metadata, $image ) ) {
+		if (
+			! empty( $metadata['sources'][ $mime_type ] ) &&
+			webp_uploads_should_discard_additional_image_file( $metadata['sources'][ $mime_type ], $image )
```
felixarntz: Oh wow, that's a critical fix I guess? The `$metadata` above looks completely wrong, since it's the full attachment metadata right? Guess we never found this before.

modules/images/webp-uploads/load.php
```diff
@@ -137,7 +140,10 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 			$properties['sources'] = array();
 		}
 
-		if ( empty( $properties['sources'][ $current_mime ] ) ) {
+		if (
+			empty( $properties['sources'][ $current_mime ] ) &&
+			in_array( $current_mime, $valid_mime_transforms[ $mime_type ], true )
```
felixarntz: Why is this needed? If the size already has a file of `$current_mime`, whatever the MIME type is WordPress would only have generated it if it should. If without this condition we get a problem, then I think the bug is somewhere else.

modules/images/webp-uploads/load.php
```diff
@@ -162,7 +168,10 @@ function webp_uploads_create_sources_property( array $metadata, $attachment_id )
 				continue;
 			}
 
-			if ( webp_uploads_should_discard_additional_image_file( $properties, $source ) ) {
+			if (
+				! empty( $metadata['sources'][ $mime_type ] ) &&
+				webp_uploads_should_discard_additional_image_file( $properties['sources'][ $mime_type ], $source )
+			) {
```
felixarntz: While I agree on the similar change further above, I think here it shouldn't be needed: `$properties` for every image size should already have a `filesize` key from WP core, and that image should be for the primary (typically original) MIME type. Again, similar to almost all other problems here, I think the issue here may be what I outlined in my very long comment above.

felixarntz (changes requested):
@eugene-manuilov The root change here is of course rather simple, but I see this is leading to all sorts of problems now. I believe some of your fixes are a bit preliminary though. We should not focus on fixing the tests, we should focus more on fixing the expected behavior, and it is very well possible that the tests are an issue here. At a high level, I think the biggest overarching problem is the following:
* We have all this logic here to generate additional MIME type files.
* That is however mainly for the case that the input MIME type needs more than one output MIME type.
* With this PR, we're changing both `image/jpeg` and `image/webp` to have only one output MIME type though, so most of the code in the module shouldn't be run if that's the case (early returns are needed). For the single MIME type behavior, the only function that is critical is `webp_uploads_filter_image_editor_output_format()` (the `image_editor_output_format` callback), and that will have core handle everything. The only other thing that still needs to run from our plugin is to populate the `sources`, but not generate _any_ additional images, since core will already generate all the necessary images out of the box (only in e.g. WebP even though you upload a JPEG).

There are potentially a few smaller problems here or there where we assume that the uploaded image MIME type is also the MIME type that WordPress has already generated files for, which is not necessarily the case. But really, by adding early returns in most entry points here we should be able to fix almost all the problems in a more appropriate way.

modules/images/webp-uploads/helper.php
eugene-manuilov: @felixarntz, the problem is much simpler but to explain it, I need to show you an example. Let's assume we have the following image mime transformations:

```php
array(
    'image/jpeg' => array( 'image/webp' ),
)
```

Such transformation leads to the following results:

```
    ["medium"]=>
    array(5) {
      ["file"]=>
      string(25) "balloons-175-200x300.webp"
      ["width"]=>
      int(200)
      ["height"]=>
      int(300)
      ["mime-type"]=>
      string(10) "image/webp"
      ["sources"]=>
      array(2) {
        ["image/webp"]=>
        array(2) {
          ["file"]=>
          string(25) "balloons-175-200x300.webp"
          ["filesize"]=>
          int(21246)
        }
        ["image/jpeg"]=>
        array(2) {
          ["file"]=>
          string(29) "balloons-175-200x300-webp.webp"
          ["filesize"]=>
          int(20417)
        }
      }
    }
```

Pay attention to the `image/jpeg` source, it is generated as webp image again even though we wanted to create a jpeg version. Why does it happen? Because our logic in the `webp_uploads_filter_image_editor_output_format` function knows nothing about which image is being generated and updates mime type only based on the transformation settings.

In other words, the `webp_uploads_filter_image_editor_output_format` function sees that the incoming image has the `image/jpeg` mime type and tries to find the corresponding transformation in the list of transformations for the `image/jpeg` mime type. Of course, it can find only the `image/webp` mime type in this case because we have `'image/jpeg' => array( 'image/webp' )` transformation for jpeg images although the original image is in the webp format and the function should look up for the `image/webp` transformations.

An early return obviously won't help in this situation because we have a case where we have multiple mime types for an image.

Yes, you can say that it won't be the case anymore because we removed `image/jpeg` from the allowed transformations list for webp images, but this can be incorrectly configured in a 3rd party plugin that use our filter to update transformations.

modules/images/webp-uploads/helper.php
eugene-manuilov: Sounds good, updated.

modules/images/webp-uploads/load.php
eugene-manuilov: Yeah, i was also surprised to see that we use the whole metadata array here whilst we needed only source information for the current mime type.

modules/images/webp-uploads/image-edit.php
eugene-manuilov: This is added for the same reason as in the case above.

modules/images/webp-uploads/image-edit.php
eugene-manuilov: Same here, see my comments above.

modules/images/webp-uploads/load.php
eugene-manuilov: I don't know why but it wasn't the case when I tested it yesterday. Here is a var_dump result for metadata of an image in one of the tests:

```php
array(8) {
  ["width"]=>
  int(1080)
  ["height"]=>
  int(720)
  ["file"]=>
  string(21) "2022/09/leafs-572.jpg"
  ["sizes"]=> ...
  ["image_meta"]=> ...
  ["dominant_color"]=>
  string(6) "283c23"
  ["has_transparency"]=>
  bool(false)
  ["sources"]=> ...
}
```

As you can see there is no `filesize` property on the root level, that's why I updated the functionality to use the correct path to main source information that contains `filesize` data. If the `filesize` key is generated by WP core, then I wonder why it doesn't do it in phpunit tests 🤔

modules/images/webp-uploads/load.php
eugene-manuilov: If I remember correctly, the problem was that we kept generating source information for the `image/jpeg` mime type despite the fact that we were no jpeg images after converting it to webp. This fix ensures that we don't add a mime type to the sources list if it doesn't exists in the transformations list.

modules/images/webp-uploads/helper.php
felixarntz: @eugene-manuilov Yes, temporarily removing the filter makes sense. Still, I think we also need to have early returns here for the reason I outlined above: With the mapping of `image/jpeg` to `array( 'image/webp' )`, our plugin currently does way too much, and it tries to in fact generate another version of exactly the same WebP - so there are two problems here, the temporary filter removal addresses only one of them.

Let me explain, based on the current mapping:
* WordPress core, just via `webp_uploads_filter_image_editor_output_format()`, will generate WebP images for all the sizes of a JPEG upload, and even for the full / original image _if_ it gets downscaled.
* At this point, all necessary images are generated.
* We of course still need to populate `sources` for the attachment then, where our logic is already flawed: We call `get_post_mime_type()` which will be set to `image/jpeg`, but the full image _may_ in fact be `image/webp` (see above). We then store the existing full image source as `image/jpeg` although it is `image/webp`. This would then lead us to generate another "full" image WebP (basically the same that already exists), just because our logic thinks the existing image is a JPEG.

There may be a similar flaw in the sub-sizes, I haven't been able to assess that from code review yet, but we need to make sure we fix some of that logic so that we don't try to create the exact same WebP images again when they were already created by WordPress core.

Essentially: _If_ there is only one output MIME type, WordPress will have taken care of it. In that case, all our additional logic should do is populate the `sources` fields with those images, but not generate _any_ additional images. Does that make sense?

modules/images/webp-uploads/image-edit.php
felixarntz: Temporary filter removal LGTM, however note there are other flaws we need to still address elsewhere.

modules/images/webp-uploads/load.php
felixarntz: @eugene-manuilov Actually, passing `$metadata` here is not wrong. What we expect inside of this function is an array that has keys `file` and `filesize`. The overall attachment metadata _should_ have that field, at least since WordPress 6.0 (see e.g. core's `wp_generate_attachment_metadata()` and `_wp_image_meta_replace_original()`). Did you potentially run into this problem with an older image, our in a version below 6.0?

It would also be okay passing the individual source here, but technically they should have the same data, so we really need to figure out what's going wrong here. Could you follow up on that?

Side note, maybe we should raise the minimum required WordPress version to 6.0 soon - it would help this one, and also 6.1 is coming out soon, and our policy is to support at least the latest 2 versions.

modules/images/webp-uploads/load.php
felixarntz: @eugene-manuilov

> we kept generating source information for the `image/jpeg` mime type despite the fact that we were no jpeg images after converting it to webp.

This is one of the problems I was trying to convey further above. We need to fix _that_, rather than add this condition here I think.

modules/images/webp-uploads/load.php
felixarntz: @eugene-manuilov This is indeed odd, not sure why it is not included. See my comment https://github.com/WordPress/performance/pull/527#discussion_r971131216, we need to look into this further.

On another note, you're referring to the overall metadata above, but this particular comment here is about each individual size's data, not the overall metadata. The sizes should each also have a `filesize` key, as long as they were uploaded with WordPress 6.0. (Of course we have to handle this gracefully, for images uploaded in an older WordPress version.)

modules/images/webp-uploads/load.php
eugene-manuilov: @felixarntz, yes, we have `file` and `filesize` properties with WordPress 6.0, I was on an older version. I'll revert this change.

> Side note, maybe we should raise the minimum required WordPress version to 6.0 soon - it would help this one, and also 6.1 is coming out soon, and our policy is to support at least the latest 2 versions.

Yes, I think we should definitely do it if we rely on a certain functionality from WP 6.0 in our codebase.

eugene-manuilov: Ok, reverted this change.

modules/images/webp-uploads/load.php
eugene-manuilov: Yes, we don't need this. Reverted.

modules/images/webp-uploads/helper.php
eugene-manuilov: Ok, yes, it does make sense. I have added one early return to avoid saving the same file to the disk:

https://github.com/WordPress/performance/pull/527/files#diff-1f653bd37f60a6476485a6794f7003a91b3713868c7414588d5835332fe68642R147-R153

Unfortunately, there is nothing else we can optimise in the code if only one mime type transformation is used.

felixarntz: @eugene-manuilov Makes sense, after further inspection of the code I agree. With that said, there is still some unexpected behavior here. Will follow up in separate comments.

modules/images/webp-uploads/load.php
felixarntz: I tested this locally, and I still think the extra condition here should be reverted. With the configuration of a single MIME type, `$current_mime` will be set to `image/webp`, as that is the MIME type of any existing image size WP core generates in that scenario. All this condition does is set the `sources` information for that existing WebP image. Checking whether it is in the list of allowed transform MIME types is unnecessary - mostly since that file already exists. If the file already exists, we don't need to check whether our filters allow for it. It's already generated, so we can just populate the information for it.

The MIME transforms filter is critical only to decide whether to transform an image of a certain format to a certain format, but for an image that already exists we can just add its information. In fact, we need to remove a similar condition around line 58 as well, since otherwise it runs into the problem that for the original JPEG no `sources` value is added, just because we don't _want_ to generate it.

felixarntz: @eugene-manuilov I fixed this in https://github.com/WordPress/performance/pull/527/commits/670b63f9ae5679af1f4ffd7556ad7fdd04477f75. There are currently unit test failures, but those are only happening because _WordPress_ still keeps the full JPEG image. Simply not putting it into `sources` should not be a fix for those tests, since that means the image is still there.

What I think we _want_ to do in the case that only WebP should be generated is actually override the original file with our generated "full" size WebP. We can use `_wp_image_meta_replace_original()` for that. And in that case, we can then also remove that original image from `sources`.

felixarntz: @eugene-manuilov I found the actual problem that was causing the test to fail, and fixed it in https://github.com/WordPress/performance/pull/527/commits/8825df789b16790dfa4e3d2fe88717fed2e23125 (see code comment on what was causing the problem).

The extra condition here is indeed not necessary and was only avoiding the failure rather than fixing the problem which was more in the test itself.

felixarntz:
@eugene-manuilov In https://github.com/WordPress/performance/pull/527/commits/9581c530188a076c1873d633ff15c24ec2b5d4f3, I implemented a fix to ensure that, when the MIME transforms are set to only generate WebP, the original/full size image is also replaced accordingly. See data below what that changes.

**Attachment metadata before**
```
array(7) {
  ["width"]=>
  int(1800)
  ["height"]=>
  int(1350)
  ["file"]=>
  string(15) "2022/10/bar.jpg"
  ["filesize"]=>
  int(326585)
  ["sizes"]=>
  array(5) {
    ["medium"]=>
    array(6) {
      ["file"]=>
      string(16) "bar-300x225.webp"
      ["width"]=>
      int(300)
      ["height"]=>
      int(225)
      ["mime-type"]=>
      string(10) "image/webp"
      ["filesize"]=>
      int(19644)
      ["sources"]=>
      array(1) {
        ["image/webp"]=>
        array(2) {
          ["file"]=>
          string(16) "bar-300x225.webp"
          ["filesize"]=>
          int(19644)
        }
      }
    }
    ["large"]=>
    array(6) {
      ["file"]=>
      string(17) "bar-1024x768.webp"
      ["width"]=>
      int(1024)
      ["height"]=>
      int(768)
      ["mime-type"]=>
      string(10) "image/webp"
      ["filesize"]=>
      int(102508)
      ["sources"]=>
      array(1) {
        ["image/webp"]=>
        array(2) {
          ["file"]=>
          string(17) "bar-1024x768.webp"
          ["filesize"]=>
          int(102508)
        }
      }
    }
    ["thumbnail"]=>
    array(6) {
      ["file"]=>
      string(16) "bar-150x150.webp"
      ["width"]=>
      int(150)
      ["height"]=>
      int(150)
      ["mime-type"]=>
      string(10) "image/webp"
      ["filesize"]=>
      int(7614)
      ["sources"]=>
      array(1) {
        ["image/webp"]=>
        array(2) {
          ["file"]=>
          string(16) "bar-150x150.webp"
          ["filesize"]=>
          int(7614)
        }
      }
    }
    ["medium_large"]=>
    array(6) {
      ["file"]=>
      string(16) "bar-768x576.webp"
      ["width"]=>
      int(768)
      ["height"]=>
      int(576)
      ["mime-type"]=>
      string(10) "image/webp"
      ["filesize"]=>
      int(68598)
      ["sources"]=>
      array(1) {
        ["image/webp"]=>
        array(2) {
          ["file"]=>
          string(16) "bar-768x576.webp"
          ["filesize"]=>
          int(68598)
        }
      }
    }
    ["1536x1536"]=>
    array(5) {
      ["file"]=>
      string(18) "bar-1536x1152.webp"
      ["width"]=>
      int(1536)
      ["height"]=>
      int(1152)
      ["mime-type"]=>
      string(10) "image/webp"
      ["filesize"]=>
      int(179844)
    }
  }
  ["image_meta"]=>
  array(12) {
    ["aperture"]=>
    string(1) "0"
    ["credit"]=>
    string(0) ""
    ["camera"]=>
    string(0) ""
    ["caption"]=>
    string(0) ""
    ["created_timestamp"]=>
    string(1) "0"
    ["copyright"]=>
    string(0) ""
    ["focal_length"]=>
    string(1) "0"
    ["iso"]=>
    string(1) "0"
    ["shutter_speed"]=>
    string(1) "0"
    ["title"]=>
    string(0) ""
    ["orientation"]=>
    string(1) "0"
    ["keywords"]=>
    array(0) {
    }
  }
  ["sources"]=>
  array(2) {
    ["image/jpeg"]=>
    array(2) {
      ["file"]=>
      string(7) "bar.jpg"
      ["filesize"]=>
      int(326585)
    }
    ["image/webp"]=>
    array(2) {
      ["file"]=>
      string(12) "bar-jpg.webp"
      ["filesize"]=>
      int(315124)
    }
  }
}
```

**Attachment metadata after**
```
array(8) {
  ["width"]=>
  int(1800)
  ["height"]=>
  int(1200)
  ["file"]=>
  string(20) "2022/10/pie-jpg.webp"
  ["filesize"]=>
  int(92266)
  ["sizes"]=>
  array(5) {
    ["medium"]=>
    array(6) {
      ["file"]=>
      string(16) "pie-300x200.webp"
      ["width"]=>
      int(300)
      ["height"]=>
      int(200)
      ["mime-type"]=>
      string(10) "image/webp"
      ["filesize"]=>
      int(4912)
      ["sources"]=>
      array(1) {
        ["image/webp"]=>
        array(2) {
          ["file"]=>
          string(16) "pie-300x200.webp"
          ["filesize"]=>
          int(4912)
        }
      }
    }
    ["large"]=>
    array(6) {
      ["file"]=>
      string(17) "pie-1024x683.webp"
      ["width"]=>
      int(1024)
      ["height"]=>
      int(683)
      ["mime-type"]=>
      string(10) "image/webp"
      ["filesize"]=>
      int(20432)
      ["sources"]=>
      array(1) {
        ["image/webp"]=>
        array(2) {
          ["file"]=>
          string(17) "pie-1024x683.webp"
          ["filesize"]=>
          int(20432)
        }
      }
    }
    ["thumbnail"]=>
    array(6) {
      ["file"]=>
      string(16) "pie-150x150.webp"
      ["width"]=>
      int(150)
      ["height"]=>
      int(150)
      ["mime-type"]=>
      string(10) "image/webp"
      ["filesize"]=>
      int(2256)
      ["sources"]=>
      array(1) {
        ["image/webp"]=>
        array(2) {
          ["file"]=>
          string(16) "pie-150x150.webp"
          ["filesize"]=>
          int(2256)
        }
      }
    }
    ["medium_large"]=>
    array(6) {
      ["file"]=>
      string(16) "pie-768x512.webp"
      ["width"]=>
      int(768)
      ["height"]=>
      int(512)
      ["mime-type"]=>
      string(10) "image/webp"
      ["filesize"]=>
      int(13368)
      ["sources"]=>
      array(1) {
        ["image/webp"]=>
        array(2) {
          ["file"]=>
          string(16) "pie-768x512.webp"
          ["filesize"]=>
          int(13368)
        }
      }
    }
    ["1536x1536"]=>
    array(5) {
      ["file"]=>
      string(18) "pie-1536x1024.webp"
      ["width"]=>
      int(1536)
      ["height"]=>
      int(1024)
      ["mime-type"]=>
      string(10) "image/webp"
      ["filesize"]=>
      int(40008)
    }
  }
  ["image_meta"]=>
  array(12) {
    ["aperture"]=>
    string(1) "0"
    ["credit"]=>
    string(0) ""
    ["camera"]=>
    string(0) ""
    ["caption"]=>
    string(0) ""
    ["created_timestamp"]=>
    string(1) "0"
    ["copyright"]=>
    string(0) ""
    ["focal_length"]=>
    string(1) "0"
    ["iso"]=>
    string(1) "0"
    ["shutter_speed"]=>
    string(1) "0"
    ["title"]=>
    string(0) ""
    ["orientation"]=>
    string(1) "0"
    ["keywords"]=>
    array(0) {
    }
  }
  ["sources"]=>
  array(1) {
    ["image/webp"]=>
    array(2) {
      ["file"]=>
      string(12) "pie-jpg.webp"
      ["filesize"]=>
      int(92266)
    }
  }
  ["original_image"]=>
  string(7) "pie.jpg"
}
```

The original image is now backed up, similarly to how core does it when scaling it down. This way we keep the JPEG available in case it's needed, but we fully use WebP, as we should when that is the only output MIME type from the `webp_uploads_get_upload_image_mime_transforms()` function.

felixarntz (approved):
@eugene-manuilov This is now good to go from my end. Since I made additional changes, please review and LMKWYT.

mehulkaklotar (approved):
Changes look good to me! 🚀 

We will need to add some conditions in this code for the checkbox integration for site settings to generate both mime types from this PR which is being worked on right now https://github.com/WordPress/performance/pull/537

jjgrainger (approved):

mukeshpanchal27 (approved):

PR #528, opened 2022-09-13
Implement new experimental `fetchpriority` module

### LINKED ISSUE #477: Fetchpriority - add a `fetchpriority=high` attribute to the non-lazy image(s)
## Overview
- Proposed module name: Fetchpriority 
- Proposed module slug: fetchpriority
- Proposed module owner GitHub username(s): pbearne, adamsilverstein

## About the module
Adding the [fetchpriority attribute](https://web.dev/priority-hints/#the-fetchpriority-attribute) `high` attribute to the LCP (or first large) [image](https://web.dev/priority-hints/#increase-the-priority-of-the-lcp-image) on the page should improve the page's LCP, because it tells the browser to prioritize loading that image.

We can build this as a module (possibly experimental), then measure its impact on performance. Note: as of this writing, `fetchpriority` is only supported in [chrome and edge](https://caniuse.com/?search=fetchpriority).
### Scope

WordPress already determines which image is most likely to be the "LCP" image in [`wp_get_loading_attr_default`](https://github.com/WordPress/WordPress/blob/b90c2adb7f65b610c84db80fa35b66cc036e1a66/wp-includes/media.php#L5495-L5516), excluding the `loading=lazy` attribute. We can leverage this to add `fetchpriority=high` to those same images, which should cause the browser to prioritize their load.
### Rationale

`fetchpriority` is a new approach to priority hinting that improves on features like `preload`. Adding this module will let the performance team test this feature as a potential improvement for core.
### Other

Read this document for the details: https://web.dev/priority-hints/#the-fetchpriority-attribute

pbearne:
## Summary

Fixes #477

## Checklist


modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,34 @@
+<?php
+/**
+ * Module Name: Fetchpriority
+ * Description: Adds Fetchpriority parameter to the none lazy loaded images.
```
adamsilverstein:
```suggestion
 * Description: Adds `fetchpriority=high` parameter to the non lazy loaded images.
```

adamsilverstein: is it images or image? typically only one image would be non lazy loaded, right?

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,34 @@
...
+ * Experimental: Yes
+ *
+ * @package performance-lab
+ * @since 1.5.0
+ */
+
+/**
+ * Filter image tags in content to add fetchpriority to the image tag if not lazy loaded set.
```
adamsilverstein:
```suggestion
 * Filter image tags in content to add fetchpriority to the image tag if lazy loaded is not set.
```

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,34 @@
...
+ * @since 1.5.0
+ *
+ * @param string $filtered_image The filtered image.
+ * @param string $context        The context of the image.
+ *
+ * @return string image tag
+ */
+function fetchpriority_img_tag_add( $filtered_image, $context ) {
+
+	// Only apply this in `the_content` for now, since otherwise it can result in duplicate runs due to a problem with full site editing logic.
```
adamsilverstein: interesting, any reference link for this issue?

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,34 @@
...
+	if ( 'the_content' !== $context ) {
+		return $filtered_image;
+	}
+
+	if ( strpos( $filtered_image, 'loading="lazy"' ) === false ) {
+		$filtered_image = str_replace( '<img ', '<img fetchpriority="high" ', $filtered_image );
```
adamsilverstein, line 32: I **think** we should only add this at most once (which maybe the case already), so once you add this you could `remove_filter( 'wp_content_img_tag', 'fetchpriority_img_tag_add', 30, 2 );`

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,34 @@
...
+	}
+
+	return $filtered_image;
+}
+add_filter( 'wp_content_img_tag', 'fetchpriority_img_tag_add', 30, 2 );
```
adamsilverstein: How did you choose 30?

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,34 @@
+<?php
+/**
+ * Module Name: Fetchpriority
+ * Description: Adds Fetchpriority parameter to the none lazy loaded images.
+ * Experimental: Yes
+ *
+ * @package performance-lab
+ * @since 1.5.0
```
mukeshpanchal27: Should use `n.e.x.t`.

```suggestion
 * @since n.e.x.t
 * @package performance-lab
```

modules/images/fetchpriority/load.php
pbearne: good call

modules/images/fetchpriority/load.php
pbearne: wanted it to be late 
no good reason 
let me try default 10

tests/modules/images/fetchpriority/Fetchpriority-Tests.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+class Fetchpriority_Tests extends WP_UnitTestCase {
+
+	public function test_wp_filter_content_tags_filter_adds_fetchpriority_fist_image_only() {
+		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg' );
+		$img           = get_image_tag( $attachment_id, '', '', '', 'large' );
+
+		$this->assertStringContainsString( 'fetchpriority="high"', fetchpriority_img_tag_add( $img, 'the_content' ) );
+		$this->assertStringNotContainsString( 'fetchpriority="high"', fetchpriority_img_tag_add( $img, 'not_content' ) );
```
adamsilverstein: Add an additional test here where the `img` tag contains `loading="lazy"` and the context is `the_content`, verifying the fetchpriority attribute is not added.

pbearne: added

modules/images/fetchpriority/load.php
pbearne: changed

modules/images/fetchpriority/load.php
pbearne: added

modules/images/fetchpriority/load.php
pbearne: no removed comment

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @return string image tag
+ */
+function fetchpriority_img_tag_add( $filtered_image, $context ) {
+
+	if ( 'the_content' !== $context ) {
+		return $filtered_image;
+	}
+
+	if ( strpos( $filtered_image, 'loading="lazy"' ) === false ) {
```
felixarntz: We should make sure to have a space at the beginning of this string, to make sure it doesn't match something like `fooloading="lazy"`. Also, while that's unlikely to already be present, we must avoid adding the `fetchpriority` attribute if the image tag already has one on it, so let's include that here as well.
```suggestion
	if ( strpos( $filtered_image, ' loading="lazy"' ) === false && strpos( $filtered_image, ' fetchpriority=' ) === false ) {
```

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Filter image tags in content to add fetchpriority to the image tag if lazy loaded is not set.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $filtered_image The filtered image.
+ * @param string $context        The context of the image.
+ *
+ * @return string image tag
```
felixarntz:
```suggestion
 * @param string $filtered_image The image tag to filter.
 * @param string $context        The context of the image.
 * @return string The filtered image tag.
```

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,34 @@
...
+ */
+function fetchpriority_img_tag_add( $filtered_image, $context ) {
```
felixarntz: Nit-pick, but better to align with core naming:
```suggestion
function fetchpriority_img_tag_add_attr( $filtered_image, $context ) {
```

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,34 @@
...
+
+	if ( 'the_content' !== $context ) {
+		return $filtered_image;
+	}
+
+	if ( strpos( $filtered_image, 'loading="lazy"' ) === false ) {
+		$filtered_image = str_replace( '<img ', '<img fetchpriority="high" ', $filtered_image );
+		remove_filter( 'wp_content_img_tag', 'fetchpriority_img_tag_add', 10, 2 );
```
felixarntz: See above. Also:
* `remove_filter()` only supports three attributes.
* We need to remove the other filter for featured image which I suggested here too so that it's really always only on the first image it encounters.

```suggestion
		remove_filter( 'wp_content_img_tag', 'fetchpriority_img_tag_add_attr', 10 );
		remove_filter( 'post_thumbnail_html', 'fetchpriority_filter_post_thumbnail_html' );
```

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,34 @@
...
+	}
+
+	return $filtered_image;
+}
+add_filter( 'wp_content_img_tag', 'fetchpriority_img_tag_add', 10, 2 );
```
felixarntz:
```suggestion
add_filter( 'wp_content_img_tag', 'fetchpriority_img_tag_add_attr', 10, 2 );
```

tests/modules/images/fetchpriority/Fetchpriority-Tests.php
```diff
@@ -0,0 +1,15 @@
...
+
+		$img = str_replace( '<img ', '<img loading="lazy" ', $img );
+		$this->assertStringNotContainsString( 'fetchpriority="high"', fetchpriority_img_tag_add( $img, 'the_content' ) );
+	}
```
felixarntz: See above regarding the function name.

Note also that the test name here is misleading, as this function does not test `wp_filter_content_tags()` at all. It only tests the new function added here, not how it affects `wp_filter_content_tags()`. It also doesn't test that it only adds the attribute to the first image only, so that also needs to be changed in the test name.

Also see my comment above on that we need to add a space before the attribute name when checking for it.

```suggestion
	public function test_fetchpriority_img_tag_add_attr_based_on_context_and_loading_lazy() {
		$attachment_id = $this->factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg' );
		$img           = get_image_tag( $attachment_id, '', '', '', 'large' );

		$this->assertStringContainsString( ' fetchpriority="high"', fetchpriority_img_tag_add_attr( $img, 'the_content' ) );
		$this->assertStringNotContainsString( ' fetchpriority="high"', fetchpriority_img_tag_add_attr( $img, 'not_content' ) );

		$img = str_replace( '<img ', '<img loading="lazy" ', $img );
		$this->assertStringNotContainsString( ' fetchpriority="high"', fetchpriority_img_tag_add_attr( $img, 'the_content' ) );
	}
```

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,34 @@
...
+ * Module Name: Fetchpriority
+ * Description: Adds `fetchpriority=high` parameter to the non lazy loaded images.
+ * Experimental: Yes
+ *
+ * @since n.e.x.t
+ * @package performance-lab
+ */
+
+/**
+ * Filter image tags in content to add fetchpriority to the image tag if lazy loaded is not set.
```
felixarntz:
```suggestion
 * Filters an image tag in content to add the fetchpriority attribute if it is not lazy-loaded.
```

modules/images/fetchpriority/load.php
felixarntz: One other piece you need to add here is support for the featured image output. You can hook into the `post_thumbnail_html` filter a new callback that just wraps the other callback above. Then you can use the `the_post_thumbnail` context which is already used in WordPress core.

Please add a test for this function as well.
```suggestion
add_filter( 'wp_content_img_tag', 'fetchpriority_img_tag_add', 10, 2 );

/**
 * Filters the post thumbnail HTML to conditionally add the fetchpriority attribute.
 *
 * @since n.e.x.t
 *
 * @param string $html The post thumbnail HTML to filter.
 * @return string The filtered post thumbnail HTML.
 */
function fetchpriority_filter_post_thumbnail_html( $html ) {
	return fetchpriority_img_tag_add_attr( $html, 'the_post_thumbnail' );
}
add_filter( 'post_thumbnail_html', 'fetchpriority_filter_post_thumbnail_html' );
```

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,34 @@
...
+ * @since n.e.x.t
+ *
+ * @param string $filtered_image The filtered image.
+ * @param string $context        The context of the image.
+ *
+ * @return string image tag
+ */
+function fetchpriority_img_tag_add( $filtered_image, $context ) {
+
+	if ( 'the_content' !== $context ) {
```
felixarntz: Based on my other comment and code suggestion around the featured image:
```suggestion
	if ( 'the_content' !== $context && 'the_post_thumbnail' !== $context ) {
```

tests/modules/images/fetchpriority/Fetchpriority-Tests.php
```diff
@@ -0,0 +1,15 @@
...
+}
```
felixarntz: We should add another test here that actually calls `wp_filter_content_tags()` to test the integration, since your test above only tests direct calls to the newly added function. For the new test, let's pass some demo content with at least two images in it. The lazy-loading shouldn't matter here. We just need to ensure that the first image has the `fetchpriority` attribute while the second one shouldn't, to test that the removal of the filter works.

Also see my comment above around adding one test for the featured image integration.

felixarntz (changes requested):
@pbearne This is a solid start. I left a few mostly small comments below, but there are two bigger things among them which are critical but missing here:
1. We need to add support for the featured image output, as it is covered by the conditional lazy-loading logic as well.
2. We should add at least one more test to cover the actual `wp_filter_content_tags()` integration of the code here.

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,47 @@
...
+ * @return string The filtered image tag.
+ */
+function fetchpriority_img_tag_add_attr( $filtered_image, $context ) {
+
+	if ( 'the_content' !== $context && 'the_post_thumbnail' !== $context ) {
+		return $filtered_image;
+	}
+
+	if ( strpos( $filtered_image, ' loading="lazy"' ) === false && strpos( $filtered_image, ' fetchpriority=' ) === false ) {
+		$filtered_image = str_replace( '<img ', '<img fetchpriority="high" ', $filtered_image );
```
mukeshpanchal27, line 32: If my post contains text and at the end of the article I use an image, then do we also need to add `fetchpriority="high"` attribute? The image is not in view port.

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Filters an image tag in content to add the fetchpriority attribute if it is not lazy-loaded.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $filtered_image The image tag to filter.
+ * @param string $context        The context of the image.
+ * @return string The filtered image tag.
+ */
+function fetchpriority_img_tag_add_attr( $filtered_image, $context ) {
```
mukeshpanchal27, line 20: Do we need to add a filter to disable this functionality? Like with Lazy Load, we have `add_filter( 'wp_lazy_loading_enabled', '__return_false' );` filter in the core.

mukeshpanchal27 (commented):
@pbearne great work. It look good to me. Left some questions.

mukeshpanchal27:
https://github.com/WordPress/performance/pull/528/commits/5480988b0e66b5d07cd93944f32fe95c471c8117 Use lowercase file name for test class file commit reverted in https://github.com/WordPress/performance/pull/528/commits/e6826eb1f164ede4582d793d709d2e925540e8d4 that needs to be correct.

Do we need `/tests/modules/images/fetchpriority` in CODEOWNERS?

modules/images/fetchpriority/load.php
pbearne: I am letting the loading="lazy" code work if to add the as don't what to have both this is an else, and I suspect that when add it to code, it will be mixed into the code that adds loading="lazy"

modules/images/fetchpriority/load.php
pbearne: not here, but in core yes.
we can just turn it off in the plugin

mukeshpanchal27:
@bethanylang The PR needs a `Type` label. PR failing now because of it.

mxbclang:
@mukeshpanchal27 Ah, that's right! Fixed.  :)

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,49 @@
...
+ * @param string $html The post thumbnail HTML to filter.
+ * @return string The filtered post thumbnail HTML.
+ */
+function fetchpriority_filter_post_thumbnail_html( $html ) {
+	$html = fetchpriority_img_tag_add_attr( $html, 'the_post_thumbnail' );
+	remove_filter( 'post_thumbnail_html', 'fetchpriority_filter_post_thumbnail_html' );
+
+	return $html;
+}
+add_filter( 'post_thumbnail_html', 'fetchpriority_filter_post_thumbnail_html' );
```
adamsilverstein, line 54: Nice!

tests/modules/images/fetchpriority/fetchpriority-tests.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
```
mukeshpanchal27: Remove space

tests/modules/images/fetchpriority/fetchpriority-tests.php
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Tests the adding of fetchpriority to img tags in the content.
+ *
+ * @Covers ::wp_filter_content_tags
```
mukeshpanchal27: Test is for `fetchpriority_img_tag_add_attr` function isn't it? And should use lowercase for `@covers` annotation.

```suggestion
 * @covers ::fetchpriority_img_tag_add_attr
```

tests/modules/images/fetchpriority/fetchpriority-tests.php
```diff
@@ -0,0 +1,37 @@
...
+		self::$attachment_id = $factory->attachment->create_upload_object(
+			$file,
+			self::$post->ID,
+			array(
+				'post_mime_type' => 'image/jpeg',
+			)
+		);
+	}
+	public function test_fetchpriority_img_tag_add_attr_based_on_context_and_loading_lazy() {
+		$img = get_image_tag( self::$attachment_id, '', '', '', 'large' );
```
mukeshpanchal27, line 41:
```suggestion
		$attachment_id = $this->factory->attachment->create_upload_object(
			TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg'
		);

		$img = get_image_tag( $attachment_id, '', '', '', 'large' );
```

tests/modules/images/fetchpriority/fetchpriority-tests.php
```diff
@@ -0,0 +1,37 @@
...
+
+		$file                = TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg';
+		self::$attachment_id = $factory->attachment->create_upload_object(
+			$file,
+			self::$post->ID,
+			array(
+				'post_mime_type' => 'image/jpeg',
+			)
+		);
+	}
```
mukeshpanchal27: Should remove this as we get `$attachment_id` from other code. Check other comment.

mukeshpanchal27 (commented):
@pbearne Left some review for the unit test.

mxbclang:
@pbearne Just a reminder on changes requested here, thank you!

tests/modules/images/fetchpriority/fetchpriority-tests.php
pbearne: code removed

pbearne:
I need help getting the tests to work.
I expect the first image to not have to lazyload set, but I am getting this so tests are failing

gilbertococchi:
Thanks Paul for all the efforts into building this module.

Were you able to identify in what use case the attribute loading=lazy is added on the first image?

I also was able to reproduce that and got some feedback from WP developers that this happens but I wasn't able to find the root cause in WP Core just yet.

I can see how this might undermine the potential impact of adding the fetchpriority if there are critical cases where it would not be added.

pbearne:
Look at the core code It should miss the first image loaded on a page it uses this function track the count wp_increase_content_media_count()
You can see an example in wp_get_attachment_image()

It may be working on a site but not in the tests

gilbertococchi:
Thanks for your answer @pbearne, I can totally get why the test would fail in the tests, @felixarntz FYI.

I will investigate on other threads why WP Sites apply so often the `loading=lazy` value on first images, that was one of my concerns to use the same logic to decide when to apply the `fetchpriority` attribute as well.

adamsilverstein:
@pbearne I'm going to see if I can tell why the tests are failing here. Were you able to resolve all the other previous feedback (I see some unresolved comments in the thread)?

modules/images/fetchpriority/load.php
adamsilverstein: > If my post contains text and at the end of the article I use an image, then do we also need to add fetchpriority="high" attribute? The image is not in view port.

The short answer is no. However, we are expecting you will benefit from marking your first image - based on the expectation that on balance, it will be your LCP element.

tests/modules/images/fetchpriority/fetchpriority-tests.php
adamsilverstein: removed in incoming code

tests/modules/images/fetchpriority/Fetchpriority-Tests.php
adamsilverstein: this is already updated in the code base

modules/images/fetchpriority/load.php
adamsilverstein: this is in place in the codebase

tests/modules/images/fetchpriority/Fetchpriority-Tests.php
adamsilverstein: this is in place as `test_fetchpriority_img_tag_add_in_wp_filter_content_tags`

pbearne:
@adamsilverstein I have lost track of where we all are. If you can sort the tests/code, I will revisit and resolve the tickets 
I suspect that the issue is with the lazyloading code, not this code as we are piggybacking on that

adamsilverstein:
@pbearne I tracked down the issue(s) with the tests and got them all passing:
- the filter in core only does the counting & limiting lazy from first image for images that are "in the loop" and where `is_main_query` is true (a restriction that is perhaps too limiting?) so I had to mock that condition
- the duplicate image id caused an issue because `wp_filter_content_tags` has an optimization that prevents it from repeating logic on identical images.Using two separate image Ids fixed the issue.
I also added some cleanup logic for the attachments.

Please review and we can try to land this soon.

adamsilverstein:
I verified this works as expected in a classic theme like twentytwelve.

when testing with twentytwentythree however, I noticed that the first image is lazy loaded, so something isn't quite working with the logic in core for block themes. I'm going to dig into this a bit further to see why. cc: @felixarntz

adamsilverstein (approved):
LGTM

adamsilverstein:
> when testing with twentytwentythree however, I noticed that the first image is lazy loaded, so something isn't quite working with the logic in core for block themes. I'm going to dig into this a bit further to see why. 

As we had discussed earlier, this is caused by the block theme "double" pass at the content (which can help performance). To fix this we need to avoid "counting" the first pass by block themes. I noticed in this case the `$context` is false, so arrived at this simple solution which fixes the issue in my testing: https://github.com/WordPress/wordpress-develop/pull/3538

I will open up a Trac ticket to discuss further.

adamsilverstein:
@pbearne this is ready for review.

adamsilverstein:
@felixarntz this is ready for review

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,49 @@
...
+ * @param string $context        The context of the image.
+ * @return string The filtered image tag.
+ */
+function fetchpriority_img_tag_add_attr( $filtered_image, $context ) {
+
+	if ( 'the_content' !== $context && 'the_post_thumbnail' !== $context ) {
+		return $filtered_image;
+	}
+
+	if ( ! empty( $filtered_image ) && strpos( $filtered_image, 'loading="lazy"' ) === false && strpos( $filtered_image, 'fetchpriority=' ) === false ) {
```
felixarntz, line 31: Since we rely on the presence of `loading="lazy"` here, we first of all have to make sure the site has lazy-loading enabled in general (via `wp_lazy_loading_enabled( 'img', $context )`). Otherwise, this will cause a bug on sites that disabled it, effectively giving every image `fetchpriority="high"` which would be detrimental. Let's add a test to cover that scenario as well.

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,49 @@
...
+		$filtered_image = str_replace( '<img ', '<img fetchpriority="high" ', $filtered_image );
+		remove_filter( 'wp_content_img_tag', 'fetchpriority_img_tag_add_attr' );
```
felixarntz, line 33: I think for safety reasons we need to rather include the other filter removal call here as well. Essentially, whenever we add `fetchpriority="high"` to any image, we need to remove all other filters that could possibly do it, so that it's always just one image.

Right now, the way the code is written it would be possible in principle for 2 images to have `fetchpriority="high"` when a featured image without `loading` attribute follows a content image without `loading` attribute. Unlikely but possible, and we need to be stricter here than with lazy-loading as preferably really only the LCP image should have `fetchpriority="high"`.
```suggestion
		remove_filter( 'wp_content_img_tag', 'fetchpriority_img_tag_add_attr' );
		remove_filter( 'post_thumbnail_html', 'fetchpriority_filter_post_thumbnail_html' );
```

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,49 @@
...
+ * Filters the post thumbnail HTML to conditionally add the fetchpriority attribute.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $html The post thumbnail HTML to filter.
+ * @return string The filtered post thumbnail HTML.
+ */
+function fetchpriority_filter_post_thumbnail_html( $html ) {
+	$html = fetchpriority_img_tag_add_attr( $html, 'the_post_thumbnail' );
+	remove_filter( 'post_thumbnail_html', 'fetchpriority_filter_post_thumbnail_html' );
```
felixarntz: See above, we should then remove the call here in favor of the one above.
```suggestion
```

felixarntz (changes requested):
@pbearne @adamsilverstein Overall this looks solid, though there are a few things we need to fix as they can introduce bugs.

In light of the currently ongoing work to fix image lazy-loading in block themes (see https://core.trac.wordpress.org/ticket/56930), we should make sure to test prior to merge that this works as expected in a block theme.

Looking at the code here, things look like this may not be affected by the double execution bug from core (due to the more explicit context check present here), but we need to verify that. Adding `fetchpriority="high"` to too many images may be even worse than lazy-loading too many images.

tests/modules/images/fetchpriority/fetchpriority-tests.php
```diff
@@ -0,0 +1,77 @@
...
+		// Ensure image filtering occurs 'in_the_loop', is_main_query.
+		$wp_the_query          = $wp_query;
+		$wp_query->in_the_loop = true;
+		$content               = wp_filter_content_tags( $img, 'the_content' );
+		$this->assertStringContainsString( 'fetchpriority="high"', $content );
+		$this->assertStringContainsString( 'loading="lazy"', $content );
+		$this->assertTrue( strpos( $content, 'fetchpriority="high"' ) < strpos( $content, 'loading="lazy"' ) );
+
+		$spilt = explode( 'fetchpriority', $content );
+		$this->assertEquals( 2, count( $spilt ) );
```
felixarntz: What is the purpose of this? To check how often `fetchpriority` is included in the content? In that case we should rather use `substr_count`. This logic looks confusing :)

adamsilverstein: i was confused at first as well, this is just counting that fetchpriority is included exactly once, agree changing this to a substr_count makes more sense 👍🏼

modules/images/fetchpriority/load.php
adamsilverstein: 👍🏼 added a check above @ L26

tests/modules/images/fetchpriority/fetchpriority-tests.php
adamsilverstein: Adjusted

adamsilverstein:
@felixarntz Addressed feedback and ready for another review

adamsilverstein:
> we should make sure to test prior to merge that this works as expected in a block theme.

I will double check this.

felixarntz:
Milestoning this for the 1.8.0 release coming in December, as it's shaping up for merge by then. I'll give it another review later this week.

modules/images/fetchpriority/load.php
```diff
@@ -0,0 +1,54 @@
+<?php
+/**
+ * Module Name: Fetchpriority
+ * Description: Adds `fetchpriority=high` parameter to the non lazy loaded images.
```
felixarntz: This description is a bit too technical and too close to the detailed implementation. Also, we can't use Markdown since this will be displayed as is in HTML. I suggest something like:
```suggestion
 * Description: Adds a fetchpriority hint for the primary content image on the page to load faster.
```

felixarntz (approved):
@pbearne @adamsilverstein The implementation here LGTM, with the following caveats / known issues:
* For block themes, it currently doesn't work on featured images since the featured image block is broken in regards to `loading="lazy"`.
    * This was recently fixed in Gutenberg (see https://github.com/WordPress/gutenberg/pull/45534).
* For block themes, it already works for regular content images, however the first content image will still have `loading="lazy"` on it due to the double content parsing bug in core.
    * A WordPress core fix is being worked on (see https://github.com/WordPress/wordpress-develop/pull/3560).

Both of these problems are not critical, since at least this PR doesn't make it worse than it is; and both of these problems need to be fixed separately, so this is good to merge IMO.

Will update the module description though, as it's a bit too technical and doesn't describe the performance impact.

felixarntz:
FYI the PR https://github.com/whatwg/html/pull/8470 is where adding `fetchpriority` to the HTML spec is being worked on. This will be critical before we can eventually merge it into WordPress core. But for the plugin shipping this already should be fine.

mukeshpanchal27 (approved):
Thanks @pbearne @adamsilverstein LGTM

PR #537, opened 2022-09-19
Add checkbox to Settings > Media to control whether to generate JPEG in addition to WebP

### LINKED ISSUE #525: Add checkbox to Media screen to enable multi MIME type output (WebP and JPEG)
#524 changes the default behavior of the WebP Uploads module so that it only generates WebP by default, following the core decision, as that is a safer default.

However, the module still offers the alternative approach of generating both WebP _and_ JPEG files. At a high level, this comes with the following tradeoff:
* **Pro:** Better backward/cross compatibility since any use-case where WebP is not supported can still leverage an equivalent JPEG.
* **Contra:** Large filesystem storage needs, since now every size requires two files, effectively almost doubling storage requirements.

While for a WordPress core implementation this decision would likely not be exposed as a user-facing control, for this plugin it makes more sense. It also would be applicable [if WebP became a "canonical plugin"](https://make.wordpress.org/core/2022/09/11/webp-in-core-for-6-1/).

Once #524 has been implemented, we can proceed with this one. It could work as follows:
* A new WordPress setting is registered to control whether to output both WebP and JPEG (e.g. `generate_webp_and_jpeg`), which could be a simple boolean value.
* If that setting is `true`, the default value of the `webp_uploads_upload_image_mime_transforms` filter should be changed so that it outputs both `image/jpeg` and `image/webp` for both `image/jpeg` and `image/webp` input format (effectively the same as it was prior to #524).
* A UI control (checkbox) for the setting could be added under "Settings > Media" in the "Uploads" section.
    * Label could be `Generate JPEG files in addition to WebP`
    * It would be good to add a brief description below that explains the tradeoffs, e.g. `Enabling JPEG in addition to WebP can improve compatibility, but will effectively double the filesystem storage use of your images.`

mehulkaklotar:
## Summary

This PR is dependent on the issue https://github.com/WordPress/performance/issues/524 and PR https://github.com/WordPress/performance/pull/527

PR introduces 

- WordPress setting `generate_webp_and_jpeg` -> boolean value
- Changes default filter output for `webp_uploads_upload_image_mime_transforms` to image/jpeg and image/webp if the `generate_webp_and_jpeg` is true

**Remaining work**(After https://github.com/WordPress/performance/issues/524 is merged)

- Add unit tests around new WP Settings and their use with filter
- Changes if needed any from the dependent PR https://github.com/WordPress/performance/pull/527

Fixes #525 

## Checklist


modules/images/webp-uploads/helper.php
```diff
@@ -20,6 +20,14 @@ function webp_uploads_get_upload_image_mime_transforms() {
 		'image/webp' => array( 'image/webp', 'image/jpeg' ),
 	);
 
+	// Check Setting for generating mimes.
+	if ( true === (bool) get_option( 'generate_webp_and_jpeg' ) ) {
```
mehulkaklotar: @mukeshpanchal27 aren't we checking for the default setting value here if the site hasn't saved settings yet? I guess it should be `1` as the default value.

modules/images/webp-uploads/helper.php
```diff
@@ -322,3 +331,24 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
...
+ *
+ * @since n.e.x.t
+ *
+ * @param WP_Image_Editor $editor    WP_Image_Editor instance.
+ * @param string          $file_name The path where the file would be stored, including the extension.
+ * @param string          $mime      The target mime in which the image should be created.
+ * @return WP_Image_Editor WP_Image_Editor instance with changes applied.
+ */
+function webp_uploads_check_multi_mime_support( $editor, $file_name, $mime ) {
+	if ( true === (bool) get_option( 'generate_webp_and_jpeg' ) ) {
```
mehulkaklotar: see the comment above for default setting value.

mehulkaklotar (commented):
@mukeshpanchal27 Changes looks good here. We should add some tests for the new setting.

modules/images/webp-uploads/helper.php
mehulkaklotar: Now that we think of it, it seems default value for the setting should be `0` -> false to avoid generating duplicate mime images.

modules/images/webp-uploads/helper.php
```diff
@@ -20,6 +20,14 @@ function webp_uploads_get_upload_image_mime_transforms() {
 		'image/webp' => array( 'image/webp' ),
 	);
 
+	// Check Setting for generating mimes.
+	if ( true === (bool) get_option( 'generate_webp_and_jpeg' ) ) {
```
felixarntz: We should always prefix options based on where they're coming from (core, which plugin, which theme). Let's prefix the setting/option name with `perflab_`.

modules/images/webp-uploads/helper.php
```diff
@@ -144,9 +152,7 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 		$destination_file_name = $editor->generate_filename( $suffix, null, $extension[0] );
 	}
 
-	remove_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
-	$image = $editor->save( $destination_file_name, $mime );
-	add_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
+	$image = webp_uploads_check_multi_mime_support( $editor, $destination_file_name, $mime );
```
felixarntz: This change shouldn't be needed. Our `image_editor_output_format` filter is only there for WordPress core to change the _default_ MIME type output format, which is only relevant when we generate _only_ WebP. The code here is for generating _additional_ MIME type output files, so for that the filter always needs to be temporarily removed, so that we don't try to map the additional MIME type output format as well.

Please revert this change.

modules/images/webp-uploads/helper.php
```diff
@@ -325,3 +331,24 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
...
+function webp_uploads_check_multi_mime_support( $editor, $file_name, $mime ) {
+	if ( true === (bool) get_option( 'generate_webp_and_jpeg' ) ) {
+		$image = $editor->save( $file_name, $mime );
+	} else {
+		remove_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
+		$image = $editor->save( $file_name, $mime );
+		add_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
+	}
+	return $image;
+}
```
felixarntz: See above, this isn't needed. Please revert.

modules/images/webp-uploads/image-edit.php
```diff
@@ -188,9 +188,7 @@ function ( $metadata, $post_meta_id ) use ( $post_id, $file_path, $mime_type, $e
 					$target_file_name     = preg_replace( "/\.$current_extension$/", ".$extension", $thumbnail_file );
 					$target_file_location = path_join( $original_directory, $target_file_name );
 
-					remove_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
-					$result = $editor->save( $target_file_location, $targeted_mime );
-					add_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
+					$result = webp_uploads_check_multi_mime_support( $editor, $target_file_location, $targeted_mime );
```
felixarntz: See above.

modules/images/webp-uploads/image-edit.php
```diff
@@ -200,9 +198,7 @@ function ( $metadata, $post_meta_id ) use ( $post_id, $file_path, $mime_type, $e
 				} else {
 					$destination = trailingslashit( $original_directory ) . "{$filename}.{$extension}";
 
-					remove_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
-					$result = $editor->save( $destination, $targeted_mime );
-					add_filter( 'image_editor_output_format', 'webp_uploads_filter_image_editor_output_format', 10, 3 );
+					$result = webp_uploads_check_multi_mime_support( $editor, $destination, $targeted_mime );
```
felixarntz: See above.

modules/images/webp-uploads/load.php
```diff
@@ -741,3 +741,43 @@ function webp_uploads_get_image_sizes_additional_mime_type_support() {
...
 }
+
+/**
+ * Register media settings field to generate multiple mime WebP & JPEG output.
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_register_media_settings_field() {
+	// Register field setting for the processing.
+	register_setting( 'media', 'generate_webp_and_jpeg' );
```
felixarntz: See above, please prefix with `perflab_` throughout this PR.

More importantly though, please use the full setting registration, also providing the necessary `$args`. Settings should be registered on `init`, not `admin_init` (the `add_settings_field()` call below should remain in `admin_init` though).

modules/images/webp-uploads/load.php
```diff
@@ -741,3 +741,43 @@ function webp_uploads_get_image_sizes_additional_mime_type_support() {
...
+		<td class="td-full">
+			<label for="generate_webp_and_jpeg">
+				<input name="generate_webp_and_jpeg" type="checkbox" id="generate_webp_and_jpeg" value="1"<?php checked( '1', get_option( 'generate_webp_and_jpeg' ) ); ?> />
+				<?php _e( 'Generate JPEG files in addition to WebP', 'performance-lab' ); ?>
+				<p class="description"><?php _e( 'Enabling JPEG in addition to WebP can improve compatibility, but will effectively double the filesystem storage use of your images.', 'performance-lab' ); ?></p>
+			</label>
+		</td>
+	</tr>
+	<?php
+}
```
felixarntz: I would advise that you move the new code in this file into a new `webp-uploads/settings.php` file which you then `require_once` here.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -894,4 +894,36 @@ private function mock_frontend_body_hooks() {
...
+
+		// The post MIME type should be JPEG.
+		$this->assertSame( 'image/jpeg', get_post_mime_type( $attachment_id ) );
+
+		// There should be JPEG and WebP sources for all sizes.
+		foreach ( array_keys( $metadata['sizes'] ) as $size_name ) {
+			$this->assertImageHasSizeSource( $attachment_id, $size_name, 'image/jpeg' );
+			$this->assertImageHasSizeSource( $attachment_id, $size_name, 'image/webp' );
+		}
+	}
```
felixarntz: Could you please move this further up in the file, next to the test methods at the top that cover similar logic?

modules/images/webp-uploads/helper.php
```diff
@@ -20,6 +20,14 @@ function webp_uploads_get_upload_image_mime_transforms() {
...
 	);
 
+	// Check Setting for generating mimes.
+	if ( true === (bool) get_option( 'generate_webp_and_jpeg' ) ) {
+		$default_transforms = array(
+			'image/jpeg' => array( 'image/jpeg', 'image/webp' ),
+			'image/webp' => array( 'image/webp', 'image/jpeg' ),
+		);
+	}
+
```
felixarntz, line 30: Please add a new test specifically for the `webp_uploads_get_upload_image_mime_transforms()` method, to cover this new condition.

modules/images/webp-uploads/load.php
```diff
@@ -741,3 +741,43 @@ function webp_uploads_get_image_sizes_additional_mime_type_support() {
...
+ * @since n.e.x.t
+ */
+function webp_uploads_generate_webp_jpeg_setting_callback() {
+	?>
+	<tr>
+		<td class="td-full">
+			<label for="generate_webp_and_jpeg">
+				<input name="generate_webp_and_jpeg" type="checkbox" id="generate_webp_and_jpeg" value="1"<?php checked( '1', get_option( 'generate_webp_and_jpeg' ) ); ?> />
+				<?php _e( 'Generate JPEG files in addition to WebP', 'performance-lab' ); ?>
+				<p class="description"><?php _e( 'Enabling JPEG in addition to WebP can improve compatibility, but will effectively double the filesystem storage use of your images.', 'performance-lab' ); ?></p>
```
felixarntz: Please use `esc_html_e()`.

Also, for better HTML semantics & accessibility:
* Please move the `p` tag outside, i.e. after the `label` tag.
* Please add an ID attribute to the `p` tag and then add an `aria-describedby` attribute to the `input` that references it.

felixarntz (changes requested):
@mehulkaklotar This is on a good path, some quirks to address though.

felixarntz (commented):
@mehulkaklotar One more thing: Please make sure to delete the new option in `uninstall.php`.

modules/images/webp-uploads/load.php
mehulkaklotar: I have now added these changes.

modules/images/webp-uploads/helper.php
mehulkaklotar: I have now added test to check the output of this function when `perflab_generate_webp_and_jpeg` option is set.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: Done.

modules/images/webp-uploads/load.php
mehulkaklotar: Done.

modules/images/webp-uploads/load.php
mehulkaklotar: Done.

modules/images/webp-uploads/helper.php
```diff
@@ -20,6 +20,14 @@ function webp_uploads_get_upload_image_mime_transforms() {
 		'image/webp' => array( 'image/webp' ),
 	);
 
+	// Check Setting for generating mimes.
```
felixarntz:
```suggestion
	// Check setting for whether to generate both JPEG and WebP.
```

modules/images/webp-uploads/settings.php
```diff
@@ -0,0 +1,55 @@
+<?php
+/**
+ * Settings for the webp uploads module.
```
felixarntz:
```suggestion
 * Settings for the WebP Uploads module.
```

modules/images/webp-uploads/settings.php
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+/**
+ * Register media settings field to generate multiple mime WebP & JPEG output.
```
felixarntz:
```suggestion
 * Registers setting for generating both JPEG and WebP versions for image uploads.
```

modules/images/webp-uploads/settings.php
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_register_media_settings_field() {
+	// Register field setting for the processing.
+	register_setting( 'media', 'perflab_generate_webp_and_jpeg' );
```
felixarntz: I think my comment in https://github.com/WordPress/performance/pull/537#discussion_r988060196 wasn't precise enough, I meant to "fully" register the setting, i.e. also including the `$args` parameter.
```suggestion
	register_setting(
		'media',
		'perflab_generate_webp_and_jpeg',
		array(
				'type'         => 'boolean',
				'default'      => false,
				'show_in_rest' => false,
		)
	);
```

modules/images/webp-uploads/settings.php
```diff
@@ -0,0 +1,55 @@
...
+}
+add_action( 'init', 'webp_uploads_register_media_settings_field' );
+
+/**
+ * Add media settings field to generate multiple mime WebP & JPEG output.
```
felixarntz: Maybe keep this more generic (also please use a third-person verb in the beginning of function/method descriptions):
```suggestion
 * Adds media settings field for the 'perflab_generate_webp_and_jpeg' setting.
```

modules/images/webp-uploads/settings.php
```diff
@@ -0,0 +1,55 @@
...
+		'webp_uploads_generate_webp_jpeg_setting_callback',
+		'media',
+		'uploads',
+		array( 'class' => 'hidden' )
+	);
+}
+add_action( 'admin_init', 'webp_uploads_add_media_settings_field' );
+
+/**
+ * Callback function for the media settings perflab_generate_webp_and_jpeg field.
```
felixarntz: See above.
```suggestion
 * Renders the settings field for the 'perflab_generate_webp_and_jpeg' setting.
```

modules/images/webp-uploads/settings.php
```diff
@@ -0,0 +1,55 @@
...
+	?>
+	<tr>
+		<td class="td-full">
+			<label for="perflab_generate_webp_and_jpeg">
+				<input name="perflab_generate_webp_and_jpeg" type="checkbox" id="perflab_generate_webp_and_jpeg" aria-describedby="perflab_generate_webp_and_jpeg_description" value="1"<?php checked( '1', get_option( 'perflab_generate_webp_and_jpeg' ) ); ?> />
+				<?php esc_html_e( 'Generate JPEG files in addition to WebP', 'performance-lab' ); ?>
+			</label>
+			<p class="description" id="perflab_generate_webp_and_jpeg_description"><?php esc_html_e( 'Enabling JPEG in addition to WebP can improve compatibility, but will effectively double the filesystem storage use of your images.', 'performance-lab' ); ?></p>
+		</td>
+	</tr>
```
felixarntz: Having the `<tr>` and `<td>` wrappers in here breaks the HTML. It does _look_ correct, but the resulting HTML is broken (see output below):
```html
<tr class="hidden"><th scope="row">Generate WebP and JPEG</th><td>	<tr>
    <td class="td-full">
        <label for="perflab_generate_webp_and_jpeg">
            <input name="perflab_generate_webp_and_jpeg" type="checkbox" id="perflab_generate_webp_and_jpeg" aria-describedby="perflab_generate_webp_and_jpeg_description" value="1" />
            Generate JPEG files in addition to WebP			</label>
        <p class="description" id="perflab_generate_webp_and_jpeg_description">Enabling JPEG in addition to WebP can improve compatibility, but will effectively double the filesystem storage use of your images.</p>
    </td>
</tr>
</td></tr>
```

I see what you're trying to do here, but we have to make sure to get valid HTML output. It's a bit challenging because WordPress's `do_settings_fields()` doesn't really give us the flexibility we need (please take a look at the code of that function to see how we can get correct HTML). Given the limitations, I would suggest the following:
* Remove the `<tr>` and `<td>` wrappers from the code here.
* Instead, add a `</td><td class="td-full">` at the beginning. This will ensure we close the `<td>` that WordPress core opens, and then open our own, which WordPress core then closes.
* In the `add_settings_field()` call above, change the `hidden` class to some other specific class name.
* We can then add our own `<style>` tag with the rules we need - essentially hide the `th` and `td` elements within, except the `td.td-full`.

I hope that makes sense, it's a bit confusing and unfortunate that WordPress makes that so hard.

felixarntz (commented):
@mehulkaklotar This is almost good to go now. Left mostly minor suggestions, so there are two quirks (one little tricky) to still address.

modules/images/webp-uploads/settings.php
mukeshpanchal27: The changes make sense.

modules/images/webp-uploads/settings.php
mukeshpanchal27: Strange behaviour 🤔 It's really hard to mange such things for any end user.

mukeshpanchal27: Changes updated in PR.

tests/modules/images/webp-uploads/helper-tests.php
```diff
@@ -420,6 +420,28 @@ function () {
 		$this->assertSame( array( 'image/jpeg' => array( 'image/jpeg', 'image/webp' ) ), $transforms );
 	}
 
+	/**
+	 * Returns custom transforms array when overwritten by webp_uploads_upload_image_mime_transforms filter.
```
mehulkaklotar:
```suggestion
	 * Returns JPG and WebP transforms array when perflab_generate_webp_and_jpeg option is true.
```

tests/modules/images/webp-uploads/helper-tests.php
```diff
@@ -420,6 +420,28 @@ function () {
...
 
+	/**
+	 * Returns JPG and WebP transforms array when perflab_generate_webp_and_jpeg option is true.
+	 *
+	 * @test
+	 */
+	public function it_should_return_jpeg_and_webp_transforms_when_option_generate_webp_and_jpeg_set() {
+		remove_all_filters( 'webp_uploads_get_upload_image_mime_transforms' );
+
+		update_option( 'perflab_generate_webp_and_jpeg', '1' );
```
mehulkaklotar:
```suggestion
		update_option( 'perflab_generate_webp_and_jpeg', true );
```

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -108,6 +108,37 @@ public function it_should_create_jpeg_and_webp_for_jpeg_images_if_opted_in() {
 		}
 	}
 
+	/**
+	 * Create JPEG and WebP for JPEG images, if perflab_generate_webp_and_jpeg option set.
+	 *
+	 * @test
+	 */
+	public function it_should_create_jpeg_and_webp_for_jpeg_images_if_generate_webp_and_jpeg_set() {
+		update_option( 'perflab_generate_webp_and_jpeg', '1' );
```
mehulkaklotar:
```suggestion
		update_option( 'perflab_generate_webp_and_jpeg', true );
```

modules/images/webp-uploads/settings.php
felixarntz: Thanks @mukeshpanchal27, this looks good now.

modules/images/webp-uploads/settings.php
```diff
@@ -0,0 +1,83 @@
...
+		<style>
+		.form-table .perflab-generate-webp-and-jpeg th,
+		.form-table .perflab-generate-webp-and-jpeg td:not(.td-full) {
+			display: none;
+		}
+		</style>
+		<?php
+	}
+}
+add_action( 'admin_head', 'webp_uploads_media_setting_style' );
```
felixarntz: We should use the more targeted `admin_head-{$hook_suffix}` action for that. This way we ensure it's only hooked in the right place and doesn't need to run a check on every admin screen.
```suggestion
add_action( 'admin_head-options-media.php', 'webp_uploads_media_setting_style' );
```

modules/images/webp-uploads/settings.php
```diff
@@ -0,0 +1,83 @@
...
+	if ( 'options-media.php' === $pagenow ) {
+		?>
+		<style>
+		.form-table .perflab-generate-webp-and-jpeg th,
+		.form-table .perflab-generate-webp-and-jpeg td:not(.td-full) {
+			display: none;
+		}
+		</style>
+		<?php
+	}
```
felixarntz: See my other comment, with that change you can remove the extra global and condition here and simply render the `style` element.

Nit-pick: Maybe indent the contents of the `<style>` tag by one tab. :)

felixarntz (commented):
@mukeshpanchal27 @mehulkaklotar One small optimization recommendation here, but other than that this is good to go.

modules/images/webp-uploads/settings.php
mukeshpanchal27: Thanks Felix for the review. Changes make sense to me.

modules/images/webp-uploads/settings.php
mukeshpanchal27: Yes. Thanks for review.

mukeshpanchal27 (approved):

felixarntz (approved):

PR #543, opened 2022-09-20
Only load Site Health checks for persistent cache and full page cache when not available in core

### LINKED ISSUE #540: Add `can-load.php` files to load Site Health checks for persistent cache and full page cache conditionally
Since the two modules for the Site Health checks for persistent cache and full page cache got merged into core, we need to add logic so that these two modules are only loaded going forward when the relevant core functionality is not already present. For that we should use the `can-load.php` approach which was established in #293.

Relevant WordPress core commits:
* https://core.trac.wordpress.org/changeset/53955
* https://core.trac.wordpress.org/changeset/54043

Since the class that has all the changes is only loaded within WP Admin, we may not be able to check for actual functions or methods reliably, as our plugin initialization would run too early. So potentially we will simply need to check that the WordPress version is lower than 6.1.

mukeshpanchal27:
## Summary

Fixes #540

## Checklist

### DIFF
--- a/modules/object-cache/persistent-object-cache-health-check/can-load.php
+++ b/modules/object-cache/persistent-object-cache-health-check/can-load.php
@@ -1,0 +1,20 @@
+<?php
+/**
+ * Can load function to determine if Persistent Object Cache Health Check module is already merged in WordPress core.
+ *
+ * @since n.e.x.t
+ * @package performance-lab
+ */
+
+/**
+ * Checks whether the given module is already merged into the WordPress core.
+ *
+ * @since n.e.x.t
+ *
+ * @global string $wp_version The WordPress version string.
+ */
+return function() {
+	global $wp_version;
+
+	return ! version_compare( $wp_version, '6.1.0', '>=' );
+};
--- a/modules/object-cache/persistent-object-cache-health-check/can-load.php
+++ b/modules/object-cache/persistent-object-cache-health-check/can-load.php
@@ -16,5 +16,5 @@
 return function() {
 	global $wp_version;
 
-	return ! version_compare( $wp_version, '6.1.0', '>=' );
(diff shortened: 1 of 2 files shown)


modules/object-cache/persistent-object-cache-health-check/can-load.php
```diff
@@ -0,0 +1,20 @@
...
+ * Checks whether the given module is already merged into the WordPress core.
+ *
+ * @since n.e.x.t
+ *
+ * @global string $wp_version The WordPress version string.
+ */
+return function() {
+	global $wp_version;
+
+	return ! version_compare( $wp_version, '6.1.0', '>=' );
```
felixarntz: Nit-pick, but we could simplify this to
```suggestion
	return version_compare( $wp_version, '6.1.0', '<' );
```

Makes it a bit easier to understand ("is the WordPress version lower than 6.1.0?").

modules/site-health/audit-full-page-cache/can-load.php
```diff
@@ -0,0 +1,20 @@
...
+ * Checks whether the given module is already merged into the WordPress core.
+ *
+ * @since n.e.x.t
+ *
+ * @global string $wp_version The WordPress version string.
+ */
+return function() {
+	global $wp_version;
+
+	return ! version_compare( $wp_version, '6.1.0', '>=' );
```
felixarntz: See above.
```suggestion
	return version_compare( $wp_version, '6.1.0', '<' );
```

felixarntz (approved):
@mukeshpanchal27 Looks great, just 2 nit-picks you may want to address.

mehulkaklotar (approved):

mukeshpanchal27:
Assign it to @felixarntz. It is now ready to merge.

PR #544, opened 2022-09-23
Run tests in different versions of PHP

### LINKED ISSUE #399: Run PHP unit tests in different versions of PHP
## Feature Description

Run PHP unit tests in different versions of PHP in CI.

mukeshpanchal27:
## Summary

Fixes #399

## Checklist


.github/workflows/php-test.yml
```diff
@@ -39,9 +39,15 @@ on:
...
 jobs:
   php-test:
-    name: PHP
+    name: "PHP ${{ matrix.php }}"
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        php: ['7.4','8.0','8.1']
```
swissspidy: Ideally we also test PHP 7.0, 8.2, and 8.3 (as experimental); not sure if wp-env supports all of these

swissspidy:
I thought I'd pick this one up, but a couple of SVG upload tests (`test_get_dominant_color_none_images`) are failing on various PHP versions 🤷

mukeshpanchal27:
Thanks @swissspidy for picking. I also don't understood why it going fails 🤔

swissspidy:
So the failures makes sense, because WP doesn't handle/support SVG uploads, so `wp_check_filetype` fails. And unfiltered uploads are not enabled either. The weird part is that the test is passing on PHP 7.4...

We don't really need to test SVG upload here anyway, so we can just remove it.

swissspidy:
Tests on PHP < 7.3 are failing because the PHPUnit version (v9.6) is too new for those PHP versions, causing syntax errors.

We probably don't need something as complex as https://github.com/WordPress/plugin-check/pull/321, but definitely something in that direction.

swissspidy:
Just updated wp-env and force a `composer update` before running tests

Good news: PHPUnit is now running on every PHP version \*
Bad news: tests are failing because webp doesn't seem to be supported on these versions? 🤔 

\* Edit: need to move PHPStan to `build-cs` here to fix PHP 7.0 & 7.1

swissspidy:
Okay so PHPUnit install is fixed now.

The last remaining issue seems to be that `wp_image_editor_supports( array( 'mime_type' => 'image/webp' ) )` in `webp_uploads_generate_additional_image_source` now suddenly seems to return `false`  on wp-env 🤔  That causes WebP uploads to fail in tests with `image_mime_type_not_supported` errors.

On the actual wp-env website the WebP uploads work just fine though, so no idea what's going on.

Unfortunately I am unable to get my Docker setup working right now so can't reproduce locally.

swissspidy:
I really have no idea what's going on with these Imagick and WebP tests 🤷 😩

swissspidy:
@felixarntz @adamsilverstein @spacedmonkey Has anyone got an idea why these dominant color and webp tests are suddenly failing with this PR? 🤷

felixarntz:
@swissspidy Not sure that's what's causing the failures, but there's that hard-coded PHP 7.4 declaration in https://github.com/WordPress/performance/blob/trunk/.wp-env.json#L9. I wonder whether because of that the test suite is somehow on a different PHP version than core and therefore these problems are caused?

Just a hunch worth checking. I don't even know why this hard-coded version is there, so maybe we should try removing it and see what happens? 🤔

swissspidy:
Doesn't seem to make a difference.

At least locally, I get the `it_should_validate_source_attribute_update_when_webp_edited` test to fail as well, even without wp-env.

As for dominant color, it's only the Imagick tests that fail. Maybe that's not properly available or so? 🤔

mukeshpanchal27:
In https://github.com/WordPress/performance/pull/544/commits/baf7ce31678f73115b6ea4d8f00c7b42f8340c7b remove filter that disturbed other tests. now test suite show two error in PHP > 7.4 but for lower version it show error for more tests 🤔

mukeshpanchal27:
@swissspidy Unit tests are now pass ✅ 7.4 and above PHP version. For lower PHP version we have to check what was missing.

swissspidy:
On PHP < 7.4 it seems that there is no built-in WebP support or so.

mukeshpanchal27:
> On PHP < 7.4 it seems that there is no built-in WebP support or so.

Yes. In https://github.com/WordPress/performance/pull/544/commits/e2872b8779347e862b97bc551d070f7f3cb78bbc I fid Unit test issue for PHP 7.3

mukeshpanchal27:
We also need to update the command here: https://github.com/WordPress/performance/blob/fix/399-unit-tests-in-php-versions/bin/plugin/commands/test-plugins.js#L323-L343 so that the `test-plugins` script works as expected.

swissspidy:
> We also need to update the command here: [`fix`/399-unit-tests-in-php-versions/bin/plugin/commands/test-plugins.js#L323-L343](https://github.com/WordPress/performance/blob/fix/399-unit-tests-in-php-versions/bin/plugin/commands/test-plugins.js?rgh-link-date=2024-01-24T12%3A30%3A36Z#L323-L343) so that the `test-plugins` script works as expected.

First we need to update `.github/workflows/php-test-standalone-plugins.yml` to use a matrix.

I'll do that now

swissspidy:
Unit tests pass now 🎉 

Not sure how the integration tests are setup 🤔  @mukeshpanchal27 perhaps you can take a look at fixing that setup?

mukeshpanchal27:
All the unit test pass ✅ now. Looking into integration tests setup.

swissspidy:
There is no error output for failing runs, see e.g. https://github.com/WordPress/performance/actions/runs/7650368243/job/20846313421?pr=544

Do we need to log `command.stderr`?

mukeshpanchal27:
@swissspidy Added log in script. In local i got below error for PHP 7.2.

```
ℹ Starting 'vendor/bin/phpunit -c phpunit.xml --verbose --testdox' on the tests-cli container. 
This version of PHPUnit requires PHP >= 7.3.
You are using PHP 7.2.34 (/usr/local/bin/php).
```

swissspidy:
> @swissspidy Added log in script. In local i got below error for PHP 7.2.
> 
> ```
> ℹ Starting 'vendor/bin/phpunit -c phpunit.xml --verbose --testdox' on the tests-cli container. 
> This version of PHPUnit requires PHP >= 7.3.
> You are using PHP 7.2.34 (/usr/local/bin/php).
> ```

Aah 💡 

Now it makes sense. We need to downgrade PHPUnit in this case using `composer update`, just like we do for unit tests.

Maybe 7d560aeb42b528eed8b1449174fc35d3e4cd795b works 🤞

swissspidy:
Or maybe not 🤷

swissspidy (approved):
Yay 🤩🥳

build-cs/composer.json
westonruter, line 1: Perhaps a naïve question: what does the "CS" stand for in `build-cs`?

swissspidy: Coding standards

westonruter (approved):

composer.json
```diff
@@ -12,35 +12,34 @@
...
+      "build-cs/vendor/bin/phpstan analyse --memory-limit=2048M -c phpstan.neon.dist"
+    ],
+    "format": [
+      "composer --working-dir=build-cs update --no-interaction",
+      "build-cs/vendor/bin/phpcbf --standard=phpcs.xml.dist --report-summary --report-source"
+    ],
+    "lint": [
+      "composer --working-dir=build-cs update --no-interaction",
+      "build-cs/vendor/bin/phpcs --standard=phpcs.xml.dist"
+    ],
```
westonruter, line 34: I found that calling `composer update` here causes problems: https://github.com/WordPress/performance/issues/961

Proposed fix: https://github.com/WordPress/performance/pull/962

PR #549, opened 2022-10-03
Bump minimum WordPress requirement to 6.0

### LINKED ISSUE #531: Bump minimum WordPress requirement to 6.0
**With this issue, I'm proposing to bump the minimum WordPress version requirement of the Performance Lab plugin to 6.0 in the upcoming 1.6.0 version (October 17).**

This is in line with https://github.com/WordPress/performance/blob/trunk/docs/Version-support-policy.md#wordpress-core-versions, and even more so, if we do this in the upcoming 1.6.0 release (October 17), we would be very close to even the WordPress 6.1 release (November 1), so realistically we would support the latest _two_ WordPress versions in the coming months (similar to how initially we supported 5.9 and 5.8 when 5.9 was the latest version).

Currently, the Performance Lab plugin requires at least WordPress 5.8.

As mentioned in the version policy, as soon as there's a benefit to bumping the WordPress version requirement, we should be able to do so as long as it is to a stable version. One of those is the introduction of the `filesize` metadata for attachments, which was added in 6.0 and heavily benefits logic in the WebP Uploads module. While this is just one benefit, with this being a feature plugin, we don't need to worry too much about keeping support for old WordPress core versions as long as the newer one provides us a clear benefit from a code perspective.

mukeshpanchal27:
## Summary

Fixes #531

## Checklist


pushpakpop (approved):

jjgrainger (approved):

felixarntz (approved):
Thanks @mukeshpanchal27!

Issue #526, opened 2022-09-12
Module Proposal: Optimized Autoloaded Options


felixarntz:
## Overview
- Proposed module name: Optimized Autoloaded Options
- Proposed module slug: optimized-autoloaded-options
- Proposed module owner GitHub username(s): markjaquith, felixarntz

This module proposal was created during WordCamp US 2022 as a collaboration between @merkys7 @markjaquith @felixarntz @sabrina-zeidan @rytisder.

## About the module

WordPress has basically forever had a mechanism to load many (likely most) of its options in one single database request, by relying on a single flag `autoload` which could be set to `yes` or `no`. This is being done to improve performance, since having individual database queries for every option used on a page would be a lot slower.

However, the decision whether to `autoload` an option is currently left to plugins and themes (when they call `add_option()` or `update_option()`), without making it a requirement: There is a default value of `yes`, and many (likely most) plugin and theme developers are not aware of what that value does or that it even exists.

This often results in an extremely bloated list of options to autoload, with many of these options unused, especially over time as plugins are tried out and then deleted, replaced with other plugins etc. As of today there is no mechanism in core to clean up any of this.

### Purpose

The primary goal of the proposed module is to keep the list of options to autoload at a sizable level, to make the relevant database request faster and to avoid the scenario where the list of those options and values is too large to fit into an object cache solution.

Developers of plugins and themes should no longer be responsible for this as this is a very low level problem that WordPress core can solve by itself better than offloading it to each individual extension. Additionally, if this was based on an API, it would be unlikely or at least take years (if not decades) that plugins and themes would use such an API correctly at scale.

### Scope
The module would do the following to achieve the above goal:
* Track which options are actually used on the site and based on that regularly rebuild the list of options to autoload, by dynamically modifying the `autoload` column in the database.
    * This routine would hook into get_option() calls to keep track of used options within a request.
    * It would be periodically restarted, so that it does not just grow forever (since one of the goals explicitly to clear out old “junk”).
    * It would only be started when there is a logged in user with sufficient capabilities, to avoid any notable database write to happen in an unauthenticated request (which could also lead to race conditions on high traffic sites).
    * Subsequently, any new options being requested that were not autoloaded yet would be added also from unauthenticated requests. This would result in minimal write requests over time, but it would be critical since different templates and screens may use different options.
* Avoid autoloading options with values that are greater than a specific threshold (TBD).
[...]

PR #553, opened 2022-10-06
Implement Server-Timing API foundation as well as basic load time metrics

### LINKED ISSUE #551: Introduce Server-Timing API as part of infrastructure
## Context

Common web performance metric and auditing tools (e.g. Lighthouse, PageSpeed Insights, WebPageTest) only focus on client-side performance in detail. Server-side performance gets reduced to a single metric ([TTFB](https://web.dev/ttfb/)), simply because these tools cannot look "behind the scenes". This means we have no clear way to figure out how e.g. a feature in the Performance Lab plugin affects server-side performance in the lab or in the field.

The [Server-Timing header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing) is a standardized way to expose server-side metrics to the client-side. This way, performance tools can use and evaluate such data, which was previously hidden from them. Of course the server has to somehow measure and expose the results though, and that is what this issue is about.

## Proposal
The Performance Lab plugin should come with a PHP API to centrally control the [Server-Timing header](https://w3c.github.io/server-timing/#the-server-timing-header-field) which, most importantly, allows developers to pass identifiers for a metric and their values to the API, with the API then managing the header output.

This API should be part of the plugin infrastructure, i.e. _not_ a module. The reason for that is that this API should be usable by any module to measure certain aspects of its performance impact, which would not be reliable if it was a module. Another reason to not make the API a module is that modules should come with functionality on their own, which would not be the case for this API. Last but not least, unlike the purpose of modules, this API should not be proposed for inclusion in WordPress core (at least not in the near term) - its main purpose is to allow us to gather server-side metrics in the lab and in the field, which we have not been able to so far.

### Defaults
Primarily, individual modules of the plugin should use the API to measure specific parts of their execution relevant to the feature they implement, in order to see how they affect PHP performance. However, a single default is worth including, which is the overall WordPress execution time, which WordPress core already tracks itself using its [timer_start()](https://developer.wordpress.org/reference/functions/timer_start/) function.

### High-level spec

* Class `Perflab_Server_Timing`:
    * `register_metric( string $metric_slug, callable $measure_callback, string|callable $access )`: Registers a metric and its measuring callback, to be added to the list of metrics. The measuring callback will only be executed if the access is satisfied by the current request.
        * Potentially for a more flexible API the signature of this method should use `array $args` as second parameter, but the exact parameters are outlined above for clarity.
    * `get_header_value() : string`: Returns the full Server-Timing header value, based on all registered metrics which have a value set.
    * `add_header()`: Adds the Server-Timing header, using `get_header_value()`.
* Function `perflab_server_timing()`: Initializes the main instance of the `Perflab_Server_Timing` class and fires an action hook for any extension to integrate, passing the instance to the action as a parameter.
* Function `perflab_wrap_server_timing( callable $callback, string $metric_slug, string|callable $access ) : callable`: Wraps a callback (e.g. of an `add_action()` or `add_filter()` call) into this function, which uses `Perflab_Server_timing::register_metric()` internally to simply measure execution time of the `$callback` function and add it to the Server-Timing header value.
    * This little utility will simplify usage of the API around hook callbacks and is inspired by [this article](https://calendar.perfplanet.com/2018/server-timing/#instrumentation).

### Quality attributes (performance, privacy, security, accessibility)

* **Header length**: While it is perfectly acceptable to include various metrics and values in the Server-Timing header, it should not become too long in order to minimize the HTTP overhead. For that reason, the API should include validation to ensure the metric names are kept short and concise (exact character limit per metric TBD), and it should furthermore limit the overall number of metrics that can be included in the header as well.
* **Timing overhead**: While measuring time in PHP is generally considered harmless, an excessive number of timers may cause a performance regression. For example, WordPress core conditionally measures the time every individual database request takes, though it does so only conditionally if the `SAVEQUERIES` flag is set. Note that this has some overlap with the header length consideration above and will potentially already be satisfied by the solution for that.
* **Sensitive values**: The values for each metric should be numeric, e.g. integers or floats. However, since the Server-Timing header is technically a string, any other value could also be rendered, which opens room for abuse, where code in the WordPress site could leak sensitive information as part of that value. To avoid that, the API should require all timing values to be numeric.
* **Sensitive metrics**: There are certain metrics which a site may want to measure but keep internal. For this purpose, the API should require for each metric to be registered with a certain access level definition, so that not every metric is public. This could either be a WordPress capability check or a shorthand such as “public” or “admin”.

felixarntz:
## Summary

Fixes #551

## Relevant technical choices

* Add `Perflab_Server_Timing` class to control the Server-Timing header.
* Add `Perflab_Server_Timing_Metric` class to represent a single metric, so that measuring callbacks can set a value on it without having access to the full API.
* Add `perflab_server_timing_register_metric()` function which can be used to register a Server-Timing metric. This function can be called at any point as long as it is before or during the `perflab_server_timing_send_header` action.
* Add `perflab_server_timing_use_output_buffer()` function and filter of the same name. By default output buffering is disabled, but it can be enabled to record additional measurements that occur while rendering the template.
* Add `perflab_server_timing()` function which provides access to the API instance, e.g. to use it for lower-level functions, which are typically not necessary due to the wrapper functions mentioned above.
* Add two functions to register default metrics. These are:
    * `wp-before-template` measures how long WordPress execution time takes prior to serving the HTML template
    * `wp-before-template-db-queries` measures total execution of database queries prior to serving the HTML template (**only** if `SAVEQUERIES` constant is enabled)
    * `wp-template` measures the overall HTML template execution (**only** if using an output buffer)
    * `wp-template-db-queries` measures total execution of database queries during HTML template execution (**only** if using an output buffer and if `SAVEQUERIES` constant is enabled)
    * `wp-load-alloptions-query` measures execution of the alloptions database query (**only** if the Performance Lab `object-cache.php` drop in is placed, see below)
* Implement `perflab_wrap_server_timing()` utility function which can be used to measure execution of any callback (e.g. for an action or filter).
* Include an `object-cache.php` drop-in, which is used to initialize the Server-Timing API early, in order to measure events that would not be possible from within a plugin. 
    * This file is placed automatically if possible, in a way that's entirely non-disruptive.
    * If there is already a (real) `object-cache.php` file, it will be renamed to `object-cache-plst-orig.php` and then required from the new `object-cache.php` file, so nothing will change in the behavior of any persistent object cache present.
    * Placing the file will only be attempted in WP Admin.
    * If placing the file fails for any reason, it will only be attempted once per hour maximum (and only if logged into WP Admin, see above).
    * If a site still does not want to use the drop-in, it can set a constant: `define( 'PERFLAB_DISABLE_OBJECT_CACHE_DROPIN', true );`
    * The PR includes one more specific metric which is facilitated by having this drop-in, which is to measure the autoloaded options query. This query is added in support of #526 (also since it would not be possible to add this query from within a plugin).

## Testing

Server-Timing header information can be inspected in JS via:
```js
performance.getEntries()[0].serverTiming
```

Use the following code snippet to test with output buffering enabled:
```php
add_filter( 'perflab_server_timing_use_output_buffer', '__return_true' );
```

You can test the `perflab_wrap_server_timing()` function e.g. by adding a code snippet like the following:
```php
add_action(
	'init',
	perflab_wrap_server_timing(
		function() {
			// Wait for 1 second.
			usleep( 1000000 );
		},
		'init-test-callback',
		'exist'
	)
);
```

Some other example use cases to test with can be found in this Gist: https://gist.github.com/felixarntz/43fb19e6491a1bc65a7e930dace17c47

## Checklist


felixarntz:
In https://github.com/WordPress/performance/pull/553/commits/51c4222c60327d5b40667dda5edc3b6b64afa269, I have revamped the API to simplify its usage quite a bit. I've updated the PR description accordingly and created a new Gist with updated examples (https://gist.github.com/felixarntz/43fb19e6491a1bc65a7e930dace17c47).

For reference, I've kept the original example Gist (https://gist.github.com/felixarntz/c343a33242b6527ad3f446c8f7f4fec2). By comparing the two, you will see that the new API usage is much simplified. Most importantly the new approach has the following benefits:
* You aren't forced to hook into a specific action to register a Server-Timing metric.
* This encourages registration of metrics as close as possible to when the relevant logic runs rather than registering all metrics early on - essentially lazy initialization, to avoid registering metrics which may not be used during a page load.
* Through that it also allows the core infrastructure to not run a measuring callback if the current user lacks the necessary capability, saving resources (even though these aren't expensive to run, it's a nice-to-have).
* There is now an action right before the Server-Timing header is sent, so that you can rely on it or check for it to ensure you register metrics in time for that.
* The new `measure_before()` and `measure_after()` methods simplify measurement as you no longer need to manually store the reference value and calculate the diff.

felixarntz:
@johnbillion @tillkruss @jjgrainger I'd like to get your feedback on this. I'm still keeping it as a draft PR until I have added test coverage, but at this point I think the implementation would be suitable for the Performance Lab plugin.

Some high-level considerations:
* Currently, using an output buffer is off by default. Could we consider enabling it by default (since we're in plugin territory, and therefore it's certainly feasible here)?
* Some metrics that would be valuable to measure cannot be captured from within a plugin because they are running too early (e.g. the alloptions query). I wonder if we can find ways to address that. Given this is a beta testing plugin, could we potentially place a drop-in file to register some early metrics?

For the latter, it would be a hack, but I'm thinking potentially we could place e.g. `object-cache.php` (without that being actually an object cache implementation). If the site is not using an object cache, that'll still work as before, and if it does, we could easily detect that it has `object-cache.php` already and in that case rename that file, then `require_once` it within our own `object-cache.php` file. Too crazy, or potentially okay? :)

tillkruss:
The `advanced-cache.php` might be more suited for that, since it loads a lot earlier than the `object-cache.php`.

tillkruss:
Can you give me a gist how an Object Cache implementation would supply the Server Timing API with let's say a single metric "total time waited for Datastore", similar to "total time waited for SQL"?

felixarntz:
@tillkruss For the actual measurement, you would probably need to add support for that in the `object-cache.php` implementation itself, e.g. use `microtime( true )` or something like that around every single cache lookup, then store that somewhere. Maybe you could rely on a constant similar to `SAVEQUERIES` to enable that?

You could then register the data like this:
```php
add_action(
	'perflab_server_timing_send_header',
	function() {
		perflab_server_timing_register_metric(
			'object-cache-lookups',
			array(
				'measure_callback' => function( $metric ) {
					// The function here should return the time in milliseconds.
					$total_cache_time = demo_function_that_returns_total_cache_time();
					$metric->set_value( $total_cache_time );
				},
				'access_cap'       => 'exist',
			)
		);
	}
);
```

The hook is called as late as possible (either before serving the HTML template or on shutdown, depending on whether the feature is configured with an output buffer or not for the site), so all lookups should be calculated by then at least if an output buffer is used. For sites without an output buffer (see `perflab_server_timing_use_output_buffer()`), if there is still value in including this, you could call it something like `before-template-object-cache-lookups`, kind of similar how it is in this PR for the DB queries (since without an output buffer you can only account for the lookups that happen prior to serving the template).

tillkruss:
@felixarntz: This is great 👍

felixarntz:
@tillkruss @johnbillion In https://github.com/WordPress/performance/pull/553/commits/096bb87fc14fbf68ff8f4df584369d8e46091104, I've added logic to place an `object-cache.php` drop-in to initialize the Server-Timing API early, as mentioned in https://github.com/WordPress/performance/pull/553#issuecomment-1297815746. While this is a bit of a hack, I think it is totally reasonable. Unless I'm missing something, the implementation should not be disruptive in any way. Please take a look. Also review the PR description, which I keep updating based on the latest implementation.

I will work on tests within the next 2 weeks and get this ready for a full code review.

server-timing/defaults.php
```diff
@@ -0,0 +1,172 @@
...
+				array(
+					'measure_callback' => function( $metric ) {
+						// Store this value in a global to later subtract it from total query time after template.
+						$GLOBALS['perflab_query_time_before_template'] = array_reduce(
+							$GLOBALS['wpdb']->queries,
+							function( $acc, $query ) {
+								return $acc + $query[1];
+							},
+							0.0
+						);
```
eugene-manuilov, line 49: Got the following error on two hosting providers - Seravo and WPEngine. WPEngine has fixed the issue by itself and i don't see it anymore. Seravo still throws this error even with `SAVEQUERIES` enabled.

```
Fatal error:  Uncaught TypeError: array_reduce(): Argument #1 ($array) must be of type array, null given in /data/wordpress/htdocs/wp-content/plugins/performance-add-551-server-timing-api/server-timing/load.php:151
Stack trace:
#0 /data/wordpress/htdocs/wp-content/plugins/performance-add-551-server-timing-api/server-timing/load.php(151): array_reduce(NULL, Object(Closure), 0)
#1 /data/wordpress/htdocs/wp-content/plugins/performance-add-551-server-timing-api/server-timing/class-perflab-server-timing.php(98): {closure}(Object(Perflab_Server_Timing_Metric))
#2 /data/wordpress/htdocs/wp-content/plugins/performance-add-551-server-timing-api/server-timing/load.php(50): Perflab_Server_Timing->register_metric('template-db-que...', Array)
#3 /data/wordpress/htdocs/wp-content/plugins/performance-add-551-server-timing-api/server-timing/load.php(155): perflab_server_timing_register_metric('template-db-que...', Array)
#4 /data/wordpress/htdocs/wordpress/wp-includes/class-wp-hook.php(308): {closure}('')
#5 /data/wordpress/htdocs/wordpress/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters('', Array)
#6 /data/wordpress/htdocs/wordpress/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#7 /data/wordpress/htdocs/wp-content/plugins/performance-add-551-server-timing-api/server-timing/class-perflab-server-timing.php(125): do_action('perflab_server_...')
#8 /data/wordpress/htdocs/wp-content/plugins/performance-add-551-server-timing-api/server-timing/class-perflab-server-timing.php(210): Perflab_Server_Timing->send_header()
#9 /data/wordpress/htdocs/wordpress/wp-includes/class-wp-hook.php(308): Perflab_Server_Timing->{closure}('')
#10 /data/wordpress/htdocs/wordpress/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters('', Array)
#11 /data/wordpress/htdocs/wordpress/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#12 /data/wordpress/htdocs/wordpress/wp-includes/load.php(1124): do_action('shutdown')
#13 [internal function]: shutdown_action_hook()
#14 {main}
  thrown in /data/wordpress/htdocs/wp-content/plugins/performance-add-551-server-timing-api/server-timing/load.php on line 151
```

eugene-manuilov: The funniest part is that the server-timing header works correctly on another site that we have on Seravo hosting. 🤔

felixarntz: @eugene-manuilov So for some reason `$GLOBALS['wpdb']->queries` is not an array in that situation? That definitely seems like a bug on their end, but happy to add a `isset && is_array` check here just to cater for weird setups :)

felixarntz: Addressed in https://github.com/WordPress/performance/pull/553/commits/94304921e4e48af89f5961470a9c8f8659a44b36.

eugene-manuilov: > So for some reason `$GLOBALS['wpdb']->queries` is not an array in that situation?

Yep. That's weird but for some reason that's `NULL`. 🤷 

> Addressed in [9430492](https://github.com/WordPress/performance/commit/94304921e4e48af89f5961470a9c8f8659a44b36).

Thanks!

server-timing/defaults.php
```diff
@@ -0,0 +1,178 @@
...
+								return;
+							}
+
+							$total_query_time = array_reduce(
+								$GLOBALS['wpdb']->queries,
+								function( $acc, $query ) {
+									return $acc + $query[1];
+								},
+								0.0
+							);
```
eugene-manuilov, line 181: @felixarntz same problem here, we need to check whether the `$GLOBALS['wpdb']->queries` is not NULL.

felixarntz: @eugene-manuilov Technically I could add the extra check, but the one above should already prevent that problem (since that global only gets set if `$GLOBALS['wpdb']->queries` is an array in the metric measure callback further up, which runs before.

felixarntz:
With initial test coverage, this is now ready for full review. Adding to 1.8.0 milestone for consideration.

I will add remaining test coverage (for `Perflab_Server_Timing` and `Perflab_Server_Timing_Metric` classes) over the next few days.

load.php
```diff
@@ -253,9 +263,56 @@ function perflab_load_active_and_valid_modules() {
...
+
+	if ( $wp_filesystem || WP_Filesystem() ) {
+		// If there is an actual object-cache.php file, rename it.
+		// The Performance Lab object-cache.php will still load it, so the
+		// behavior does not change.
+		if ( $wp_filesystem->exists( WP_CONTENT_DIR . '/object-cache.php' ) ) {
+			$wp_filesystem->move( WP_CONTENT_DIR . '/object-cache.php', WP_CONTENT_DIR . '/object-cache-orig.php' );
+		}
+
+		$wp_filesystem->copy( PERFLAB_PLUGIN_DIR_PATH . 'server-timing/object-cache.php', WP_CONTENT_DIR . '/object-cache.php' );
```
eugene-manuilov: The problem here is that it won't work as expected if a site has multiple servers for WordPress. Maybe instead of self-installing this drop-in, we should tell users what they need to do if they want to have more accurate server timing metrics?

server-timing/load.php
```diff
@@ -0,0 +1,107 @@
...
+
+		perflab_server_timing_register_metric(
+			$metric_slug,
+			array(
+				'measure_callback' => function( $metric ) use ( &$server_timing_metric ) {
+					$server_timing_metric = $metric;
+				},
+				'access_cap'       => $access_cap,
+			)
+		);
```
eugene-manuilov: Shouldn't it be set in the upper scope and passed into this anonymous function?

```php
function perflab_wrap_server_timing( $callback, $metric_slug, $access_cap ) {
    $server_timing_metric = null;
    
    perflab_server_timing_register_metric(
        ...
    );
    
    return function( ...$calback_args ) use ( $server_timing_metric, $callback ) {
        if ( ! $server_timing_metric ) {
            return call_user_func_array( $callback, $callback_args );
        }
        
        $server_timing_metric->measure_before();
        $result = call_user_func_array( $callback, $callback_args );
        $server_timing_metric->measure_after();
        
        return $result;
    };
}
```

The problem with the current implementation is that it registers the metric every time when the measurement function returned from `perflab_wrap_server_timing` function is called. If we use as an action hook, we will re-register the same metric number of times the action is done.

server-timing/defaults.php
eugene-manuilov: Ah... I see now. Maybe we should add the extra check anyway to make sure we are safe?

load.php
felixarntz: @eugene-manuilov I think that's okay. The vast majority of sites are on a single server, so automatically setting that up benefits the majority. Also without this being automated, it would result in much lower usage just because expecting the end users to place that file may be a bit too much to ask as a baseline experience.

The feature is already built in a way that, if placing the file does not work (for whatever reason, real-world servers are configured in _weird_ ways sometimes), everything would keep working as expected, just without any early metrics reporting. So I'd argue that wouldn't be a major problem for sites on multiple servers either. Also, to be fair, probably the average site with multiple servers is a "professional" setup where that file couldn't be automatically placed anyway (writing to the filesystem to other than the `uploads` directory is often disabled).

server-timing/load.php
felixarntz: @eugene-manuilov Great catch, I didn't consider that. For some context, I wanted to put the registration in the function so that the metric is only registered when the function is actually executed, which I'd like to keep.

But we can definitely find a way to still ensure the metric is only registered once. However, the bigger question now is: What is the expected behavior for its measurement when the function is called more than once?
1. Should only the first call be measured?
2. Should each call be measured and the sum of the duration be reported as the metric value?

WDYT?

server-timing/defaults.php
felixarntz: @eugene-manuilov Good idea, will add this.

load.php
```diff
@@ -253,9 +263,56 @@ function perflab_load_active_and_valid_modules() {
...
+ * Places the Performance Lab's object cache drop-in in the drop-ins folder.
+ *
+ * This only runs in WP Admin to not have any potential performance impact on
+ * the frontend.
+ *
+ * This function will short-circuit if the constant
+ * 'PERFLAB_DISABLE_OBJECT_CACHE_DROPIN' is set as true.
+ *
+ * @since n.e.x.t
+ */
```
mukeshpanchal27:
```suggestion
 * @since n.e.x.t
 *
 * @global WP_Filesystem_Base $wp_filesystem WordPress filesystem subclass.
 */
```

load.php
```diff
@@ -253,9 +263,56 @@ function perflab_load_active_and_valid_modules() {
...
+	if ( false !== $timeout ) {
+		return;
+	}
+
+	if ( $wp_filesystem || WP_Filesystem() ) {
+		// If there is an actual object-cache.php file, rename it.
+		// The Performance Lab object-cache.php will still load it, so the
+		// behavior does not change.
+		if ( $wp_filesystem->exists( WP_CONTENT_DIR . '/object-cache.php' ) ) {
+			$wp_filesystem->move( WP_CONTENT_DIR . '/object-cache.php', WP_CONTENT_DIR . '/object-cache-orig.php' );
```
mukeshpanchal27: What happens when the user deactivates the plugin? Do we revert this change and rename `object-cache-orig.php` to `object-cache.php`?

mukeshpanchal27 (commented):
Thanks @felixarntz, The PR looks good, though I have a question on the one change.

load.php
felixarntz: @mukeshpanchal27 Great catch! I've addressed this in https://github.com/WordPress/performance/pull/553/commits/6f2f1c59cbb294c4871bf14ac4ba3912176796b5.

Even if the file is not removed, it wouldn't cause problems since its code will still include the original drop-in, and only load the Performance Lab code if it is present. However I agree with you that it's a best practice to remove our custom drop-in on plugin deactivation.

server-timing/class-perflab-server-timing.php
```diff
@@ -0,0 +1,255 @@
...
+				''
+			);
+			return;
+		}
+
+		$args = wp_parse_args(
+			$args,
+			array(
+				'measure_callback' => null,
+				'access_cap'       => null,
```
ankitrox, line 76: Is it possible to have a default `access_cap` (maybe the top most capability like `manage_options` or `manage_network` for multisite), so that every time the custom metric is getting registered, access cap need not to be passed?

server-timing/class-perflab-server-timing.php
```diff
@@ -0,0 +1,255 @@
...
+		}
+
+		$this->registered_metrics[ $metric_slug ]      = new Perflab_Server_Timing_Metric( $metric_slug );
+		$this->registered_metrics_data[ $metric_slug ] = $args;
+
+		// If the current user has already been determined and they lack the necessary access,
+		// do not even attempt to calculate the metric.
+		if ( did_action( 'set_current_user' ) && ! current_user_can( $args['access_cap'] ) ) {
+			return;
+		}
```
ankitrox, line 105: What if the current user is not determined? I think it will still run the callback?

ankitrox (commented):
LGTM. Left couple of questions.

server-timing/class-perflab-server-timing.php
10upsimon: Initially I was unsure about this, but I think @ankitrox _might_ have a point here, especially given that we instantiate the Server Timing API at object-cache.php drop-in level, which essentially allows the use of the Server Timing API from that point in the execution cycle. 

What I am unsure about though, is whether a user has been established at this point or not?

10upsimon:
@felixarntz  I have reviewed the code in this PR and tested the API locally (adding 3 custom metrics). I am happy with the outcome and I had no issues implementing it.

I'm going to await the outcome of @ankitrox 's review comment at https://github.com/WordPress/performance/pull/553/files/7c87c3104ba9aaf337904601daecad128c258d0b#r1047058595 before I add my final approval, as I am curious about the answer to said comment myself.

ankitrox:
I tested this API locally with custom metrics and seems to be working as expected.

.wp-env.json
```diff
@@ -3,6 +3,9 @@
 	"plugins": [ "." ],
 	"env": {
 		"tests": {
+      "config": {
+        "FS_METHOD": "direct"
+      },
```
eugene-manuilov, line 8: Need to fix indentation.

load.php
eugene-manuilov: Ok, sounds good then. If this doesn't break anything on multi-server setups then it's fine with me.

server-timing/load.php
eugene-manuilov: Yes, these are good questions... Probably it depends on circumstances in which this function is used. I think lets leave this as is for now and if someone faces any problems with it, we will get back to it and think about a better approach.

10upsimon:
@felixarntz this is a real nit pick comment, but I am raising it _**in case**_ your summary gets lifted word for word for any docs. 

You have this mentioned under `Relevant technical choices` in your summary:

```
If there is already a (real) object-cache.php file, it will be renamed to object-cache-orig.php and then required from the new object-cache.php file, so nothing will change in the behavior of any persistent object cache present.
```

The correct name that the original is renamed to is `object-cache-plst-orig.php` and _not_ `object-cache-orig.php` as in the overview. 

See https://github.com/WordPress/performance/pull/553/files#diff-733209cd1718b5cb445345c12b0fc02666cf6aab6e597c59fef8a9c13b33f047R307

load.php
```diff
@@ -253,9 +263,101 @@ function perflab_load_active_and_valid_modules() {
...
+	if ( false !== $timeout ) {
+		return;
+	}
+
+	if ( $wp_filesystem || WP_Filesystem() ) {
+		// If there is an actual object-cache.php file, rename it.
+		// The Performance Lab object-cache.php will still load it, so the
+		// behavior does not change.
+		if ( $wp_filesystem->exists( WP_CONTENT_DIR . '/object-cache.php' ) ) {
+			$wp_filesystem->move( WP_CONTENT_DIR . '/object-cache.php', WP_CONTENT_DIR . '/object-cache-plst-orig.php' );
```
adamsilverstein, line 307: curious what `object-cache-plst-orig` stands for, why not `object-cache-original`?

server-timing/class-perflab-server-timing.php
```diff
@@ -0,0 +1,255 @@
...
+	 *
+	 * @param string $metric_slug The metric slug.
+	 * @param array  $args        {
+	 *     Arguments for the metric.
+	 *
+	 *     @type callable $measure_callback The callback that initiates calculating the metric value. It will receive
+	 *                                      the Perflab_Server_Timing_Metric instance as a parameter, in order to set
+	 *                                      the value when it has been calculated. Metric values must be provided in
+	 *                                      milliseconds.
+	 *     @type string   $access_cap       Capability required to view the metric. If this is a public metric, this
```
adamsilverstein, line 47: should there be a default for this value if not passed?

server-timing/class-perflab-server-timing.php
```diff
@@ -0,0 +1,255 @@
...
+			);
+			return;
+		}
+
+		$this->registered_metrics[ $metric_slug ]      = new Perflab_Server_Timing_Metric( $metric_slug );
+		$this->registered_metrics_data[ $metric_slug ] = $args;
+
+		// If the current user has already been determined and they lack the necessary access,
+		// do not even attempt to calculate the metric.
+		if ( did_action( 'set_current_user' ) && ! current_user_can( $args['access_cap'] ) ) {
```
adamsilverstein, line 103: I assume this is to prevent non user invocation - out of curiosity, when is `did_action( 'set_current_user' )` not true? during something like a cron job, or even on unauthenticated requests? why not a more explicit call like `wp_get_current_user`?

server-timing/class-perflab-server-timing.php
adamsilverstein: > What if the current user is not determined? I think it will still run the callback?

Don't we want to gather metrics for non-authenticated users as well? authenticated loads are very different that anonymous loads.

server-timing/class-perflab-server-timing.php
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	/**
+	 * Gets the value for the Server-Timing header.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string The Server-Timing header value.
+	 */
+	public function get_header_value() {
```
adamsilverstein: this gets all the header values, maybe rename `get_header_values` (plural)? even though the return is a string, it contains more than one "value"

server-timing/class-perflab-server-timing.php
```diff
@@ -0,0 +1,255 @@
...
+			function( $value ) {
+				return null !== $value;
+			}
+		);
+
+		return implode( ', ', $metric_header_values );
+	}
+
+	/**
+	 * Returns whether an output buffer should be used to gather Server-Timing metrics during template rendering.
```
adamsilverstein, line 187: Might be useful to add one line doc here indicating **why** you might want to do this

server-timing/class-perflab-server-timing.php
```diff
@@ -0,0 +1,255 @@
...
+	 * This effectively initializes the class to send the Server-Timing header at the right point.
+	 *
+	 * This method is solely intended for internal use within WordPress.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param mixed $passthrough Optional. Filter value. Default null.
+	 * @return mixed Unmodified value of $passthrough.
+	 */
+	public function on_template_include( $passthrough = null ) {
```
adamsilverstein, line 224: core functions intended for internal use only are usually prefixed with `_`, maybe we should follow that here?

server-timing/class-perflab-server-timing.php
```diff
@@ -0,0 +1,255 @@
...
+
+		ob_start();
+		add_action(
+			'shutdown',
+			function() {
+				$output = ob_get_clean();
+				$this->send_header();
+				echo $output;
+			},
+			-1000
```
adamsilverstein: curious why this needs to be so "early" (and how -1000 was chosen)

server-timing/class-perflab-server-timing.php
```diff
@@ -0,0 +1,255 @@
...
+	 * Formats the header segment for a single metric.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Perflab_Server_Timing_Metric $metric The metric to format.
+	 * @return string Segment for the Server-Timing header.
+	 */
+	private function format_metric_header_value( Perflab_Server_Timing_Metric $metric ) {
+		$value = $metric->get_value();
+		if ( is_float( $value ) ) {
```
adamsilverstein, line 260: should this have another check to make sure the value is a number (ie set at all)

server-timing/defaults.php
```diff
@@ -0,0 +1,187 @@
...
+ */
+function perflab_register_default_server_timing_before_template_metrics() {
+	$calculate_before_template_metrics = function() {
+		// WordPress execution prior to serving the template.
+		perflab_server_timing_register_metric(
+			'before-template',
+			array(
+				'measure_callback' => function( $metric ) {
+					// The 'timestart' global is set right at the beginning of WordPress execution.
+					$metric->set_value( ( microtime( true ) - $GLOBALS['timestart'] ) * 1000.0 );
```
adamsilverstein, line 24: is `$GLOBALS['timestart']` always defined? if not, add a check here

server-timing/defaults.php
```diff
@@ -0,0 +1,187 @@
...
+		function() use ( $calculate_before_template_metrics ) {
+			if ( ! perflab_server_timing_use_output_buffer() ) {
+				$calculate_before_template_metrics();
+			}
+		},
+		PHP_INT_MAX
+	);
+
+	// Measure duration of autoloaded options query.
+	// Requires the Performance Lab object-cache.php drop-in to be present in order to work.
```
adamsilverstein: Add a check early in here for `PERFLAB_OBJECT_CACHE_DROPIN_VERSION` and skip if not present

server-timing/defaults.php
```diff
@@ -0,0 +1,187 @@
...
+				)
+			);
+			return $query;
+		}
+	);
+}
+perflab_register_default_server_timing_before_template_metrics();
+
+/**
+ * Registers the default Server-Timing metrics for while rendering the template.
```
adamsilverstein:
```suggestion
 * Registers the default Server-Timing metrics while rendering the template.
```

server-timing/defaults.php
```diff
@@ -0,0 +1,187 @@
...
+	add_filter(
+		'template_include',
+		function( $passthrough = null ) {
+			// WordPress execution while serving the template.
+			perflab_server_timing_register_metric(
+				'template',
+				array(
+					'measure_callback' => function( $metric ) {
+						$metric->measure_before();
+						add_action( 'perflab_server_timing_send_header', array( $metric, 'measure_after' ), PHP_INT_MAX );
```
adamsilverstein, line 144: this `add_action` call looks off, shouldn't the second parameter be a callback function?

server-timing/defaults.php
```diff
@@ -0,0 +1,187 @@
...
+					'access_cap'       => 'exist',
+				)
+			);
+
+			return $passthrough;
+		},
+		PHP_INT_MAX
+	);
+
+	if ( defined( 'SAVEQUERIES' ) && SAVEQUERIES ) {
```
adamsilverstein, line 156: why is this metric limited to when `SAVEQUERIES` is set? suggest adding an inline comment here explaining.

load.php
adamsilverstein: Also, since this is used repeatedly, maybe put in a constant?

server-timing/class-perflab-server-timing.php
felixarntz: @ankitrox I wanted to intentionally avoid a default here, to make sure developers think about it. Having a default is not always a good idea. If we use a strict default (like `manage_options`), it will lead to most metrics not being visible in the frontend, making it almost useless for field analysis. If we use a loose default (like `exist`), it will lead to sensitive metrics being visible in the frontend when they shouldn't.

server-timing/class-perflab-server-timing.php
felixarntz: @ankitrox @10upsimon Yes, it will still be run. However, that does not present a real issue. Here's why:

* The primary purpose of the `access_cap` is to control whether the metric is exposed in the `Server-Timing` header.
* Because of that, the cap is checked when rendering the header (in the `get_header_value()` method). That is the actually critical check. And at that point, for sure the user will have been set.
* The check here is just a very minor performance consideration: Basically _if_ at this point the user is already set, we can check the capability right away and not even calculate the metric.
* If the user isn't set at that point, running the function to calculate the metric is perfectly fine; alternatively we _could_ ensure the current user is set early (normally it's set right before the `init` action), but that's not something I like to do especially given the little benefit it would bring here.

Let me know if that satisfies your concerns.

felixarntz:
@10upsimon @ankitrox I've addressed all your feedback.

@eugene-manuilov @adamsilverstein Will work on addressing your feedback next.

load.php
felixarntz: @adamsilverstein `plst` = `Performance Lab Server Timing`

The reason I didn't go with something more generic (I actually used `-orig` first but then added the `plst` to it) is that it makes it less likely to have any sort of conflict. If another plugin has a similar mechanism to automatically move or back up the `object-cache.php` file, it's likely they would use something like `-orig` or `-original`. So this more specific name is just to avoid the conflict.

server-timing/class-perflab-server-timing.php
felixarntz: @adamsilverstein See my response to a similar question in https://github.com/WordPress/performance/pull/553#discussion_r1048626240, I intentionally did not want to add a default for that reason. Sometimes defaults can be rather detrimental (a good example is the `autoload = yes` default in core - would have been better if there was none, requiring developers to think about it.

server-timing/class-perflab-server-timing.php
felixarntz: @adamsilverstein The current user is initialized right before `init`. This means for example when calling this on `plugins_loaded` it won't be initialized.

It is certainly possible to initialize the current user earlier than that (by calling any function that relies on the current user), but that's not a great pattern and can lead to other unexpected behavior.

Related, see my response in https://github.com/WordPress/performance/pull/553#discussion_r1048634541 in regards to this check not being critical anyway - it's just a nice to have.

server-timing/class-perflab-server-timing.php
felixarntz: @adamsilverstein I would kindly disagree here; the function returns multiple _Server Timing_ values, but it only returns the one header value for the `Server-Timing` header. That's why I went with this name. LMKWYT.

server-timing/class-perflab-server-timing.php
felixarntz: @adamsilverstein Good point, will add.

server-timing/class-perflab-server-timing.php
felixarntz: @adamsilverstein This is actually mixed in core. There is no clear guidance. The last times I added `_` prefixed core functions, I always got pushback that they should simply have an `@access private` annotation and no longer use such a prefix.

server-timing/class-perflab-server-timing.php
felixarntz: @adamsilverstein My idea was that it should be as soon as possible on `shutdown` since at that point the entire WP response has been served. If other things happen on `shutdown` they shouldn't "pollute" the header output.

Preferably this would use `PHP_INT_MIN`, but that is only available in PHP 7+.

Good point though, I should probably change this to at least use `PHP_INT_MIN` if available. The `-1000` is an arbitrary value just to be early, which I'll change to be just a fallback for when the constant does not exist.

server-timing/class-perflab-server-timing.php
felixarntz: @adamsilverstein Such a check is already present in `get_header_value()` method, which ensures this won't be called if no value is set. But good point, it's probably a good idea to move that in here.

server-timing/defaults.php
felixarntz: @adamsilverstein Yes it will always be set, since WP sets it unconditionally even before the database is initialized.

The only potential edge-case would be if a plugin simply unset the global, but that would be a problem in that plugin. I'm undecided whether it's worth adding a check for.

server-timing/defaults.php
felixarntz: @adamsilverstein Good idea, will add.

server-timing/defaults.php
felixarntz: @adamsilverstein It is, it's a callback method.

server-timing/defaults.php
felixarntz: @adamsilverstein It's because database query time is only recorded in that case. Will add a comment explaining.

.wp-env.json
felixarntz: It's actually the other indentation that was incorrect, not this one :)

Fixed it.

felixarntz:
@adamsilverstein @eugene-manuilov Addressed all your feedback either with new commits or comment replies.

load.php
adamsilverstein: makes sense, thanks for clarifying

server-timing/class-perflab-server-timing.php
adamsilverstein: makes sense, thanks for clarifying

server-timing/class-perflab-server-timing.php
adamsilverstein: 👍🏼

server-timing/class-perflab-server-timing.php
adamsilverstein: don't feel strongly about this. I see it is a single value, just feels slightly wrong because the header contains value**s**. One other idea, just `get_header`?

server-timing/class-perflab-server-timing.php
adamsilverstein: 🤷🏼 OK!

server-timing/defaults.php
adamsilverstein: probably safe to assume it is available, I don't see any plugins modifying it and query monitor uses it directly: https://wpdirectory.net/search/01GM8ZBD0ZT8XF4YGXKHA06XHK

server-timing/defaults.php
adamsilverstein: Oh! I was reading that as an array of values, got it.

adamsilverstein (approved):
Looks good to me!

server-timing/defaults.php
felixarntz: Thanks for the extra research! I had no idea this website existed, super useful 😍

server-timing/class-perflab-server-timing.php
felixarntz: @adamsilverstein Good call! Will rename.

server-timing/class-perflab-server-timing.php
10upsimon: @felixarntz this makes total sense to me now, thank you for clarifying.

10upsimon (approved):

ankitrox (approved):

load.php
```diff
@@ -253,9 +263,101 @@ function perflab_load_active_and_valid_modules() {
...
+
+	if ( $wp_filesystem || WP_Filesystem() ) {
+		// If there is an actual object-cache.php file, restore it
+		// and override the Performance Lab file.
+		// Otherwise just delete the Performance Lab file.
+		if ( $wp_filesystem->exists( WP_CONTENT_DIR . '/object-cache-plst-orig.php' ) ) {
+			$wp_filesystem->move( WP_CONTENT_DIR . '/object-cache-plst-orig.php', WP_CONTENT_DIR . '/object-cache.php', true );
+		} else {
+			$wp_filesystem->delete( WP_CONTENT_DIR . '/object-cache.php' );
+		}
```
eugene-manuilov, line 353: This code should not delete the `object-cache.php` file if it can't find the original `object-cache-plst-orig.php` file because if we can't find that file then it means that we haven't replaced the object cache file.

load.php
```diff
@@ -253,9 +263,101 @@ function perflab_load_active_and_valid_modules() {
...
+	}
+
+	// Delete transient for drop-in check in case the plugin is reactivated shortly after.
+	delete_transient( 'perflab_set_object_cache_dropin' );
+}
+register_deactivation_hook( __FILE__, 'perflab_maybe_remove_object_cache_dropin' );
```
eugene-manuilov, line 359: Perhaps we also need to add this function call to the `uninstall.php` file (or using `register_uninstall_hook` func) to make sure that the drop in is deleted when the plugin has been removed?

server-timing/class-perflab-server-timing-metric.php
```diff
@@ -0,0 +1,140 @@
...
+	 *
+	 * Alternatively to setting the metric value directly, the {@see Perflab_Server_Timing_Metric::measure_before()}
+	 * and {@see Perflab_Server_Timing_Metric::measure_after()} methods can be used to further simplify measuring.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param int|float $value The metric value to set, in milliseconds.
+	 */
+	public function set_value( $value ) {
+		if ( ! is_int( $value ) && ! is_float( $value ) ) {
```
eugene-manuilov: Maybe better to use `! is_numeric( $value )`? Because if someone calls the function `$metric->set_value( '15.2' )` it won't allow them to set it.

server-timing/class-perflab-server-timing.php
```diff
@@ -0,0 +1,265 @@
...
+
+		// If no value is set, make sure it's just passed through.
+		if ( null === $value ) {
+			return null;
+		}
+
+		if ( is_float( $value ) ) {
+			$value = round( $value, 2 );
+		}
+		return sprintf( 'wp-%1$s;dur=%2$s', $metric->get_slug(), $value );
```
eugene-manuilov, line 263: Having the `wp-` prefix here means that all metrics will be prefixed. It also means that if someone decide to add their own metric in a 3rd party plugin with a custom name, it will also be prefixed with `wp-` which can be confusing. Maybe we shouldn't hardcode this prefix here and let metrics decide which name to use in the header?

load.php
mukeshpanchal27: The deactivate hook calls before the uninstall hook, so I don't think we need to add in `uninstall.php`.

server-timing/class-perflab-server-timing.php
mukeshpanchal27: We can add filter for it? Also we can do it in next release WDYT?

server-timing/class-perflab-server-timing-metric.php
```diff
@@ -0,0 +1,140 @@
...
+			_doing_it_wrong(
+				__METHOD__,
+				/* translators: %s: PHP method name */
+				sprintf( __( 'The %s method must be called before.', 'performance-lab' ), __CLASS__ . '::measure_before()' ),
+				''
+			);
+			return;
+		}
+
+		$this->set_value( ( microtime( true ) - $this->before_value ) * 1000.0 );
```
mukeshpanchal27, line 143: Is there any specific reason to use `1000.0` instead of `1000`?

load.php
felixarntz: @eugene-manuilov I don't think that's correct. We place an `object-cache.php` file also when there is no original `object-cache.php` file. And that file still needs to be deleted when deactivating the plugin.

In order to know whether the `object-cache.php` file is _our_ `object-cache.php` file, we rely on the constant check for the one that is present in our file.

load.php
felixarntz: @eugene-manuilov Agree with @mukeshpanchal27, the file should no longer be used as soon as the plugin is no longer active. Otherwise it'll keep loading the Server-Timing code from the plugin even when the plugin is not active.

server-timing/class-perflab-server-timing-metric.php
felixarntz: @eugene-manuilov I wanted to be strict here and require these values to be _actual_ numbers, not a numeric string. Since this is a new API, I think it's good to start with a stricter baseline to avoid issues in the future.

IMO passing a numeric string in pretty much any situation is not a good idea and should be discouraged by the API. But let me know if you disagree.

server-timing/class-perflab-server-timing.php
felixarntz: @eugene-manuilov The idea is that the `wp-` prefix will apply to all metrics added by WordPress, even those added by third party plugins (e.g. there could be something like `wp-jetpack-...` or `wp-site-kit-...`). The purpose of the prefix is to differentiate from other Server-Timing metrics (e.g. the server environment itself can add its own metrics and those shouldn't be confused with the metrics that WordPress adds.

server-timing/class-perflab-server-timing-metric.php
felixarntz: @mukeshpanchal27 I think this is just me trying to make sure the resulting value is a float. I still get confused on whether if we put an integer `1000` here the value would still be a float. If you are 100% sure this isn't needed, I will change it, but I have run into too many issues in the past to be confident, so `1000.0` seems safer IMO :)

mukeshpanchal27: Make sense. thanks for the information.

mukeshpanchal27 (approved):
Thanks, @felixarntz for all your answers. looks good to me.

server-timing/class-perflab-server-timing.php
eugene-manuilov: Ok, got it. Makes sense to me. Thanks.

server-timing/class-perflab-server-timing-metric.php
eugene-manuilov: PHP is loosely typed language and it's pretty common situation when some numbers come as strings. So, I think it will just add more problems than benefits because people won't expect strict types checking here.

load.php
eugene-manuilov: Ok, got it. Thanks.

server-timing/class-perflab-server-timing-metric.php
felixarntz: @eugene-manuilov Fair point. I've updated this in https://github.com/WordPress/performance/pull/553/commits/569637dcca8249a7c2dd9228905bb5e3eaf764b4.

eugene-manuilov: Thanks, @felixarntz. Looks good to me.

eugene-manuilov (approved):
LGTM. Great work, @felixarntz!

server-timing/class-perflab-server-timing.php
```diff
@@ -0,0 +1,265 @@
...
+		$this->registered_metrics_data[ $metric_slug ] = $args;
+
+		// If the current user has already been determined and they lack the necessary access,
+		// do not even attempt to calculate the metric.
+		if ( did_action( 'set_current_user' ) && ! current_user_can( $args['access_cap'] ) ) {
+			return;
+		}
+
+		// Otherwise, call the measuring callback and pass the metric instance to it.
+		call_user_func( $args['measure_callback'], $this->registered_metrics[ $metric_slug ] );
```
westonruter, line 108: I'm curious why this method doesn't rather `return` the `$this->registered_metrics[ $metric_slug ]` value here? Why is `measure_callback` needed?

Then instead of:

https://github.com/WordPress/performance/blob/0e10d5a0a924bd1d35345587eabb2984814c502d/server-timing/load.php#L105-L115

It could be simplified to:

```php
		if ( ! perflab_server_timing()->has_registered_metric( $metric_slug ) ) {
			$server_timing_metric = perflab_server_timing_register_metric(
				$metric_slug,
				array(
					'access_cap' => $access_cap,
				)
			);
		}
```

felixarntz: @westonruter I think that scenario is an edge-case, which is only needed because of the way this function is a wrapper to an action callback.

Overall, I consider that a discouraged pattern. Passing a measurement callback allows for that callback to only be called on other conditions (e.g. depending on the "access cap"). If the metric was returned, it would encourage implementing the measurement logic there, which then would be decoupled from those conditions for whether the metric should be measured in the first place. Alternatively, we could return the metric only if the conditions are satisfied, but that would be a strange pattern IMO, as not returning it wouldn't mean that the metric wasn't registered correctly - whereas the return value of a "register" function would be expected to only return `null` or `false` if the registration itself failed.

PR #556, opened 2022-10-13
Add Web Worker Offloading module

### LINKED ISSUE #176: Partytown support

thelovekesh:
## Summary

Fixes #176

**Please check #271 for previous conversations.**

## Relevant technical choices
- Add Web Worker module which allows running scripts within the worker thread with the help of the Partytown library.
- Add the Partytown package to the dependencies
- Add npm script to copy Partytown lib files to module `assets/js` folder.
    - Currently, Partytown lib files are committed to the module.
    - We can also set up webpack to copy partytown js files in the module build folder when creating the release of the plugin.
    - Partytown library needs to be updated when a new version is released.
- Add tests for the Web Worker Module.
- Add `partytown_configuration` filter which can be used to update partytown [configurations](https://partytown.builder.io/configuration).

## Screencast
https://user-images.githubusercontent.com/54371619/159474078-d88fa7f7-293f-4347-90f4-4ab0d4881240.mp4

## Checklist


.github/workflows/deploy-dotorg.yml
```diff
@@ -12,12 +12,28 @@ jobs:
...
+          cache: npm
+
+      - name: Install Node production dependencies
+        run: npm ci --production
+        env:
+          CI: true
+
+      - name: Add Partytown library files to release
+        run: npm run partytown
+
```
thelovekesh: > https://github.com/WordPress/performance/pull/271#discussion_r993678322

@adamsilverstein Please check I have moved the process of copying partytown files to the release process. 

Since the plugin doesn't have its build system I think it's the best possible place to hook this functionality.

modules/javascript/partytown-web-worker/load.php
```diff
@@ -0,0 +1,113 @@
...
+			},
+			10,
+			3
+		);
+	}
+}
+add_action( 'wp_print_scripts', 'perflab_web_worker_partytown_worker_scripts' );
+
+/**
+ * Helper function to get all scripts tags which has `partytown` dependency.
```
adamsilverstein: Rather than using a partytown dependency, can we look for scripts that have a `partytown-worker` strategy (and maybe no dependencies)? 

That way partytown could be enabled with:
`wp_script_add_data( $handle, 'strategy', 'partytown-worker' );`

modules/javascript/partytown-web-worker/load.php
```diff
@@ -0,0 +1,113 @@
...
+ * Get all scripts tags which have a `partytown` dependency.
+ *
+ * @since 1.0.0
+ * @return void
+ */
+function perflab_web_worker_partytown_worker_scripts() {
+	$partytown_handles = perflab_get_partytown_handles();
+
+	foreach ( $partytown_handles as $partytown_handle ) {
+		add_filter(
```
adamsilverstein: adding a filter for each partytown enabled handle seems excessive - it should be a bit cleaner and more efficient to add the filter **once**, then check if the handle is in the list of partytown handles.

modules/javascript/partytown-web-worker/load.php
```diff
@@ -0,0 +1,113 @@
+<?php
+/**
+ * Module Name: Partytown Web Worker
+ * Description: Run JavaScript in a separate Web Worker with the help of Partytown library.
+ * Experimental: Yes
+ *
+ * @since   1.0.0
```
adamsilverstein: I believe all of the "@ since" annotations can use n.e.x.t to be transformed to the proper version on release.

modules/javascript/partytown-web-worker/load.php
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * PartyTown Configuration
+ *
+ * @since 1.0.0
+ * @see https://partytown.builder.io/configuration
+ * @return void
+ */
+function perflab_web_worker_partytown_configuration() {
+	$config = array(
+		'lib'     => str_replace( site_url(), '', plugin_dir_url( __FILE__ ) ) . 'assets/js/partytown/',
```
adamsilverstein: Does partytown require https? if so we should check that the site url does as well

adamsilverstein:
@thelovekesh apologies for dropping my attention here: I finally got back to this and left some feedback on the PR; I'd love to help get this over the finish line to land as a module.

thelovekesh:
Thanks @adamsilverstein. I will try to resolve the feedback and update you once done.

.github/workflows/deploy-dotorg.yml
adamsilverstein: possibly this could now be added to part of the `npm run build-plugins` command, what do you think?

modules/javascript/partytown-web-worker/load.php
```diff
@@ -0,0 +1,113 @@
...
+	 * @param array $config Configuration for PartyTown.
+	 * @return array
+	 */
+	$config = apply_filters( 'perflab_partytown_configuration', $config );
+
+	?>
+	<script>
+		window.partytown = <?php echo wp_json_encode( $config ); ?>;
+	</script>
+	<?php
```
westonruter:
```suggestion
	wp_print_inline_script_tag( sprintf( 'window.partytown = %s;', wp_json_encode( $config ) ) );
```

westonruter: Or alternatively, this would be even better added as an inline script for the `partytown` script.

modules/javascript/partytown-web-worker/load.php
```diff
@@ -0,0 +1,113 @@
...
+			 * @return string $tag Script tag with type="text/partytown".
+			 */
+			function( $tag, $handle, $src ) use ( $partytown_handle ) {
+				if ( $handle === $partytown_handle ) {
+					$create_script_tag = sprintf(
+						'<script type="text/partytown" src="%1s" id="%2s"></script>',
+						$src,
+						$handle . '-js'
+					);
+					$tag               = $create_script_tag;
```
westonruter:
```suggestion
					$tag = wp_get_script_tag( array( 'type' => 'text/partytown', 'src' => $src, 'id' => "{$handle}-js" ) );
```

.github/workflows/deploy-dotorg.yml
westonruter: Having a build process is also needed for Image Loading Optimization. Referenced in #869 pointing to https://github.com/WordPress/performance/pull/876#discussion_r1380553929. We need to use a local copy of web-vitals.js.

modules/javascript/partytown-web-worker/load.php
westonruter: Yes, it would, since service workers require HTTPS. That is, unless the site is `localhost` in which browsers will it like HTTPS. 

Partytown does not require service workers, but they can be easier to deploy. The better approach is using [Atomics](https://partytown.builder.io/atomics), but it has other requirements with COOP/COEP.

modules/javascript/partytown-web-worker/load.php
```diff
@@ -0,0 +1,113 @@
...
+				}
+				return $tag;
+			},
+			10,
+			3
+		);
+	}
+}
+add_action( 'wp_print_scripts', 'perflab_web_worker_partytown_worker_scripts' );
```
westonruter: The `wp_print_scripts` action can fire multiple times on the page, such as when someone manually calls `wp_print_scripts( $handles )`. Probably better to do:

```suggestion
add_action( 'wp_print_footer_scripts', 'perflab_web_worker_partytown_worker_scripts' );
```

modules/javascript/partytown-web-worker/load.php
thelovekesh: Updated the logic to hook only single filter.

modules/javascript/partytown-web-worker/load.php
thelovekesh: Updated

modules/javascript/partytown-web-worker/load.php
thelovekesh: Now adding it as an inline script on `partytown` handle.

modules/javascript/partytown-web-worker/load.php
thelovekesh: After removing the dependence from the `wp_print_scripts`, the `script_loader_tag` works. I'm not sure why I initially included this.

modules/javascript/partytown-web-worker/load.php
thelovekesh: While it's feasible to execute this, my inclination is towards considering `partytown` to be included in the `$deps` array as a dependency. 

This preference stems from the fact that when a script loads as `script[type="text/partytown"]`, it necessitates the presence of `partytown`. Thus, referring to it as a direct dependency seems logical.

 Additionally, by adhering to the dependency approach, merely registering the `partytown` script suffices, as WordPress automatically enqueues it if another script depends on it else it will not be enqueued to the front-end.

modules/javascript/partytown-web-worker/load.php
thelovekesh: Humm, that is a good case to consider. In that case, would it be better to disable the module itself? if the site is not using SSL with a notice like:
```txt
The partytown web worker module only works if the site is SSL enabled.
```

.github/workflows/deploy-dotorg.yml
thelovekesh: @westonruter @adamsilverstein Added a CLI command called `copy-vendor-libs` in https://github.com/WordPress/performance/pull/556/commits/e6a51cecf3e716b2a263838b970e4b6574bebeb5.

Tasks it will perform:
- Copy vendor files from source to destination based on `copy-vendor-libs` config in `package.json`.
- Copy files from source to destination based on command args i.e. `./bin/plugin/cli.js copy-vendor-libs --source foo/bar/lib --destination modules/module`
- Run after executing the `npm install` command (prepare hook).
- Run with the `build-plugins` command, just before creating the standalone plugin builds.

admin/load.php
```diff
@@ -210,6 +210,9 @@ function perflab_get_focus_areas() {
 		'images'       => array(
 			'name' => __( 'Images', 'performance-lab' ),
 		),
+		'javascript'   => array(
```
thelovekesh: Maybe this is not required. Should we include the `partytown` module in the `js-and-css` focus area?

modules/javascript/partytown-web-worker/load.php
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Initialize PartyTown
+ *
+ * @since n.e.x.t
+ * @return void
+ */
+function perflab_partytown_web_worker_init() {
+	$partytown_js = __DIR__ . '/assets/js/partytown/partytown.js';
+
+	if ( file_exists( $partytown_js ) ) {
```
westonruter: If the file doesn't exist, shouldn't this do `wp_die()` instead of add an empty string? This file not existing indicates a development issue which wouldn't occur in production.

package.json
```diff
@@ -38,7 +41,8 @@
     "pretest-php-multisite": "wp-env run composer 'install --no-interaction'",
     "test-php-multisite": "wp-env run phpunit 'WP_MULTISITE=1 phpunit -c /var/www/html/wp-content/plugins/$(basename $(pwd))/tests/multisite.xml --verbose'",
     "wp-env": "wp-env",
-    "prepare": "husky install"
+    "prepare": "husky install && npm run copy-vendor-libs",
+    "copy-vendor-libs": "node ./bin/plugin/cli.js copy-vendor-libs"
```
westonruter: Note: Partytown also has a built-in Copy Task command which perhaps could be reused? https://partytown.builder.io/copy-library-files

modules/javascript/partytown-web-worker/load.php
westonruter: Yes, that makes sense to me, if not `localhost`.

package.json
thelovekesh: > Partytown also has a built-in Copy Task command which perhaps could be reused?

If we will be having our build system(most probably we will be having that for the ILO module), that wouldn't be required.

modules/javascript/partytown-web-worker/load.php
thelovekesh: That makes sense. and maybe with a notice?

westonruter: A notice in `wp_die()`? Ideally we wouldn't include the `file_exists()` check in the build of the plugin, since we wouldn't need to keep checking for it all the time as it would be guaranteed to be there. The condition could be stripped out for distribution. This is something that could be done in [`build-plugins`](https://github.com/WordPress/performance/blob/0e7f28516e452e107fbaf66a3a548c274f230878/bin/plugin/commands/build-plugins.js#L67-L94), although that is for standalone _plugins_ and not modules. In any case, I don't think that this should necessarily hold up this PR. There should just be some error communicated that the user hasn't done a build.

package.json
westonruter: Yeah, you're probably right.

modules/javascript/partytown-web-worker/load.php
thelovekesh: Added in https://github.com/WordPress/performance/pull/556/commits/4191312440cca374e0c377f8b277acacc887ab5e

modules/javascript/partytown-web-worker/load.php
thelovekesh: Agree since FS checks are costly. To address this:
- Add logic to not load Partytown module at all if vendor files does not exists and add admin notice - https://github.com/WordPress/performance/pull/556/commits/8c76f52159f0b386dc4e026e9dc2080f3eae99b9
- Remove file exists check for Partytown library - https://github.com/WordPress/performance/pull/556/commits/16a466c7f5181fda8f3ff3af2139ffe821fc8b6b

modules/javascript/partytown-web-worker/can-load.php
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @since   n.e.x.t
+ * @package partytown-web-worker
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+if ( ! function_exists( 'perflab_partytown_web_worker_can_load' ) ) {
```
westonruter: Why are these pluggable functions?

modules/javascript/partytown-web-worker/can-load.php
```diff
@@ -0,0 +1,99 @@
...
+			);
+		}
+
+		// Check for HTTPS.
+		if (
+			! is_ssl()
+			&&
+			( strpos( get_bloginfo( 'wpurl' ), 'https' ) !== 0 )
+			&&
+			( strpos( get_bloginfo( 'url' ), 'https' ) !== 0 )
```
westonruter: Service workers normally don't work on HTTPS, but the exception is on `localhost` for development. Also passing `url` is just a wrapper for `home_url()`:

```suggestion
			( wp_parse_url( home_url(), PHP_URL_SCHEME ) !== 'https' )
			&&
			'localhost' !== wp_parse_url( home_url(), PHP_URL_HOST )
```

modules/javascript/partytown-web-worker/can-load.php
```diff
@@ -0,0 +1,99 @@
...
+					)
+				)
+			);
+		}
+
+		// Check for HTTPS.
+		if (
+			! is_ssl()
+			&&
+			( strpos( get_bloginfo( 'wpurl' ), 'https' ) !== 0 )
```
westonruter: Passing `wpurl` is just an alias for `site_url()`:

```suggestion
			( wp_parse_url( site_url(), PHP_URL_SCHEME ) !== 'https' )
```

admin/load.php
westonruter: Yeah, let's include as part of `js-and-css`

modules/javascript/partytown-web-worker/can-load.php
westonruter: Nevertheless, I think probably the `is_ssl()` and `localhost` checks would be sufficient?

modules/javascript/partytown-web-worker/load.php
adamsilverstein: That makes sense, however I can also see the use case for the `wp_script_add_data` approach. For example, an optimization plugin could use this to specify it wants a specific script loaded with partytown. I'm not sure you could add a dependency dynamically like that (I have to look). 

This is fine as is for now, I will check if maybe we can make it work both ways.

adamsilverstein:
Looks really good, I'm going to try to test this out a bit locally.

thelovekesh:
Note: Partytown lib assets can be copied using the Webpack which landed recently in the repo.
See https://github.com/WordPress/performance/pull/953

module-i18n.php
```diff
@@ -5,6 +5,8 @@
 	_x( 'Adds support to store the dominant color of newly uploaded images and create a placeholder background of that color.', 'module description', 'performance-lab' ),
 	_x( 'WebP Support Health Check', 'module name', 'performance-lab' ),
 	_x( 'Adds a WebP support check in Site Health status.', 'module description', 'performance-lab' ),
+	_x( 'Partytown Web Worker', 'module name', 'performance-lab' ),
+	_x( 'Run JavaScript in a separate Web Worker with the help of Partytown library.', 'module description', 'performance-lab' ),
```
adamsilverstein:
```suggestion
	_x( 'Add support for running JavaScript in a separate Web Worker with the help of the Partytown library.', 'module description', 'performance-lab' ),
```

thelovekesh: I think it's an auto-generated file. Better to update this in the plugin header?

westonruter: Wouldn't it be in both? Curious why `module-i18n.php` is committed to git in the first place. Seems like it should be part of the build process only? cc @mukeshpanchal27

modules/javascript/partytown-web-worker/can-load.php
thelovekesh: Fixed in https://github.com/WordPress/performance/pull/556/commits/ac2c9d0d5b900969699c35979a5b4afabb2aae97

modules/javascript/partytown-web-worker/can-load.php
thelovekesh: Updated in https://github.com/WordPress/performance/pull/556/commits/ac2c9d0d5b900969699c35979a5b4afabb2aae97

.github/workflows/deploy-dotorg.yml
thelovekesh: We have removed the above logic and now using Webpack to copy the assets. See https://github.com/WordPress/performance/pull/556/commits/5dfcfdfee96f090cd18889dbc6bc9b3a68d875ac and https://github.com/WordPress/performance/pull/953

admin/load.php
thelovekesh: Done in https://github.com/WordPress/performance/pull/556/commits/335daaa1ace427486f7b2023963ffeda72f6befc

.github/CODEOWNERS
```diff
@@ -48,6 +48,8 @@
 /modules/js-and-css/audit-enqueued-assets                                   @manuelRod
 /tests/modules/js-and-css/audit-enqueued-assets                             @manuelRod
 /tests/testdata/modules/js-and-css/audit-enqueued-assets                    @manuelRod
+/modules/js-and-css/partytown-web-worker                                    @thelovekesh
+/tests/modules/js-and-css/partytown-web-worker                              @thelovekesh
```
swissspidy, line 52: Note that these would only work when given write access to the repo

modules/js-and-css/partytown-web-worker/load.php
```diff
@@ -0,0 +1,139 @@
+<?php
+/**
+ * Module Name: Partytown Web Worker
```
swissspidy, line 3: Do we really want to call the module Partytown?

IMHO usage of the Partytown library is just an implementation detail.

westonruter: Agreed. A less-specific name would be "Web Worker Offloading". I asked Bard for some alternative names, and it suggested "Threadshift". I like that, but it is trademarked.

westonruter:
Since we're currently in the middle of re-plumbing for [standalone plugins](https://github.com/WordPress/performance/issues?q=is%3Aissue+is%3Aopen+label%3A%22Creating+standalone+plugins%22), I'm going to re-base this branch to a ~`feature/partytown`~ `feature/web-worker-offloading` branch. I don't think this feature should be released as a module but rather as a standalone plugin, right? cc @mukeshpanchal27 @joemcgill

modules/js-and-css/partytown-web-worker/load.php
thelovekesh: That makes sense. If you have a certain name in mind, do let me know, and I will edit it.

westonruter: How about Web Worker Offloading for now?

thelovekesh: SGTM, some more names:
- Scripts offloading/offloader (in the description add a reference to Web Worker and/or Partytown).
- Third-party JS offloader
- Non-critical JS offloader

I will defer name selection to you.

westonruter: Those are some good options but I like Web Worker Offloading.

The JS being offloaded may not be third-party. It may also be critical, but just offloaded so it can run in another thread.

.gitignore
```diff
@@ -33,6 +34,7 @@ build-cs/composer.lock
 node_modules/
 vendor/
 
+modules/js-and-css/partytown-web-worker/assets/js/partytown
```
westonruter, line 37: This will necessitate introducing a `.distignore` file to replace the `.gitattributes` file as explained in https://github.com/WordPress/performance/issues/893#issuecomment-1947398713.

module-i18n.php
```diff
@@ -9,6 +9,8 @@
 	_x( 'Creates WebP versions for new JPEG image uploads if supported by the server.', 'module description', 'performance-lab' ),
 	_x( 'Enqueued Assets Health Check', 'module name', 'performance-lab' ),
 	_x( 'Adds a CSS and JS resource check in Site Health status.', 'module description', 'performance-lab' ),
+	_x( 'Partytown Web Worker', 'module name', 'performance-lab' ),
```
westonruter, line 12:
```suggestion
	_x( 'Web Worker Offloading', 'module name', 'performance-lab' ),
```

modules/js-and-css/partytown-web-worker/load.php
westonruter:
```suggestion
 * Module Name: Web Worker Offloading
```

tests/modules/js-and-css/partytown-web-worker/partytown-web-worker-test.php
```diff
@@ -0,0 +1,139 @@
+<?php
+/**
+ * Test cases for Partytown Web Worker module.
```
westonruter, line 3:
```suggestion
 * Test cases for Web Worker Offloading module.
```

modules/js-and-css/partytown-web-worker/load.php
```diff
@@ -0,0 +1,139 @@
...
+ * @since n.e.x.t
+ *
+ * @return array Array of script handles with `partytown` dependency.
+ */
+function perflab_get_partytown_handles() {
+	global $wp_scripts;
+
+	$partytown_handles = array();
+	foreach ( $wp_scripts->registered as $handle => $script ) {
+		if ( ! empty( $script->deps ) && in_array( 'partytown', $script->deps, true ) ) {
```
westonruter, line 88: Correct me if I'm wrong, but let's say you have an existing script in a theme like:

```php
wp_enqueue_script( 'foo', 'https://example.com/foo.js' );
```

In order to opt-in this script to use Partytown, you'd have to do this if you wanted to opt-in from a separate plugin:

```php
wp_scripts()->registered['foo']->deps[] = 'partytown';
```

Is that right?

westonruter (commented):
Since this would certainly get released as a standalone plugin on Wordpress.org, it seems we'd need to have the plugin actually do something without needing to code. Otherwise, I understand that the plugin review team would reject the plugin as a "framework plugin". I recall there was some discussion of what this could do by default, without there being a clear answer. Would there need to be a UI to select scripts to opt-in? Humm.

adamsilverstein:
> Since this would certainly get released as a standalone plugin on Wordpress.org, it seems we'd need to have the plugin actually do something without needing to code 

We could have it automatically apply to analytics scripts where it is expected to work well - although those may not be enqueued with wp_scripts.

modules/js-and-css/partytown-web-worker/load.php
thelovekesh: Yup, that should do the trick.

thelovekesh:
> We could have it automatically apply to analytics scripts where it is expected to work well - although those may not be enqueued with wp_scripts.

If the user wishes to offload these [common scripts](https://partytown.builder.io/common-services), it would be preferable if there was an opt-in form on some sort of options page. Alternatively, the user may utilize this as a framework to offload scripts.

There are two scenarios I can think of if the user opts in for script auto-offloading:
- They should be enqueued with wp_scripts, and we could use something like `wp_scripts()->registered['foo']->deps[] = 'partytown';` 
- Alternatively, if any script tags are inline scripts added to the page's head or footer, we will need to parse every script tag on the page and offload the above-mentioned common scripts if available before the output buffer.

westonruter:
Just noting this can be updated to use the new standalone plugin structure.

westonruter (approved):
Approving to continue work in the feature branch

adamsilverstein (approved):
Looking good! Lets keep iterating and testing in the branch.

PR #565, opened 2022-10-20
Fix opt-in checkbox for generating WebP and JPEG to also show on Multisite

### LINKED ISSUE #560: WebP upload configuration checkbox not showing in Settings > Media for multisite
## Bug Description

The checkbox introduced in #525 does not show up when using Multisite. This is because the entire settings section (group `media`, section `uploads`) is not rendered for multisites. We have to fix this, which I think could be done in two alternative ways:
* Either we keep control about this per site in the multisite, thus just conditionally render it in another settings section on the same page when using a multisite (e.g. the `default` section).
* Or we only grant control about this at the network level, thus add the checkbox to the Network Admin's settings screen.

## Steps to reproduce

1. Use a Multisite.
2. Ensure the WebP Uploads module is active.
3. Go to Settings > Media, and notice that the checkbox to opt in to WebP _and_ JPEG generation is not there.

## Screenshots

**Settings > Media when using a multisite**
Screen Shot 2022-10-17 at 10 35 17 AM

mehulkaklotar:
## Summary

Fixes #560 

## Relevant technical choices

PR adds checkbox setting for WebP and JPEG in Settings > Media in the default section in the multisite for each site.

Screenshot 2022-10-20 at 6 58 21 PM

## Checklist

### DIFF
--- a/modules/images/webp-uploads/settings.php
+++ b/modules/images/webp-uploads/settings.php
@@ -33,10 +33,10 @@ function webp_uploads_add_media_settings_field() {
 	// Add settings field.
 	add_settings_field(
 		'perflab_generate_webp_and_jpeg',
-		__( 'Generate WebP and JPEG', 'performance-lab' ),
+		__( 'WebP and JPEG', 'performance-lab' ),
 		'webp_uploads_generate_webp_jpeg_setting_callback',
 		'media',
-		'uploads',
+		! is_multisite() ? 'uploads' : 'default',
 		array( 'class' => 'perflab-generate-webp-and-jpeg' )
 	);
 }
@@ -50,7 +50,7 @@ add_action( 'admin_init', 'webp_uploads_add_media_settings_field' );
 function webp_uploads_generate_webp_jpeg_setting_callback() {
 	?>
 	</td>
-	<td class="td-full">
+	<td <?php echo ! is_multisite() ? esc_html( 'class="td-full"' ) : ''; ?>>
 		<label for="perflab_generate_webp_and_jpeg">
 			<input name="perflab_generate_webp_and_jpeg" type="checkbox" id="perflab_generate_webp_and_jpeg" aria-describedby="perflab_generate_webp_and_jpeg_description" value="1"<?php checked( '1', get_option( 'perflab_generate_webp_and_jpeg' ) ); ?> />
 			<?php esc_html_e( 'Generate JPEG files in addition to WebP', 'performance-lab' ); ?>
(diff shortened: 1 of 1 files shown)


modules/images/webp-uploads/settings.php
```diff
@@ -67,10 +67,19 @@ function webp_uploads_generate_webp_jpeg_setting_callback() {
...
+			display: none;
+		}
+		<?php else : ?>
+		.form-table .perflab-generate-webp-and-jpeg > th {
+			font-size: 16px;
+		}
+		.form-table .perflab-generate-webp-and-jpeg > td:nth-child(2) {
+			display: none;
+		}
+		<?php endif; ?>
```
mukeshpanchal27: Should check `if ( is_multisite() ) :` instead of `if ( ! is_multisite() ) :` and change the CSS respectively

modules/images/webp-uploads/settings.php
```diff
@@ -33,10 +33,10 @@ function webp_uploads_add_media_settings_field() {
 	// Add settings field.
 	add_settings_field(
 		'perflab_generate_webp_and_jpeg',
-		__( 'Generate WebP and JPEG', 'performance-lab' ),
+		__( 'WebP and JPEG', 'performance-lab' ),
 		'webp_uploads_generate_webp_jpeg_setting_callback',
 		'media',
-		'uploads',
+		! is_multisite() ? 'uploads' : 'default',
```
felixarntz: For better readability, let's reverse this.
```suggestion
		is_multisite() ? 'default' : 'uploads',
```

modules/images/webp-uploads/settings.php
```diff
@@ -50,7 +50,7 @@ function webp_uploads_add_media_settings_field() {
 function webp_uploads_generate_webp_jpeg_setting_callback() {
 	?>
 	</td>
-	<td class="td-full">
+	<td class="<?php echo ! is_multisite() ? esc_attr( 'td-full' ) : ''; ?>">
```
felixarntz: Instead of this hack, we should simply not render the entire `</td><td class="td-full">` markup when using multisite so that it fits into the default layout out of the box. This way all those CSS hacks below aren't needed either, and we just render the field like it is by default.

modules/images/webp-uploads/settings.php
```diff
@@ -67,10 +67,19 @@ function webp_uploads_generate_webp_jpeg_setting_callback() {
...
+		<?php else : ?>
+		.form-table .perflab-generate-webp-and-jpeg > th {
+			font-size: 16px;
+		}
+		.form-table .perflab-generate-webp-and-jpeg > td:nth-child(2) {
+			display: none;
+		}
+		<?php endif; ?>
 	</style>
 	<?php
```
felixarntz: Related to my above comment: Instead of all this CSS "hack", we should simply rely on the default markup (without the extra `td`) when in multisite. In other words, the only change to this function should be to return early (without rendering any CSS) when in multisite.

felixarntz (changes requested):
@mehulkaklotar Left a few points of feedback, this can be simplied quite a bit.

modules/images/webp-uploads/settings.php
mehulkaklotar: @felixarntz Makes sense. I have added these changes in the latest commit.

felixarntz (approved):
Thanks @mehulkaklotar, looks great!

ankitrox (approved):
💯

PR #568, opened 2022-10-28
Fix fatal error in REST API response when an image has no attachment metadata.

### LINKED ISSUE #567: Improve error handling: Cannot use object of type stdClass as array
## Bug Description

[One user in the support forums ](https://wordpress.org/support/topic/php-error-cannot-use-object-of-type-stdclass-as-array/) reported the below error:

Uncaught Error: Cannot use object of type stdClass as array in /wp-content/plugins/performance-lab/modules/images/webp-uploads/rest-api.php:21 Stack trace: 

```
#0 /wp-includes/class-wp-hook.php(307): webp_uploads_update_rest_attachment() 
#1 /wp-includes/plugin.php(191): WP_Hook->apply_filters() 
#2 /wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php(843): apply_filters() 
#3 /wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php(377): WP_REST_Attachments_Controller->prepare_item_for_response() 
#4 /wp-includes/rest-api/class-wp-rest-server.php(1143): WP_REST_Posts_Controller->get_items() 
#5 /mnt/data/home/502433.cloudw
```

Awaiting feedback from impacted user. No further details at present. 

## Additional Context

 - PHP Version: 
 - Browser: [e.g. chrome, safari]
 - Plugin Version: [e.g. 22]

mehulkaklotar:
## Summary
This bug is very unlikely to happen as it depends on attachment metadata. It is assumed that metadata should always be present for the attachment but when metadata is deleted from DB, we should ensure proper error handling by checking the data from REST.

Fixes #567 

## Relevant technical choices

Added more checks for `media_details` in response of REST API for attachments

## Checklist

### DIFF
--- a/modules/images/webp-uploads/rest-api.php
+++ b/modules/images/webp-uploads/rest-api.php
@@ -18,7 +18,7 @@
  */
 function webp_uploads_update_rest_attachment( WP_REST_Response $response, WP_Post $post, WP_REST_Request $request ) {
 	$data = $response->get_data();
-	if ( ! isset( $data['media_details']['sizes'] ) || ! is_array( $data['media_details']['sizes'] ) ) {
+	if ( ! isset( $data['media_details'] ) || ! is_array( $data['media_details'] ) || ! isset( $data['media_details']['sizes'] ) || ! is_array( $data['media_details']['sizes'] ) ) {
 		return $response;
 	}
 
--- a/tests/modules/images/webp-uploads/rest-api-tests.php
+++ b/tests/modules/images/webp-uploads/rest-api-tests.php
@@ -71,4 +71,39 @@ class WebP_Uploads_REST_API_Tests extends WP_UnitTestCase {
 		$this->assertArrayNotHasKey( 'sources', $data['media_details'] );
 	}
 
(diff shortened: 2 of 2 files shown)


felixarntz (commented):
@mehulkaklotar Fix generally looks reasonable, but please add a unit test for the bug, so that we can verify it is fixed.

In the test, you can use the `rest_prepare_attachment` to e.g. make `$data['media_details']` an object. With that test, we should then verify whether the old code would fail vs the new code would pass.

mehulkaklotar:
@felixarntz I have added the new unit test for checking `media_details` in the attachment response data in the latest commit. Please review it once.

I have used the REST controller's `get_item` and deleted the attachment metadata before to get the `media_details` as an object because `rest_prepare_attachment` doesn't have control over changing the data in the response object in the test.

felixarntz (approved):
@mehulkaklotar Thank you, the test looks excellent!

adamsilverstein (approved):
Great!

PR #569, opened 2022-10-28
Added the benchmark command

eugene-manuilov:
## Summary

Added the benchmarking command that allows us to run load tests and get an average response time for an URL. In addition to the average response time, this command can also collect the `Server-Timing` metrics and shows an average time for each metric that is returned in the response.

To run it use the following command:

```
npm run benchmark -- -c 5 -n 100 -u https://example.com/
```

This command will send `n` requests to the `u` page with `c` concurrency.

We can also use the `-f, --file <file>` argument to use URLs from a file and the `-o, --output <output>` argument to output results in the CSV format. For example, the following command will benchmark URLs from file and output results in the CSV format:

```
npm run benchmark -- -c 5 -n 100 -f urls.txt -o csv
```

Whereas the `urls.txt` file is just a simple list of URLs:

```
https://example.com/pageA/
https://example.com/pageB/
https://mysite/blog/
...
```

## Checklist


bin/plugin/commands/benchmark.js
```diff
@@ -0,0 +1,231 @@
...
+/**
+ * Runs http benchmarks for an URL.
+ *
+ * @param {BenchmarkCommandOptions} opt Command options.
+ */
+exports.handler = async ( opt ) => {
+	const { concurrency: connections, requests: amount } = opt;
+	const results = [];
+
+	for await ( const url of getURLs( opt ) ) {
```
felixarntz, line 77: Given the `url` option is required, could you add validation here similar to what we have for other CLI commands? If the user doesn't provide any `url`, we should bail early and show an error (using `log( formats.error( ... ) )`).

bin/plugin/commands/benchmark.js
```diff
@@ -0,0 +1,231 @@
...
+		argname: '-u, --url <url>',
+		description: 'An URL to run benchmark tests for',
+	},
+	{
+		argname: '-c, --concurrency <concurrency>',
+		description: 'Number of multiple requests to make at a time',
+		defaults: 1,
+	},
+	{
+		argname: '-n, --requests <requests>',
```
felixarntz: A bit odd to have the shorthand `n` but the long version `requests`. How about we make this `number`?

bin/plugin/commands/benchmark.js
```diff
@@ -0,0 +1,231 @@
...
+	} );
+
+	for ( let i = 0; i < len; i++ ) {
+		const [ url, responseTimes, metrics ] = results[ i ];
+
+		tableData[ 0 ][ i + 1 ] = url;
+		tableData[ 1 ][ i + 1 ] = floor( mean( responseTimes ), 2 );
+
+		for ( const metric of Object.keys( metrics ) ) {
+			const metricAvgMs = floor( mean( metrics[ metric ] ), 2 );
```
felixarntz: We should be using median values for performance metrics from multiple runs, rather than averages.

Also, I would suggest we use `round` rather than `floor`.

bin/plugin/commands/benchmark.js
```diff
@@ -0,0 +1,231 @@
...
+	const onStop = instance.stop.bind( instance );
+	process.once( 'SIGINT', onStop );
+
+	return new Promise( ( resolve ) => {
+		instance.on( 'done', () => {
+			process.off( 'SIGINT', onStop );
+			resolve( { responseTimes, metrics } );
+		} );
+	} );
+}
```
felixarntz, line 169: There's currently no error handling in this function. When calling this e.g. with a URL that doesn't exist (or basically any more realistic error that could happen), the table still gets a column for that URL contain just an `NaN` average time. It would be better to catch request failures or 400/500 responses and maybe log warnings for those.

I would say results for a URL where any errors occurred should not be included in the table (because if e.g. 2 out of 9 requests fail, the data would be inconsistent compared to the other URLs), and the warning should indicate that the user should run the check for those again or investigate what went wrong.

bin/plugin/commands/benchmark.js
```diff
@@ -0,0 +1,231 @@
...
+	const len = results.length;
+	const allMetricNames = {};
+
+	const newRow = ( title ) => {
+		const line = new Array( len + 1 ).fill( '' );
+		line[ 0 ] = title;
+		return line;
+	};
+
+	const tableData = [ newRow( '' ), newRow( 'Avg Time' ) ];
```
felixarntz: Maybe use `Response Time` instead? The average is used for every metric, so that's a bit pointless to mention here (plus see my comment below that we should be using median though).

bin/plugin/commands/benchmark.js
```diff
@@ -0,0 +1,231 @@
...
+ *
+ * @property {Array.<number>} responseTimes An array of response times.
+ * @property {Object}         metrics       The Server-Timing metrics object.
+ */
+
+/**
+ * @typedef BenchmarkCommandOptions
+ *
+ * @property {string} url         An URL to benchmark.
+ * @property {number} concurrency A number of multiple requests to make at a time.
```
felixarntz, line 37: What is the purpose of this? Does it just speed up the process? Or is it to simulate traffic in some way? What would be the recommendation for how to use this option?

felixarntz (commented):
@eugene-manuilov This looks excellent so far and works well. I left comments below on a few things that we can improve.

bin/plugin/commands/benchmark.js
eugene-manuilov: Yes, this is to simulate traffic. It says how many requests should be sent simultaneously to replicate real life concurrency when multiple people access the site at the same time. It's up to us to decide which number of concurrent requests we want to use. We can go with just `1` request at time, or maybe `5` to simulate more real life results.

bin/plugin/commands/benchmark.js
eugene-manuilov: Added

bin/plugin/commands/benchmark.js
eugene-manuilov: Sounds good to me, updated.

bin/plugin/commands/benchmark.js
eugene-manuilov: Ok, updated.

bin/plugin/commands/benchmark.js
eugene-manuilov: Updated.

bin/plugin/commands/benchmark.js
eugene-manuilov: > ... It would be better to catch request failures or 400/500 responses and maybe log warnings for those ...
> ... results for a URL where any errors occurred should not be included in the table ...

I don't think that the status code really matters for us. This tool is intended to benchmark the response time, not the correct behaviour of the site. If the site returns an error code, it still spends some time to process the request, right? Anyway we are going to use it for controlled set of pages where errors are not expected 🤓 

Added the completion rate to the results table, just in case something goes wrong suddenly.

felixarntz: That sounds good, thanks!

bin/plugin/commands/benchmark.js
```diff
@@ -0,0 +1,278 @@
...
+			}
+		);
+
+		results.push( [ url, completeRequests, responseTimes, metrics ] );
+	}
+
+	if ( results.length === 0 ) {
+		log(
+			formats.error(
+				'No URLs were found. Please, provide URLs to benchmark.'
```
felixarntz: It would be helpful to include the concrete parameters here which makes this command very intuitive to use (trial and error, you wouldn't even need to read the code or look at help).
```suggestion
				'You need to provide a URL to benchmark via the --url (-u) argument, or a file with multiple URLs via the --file (-f) argument.'
```

bin/plugin/commands/benchmark.js
```diff
@@ -0,0 +1,278 @@
...
+
+	const newRow = ( title ) => {
+		const line = new Array( len + 1 ).fill( '' );
+		line[ 0 ] = title;
+		return line;
+	};
+
+	const tableData = [
+		newRow( '' ),
+		newRow( 'Complete Requests' ),
```
felixarntz: Maybe better "Success Rate" (since the value is a percentage)?

felixarntz (approved):
@eugene-manuilov Two nit-picks here which would be great to address, but overall consider this approved already - works great!

bin/plugin/commands/benchmark.js
eugene-manuilov: Yep, makes sense. Updated.

mehulkaklotar (approved):
@eugene-manuilov Looks good to me! 🚀

PR #571, opened 2022-11-01
Change WP Image editor quality for mime types

### LINKED ISSUE #563: Change default WebP quality to 82
### Feature Description
In https://github.com/WordPress/performance/issues/7 and the related Trac ticket (https://core.trac.wordpress.org/ticket/54356) we have been researching the best default quality settings for WebP images. While results aren't conclusive, I propose testing a quality of 82 in ther plugin to gather additional (subjective) feedback.

### Current Status
Currently, when uploading WebP images or using the `image_editor_output_format` filter to output WebP images, or using the Performance Lab plugin to output WebPs, the default quality set for WebP images is 86 (compared to the default quality setting used for JPEG of 82). The 86 quality setting was chosen to try to achieve the same visual quality for WebP images (compared to JPEGs). 

### Desired Outcome
Users get the smallest possible images without any **noticeable** degradation in quality. WebP was designed to meet this need: the quality setting is designed to match JPEG, providing a similar quality with a (much) smaller file size.

### Challenges in 6.0
The filter used to control the default quality setting  `wp_editor_set_quality` had a bug until WordPress 6.1 that resulted in the incorrect mime type being passed to the filter. Because of this, setting quality by mime type doesn't work correctly until 6.1. However, we can use the filter to _always return/set quality to 82 for all mime types_.

### Proposed solution
I propose setting the default quality for WebP images to 82. For versions before 6.1, we should set all mime types to 82, for 6.1+ we can base the quality on mime type and return 82 for WebP

mehulkaklotar:
## Summary

Fixes #563 

## Relevant technical choices

For WP 6.1 below, all mime types for images, quality is set to 82
For WP 6.1+, all mime types for images should be default as per this [ticket](https://core.trac.wordpress.org/ticket/54356) and [PR](https://github.com/WordPress/wordpress-develop/pull/3115) and WebP to 82

## Checklist


modules/images/webp-uploads/load.php
```diff
@@ -742,3 +742,33 @@ function webp_uploads_get_image_sizes_additional_mime_type_support() {
...
+ */
+function webp_uploads_set_wp_image_editor_quality( $quality, $mime_type ) {
+	global $wp_version;
+
+	$valid_mime_transforms = webp_uploads_get_upload_image_mime_transforms();
+
+	// Return early for non image or invalid mime types.
+	if ( ! isset( $valid_mime_transforms[ $mime_type ] ) ) {
+		return $quality;
+	}
```
felixarntz: This should not be relevant here. We don't really care if the MIME type should be transformed or not, this filter is simply to update the WebP image quality.

modules/images/webp-uploads/load.php
```diff
@@ -742,3 +742,33 @@ function webp_uploads_get_image_sizes_additional_mime_type_support() {
 
 	return $allowed_sizes;
 }
+
+/**
+ * Updates the quality of images used in the wp editor.
+ * For WP version less than 6.2 returns 82,
+ * Otherwise returns default quality.
```
felixarntz: This is not what the goal of this PR is, see #563:
> I propose setting the default quality for WebP images to 82. For versions before 6.1, we should set all mime types to 82, for 6.1+ we can base the quality on mime type and return 82 for WebP

The reason to make the differentiation between 6.1+ and pre-6.1 is the fix in https://core.trac.wordpress.org/changeset/54417 which was part of the 6.1 release: Prior to 6.1, we can't rely on the `$mime_type` filter parameter to be correct.

modules/images/webp-uploads/load.php
```diff
@@ -742,3 +742,33 @@ function webp_uploads_get_image_sizes_additional_mime_type_support() {
...
+	if ( ! isset( $valid_mime_transforms[ $mime_type ] ) ) {
+		return $quality;
+	}
+
+	if ( version_compare( '6.2', $wp_version, '>=' ) ) {
+		return $quality;
+	}
+
+	// Default quality set to 82 for all mimes for WP version less than 6.2.
+	return 82;
```
felixarntz: The code here is not doing what the comment says:
* The code returns the default quality for 6.2 and below, and 82 for a version greater than 6.2.
* The comment says to return the default quality for 6.2 and above, and 82 for a version lower than 6.2.

In any case, please review the solution proposed in the issue:
* Below 6.1, we should universally return 82.
* In 6.1+, we should return 82 only for `image/webp`, and the default quality otherwise.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -775,6 +775,7 @@ public function it_should_create_webp_for_full_size_which_is_smaller_in_webp_for
 		$this->opt_in_to_jpeg_and_webp();
 
 		add_filter( 'webp_uploads_discard_larger_generated_images', '__return_true' );
+		remove_all_filters( 'wp_editor_set_quality' );
```
felixarntz: Why is the new filter being removed here?

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -801,6 +802,7 @@ public function it_should_create_webp_for_some_sizes_which_are_smaller_in_webp_f
...
```
felixarntz: See above, why is this necessary?

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -918,11 +920,63 @@ public function it_should_create_mime_types_for_allowed_sizes_only_via_global_va
...
+		// Quality setting for the source image. For JPG the fallback default of 82 is used.
+		$this->assertSame( 82, $editor->get_quality(), 'Default quality setting for JPG is 82.' );
+
+		// A JPG image will be converted to WEBP whose quality should be 82 for wp version 6.2 and above.
+		$editor->save();
+		$this->assertSame( $expected_webp, $editor->get_quality(), 'Output image format is WEBP. Quality setting for it should be 82 for WP version 6.2 and above.' );
+
+		unset( $editor );
+		remove_filter( 'image_editor_output_format', array( $this, 'image_editor_output_formats' ) );
+	}
```
felixarntz, line 986: The logic in this test also needs to be updated based on my above comments, as it's currently not testing the right thing.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -918,11 +920,63 @@ public function it_should_create_mime_types_for_allowed_sizes_only_via_global_va
...
+	 * @test
+	 */
+	public function test_set_quality_with_image_conversion() {
+		global $wp_version;
+		// Set conversions for uploaded images.
+		add_filter( 'image_editor_output_format', array( $this, 'image_editor_output_formats' ) );
+		$expected_webp = 82;
+		if ( version_compare( '6.2', $wp_version ) ) {
+			$expected_webp = 86;
+		}
```
felixarntz: See above, this is not right: The WebP quality after this fix should _always_ be 82. The only difference is in how we fix this in the filter, depending on whether it's WP <6.1 or WP >=6.1.

felixarntz (changes requested):
@mehulkaklotar Unfortunately it looks like this PR does not address the issue correctly. Please see my comments below.

modules/images/webp-uploads/load.php
```diff
@@ -742,3 +742,33 @@ function webp_uploads_get_image_sizes_additional_mime_type_support() {
 
 	return $allowed_sizes;
 }
+
+/**
+ * Updates the quality of images used in the wp editor.
+ * For WP version less than 6.2 returns 82,
```
mukeshpanchal27:
```suggestion
 * Updates the quality of images used in the wp editor.
 *
 * For WP version less than 6.2 returns 82,
```

modules/images/webp-uploads/load.php
```diff
@@ -742,3 +742,33 @@ function webp_uploads_get_image_sizes_additional_mime_type_support() {
...
+ * Otherwise returns default quality.
+ *
+ * @since n.e.x.t
+ *
```
mukeshpanchal27, line 753:
```suggestion
 *
 * @global string $wp_version The WordPress version string.
 *
```

Add missing doc for `global`

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -918,11 +920,63 @@ public function it_should_create_mime_types_for_allowed_sizes_only_via_global_va
 		$this->assertImageNotHasSizeSource( $attachment_id, 'not_allowed_size_200x150', 'image/webp' );
 	}
 
+	/**
+	 * Test image quality for image conversion.
+	 *
+	 * @test
+	 */
+	public function test_set_quality_with_image_conversion() {
```
mukeshpanchal27: Remove `test_` from the function name.
```suggestion
	public function it_should_set_quality_with_image_conversion() {
```

@mehulkaklotar We have two options for running a test.

- At starting of function name use `test_`
- Or you can use the `@test` annotation in a method’s docblock.

mukeshpanchal27 (commented):
@mehulkaklotar Left some minor feedbacks

modules/images/webp-uploads/load.php
mehulkaklotar: Make sense! I have updated the code to reflect this ^ Please have a look once again.

modules/images/webp-uploads/load.php
mehulkaklotar: Right, I understood it opposite. I have changed it now.

modules/images/webp-uploads/load.php
mehulkaklotar: New changes reflect these suggestions. Please have a look.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: New image quality for WebP is actually generating smaller image size then jpeg image in this test case. So we are testing it if full image has webp and other sizes doesn't have webp image in sources. But with new quality, it generates sources for all image sizes. so this case fails. That is why I have removed the quality filter from here as well as from the test case below.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: same case as above, webp images generated is smaller in size.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: Updated the logic to reflect the accurate changes in the test case as well.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: Updated the logic to reflect the changes in the test case. Please review once.

tests/modules/images/webp-uploads/load-tests.php
aaemnnosttv: Should this perhaps be removed on a per-test basis instead? It seems odd that we would add a filter as part of the module, only to disable it for all of the module's tests. Is that not the case here?

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -918,6 +920,41 @@ public function it_should_create_mime_types_for_allowed_sizes_only_via_global_va
...
+		$this->assertSame( 82, $editor->get_quality(), 'Default quality setting for PNG is 82.' );
+
+		$file = wp_tempnam();
+
+		// A PNG image will be converted to WebP whose quality should be 82 universally.
+		$editor->save( $file, 'image/webp' );
+		$this->assertSame( 82, $editor->get_quality(), 'Output image format is WebP. Quality setting for it should be 82 universally.' );
+
+		unlink( $file );
+		unset( $editor );
```
aaemnnosttv: Cleanup like this should happen in a `tearDown` method as it will not run if an assertion fails before it as part of the test body. These references can be stored in a property that is processed later in the tear down method. You might also consider splitting into separate test methods if such cleanup is needed before each "inner test".

modules/images/webp-uploads/load.php
```diff
@@ -742,3 +742,29 @@ function webp_uploads_get_image_sizes_additional_mime_type_support() {
 
 	return $allowed_sizes;
 }
+
+/**
+ * Updates the quality of images used in the wp editor.
+ * For WP version less than 6.1 or WebP returns 82,
+ * Otherwise returns default quality.
```
```suggestion
 * Updates the quality of WebP image sizes generated by WordPress to 82.
 *
 * Prior to WordPress 6.1, the MIME type was not correctly provided to this filter, so for those versions this function
 * simply returns 82 universally (since all other MIME types already use 82 anyway).
```

modules/images/webp-uploads/load.php
```diff
@@ -742,3 +742,29 @@ function webp_uploads_get_image_sizes_additional_mime_type_support() {
...
+ *
+ * @since n.e.x.t
+ *
+ * @global string $wp_version The WordPress version string.
+ *
+ * @param int    $quality   Quality level between 1 (low) and 100 (high).
+ * @param string $mime_type Image mime type.
+ * @return int The updated quality for mime types.
+ */
+function webp_uploads_set_wp_image_editor_quality( $quality, $mime_type ) {
```
felixarntz: A bit more precise name: This function is focused on changing WebP image quality, not any image quality.
```suggestion
function webp_uploads_modify_webp_quality( $quality, $mime_type ) {
```

modules/images/webp-uploads/load.php
```diff
@@ -742,3 +742,29 @@ function webp_uploads_get_image_sizes_additional_mime_type_support() {
...
+	global $wp_version;
+
+	// Below WP 6.1 or WebP, always return 82.
+	if ( version_compare( $wp_version, '6.1', '<' ) || 'image/webp' === $mime_type ) {
+		return 82;
+	}
+
+	// Return default quality for WP 6.1+.
```
felixarntz: This comment is misleading: The code doesn't always return the default quality for WP 6.1+, it only returns the default quality if it's WP 6.1+ _and_ the image is not WebP.
```suggestion
	// Return default quality for non-WebP images in WP 6.1+.
```

modules/images/webp-uploads/load.php
```diff
@@ -742,3 +742,29 @@ function webp_uploads_get_image_sizes_additional_mime_type_support() {
...
+ * @global string $wp_version The WordPress version string.
+ *
+ * @param int    $quality   Quality level between 1 (low) and 100 (high).
+ * @param string $mime_type Image mime type.
+ * @return int The updated quality for mime types.
+ */
+function webp_uploads_set_wp_image_editor_quality( $quality, $mime_type ) {
+	global $wp_version;
+
+	// Below WP 6.1 or WebP, always return 82.
```
```suggestion
	// Below WP 6.1 or for WebP images, always return 82 (other MIME types were already using 82 by default anyway).
```

modules/images/webp-uploads/load.php
```diff
@@ -742,3 +742,29 @@ function webp_uploads_get_image_sizes_additional_mime_type_support() {
...
+	if ( version_compare( $wp_version, '6.1', '<' ) || 'image/webp' === $mime_type ) {
+		return 82;
+	}
+
+	// Return default quality for WP 6.1+.
+	return $quality;
+}
+add_filter( 'wp_editor_set_quality', 'webp_uploads_set_wp_image_editor_quality', 10, 2 );
```
felixarntz: See above
```suggestion
add_filter( 'wp_editor_set_quality', 'webp_uploads_modify_webp_quality', 10, 2 );
```

tests/modules/images/webp-uploads/load-tests.php
felixarntz: @aaemnnosttv Isn't the filter already being removed on a per-test basis here? I think that looks reasonable to me.

However: @mehulkaklotar Instead of relying on whatever quality is set by default, I think we should rather solve this in a more reliable way: Instead of removing the filter and relying on whatever WP core provides, why don't we add a custom filter in these tests instead, which sets the quality for all `image/webp` MIME type images to e.g. 90? That should ensure that no matter what core's behavior is, the WebP images generated here will indeed be larger.

tests/modules/images/webp-uploads/load-tests.php
felixarntz: See above on how I think we can solve this in a more reliable way that doesn't assume as much about the defaults.

tests/modules/images/webp-uploads/load-tests.php
felixarntz: Why is this unsetting of variables even needed? They are just rewritten right below. It seems to me that these two lines could simply be removed.

tests/utils/TestCase/DominantColorTestCase.php
```diff
@@ -24,7 +24,7 @@ public function provider_get_dominant_color() {
 			),
 			'green_jpg'     => array(
 				'image_path'            => TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/dominant-color/green.jpg',
-				'expected_color'        => array( '00ff00', '00ff01' ),
+				'expected_color'        => array( '00ff00', '00ff01', '02ff01' ),
```
felixarntz, line 27: Why is this in here? Seems unrelated? If there's another PR that caused breakage here, I think we should fix this separately.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -918,6 +920,41 @@ public function it_should_create_mime_types_for_allowed_sizes_only_via_global_va
 		$this->assertImageNotHasSizeSource( $attachment_id, 'not_allowed_size_200x150', 'image/webp' );
 	}
 
+	/**
+	 * Test image quality for image conversion.
+	 *
+	 * @test
+	 */
+	public function it_should_set_quality_with_image_conversion() {
```
felixarntz, line 946: This test is useful for an integration test, however it's almost impossible to validate here that the logic in the underlying function `webp_uploads_modify_webp_quality()` is correct. I would suggest we add another (simpler) test that calls `webp_uploads_modify_webp_quality()` directly, passing a specific default and MIME type to it so that we can really cover the internals of that function and that they work as expected.

felixarntz (commented):
@mehulkaklotar This looks good now in terms of production code. Left a few additional suggestions on wording and tests.

tests/modules/images/webp-uploads/load-tests.php
aaemnnosttv: > Isn't the filter already being removed on a per-test basis here? I think that looks reasonable to me.

Ah yes, my mistake – I had misread the context of the diff 👍 

> Instead of removing the filter and relying on whatever WP core provides, why don't we add a custom filter in these tests instead, which sets the quality for all `image/webp` MIME type images to e.g. 90? That should ensure that no matter what core's behavior is, the WebP images generated here will indeed be larger.

This sounds better to me as well, although should we perhaps also filter the quality for jpg images to be lower as well to make the difference explicit and potentially even larger?

tests/modules/images/webp-uploads/load-tests.php
aaemnnosttv: The unset for `$file` shouldn't be needed, but my guess regarding the `$editor` was to invoke `\WP_Image_Editor_Imagick::__destruct` or another classes' destructor. I'm not sure if this would be called immediately otherwise just by overwriting the variable. Again, I don't think it would be necessary with separate tests but if it were for some reason, it should get a comment to explain that, otherwise it doesn't make much sense 👍

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: I have pushed new simple test to test `webp_uploads_modify_webp_quality()` function. It is actually really simple just calling the function with quality and check whether it return correct quality for mimes or not.

tests/utils/TestCase/DominantColorTestCase.php
mehulkaklotar: Actually, some of the dominant color tests were checking this green color, but with quality changes, it seems that `02ff01` is another green variant being created in tests for the jpg images.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: I have added a new filter for these tests to return WebP quality to 86. Actually with 90, it was still failing. so I am just sticking with 86 for WebP in these tests.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: unset was not required there. Unlink is still required for temporary file that is being generated with image editor save, but that is also needed at the end of test case. I have removed the in between unset and unlink code now.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -918,11 +924,66 @@ public function it_should_create_mime_types_for_allowed_sizes_only_via_global_va
...
 	 */
 	private function mock_frontend_body_hooks() {
 		remove_all_actions( 'template_redirect' );
 		do_action( 'template_redirect' );
 	}
+
+	/**
+	 * Return WebP image quality 86 for testing.
+	 */
+	public function return_webp_image_quality( $quality, $mime_type ) {
```
felixarntz: Let's call this something like `force_high_webp_quality`. The current name doesn't really describe the purpose of this filter method.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -918,11 +924,66 @@ public function it_should_create_mime_types_for_allowed_sizes_only_via_global_va
...
+		if ( 'image/webp' === $mime_type ) {
+			return 86;
```
felixarntz, line 1001: Maybe use something even higher, to make it even more safe that the created WebP image is larger than the JPEG. How about `90`?

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -789,6 +790,8 @@ public function it_should_create_webp_for_full_size_which_is_smaller_in_webp_for
 			$this->assertImageNotHasSizeSource( $attachment_id, $size, 'image/webp' );
 		}
 		$this->assertNotSame( $tag, webp_uploads_img_tag_update_mime_type( $tag, 'the_content', $attachment_id ) );
+
+		remove_filter( 'wp_editor_set_quality', array( $this, 'return_webp_image_quality' ), PHP_INT_MAX );
```
felixarntz: Please remove. WP's test suite removes filters from test methods automatically. Also, this wouldn't work reliably, if the assertion fails for example.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -821,6 +825,8 @@ public function it_should_create_webp_for_some_sizes_which_are_smaller_in_webp_f
 
 		$this->assertNotSame( $tag, $updated_tag );
 		$this->assertSame( $expected_tag, $updated_tag );
+
+		remove_filter( 'wp_editor_set_quality', array( $this, 'return_webp_image_quality' ), PHP_INT_MAX );
```
felixarntz: See above, please remove.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -918,11 +924,66 @@ public function it_should_create_mime_types_for_allowed_sizes_only_via_global_va
...
+	 *
+	 * @covers ::webp_uploads_modify_webp_quality()
+	 *
+	 * @test
+	 */
+	public function it_should_return_correct_quality_for_mime_types() {
+		$this->assertSame( 82, webp_uploads_modify_webp_quality( 90, 'image/webp' ), 'WebP image quality should always be 82.' );
+		$this->assertSame( 82, webp_uploads_modify_webp_quality( 82, 'image/webp' ), 'WebP image quality should always be 82.' );
+		$this->assertSame( 80, webp_uploads_modify_webp_quality( 80, 'image/jpeg' ), 'JPEG image quality should return default quality provided from WP filter wp_editor_set_quality.' );
+		$this->assertNotSame( 50, webp_uploads_modify_webp_quality( 40, 'image/jpeg' ), 'JPEG image quality should return default quality provided from WP filter wp_editor_set_quality.' );
```
felixarntz: This assertion is not really useful, since the assertion above already satisfies this scenario in a more precise way. I suggest you remove this one.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -918,11 +924,66 @@ public function it_should_create_mime_types_for_allowed_sizes_only_via_global_va
...
+	 * Test webp_uploads_modify_webp_quality function for image quality.
+	 *
+	 * @covers ::webp_uploads_modify_webp_quality()
+	 *
+	 * @test
+	 */
+	public function it_should_return_correct_quality_for_mime_types() {
+		$this->assertSame( 82, webp_uploads_modify_webp_quality( 90, 'image/webp' ), 'WebP image quality should always be 82.' );
+		$this->assertSame( 82, webp_uploads_modify_webp_quality( 82, 'image/webp' ), 'WebP image quality should always be 82.' );
+		$this->assertSame( 80, webp_uploads_modify_webp_quality( 80, 'image/jpeg' ), 'JPEG image quality should return default quality provided from WP filter wp_editor_set_quality.' );
```
felixarntz: This assertion will only pass when using WordPress 6.1+. While we currently only run CI tests with the latest WordPress version, let's make this test resilient to also work e.g. with WordPress 6.0 (since the Performance Lab plugin supports that version as well).

In other words, we need to have an if condition here for the `image/jpeg` case:
* If WordPress version is lower than 6.1, the expected value should be 82, no matter what you pass.
* Otherwise, the expected value should be the same that is passed, like what you have here.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -918,11 +924,66 @@ public function it_should_create_mime_types_for_allowed_sizes_only_via_global_va
...
+	 * @test
+	 */
+	public function it_should_set_quality_with_image_conversion() {
+		$editor = wp_get_image_editor( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/dice.png', array( 'mime_type' => 'image/png' ) );
+
+		// Quality setting for the source image. For PNG the fallback default of 82 is used.
+		$this->assertSame( 82, $editor->get_quality(), 'Default quality setting for PNG is 82.' );
+
+		// Temporary file.
+		$file = wp_tempnam();
```
felixarntz: Why do we need to create this file here? I would suggest we simply pass a file name or potentially `null` to `WP_Image_Editor::save()`.

Creating a temporary file just for this seems unnecessary.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -918,11 +924,66 @@ public function it_should_create_mime_types_for_allowed_sizes_only_via_global_va
...
+
+		// Quality setting for the source image. For JPG the fallback default of 82 is used.
+		$this->assertSame( 82, $editor->get_quality(), 'Default quality setting for JPG is 82.' );
+
+		// A JPG image will be converted to WebP whose quality should be 82 universally.
+		$editor->save( $file, 'image/webp' );
+		$this->assertSame( 82, $editor->get_quality(), 'Output image format is WebP. Quality setting for it should be 82 universally.' );
+
+		// Delete the temporary file.
+		unlink( $file );
```
felixarntz: This code will not run if any of the above assertion fails. See my comment above though, I think we should not use a temporary file anyway, so I think the code here should be removed.

felixarntz (commented):
@mehulkaklotar Left a few follow up comments. Once these have been addressed, this should be good to go.

tests/modules/images/webp-uploads/load-tests.php
aaemnnosttv: Per https://github.com/WordPress/performance/pull/571#discussion_r1019633576 how about we also filter the quality of jpg (with another method) to ensure webp will always be larger?

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: @felixarntz @aaemnnosttv We don't need to change quality for all the test cases here. Actually, the two test cases which are failing are both special/different.

1. **it_should_create_webp_for_full_size_which_is_smaller_in_webp_format** - this test case requires only full size image to generate WebP image for sources, and all other sub sizes should not generate WebP sizes. so when we pass 90 as WebP quality it will generate WebP for all and it will fail there. so we need 86 here for WebP for full size image only.
2. **it_should_create_webp_for_some_sizes_which_are_smaller_in_webp_format** - this test case requires sub sizes to generate WebP image as sub sources, and full size will not generate WebP. so when we pass 90 as WebP quality it will generate WebP for all and it will fail there. so we need 86 here for WebP for sub sizes image only.

It is a kind of special case where we can't just force WebP to any higher quality as the test image used there is specifically used there.

I have made some specific changes to only force return WebP quality for these two test cases.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: See my comment above for this ^

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: Removed!

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: Removed!

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: Removed now!

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: Added changes for WP version here. Please have a look.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: When we save the editor, it generates the image for WebP for that mime and it is generated in our testdata folder.
Screenshot 2022-11-15 at 4 20 38 PM

so to avoid generating these files here better to generate them in temp folder and even if unlink code doesn't work if assertions fails, temp folder will take of itself, I guess.

tests/modules/images/webp-uploads/load-tests.php
mehulkaklotar: see my comment above about test files.

tests/modules/images/webp-uploads/load-tests.php
adamsilverstein: Also don't understand why this is needed here

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -918,11 +920,71 @@ public function it_should_create_mime_types_for_allowed_sizes_only_via_global_va
...
+		$editor = wp_get_image_editor( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/dice.png', array( 'mime_type' => 'image/png' ) );
+
+		// Quality setting for the source image. For PNG the fallback default of 82 is used.
+		$this->assertSame( 82, $editor->get_quality(), 'Default quality setting for PNG is 82.' );
+
+		// Temporary file.
+		$file = wp_tempnam();
+
+		// A PNG image will be converted to WebP whose quality should be 82 universally.
+		$editor->save( $file, 'image/webp' );
```
adamsilverstein, line 956: `$file` can be null here and save will generate a filename, returning the saved file which you can then use for deleting.

tests/modules/images/webp-uploads/load-tests.php
```diff
@@ -918,11 +920,71 @@ public function it_should_create_mime_types_for_allowed_sizes_only_via_global_va
...
+		$this->assertSame( 82, $editor->get_quality(), 'Output image format is WebP. Quality setting for it should be 82 universally.' );
+
+		$editor = wp_get_image_editor( TESTS_PLUGIN_DIR . '/tests/testdata/modules/images/leafs.jpg' );
+
+		// Quality setting for the source image. For JPG the fallback default of 82 is used.
+		$this->assertSame( 82, $editor->get_quality(), 'Default quality setting for JPG is 82.' );
+
+		// A JPG image will be converted to WebP whose quality should be 82 universally.
+		$editor->save( $file, 'image/webp' );
```
adamsilverstein, line 965: same as above

tests/utils/TestCase/DominantColorTestCase.php
adamsilverstein: are all three colors still expected here? it seems like if this changed one of the other colors would go away.

tests/modules/images/webp-uploads/load-tests.php
felixarntz: I talked about this with @aaemnnosttv, makes sense to me now. The point is that this function generates a unique filename in a temporary directory. If we don't do this, the files are generated within our repository directory, which is not advisable as then they may by accident end up in our repository.

WordPress core uses a similar mechanism for some of its tests, then also unlinking the file, so that seems good to me now. 👍

tests/modules/images/webp-uploads/load-tests.php
felixarntz: @adamsilverstein `$file` should always be a string here since `wp_tempnam()` always returns a string.

tests/modules/images/webp-uploads/load-tests.php
felixarntz: See above.

tests/modules/images/webp-uploads/load-tests.php
felixarntz: @mehulkaklotar I understand now why you're using `wp_tempnam()`, that makes sense. However, the comment here still needs to be addressed: If an assertion fails, this line is never reached, which means it would leave the file in place.

We need to change this as follows:
* Assign the results from any method calls from assertions here to variables (_before_ the `unlink` call).
* Run all assertions only _after_ the `unlink` call.

This way, when an assertion fails, at least the file was still deleted as usual.

felixarntz (commented):
@mehulkaklotar One last thing here, then this should be good to go.

tests/modules/images/webp-uploads/load-tests.php
aaemnnosttv: @felixarntz Another alternative could be to store the file name in a property and clean it up in a tear down.

<details>
<summary>Example class</summary>

```php
class TestCase { // could also be implemented as a trait

	protected $to_unlink = [];
	
	public function test_something_with_a_file() {
		$filename = $this->temp_filename();
		
		file_put_contents( $filename, 'foobar' );
		
		// $filename will always be deleted after the test
	}

	protected function temp_filename() {
		$filename = wp_tempnam();

		$this->to_unlink[] = $filename;

		return $filename;
	}

	public function tear_down() {
		$this->to_unlink = array_filter(
			$this->to_unlink,
			function ( $filename ) {
				return unlink( $filename );
			}
		);
	}

}
```

</details>

mehulkaklotar:
@felixarntz I have used the tear down method instead to unlink the file now. Please have a look.

felixarntz (approved):
@mehulkaklotar Thank you, LGTM!

adamsilverstein (approved):
Nice work!

PR #566, opened 2022-10-21
Introduce database focus area, rename JavaScript focus area to JS & CSS, and phase out Site Health focus area

mukeshpanchal27:
## Summary

PR that updates focus areas in the codebase
- Update perflab_get_focus_areas() to return the latest focus areas
- Rename javascript focus area to css-and-js (CSS & JS), including the directory in modules
- Move site-health modules into their new focus area directories
- Implement logic in perflab_get_module_settings() to migrate all old ("legacy") module slugs to the new ones (e.g. if you have site-health/audit-autoloaded-options in the DB, it should become database/audit-autoloaded-options)

Fixes #554

## Relevant technical choices
To fix the unit test you have to update `default-enabled-modules.php` using https://make.wordpress.org/performance/handbook/performance-lab/releasing-the-plugin/#update-list-of-default-enabled-modules.

## Checklist

- [ ] PR has either `[Focus]` or `Infrastructure` label.
- [ ] PR has a `[Type]` label.
- [ ] PR has a milestone or the `no milestone` label.

PR #572, opened 2022-11-02
Load the WP_Image_Editor_Imagick class if it doesn't exist

aristath:
## Summary

Recently I noticed that some tests are failing. I'm _assuming_ (though not sure) that it's because of the WP 6.1 release, at least the timing lines-up. 
Example: https://github.com/WordPress/performance/actions/runs/3374928930/jobs/5601071400
The problem seems to be the fact that the `WP_Image_Editor_Imagick` class is not loaded.

## Relevant technical choices

Added a `class_exists` check before the class in tests, and I'm just require_once the missing files.

## Checklist

### DIFF
--- a/modules/images/dominant-color/class-dominant-color-image-editor-gd.php
+++ b/modules/images/dominant-color/class-dominant-color-image-editor-gd.php
@@ -9,6 +9,11 @@
  * @since 1.2.0
  */
 
+if ( ! class_exists( 'WP_Image_Editor_GD' ) ) {
+	require_once ABSPATH . WPINC . '/class-wp-image-editor.php';
+	require_once ABSPATH . WPINC . '/class-wp-image-editor-gd.php';
+}
+
 /**
  * WordPress Image Editor Class for Image Manipulation through GD
  * with dominant color detection.
--- a/modules/images/dominant-color/class-dominant-color-image-editor-gd.php
+++ b/modules/images/dominant-color/class-dominant-color-image-editor-gd.php
@@ -9,11 +9,6 @@
  * @since 1.2.0
  */
 
-if ( ! class_exists( 'WP_Image_Editor_GD' ) ) {
-	require_once ABSPATH . WPINC . '/class-wp-image-editor.php';
-	require_once ABSPATH . WPINC . '/class-wp-image-editor-gd.php';
-}
-
 /**
  * WordPress Image Editor Class for Image Manipulation through GD
  * with dominant color detection.
--- a/tests/admin/load-tests.php
+++ b/tests/admin/load-tests.php
@@ -68,7 +68,7 @@ class Admin_Load_Tests extends WP_UnitTestCase {
 		remove_all_filters( 'plugin_action_links_' . plugin_basename( PERFLAB_MAIN_FILE ) );
 
 		// Rely on current user to be an administrator (with 'manage_options' capability).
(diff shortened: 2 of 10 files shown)


mukeshpanchal27:
Thanks @aristath for reporting. earlier today i face the same issue https://github.com/WordPress/performance/pull/566#issuecomment-1299597466

felixarntz:
@aristath @mukeshpanchal27 I think we should dig into the WP 6.1 release to identify why that is just failing now. The fix here is probably good to go, but IMO we should verify that it is indeed the appropriate solution and that we aren't missing something else it could be caused by.

mukeshpanchal27:
Some highlight of the changes that i made on this PR.

- [9b48e3f](https://github.com/WordPress/performance/pull/572/commits/9b48e3f0d5887ac34eb08458db72b62afdfd4dbf) - This adds new keys added to `add_settings_section()` in [[54247]](https://core.trac.wordpress.org/changeset/54247).
- [3ec7d40](https://github.com/WordPress/performance/pull/572/commits/3ec7d40ebba04c02d3ace8858f3537b8bfee96f6) - This replaces `$this->factory` and `$this->factory()` references with `self::factory()`.
See [[54087]](https://core.trac.wordpress.org/changeset/54087). `$this->factory()` is fine, but I used `self::factory()` as that's the change applied in the changeset mentioned.

Props to @costdev for assisting and resolving this problem. PR now ready for review.

costdev (approved):
Just reviewed that the four additional changes I found to resolve the failures were applied. LGTM! :+1:

modules/images/dominant-color/class-dominant-color-image-editor-gd.php
```diff
@@ -9,6 +9,11 @@
  * @since 1.2.0
  */
 
+if ( ! class_exists( 'WP_Image_Editor_GD' ) ) {
+	require_once ABSPATH . WPINC . '/class-wp-image-editor.php';
+	require_once ABSPATH . WPINC . '/class-wp-image-editor-gd.php';
+}
+
```
costdev: Upon further testing, this change shouldn't be needed.

@mukeshpanchal27 Can you remove this change locally and confirm the tests still pass?

If the tests still pass, you can go ahead and revert [3ec7d40](https://github.com/WordPress/performance/pull/572/commits/e76b654aa4314101eb52393acc64196ce5b28fe6).

mukeshpanchal27: Thanks @costdev. Reverted [3ec7d40](https://github.com/WordPress/performance/pull/572/commits/e76b654aa4314101eb52393acc64196ce5b28fe6).

aristath:
Thank you @mukeshpanchal27 and @costdev for the extra commits and the review! All tests now pass and this looks good! ❤️

mukeshpanchal27 (approved):

mehulkaklotar (approved):
@mukeshpanchal27 @costdev @aristath  Thanks for fixing this. LGTM 🚀 

Do we have any information on how the Image editor class isn't available in 6.1 and if we need to require it?

aristath:
> Do we have any information on how the Image editor class isn't available in 6.1 and if we need to require it?

I haven't seen any fatalities in real-life, so I'm inclined to believe this was more related to the tests 👍

PR #574, opened 2022-11-04
Feature: Allow updating with git-updater

batonac:
See: https://github.com/afragen/git-updater

## Summary
Adds GitHub Plugin URI to plugin header to allow for optionally tracking with the development of the plugin on GitHub and updating directly from the WordPress backend.

## Relevant technical choices
This uses the required header URI syntax as specified by [The Git Updater Plugin](https://github.com/afragen/git-updater)

### DIFF
--- a/load.php
+++ b/load.php
@@ -11,6 +11,7 @@
  * License: GPLv2 or later
  * License URI: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
  * Text Domain: performance-lab
+ * GitHub Plugin URI: WordPress/performance
  *
  * @package performance-lab
  */


mxbclang:
@felixarntz Tagging you in to take a look!

felixarntz:
@batonac What is the use-case for updating via GitHub vs WordPress.org? I am familiar with the GitHub Updater, but I see its primary value for plugins that are exclusively hosted on GitHub, which this plugin is not. Hence my question on what this would be useful for.

batonac:
Hey @felixarntz, my use-case is switching to different branches to test new/in-development feature sets. In this case, I'm looking to test out the SQLite development branch. If there's a better way to do this, I'm all ears.

felixarntz:
@batonac This is a quite specific scenario, couldn't you pull in the branch manually to test it?

I'm not necessarily opposed to adding that here, but I'm not sure it's the right move as it could falsely encourage folks to use GitHub for updates of the plugin. Additionally, this plugin is part of the WordPress organization, and looking at other repos like Gutenberg I don't see the GitHub Updater plugin header present there either.

batonac:
Fair enough @felixarntz.

I was considering this to be a low-impact change that could benefit your community testing volunteers in some scenarios. It would be nice if all these parameters could be controlled directly in git-updater, but unfortunately that doesn't appear to be possible ATM...

I'm not hard-up. Use your best judgement.

westonruter:
Unfortunately this will not work with the move to this repo becoming a monorepo of standalone plugins, especially with #1182 where the Performance Lab plugin itself will be moved into into the `plugins/performance-lab` subdirectory.

PR #573, opened 2022-11-02
Remove unnecessary leftover CSS for dominant color backgrounds on images

pbearne:
## Summary

~~Fixes #562~~

## Checklist

- [x] PR has either `[Focus]` or `Infrastructure` label.
- [x] PR has a `[Type]` label.
- [x] PR has a milestone or the `no milestone` label.

Issue #123, opened 2022-01-22
WP_Query performance for multiple taxonomies and post meta filters


khoipro:
This is only a proposal to see how could we improve a way to handle a filter for large database.

**My issues**

We need to build a large app with main purpose is real estate classified. Each post will contain some data, but mostly it will be a few terms and have some post meta.

The filters based on those taxonomies work, however the performance will be slow and slow when having 100.000 and more.

I know there is an exiting solution to build custom table and using custom SQL query to filters, but if we rely on default taxonomy route, the query seems need to destroy existing WP_Query.

So what is a best way, or any flexible way to handle that kind of performance well?

XVII:
As a thought exercise, it would potentially become more efficient to store each combination of taxonomy and give them an ID, then use that ID to identify the posts instead.

Therefore, a query would just be looking for posts with a particular ID (representing a specific combination of taxonomies, i.e. Taxonomy Set ID). Taxonomies would be responsible for maintaining the unique combination lookup table as taxonomies are added/removed.

We saw something similar used in a system that had scale issues when storing "tags" of sorts across millions of rows.

swissspidy:
Architecture support questions like this are better suited for StackOverflow. Closing this issue as there is nothing actionable here for core or Performance Lab.

PR #578, opened 2022-11-08
Avoid potentially adding invalid attributes or duplicates for dominant color images

### LINKED ISSUE #577: Revise "Dominant Color" styling and inline styling
## Bug Description

As [one user in the support forums](https://wordpress.org/support/topic/dominant-color-option-enabled-it-overwrites-style-only-on-the-frontend/#post-16172789) reported, the plugin is adding the below inline styling to images if the "_Dominants Color_" option is enabled. 
`style="--dominant-color: #332f2b;"`

One of the problems with this, is that it is overwriting any existing inline styling associated with any images. An example of this can be demonstrated in the below recording, where you'll find that my own custom inline styling is removed, in place of the PL inline styling. As soon as I disabled the "_Dominant Color_" option my custom inline styling is applied once more.  

https://user-images.githubusercontent.com/41326532/200372452-ce2b835e-2021-4167-917c-2f37ca30257f.mp4

To resolve this, consider moving away from inline styling. Printing the styling to the `wp_head` as opposed to within the `img` element also resolves this error. Tested with the below, which doesn't conflict with the "Dominant color" configuration styling:

```
function inline_css() {
  echo "<style>
  .wp-block-image img{opacity: 0.5;}
  </style>";
}
add_action( 'wp_head', 'inline_css', 0 );
```

Note also that there is [a GitHub issue](https://github.com/WordPress/gutenberg/issues/23893) in relation to not allowing themes or plugins to insert inline styling. 

## Steps to reproduce

1. Ensure you have the Performance Lab plugin active but the "Dominant Color" option disabled
2. Create a new post and switch to the code editor. 
3. Add some inline styling to an image or paste the code snippet highlighted below (_Example snippet_) into your content. Check that the styling does work as intended by visiting the front end of your site and viewing that page
4. Enable the "Dominent Color" option
5. Revisit the page on your site where you added your own inline styling
6. Refresh the page and notice that your own custom inline styling is replaced

<Details>
<summary>Example snippet</summary>

```
<!-- wp:image {"id":6632,"sizeSlug":"full","linkDestination":"none"} -->
<figure class="wp-block-image size-full"><img src="https://www.gstatic.com/webp/gallery/1.webp" alt="" class="wp-image-6632"/ style="opacity: 0.5;border:5px solid #000;border-radius:20px"></figure>
<!-- /wp:image -->
```

</details>

pbearne:
## Summary

Fixes #577

## Checklist


modules/images/dominant-color/load.php
```diff
@@ -117,7 +117,7 @@ function dominant_color_img_tag_add_dominant_color( $filtered_image, $context, $
 
 	if ( ! empty( $image_meta['dominant_color'] ) ) {
 		$data .= sprintf( 'data-dominant-color="%s" ', esc_attr( $image_meta['dominant_color'] ) );
-		$style = 'style="--dominant-color: #' . esc_attr( $image_meta['dominant_color'] ) . ';" ';
+		$style = '--dominant-color: #' . esc_attr( $image_meta['dominant_color'] ) . ';" ';
```
costdev: - Line 131 appears to add `$style` just after `style="`. Since `$style` ends with `;" `, this will close the `style` value, breaking other inline styles.
- This also produces extra spacing (presumably unintended) between `data-dominant-color` and `style`. Not a big deal, but might be unintended.

See the results of [this 3v4l](https://3v4l.org/ZX70X).

modules/images/dominant-color/load.php
```diff
@@ -126,8 +126,12 @@ function dominant_color_img_tag_add_dominant_color( $filtered_image, $context, $
 		$extra_class  = $image_meta['has_transparency'] ? 'has-transparency' : 'not-transparent';
 	}
 
-	if ( ! empty( $data ) || ! empty( $style ) ) {
-		$filtered_image = str_replace( '<img ', '<img ' . $data . $style, $filtered_image );
+	if ( ! empty( $data ) || ! empty( $style ) || ! str_contains( $data, $filtered_image ) ) {
+		if ( strpos( $filtered_image, 'style=' ) ) {
```
costdev: Suggest using `str_contains()` instead of `strpos()`.

modules/images/dominant-color/load.php
pbearne: Thanks for catching this
Fixed and remove extra space

modules/images/dominant-color/load.php
pbearne: yes swaped

modules/images/dominant-color/load.php
```diff
@@ -126,8 +126,12 @@ function dominant_color_img_tag_add_dominant_color( $filtered_image, $context, $
 		$extra_class  = $image_meta['has_transparency'] ? 'has-transparency' : 'not-transparent';
 	}
 
-	if ( ! empty( $data ) || ! empty( $style ) ) {
-		$filtered_image = str_replace( '<img ', '<img ' . $data . $style, $filtered_image );
+	if ( ! empty( $data ) || ! empty( $style ) || ! str_contains( $data, $filtered_image ) ) {
```
felixarntz: Why is the `str_contains` check needed here? That change appears to be unrelated to the fix to not duplicate the `style` attribute, which based on the PR title is the goal here.

Also, it looks incorrect: It is impossible for `$data` to ever contain `$filtered_image`, as `$data` is declared just above here. Do these two parameters have to be switched maybe? Regardless of that though, please also clarify why this is needed in the first place. I don't see what it fixes.

modules/images/dominant-color/load.php
```diff
@@ -126,8 +126,12 @@ function dominant_color_img_tag_add_dominant_color( $filtered_image, $context, $
...
+		if ( str_contains( $filtered_image, 'style=' ) ) {
+			$filtered_image = str_replace( 'style="', $data . 'style="' . $style, $filtered_image );
```
felixarntz: We need to make sure this actually looks at the `style` attribute. Right now it would also match something like `data-style`, which we don't want.
```suggestion
		if ( str_contains( $filtered_image, ' style=' ) ) {
			$filtered_image = str_replace( ' style="', ' ' . $data . 'style="' . $style, $filtered_image );
```

modules/images/dominant-color/load.php
```diff
@@ -117,7 +117,7 @@ function dominant_color_img_tag_add_dominant_color( $filtered_image, $context, $
```
felixarntz, line 127: Not strictly related to this fix, but there is a problem in this function, which is exactly similar to the bug flagged in https://core.trac.wordpress.org/ticket/56969: We must avoid adding any of these attributes to `img` tags which don't use regular double quotes.

For this reason, let's add a check for `' src="'` to the top of this function. If that is not part of the `img` string, we should bail early and not even run the filter, similar to how it's done for lazy-loading. See https://core.trac.wordpress.org/ticket/56969#comment:16.

modules/images/dominant-color/load.php
```diff
@@ -126,8 +126,12 @@ function dominant_color_img_tag_add_dominant_color( $filtered_image, $context, $
 		$extra_class  = $image_meta['has_transparency'] ? 'has-transparency' : 'not-transparent';
 	}
 
-	if ( ! empty( $data ) || ! empty( $style ) ) {
-		$filtered_image = str_replace( '<img ', '<img ' . $data . $style, $filtered_image );
+	if ( ! empty( $data ) || ! empty( $style ) || ! str_contains( $data, $filtered_image ) ) {
+		if ( str_contains( $filtered_image, 'style=' ) ) {
+			$filtered_image = str_replace( 'style="', $data . 'style="' . $style, $filtered_image );
+		} else {
+			$filtered_image = str_replace( '<img ', '<img ' . $data . ' style="' . $style . '" ', $filtered_image );
```
felixarntz: See my overarching PR comment, we should avoid adding an empty style attribute.

Also, if `$data` is not empty, it always has a space in the end already, so let's avoid double spaces here.
```suggestion
		} elseif ( ! empty( $style ) ) {
			$filtered_image = str_replace( '<img ', '<img ' . $data . 'style="' . $style . '" ', $filtered_image );
		} else {
			$filtered_image = str_replace( '<img ', '<img ' . $data, $filtered_image );
```

felixarntz (changes requested):
@pbearne While the approach for the fix makes sense to me, there are still a few flaws in the concrete logic.

One larger problem is that the logic now doesn't check whether the `$style` variable is not empty individually, which means that the `else` clause now can add a ` style=""` - we don't want that.

See related comments below.

modules/images/dominant-color/load.php
costdev: Curious about this also.

1. `str_contains( $haystack, $needle )` [Ref](https://www.php.net/manual/en/function.str-contains.php), means that this condition is looking for the filtered image markup inside `data-dominant-color="#123"`, which would be impossible. The arguments would need to be switched as Felix says.
2. Regarding the purpose, is it possible that this `str_contains()` is intended to cover a manually added `data-dominant-color: #123` attribute/value pair?

If not, then I'm not sure what the purpose of this condition is.

If so, does this PR account for a `data-dominant-color` attribute being added manually, but not added as a CSS variable in the `style` attribute? If the PR doesn't account for this, should it?

e.g.
```php
$featured_image = '<img src="hello.jpg" alt="Hello.jpg" width="100" height="100" data-dominant-color="#000" style="display: block;">';
```

pbearne: I migrated this back from the core patch 
the idea here is to stop adding the style and data twice Just in case the the_content filter is called twice

modules/images/dominant-color/load.php
pbearne: Added

modules/images/dominant-color/load.php
```diff
@@ -88,6 +88,12 @@ function dominant_color_img_tag_add_dominant_color( $filtered_image, $context, $
 		return $filtered_image;
 	}
 
+	// Only apply the dom color to images that have a src attribute that
+	// starts with a double quote, ensuring escaped JSON is also excluded.
+	if ( false === strpos( $filtered_image, ' src="' ) ) {
```
costdev: - Suggest changing to `! str_contains( $filtered_image, ' src="' )`.
- Also suggest changing "dom color" to "dominant color".

modules/images/dominant-color/load.php
felixarntz: @pbearne That isn't strictly necessary since `the_content` filter is only called once. The duplicate execution of `wp_filter_content_tags()` is not because of the `the_content`, it comes from block themes and their full page template parsing (see https://github.com/WordPress/wordpress-develop/pull/3549 for a fix).

With that said, #582 implements the addition of the attributes in a cleaner way, clearly separating the logic around the data attributes and the style attribute.

I would advise you revert all changes around the lines here and instead add a condition to the top of this function:
```php
if ( str_contains( $filtered_image, ' data-dominant-color="' ) || str_contains( $filtered_image, ' data-has-transparency="' ) ) {
	return $filtered_image;
}
```

With this logic we holistically ensure nothing is added to the image tag if this function already added its attributes before.

felixarntz (commented):
@pbearne There is some overlap between this PR and #582.

Both of the PRs have something useful in them, but one part of #582 is a cleaner solution than the one here, while the PR here covers a few more problems. Therefore I added a suggestion below to limit your PR here to avoiding the addition of invalid attributes or duplicates from calling this function multiple times on the same image.

modules/images/dominant-color/load.php
felixarntz: Would be great to add a unit test for this:
* One `img` without any `src` attribute (should not be modified).
* One `img` which has its attributes (including `src`) using single quotes instead of double quotes.

felixarntz (commented):
@pbearne One more point of feedback here.

felixarntz:
@pbearne FYI I just added that one change from the closed #573 into this PR in https://github.com/WordPress/performance/pull/578/commits/77de817af433b45bb635b3d1e51ed7602db44946.

felixarntz:
Will take over this one since it's a bug so we should get it merged today for the 1.7.0 plugin release.

felixarntz:
@costdev I've made the relevant updates here, this should be ready for another review. Would be awesome if you could get to it today so we can merge it to get the release branch ready.

felixarntz (approved):

felixarntz:
Changed the base branch here to `release/1.7.0` so that this will go into the 1.7.0 release once approved and merged.

costdev (approved):
LGTM :+1:

mukeshpanchal27 (approved):

Issue #354, opened 2022-06-02
Use dominant color as background color in media modal


spacedmonkey:
## Feature Description
Use dominant color as background color in media modal for preview thumbnail. 

<img width="1203" alt="Screenshot 2022-06-02 at 16 59 53" src="https://user-images.githubusercontent.com/237508/171674005-4ad62af5-1115-4500-971d-d695e45b0f69.png">

Follow on from https://github.com/WordPress/performance/issues/19 https://github.com/WordPress/performance/pull/282

mxbclang:
@felixarntz Looks like this needs to push to 1.4.0 as there isn't a PR, can you confirm?

pbearne:
I don't think we can add this to the performance plugin as it needs a core patch 
https://github.com/WordPress/wordpress-develop/compare/trunk...pbearne:wordpress-develop:use_dominat_color_in_media_libary?expand=1

felixarntz:
Thanks @bethanylang @pbearne. Definitely won't make it in time for 1.3.0, but if it's not feasible in the plugin anyway, let's remove the milestone entirely and keep this simply as a tracking task for the core patch work.

adamsilverstein:
> I don't think we can add this to the performance plugin as it needs a core patch
https://github.com/WordPress/wordpress-develop/compare/trunk...pbearne:wordpress-develop:use_dominat_color_in_media_libary?expand=1

Thanks for sharing this code sample @pbearne I see how adding this in core would be much easier, however I'm going to dig in a little more to see if we can add the correct classes to the template. A possible solution would be to unhook core's template and insert our own instead (`add_action( 'admin_footer', 'wp_print_media_templates' );`)

estlacksensory:
I didn't know there was a git repo for performance lab. Issue with dominant color reading transparent pixels incorrectly and then adding background color to them. Deactivated this, to an extent. Rest of performance lab seems to be working nicely though. https://github.com/estlacksensory/WP-Performance-Lab

estlacksensory:
Is this "dominant color" idea only supposed to exist in the media library? It keeps adding "dominant colors" to PNG images etc., in site layouts etc., where there shouldn't be solid colors, according to design. I keep deinstalling updates of Performance lab and reverting files back to version mentioned in comment above. I'm advising admins and other users against updating this plugin until this is sorted out.

This feature might be useful if it was an option that could be activated or not - per user/admin choice - checkbox?

[...]

PR #587, opened 2022-11-29
Integrate dominiant color with the media library

adamsilverstein:
## Summary

Fixes https://github.com/WordPress/performance/issues/354

## Relevant technical choices
* In order to modify the templates used by core, the plugin unhooks the core templates and adds a modified copy. All code from L465 on are directly from core, with the addition of some inline styles (L491) and styles added to the HTML (L877). The specific template changes can also be reviewed [here](https://github.com/WordPress/wordpress-develop/compare/trunk...pbearne:wordpress-develop:use_dominat_color_in_media_libary?expand=1).
* Core doesn't provide a way to filter or modify these templates, so the only approach to adding this is to replace them entirely. If/when this feature is merged to core, the changes would be much [smaller](https://github.com/WordPress/wordpress-develop/compare/trunk...pbearne:wordpress-develop:use_dominat_color_in_media_libary?expand=1)

## Screencast:

https://user-images.githubusercontent.com/2676022/208721702-0fb7c562-21ed-4a95-b952-8ab84c01bf2f.mp4

## Checklist


modules/images/dominant-color/load.php
```diff
@@ -405,3 +405,1418 @@ function dominant_color_rgb_to_hex( $red, $green, $blue ) {
 
 	return sprintf( '%02x%02x%02x', $red, $green, $blue );
 }
+
+/**
+ * Integrate with the media library.
+ */
+function dominant_color_add_media_library_integration() {
+	remove_action( 'admin_footer', 'wp_print_media_templates' );
```
adamsilverstein: this removal isn't working here and probably needs to run earlier or later.

adamsilverstein: moved later, works now

pbearne: thanks

modules/images/dominant-color/load.php
```diff
@@ -405,3 +405,1424 @@ function dominant_color_rgb_to_hex( $red, $green, $blue ) {
...
+	if ( ! wp_attachment_is_image( $attachment->ID ) ) {
+		return $response;
+	}
+
+	$dominant_color = dominant_color_get_dominant_color( $attachment->ID );
+	if ( ! $dominant_color ) {
+		return $response;
+	}
+
+	$response['dominant_color'] = $dominant_color;
```
pbearne: theis needs to be $response['dominantColor'] = $dominant_color;

pbearne (approved):
don't like having to have all the duplicate code in the plugin but is the only way

felixarntz:
This is still in draft stage and not ready, therefore punting it to 1.9.0 for now.

Leeroy28 (approved):
Need fix update

adamsilverstein:
> don't like having to have all the duplicate code in the plugin but is the only way

👍🏼 agree, although feel it is ok for the plugin, proposing this in core would be a much smaller change.

modules/images/dominant-color/load.php
adamsilverstein: That fixed it, working for me now. Capturing a screencast.

adamsilverstein:
This now works correctly, thanks for the pointer @pbearne. 

I attached a screencast to the description. I noticed this also works during uploads, but only after image processing on the server (while the thumbnail loads). Even better would be a preview in the browser on upload: Gutenberg does this already.

adamsilverstein:
@pbearne this is ready for another review/testing

pbearne (approved):

pbearne:
@adamsilverstein thanks for doing this

modules/images/dominant-color/load.php
```diff
@@ -417,3 +417,1428 @@ function dominant_color_rgb_to_hex( $red, $green, $blue ) {
 
 	return sprintf( '%02x%02x%02x', $red, $green, $blue );
 }
+
+/**
+ * Integrate with the media library.
+ */
```
mukeshpanchal27:
```suggestion
/**
 * Integrates with the media library.
 *
 * @since n.e.x.t
 */
```

modules/images/dominant-color/load.php
```diff
@@ -417,3 +417,1428 @@ function dominant_color_rgb_to_hex( $red, $green, $blue ) {
...
+function dominant_color_add_media_library_integration() {
+	add_action( 'admin_footer', 'dominant_color_print_media_templates' );
+	add_filter( 'wp_prepare_attachment_for_js', 'dominant_color_prepare_attachment_for_js', 10, 3 );
+}
+add_filter( 'plugins_loaded', 'dominant_color_add_media_library_integration' );
+remove_action( 'admin_footer', 'wp_print_media_templates' );
+
+
+/**
```
mukeshpanchal27:
```suggestion
remove_action( 'admin_footer', 'wp_print_media_templates' );

/**
```

Remove extra empty line

modules/images/dominant-color/load.php
```diff
@@ -417,3 +417,1428 @@ function dominant_color_rgb_to_hex( $red, $green, $blue ) {
...
+ * Add the dominant color when preparing attachment data.
+ *
+ * @param array       $response   Array of prepared attachment data. @see wp_prepare_attachment_for_js().
```
mukeshpanchal27:
```suggestion
 * Add the dominant color when preparing attachment data.
 *
 * @since n.e.x.t
 *
 * @param array       $response   Array of prepared attachment data. @see wp_prepare_attachment_for_js().
```

modules/images/dominant-color/load.php
```diff
@@ -417,3 +417,1428 @@ function dominant_color_rgb_to_hex( $red, $green, $blue ) {
...
+ * @param WP_Post     $attachment Attachment object.
+ * @param array|false $meta       Array of attachment meta data, or false if there is none.
+ *
+ * @return array Array of prepared attachment data.
```
mukeshpanchal27:
```suggestion
 * @param array|false $meta       Array of attachment meta data, or false if there is none.
 * @return array Array of prepared attachment data.
```

modules/images/dominant-color/load.php
```diff
@@ -417,3 +417,1428 @@ function dominant_color_rgb_to_hex( $red, $green, $blue ) {
...
+	} else {
+		$response['hasTransparency']      = 'false';
+		$response['hasTransparencyClass'] = 'no-transparency';
+	}
+
+	return $response;
+}
+
+/**
+ * Print media templates.
```
mukeshpanchal27:
```suggestion
 * Prints the templates used in the media manager.
```

modules/images/dominant-color/load.php
```diff
@@ -417,3 +417,1428 @@ function dominant_color_rgb_to_hex( $red, $green, $blue ) {
...
+ *
+ * Copied from core's wp_print_media_templates() and modified to add the dominant color.
+ *
+ * Eslint is disabled because this block of code is copied from core and only included here to enable modification.
+ */
```
mukeshpanchal27:
```suggestion
 * Eslint is disabled because this block of code is copied from core and only included here to enable modification.
 *
 * @since n.e.x.t
 */
```

mukeshpanchal27 (commented):
Great work @adamsilverstein Left some nit-pick feedback.

adamsilverstein:
Thanks for the review and feedback @mukeshpanchal27 - I have applied your suggestions to the PR

modules/images/dominant-color/load.php
```diff
@@ -417,3 +417,1432 @@ function dominant_color_rgb_to_hex( $red, $green, $blue ) {
...
+	);
+	?>
+
+	<?php
+	// Template for the media frame: used both in the media grid and in the media modal.
+	// Include the inline styles for the dominant color feature.
+	?>
+	<style>
+		.wp-core-ui .attachment-preview[data-dominant-color]:not(.has-transparency) {
+			background: var(--dominant-color);
```
mukeshpanchal27: Instead of `background` use `background-color` so it will not override other background properties. For the frontend request, we used `background-color` https://github.com/WordPress/performance/blob/trunk/modules/images/dominant-color/load.php#L243.

```suggestion
			background-color: var(--dominant-color);
```

modules/images/dominant-color/load.php
```diff
@@ -417,3 +417,1432 @@ function dominant_color_rgb_to_hex( $red, $green, $blue ) {
...
+ * Prints the templates used in the media manager.
+ *
+ * Copied from core's wp_print_media_templates() and modified to add the dominant color.
+ *
+ * Eslint is disabled because this block of code is copied from core and only included here to enable modification.
+ *
+ * @since n.e.x.t
+ */
+// phpcs:disable
+ function dominant_color_print_media_templates() {
```
mukeshpanchal27:
```suggestion
function dominant_color_print_media_templates() {
```

Remove space before the function name.

mukeshpanchal27 (commented):
Thanks @adamsilverstein, The changes look good to me. I left some final feedback and then it was ready for merge.

mukeshpanchal27:
@felixarntz When you get time today, could you please review it so we can merge it for `1.9.0`? Thanks!

felixarntz:
@mukeshpanchal27 @adamsilverstein @pbearne I think for such a massive PR it's a bit late now.

I understand that most of this code is duplicated from WordPress core, but that in itself is not great. Yes, it works, but it creates a maintenance nightmare. How are we able to keep track if anything in that part of core changes?

I personally am not a fan of this approach, though I am aware that there's probably no other way to do it from a plugin. With that said, I don't know whether there's a point in further enhancing this module with _additional_ functionality, when there is still a debate on the very foundation of the module and whether it makes sense for core.

For something like this here, I would advise creating a WordPress core draft pull request, which allows testing this without the need for 100s of lines of duplicate code.

adamsilverstein:
> For something like this here, I would advise creating a WordPress core draft pull request, which allows testing this without the need for 100s of lines of duplicate code.

This was the approach in the original PR Paul opened: https://github.com/WordPress/wordpress-develop/compare/trunk...pbearne:wordpress-develop:use_dominat_color_in_media_libary?expand=1

I don't think we can add this to core by itself though. If we want to test this approach in the plugin without core changes, we need to include a copy of the (modified) media template

felixarntz:
@adamsilverstein 
> I don't think we can add this to core by itself though. If we want to test this approach in the plugin without core changes, we need to include a copy of the (modified) media template

I understand. However as mentioned above that creates a maintenance nightmare that I don't think is justified. The dominant color feature in itself does still not have buy in from the core developers, and there have been, to my knowledge, little attempt to address the pushback on its original feature proposal. So adding it to the media library is going a step too far IMO, based on where the overall feature is at. If it was straightforward from a plugin, surely we could add it. But the limitations of missing hooks in that part of WordPress core to me make this a no-go.

If at some point the feature has merit to move forward, we can always open a core PR with these changes to implement this in a more reasonable manner.

adamsilverstein:
Makes sense. I guess one possibly for core would be adding a hook to make the media templates more customizable.

felixarntz:
Closing this for now as in the current stage there is no reasonable path forward for this to be in the plugin.

PR #591, opened 2022-12-09
Use PHP 7.4 for unit tests

mukeshpanchal27:
## Summary
Resolve unit test failures that occurred after the release of PHP 8.2. 

## Checklist

### DIFF
--- a/.github/workflows/php-test.yml
+++ b/.github/workflows/php-test.yml
@@ -50,6 +50,8 @@ jobs:
         with:
           node-version-file: '.nvmrc'
           cache: npm
+      - name: Composer update
+        run: composer update
       - name: npm install
         run: npm ci
       - name: Install WordPress
--- a/.github/workflows/php-test.yml
+++ b/.github/workflows/php-test.yml
@@ -50,8 +50,6 @@ jobs:
         with:
           node-version-file: '.nvmrc'
           cache: npm
-      - name: Composer update
-        run: composer update
       - name: npm install
         run: npm ci
       - name: Install WordPress
--- a/.github/workflows/php-test.yml
+++ b/.github/workflows/php-test.yml
@@ -9,6 +9,9 @@ on:
     paths:
       - '.github/workflows/php-test.yml'
       - '**.php'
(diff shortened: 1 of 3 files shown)


mehulkaklotar (approved):
@mukeshpanchal27 Looks good to me! 🚀

.github/workflows/php-test.yml
```diff
@@ -44,6 +44,8 @@ jobs:
         with:
           node-version-file: '.nvmrc'
           cache: npm
+      - name: Composer update
+        run: composer update
```
felixarntz, line 54: Why is this addition needed?

felixarntz (commented):
Thanks @mukeshpanchal27, fix looks good, though I have a question on the one change.

.github/workflows/php-test.yml
mukeshpanchal27: Thanks, @felixarntz for the review and question.

Without `composer update` it gives the same error [shown here](https://github.com/WordPress/performance/actions/runs/3653782252/jobs/6173598812) as it did not pull updated composer packages.

similar issue replicated on local systems:
1. git clone git@github.com:WordPress/performance.git
2. git checkout fix/unit-tests-failure
3. npm run test-php

If we run `composer update` before step 3, then unit testing works fine.

felixarntz: Thanks @mukeshpanchal27! I still gave this another try, I was thinking maybe it's due to some cache of Composer dependencies. I'm still a bit confused why it is necessary, but you're right that it seems needed for the tests to pass. Let's leave it as is for now 👍

felixarntz (approved):
Thanks @mukeshpanchal27!

I pushed one more tiny commit https://github.com/WordPress/performance/pull/591/commits/bc9e2dc35fa872bc22ffcde0e18ac591f78736ee to ensure that unit tests also run in the future when making changes to files like `.wp-env.json` since it affects the Docker container we run the logic in.

Btw I notice we get warnings from PHPUnit, opened #594 to fix that.

PR #592, opened 2022-12-09
Update npm dependencies to their latest versions to fix vulnerability

mukeshpanchal27:
## Summary

NPM audit report shows that we have 3 high and 1 critical vulnerabilities in dependency packages:
```
found 4 vulnerabilities (3 high, 1 critical) in 1552 scanned packages
```

We need to update all dependencies to the latest versions to fix vulnerabilities. If it doesn't help, we will need to follow instructions provided by `npm audit` command to fix remaining issues.

## Checklist


mehulkaklotar (approved):
@mukeshpanchal27 Looks good to me! 🚀

eugene-manuilov (approved):
LGTM

felixarntz (changes requested):
@mukeshpanchal27 There may be some problems introduced by (some of) these dependency updates. At least on my system, I now get errors when using our CLI commands.

For example, when running `npm run translations`:
```
> ./bin/plugin/cli.js translations

internal/modules/cjs/loader.js:1102
      throw new ERR_REQUIRE_ESM(filename, parentPath, packageJsonPath);
      ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: .../performance/node_modules/chalk/source/index.js
require() of ES modules is not supported.
require() of .../performance/node_modules/chalk/source/index.js from .../performance/bin/plugin/lib/logger.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from .../performance/node_modules/chalk/package.json.
```

Can you check on your system? Just wanna rule out it's not just an issue on mine. I have verified though that at least for mine, the errors are introduced from this dependency update - using the dependencies from `trunk` works as expected.

If this is indeed a general problem (not specific to my machine), we need to either fix our code, or for now we could only update _some_ of these dependencies that do not result in the above error.

Generally, I would suggest when making dependency updates like this to always test some of our scripts to make sure they don't have any problems.

mukeshpanchal27:
Thanks, @felixarntz for the review. I face a similar error for `npm run translations` command. The culprit is in the latest [chalk](https://www.npmjs.com/package/chalk) version.

I have troubleshooted the issue in details and found that after the [chalk version 5.0](https://github.com/chalk/chalk/releases#hd-7cc85ca8) it supports pure [ESM](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules). It can't support `require()` anymore. So I downgraded it to `4.1.2` and tested all of the commands, and it works perfectly in my system.

mehulkaklotar (approved):
@mukeshpanchal27 I also tested the latest changes and npm commands seems to be working fine without notices.

felixarntz (approved):
Thanks @mukeshpanchal27, LGTM! Everything appears to work great now.

felixarntz:
Just fixed the merge conflict here after #593 got merged, which removed some dependencies entirely.

PR #547, opened 2022-09-29
Implement new experimental SQLite integration module

aristath:
## Summary

This PR introduces a module to allow testing the SQLite implementation - as discussed in the performance team's Slack meeting this week.
For background, you can read the [SQLite proposal on make.w.org](https://make.wordpress.org/core/2022/09/12/lets-make-wordpress-officially-support-sqlite/)

This is a continuous work in progress, but it is working well enough to be included as an experimental module. The issues are not so much with the SQLite implementation itself, as much as with the module's activation/deactivation and the overall User Experience during that process.

The code for the SQLite implementation was copied from https://github.com/aaemnnosttv/wp-sqlite-db/blob/master/src/db.php, which is a fork of the original work on the [sqlite-integration](https://wordpress.org/plugins/sqlite-integration/) plugin. It was then refactored a bit, coding-standards were applied, and an integration with the plugin was built.

## Checklist

- [x] PR has either `[Focus]` or `Infrastructure` label.
- [x] PR has a `[Type]` label.
- [x] PR has a milestone or the `no milestone` label.

PR #595, opened 2022-12-12
Make WP filesystem setup more robust to prevent potential errors

felixarntz:
## Summary

Follow up to #547: The `WP_Filesystem()` function is not guaranteed to succeed. If it fails in an environment, we need to account for that, as otherwise methods called on the global will at best "just not work", at worst result in a fatal error.

## Checklist


aristath (approved):
Looks good to me 👍

mukeshpanchal27 (approved):
Thanks @felixarntz, LGTM.

PR #596, opened 2022-12-12
Actually update Composer dependencies instead of calling composer update

felixarntz:
## Summary

Follow up to #591: In that PR, `composer update` was added to the GitHub workflow for unit tests. This isn't really an appropriate fix since it updates the Composer dependencies locally before running tests on the container.

The actual reason this was needed was because the Composer dependencies in `composer.lock` are out of date, so we need to simply commit that updated file instead of running `composer update` on every GitHub workflow.

## Checklist

### DIFF
--- a/.github/workflows/php-test.yml
+++ b/.github/workflows/php-test.yml
@@ -50,8 +50,6 @@ jobs:
         with:
           node-version-file: '.nvmrc'
           cache: npm
-      - name: Composer update
-        run: composer update
       - name: npm install
         run: npm ci
       - name: Install WordPress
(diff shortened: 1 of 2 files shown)


felixarntz:
@mukeshpanchal27 Figured out my comment from https://github.com/WordPress/performance/pull/591#discussion_r1046141845 :)

mukeshpanchal27 (approved):
Thanks @felixarntz for the PR, LGTM! 🚀

mehulkaklotar (approved):
@felixarntz  Perfect! 🚀

PR #603, opened 2022-12-14
Fix incorrect handling of `admin_email` and actual admin user's email when original `admin_email` user was deleted

### LINKED ISSUE #601: SQLite module: Wrong user details
## Bug Description

User information get wrong when SQLite is enabled.

## Steps to reproduce

1. Setup WordPress with an email `wordpress@example.com`
2. Go to Admin Panel > Users > Add New user with email `test@test.com` and First and Last name
3. Log out
4. Login with `test@test.com` user
5. Go to Admin Panel > Users and delete `wordpress@example.com` user
6. Now Activate SQLite module (Go to Admin Panel > Settings > Performance > SQLite Integration)
7. Check User details at Admin Panel > Users
8. The new WordPress installation uses the email address `wordpress@example.com` rather than `test@test.com` 
9. It should use the original user information instead of the old WP setup admin user details.

## Additional Context

 - PHP Version: 7.4.33
 - OS: Mac
 - Browser: chrome
 - Plugin Version: 1.7.0
 - Device: Macbook Pro
 - SQLite version: 3.34.1

felixarntz:
## Summary

Fixes #601

## Relevant technical choices

* This PR ensures the `admin_email` option and the admin user's email keep their original values even in case they were different (e.g. when the original admin user was deleted).
* See https://github.com/WordPress/performance/issues/601#issuecomment-1352012473 for additional context on the fix.
* The PR includes one additional (somewhat unrelated) safe guard of not running the automatic setup in a multisite. While it would be great to eventually cover that too, it is a lot more complicated; for now it's the safest option to let the user handle it for a multisite environment.

## Checklist

### DIFF
--- a/modules/database/sqlite/activate.php
+++ b/modules/database/sqlite/activate.php
@@ -76,28 +76,24 @@ return function() {
 			}
 
 			// Get current basic setup data to install WordPress in the new DB.
-			$blog_title = get_bloginfo( 'name' );
-			$is_public  = (bool) get_option( 'blog_public' );
-			$language   = get_option( 'WPLANG' );
-			$user_email = get_option( 'admin_email' );
-			$admin_user = get_user_by( 'email', $user_email );
+			$blog_title  = get_bloginfo( 'name' );
+			$is_public   = (bool) get_option( 'blog_public' );
+			$language    = get_option( 'WPLANG' );
+			$admin_email = get_option( 'admin_email' );
+			$admin_user  = get_user_by( 'email', $admin_email );
 			if ( ! $admin_user ) {
 				$admin_user = wp_get_current_user();
 			}
 
-			// If the current user is not the admin email user, look up the
-			// data for the current user. Additionally, attempt to keep them
+			// Look up the data for the current user (in case they're not the
+			// admin email user). Additionally, attempt to keep them
 			// logged-in by retaining their current session. Depending on the
 			// site configuration, this is not 100% reliable as sites may store
 			// session tokens outside of user meta. However that does not lead
 			// to any problem, the user would simply be required to sign in
 			// again.
-			$current_user    = null;
-			$current_user_id = get_current_user_id();
-			if ( $current_user_id !== (int) $admin_user->ID ) {
-				$current_user = wp_get_current_user();
-			}
-			$user_sessions = get_user_meta( $current_user_id, 'session_tokens', true );
+			$current_user          = wp_get_current_user();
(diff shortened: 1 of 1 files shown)


modules/database/sqlite/activate.php
```diff
@@ -41,6 +41,12 @@
 		}
 	}
 
+	// Do not automatically set up SQLite DB on a multisite as that is notably
+	// more complex. Potentially it can be added in the future.
+	if ( is_multisite() ) {
+		return;
+	}
```
aristath, line 48: Should we also add a check for `is_multisite` in the `can-load.php` file? 🤔

mukeshpanchal27: If we add the `is_multisite` check in the `can-load.php` file, we don't need to add it to the `activate.php` file. WDYT?

mukeshpanchal27 (commented):
Thanks @felixarntz The PR resolved the user email issue, but it still did not pull user details such as first name, last name, website, biographical information, and so on.

Can we fix this edge case issue in the next release? WDYT?

modules/database/sqlite/activate.php
ankitrox: Yes, I think this is better candidate to go in `can-load.php` as if site is multisite, then we should instruct not to load module at all rather than instructing to load module and then simply doing nothing!

ankitrox:
Apart from [this comment](https://github.com/WordPress/performance/pull/603#discussion_r1049306164), everything else looks fine to me.

ankitrox (approved):

aristath (approved):
I'll go ahead and pre-approve this. Moving the `is_multisite` check to `can-load` is not a deal-breaker, I'm happy to merge this one as-is if we don't want to move the multisite check. cc @felixarntz

modules/database/sqlite/activate.php
felixarntz: @aristath @mukeshpanchal27 @ankitrox This shouldn't be necessary. It's not that the entire module doesn't work on multisite, it's just that the automatic installation doesn't work with multisite.

felixarntz:
@mukeshpanchal27 
> The PR resolved the user email issue, but it still did not pull user details such as first name, last name, website, biographical information, and so on.

I think that's expected. We explicitly say that this setup is not a migration of data, we only set up the bare minimum WordPress SQLite database based on the data from the original database. User details like those you're saying are not essential to a WordPress install (i.e. they are not encompassed by `wp_install()`).

mukeshpanchal27 (approved):

dianatupas:
This bug is still existing

PR #604, opened 2022-12-15
Add an indicator in the adminbar to show when using SQLite

aristath:
## Summary

Adds an indicator in the adminbar to show when using SQLite:

Screenshot 2022-12-15 at 2 54 54 PM

## Checklist

### DIFF
--- a/modules/database/sqlite/load.php
+++ b/modules/database/sqlite/load.php
@@ -58,3 +58,21 @@ function perflab_sqlite_plugin_admin_notice() {
 	);
 }
 add_action( 'admin_notices', 'perflab_sqlite_plugin_admin_notice' ); // Add the admin notices.
+
+/**
+ * Add a link to the admin bar.
+ *
+ * @param WP_Admin_Bar $admin_bar The admin bar object.
+ */
+function perflab_sqlite_plugin_adminbar_item( $admin_bar ) {
+	$args = array(
+		'id'     => 'performance-lab-sqlite',
+		'parent' => 'top-secondary',
+		/* translators: %s: SQLite icon. */
+		'title'  => '<span style="color:#46B450;">' . __( 'SQLite activated', 'performance-lab' ) . '</span>',
+		'href'   => esc_url( admin_url( 'options-general.php?page=' . PERFLAB_MODULES_SCREEN ) ),
+		'meta'   => false,
+	);
+	$admin_bar->add_node( $args );
+}
+add_action( 'admin_bar_menu', 'perflab_sqlite_plugin_adminbar_item', 999 );
--- a/modules/database/sqlite/load.php
+++ b/modules/database/sqlite/load.php
@@ -65,6 +65,9 @@ add_action( 'admin_notices', 'perflab_sqlite_plugin_admin_notice' ); // Add the 
  * @param WP_Admin_Bar $admin_bar The admin bar object.
  */
 function perflab_sqlite_plugin_adminbar_item( $admin_bar ) {
+	if ( ! defined( 'PERFLAB_SQLITE_DB_DROPIN_VERSION' ) ) {
+		return;
+	}
 	$args = array(
 		'id'     => 'performance-lab-sqlite',
 		'parent' => 'top-secondary',
--- a/modules/database/sqlite/load.php
+++ b/modules/database/sqlite/load.php
@@ -65,14 +65,13 @@ add_action( 'admin_notices', 'perflab_sqlite_plugin_admin_notice' ); // Add the 
  * @param WP_Admin_Bar $admin_bar The admin bar object.
  */
 function perflab_sqlite_plugin_adminbar_item( $admin_bar ) {
-	if ( ! defined( 'PERFLAB_SQLITE_DB_DROPIN_VERSION' ) ) {
-		return;
-	}
 	$args = array(
 		'id'     => 'performance-lab-sqlite',
 		'parent' => 'top-secondary',
 		/* translators: %s: SQLite icon. */
-		'title'  => '<span style="color:#46B450;">' . __( 'SQLite activated', 'performance-lab' ) . '</span>',
+		'title'  => defined( 'PERFLAB_SQLITE_DB_DROPIN_VERSION' ) && defined( 'DATABASE_TYPE' ) && 'sqlite' === DATABASE_TYPE
+			? '<span style="color:#46B450;">' . __( 'Database: SQLite', 'performance-lab' ) . '</span>'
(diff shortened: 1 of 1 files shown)


felixarntz:
@aristath The change looks good to me, but I think we need to evaluate whether this is generally considered useful enough to take a spot in the admin bar. I think it's a bit too late for 1.8.0 since it's an enhancement that was just opened today and needs a bit of thought, so will move this to 1.9.0.

felixarntz:
@aristath We still haven't had any conversation about this. For example here's some questions I have which would be good to discuss:
* Is this something that users of the module would want to see?
* Is it justified to reserve some of the very limited space in the admin bar for a very basic piece of information like this?
* Since it sets a precedent on adding PL information to the admin bar, should we rather think about a dedicated PL admin bar menu which could potentially incorporate other important heuristics than just this?

Last but not least, should we do this or should we not do this?

@bethanylang This is probably not a massive conversation to reserve an entire chat for, but still it could be a good topic to cover in one of the next ones.

For now I'll move this to the next release (2.0.0) since we need to create the 1.9.0 release branch today.

felixarntz:
@aristath Wondering what your take is on the above? What led you to opening this PR? :)

aristath:
Initially, when someone was activating the module, they were getting the install screen where they would set their site-name etc. With that workflow, I was just setting the site-name to `SQLite test` and I could immediately see (by the site-title) which db I'm working with.
In order to make the testing experience smoother, that process was automated and the installation of WP runs automatically, copying over the pre-existing sitename, admin-user etc. As a result, I no longer know which db I'm working with unless I go to site-health to check - or check the filesystem, to see if wp-content/db.php is there.
If we keep the current flow that runs during the module's activation, a visual indication of which DB is currently used would be useful.

felixarntz:
Thanks @aristath, makes sense. I was just a bit concerned to use the limited WP Admin bar space, but that use-case is reasonable to me.

However, I would advise we also show that menu item when _not_ using SQLite, just as long as the module is active. We also need to consider the value of the `DATABASE_TYPE` constant. Maybe like this:
* If `PERFLAB_SQLITE_DB_DROPIN_VERSION` is defined _and_ `DATABASE_TYPE` is set to "sqlite", the menu displays in green something like "Database: SQLite".
* Otherwise, the menu displays in red something like "Database: MySQL".

Or alternatively, we could just rely on `$wpdb->db_server_info()` to determine which DB is being used, this way we wouldn't need to deal with the constants at all, and it would look at the real source of truth (arguably the `$wpdb` class is the source of truth for what is _actually_ being used). If we do that, we can even differentiate between "SQLite", "MySQL", and "MariaDB" if we want (the latter isn't _really_ necessary but nice to have).

Let me know what you think. My main feedback would be to show the menu even when SQLite is _not_ enabled but the module is enabled, as that would be a good indicator that something is wrong.

felixarntz:
@aristath Friendly ping ^

OllieJones (approved):

SergeyBiryukov:
> Or alternatively, we could just rely on `$wpdb->db_server_info()` to determine which DB is being used, this way we wouldn't need to deal with the constants at all, and it would look at the real source of truth (arguably the `$wpdb` class is the source of truth for what is _actually_ being used). If we do that, we can even differentiate between "SQLite", "MySQL", and "MariaDB" if we want (the latter isn't _really_ necessary but nice to have).

Sounds good to me :+1:

aristath:
> * If `PERFLAB_SQLITE_DB_DROPIN_VERSION` is defined _and_ `DATABASE_TYPE` is set to "sqlite", the menu displays in green something like "Database: SQLite".
> * Otherwise, the menu displays in red something like "Database: MySQL".

Done.
I also rebased the PR to bring it up to speed with `trunk` 👍

modules/database/sqlite/load.php
```diff
@@ -58,3 +58,23 @@ function perflab_sqlite_plugin_admin_notice() {
...
+ * @param WP_Admin_Bar $admin_bar The admin bar object.
+ */
+function perflab_sqlite_plugin_adminbar_item( $admin_bar ) {
+	$args = array(
+		'id'     => 'performance-lab-sqlite',
+		'parent' => 'top-secondary',
+		/* translators: %s: SQLite icon. */
+		'title'  => defined( 'PERFLAB_SQLITE_DB_DROPIN_VERSION' ) && defined( 'DATABASE_TYPE' ) && 'sqlite' === DATABASE_TYPE
+			? '<span style="color:#46B450;">' . __( 'Database: SQLite', 'performance-lab' ) . '</span>'
+			: '<span style="color:#DC3232;">' . __( 'Database: MySQL', 'performance-lab' ) . '</span>',
```
felixarntz: See my comment https://github.com/WordPress/performance/pull/604#issuecomment-1406864073 (also https://github.com/WordPress/performance/pull/604#issuecomment-1430594496), we should use `$wpdb->db_server_info()` to display more accurate information. It's not always MySQL, it can also be MariaDB.

Also noting that inlining this conditional here makes this hard to parse, we can simplify this by e.g. using a `$title` variable.

modules/database/sqlite/load.php
```diff
@@ -58,3 +58,23 @@ function perflab_sqlite_plugin_admin_notice() {
 	);
 }
 add_action( 'admin_notices', 'perflab_sqlite_plugin_admin_notice' ); // Add the admin notices.
+
+/**
+ * Add a link to the admin bar.
```
felixarntz: Function descriptions should start with a third-person verb per WP docs standards. Also, we need to add a `@since` annotation.
```suggestion
 * Adds a link to the admin bar.
 *
 * @since n.e.x.t
```

felixarntz (commented):
@aristath A bit of additional follow up feedback.

felixarntz:
@aristath In the interest of time, I've quickly addressed the outstanding feedback myself in https://github.com/WordPress/performance/pull/604/commits/6bf932872abec6a79a7a535aff8c99e63ea15a36, since we preferably need to merge this today to get it into the 2.0.0 release.

@SergeyBiryukov Can you give this a quick review, since you previously commented (and I rather don't just approve my own additions 😄)?

felixarntz (approved):

felixarntz:
@aristath Can you add this change to the standalone plugin as well please?

aristath:
Will do 👍

felixarntz:
I'm gonna go ahead and merge this in order to create the new release branch. If there's anything that should be changed, we can always open a quick follow up PR against that branch `release/2.0.0`.

PR #617, opened 2022-12-26
Preloading images and stylesheets: <link> vs Early Hints response header

felixarntz:
## Summary

This is an experiment. **Do not merge.**

## Checklist

### DIFF
--- a/modules/images/hero-image-early-hints/load.php
+++ b/modules/images/hero-image-early-hints/load.php
@@ -81,12 +81,42 @@ function perflab_hieh_add_hooks() {
 add_action( 'init', 'perflab_hieh_add_hooks' );
 
 /**
+ * Replacement for the WP core status_header() function.
+ *
+ * This function in theory supports setting multiple HTTP status headers, which WP core's function does not support.
+ *
+ * However, in practice this still doesn't matter, as PHP itself allows only for a single HTTP response code, making
+ * this entire feature not usable in the server environment.
+ *
+ * @since n.e.x.t
+ *
+ * @param int    $code        HTTP status code.
+ * @param string $description Description for the HTTP status.
+ */
+function perflab_hieh_fixed_status_header( $code, $description ) {
+	if ( ! $description ) {
+		return;
+	}
+
+	$protocol      = wp_get_server_protocol();
+	$status_header = "$protocol $code $description";
+
+	// This filter is copied over from the WP core function status_header().
+	$status_header = apply_filters( 'status_header', $status_header, $code, $description, $protocol );
+
+	if ( ! headers_sent() ) {
+		// The `false` here is the critical change needed, to not override previous status headers.
+		header( $status_header, false, $code );
+	}
+}
+
+/**
  * Checks the request URI and based on it attempts to send a 103 Early Hints header for the hero image.
  *
  * @since n.e.x.t
  */
 function perflab_hieh_send_early_hints_header() {
-	global $perflab_hieh_request_uri;
+	global $perflab_hieh_request_uri, $wp_header_to_desc;
 
 	// Bail if not a frontend request.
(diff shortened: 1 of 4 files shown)


modules/js-and-css/theme-style-early-hints/load.php
```diff
@@ -0,0 +1,112 @@
...
+		}
+		if ( $registered[ $handle ]->src === $stylesheet_uri ) {
+			$preload = $registered[ $handle ]->src;
+			$ver     = $registered[ $handle ]->ver;
+			if ( null !== $ver ) {
+				if ( false === $ver ) {
+					$ver = wp_styles()->default_version;
+				}
+				$preload = add_query_arg( 'ver', $ver, $preload );
+			}
```
mukeshpanchal27, line 45: When the version of the stylesheet is not define then we didn't get theme version. ex. `wp_enqueue_style( 'style', get_stylesheet_uri(), array(), '' );`

modules/js-and-css/theme-style-early-hints/load.php
```diff
@@ -0,0 +1,112 @@
...
+}
+add_action( 'wp_print_styles', 'perflab_tseh_stylesheet_check' );
+
+/**
+ * Resets the theme's style.css check (typically upon changing the current theme).
+ *
+ * @since n.e.x.t
+ */
+function perflab_tseh_reset_stylesheet_check() {
+	delete_option( 'perflab_preload_stylesheet' );
```
mukeshpanchal27, line 60: Please make sure to delete the new option in `uninstall.php`.

modules/js-and-css/theme-style-early-hints/load.php
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Checks whether the theme's style.css file is actually being enqueued and records the URL.
+ *
+ * @since n.e.x.t
+ */
+function perflab_tseh_stylesheet_check() {
+	// Bail if not a frontend request.
+	if ( is_admin() || defined( 'XMLRPC_REQUEST' ) || defined( 'REST_REQUEST' ) || defined( 'MS_FILES_REQUEST' ) ) {
+		return;
+	}
```
mukeshpanchal27, line 20: Similar conditions are checked in `perflab_tseh_send_early_hints_header()`. Add a new function and use it.

mukeshpanchal27 (commented):
Thanks @felixarntz Left some feedback.

modules/images/hero-image-early-hints/load.php
```diff
@@ -0,0 +1,174 @@
...
+					}
+				},
+				// phpcs:ignore PHPCompatibility.Constants.NewConstants.php_int_minFound
+				PHP_INT_MIN
+			);
+			return $headers;
+		}
+	);
+}
+add_action( 'plugins_loaded', 'perflab_hieh_send_early_hints_header' );
```
eugene-manuilov, line 215: Do we really need all this logic if we can just check the current post thumbnail ID and preload that image if it is available? If there is no thumbnail for the post, then I think we should preload nothing because the first image in the content is not always the hero image. Furthermore, the first image in the content can be far below the fold and in this case preloading it is inefficient. 🤔

felixarntz: @eugene-manuilov What you're saying about the first content image can certainly happen, though at the same time it even applies to the featured image, it entirely depends on the theme where that is displayed. It _usually_ is in the header area but not always.

Each of these concerns already apply to the `fetchpriority="high"` and `loading="lazy"` features too to be fair.

felixarntz:
@mukeshpanchal27 @eugene-manuilov Thank you for the feedback. Just to highlight, this PR is not meant to be merged. It's fine to review it, but there's a couple things in here that I just quickly hacked together and that would certainly not be a proper approach if we wanted to truly make this a module. Just as some context; I'll probably eventually close this, not really worth too much polishing (unless something is broken that even affects the lab tests we're running with this code).

phanduynam:

PR #629, opened 2023-01-24
Add `perflab_disable_object_cache_dropin` filter

### LINKED ISSUE #628: Disable Server Timing via filter
## Feature Description

Object Cache Pro and Redis Object Cache both allow users to enable and disable the Redis cache via CLI and UI. 

This plugin hijacking the `object-cache.php` file prevents that core functionality.

I’d like to disable the Server Timing hijack using a filter, not a constant, from my plugins.

It created an unnecessary support burden over the past weeks.

tillkruss:
## Summary

Fixes #628 and #629.

## Relevant technical choices

Using a filter is easy for other plugin to use, while the existing constant would need to be set in the `wp-config.php` which isn't easily controlled by a plugin.

## Checklist

### DIFF
--- a/load.php
+++ b/load.php
@@ -294,7 +294,7 @@ function perflab_maybe_set_object_cache_dropin() {
 	/**
 	 * Filters whether the Perflab server timing drop-in should be set.
 	 *
-	 * @since x.x.x
+	 * @since n.e.x.t
 	 *
 	 * @param bool Whether the server timing drop-in should be set.
 	 */
--- a/load.php
+++ b/load.php
@@ -290,6 +290,17 @@ function perflab_maybe_set_object_cache_dropin() {
 	if ( PERFLAB_OBJECT_CACHE_DROPIN_VERSION ) {
 		return;
 	}
+	
+	/**
+	 * Filters whether the Perflab server timing drop-in should be set.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param bool Whether the server timing drop-in should be set.
+	 */
+	if ( apply_filters( 'perflab_object_cache_dropin', '__return_false' ) ) {
+		return;
+	}
 
 	// Bail if already attempted before timeout has been completed.
 	// This is present in case placing the file fails for some reason, to avoid
--- a/load.php
+++ b/load.php
@@ -290,7 +290,7 @@ function perflab_maybe_set_object_cache_dropin() {
 	if ( PERFLAB_OBJECT_CACHE_DROPIN_VERSION ) {
 		return;
 	}
-	
+
 	/**
 	 * Filters whether the Perflab server timing drop-in should be set.
 	 *
--- a/load.php
+++ b/load.php
@@ -298,7 +298,7 @@ function perflab_maybe_set_object_cache_dropin() {
 	 *
 	 * @param bool Whether the server timing drop-in should be set.
 	 */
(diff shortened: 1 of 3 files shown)


load.php
```diff
@@ -291,6 +291,17 @@ function perflab_maybe_set_object_cache_dropin() {
...
 	}
 
+	/**
+	 * Filters whether the Perflab server timing drop-in should be set.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param bool Whether the server timing drop-in should be set.
+	 */
+	if ( apply_filters( 'perflab_object_cache_dropin', '__return_false' ) ) {
```
felixarntz: Maybe name this to `perflab_disable_object_cache_dropin`? Also the default value is incorrect, should be just the boolean.
```suggestion
	if ( apply_filters( 'perflab_disable_object_cache_dropin', false ) ) {
```

load.php
```diff
@@ -291,6 +291,17 @@ function perflab_maybe_set_object_cache_dropin() {
 		return;
 	}
 
+	/**
+	 * Filters whether the Perflab server timing drop-in should be set.
+	 *
+	 * @since x.x.x
```
```suggestion
	 * @since n.e.x.t
```

felixarntz (commented):
@tillkruss Two minor points of feedback here, also can you please add a test covering the filter?

E.g. like this:
```php
	public function test_perflab_maybe_set_object_cache_dropin_disabled_via_filter() {
		global $wp_filesystem;

		$this->set_up_mock_filesystem();

		// Ensure PL object-cache.php drop-in is not present and constant is not set.
		$this->assertFalse( $wp_filesystem->exists( WP_CONTENT_DIR . '/object-cache.php' ) );
		$this->assertFalse( PERFLAB_OBJECT_CACHE_DROPIN_VERSION );

		// Add filter to disable drop-in.
		add_filter( 'perflab_disable_object_cache_dropin', '__return_true' );

		// Run function to place drop-in and ensure it still doesn't exist afterwards.
		perflab_maybe_set_object_cache_dropin();
		$this->assertFalse( $wp_filesystem->exists( WP_CONTENT_DIR . '/object-cache.php' ) );
	}
```

felixarntz (approved):
Thanks @tillkruss!

load.php
```diff
@@ -291,6 +291,17 @@ function perflab_maybe_set_object_cache_dropin() {
...
 	}
 
+	/**
+	 * Filters whether the Perflab server timing drop-in should be set.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param bool Whether the server timing drop-in should be set.
+	 */
+	if ( apply_filters( 'perflab_disable_object_cache_dropin', false ) ) {
```
peterwilsoncc, line 301: For clarity of intent, I'd suggest `enable` with a default of `true` rather than `disable` with a default of `false`.

tillkruss: Even tho a few lines up it says `PERFLAB_DISABLE_OBJECT_CACHE_DROPIN`?

peterwilsoncc: Sorry, missed that as it's outside the diff. Let's go with the awkward phrasing.

peterwilsoncc (approved):

mukeshpanchal27 (approved):
Thanks @tillkruss, LGTM!

PR #641, opened 2023-02-02
Update WordPress version to 6.1 and remove Cache modules 

### LINKED ISSUE #627: Bump minimum WP version to 6.1, remove merged modules
For context on why I'm opening this issue, please see https://make.wordpress.org/performance/handbook/performance-lab/version-support-policy/. Specifically the following:
> The Performance Lab plugin commits to supporting the latest stable version of WordPress core.

> The Performance Lab plugin commits to only removing a module in combination with bumping the minimum WordPress version requirement to the same stable core release that the merged module code was published with.

The two Performance Lab modules for the Site Health checks on full page cache and persistent object cache have been launched in WordPress core's version 6.1, see https://make.wordpress.org/core/2022/10/06/new-cache-site-health-checks-in-wordpress-6-1/. Any iterations of them make most sense to happen in core, and as such the modules in the Performance Lab are now obsolete (they are already disabled from loading when using WP 6.1 anyway).

Therefore I suggest to remove the two modules now entirely. Per the above policy, this also means we have to bump the minimum required WordPress version of the Performance Lab plugin to 6.1, since that is when those two modules were made available in WordPress core.

kt-12:
## Summary

Fixes #627

1. WordPress version bumped up to 6.1
2. Removed two Performance Lab modules for the Site Health checks on full page cache and persistent object cache.
3. Updated CODEOWNER
4. Removed test files

## Checklist


tillkruss:
Make sure all tests are deleted as well.

kt-12:
> Make sure all tests are deleted as well.

Thanks for pointing that out. I'll update that too.

mukeshpanchal27 (commented):
@kt-12 Please also remove [CODEOWNERS](https://github.com/WordPress/performance/blob/trunk/.github/CODEOWNERS) for those module.

mukeshpanchal27 (commented):
Thanks @kt-12, PR looks solid to me. Could you update https://github.com/WordPress/performance/blob/trunk/readme.txt#L4 to `6.1`?

kt-12:
> Thanks @kt-12, PR looks solid to me. Could you update https://github.com/WordPress/performance/blob/trunk/readme.txt#L4 to `6.1`?

@mukeshpanchal27  Should we also be adding any text in readme.txt indicating that we removed object-cache?

felixarntz (approved):
Thanks @kt-12, LGTM!

The `readme.txt` file and `module-i18n.php` file will be automatically updated as part of the release (with latest module list and changelog), so we don't need to do that here. cc @mukeshpanchal27

mukeshpanchal27 (approved):
Thanks @kt-12, Great work!

PR #649, opened 2023-02-11
Add file header to object-cache drop-in to clarify purpose

### LINKED ISSUE #615: The server performance object-cache.php dropin needs a useful header
## Bug Description

When Performance Lab is activated, visiting `/wp-admin/plugins.php?plugin_status=dropins` shows no useful information. 

This isn't just a dashboard cosmetic issue; other code that handles drop-ins uses [get_plugin_data()](https://developer.wordpress.org/reference/functions/get_plugin_data/) to figure out what's going on. The ability to understand the drop-in situation from code will be critical to the success of other drop-ins if this particular hack lives a long time (like hacks tend to do...).

## Steps to reproduce

1. Install and activate some any object cache drop-in.
2. View the drop-ins via  `/wp-admin/plugins.php?plugin_status=dropins` and observe useful info about the drop-in.
3. Install and activate Performance Lab.
4. View the drop-ins via  `/wp-admin/plugins.php?plugin_status=dropins` and observe just rudimentary data.

## Additional Context

 - PHP Version:  7.4.3
 - OS: Linux
 - Browser: any
 - Plugin Version: 1.8.0
 - Device: any

##

Till did a good job of this in the redis object cache.  https://github.com/rhubarbgroup/redis-cache/blob/develop/includes/object-cache.php

felixarntz:
## Summary

Fixes #615

## Checklist

### DIFF
--- a/server-timing/object-cache.copy.php
+++ b/server-timing/object-cache.copy.php
@@ -1,5 +1,14 @@
 <?php
 /**
+ * Plugin Name: Performance Lab Server Timing Object Cache Drop-In
+ * Plugin URI: https://github.com/WordPress/performance
+ * Description: Performance Lab drop-in to register Server-Timing metrics early. This is not a real object cache drop-in and will not override other actual object cache drop-ins.
+ * Version: 1
+ * Author: WordPress Performance Team
+ * Author URI: https://make.wordpress.org/performance/
+ * License: GPLv2 or later
+ * License URI: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+ *
  * Object cache drop-in from Performance Lab plugin.
  *
  * This drop-in is used, admittedly as a hack, to be able to measure server


server-timing/object-cache.copy.php
```diff
@@ -1,5 +1,14 @@
...
 /**
+ * Plugin Name: Performance Lab Server Timing Object Cache Drop-In
+ * Plugin URI: https://github.com/WordPress/performance
+ * Description: Performance Lab drop-in to register Server-Timing metrics early. This is not a real object cache drop-in and will not override other actual object cache drop-ins.
+ * Version: 1
+ * Author: WordPress Performance Team
+ * Author URI: https://make.wordpress.org/performance/
+ * License: GPLv2 or later
+ * License URI: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+ *
```
mukeshpanchal27, line 11:
```suggestion
 * License URI: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 */
 
 /**
```

Could we distinguish between header information and other extra information?

server-timing/object-cache.copy.php
```diff
@@ -1,5 +1,14 @@
 <?php
 /**
+ * Plugin Name: Performance Lab Server Timing Object Cache Drop-In
+ * Plugin URI: https://github.com/WordPress/performance
+ * Description: Performance Lab drop-in to register Server-Timing metrics early. This is not a real object cache drop-in and will not override other actual object cache drop-ins.
+ * Version: 1
```
mukeshpanchal27, line 6:
```suggestion
 * Version: 1.0
```
Nit-pick: Can we use proper version number here?

mukeshpanchal27 (commented):
Thanks @felixarntz, The changes look solid to me. Left few suggestions.

server-timing/object-cache.copy.php
felixarntz: @mukeshpanchal27 I don't think that's necessary. It's reasonable to put additional information in the same block comment as the file header information to parse.

server-timing/object-cache.copy.php
felixarntz: @mukeshpanchal27 I used that since this version is already used in the constant in this file, so we shouldn't diverge from that.

In the future, if we make (real) changes to this file, we can update the version number to e.g. `2.0.0`, but I think at this point it's not worth doing.

Part of me also thinks it's worth using a different versioning approach here, since this file needs to be copied over, and we shouldn't do that unless necessary. So we should keep the version here independent of the "main" version of the overall plugin, and by using another versioning approach that becomes obvious.

server-timing/object-cache.copy.php
mukeshpanchal27: Thanks Felix for the information.

server-timing/object-cache.copy.php
mukeshpanchal27: 👍

mukeshpanchal27 (approved):
Thanks @felixarntz, LGTM!

felixarntz:
@joemcgill Can you please give this a 2nd pass and approval when you have a second?

felixarntz:
@OllieJones @joemcgill Can one of you please give this a 2nd pass and approval when you have a second?

OllieJones (approved):
Looks exactly right to me. Thanks!

PR #662, opened 2023-03-02
CLI command for a build process to transform modules into standalone plugins

### LINKED ISSUE #635: Implement CLI command to for a build process to transform modules into standalone plugins
## Feature Description

Following https://github.com/WordPress/performance/issues/618#issuecomment-1410903536, we now have an initial path forward for unbundling the Performance Lab plugin and creating standalone plugins from modules. While it is still being discussed whether we will publish all modules as standalone plugins or just a subset of them, we can definitely start working on the infrastructure changes needed.

For the purpose of this issue, let's simply start with one module, `images/webp-uploads`. This is just so that we can get started with one plugin. `database/sqlite` would be the other big priority to publish as standalone plugin, but that one is already available through the separate repo, so it's not a good idea to cover it here. In fact, it should be handle in a separate issue as bringing the codebase back into this monorepo will have its own little quirks to consider.

### Requirements

* Add a config file (e.g. something like `plugins.json`) which contains a map of module directories and their associated standalone plugin configuration (for now probably just the WordPress.org plugin repository slug and intended version). The file should for now only contain the `image/webp-uploads` module. For example:
```json
{
  "images/webp-uploads": {
    "slug": "webp-uploads",
    "version": "1.0.0"
  }
}
```
* Implement a new Node CLI command `build-plugins` in `bin/commands` that does the following:
    * Copy over all module directories based on the above config file (i.e. for now only `image/webp-uploads`) into a (git-ignored) build directory.
    * After copying over, replace each `load.php` file's module header with a corresponding plugin header:
        * `Plugin Name` and `Plugin Description` should use the module name and description respectively.
        * `Plugin URI` should point to the module's directory in this GitHub repository.
        * `Text Domain` should match the target plugin's slug.
        * `Version` should use the target version from the config file.
        * All other fields (requirements, author, license, ...) should match what is in the Performance Lab plugin's own plugin header.
    * Replace all usages of the `'performance-lab'` text domain with the correct text domain for the individual plugin based on its slug (see above).

mukeshpanchal27:
## Summary

Fixes #635

## Relevant technical choices
To test this use `npm run build-plugins` command.

## Checklist


bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,219 @@
...
+				moduleDataValue = value + moduleDataValue;
+			}
+			fs.writeFileSync( buildLoadFile, updatedBuildLoadFileContent.replace( new RegExp( value ), moduleDataValue ) );
+			updatedBuildLoadFileContent = fs.readFileSync( buildLoadFile, 'utf-8' );
+		}
+	} );
+}
+
+/**
+ * Updates the `readme.txt` file with the given changelog.
```
10upsimon: @mukeshpanchal27 is this function description correct? `getModuleData` does not appear to update anything, instead it returns data.

10upsimon: It also does not perform any changelog related tasks from what I am observing

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,219 @@
...
+				)
+			);
+		}
+
+		try {
+			const customer = JSON.parse( jsonString );
+			for ( const key in customer ) {
+				const pluginKey = key;
+				const pluginVersion = customer[ pluginKey ].version;
+				const pluginSlug = customer[ pluginKey ].slug;
```
10upsimon: @mukeshpanchal27 (cc @joemcgill ) I don't want to nit pick at this too much given the context of when and how this command will be run, but we are assuming the presence of the `version` and `slug` object properties in all plugin objects in the json file.

While the reality is that they definitely _should always be present_ (and would/should be caught in PR's if not), the command would break the build flow for all plugins following any broken entry.

Do we need to consider checking the values of these properties before assuming their presence, or are we confident that this is such an "internal" usage of the commands that we're unlikely to ever see such an event occurring?

If I've missed some context that negates the above, please let me know.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,219 @@
...
+				try {
+					// Copy module files from the folder.
+					const modulePath = path.join( '.', 'modules/' + key );
+					fs.copySync( modulePath, './build', { overwrite: true } );
```
10upsimon: @mukeshpanchal27 given that we are using the non async versions of the copy function, and given that they _could_ throw errors (for reasons beyond our control), should we consider trying to catch errors on copy? If this fails, it makes all the steps that follow, fruitless.

```
try {
  fs.copySync( modulePath, './build', { overwrite: true } );
} catch (err) {
    log(
        formats.error(
            `\nError copying plugin file: "${ err }"\n\n`
        )
    );
}
```

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Thanks @10upsimon for review. Addressed in https://github.com/WordPress/performance/pull/662/commits/b19e57cd4f65a36829ddfa24cd1dad3f6771d60b

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Thanks @10upsimon for review. Agreed added check in https://github.com/WordPress/performance/pull/662/commits/b19e57cd4f65a36829ddfa24cd1dad3f6771d60b

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Addressed in https://github.com/WordPress/performance/pull/662/commits/b19e57cd4f65a36829ddfa24cd1dad3f6771d60b

package.json
```diff
@@ -18,6 +19,7 @@
     "since": "./bin/plugin/cli.js since",
     "readme": "./bin/plugin/cli.js readme",
     "translations": "./bin/plugin/cli.js translations",
+    "build-plugins": "./bin/plugin/cli.js build-plugins",
```
mukeshpanchal27, line 22: Once PR is ready for merge, then we will remove this.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,237 @@
...
+				)
+			);
+		}
+
+		try {
+			const modules = JSON.parse( jsonString );
+			for ( const key in modules ) {
+				const pluginKey = key;
+				const pluginVersion = modules[ pluginKey ].version;
+				const pluginSlug = modules[ pluginKey ].slug;
```
10upsimon:
```suggestion
				const pluginVersion = modules[ pluginKey ]?.version;
				const pluginSlug = modules[ pluginKey ]?.slug;
```

@mukeshpanchal27 thank you for adding the conditional checks on these `const`'s but if they keys are missing or incorrectly spelled as per prior suggestion, even the property access would fail, hence I am suggesting the `?.` optional chaining method as well.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,237 @@
...
+
+				if ( ! pluginVersion || ! pluginSlug ) {
+					log(
+						formats.error(
+							'\nThe given module configuration is not valid.\n\n'
```
10upsimon:
```suggestion
							'\nThe given module configuration is invalid, the module JSON object is missing the "version" and/or "slug" properties, or they are misspelled.\n\n'
```

10upsimon (changes requested):

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Thanks @10upsimon, Updated in https://github.com/WordPress/performance/pull/662/commits/a3b2c74366d8920e39a393716313898ff6d9399c

10upsimon:
@mukeshpanchal27  when functionally testing, I noticed that the plugins are not added to their own folders within build dir. therefore, while this works fine when one plugin is defined, if multiple plugins were defined it would override the PHP and other files.

I'd expect built source to be placed in `builds/[slug]/` where `[slug]` is the plugins slug as defined in the `plugins.json`.

Instead, this is what I a seeing:

image

10upsimon:
> @mukeshpanchal27 when functionally testing, I noticed that the plugins are not added to their own folders within build dir. therefore, while this works fine when one plugin is defined, if multiple plugins were defined it would override the PHP and other files.
> 
> I'd expect built source to be placed in `builds/[slug]/` where `[slug]` is the plugins slug as defined in the `plugins.json`.
> 
> Instead, this is what I a seeing:
> 
> image

I tested and confirmed that the last plugin in plugins.json is what will be present in the build folder in its current form.

10upsimon (changes requested):
Requesting changes as per my latest comments

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,296 @@
...
+/**
+ * Command for copying the files required for the standalone plugin.
+ */
+exports.handler = async () => {
+	const pluginsFile = path.join( '.', 'plugins.json' );
+	fs.readFile( pluginsFile, 'utf8', ( err, jsonString ) => {
+		if ( err ) {
+			log(
+				formats.error(
+					`\nError reading file from disk: "${ err }"\n\n`
```
felixarntz: Why all the `\n` here? Are those really needed? Is the formatting confusing without them?

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,296 @@
...
+			const modules = JSON.parse( jsonString );
+			for ( const key in modules ) {
+				const pluginKey = key;
+				const pluginVersion = modules[ pluginKey ]?.version;
+				const pluginSlug = modules[ pluginKey ]?.slug;
+
+				if ( ! pluginVersion || ! pluginSlug ) {
+					log(
+						formats.error(
+							'\nThe given module configuration is invalid, the module JSON object is missing the "version" and/or "slug" properties, or they are misspelled.\n\n'
```
felixarntz: See above.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,296 @@
...
+}
+
+/**
+ * Adds can load functions.
+ *
+ * @param {string} slug File header.
+ * @param {string} buildLoadFile File header.
+ * @param {string} buildLoadFileContent File header.
+ */
+async function addCanLoadScript( slug, buildLoadFile, buildLoadFileContent ) {
```
felixarntz: I think this function is unnecessary. Why don't we instead simply build this logic into the individual modules? Even in the modules in the PL plugin, having this logic won't do any harm. If anything, it helps in case one of the requirements is no longer satisfied when the module is already active.

I think having this function here to manually inject PHP code into the files is far more of a maintenance burden and more error-prone than directly adding it to the modules which need it.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,296 @@
...
+			if ( prop === 'pluginuri' ) {
+				moduleDataValue = value + moduleDataValue;
+			}
+			fs.writeFileSync( buildLoadFile, updatedBuildLoadFileContent.replace( new RegExp( value ), moduleDataValue ) );
+			updatedBuildLoadFileContent = fs.readFileSync( buildLoadFile, 'utf-8' );
+		}
+	} );
+
+	// Add load module script.
+	await addCanLoadScript( slug, buildLoadFile, updatedBuildLoadFileContent );
```
felixarntz: See my other comment, I don't think we should have a script for this.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,296 @@
...
+	while ( match ) {
+		const content = match[ 2 ].trim();
+		const prop = moduleHeaders[ match[ 1 ] ];
+		if ( content && prop ) {
+			moduleData[ prop ] = content;
+		}
+		match = regex.exec( fileHeaderContent );
+	}
+	return moduleData;
+}
```
felixarntz, line 160: This function has a lot of duplicated code from the existing `getModuleData()` function. We should reuse code we have, not rewrite it.

I understand that the existing `getModuleData()` function does more things than we need here. We should therefore break it apart into a few smaller utility functions so that we can reuse them here. For example `parseFileHeader( file ) : Object` and `getModuleDataFromHeaders( Object ) : Object`. The latter we can use instead of the duplicate here. The first we can probably use above in `updatePluginFiles()`.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,296 @@
...
+	files.forEach( ( file ) => {
+		const content = fs.readFileSync( file, 'utf-8' );
+		if ( regexp.test( content ) ) {
+			fs.writeFileSync(
+				file,
+				content.replace( regexp, `@since ${ settings.version }` )
+			);
+		}
+	} );
+}
```
felixarntz: Updating `@since` versions is not part of the requirements. While this may be needed eventually, the code here is also incorrect, since we can't just put `@since` to always be the latest version. That would make that annotation pointless. I would suggest we remove this.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,296 @@
...
+async function updateTextDomain( settings ) {
+	const patterns = [
+		path.resolve( __dirname, '../../../build/**/*.php' ),
+	];
+
+	const files = await glob( patterns, {
+		ignore: [ __filename ],
+	} );
+
+	const regexp = new RegExp( 'performance-lab', 'g' );
```
felixarntz: We should make this a bit more specific and replace `'performance-lab'`, not just the unquoted `performance-lab`. This way we only replace encapsulated strings which is less error-prone.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,296 @@
...
+		}
+	} );
+};
+
+/**
+ * Updates the `load.php` file header information.
+ *
+ * @param {[]Settings} settings Plugin settings.
+ */
+async function updatePluginFiles( settings ) {
```
felixarntz: I think a better name for this would be `updatePluginHeader`, since I think that's what this function does.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,296 @@
...
+
+					// Update since version.
+					updateSinceVersion( {
+						version: pluginVersion,
+					} );
+
+					// Update text domain.
+					updateTextDomain( {
+						slug: pluginSlug,
+					} );
```
felixarntz: It would be a better practice to pass the base directory to operate in to all of these functions rather than coupling each function to a specific location. The second parameter of each function could be something like `pluginDir`.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,296 @@
...
+				formats.error(
+					`\nError reading file from disk: "${ err }"\n\n`
+				)
+			);
+		}
+
+		try {
+			const modules = JSON.parse( jsonString );
+			for ( const key in modules ) {
+				const pluginKey = key;
```
felixarntz: Both `key` and `pluginKey` are not really describing what this value is, as it is the module directory. Also we don't need two variables for this.
```suggestion
			for ( const moduleDir in modules ) {
```

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,296 @@
...
+						)
+					);
+					return;
+				}
+
+				try {
+					// Copy module files from the folder.
+					const modulePath = path.join( '.', 'modules/' + key );
+					try {
+						fs.copySync( modulePath, './build/' + pluginSlug, { overwrite: true } );
```
felixarntz: Instead of overwriting, I think we should rather delete the existing folder before. Otherwise, if we delete a file from a module, it may still remain in the built plugin. Overwriting alone would not protect us from that I think.

package.json
felixarntz: Why? I think it's a good  idea to keep the short-hand in here, similar to how we have for the other CLI scripts.

felixarntz (changes requested):
@mukeshpanchal27 Great work so far! I left a few comments for things that we can tidy up a bit. There is some duplicate code which we should avoid, and most importantly we need to remove the `@since` replacement as that isn't part of the requirements and doesn't work like it is here (it's not that trivial, let's ignore those for now).

package.json
mukeshpanchal27: Initially, I assumed it was part of the GH workflow, so I included a script for testing purposes.

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Thanks @felixarntz, In this case it will not update `* @package performance-lab`, Do we needs to update this?

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Removed as it will affect formatting.

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Thanks for pointing it out. I have added this function to the PHP code and removed it from the JS script.

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Thanks for pointing it out. I have removed the version update script.

bin/plugin/commands/build-plugins.js
mukeshpanchal27: 👍

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Thanks for the feedback. I have updated it. Can you please review it and let me know if I missed anything?

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Thanks @felixarntz for the feedback. I have created a common function for module data. But for the plugin header, I'm updating match values in the loop; how can I update if I make a common function?

bin/plugin/commands/build-plugins.js
felixarntz: @mukeshpanchal27 That's a good point. Let's capture that specifically then, i.e. let's make one replacement of `'performance-lab'` and another replacement of `@package performance-lab`.

bin/plugin/commands/build-plugins.js
felixarntz: @mukeshpanchal27 Three follow up things:
* Why are you passing the module's directory to `updatePluginHeader()` and not the module's build directory? We should only make replacements in the build directory. We don't need access to the module directory at all in that function I think.
* You're still not fully using the new `pluginModulePath` parameter in the `updatePluginHeader()` function. That function should no longer assume anything on the locations of the files (i.e. please avoid things like `path.join( '.', ... )`, it should always be `path.join( pluginModulePath, ... )` instead.
* You also need to make the same change for the `updateTextDomain()` function.

bin/plugin/commands/build-plugins.js
felixarntz: @mukeshpanchal27 You could add two new functions to the `common.js` file:
* A new function like `getModuleHeader( moduleFileContent )` which gets the entire multi-line module header string from the module file. For this you will need a new regex to get that entire first comment.
* A new function like `getModuleDataFromHeader( moduleHeader )` which runs the existing regex and logic that is currently in `getModuleData()` on the header string you got from `getModuleHeader( moduleFileContent )`.

You can then call both of those functions in the existing `getModuleData()` function, replacing the relevant logic with those functions.

And then you can also use both of the functions to create the plugin header. The first function just gives you the full header string which you then need to replace in the file. The second function gives you the individual module data values from the header string.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,178 @@
...
+						);
+						continue;
+					}
+
+					// Update file content.
+					updatePluginHeader( {
+						key: moduleDir,
+						slug: pluginSlug,
+						version: pluginVersion,
+						pluginModulePath: modulePath,
```
felixarntz: This needs to be the path to the module in the `build` folder, not the original module path. I would also suggest to shorten this property from `pluginModulePath` to a simpler `pluginPath`.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,178 @@
...
+					} );
+
+					// Update text domain.
+					updateTextDomain( {
+						slug: pluginSlug,
+					} );
```
felixarntz, line 83: See my other comment, you should pass the `build` module path to this function and use it in the function as well.

bin/plugin/commands/common.js
```diff
@@ -133,3 +96,52 @@ exports.getModuleData = async ( modulesDir ) => {
...
 };
+
+/**
+ * Returns the list of module data for all modules.
+ *
+ * @param {string} moduleFile Modules file.
+ *
+ * @return {[]WPModuleData} Module data list.
+ */
+exports.getModuleDataFromHeaders = ( moduleFile ) => {
```
felixarntz: See my comment in https://github.com/WordPress/performance/pull/662#discussion_r1126861733, you can add another function to get the full module header as a string. Then this function here can be changed to receive not the entire `moduleFile`, but rather just the `moduleHeader`.

bin/plugin/commands/common.js
```diff
@@ -133,3 +96,52 @@ exports.getModuleData = async ( modulesDir ) => {
...
+	// Populate slug and focus based on file path.
+	const moduleDir = path.dirname( moduleFile );
+	const moduleData = {
+		slug: path.basename( moduleDir ),
+		focus: path.basename( path.dirname( moduleDir ) ),
+	};
```
felixarntz: This logic should rather be kept in the original `getModuleData()` function, since this is **not** information that it gets from the module header. Let's keep those concerns separated, so that each function only has a single purpose.

bin/plugin/commands/common.js
```diff
@@ -133,3 +96,52 @@ exports.getModuleData = async ( modulesDir ) => {
...
+
+	// Map of module header => object property.
+	const headers = {
+		'Module Name': 'name',
+		Description: 'description',
+		Experimental: 'experimental',
+	};
+
+	// Populate name, description and experimental based on module file headers.
+	const fileContent = fs.readFileSync( moduleFile, 'utf8' );
```
felixarntz: This line won't be needed anymore in this function when you change it to receive the `moduleHeader` as a string.

bin/plugin/commands/common.js
```diff
@@ -41,44 +41,7 @@ exports.getModuleData = async ( modulesDir ) => {
...
-				match = regex.exec( fileContent );
-			}
-
-			// Parse experimental field into a boolean.
-			if ( typeof moduleData.experimental === 'string' ) {
-				moduleData.experimental =
-					moduleData.experimental.toLowerCase() === 'yes';
-			}
-
+			const moduleData = exports.getModuleDataFromHeaders( moduleFile );
```
felixarntz: To clarify my comment in https://github.com/WordPress/performance/pull/662#discussion_r1126861733: I suggest you refactor this to something like:
```suggestion
			const moduleFileContent = fs.readFileSync( moduleFile, 'utf8' );
			const moduleHeader = exports.getModuleHeader( moduleFileContent );

			// Populate slug and focus based on file path.
			const moduleDir = path.dirname( moduleFile );
			const moduleData = {
				slug: path.basename( moduleDir ),
				focus: path.basename( path.dirname( moduleDir ) ),
				...exports.getModuleDataFromHeader( moduleHeader ),
			};
```

Once you have created/updated those functions accordingly, this should work.

modules/images/webp-uploads/load.php
```diff
@@ -8,6 +8,39 @@
  * @package performance-lab
  */
 
+/**
+ * Can load function to determine if the module is already merged into WordPress core or loaded by
+ * another plugin.
+ *
+ * @since 1.0.0
```
```suggestion
 * @since n.e.x.t
```

modules/images/webp-uploads/load.php
```diff
@@ -8,6 +8,39 @@
  * @package performance-lab
  */
 
+/**
+ * Can load function to determine if the module is already merged into WordPress core or loaded by
+ * another plugin.
```
felixarntz: This function doesn't check whether the code is already loaded by another plugin, it only checks whether the code is already merged into WordPress core.
```suggestion
 * Determines whether the feature can be loaded, or whether is already merged into WordPress core.
 *
 * If it cannot be loaded, it will add an action to display a WordPress admin notice.
```

modules/images/webp-uploads/load.php
```diff
@@ -8,6 +8,39 @@
...
+	if ( $can_load() ) {
+		return true;
+	}
+	add_action(
+		'admin_notices',
+		function() {
+			printf(
+				'<div class="notice notice-error"><p>%s</p></div>',
+				sprintf(
+					__( 'The module is already merged into WordPress core or loaded by another plugin.', 'performance-lab' )
```
felixarntz: See above. Also please clarify which module, plus we shouldn't use the word "module" since this message would primarily be for the plugin.
```suggestion
					__( 'The WebP Uploads feature cannot be loaded from within the plugin since it is already merged into WordPress core.', 'performance-lab' )
```

felixarntz: Why `sprint` if there is no placeholder? Also please use `esc_html__()` instead of `__()` for security best practices.

modules/images/webp-uploads/load.php
```diff
@@ -8,6 +8,39 @@
...
+				)
+			);
+			return;
```
felixarntz: This is unnecessary.
```suggestion
```

modules/images/webp-uploads/load.php
```diff
@@ -8,6 +8,39 @@
  * @package performance-lab
  */
 
+/**
+ * Can load function to determine if the module is already merged into WordPress core or loaded by
+ * another plugin.
+ *
+ * @since 1.0.0
+ *
+ * @return boolean
```
```suggestion
 * @return bool Whether the feature can be loaded.
```

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,178 @@
...
+ * Updates the `load.php` file header information.
+ *
+ * @param {[]Settings} settings Plugin settings.
+ */
+async function updatePluginHeader( settings ) {
+	const { key, version, slug, pluginModulePath } = settings;
+	const loadFile = path.join( '.', 'load.php' );
+	const loadFileContent = fs.readFileSync( loadFile, 'utf-8' );
+	const buildLoadFile = path.join( '.', 'build/' + slug + '/load.php' );
+	const buildLoadFileContent = fs.readFileSync( buildLoadFile, 'utf-8' );
```
felixarntz: Please use the new plugin path parameter for all of this.

Also, why do we need to use both the original module's `load.php` file and the build plugin's `load.php` file? We just copied over the module to the `build` folder in the function that calls this, so it'll be the same.

This can simply become something like:
```suggestion
	const loadFile = path.join( pluginPath, 'load.php' );
	const loadFileContent = fs.readFileSync( loadFile, 'utf-8' );
```

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,178 @@
...
+
+	const regex = /\/\\*\\*[\s\S]+?(?=\*\/)/mi;
+	const pluginComment = loadFileContent.match( regex )?.[ 0 ];
```
felixarntz: This logic is what you would be able to move to the `getModuleHeader()` function I'm mentioning in https://github.com/WordPress/performance/pull/662#discussion_r1126861733.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,178 @@
...
+	const newPluginComment = buildLoadFileContent.match( regex )?.[ 0 ];
+
+	fs.writeFileSync( buildLoadFile, buildLoadFileContent.replace( newPluginComment, pluginComment ) );
+
+	let updatedBuildLoadFileContent = fs.readFileSync( buildLoadFile, 'utf-8' );
+
+	const moduleFilePattern = path.join( pluginModulePath, 'load.php' );
+	const moduleFiles = path.resolve( '.', moduleFilePattern );
```
felixarntz: See above, I don't understand what you are trying to do here. To me all of this logic seems unnecessary.

We only need to update the file once, to replace the module header with the plugin header at the very end of this function.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,178 @@
...
+		let moduleDataValue = moduleData[ prop ];
+
+		if ( value && prop && moduleDataValue ) {
+			if ( prop === 'pluginuri' ) {
+				moduleDataValue = value + moduleDataValue;
+			}
+			fs.writeFileSync( buildLoadFile, updatedBuildLoadFileContent.replace( new RegExp( value ), moduleDataValue ) );
+			updatedBuildLoadFileContent = fs.readFileSync( buildLoadFile, 'utf-8' );
+		}
+	} );
```
felixarntz: Rather than having all those individual regexes, let's create the single plugin header string and make only one replacement. This should be straightforward after you have implemented and are using the functions I outlined in https://github.com/WordPress/performance/pull/662#discussion_r1126861733.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,178 @@
...
+}
+
+/**
+ * Updates the text domain in the build folder files content.
+ *
+ * @param {[]Settings} settings Plugin settings.
+ */
+async function updateTextDomain( settings ) {
+	const patterns = [
+		path.resolve( __dirname, '../../../build/**/*.php' ),
```
felixarntz: This is incorrect as it matches all built plugins and their files. You should rather pass the individual built plugin's path to this function and use that here.

Something like this:
```suggestion
		path.resolve( pluginPath, './**/*.php' ),
```

felixarntz (changes requested):
@mukeshpanchal27 There are still a couple things to enhance here, some of that replacement logic is a bit clunky and needs further refinement.

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Thanks @felixarntz for review feedback.

I have added two new function as requested.

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Addressed in https://github.com/WordPress/performance/pull/662/commits/0032104c7595910d2d52b6bfa57eeb5d633256ea

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Updated in https://github.com/WordPress/performance/pull/662/commits/0032104c7595910d2d52b6bfa57eeb5d633256ea

bin/plugin/commands/common.js
mukeshpanchal27: Addressed in https://github.com/WordPress/performance/pull/662/commits/0032104c7595910d2d52b6bfa57eeb5d633256ea

bin/plugin/commands/common.js
mukeshpanchal27: Thanks @felixarntz for suggestion.

modules/images/webp-uploads/load.php
mukeshpanchal27: Great catch. Removed in https://github.com/WordPress/performance/pull/662/commits/0032104c7595910d2d52b6bfa57eeb5d633256ea

bin/plugin/commands/build-plugins.js
mukeshpanchal27: @felixarntz `path.join( '.', 'load.php' )` is use to get the plugin root `load.php` header content that we copy to module header then we update some with module data.

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Thanks @felixarntz Can you please share an example of it or more details?

bin/plugin/commands/build-plugins.js
felixarntz: @mukeshpanchal27 Ah, I see now what you mean. For which values is that needed? Couldn't we create the entire plugin header string in this function without using the Performance Lab's plugin header?

Reading the plugin header and then replacing strings inside it is what requires all those additional regexes you're using here, I think it's rather messy. If we generate the entire plugin header string from scratch, we don't need all that and can simplify this logic.

bin/plugin/commands/build-plugins.js
felixarntz: @mukeshpanchal27 See my reply in https://github.com/WordPress/performance/pull/662#discussion_r1128161824, I would suggest we don't read the original Performance Lab plugin header, but compute the entire header string ourselves. This simplifies things drastically and you can get rid of the additional regexes in `updatePluginHeader()`.

Here's the flow of logic I would suggest for your `updatePluginHeader()` function:
1. Get the content of the `load.php` file within `settings.pluginPath` (you're already doing that).
2. Extract the module header from it (you're already doing that).
3. Get the module data from the module header (you're already doing that).
4. Create the plugin header from scratch, e.g. with a new variable where you put the entire multiline string inside, starting with `/**\n * Plugin Name: ...`.
    * You can use a JS template string to fill in all the variable data for `Plugin Name`, `Description`, `Version`, `Text Domain`.
    * The `Plugin URI` field can be hard-coded to `https://github.com/WordPress/performance/tree/trunk/modules/{originalModulePath}`. For this you may need to pass a new parameter to the function.
    * The `Requires at least`, `Requires PHP`, `Author`, `Author URI`, `License`, and `License URI` parameters can all be hard-coded, to the same values that are used in the Performance Lab plugin header.
    * Don't forget to include the `@package` note at the end of the comment.
5. Replace the entire module header from point 2. with the new plugin header from point 4. within the overall file content you got in point 1.
6. Update the file from point 1. with the new content from point 5.

All this should be simpler than the regular expressions etc that you are currently using in this function.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,200 @@
...
+				)
+			);
+		}
+	} );
+};
+
+/**
+ * Updates the `load.php` file header information.
+ *
+ * @param {[]Settings} settings Plugin settings.
```
felixarntz: The `[]Settings` is not a valid return type, as `Settings` is not a defined thing. Either you need to define it in a comment, or you just use the more generic `Object`. Also, the parameter is not an array, but an object.
```suggestion
 * @param Object settings Plugin settings.
```

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,200 @@
...
+			updatedBuildLoadFileContent = fs.readFileSync( buildLoadFile, 'utf-8' );
+		}
+	} );
+}
+
+/**
+ * Gets the specific module file header information.
+ *
+ * @param {string} moduleFileContent File content.
+ *
```
felixarntz: There should be no blank lines between `@param` and `@return`.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,200 @@
...
+ * @return {[]HeaderData} File header data.
```
felixarntz: See above.
```suggestion
 * @return {Object} File header data.
```

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,200 @@
...
+	const rootLoadFileContent = fs.readFileSync( rootLoadFile, 'utf-8' );
+	// Specific module `load.php` file content.
+	const buildLoadFile = path.join( pluginPath, 'load.php' );
+	const buildLoadFileContent = fs.readFileSync( buildLoadFile, 'utf-8' );
+
+	const rootFileHeader = await getModuleHeader( rootLoadFileContent );
+	const buildFileLoad = await getModuleHeader( buildLoadFileContent );
+
+	// Get module header data.
+	const { name, description } = await getModuleData( buildLoadFileContent );
```
felixarntz: You already get the header above, you should be able to extract the module data from that, which is faster.
```suggestion
	const { name, description } = await getModuleDataFromHeader( buildFileLoad );
```

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,200 @@
...
+	const { key, version, slug, pluginPath } = settings;
+	// Plugin root `load.php` file content.
+	const rootLoadFile = path.join( '.', 'load.php' );
+	const rootLoadFileContent = fs.readFileSync( rootLoadFile, 'utf-8' );
+	// Specific module `load.php` file content.
+	const buildLoadFile = path.join( pluginPath, 'load.php' );
+	const buildLoadFileContent = fs.readFileSync( buildLoadFile, 'utf-8' );
+
+	const rootFileHeader = await getModuleHeader( rootLoadFileContent );
+	const buildFileLoad = await getModuleHeader( buildLoadFileContent );
```
felixarntz: This is a confusing variable name. Why not a clear `moduleHeader`?

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,200 @@
...
+ * Gets the specific module file header information.
+ *
+ * @param {string} moduleFileContent File content.
+ *
+ * @return {[]HeaderData} File header data.
+ */
+async function getModuleData( moduleFileContent ) {
+	const moduleHeader = await getModuleHeader( moduleFileContent );
+	return await getModuleDataFromHeader( moduleHeader );
+}
```
felixarntz: I don't think we need this function. See my related comment above, we already get the module header above, and we need it separately to replace it later (see also https://github.com/WordPress/performance/pull/662#discussion_r1128192056). So we should remove this function and instead call the two individual functions in here directly.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Updates the text domain and package details in the build folder files content.
+ *
+ * @param {[]Settings} settings Plugin settings.
```
felixarntz: See above.
```suggestion
 * @param {Object} settings Plugin settings.
```

bin/plugin/commands/common.js
```diff
@@ -133,3 +105,57 @@ exports.getModuleData = async ( modulesDir ) => {
 			return firstModule.slug.localeCompare( secondModule.slug );
 		} );
 };
+
+/**
+ * Returns the list of module data for module.
+ *
+ * @param {string} moduleHeader Modules file header contetnt.
+ *
+ * @return {[]WPModuleData} Module data list.
```
felixarntz: This doesn't return an array of `WPModuleData`, just a single `WPModuleData`.
```suggestion
 * @return {WPModuleData} Module data.
```

bin/plugin/commands/common.js
```diff
@@ -133,3 +105,57 @@ exports.getModuleData = async ( modulesDir ) => {
...
+	}
+
+	return moduleData;
+};
+
+/**
+ * Returns the file header.
+ *
+ * @param {string} moduleFileContent Module file content.
+ *
```
felixarntz: See above.
```suggestion
```

felixarntz (commented):
@mukeshpanchal27 See my comment replies https://github.com/WordPress/performance/pull/662#discussion_r1128161824 and https://github.com/WordPress/performance/pull/662#discussion_r1128192056 for my most important feedback.

Below I pointed out just a few small details, the above comments are more important.

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Thanks for the feedback. This solves all the problems that I am facing with header replacement and removes much code from the function.

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Thanks @felixarntz This feedback really solve all the problems that i face.

mukeshpanchal27:
Thanks to @felixarntz and @10upsimon for the feedback. PR is ready for another round of review.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,149 @@
...
+					const buildModulePath = path.join( '.', 'build/' + pluginSlug );
+					try {
+						// Clean up build module files directory.
+						fs.rmSync( buildModulePath, { force: true, recursive: true } );
+
+						fs.copySync( modulePath, buildModulePath, { overwrite: true } );
+					} catch ( copyError ) {
+						log(
+							formats.error(
+								`Error copying plugin file: "${ copyError }"`
```
10upsimon:
```suggestion
								`Error copying files for plugin "${pluginSlug}": "${ copyError }"`
```

This error is not necessarily about copying a single plugin file, but rather an entire plugin, a nit pick suggestion, not going to block approval on this alone.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,149 @@
...
+							)
+						);
+						continue;
+					}
+
+					// Update file content.
+					updatePluginHeader( {
+						originalModulePath: moduleDir,
```
10upsimon:
```suggestion
						modulePath: moduleDir,
```

More of an observation that we probably don't need to call this `originalModulePath` and can just call it `modulePath` because the new path is the plugin path, the module path will always remain as it currently is.

Not a blocking change, more of an observation.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,149 @@
...
+	} );
+};
+
+/**
+ * Updates the `load.php` file header information.
+ *
+ * @param {Object} settings Plugin settings.
+ */
+async function updatePluginHeader( settings ) {
+	const { originalModulePath, version, slug, pluginPath } = settings;
```
10upsimon: See my nitpick comment further up around `originalModulePath` vs just `modulePath`

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,149 @@
...
+	const files = await glob( patterns, {
+		ignore: [ __filename ],
+	} );
+
+	const regexp = new RegExp( settings.regex, 'gm' );
+
+	files.forEach( ( file ) => {
+		const content = fs.readFileSync( file, 'utf-8' );
+		if ( regexp.test( content ) ) {
+			fs.writeFileSync(
```
10upsimon: @mukeshpanchal27 as this is a synchronous function, it may throw errors. We should catch these, therefore I'd suggest wrapping this in a `try catch`.

bin/plugin/commands/common.js
```diff
@@ -41,44 +40,16 @@ exports.getModuleData = async ( modulesDir ) => {
 
 	return moduleFiles
 		.map( ( moduleFile ) => {
+			const moduleFileContent = fs.readFileSync( moduleFile, 'utf8' );
```
10upsimon: Another note that we should always catch errors for synchronous functions, especially I/O ones. I'd suggest wrapping this in a `try catch`.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,149 @@
...
+
+	const moduleHeader = await getModuleHeader( buildLoadFileContent );
+
+	// Get module header data.
+	const { name, description } = await getModuleDataFromHeader( moduleHeader );
+
+	const pluginHeader = `/**\n * Plugin Name: ${ name }\n * Plugin URI: https://github.com/WordPress/performance/tree/trunk/modules/${ originalModulePath }\n * Description: ${ description }\n * Requires at least: 6.1\n * Requires PHP: 5.6\n * Version: ${ version }\n * Author: WordPress Performance Team\n * Author URI: https://make.wordpress.org/performance/\n * License: GPLv2 or later\n * License URI: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html\n * Text Domain: ${ slug }\n *\n * @package ${ slug }\n `;
+
+	// Replace the module file header.
+	fs.writeFileSync( buildLoadFile, buildLoadFileContent.replace( moduleHeader, pluginHeader ) );
```
10upsimon: @mukeshpanchal27 we should handle exceptions for synchronous i/o operations, let's wrap this `fs.writeFileSync` operation in a `try catch`.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Updates the `load.php` file header information.
+ *
+ * @param {Object} settings Plugin settings.
+ */
+async function updatePluginHeader( settings ) {
+	const { originalModulePath, version, slug, pluginPath } = settings;
+	// Specific module `load.php` file content.
+	const buildLoadFile = path.join( pluginPath, 'load.php' );
+	const buildLoadFileContent = fs.readFileSync( buildLoadFile, 'utf-8' );
```
10upsimon: @mukeshpanchal27 we should handle exceptions for synchronous i/o operations, let's wrap this fs.readFileSync operation in a `try catch`.

10upsimon (changes requested):
@mukeshpanchal27 I have a few nitpick comments and 4 - what I believe are - important observations around handling errors/exceptions when using i/o synchronous operations like `fs.readFileSync()` and `fs.writeFileSync()`

mukeshpanchal27:
Thanks @10upsimon for feedback, In https://github.com/WordPress/performance/pull/662/commits/6630fef153cd4b2b9c459da602f76bb57e03fca2 i have added exceptions for synchronous i/o operations.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,183 @@
...
+		buildLoadFileContent = fs.readFileSync( buildLoadFile, 'utf-8' );
+	} catch ( err ) {
+		log(
+			formats.error(
+				`Error reading the file "${ buildLoadFile }": "${ err }"`
+			)
+		);
+	}
+
+	const moduleHeader = await getModuleHeader( buildLoadFileContent );
```
10upsimon, line 130: @mukeshpanchal27 if we are setting `buildLoadFileContent` to an empty string by default, further above, we probably want to check for an empty string before proceeding?

Thinking some more around this, if there is a problem and the file can not be read, and therefore defaults to `''`, how does this affect the rest of the codebase where this `updatePluginHeader` function is called? Do we need to return `false` or similar and have conditional checks in areas that call this so that logic that follows calls to this function does not apply?

Keen to hear your thoughts.

felixarntz: +1, we should do an early `return` after logging the error since the remaining logic otherwise won't work anyway.

bin/plugin/commands/build-plugins.js
```diff
@@ -0,0 +1,183 @@
...
+
+	// Get module header data.
+	const { name, description } = await getModuleDataFromHeader( moduleHeader );
+
+	const pluginHeader = `/**\n * Plugin Name: ${ name }\n * Plugin URI: https://github.com/WordPress/performance/tree/trunk/modules/${ modulePath }\n * Description: ${ description }\n * Requires at least: 6.1\n * Requires PHP: 5.6\n * Version: ${ version }\n * Author: WordPress Performance Team\n * Author URI: https://make.wordpress.org/performance/\n * License: GPLv2 or later\n * License URI: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html\n * Text Domain: ${ slug }\n *\n * @package ${ slug }\n `;
```
felixarntz: Maybe you can break this into multiple lines for better readability? JS template strings can be broken apart into multiple lines I think. This way you can also get rid of all the `\n` usage.

felixarntz (approved):
@mukeshpanchal27 LGTM, great work!

A few small nit-picks, but this looks good from my end.

bin/plugin/commands/build-plugins.js
mukeshpanchal27: Thanks both for the feedback. Addressed in https://github.com/WordPress/performance/pull/662/commits/2b0507034912df7a457fc89f6158a473c88d5a4f

10upsimon (approved):

Issue #636, opened 2023-02-01
Add module-specific readme files and enhance CLI command to include standalone plugin readme files


felixarntz:
## Feature Description

Follow up to #635: Since we're going to publish modules as individual plugins, we will need to have `readme.txt` files for them. While this is required for the plugins anyway, it is also a great way for us to provide an overview for each individual module in the GitHub repository.

We can add a `README.md` file in every module directory and then potentially transform it into a `readme.txt` in a build step; or alternatively we can simply format the `README.md` file as a plugin readme from the very start (see https://johnblackbourn.com/readme-md-plugin-on-wordpress-org/).

For the initial scope of this implementation, let's limit the work to just the `images/webp-uploads` module, similar to #635.

### Requirements

* A `readme.txt` file for the WebP Uploads standalone plugin should be added to the `image/webp-uploads` module folder. Some notes on the content:
    * We should focus on a basic foundation, let's not get too caught up in the initial readme content. We can always iterate.
    * The readme can generally follow what the overall Performance Lab readme contains, however of course scoped to just the future standalone WebP Uploads plugin.
    * Let's use version number 1.0.0.
    * For the short description of the plugin, we can use what's currently the short description of the module. Maybe we can tweak it to start with "WordPress core feature plugin to ..." (do xyz)
    * For the longer description, we don't need to provide too much detail for now since this issue is mainly about setting up a minimum viable readme version. It would be a good idea to include links that provide more context, e.g. to the different WebP for WordPress core proposal posts and the Trac ticket.
    * The Installation and FAQ sections can be almost taken over from Performance Lab as is, since the questions are all either generic (just need minor tweaks to focus on "WebP Uploads" instead of "Performance Lab") or already specific to the WebP feature.
    * Let's start with a fresh changelog, just to cover the upcoming initial 1.0.0 release.
* A new `export-ignore` rule should be added to `.gitattributes` that ensures that no `readme.txt` file from any module directory is included in an overall ZIP build, so that the overall Performance Lab plugin ZIP build does not include those readme files. Of course it still needs to include its own `readme.txt` file though.
* As a result of the above, we need to ensure that the `webp-uploads` standalone plugin built via the `build-plugins` script from #635 now contains the new `readme.txt` file.

joemcgill:
[...]

Issue #656, opened 2023-02-16
Overview: Creating standalone plugins


felixarntz:
Originally discussed in #618, this issue acts purely as an overview of all the issues related to creating standalone plugins for the modules that are currently part of the Performance Lab plugin, and then removing the corresponding modules from the PL plugin in favor of the new standalone plugins.

## Milestones

This effort should be broken down into 2 milestones:
1. Publish the modules available as standalone plugins
2. Remove the modules from the plugin

The first milestone will unblock the more urgent goal of having standalone plugins for every performance feature project, as well as therefore having a clear path forward also for new features.

The second milestone will ensure those standalone plugins are the canonical way to test those features, with the Performance Lab plugin still providing supporting features e.g. related to performance measurement and management of the performance related features and their plugins.

While the first milestone does not change anything notable for existing Performance Lab plugin users, the second one will require administrators to install/activate the new plugins and can therefore be considered a breaking change. As such, the second milestone should eventually be released with a major version number change, as `3.0.0`.

## List of issues

The list below includes all related issues, somewhat ordered by their intended sequence of execution. Important to note:
* Some of the issues can be worked on in parallel.
* The issue for milestone 2 can be defined as milestone 1 is being worked on, however milestone 2 issues should only be implemented / merged after milestone 1 has been completed.
* The list is not necessarily comprehensive, more issues will be added as needed.
* You can always get a list of issues also using the ["Creating standalone plugins" label](https://github.com/WordPress/performance/labels/Creating%20standalone%20plugins).

### Milestone 1

* [x] #635
* [x] #639
* [x] #636
* [x] #666
* [x] #667
* [x] #637
* [x] #638
* [x] #692
* [x] #640
* [x] ~~#661~~

#### 2a (2.8.0 release)
* [x] #651
* [x] #652
* [x] #880
* [x] #653
* [x] #901

Follow-up PRs:
* #896
* #912
* #917
* #920

#### 2b (3.0.0 release)
* [x] #934
* [x] #654
* [x] #935
* [x] #655

#### 2c (3.0.0 release polishing)
* [x] #1029
* [x] #1030
* [x] #1031
* [x] #1045

Follow-up Issues & PRs:
* [x] #997
* [x] #998
* [x] #1006
* [x] #904
* [x] #972
* [x] #971
* [x] #999
* [x] #1014
* [x] #1020
* [x] #1021 
* [x] #1026

[...]

PR #664, opened 2023-03-02
Feature/ #636 Module Specific Readme Files

10upsimon:
## Summary

Addresses #636 

## Relevant technical choices

- Adds a draft v1 `readme.txt` file to the `webp-uploads` module folder for copying over to the relevant build directory as per the work carried out in #662 .
- Flags all `readme.txt` files in all performance plugin module folders for `export-ignore` via `.gitattributes` glob pattern match.

## Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -31,10 +31,4 @@
 /.gitattributes export-ignore
 /.gitignore export-ignore
 
-/modules/database/audit-autoloaded-options/readme.txt export-ignore
-modules/database/sqlite/readme.txt export-ignore
-/modules/images/dominant-color/readme.txt export-ignore
-/modules/images/fetchpriority/readme.txt export-ignore
-/modules/images/webp-support/readme.txt export-ignore
-/modules/images/webp-uploads/readme.txt export-ignore
-/modules/js-and-css/audit-enqueued-assets/readme.txt export-ignore
+/modules/**/readme.txt export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -30,3 +30,11 @@
 
 /.gitattributes export-ignore
 /.gitignore export-ignore
+
+/modules/database/audit-autoloaded-options/readme.txt export-ignore
+modules/database/sqlite/readme.txt export-ignore
+/modules/images/dominant-color/readme.txt export-ignore
+/modules/images/fetchpriority/readme.txt export-ignore
+/modules/images/webp-support/readme.txt export-ignore
+/modules/images/webp-uploads/readme.txt export-ignore
+/modules/js-and-css/audit-enqueued-assets/readme.txt export-ignore
--- a/modules/images/webp-uploads/readme.txt
+++ b/modules/images/webp-uploads/readme.txt
@@ -1,4 +1,4 @@
-=== Performance Lab - WebP Uploads Module ===
+=== WebP Uploads ===
 
 Contributors:      wordpressdotorg
 Requires at least: 6.1
@@ -7,9 +7,9 @@ Requires PHP:      5.6
 Stable tag:        1.0.0
 License:           GPLv2 or later
 License URI:       https://www.gnu.org/licenses/gpl-2.0.html
-Tags:              performance, images, site health, measurement, webp
+Tags:              performance, images, webp
 
-WebP uploads module for the [performance plugin](https://wordpress.org/plugins/performance-lab/) from the WordPress Performance Team.
+Creates WebP versions for new JPEG image uploads if supported by the server.
 
 == Description ==
 
@@ -22,34 +22,20 @@ This plugin is intended to be used alongside the [performance plugin](https://wo
 = Installation from within WordPress =
 
 1. Visit **Plugins > Add New**.
-2. Search for **Performance Lab - WebP Uploads Module**.
-3. Install and activate the **Performance Lab - WebP Uploads Module** plugin.
+2. Search for **WebP Uploads**.
+3. Install and activate the **WebP Uploads** plugin.
 
 = Manual installation =
 
-1. Upload the entire `performance-lab-webp-support-module` folder to the `/wp-content/plugins/` directory.
+1. Upload the entire `webp-uploads` folder to the `/wp-content/plugins/` directory.
 2. Visit **Plugins**.
-3. Activate the **Performance Lab - WebP Uploads Module** plugin.
-
-= After activation =
-
-1. Visit the new **Settings > Performance** menu.
-2. Enable the individual modules you would like to use, one of which should be this WebP support module.
+3. Activate the **WebP Uploads** plugin.
 
 == Frequently Asked Questions ==
 
-= What is the purpose of this plugin? =
-
-The primary purpose of the Performance Lab - WebP Support Module plugin is to add an additional module to the performance plugin that enables the ability to toggle on/off WebP support.
-
-= Can I use this plugin on my production site? =
-
-Per the primary purpose of the plugin (see above), it can be considered stable for production usage given that this particular module is not marked as experimental.
-
 = Where can I submit my plugin feedback? =
 
-Feedback is encouraged and much appreciated, especially since this plugin may contain future WordPress core features. If you have suggestions or requests for new features, you can [submit them as an issue in the Performance Lab GitHub repository](https://github.com/WordPress/performance/issues/new/choose). If you need help with troubleshooting or have a question about the plugin, please [create a new topic on our support forum](https://wordpress.org/support/plugin/performance-lab/#new-topic-0).
-
+Feedback is encouraged and much appreciated, especially since this plugin may contain future WordPress core features. If you have suggestions or requests for new features, you can [submit them as an issue in the WordPress Performance Team's GitHub repository](https://github.com/WordPress/performance/issues/new/choose). If you need help with troubleshooting or have a question about the plugin, please [create a new topic on our support forum](https://wordpress.org/support/plugin/webp-uploads/#new-topic-0).
 = How can I contribute to the plugin? =
 
 Contributions are always welcome! Learn more about how to get involved in the [Core Performance Team Handbook](https://make.wordpress.org/performance/handbook/get-involved/).
--- a/modules/images/webp-uploads/readme.txt
+++ b/modules/images/webp-uploads/readme.txt
@@ -36,9 +36,21 @@ This plugin is intended to be used alongside the [performance plugin](https://wo
 = Where can I submit my plugin feedback? =
 
 Feedback is encouraged and much appreciated, especially since this plugin may contain future WordPress core features. If you have suggestions or requests for new features, you can [submit them as an issue in the WordPress Performance Team's GitHub repository](https://github.com/WordPress/performance/issues/new/choose). If you need help with troubleshooting or have a question about the plugin, please [create a new topic on our support forum](https://wordpress.org/support/plugin/webp-uploads/#new-topic-0).
+
 = How can I contribute to the plugin? =
 
 Contributions are always welcome! Learn more about how to get involved in the [Core Performance Team Handbook](https://make.wordpress.org/performance/handbook/get-involved/).
+
+= I've activated the WebP Uploads module, but WebP images are not always generated when I upload a JPEG image. Why? =
+
+There are two primary reasons that a WebP image may not be generated:
+
+1. The WebP Uploads plugin has identified that the WebP version of the uploaded JPEG image would have a larger file size than the original JPEG image, so it does not generate the WebP version.
+2. The JPEG image was not uploaded to the [Media Library](https://wordpress.com/support/media/). At this time, WebP versions are only generated for images to the Media Library. WebP versions are not generated for JPEG images that are added to your site in other ways, such as in a template file or the [Customizer](https://wordpress.com/support/customizer/).
+
+= With the WebP Uploads plugin activated, will the plugin generate JPEG and WebP versions of every image that I upload? =
+
+By default, the WebP Uploads plugin will only generate WebP versions of the images that you upload. If you wish to have both WebP **and** JPEG versions generated, you can navigate to **Settings > Media** and enable the **Generate JPEG files in addition to WebP** option.
 
 == Changelog ==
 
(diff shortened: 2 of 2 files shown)


.gitattributes
```diff
@@ -30,3 +30,11 @@
 
 /.gitattributes export-ignore
 /.gitignore export-ignore
+
+/modules/database/audit-autoloaded-options/readme.txt export-ignore
+modules/database/sqlite/readme.txt export-ignore
+/modules/images/dominant-color/readme.txt export-ignore
+/modules/images/fetchpriority/readme.txt export-ignore
+/modules/images/webp-support/readme.txt export-ignore
+/modules/images/webp-uploads/readme.txt export-ignore
```
felixarntz: The requirements state to only add a readme for WebP Uploads as part of this issue. Let's stick to that and ignore all other modules.

felixarntz (commented):
@10upsimon Can you please limit the changes here to just the WebP Uploads module? Other modules to deploy as standalone plugins will be handled in a separate issue.

.gitattributes
10upsimon: @felixarntz  I've changed this to a single 1 line glob pattern now, based on @joemcgill 's feedback so that we can match current and future `readme.txt` files

```
/modules/**/readme.txt export-ignore
```

Let me know if this is good, or if you want me to specifically define it as webp-uploads only, and single out modules in the future.

10upsimon:
@felixarntz I've addressed your comment RE restricting this to `webp-uploads` only, however I did implement a catch-all glob pattern match for all module `readme.txt` files in order to `export-ignore` them.

Once @mukeshpanchal27 's PR at #662 is approved and merged, I'll enhance the CLI command to handle the copy of the `readme.txt` files.

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,63 @@
...
+Contributors:      wordpressdotorg
+Requires at least: 6.1
+Tested up to:      6.1
+Requires PHP:      5.6
+Stable tag:        1.0.0
+License:           GPLv2 or later
+License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+Tags:              performance, images, site health, measurement, webp
+
+WebP uploads module for the [performance plugin](https://wordpress.org/plugins/performance-lab/) from the WordPress Performance Team.
```
felixarntz: I don't think we should mention the Performance Lab plugin here given that this is the description for the standalone plugin. I think we can use the description that is also present in the module header:
```suggestion
Creates WebP versions for new JPEG image uploads if supported by the server.
```

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,63 @@
...
+
+== Description ==
+
+This plugin adds a WebP uploads module to the [performance plugin](https://wordpress.org/plugins/performance-lab/), enabling WebP support for media uploads within the WordPress application, when this module is enabled.
+
+This plugin is intended to be used alongside the [performance plugin](https://wordpress.org/plugins/performance-lab/), and will not function as a standalone plugin. The performance plugin may be obtained [here](https://wordpress.org/plugins/performance-lab/).
```
felixarntz: The latter is not correct, this plugin _will_ function as a standalone plugin. That's part of why we're breaking them apart.

I would avoid the mention of the Performance Lab plugin here and focus purely on a brief description of what this plugin does.

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,63 @@
+=== Performance Lab - WebP Uploads Module ===
```
felixarntz: The plugin should be called like the module, "WebP Uploads". It does not require Performance Lab, so shouldn't look as if it was an extension.

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,63 @@
...
+
+1. Visit **Plugins > Add New**.
+2. Search for **Performance Lab - WebP Uploads Module**.
+3. Install and activate the **Performance Lab - WebP Uploads Module** plugin.
+
+= Manual installation =
+
+1. Upload the entire `performance-lab-webp-support-module` folder to the `/wp-content/plugins/` directory.
+2. Visit **Plugins**.
+3. Activate the **Performance Lab - WebP Uploads Module** plugin.
```
felixarntz: See above:
```suggestion
= Installation from within WordPress =

1. Visit **Plugins > Add New**.
2. Search for **WebP Uploads**.
3. Install and activate the **WebP Uploads** plugin.

= Manual installation =

1. Upload the entire `webp-uploads` folder to the `/wp-content/plugins/` directory.
2. Visit **Plugins**.
3. Activate the **WebP Uploads** plugin.
```

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,63 @@
...
+
+= After activation =
+
+1. Visit the new **Settings > Performance** menu.
+2. Enable the individual modules you would like to use, one of which should be this WebP support module.
```
felixarntz: This should be removed, as the plugin will work as soon as it's activated.

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,63 @@
...
+
+== Frequently Asked Questions ==
+
+= What is the purpose of this plugin? =
+
+The primary purpose of the Performance Lab - WebP Support Module plugin is to add an additional module to the performance plugin that enables the ability to toggle on/off WebP support.
+
```
felixarntz: I think we can remove this.

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,63 @@
...
+= Can I use this plugin on my production site? =
+
+Per the primary purpose of the plugin (see above), it can be considered stable for production usage given that this particular module is not marked as experimental.
```
felixarntz: This refers to a module, which it shouldn't. The standalone plugin will have no notion of what is experimental or not.

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,63 @@
...
+
+= Where can I submit my plugin feedback? =
+
+Feedback is encouraged and much appreciated, especially since this plugin may contain future WordPress core features. If you have suggestions or requests for new features, you can [submit them as an issue in the Performance Lab GitHub repository](https://github.com/WordPress/performance/issues/new/choose). If you need help with troubleshooting or have a question about the plugin, please [create a new topic on our support forum](https://wordpress.org/support/plugin/performance-lab/#new-topic-0).
```
felixarntz:
```suggestion
Feedback is encouraged and much appreciated, especially since this plugin may contain future WordPress core features. If you have suggestions or requests for new features, you can [submit them as an issue in the WordPress Performance Team's GitHub repository](https://github.com/WordPress/performance/issues/new/choose). If you need help with troubleshooting or have a question about the plugin, please [create a new topic on our support forum](https://wordpress.org/support/plugin/webp-uploads/#new-topic-0).
```

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,63 @@
...
+
+= How can I contribute to the plugin? =
+
+Contributions are always welcome! Learn more about how to get involved in the [Core Performance Team Handbook](https://make.wordpress.org/performance/handbook/get-involved/).
+
```
felixarntz, line 52: Can you add (slightly modified versions of) the two WebP Uploads specific entries in our current Performance Lab `readme.txt` here? Those are relevant and should be part of this readme. The mentioning of activating the module in Performance Lab should be removed from them though, as just by activating this plugin on its own those two questions may be relevant.

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,63 @@
...
+== Changelog ==
+
+= 1.0.0 =
+
+**Initial Release**
+
+* Initial release of the WebP support performance plugin module as a standalone plugin. ([636](https://github.com/WordPress/performance/issues/636))
```
felixarntz: We typically refer to PRs in the changelog instead of issues.
```suggestion
* Initial release of the WebP Uploads plugin as a standalone plugin. ([664](https://github.com/WordPress/performance/pull/664))
```

felixarntz (commented):
@10upsimon Thank you for the updates. I think there is a misunderstanding here, as the standalone plugins will be in fact standalone. They should not be branded/marketed as extensions to the Performance Lab plugin.

Right now, as part of Milestone 1 (see #656), they will exist in both places, as a standalone plugin and as a module in the Performance Lab plugin. But once we're done with Milestone 2, they will only exist as standalone plugin.

Using Performance Lab alongside them is perfectly fine and we would still encourage users to use Performance Lab if they want to follow along what the performance team is working on. But someone that wants to just use WebP would be perfectly fine to only install this plugin  on its own.

Regarding the dependency on #662, I don't think we technically have to wait for that PR. Once this PR is finalized, it can be merged as is, since #662 will copy the entire directory already, so the `readme.txt` here will be included.

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,63 @@
+=== Performance Lab - WebP Uploads Module ===
+
+Contributors:      wordpressdotorg
+Requires at least: 6.1
+Tested up to:      6.1
+Requires PHP:      5.6
+Stable tag:        1.0.0
+License:           GPLv2 or later
+License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+Tags:              performance, images, site health, measurement, webp
```
mukeshpanchal27:
```suggestion
Tags:              performance, images, webp
```

I don't think we have to add `site health` and `measurement` for WebP Uploads

mukeshpanchal27 (commented):
Thanks @10upsimon, I left one minor suggestion, and could you please remove the `readme.txt` file for sqlite?

10upsimon:
@mukeshpanchal27 @felixarntz many thanks for your feedback and comments. I believe I have addressed them all and have requested a re-review of the PR. I confirm the readme.txt was copied across as expected with @mukeshpanchal27 latest merge of #662  @felixarntz

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,60 @@
...
+License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+Tags:              performance, images, webp
+
+Creates WebP versions for new JPEG image uploads if supported by the server.
+
+== Description ==
+
+This plugin adds WebP support for media uploads within the WordPress application.
+
+
```
mukeshpanchal27, line 17:
```suggestion
```
Remove extra line.

mukeshpanchal27 (approved):
Thanks @10upsimon, The changes look good to me.

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,59 @@
...
+Stable tag:        1.0.0
+License:           GPLv2 or later
+License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+Tags:              performance, images, webp
+
+Creates WebP versions for new JPEG image uploads if supported by the server.
+
+== Description ==
+
+This plugin adds WebP support for media uploads within the WordPress application.
```
felixarntz: Technically not a blocker, but since this is pretty much the only module / plugin that has a UI to it, we still may want to provide at least a bit more explanation in 2-3 short paragraphs. Things to potentially highlight:
* WebP images will be generated only for new uploads
* By default, only WebP images will be generated for JPEG uploads, only the original uploaded file will still exist as JPEG
* There is a checkbox in `Settings > Media` that can change the behavior though to generate _both_ JPEG and WebP for every sub-size (noting again that this will only affect newly uploaded images, i.e. after making that change)

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,59 @@
...
+
+= With the WebP Uploads plugin activated, will the plugin generate JPEG and WebP versions of every image that I upload? =
+
+By default, the WebP Uploads plugin will only generate WebP versions of the images that you upload. If you wish to have both WebP **and** JPEG versions generated, you can navigate to **Settings > Media** and enable the **Generate JPEG files in addition to WebP** option.
+
+== Changelog ==
+
+= 1.0.0 =
+
+**Initial Release**
```
felixarntz: No need for an extra "heading" here
```suggestion
= 1.0.0 =
```

felixarntz (commented):
@10upsimon This is pretty much good to go now, just one small thing and one slightly bigger thing. It would be great if you could expand the description just a tiny bit to explain a bit more clearly what the plugin does.

We can also iterate in a follow up if you prefer for just that content, in order to unblock this PR. Let me know what you prefer.

modules/images/webp-uploads/readme.txt
10upsimon: @felixarntz (cc @mukeshpanchal27 ) this is ready for re-review following changes pushed in recent commit(s).

modules/images/webp-uploads/readme.txt
```diff
@@ -0,0 +1,57 @@
...
+Stable tag:        1.0.0
+License:           GPLv2 or later
+License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+Tags:              performance, images, webp
+
+Creates WebP versions for new JPEG image uploads if supported by the server.
+
+== Description ==
+
+This plugin adds WebP support for media uploads within the WordPress application. WebP images will be generated only for new uploads, preexisting imagery will not be converted to WebP format. By default, WebP images will only be generated for JPEG uploads, only the original uploaded file will still exist as a JPEG image. ALl generated image sizes will exist as WebP only. If you wish to change this behaviour, there is a checkbox in `Settings > Media` that - when checked - will alter the behaviour of this plugin to generate both JPEG and WebP images for every sub-size (noting again that this will only affect newly uploaded images, i.e. after making said change).
```
mukeshpanchal27:
```suggestion
This plugin adds WebP support for media uploads within the WordPress application. WebP images will be generated only for new uploads, pre-existing imagery will not be converted to WebP format. By default, WebP images will only be generated for JPEG uploads, only the original uploaded file will still exist as a JPEG image. All generated image sizes will exist as WebP only. If you wish to change this behaviour, there is a checkbox in `Settings > Media` that - when checked - will alter the behaviour of this plugin to generate both JPEG and WebP images for every sub-size (noting again that this will only affect newly uploaded images, i.e. after making said change).
```

10upsimon: Thanks @mukeshpanchal27 , updated

felixarntz (approved):
@10upsimon Thank you for the updates, LGTM!

PR #669, opened 2023-03-10
Add `wp-total` metric to default Server-Timing metrics

### LINKED ISSUE #668: Add Server-Timing metric for total load time
## Feature Description

The Server Timing API in the plugin currently exposes metrics `wp-before-template` and `wp-template`, which when added make up for the total response time. While it's easy to calculate the sum of the two, when running many performance tests in an automated manner, this is not realistically feasible.

We should avoid using the sum of the two medians as that is statistically incorrect. We need to instead have the total load time in every run and use the median of that.

A `wp-total` metric was already included in the WordPress core automated performance CI workflow (see https://core.trac.wordpress.org/browser/trunk/tests/performance/wp-content/mu-plugins/server-timing.php?rev=55459#L26), so we should implement that in the Performance Lab plugin too.

felixarntz:
## Summary

Fixes #668

* Note that in order to see the metric, you need to enable output buffering via the Server Timing API filter. See [this Gist](https://gist.github.com/felixarntz/9c3d7150c74082e69bb426393b68b12e).
* [This data](https://docs.google.com/spreadsheets/d/1Oe-6ZO_Jqt6qSp2yF-ohb05McEB2y22aqU0S7gXuyGs/edit#gid=1501750953) (also see [WordPress Trac ticket](https://core.trac.wordpress.org/ticket/57814)) is an example where the code from this PR was already used.

## Checklist

### DIFF
--- a/server-timing/defaults.php
+++ b/server-timing/defaults.php
@@ -152,6 +152,23 @@ function perflab_register_default_server_timing_template_metrics() {
 		PHP_INT_MAX
 	);
 
+	add_action(
+		'perflab_server_timing_send_header',
+		function() {
+			// WordPress total load time.
+			perflab_server_timing_register_metric(
+				'total',
+				array(
+					'measure_callback' => function( $metric ) {
+						// The 'timestart' global is set right at the beginning of WordPress execution.
+						$metric->set_value( ( microtime( true ) - $GLOBALS['timestart'] ) * 1000.0 );
+					},
+					'access_cap'       => 'exist',
+				)
+			);
+		}
+	);
+
 	// SQL query time is only measured if the SAVEQUERIES constant is set to true.
 	if ( defined( 'SAVEQUERIES' ) && SAVEQUERIES ) {
 		add_action(


felixarntz:
cc @oandregal

mukeshpanchal27 (approved):
Thanks @felixarntz, LGTM!

server-timing/defaults.php
```diff
@@ -152,6 +152,23 @@ function( $passthrough = null ) {
 		PHP_INT_MAX
 	);
 
+	add_action(
```
joemcgill, line 155: I'd prefer to see this implemented as a named callback that could more easily be removed on sites that want to run the plugin but not expose this data. Also, I think we should be consistent about prefixing our metrics with `wp-` or similar, to avoid potential collisions if other tools are adding data to the Server Timing header.

I also notice that the [MDN docs for Server Timing](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing) suggest using a semi-colon separated syntax for reporting these metrics. Not sure if this is a convention that we should be following or not, but it would suggest that we do something like:

```
Server-Timing: wp-total;desc="WordPress Total";dur={value}
```
or (simplified)

```
Server-Timing: wp-total;dur={value}
```

Curious what your opinion is of those conventions.

felixarntz: @joemcgill Replying to your points:
* If there is a need to remove metrics, I'd prefer we handle that holistically in a separate issue. The Server-Timing header is controlled via an API, and I'm not sure that removing hooks is the cleanest way to go about it. More importantly though, for this specific metric, I don't see why someone would want to remove `wp-total` when that's the same as `wp-before-template + wp-template` anyway, which would still not be possible to be removed. That's why I think we should rather approach it holistically.
* The Server-Timing API enforces the `wp-` prefix, so it will show as you mention :)
* Similarly, the Server-Timing API controls the exact output, and it will be output exactly like you're saying, following the conventions (your 2nd example, since we don't use metric descriptions at this point).

joemcgill: Thanks for your responses, @felixarntz. I had wondered if points 2 and 3 were handled elsewhere in the process. Glad to know that it is. I do think we should revisit how certain metrics can be managed and am fine for that to be done outside the scope of this issue.

joemcgill (approved):

PR #670, opened 2023-03-13
Check for each module to only load it if the corresponding standalone plugin is not loaded

### LINKED ISSUE #666: Check for each module to only load it if the corresponding standalone plugin is not loaded
## Feature Description

As part of #656, a temporary state between completion of Milestone 1 and Milestone 2 will be that certain features will exist both as a Performance Lab module and as a standalone plugin. We should make sure there are no conflicts. Specifically, we can ensure that, if a certain standalone plugin is active, we don't load or allow enabling the corresponding Performance Lab module.

### Requirements

* The function that loads the active modules should for each module check whether it also exists as a standalone plugin and whether that standalone plugin is active. If yes, the respective module should not be loaded.
* The same check should happen on the Performance Lab admin screen. Any module where the standalone plugin is active should as disabled and with a message similar to when it cannot be loaded for another reason: `The module cannot be managed with Performance Lab since it is already active as a standalone plugin.`

mukeshpanchal27:
## Summary

Fixes #666 

## Relevant technical choices
The PR changes only work when the standalone plugin is activated, and we try to activate the Performance Lab plugin. The reverse process is not working, as it generates a fatal error for a redeclare function error because all the functions are duplicated in the PL module.

## Checklist


load.php
```diff
@@ -162,9 +162,36 @@ function( $module_settings ) {
 	 */
 	$modules = apply_filters( 'perflab_active_modules', $modules );
 
+	if ( ! function_exists( 'is_plugin_active' ) ) {
+		include_once( ABSPATH . 'wp-admin/includes/plugin.php' );
+	}
+
+	$standalone_plugin_config = perflab_get_standalone_plugins_config();
+	foreach ( $standalone_plugin_config as $plugin_slug => $plugin_main_file ) {
+		if ( in_array( $plugin_slug, $modules, true ) && is_plugin_active( $plugin_main_file ) ) {
```
felixarntz: The `is_plugin_active()` function is not available in frontend requests, so we shouldn't use it. I see you load it above, but that's not a good practice, loading admin functions in every frontend request. We have to find another solution.

I wonder, maybe the most reliable is to introduce a constant at the top of each module's `load.php` file (e.g. `WEBP_UPLOADS_VERSION`, `SQLITE_INTEGRATION_VERSION` etc.) and then we could check whether the constant is already defined in this logic.

While it involves adding the constants to the modules / plugins, I think this will be the most reliable approach. Sometimes plugins can be loaded in another folder than their expected slug, in which case the constant approach will still work, but relying on the plugin slug won't.

load.php
```diff
@@ -162,9 +162,36 @@ function( $module_settings ) {
...
+			$key = array_search( $plugin_slug, $modules, true );
+			if ( false !== $key ) {
+				unset( $modules[ $key ] );
+			}
+		}
+	}
+
```
felixarntz: This logic shouldn't be added to this function. Even when a standalone plugin is loaded, the module itself is technically still active, so this function should continue returning it. Maybe we can instead add it to `perflab_can_load_module()`? That would make more sense contextually.

Since we need a way to know _why_ a module cannot be loaded then outside of that function, I would suggest we modify the `@return` type of `perflab_can_load_module()` to return `bool|WP_Error`. This way, we can provide an error code and message for the different reasons that the module cannot be loaded. For example, when `can-load.php` returns `false`, it would be a different reason than when the standalone plugin is already loaded. WDYT?

load.php
```diff
@@ -162,9 +162,36 @@ function( $module_settings ) {
...
+/**
+ * Gets the standalone plugin configuration.
+ *
+ * @since n.e.x.t
+ *
+ * @return array Array of standalone plugin configuration.
+ */
+function perflab_get_standalone_plugins_config() {
+	return array(
+		'images/webp-uploads' => 'webp-uploads/load.php',
```
felixarntz: See above. If we go with the approach I suggested there, we could change this function to return a map of module paths to their constants, rather than a map to their standalone plugin basenames.

admin/load.php
```diff
@@ -126,16 +126,18 @@ function perflab_render_modules_page() {
...
+	$plugins_config   = perflab_get_standalone_plugins_config();
+	$is_active_plugin = ( isset( $plugins_config[ $module_slug ] ) && ( '' !== $plugins_config[ $module_slug ] ) && is_plugin_active( $plugins_config[ $module_slug ] ) ) ? true : false;
 	?>
 	<fieldset>
 		<legend class="screen-reader-text">
 			<?php echo esc_html( $module_data['name'] ); ?>
 		</legend>
 		<label for="<?php echo esc_attr( "{$base_id}_enabled" ); ?>">
-			<?php if ( perflab_can_load_module( $module_slug ) ) { ?>
+			<?php if ( perflab_can_load_module( $module_slug ) && ! $is_active_plugin ) { ?>
```
felixarntz: See my other comment: If we go with the constants approach and with modifying `perflab_can_load_module()`, we could change the logic here:
* If `perflab_can_load_module()` returns a `WP_Error` or `false`, it means the module cannot be loaded.
* If a `WP_Error` is returned, we should then display the error message from that error. And only show the default fallback messages below if `false` is returned.

felixarntz (commented):
@mukeshpanchal27 I have a suggestion here how we can make this more resilient and also refine our `perflab_can_load_module()` function to have it provide more context on why a module cannot load.

load.php
mukeshpanchal27: @felixarntz I try the approach you suggested for constant but i fall in same error when i try to active PL plugin after the WebP Uploads plugin.

Even if i define constant at top of the file i get same duplicate function error. Please guide me if i'm on wrong track or did not get your feedback

felixarntz: @mukeshpanchal27 Sorry, I meant the constant should be defined in a way that they don't conflict:
```php
if ( ! defined( 'WEBP_UPLOADS_VERSION' ) ) {
    defined( 'WEBP_UPLOADS_VERSION', 'the.version.number' );
}
```

mukeshpanchal27: Thanks @felixarntz, I have implemented constant approach as per suggestion in  https://github.com/WordPress/performance/pull/670/commits/f93cf0f9ca5edeffcca3b6bf408bde4e49d45b56.

When i first activate `WebP Uploads` plugin then `Performance Lab`, It still get defined value for PL plugin instead of the `WebP Uploads`. So WP load `Performance Lab` plugin first then `WebP Uploads`.

felixarntz: @mukeshpanchal27 Ahh right, I forgot about that. We need to change the function call to load the modules to not happen immediately in `load.php`, but on the `plugins_loaded` action instead.

felixarntz:
@mukeshpanchal27 This PR now includes a bunch of changes that are originally from the other PR #674. We should instead refresh the feature branch from `trunk` where that PR was merged into. I opened #676 for that, please review.

Let's merge that PR first, then you can refresh your PR here against the latest feature branch so that the PR here is "clean" again.

load.php
mukeshpanchal27: Thanks @felixarntz, I have implemented it in https://github.com/WordPress/performance/pull/670/commits/803a26afb68e61ce7d2b30637cdfd9174bc8dd31 and it is working fine for frontend requests.

load.php
mukeshpanchal27: Agreed and updated based on your feedback.

admin/load.php
mukeshpanchal27: @felixarntz The constant logic did not work here because the module `load.php` file ran before this check, so it was for the single Performance Lab plugin, and it shows `The module cannot be managed with Performance Lab since it is already active as a standalone plugin.`

felixarntz: @mukeshpanchal27 Even after you did what I recommended in https://github.com/WordPress/performance/pull/670#discussion_r1137503887?

felixarntz:
@mukeshpanchal27 Now that the feature branch is refreshed with `trunk`, can you update this branch and fix the one merge conflict?

admin/load.php
mukeshpanchal27: @felixarntz Yes. Please accept my apologies if I misunderstood anything.It works for frontend requests, but when I check in the "Admin Panel > Settings > Performance", it shows me that `The module cannot be managed with Performance Lab since it is already active as a standalone plugin` This is true even if only the PL plugin is activated.

Can you please take a look. Thanks! 

I have merged the branch and resolved the conflict.

load.php
```diff
@@ -262,7 +282,16 @@ function perflab_load_active_and_valid_modules() {
...
+
+/**
+ * Loads the modules once the activated plugins have loaded.
+ *
+ * @since n.e.x.t
+ */
+function perflab_plugins_loaded() {
+	perflab_load_active_and_valid_modules();
+}
+add_action( 'plugins_loaded', 'perflab_plugins_loaded' );
```
felixarntz: Why introduce a new function that does exactly the same as the existing function?
```suggestion
add_action( 'plugins_loaded', 'perflab_load_active_and_valid_modules' );
```

modules/images/webp-uploads/load.php
```diff
@@ -8,6 +8,13 @@
  * @package performance-lab
  */
 
+// Define the constant.
+if ( defined( 'WEBP_UPLOADS_VERSION' ) ) {
+	return;
+}
+
+define( 'WEBP_UPLOADS_VERSION', '1.0.0' );
```
felixarntz: I think we should set this to a version for the module, since it's in the module file, not the plugin:
```suggestion
define( 'WEBP_UPLOADS_VERSION', 'Performance Lab ' . PERFLAB_VERSION );
```

What that means though is that we should replace the version in this constant with the standalone plugin version as part of the build script. Can you open a separate follow up PR to do that? Let's not block this one on this new requirement.

Basically the new requirement for the separate PR should be:
* For any standalone plugin, the version constant name that it uses should be added to the `plugins.json` definition.
* The `build-plugins` command should be enhanced to replace the version set in the constant with the current plugin version, e.g. `WEBP_UPLOADS_VERSION` value should change from `Performance Lab 2.1.0` to `1.0.0`.

load.php
```diff
@@ -236,6 +236,13 @@ function perflab_can_load_module( $module ) {
 		return true;
 	}
 
+	$standalone_plugin_config = perflab_get_standalone_plugins_config();
+	foreach ( $standalone_plugin_config as $slug => $constant ) {
+		if ( $module === $slug && defined( $constant ) ) {
+			return new WP_Error( 'standalone_plugin_activated', __( 'The module cannot be managed with Performance Lab since it is already active as a standalone plugin.', 'performance-lab' ) );
+		}
+	}
```
felixarntz: A `foreach` here is not good for performance and shouldn't be necessary. You only need to check for the slug.
```suggestion
	if ( isset( $standalone_plugin_config[ $module ] && defined( $standalone_plugin_config[ $module ] ) ) {
		return new WP_Error( 'standalone_plugin_activated', __( 'The module cannot be managed with Performance Lab since it is already active as a standalone plugin.', 'performance-lab' ) );
	}
```

admin/load.php
```diff
@@ -170,6 +171,8 @@ function perflab_render_modules_page_field( $module_slug, $module_data, $module_
 					);
 				} elseif ( 'database/sqlite' === $module_slug && ! class_exists( 'SQLite3' ) ) {
 					esc_html_e( 'The SQLite module cannot be activated because the SQLite extension is not loaded.', 'performance-lab' );
+				} elseif ( is_wp_error( $can_load_module ) ) {
+					esc_html_e( 'The module cannot be managed with Performance Lab since it is already active as a standalone plugin.', 'performance-lab' );
```
felixarntz: This condition should be the very first one in the list, as it is the most specific ones. All other messages here are fallbacks for when there is no `WP_Error` being returned.

Also, please don't hardcode the message here, as that misses the point of using a `WP_Error`. We should display `$can_load_module->get_error_message()` here.

load.php
```diff
@@ -248,6 +255,19 @@ function perflab_can_load_module( $module ) {
...
 }
 
+/**
+ * Gets the standalone plugin configuration.
+ *
+ * @since n.e.x.t
+ *
+ * @return array Array of standalone plugin configuration.
+ */
+function perflab_get_standalone_plugins_config() {
```
felixarntz: Maybe we can clarify what this is for?
```suggestion
/**
 * Gets the standalone plugin constants used for each module / plugin.
 *
 * @since n.e.x.t
 *
 * @return array Map of module path to version constant used.
 */
function perflab_get_standalone_plugins_constants() {
```

modules/images/webp-uploads/load.php
```diff
@@ -8,6 +8,13 @@
...
+
 // Do not load the code if it is already loaded through another means.
 if ( function_exists( 'webp_uploads_create_sources_property' ) ) {
 	return;
```
felixarntz, line 20: This can be removed now that the constant is already being defined and checked for. No need to check for two things from the module that are already present, one is sufficient. Maybe move the comment here above the `defined( ...` check?

felixarntz (commented):
@mukeshpanchal27 A few last things here, but functionally this already works as expected. Great work!

admin/load.php
felixarntz: @mukeshpanchal27 That is functioning as expected: If the WebP Uploads plugin is active, the module checkbox _should_ be disabled with the message you're saying. So this is all good as is.

load.php
```diff
@@ -226,7 +226,7 @@ function perflab_render_generator() {
  * @since 1.3.0
  *
  * @param string $module Slug of the module.
- * @return bool Whether the module can be loaded or not.
+ * @return bool|WP_Error Whether the module can be loaded or not, otherwise a WP_Error.
```
joemcgill: I personally don't like the idea of overloading this function to return an error if the module is already loaded via a standalone plugin. Instead, this function should return false if the module code is already loaded.

Here's what I would propose:

1. Create a new function, like `perflab_is_module_standalone( string $slug ): bool {}` that checks whether the module slug's constant is already defined.
2. Use this function inside `perflab_can_load_module()` to decide whether to return `false` in place of the current logic that throws a WP_Error.
3. Replace the else condition in `perflab_render_modules_page_field` that looks for WP_Errors with a new condition that checks for a false response from `perflab_is_module_standalone` and displays a message that says the module can't be activated because the code is being loaded by a separate plugin.

modules/images/webp-uploads/load.php
joemcgill: I'm a big fan of trying to automatically set these constants via the `plugins.json` file so we're handling all configuration in one place. We could probably also automatically fill out the array in `perflab_get_standalone_plugins_config` using the same JSON file so the constant is always based on the slug of each plugin from that file.

modules/images/webp-uploads/load.php
joemcgill: @felixarntz is there a risk that we need to account for where someone has a version of a plugin installed previous to when these constants were implemented where there would be a conflict with an already defined function?

load.php
joemcgill: Another concern I have is that we are also using `perflab_can_load_module()` in `perflab_is_valid_module()` to determine if we should display the module as generator content in `perflab_get_generator_content()`. Wouldn't we still want to show that a feature is active on a site regardless of whether it's loaded as a Perf Lab module or a standalone plugin?

load.php
```diff
@@ -262,7 +282,16 @@ function perflab_load_active_and_valid_modules() {
...
 	}
 }
-perflab_load_active_and_valid_modules();
+
+/**
+ * Loads the modules once the activated plugins have loaded.
+ *
+ * @since n.e.x.t
+ */
+function perflab_plugins_loaded() {
```
joemcgill: I don't see a reason for adding a new callback here, let's just call `perflab_load_active_and_valid_modules` directly on `plugins_loaded`. Also, I worry that relying only on loading modules on `plugins_loaded` could allow for some edge cases where a standalone plugin could load after the module loads and still cause a fatal. It would be better if the modules themselves would short circuit if the constant were already defined.

load.php
felixarntz: I'm potentially okay not returning a `WP_Error`, but then I also think we shouldn't intertwine the two functions at all. I think a simple boolean lacks the necessary context we would want to provide in end-user facing messaging, which is why I suggested the `WP_Error` usage here.

Alternatively, we can always use only `bool`s, but then it should be individual functions that are individually called, and I don't think we should put the `perflab_is_module_standalone()` call into `perflab_can_load_module()` as otherwise the context is again intertwined and the return value does not convey _why_.

While the above might be okay, I still think even with that it has drawbacks compared to using a `WP_Error`: When you look at our `admin/load.php` file, you see all those manual messages in there for SQLite, which with a `WP_Error` return value we could work towards avoiding those. Those messages should rather be located where those problems are determined, not manually be re-checked when we already have that logic somewhere else. I think using `false` instead of `WP_Error` lacks context and will require us to repeat code unnecessarily. What is the concern about returning a `WP_Error`?

joemcgill: > What is the concern about returning a WP_Error?

I have two main concerns:

1. It complicates the function signature and purpose unnecessarily and requires any code using this function to check for a possible error message OR a boolean value, when the intent of the function is mainly to determine if a module can be loaded. Keeping the function simple makes it easier to reason about and easier to test to avoid future bugs.
2. `WP_Error` messages should be used when we actually _want_ to relay that an error has occurred. In this case, we're using it as a logical value in our code so that we can _avoid_ an error. That seems wrong.

I'd be happy to decouple all of this logic as well, if you'd prefer, I just think it would be weird for a function called `perflab_can_load_module()` to return `true` if we know that the code has already been loaded, which is why I'm proposing the approach I suggested.

modules/images/webp-uploads/load.php
mukeshpanchal27: Thanks @felixarntz, I open #681 as a follow-up PR that updates the constant value in `build-plugins` command.

mukeshpanchal27:
Thanks @felixarntz and @joemcgill for all your feedback and thoughts.

At the moment, I'm facing issues with the constant approach.

- Take a pull from this PR and spin up the ENV and check the plugin admin page( Admin Panel > Settings > Performance). You can see that the `WebP Uploads` module is disabled with the message "The module cannot be managed with Performance Lab since it is already active as a standalone plugin." Make sure only the PL plugin is activated.

It is because the Pl plugin loads all the modules in "load.php" in the "plugins_loaded` action, so "WEBP_UPLOADS_VERSION` is already defined for the admin screen.

I hope the above explanation helps to replicate the issue.

felixarntz:
@mukeshpanchal27 Thank you for clarifying, now I see what you mean. It's because once the admin page is loaded, the constant will always be defined.

I think what we could do is add another check alongside the constant check that the constant value does not start with "Performance Lab ":
* If it starts with "Performance Lab ", it means the module is loaded.
* If it doesn't start with "Performance Lab ", it means the standalone plugin is loaded. In that case the checkbox should be disabled.

joemcgill:
I'm curious if there's a way we could use the settings to determine whether the module is loaded from the performance lab plug-in versus being loaded from an external plug-in? Additionally if we discover that a conflict exists because the standalone plug-in is active we should disable the option from performance lab.

mukeshpanchal27:
@felixarntz Thanks for the feedback. It solve the issue for admin screen. PR is ready for review.

load.php
felixarntz: @joemcgill Your rationale for 1. and 2. makes sense. However we still need to think through how we can solve that alternatively in a way that also works and doesn't lose context. The purposes of the two calls are different:
* When we load the modules, we just want to know whether they can be loaded or not (which encompasses both checking a potential `can-load.php` file for the module and checking whether its code is already loaded by the standalone plugin). In that case `true` or `false` works well.
* When we show the WP admin screen, we want to disable the checkboxes for modules that cannot be loaded, and we also want to show an informative message for why they're not loaded. For that purpose, having only `true` or `false` is not enough.

> I'd be happy to decouple all of this logic as well, if you'd prefer, I just think it would be weird for a function called `perflab_can_load_module()` to return `true` if we know that the code has already been loaded, which is why I'm proposing the approach I suggested.

I'm not sure I understand the last part of what you mean here. As far as I can tell, we _wouldn't_ know whether the module code was already loaded on the WP Admin screen: For example, if the module is not active, this function wouldn't have been called anytime before in the same request. Or when the module _is_ active, we don't know if the code is loaded because of the module or because of its standalone plugin. That's why I think we need to call that function on the WP admin screen as well.

We could avoid the usage of `WP_Error` by having two separate functions `perflab_can_load_module()` (responsible for running the `can-load.php` logic) and `perflab_is_module_standalone()` (responsible for checking whether the module's standalone plugin is already loaded), and we would need to call both of them in both places. For the WP admin screen, we could then show different messages based on which function returns `false`.

That still keeps us limited though in providing more helpful messages for the `can-load.php` approach itself, the lack of which right now causes us to have this messy SQLite module specific logic hard-coded in the general WP admin screen code, just to show better messages based on duplicated logic. So I would like us to figure that one out before we discard the `WP_Error` approach.

Again, I agree the `WP_Error` approach is probably not the best approach, but we also shouldn't settle for an approach that lacks the flexibility that we're looking for.

modules/images/webp-uploads/load.php
felixarntz: @joemcgill That's a great point, I hadn't thought through this. These `function_exists` checks in the modules are just being introduced in next week's release 2.1.0, so they're also new. However, since the work here including the constants is only going into a feature branch, this won't ship in the same release, so your concern could indeed be valid.

We won't be releasing any standalone plugins before this feature branch gets merged back into `trunk`. So the standalone plugins will always have the constants and constant checks from the start. The load order is generally that plugins load first, before the PL modules. So "older" versions of Performance Lab ("older" only in the future, e.g. 2.1.0+ after we have started to publish standalone plugins) would have the `function_exists` check in the modules, and since the function would exist from the standalone plugin loaded before, no fatal error would occur. So this general use-case is fine.

I think the only problem could be when you have a module in the Performance Lab active and _then_ activate the corresponding standalone plugin on the site. The activation of a plugin is a special case where it happens after initializing WordPress, so in that scenario the module would be loaded first. That is where your point indeed becomes relevant: If we _only_ have the constant check in the standalone plugin and an "older" version of PL is used (e.g. upcoming 2.1.0) that does not include the constant, it would lead to a fatal error. So for that reason you're right - we should keep the `function_exists()` check, at least for some time in the future maybe until adoption of those older PL versions has gone to an extremely low level (i.e. probably at least several months from now).

Summaryizing: In general we shouldn't need both checks, but because we first ship only the `function_exists` check, there is a chance for fatal errors if we remove them now.

Thanks for reading to me thinking through this :)

admin/load.php
```diff
@@ -156,7 +157,9 @@ function perflab_render_modules_page_field( $module_slug, $module_data, $module_
 				<input type="checkbox" id="<?php echo esc_attr( "{$base_id}_enabled" ); ?>" aria-describedby="<?php echo esc_attr( "{$base_id}_description" ); ?>" disabled>
 				<input type="hidden" name="<?php echo esc_attr( "{$base_name}[enabled]" ); ?>" value="<?php echo $enabled ? '1' : '0'; ?>">
 				<?php
-				if ( 'database/sqlite' === $module_slug && file_exists( WP_CONTENT_DIR . '/db.php' ) && ! defined( 'PERFLAB_SQLITE_DB_DROPIN_VERSION' ) ) {
+				if ( is_wp_error( $can_load_module ) ) {
+					echo $can_load_module->get_error_message();
```
felixarntz: We need to escape this.
```suggestion
					echo esc_html( $can_load_module->get_error_message() );
```

load.php
```diff
@@ -236,6 +236,15 @@ function perflab_can_load_module( $module ) {
 		return true;
 	}
 
+	$standalone_plugins_constants = perflab_get_standalone_plugins_constants();
+	if (
+		isset( $standalone_plugins_constants[ $module ] ) &&
+		defined( $standalone_plugins_constants[ $module ] ) &&
+		0 !== strpos( constant( $standalone_plugins_constants[ $module ] ), 'Performance Lab ' )
```
felixarntz: Nit-pick: You can simplify this by using `str_starts_with()`, which is polyfilled regardless of PHP version since WordPress 5.9.
```suggestion
		! str_starts_with( constant( $standalone_plugins_constants[ $module ] ), 'Performance Lab ' )
```

felixarntz (commented):
@mukeshpanchal27 This is pretty much good to go from my perspective, just two tiny things here. More importantly though, I am not yet marking this as approved since we first have to figure out what to do regarding the `WP_Error` usage (see https://github.com/WordPress/performance/pull/670#discussion_r1139367128).

load.php
mukeshpanchal27: Thanks, Felix. I made the same changes initially, but when I ran `npm run lint-php` it gave me the error `Fatal error: Uncaught TypeError: vsprintf(): Argument #2 ($values) must be of type array, string given in /app/vendor/squizlabs/php_codesniffer/src/Files` so i updated it with `strpos` but now it working.

load.php
mukeshpanchal27: Thanks @joemcgill and @felixarntz for the feedback. It look promising to me and I implemented it in https://github.com/WordPress/performance/pull/670/commits/0c5bc48c15ddf9703dfc5fb6880bedb9d577d0e2. It removes the `WP_Error` and introduce new function `perflab_is_module_standalone` as suggested.

Can we open a new issue and work on SQLite module stuff if it isn't a blocker for this issue?

load.php
```diff
@@ -248,6 +254,39 @@ function perflab_can_load_module( $module ) {
...
 
+/**
+ * Checks whether the given module can be loaded by a separate plugin.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $module Slug of the module.
+ * @return bool Whether the module can be loaded by a separate plugin or not.
+ */
+function perflab_is_module_standalone( $module ) {
```
felixarntz: While this is the name we discussed this for, maybe let's call it `perflab_is_standalone_plugin_loaded( $module )`? I think that name would be more clear than the current one here. Happy to discuss alternatives.

cc @joemcgill

load.php
```diff
@@ -236,6 +237,11 @@ function perflab_can_load_module( $module ) {
 		return true;
 	}
 
+	// Do not load the module if it can be loaded by a separate plugin.
+	if ( perflab_is_module_standalone( $module ) ) {
+		return false;
+	}
```
felixarntz: Based on my comment https://github.com/WordPress/performance/pull/670#discussion_r1139423419, can you please remove this from here to not intertwine it with the `perflab_can_load_module()` function? Instead, the new function should be called wherever `perflab_can_load_module()` is called (i.e. in `perflab_is_valid_module()`).

load.php
```diff
@@ -224,6 +224,7 @@ function perflab_render_generator() {
  * Checks whether the given module can be loaded in the current environment.
  *
  * @since 1.3.0
+ * @since n.e.x.t Adds an additional check for standalone plugins.
```
felixarntz: See my other comment to remove this at least for now.

felixarntz (commented):
@mukeshpanchal27 Almost LGTM, I am suggesting two small adjustments.

load.php
felixarntz: Thanks @mukeshpanchal27, a new follow up issue sounds like a great idea. We can use that issue to implement a better architecture for handling _why_ a module cannot be loaded and making this information available to the end user.

The approach you implemented looks good, I just left a review in https://github.com/WordPress/performance/pull/670#pullrequestreview-1348814786.

load.php
mukeshpanchal27: The function name suggestion look good to me.

load.php
```diff
@@ -248,6 +254,39 @@ function perflab_can_load_module( $module ) {
 	return (bool) $module();
 }
 
+/**
+ * Checks whether the given module can be loaded by a separate plugin.
```
```suggestion
 * Checks whether the given module has already been loaded by a separate plugin.
```

load.php
```diff
@@ -248,6 +254,39 @@ function perflab_can_load_module( $module ) {
...
+ *
+ * @since n.e.x.t
+ *
+ * @param string $module Slug of the module.
+ * @return bool Whether the module can be loaded by a separate plugin or not.
```
```suggestion
 * @return bool Whether the module has already been loaded by a separate plugin.
```

felixarntz (approved):
LGTM, thanks @mukeshpanchal27!

Just two suggestions to clarify doc comments, which I can commit.

felixarntz:
@joemcgill Can you give this another pass please?

@mukeshpanchal27 Could you open an issue to further explore https://github.com/WordPress/performance/pull/670#discussion_r1141636353? This can wait, but we should log it for the future.

load.php
```diff
@@ -248,6 +254,39 @@ function perflab_can_load_module( $module ) {
...
+ *
+ * @param string $module Slug of the module.
+ * @return bool Whether the module has already been loaded by a separate plugin.
+ */
+function perflab_is_standalone_plugin_loaded( $module ) {
+	$standalone_plugins_constants = perflab_get_standalone_plugins_constants();
+	if (
+		isset( $standalone_plugins_constants[ $module ] ) &&
+		defined( $standalone_plugins_constants[ $module ] ) &&
+		! str_starts_with( constant( $standalone_plugins_constants[ $module ] ), 'Performance Lab ' )
```
joemcgill, line 270: Using the content of the constant is fine for now, but it may be good to iterate on this approach so we are not dependent on modifying the constant value during the build process when a module is converted to a standalone plugin.

modules/images/webp-uploads/load.php
```diff
@@ -8,6 +8,13 @@
  * @package performance-lab
  */
 
+// Define the constant.
+if ( defined( 'WEBP_UPLOADS_VERSION' ) ) {
+	return;
+}
+
+define( 'WEBP_UPLOADS_VERSION', 'Performance Lab ' . PERFLAB_VERSION );
```
joemcgill, line 16: This will need to be modified as part of the build step, correct? Do we already have this covered in an issue somewhere?

modules/images/webp-uploads/load.php
joemcgill: Thanks for the explanation, we might consider running a quick set of manual checks by activating the standalone plugins with different versions of the PL plugin installed, but really, I don't know why someone would install and activate these plugins if they already have a version of the PL plugin where these are included as a module, so probably an edge case that we can document as part of the plugin repo readme.

joemcgill (approved):
This is looking ok to me now. I left one question about how we plan to handle changing the constant value, but that can be handled outside this PR.

modules/images/webp-uploads/load.php
felixarntz: It's part of this PR :)

https://github.com/WordPress/performance/pull/670/files#diff-69937e306961cbecc9a03ad8e84f06ee230caeb86a72a6f222f071df3a518bc3R88

load.php
felixarntz: What is the drawback of converting it as part of the build step?

joemcgill: The only real drawback is that it couples this execution logic with our build process, which means that changes to our build process could introduce bugs. If we could avoid this coupling, I think it would be nice. Not a blocker.

modules/images/webp-uploads/load.php
joemcgill: 🤦🏻‍♂️ of course it is. Thanks for pointing out.

load.php
mukeshpanchal27: Thanks @joemcgill, Do we need to open a follow-up issue for this? If yes, can you do it, or would you like me to open it?

joemcgill: I see that you already have opened an issue for this, which I think is the right next step. Thank you!

mukeshpanchal27: Thanks for confirming Joe.

Issue #638, opened 2023-02-01
Implement GitHub workflow to build and deploy standalone plugins


felixarntz:
## Feature Description

Follow up to #635, #636, and #637: Once the CLI command from that issue has been implemented, we should use this issue to implement a new GitHub workflow like `deploy-standalone-plugins.yml` which uses the CLI command to build the plugins and then deploys them individually to their respective WordPress.org plugin repository folder.

### Requirements

* A new GitHub workflow `.deploy-plugins.yml` should be implemented, for now to deploy modules as standalone plugins to the WordPress plugin repository.
    * It should use the `build-plugins` script from #635 to transform certain modules (based on the `plugins.json` file) to standalone plugins.
    * It should then check out the individual SVN repositories for those plugins from the `plugins.json` file, and for each plugin replace the current `trunk` folder content with the newly built plugin.
    * From there on, it should only proceed for each plugin _if_ the version in the plugin's main file and the plugin's `readme.txt` file matches and is greater than the version that was previously in those files. (The only exception is if this is the initial release, i.e. there was no version before.) If so, the next steps for each plugin are:
        * Copy the `trunk` folder to a new `tags/$version` folder.
        * SVN-add / SVN-delete any files as needed based on `svn status`.
        * Commit the new code.
    * A lot of the above logic can probably be based on the code from https://github.com/10up/action-wordpress-plugin-deploy - however we cannot use that GitHub action directly here since it is intended to work only for one plugin in one repository (while we have a monorepo here). Another potentially useful resource for some of the logic needed could be https://github.com/Automattic/jetpack/tree/trunk/.github.
    * Let's implement scripts (e.g. `.sh` or `.js`) as needed - it's probably not a good idea to bake all of this logic directly into a GitHub workflow YAML file, as that would get overly complicated.
    * For triggering this workflow, as a starting point we can probably set it to only run when a new Performance Lab plugin release is made, i.e. this would also trigger a new release of any standalone plugins if their versions were intentionally increased. We can always enhance this later if we e.g. also need to trigger manual releases for those plugins. Note that in the future, once the standalone plugins are no longer part of the PL plugin, the trigger will have to be changed anyway, so this is just a temporary measure.

joemcgill:
Suggestions for the requirements:

* This workflow should be part of a release process, not a merge, so it should only run when a new release tag is added
* Only modules with changes since the last release should be deployed
[...]

PR #671, opened 2023-03-14
Add / #639 Standalone Built Plugin Tests

### LINKED ISSUE #639: Implement GitHub workflow to have a basic test for standalone plugins
## Feature Description

Follow up to #635: As we are publishing certain modules as standalone plugins, it is vital that we ensure the plugins work as expected. This issue's goal is to provide at least a foundation for some automated testing of the individual plugins: It should run the build step from #635 and then load the plugin in e.g. a customized PHPUnit configuration, to just ensure that there is no fatal error when loading the plugin. While this is far from a comprehensive solution, the code itself is already covered by the unit tests on the module, and having this extra check is the most critical and minimum starting point to ensure we don't accidentally use anything from the Performance Lab core in a module that wouldn't work when it's standalone.

This should already be the case per our [writing a module guidelines](), but having a CI workflow to ensure that will be crucial to ensure we don't let any mistake accidentally slip through.

### Requirements

* Implement a new GitHub workflow that runs the `build-plugins` command from #635 and then loads each individual built plugin in a sandboxed PHP environment.
    * For example, it could load the WordPress unit test suite and force those plugins to be active, just to ensure no fatal error occurs.
* This workflow should probably run only on `push` events to `trunk`, not for every pull request, to avoid the overhead of having this on every PR. Running this on every `trunk` push is sufficient to capture failures before they would accidentally get released to end users.
    * Additionally, only modules that have seen changes since the last `push` should be covered by this test, since for unchanged modules it would be redundant and just slow down CI unnecessarily.

10upsimon:
## Summary

Fixes #639 

### Introduction of a new `test-plugins` cli command
This CLI command is an extension of the existing `./bin/plugin/cli.js` entry point and can be execute in one of two manners;

1. Via `npm run test-plugins` for tests in a single site environment
2. Via `npm run test-plugins-multisite` for tests in a multisite environment
3. Via the more _vanilla_ form `./bin/plugin/cli.js test-plugins` for single site site testing
4. 3. Via the more _vanilla_ form `./bin/plugin/cli.js test-plugins --sitetype=multi` for multisite testing

The former 2 is the preferred means, and the means used by the Github workflow that executes standalone plugin integration testing.

### Why a `test-plugins` cli command as opposed to just Github Workflow set of steps?
Authors may want to validate standalone plugins locally following changes to one or more of the modules. A cli command off of the existing `./bin/plugin/cli.js` seemed like a logical and accessible choice and flows naturally from the recently introduced `build-plugins` command.

### `test-plugins` logic overview
The logical flow undertaken by the script is as follows:

1. A check for the presence of a `.wp-env.json` file is performed as this is critical to the logic that follows. If one is not present, a suitable error is produced and an exit with code `1` is performed to ensure that the test fails.
2. The presence of the `build` directory is checked. If no `build` directory is found, a suitable error is produced and an exit with code `1` is performed to ensure that the test pipeline fails. Users should run `npm run build-plugins` as a prerequisite to the standalone plugin integration tests. This step is however performed as part of the Github workflow introduced in this PR.
5. Each plugin in the `build` dir is then looped over and the steps that follow are performed.
6. Test assets from the `plugin-tests` folder are copied to each standalone plugin directory. These include a suitable `composer.json` file to install necessary testing dependencies, a `tests` folder containing a single test suite used for all standalone plugins and `phpunit.xml.dist` used for all standalone plugins, that defines the test(s) to execute per plugin. _All plugins perform the same tests, so updating the tests within the `plugin-tests/tests` folder will have bearing on all built plugins_.
7. `composer install --no-interaction` is performed against each plugin directory within the `build` directory. This is handled by setting the `--working-dir` argument of the command to reference the suitable plugin within the loop.
8. The `.wp-env.json` file within `plugin-test` is update and its `plugins` property update to an array reflective of the plugins inside of the `build` directory. It is then copied to the root directory. This ensures that when we start the wp-env environment, all built plugins are already active. This array of plugins within the `.wp-env.json` file _does not_ include the top level `performance` plugin.
9. The `wp-env` environment is started. This command is executed synchronously.
10. `npm run pretest-php` is executed in order for the applicable packages at a top level to be installed.
11. Plugins are again iterated and test are executed against each via the command `wp-env run phpunit 'phpunit -c /var/www/html/wp-content/plugins/${ plugin }/phpunit.xml.dist --verbose --testdox'` where `${plugin}` is the applicable plugin in the loop. This command is executed synchronously.
12. If a test fails, i.e the above command returned exit status `1`, the process is halted and an overall exit with code `1` is returned, forcing a fail on the test pipeline.
13. This process continues for all plugins, until an error/failure (if any) is encountered.
14. If no errors/failures are encountered, the `wp-env` environment is stopped and an exit with code `0` is performed. This signals a successful execution of the test cycle. At this stage, the integration tests for all plugins would have passed `OK`. 

### Why does the `test-plugins` command handle `wp-env start`?
Because we need to first ensure that the`.wp-env.json` file contains valid plugin pointers to the plugins within the `build` directory or else there would be no way to activate them as the environment would be unaware of their existence. Further to this, at time of authoring this code it is not possible to run the standalone plugins alongside the performance plugin.

### What if there are I/O operation errors?
All I/O operations are executed synchronously and with suitable exception catching. Exits with code `1` are performed, therefore producing a suitable error message, and failing the test pipeline.

### Can I test plugin manually outside of the Github Workflow?
Yes, you're encouraged to. As long as you've run `npm run build-plugins` first, running `npm run test-plugins` should result in a successful execution of the integration test suite for all standalone plugins in `build`. HOWEVER, be aware that this command does change the `plugins` property of the `.wp-env.json` file to reference the built plugins, and removes the reference to the top level `performance` plugin.

For example, your `.wp-env.json` files `plugins` property may change from something like this:

```
"plugins": [ "." ],
```

To this:

```
"plugins": [ "./build/webp-uploads" ],
```

Simply reinstate it to `[ "." ]` and re-run `wp-env start` in order to resume a WordPress environment with just the `performance` plugin active.

### Introduction of a new `Standalone Plugin Integration Testing / PHP` Github workflow
*NOTE:* This workflow currently runs against `push` and `pull` actions agains `feature`, `release` and `trunk` branches. Once the process of the checks are confirmed via this PR, the workflow will be suitably updated to be more concise and specific.

This workflow executes the following operations:

1. Ensures the presence of a suitable Ubuntu based PHP container.
2. Sets up Node.js
3. Caches npm
4. Runs `npm ci` agains the root `package.json` file
5. Executes the `npm run build-plugins` command to generate standalone plugins, as these are not part of the repository source
6. Executes `npm run test-plugins` following a successful step above, and continues to run integration tests for all built plugins as described further above, in a single site environment.
7. 6. Executes `npm run test-plugins-multisite` following a successful step above, and continues to run integration tests for all built plugins as described further above, in a multisite environment.

## Checklist


bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
...
+		);
+
+		// Copy over test files.
+		try {
+			fs.copySync( pluginTestAssets, `./build/${ plugin }/`, {
+				overwrite: true,
+			} );
+			log(
+				formats.success(
+					`Copied test assets for plugin "${ plugin }, executing "composer install --no-interaction" on plugin.\n`
```
mukeshpanchal27:
```suggestion
					`Copied test assets for plugin "${ plugin }", executing "composer install --no-interaction" on plugin.\n`
```

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
...
+		formats.success(
+			`Starting wp-env environment with active plugins: ${ builtPlugins.join(
+				', '
+			) }`
+		)
+	);
+
+	// Execute wp-env start.
+	execSync( `npm run wp-env start`, ( err, output ) => {
+		// once the command has completed, the callback function is called
```
mukeshpanchal27:
```suggestion
		// once the command has completed, the callback function is called.
```

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
...
+		if ( err ) {
+			log( formats.error( `${ err }` ) );
+
+			// Return with exit code 1 to trigger a failure in the test pipeline.
+			process.exit( 1 );
+		}
+		// log the output received from the command
```
mukeshpanchal27:
```suggestion
		// log the output received from the command.
```

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
...
+	// Run composer install on the main coposer container for the project.
+	execSync( `npm run pretest-php`, ( err, output ) => {
+		// once the command has completed, the callback function is called
+		if ( err ) {
+			log( formats.error( `${ err }` ) );
+
+			// Return with exit code 1 to trigger a failure in the test pipeline.
+			process.exit( 1 );
+		}
+		// log the output received from the command
```
mukeshpanchal27:
```suggestion
		// log the output received from the command.
```

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
...
+			// Return with exit code 1 to trigger a failure in the test pipeline.
+			process.exit( 1 );
+		}
+
+		// Run phpunit tests for specific plugin.
+		try {
+			execSync(
+				`wp-env run phpunit 'phpunit -c /var/www/html/wp-content/plugins/${ plugin }/phpunit.xml.dist --verbose --testdox'`,
+				( err, output ) => {
+					// once the command has completed, the callback function is called
```
mukeshpanchal27:
```suggestion
					// once the command has completed, the callback function is called.
```

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
...
+						log(
+							formats.error(
+								`Error executing phpunit test command: ${ err }`
+							)
+						);
+
+						// Return with exit code 1 to trigger a failure in the test pipeline.
+						process.exit( 1 );
+					}
+					// log the output received from the command
```
mukeshpanchal27:
```suggestion
					// log the output received from the command.
```

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
...
+			`All standalone plugin tests appeared to have passed :)`
+		)
+	);
+
+	// Start winding down.
+	log( `Stopping wp-env...` );
+
+	// Stop wp-env.
+	execSync( `wp-env stop`, ( err, output ) => {
+		// once the command has completed, the callback function is called
```
mukeshpanchal27:
```suggestion
		// once the command has completed, the callback function is called.
```

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
...
+		if ( err ) {
+			log( formats.error( `${ err }` ) );
+			return;
+		}
+		// log the output received from the command
```
mukeshpanchal27:
```suggestion
		// log the output received from the command.
```

plugin-tests/phpunit.xml.dist
```diff
@@ -0,0 +1,14 @@
...
+		convertErrorsToExceptions="true"
+		convertNoticesToExceptions="true"
+		convertWarningsToExceptions="true"
+>
+	<testsuites>
+		<testsuite name="default">
+			<directory suffix=".php">./tests</directory>
+		</testsuite>
+	</testsuites>
+</phpunit>
```
mukeshpanchal27: Missing end line.

plugin-tests/tests/test-standalone.php
```diff
@@ -0,0 +1,16 @@
+<?php
+/**
+ * Class PluginTest
+ *
+ * @package Webp_Uploads
```
mukeshpanchal27:
```suggestion
 * @package performance-lab
```

WDYT?

.gitattributes
```diff
@@ -30,3 +30,5 @@
 
 /.gitattributes export-ignore
 /.gitignore export-ignore
+
+/plugin-tests export-ignore
```
mukeshpanchal27: Missing end line.

.github/workflows/php-test-standalone-plugins.yml
```diff
@@ -0,0 +1,58 @@
+name: Standalone Plugin Integration Testing
+
+on:
```
mukeshpanchal27, line 3: Does this workflow only work when the `plugins.json` file is updated?

mukeshpanchal27 (commented):
Thanks @10upsimon, Great start. Left some nit-pick feedback.

.github/workflows/php-test-standalone-plugins.yml
10upsimon: @mukeshpanchal27 for now, it runs at the same time php unit tests run, but this will change once I have approval on the process itself. Thank you!

joemcgill:
Thanks @10upsimon for the thorough writeup of the approach you've chosen to take here. There's a lot to dig into, so before getting too deep into details, I'd like to share some initial observations and questions.

1. Overall, it seems like we're adding a lot of complexity in the setup to account for what is a limited problem—testing the build of individual plugins while we're in the in-between state of having these modules available as both part of the Performance Lab plugin AND as individual plugins. Currently, the unit tests don't appear to be running full test suites, just ensuring that there are no fatals, but we're taking a long path to get to that result. I'm curious if we could simplify this quite a bit if we chose to focus on simply ensuring that no fatals are produced when we load a site with all plugins (including the performance lab plugin) activated at the same time?
2. It looks like this approach is spinning up separate wp-env environments for each built plugin. Is that correct? If so, that seems fairly heavy handed to me. Assuming we eventually need to run separate unit test suites for all built plugins, could we modify the approach so that all of the PHPUnit tests are run in a single container that loads the plugin code in a single bootstrap process, rather than as individual independent tests?
3. If we could run tests on all of these plugins from the same container, would there still be a need to copy a composer.json to each plugin folder?

10upsimon:
@joemcgill  many thanks for your early review feedback, to address your points:

> Overall, it seems like we're adding a lot of complexity in the setup to account for what is a limited problem—testing the build of individual plugins while we're in the in-between state of having these modules available as both part of the Performance Lab plugin AND as individual plugins. Currently, the unit tests don't appear to be running full test suites, just ensuring that there are no fatals, but we're taking a long path to get to that result. I'm curious if we could simplify this quite a bit if we chose to focus on simply ensuring that no fatals are produced when we load a site with all plugins (including the performance lab plugin) activated at the same time?

I believe we should be running tests in 2x scenarios:

1. Testing standalone plugins _without_ the PL plugin active, this is what we are accounting for in this iteration of work
2. Testing standalone plugins _with_ the PL plugin active, which was not possible prior, but is now that #674 

To be clear,  `simply ensuring that no fatals are produced when we load a site with all plugins` is exactly what we are doing here. I believe the perceived complexity stems from the fact that because we are not enabling the PL plugin, there are some prerequisites needing to be performed to successfully run tests, just like we do for the PL plugin. In summary, the actual steps taken to arrive at the tests is quite minimal and only involves copying a small handful of assets and executing the tests.

> It looks like this approach is spinning up separate wp-env environments for each built plugin. Is that correct? If so, that seems fairly heavy handed to me. Assuming we eventually need to run separate unit test suites for all built plugins, could we modify the approach so that all of the PHPUnit tests are run in a single container that loads the plugin code in a single bootstrap process, rather than as individual independent tests?

This is not the case. This approach builds a plugin string of _all_ plugins that reside in the `build` directory, applies that to the existing `.wp-env.json` file, and starts the (single) WordPress environment that references `n` plugins from the `build` directory as active plugins.

Regarding running separate unit tests for all plugins, this is currently handled by the PL plugin which runs tests against each module. Will it not always be the case that we'll maintain the PL plugin and handle standalone plugins as builds therefrom, i.e via `npm run build-plugins`? If so, we should leave the module based tests as they are, as there is no need to cover those again in standalone plugin tests. If we need to adjust for that in future, we can. If we are _not_ going to be building standalone plugins from this "monorepo" using `npm run build-plugins` indefinitely, then I'd argue that the tests can be adapted for the standalone plugin repo's when that time comes.

> If we could run tests on all of these plugins from the same container, would there still be a need to copy a composer.json to each plugin folder?

I could not get it to work correctly _without_ having the PL plugin active as well (which already sets up the prerequisites for tests) and I think forcing the presence of the PL plugin is the incorrect path to take. That being said, I could attempt to simply run `composer install --no-interaction` on the top level composer file and that might actually solve for what you're suggesting, I'll give that a whirl today.

Thanks for your thoughtful feedback @joemcgill , much appreciated!

joemcgill:
Thanks for the update, @10upsimon, I had definitely misunderstood part of how this was working, in terms of spinning up separate environments for each plugin. I've done some local testing and things are much more clear. I'll consolidate any additional feedback in a proper code review soon.

.gitattributes
felixarntz: Nit-pick, maybe put this exclude together with the other directories that contain tooling code (lines 8-10 above).

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
...
+	} catch ( e ) {
+		log(
+			formats.error(
+				`Error replacing content in ${ wpEnvFile } using regex "${ wpEnvPluginsRegex }": "${ e }"`
+			)
+		);
+
+		// Return with exit code 1 to trigger a failure in the test pipeline.
+		process.exit( 1 );
+	}
```
felixarntz: Maybe move this code into its own function? The handler function is extremely long, and I think especially this here is pretty much one logical step that could be broken apart into its own function unit.

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
...
+
+			// Return with exit code 1 to trigger a failure in the test pipeline.
+			process.exit( 1 );
+		}
+		// log the output received from the command.
+		log( output );
+	} );
+
+	// Run tests per plugin.
+	log( formats.success( `About to execute standalone plugins tests.` ) );
```
felixarntz: This output is probably unnecessary :)

felixarntz (commented):
@10upsimon This looks great to me! Three nit-picks below, I only have one broader point of feedback for consideration: I think the approach of reading `.wp-env.json` and replacing the `plugins` entry in it is a creating solution, however it may be simpler to create a static `.wp-env.json` for this purpose in another location (e.g. in `plugin-tests/.wp-env.json`) with the plugins already in it, since the list of plugins we already know from the beginning (see the `plugins.json` file), so I don't think we technically have to "detect" them from the `build` directory. Your approach here would allow to e.g. only manually build one plugin and then the test code would only test that one - which may be useful, but at the same time I wonder whether that'll ever be used, given that we would always want to have standalone plugins available for all plugins in the `plugins.json` configuration.

I'm curious what you think about it and what your rationale for the approach was. Definitely not saying that we have to change it, I'm just pointing out an alternative solution that we can consider.

bin/plugin/cli.js
```diff
@@ -80,6 +84,13 @@ withOptions( program.command( 'build-standalone-plugins' ), buildPluginsOptions
 	)
 	.action( catchException( buildPluginsHandler ) );
 
+withOptions( program.command( 'test-standalone-plugins' ), testPluginsOptions )
```
mukeshpanchal27, line 88: @10upsimon Can you run `npm run lint-js` and fix the error.

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
+/**
+ * External dependencies
+ */
+const fs = require( 'fs-extra' );
+const { execSync, spawnSync } = require( 'node:child_process' );
```
mukeshpanchal27:
```suggestion
const { execSync, spawnSync } = require( 'child_process' );
```
When i run command in local environment it show me `Error: Cannot find module 'node:child_process'` error. After applying the changes it working for me.

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
...
+			formats.error(
+				`WP Env config file "${ wpEnvFile }" not detected in root of project.`
+			)
+		);
+
+		// Return with exit code 1 to trigger a failure in the test pipeline.
+		process.exit( 1 );
+	}
+
+	// Only try to read plugin dirs if build ir exists.
```
mukeshpanchal27:
```suggestion
	// Only try to read plugin dirs if build is exists.
```

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
...
+	builtPlugins.forEach( ( plugin ) => {
+		log(
+			formats.success(
+				`Detected built plugin "${ plugin }", copying test files for standalone/integration testing.`
+			)
+		);
+
+		// Copy over test files.
+		try {
+			fs.copySync( pluginTestAssets, `./build/${ plugin }/`, {
```
mukeshpanchal27:
```suggestion
			fs.copySync( pluginTestAssets, `${ builtPluginsDir }${ plugin }/`, {
```

Use already define variable.

mukeshpanchal27 (commented):
Thanks @10upsimon, I left some findings and feedback.

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,323 @@
...
+				wpEnvPluginsRegex,
+				wpEnvPluginsRegexReplacement
+			)
+		);
+	} catch ( e ) {
+		log(
+			formats.error(
+				`Error replacing content in ${ wpEnvFile } using regex "${ wpEnvPluginsRegex }": "${ e }"`
+			)
+		);
```
joemcgill: Rather than overwriting the `wp-env.json` file, we can create a new `wp-env.override.json` file, which we could also optionally add to the `.gitignore` so anyone running this locally won't accidentally have changes in their working branch.

See: https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env/#wp-env-override-json

plugin-tests/tests/bootstrap.php
```diff
@@ -0,0 +1,42 @@
...
+} elseif ( file_exists( TESTS_PLUGIN_DIR . '/../../../../tests/phpunit/includes/functions.php' ) ) {
+	$_test_root = TESTS_PLUGIN_DIR . '/../../../../tests/phpunit';
+} else { // Fallback.
+	$_test_root = '/tmp/wordpress-tests-lib';
+}
+
+// Force plugin to be active.
+$GLOBALS['wp_tests_options'] = array(
+	'active_plugins' => array( basename( TESTS_PLUGIN_DIR ) . '/load.php' ),
+);
```
joemcgill: Do we also need to  optionally load the main Plugin Lab plugin as well, in order to ensure there are no fatals due to both being loaded at the same time, or is that out of scope for this specific issue?

plugin-tests/composer.json
```diff
@@ -0,0 +1,34 @@
...
+    "performance",
+    "wordpress"
+  ],
+  "support": {
+    "issues": "https://github.com/WordPress/performance/issues"
+  },
+  "require-dev": {
+    "dealerdirect/phpcodesniffer-composer-installer": "^0.7",
+    "phpcompatibility/php-compatibility": "^9.3",
+    "phpunit/phpunit": "^4|^5|^6|^7|^8|^9",
```
joemcgill: Fun fact: PHPUnit doesn't need to be required if we're using `yoast/phpunit-polyfills`, because it [requires PHPUnit as a dependency itself](https://github.com/Yoast/PHPUnit-Polyfills/blob/1.0.4/composer.json#L27) and handles proper versioning for you.

joemcgill (commented):
A couple small suggestions, but after reading all this again, I think the approach makes sense and is looking good. One other thing I'd like to confirm is whether running these tests during a deployment workflow (see: #638) has the potential of polluting the build artifacts that would be uploaded to the SVN repo, given that we're copying certain files into the plugin folder in order to run these tests.

bin/plugin/cli.js
10upsimon: @mukeshpanchal27 I got no changes at all when running `npm run lint-js` and my eslint is also not showing errors. Can you please confirm this again? What is the error you are seeing?

plugin-tests/tests/bootstrap.php
10upsimon: @joemcgill I figured this was out of the scope of this task/issue, but given that we now have a working ability to run them side by side, maybe we should add this test case?

plugin-tests/composer.json
10upsimon: @joemcgill have removed, thanks. We should probably consider this rule for the main plugin, as that is where I borrowed the file from.

server-timing/.wp-env.json
```diff
@@ -0,0 +1,12 @@
+{
```
mukeshpanchal27: Do we need `.wp-env.json` in server-timing folder?

plugin-tests/multisite.xml
```diff
@@ -0,0 +1,18 @@
+<phpunit
+		bootstrap="tests/bootstrap.php"
```
mukeshpanchal27: Correct indention as per https://github.com/WordPress/performance/blob/trunk/.editorconfig

plugin-tests/phpunit.xml
```diff
@@ -0,0 +1,15 @@
+<phpunit
+		bootstrap="tests/bootstrap.php"
```
mukeshpanchal27: Check similar comment above.

mukeshpanchal27 (commented):
Thanks @10upsimon for the changes. Left some quick feedback and one question.

bin/plugin/cli.js
joemcgill: I can confirm that I also am not getting errors, so this may be a configuration problem.

plugin-tests/tests/bootstrap.php
joemcgill: Yes I think that it's a good idea that we have the ability to do both, but let's handle that in a separate issue so it doesn't block progress here.

plugin-tests/composer.json
joemcgill: Agreed! It's some thing that we could check for in all of our plug-ins that use the polyfills library

plugin-tests/tests/bootstrap.php
```diff
@@ -0,0 +1,37 @@
...
+if ( false !== getenv( 'WP_PLUGIN_DIR' ) ) {
+	define( 'WP_PLUGIN_DIR', getenv( 'WP_PLUGIN_DIR' ) );
+} else {
+	define( 'WP_PLUGIN_DIR', dirname( TESTS_PLUGIN_DIR ) );
+}
+
+// Load Composer dependencies if applicable.
+if ( file_exists( TESTS_PLUGIN_DIR . '/vendor/autoload.php' ) ) {
+	require_once TESTS_PLUGIN_DIR . '/vendor/autoload.php';
+}
```
joemcgill, line 20: This is likely the reason that you have needed to call composer install in each plug-in directory. If this were looking for the auto load files somewhere else, everything probably would work. Not a blocker though, for my point of you.

10upsimon:
After discussions with @joemcgill and @felixarntz it was discussed that:

- Logic would be changed so that the plugin list is read from the `plugins.json` file as opposed to traversing the `build` directory, as this is theoretically the source of truth. Directory checks were still added, and if said plugin dir does not exist within `build`, it is removed from the plugins array prior to starting `wp-env` and running the unit tests.
- A version of the standalone tests _alongside_ the main WPP plugin would be run as well, to ensure that the standalone plugins do not fatal when the WPP plugin is present and active as well. This is done as a round 2 of testing, with the same single or multisite tests being performed, simply _with_ the WPP plugin enabled.
- A totally separate workflow would be created to deploy the standalone plugins to the applicable SVN repo's due to the fact that the plugin integration tests do pollute the built plugin folders due to test assets needing to be added.

plugin-tests/tests/bootstrap.php
10upsimon: @joemcgill  as discussed between us, I'm having a hard time getting it to work without plugin specific installs. For now, I am going to proceed with plugin specific composer installs, as it's working. If others want to open up a PR for this after the fact, this is most welcomed.

server-timing/.wp-env.json
10upsimon: @mukeshpanchal27 this was incorrectly copied by myself, I have since removed it.

bin/plugin/commands/test-plugins.js
10upsimon: Thanks @joemcgill , taking a look at this as it seems a necessary change to me  👍

plugin-tests/tests/bootstrap.php
10upsimon: @joemcgill this has been handled in a recent commit.

bin/plugin/commands/test-plugins.js
10upsimon: @joemcgill this has been implemented in a recent commit.

10upsimon:
@joemcgill @felixarntz @mukeshpanchal27 this is ready for a final review I believe. I've requested a re-review from all 3 of you. Notable changes:

- I've split the logic into smaller functions and lightened the handler substantially. I decided specifically on splitting the logic of replacing/building the `wp-env.override.json` file and running the unit tests to ease the process of redoing the full test cycle _with_ the WPP plugin active as well.
- Carrying on from the above, I've now added a second round of testing _with_ the WPP plugin active as well. This is done within the same cli cycle, there is no separate workflow for it.
- Multisite tests have been added via a new `--sitetype` flag, it defaults to `'single'` but can accept `'multi'`. I've made this more accessible via a new npm command `npm run test-plugins-multisite`
- Logic has been altered to obtain the built plugins list from the `plugins.json` file as opposed to directory traversing. I've still added a check against each plugin to ensure that the directory exists though. If not, it is removed (via `.filter()`) from the array of plugins and a suitable warning is issued in the cli output. NOTE: I have _NOT_ failed the test pipeline in this case.
- Various minor changes relating to prior review comments

The only outstanding item is a full and proper implementation of the GitHub workflow rules for the changed files. If we're happy with the rest of the code, I can work on that as a final deliverable.

Thank you.

.github/workflows/php-test-standalone-plugins.yml
```diff
@@ -0,0 +1,60 @@
...
+      - 'composer.json'
+      - 'composer.lock'
+    types:
+      - opened
+      - reopened
+      - synchronize
+
+jobs:
+  php-test:
+    name: PHP
```
mukeshpanchal27: This should be updated with proper naming.

10upsimon: @mukeshpanchal27 resolved in a recent commit, thank you.

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,440 @@
...
+
+		// Return with exit code 1 to trigger a failure in the test pipeline.
+		process.exit( 1 );
+	}
+
+	// Create an array of plugins from entries in plugins JSON file.
+	builtPlugins = Object.keys( pluginsJsonFileContentAsJson )
+		.filter( ( item ) => {
+			if ( ! fs.pathExistsSync( `${ settings.builtPluginsDir }${ pluginsJsonFileContentAsJson[ item ].slug }` ) ) {
```
joemcgill, line 420: Do we need to add any handling here in case the `builtPluginsDir` value does not end in a slash?

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,440 @@
...
+			formats.error(
+				`Error replacing content in ${ settings.wpEnvFile } using regex "${ wpEnvPluginsRegex }": "${ e }"`
+			)
+		);
+
+		// Return with exit code 1 to trigger a failure in the test pipeline.
+		process.exit( 1 );
+	}
+
+	// Copy the newly modified ./plugins-tests/.wp-env.json file to the root level.
```
joemcgill: Why not write directly to this file rather than writing to the original and then copying?

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,440 @@
...
+
+	// Run unit tests against built plugins.
+	doRunUnitTests( { ...settings, builtPlugins } );
+
+	log(
+		formats.success(
+			`Re-running standalone plugin integration tests with WPP plugin active.`
+		)
+	);
+	// Add the root level WPP plugin to the built plugins array.
```
joemcgill, line 475: This works, but we could probably speed this up a bit by doing the first run with the WPP plugin active and then using a WP CLI command to disable it for the second run. That way we don't have to worry about restarting the environment.

joemcgill (commented):
Thanks Simon! Looks like it's working now. Left a couple additional comments, but nothing blocking.

bin/plugin/commands/test-plugins.js
10upsimon: @joemcgill  I do not think this is absolutely necessary, as this is not a dynamically defined value, we hard define it as part of the settings object we pass to `doRunStandalonePluginTests()`. We can potentially handle this in a future PR if the need arises?

bin/plugin/commands/test-plugins.js
10upsimon: @joemcgill we have to copy it regardless the first time, but I agree that a check for file existing and then a copy over and final write is probably best, as if it's existing already, we may as well write to it. To be clear, it won't exist already as we need to create this overrides file, so I am not actually sure how much of a difference this makes, but I am going to effect this in case this os ever enhanced to handle multiple writes (which it is currently but you hav further feedback in this review to change that).

bin/plugin/commands/test-plugins.js
10upsimon: Great suggestion, I am in the process of changing this.

10upsimon:
@joemcgill @mukeshpanchal27 I've pushed a round of changes, all of which can be seen in the commit at https://github.com/WordPress/performance/pull/671/commits/f485ccf3fe7c4086e523b11682d90bae9282ec24 

I've refactored a few areas, but high level takeaways are:

- I now only copy the wp-env overrides file once, and I copy it and then write to it as per your suggestion @joemcgill 
- Added a variable to act as a state tracking agains whether we've started wp-env already, so that it is not unnecessary re-started
- Moved the wp-env stop command out of the testing function, as that is called multiple times. wp-env is now simply stopped as the final action of the handler logic
- Added two new settings objects for passing into `doRunUnitTests()` function: `disablePlugins` and `enablePlugins`. Instead of stopping and starting the env with different plugins in the plugins JSON object, we now start them with all plugins including the main WPP one, and we disable and enable it as needed via these new settings objects
- Use of wp-cli commands to deactivate and activate plugins (in our case only the main WPP one) to speed up the testing process. This has greatly sped up the total time to test

bin/plugin/commands/test-plugins.js
10upsimon: @joemcgill this has been refactored in the latest commit at the time of writing, please see my general update comment further down.

joemcgill (approved):
Nice updates @10upsimon , this is looking good, and I can confirm that everything is working as expected locally. The only thing that I noticed while testing this command locally that does not exceed affect our GitHub workflow is that the `performancePluginSlug` value may be incorrect if somebody has cloned this repo to a folder named something other than "performance". Unless that's an easy fix here, I think we could dress that has an enhancement later.

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,503 @@
...
+exports.handler = async ( opt ) => {
+	doRunStandalonePluginTests( {
+		pluginsJsonFile: './plugins.json', // Path to plugins.json file.
+		siteType: opt.sitetype || 'single', // Site type.
+		pluginTestAssets: './plugin-tests', // plugin test assets.
+		builtPluginsDir: './build/', // Built plugins directory.
+		wpEnvFile: './plugin-tests/.wp-env.json', // Base .wp-env.json file for testing plugins.
+		wpEnvDestinationFile: './.wp-env.override.json', // Destination .wp-env.override.json file at root level.
+		wpEnvPluginsRegexPattern: '"plugins": \\[(.*)\\],', // Regex to match plugins string in .wp-env.json.
+		performancePluginSlug: 'performance', // The plugin slug of the main WPP plugin.
```
felixarntz: This must not be hard-coded, it needs to be set based on whichever folder the plugin is in. See a related issue #673 that had a similar problem elsewhere. We should not assume the plugin is in a `performance` folder.

bin/plugin/commands/test-plugins.js
```diff
@@ -0,0 +1,503 @@
...
+		);
+
+		// Return with exit code 1 to trigger a failure in the test pipeline.
+		process.exit( 1 );
+	}
+
+	// Copy the newly modified wp-env overrides file to the root level if it does not exist.
+	if ( ! fs.pathExistsSync( settings.wpEnvDestinationFile ) ) {
+		try {
+			fs.copySync( settings.wpEnvFile, settings.wpEnvDestinationFile, {
```
felixarntz, line 133: At the very end of the entire process, we need to make sure we delete the `wpEnvDestinationFile` again. While it's in `.gitignore` which is a good idea, we need to still delete it since otherwise we'll "pollute" the development directory in a way that using `wp-env` will no longer result in the expected Performance Lab plugin environment but rather the one from the overridden configuration.

felixarntz (changes requested):
@10upsimon Code-wise this looks pretty much good to go, however I found two things we should address, both related to running locally. We need to also make sure this works smoothly locally, not only in the CI workflow.

Nit-pick, one thing in your PR description is probably out of date now? See this:
> HOWEVER, be aware that this command does change the `plugins` property of the `.wp-env.json` file to reference the built plugins, and removes the reference to the top level `performance` plugin.

I believe this is no longer accurate, maybe from a previous iteration of the PR? We can probably remove it now that the code here uses the .wp-env.override.json` file.

10upsimon:
@joemcgill @felixarntz thanks for the review, I've addressed final pieces of feedback in https://github.com/WordPress/performance/pull/671/commits/e36d516489a7120ff4926185835cde9527693eeb

- `performancePluginSlug` is now dynamic and based off of the base directory in which the `npm run test-plugins` command is run from, which is the root level folder in which the repo was cloned to.
- As part of cleanup operations, we are now removing the `.wp-env.override.json` file from the base/root directory.

bin/plugin/commands/test-plugins.js
```diff
@@ -498,6 +502,15 @@ function doRunStandalonePluginTests( settings ) {
 		log( output );
 	} );
 
+	// Remove the ep-env overrides file.
```
joemcgill: Typo: `ep-env`

Also I wonder if you should run this before running any of the tests so that this is cleaned up as soon as the environment is started?

bin/plugin/commands/test-plugins.js
10upsimon: @felixarntz addressed in a recent commit, thank you.

bin/plugin/commands/test-plugins.js
10upsimon: @felixarntz addressed in a recent commit, thank you.

bin/plugin/commands/test-plugins.js
10upsimon: @joemcgill this is addressed in a recent commit. Thank you!

joemcgill (approved):
Approving again. This is working great locally now too. 🎉

felixarntz (approved):
Thank you @10upsimon for the last updates, great work!

PR #672, opened 2023-03-14
Fix problems with placing `object-cache.php` drop-in

### LINKED ISSUE #630: Revisit how we handle object-cache.php conflicts
## Bug Description

Based on [one users testing](https://wordpress.org/support/topic/plugin-overwrites-existing-third-party-wp-content-object-cache-php/#post-16403840) with the plugin when also using Redis Object Caching pro, Redis Cache pro is disabled when it determines an object-cache.php file present that isn't generated from Redis Object Caching pro. All the features of Redis Cache Pro are disabled when it comes across a non Redis Cache Pro object-cache file. 

While it's possible any resolution on this may be applicable on the Redis Object Cache pro side, it may be an opportunity to check how the plugin handles such conflicts. At present we're moving the contents of a third party plugins `object-cach.php` file to `object-cache-plst-orig.php`, and then requesting that within the footer of the PL placed `object-cache.php` file. Maybe there's another way of using a third party caching plugin alongside the PL object-cache.php file. 

Note also that we did make changes recently (PR #616) to resolve issues with third party plugin caching file contents being added to both the `object-cache.php` and `object-cache-plst-orig.php` files. 

There is also an [existing PL GitHub issue](https://github.com/WordPress/performance/issues/628) specific to Redis Cache Pro which, once actioned will provide a solution.  

## Steps to reproduce

1. Install the PL plugin on a site that already have Redis Object Cache Pro active
2. Check the `object-cache.php` file contents to ensure the PL plugin has overwritten the existing `obect-cache` file (while moving the existing one to `object-cache-plst-orig.php`
3. Check your WP dashboard and notice the [invalid status](https://i.snipboard.io/fjVKcY.jpg) of the drop-in file

## Screenshots

image

felixarntz:
## Summary

Fixes #630

## Relevant technical choices

* So far, if there was an existing `object-cache.php` drop-in, the plugin would attempt to rename it to `object-cache-plst-orig.php`, then place its own `object-cache.php` file which would then also `require_once` the renamed original drop-in. This worked correctly in principle, however several popular plugins regularly check the file contents of the `object-cache.php` file to check whether their drop-in is present rather than checking for e.g. a constant that the file sets. Due to that behavior, in practice the approach the Performance Lab plugin has taken led to several incompatibilities. This PR fixes that problem going forward by removing that approach. Now the Performance Lab drop-in will only be placed if no conflicting drop-in is present.
* Some code around the `object-cache-plst-orig.php` file is maintained for backward compatibility: For example, for sites for which that approach has been working as expected, it will need to be ensured that the deactivation routine not only deletes the Performance Lab `object-cache.php` drop-in, but also restores the original drop-in (i.e. renames `object-cache-plst-orig.php` back to `object-cache.php`).
* Several comments have been added to explain the changes so that the context is clear for any future reference when looking at the code.
* The PR also adds a `function_exists()` check around the one function that the Performance Lab drop-in defines to prevent any fatal error e.g. in case another drop-in like `advanced-cache.php` loads the `object-cache.php` drop-in extra early.

### Without existing drop-in
1. Make sure there is no `wp-content/object-cache.php` file in your setup.
2. Activate PL plugin in WP Admin.
3. Validate that the `wp-content/object-cache.php` file now exists and has the same content as the `server-timing/object-cache.copy.php` file within the plugin.
4. Deactivate the PL plugin.
5. Validate that the `wp-content/object-cache.php` file no longer exists.

### With drop-in already present
1. Manually copy the plugin's `server-timing/object-cache.copy.php` file to `wp-content/object-cache.php`.
2. Activate PL plugin in WP Admin.
3. Validate that the `wp-content/object-cache.php` file still exists and has the same content as the `server-timing/object-cache.copy.php` file within the plugin.
4. Deactivate the PL plugin.
5. Validate that the `wp-content/object-cache.php` file no longer exists.

### With conflicting drop-in present
1. Place a custom PHP file (e.g. with just a comment, or you could use an actual `object-cache.php` drop-in from a plugin like e.g. W3 Total Cache) in `wp-content/object-cache.php`.
2. Activate PL plugin in WP Admin.
3. Validate that the `wp-content/object-cache.php` file still exists and has the same content as before. The plugin's own file should _not_ replace it.
4. Deactivate the PL plugin.
5. Validate that the `wp-content/object-cache.php` file _still_ exists (as it is not our own, so we shouldn't remove it).

### With old code (our drop-in plus renamed other drop-in)
1. Manually copy the plugin's `server-timing/object-cache.copy.php` file to `wp-content/object-cache.php`.
2. Place a custom PHP file (e.g. with just a comment, or you could use an actual `object-cache.php` drop-in from a plugin like e.g. W3 Total Cache) in `wp-content/object-cache-plst-orig.php`.
3. Activate PL plugin in WP Admin.
4. Validate that the `wp-content/object-cache.php` and `wp-content/object-cache-plst-orig.php` files still exist and have the same content as before.
5. Deactivate the PL plugin.
6. Validate that the `wp-content/object-cache.php` file now has the content that previously was in the `wp-content/object-cache-plst-orig.php` file, and that the `wp-content/object-cache-plst-orig.php` file is now removed (as in that case we need to restore the other drop-in in its original location, which was something the plugin used to do, so this is critical for backward compatibility with older PL plugin versions).

## Checklist

### DIFF
--- a/load.php
+++ b/load.php
@@ -274,6 +274,7 @@ perflab_load_active_and_valid_modules();
  * 'PERFLAB_DISABLE_OBJECT_CACHE_DROPIN' is set as true.
  *
  * @since 1.8.0
+ * @since n.e.x.t No longer attempts to use two of the drop-ins together.
  *
  * @global WP_Filesystem_Base $wp_filesystem WordPress filesystem subclass.
  */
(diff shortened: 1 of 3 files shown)


tillkruss (approved):

mukeshpanchal27 (approved):
Thanks @felixarntz, For each test case, PR functions properly.

10upsimon (approved):

10upsimon:
@felixarntz I've tested all scenarios in a `wp-env` and all passed. Approving.

PR #674, opened 2023-03-14
Ensure module `load.php` files really only load other code to prevent conflicts in standalone plugins

### LINKED ISSUE #667: Remove any functions, classes, constants etc. from module `load.php` files
## Feature Description

As part of #656, the milestone 1 will temporarily result in certain features being available as plugins and as modules. While the Performance Lab plugin will do all it can to ensure that a module doesn't load if the similar plugin is already loaded (see #666), this approach is not and cannot be 100% reliable: For example, when you already use one of the modules, and then activate the corresponding standalone plugin, it will throw an exception during activation as it would effectively redeclare the functions from the module. This is only relevant during activation, since at that point the plugin would be loaded after the module (while normally it is the other way around).

For that reason, and as a general good practice, we should reduce each module's `load.php` file to just that: A file that loads other files. Additionally, we should add a clause to each `load.php` file to not load the files if e.g. a function or constant from that feature is already defined (e.g. in the equivalent Performance Lab module or standalone plugin).

### Requirements

* All functions, classes, constants, etc. that are declared in a module's `load.php` file should be moved to other more specific files.
* Each module's `load.php` file should include a check for whether a certain key function or constant or class that is part of the module is already loaded. If so, the file should return early.
* Otherwise, all that the module `load.php` files should do going forward is load the other files.

felixarntz:
## Summary

Fixes #667

## Relevant technical choices

* Unless a module has already been using other files in a more specific way, this PR goes with the approach of splitting functions between a `hooks.php` file (functions that are used as hook callbacks) and a `helper.php` file (other functions, e.g. used by the hook functions).
* This PR looks like a massive change, but no code whatsoever was functionally modified. All functions touched were simply moved between files without modifications.
* The only code change that technically needs a closer review is all the `load.php` files, for whether they include the correct files and check for the correct function (any key function that is part of the respective module).

To avoid conflicts with other PRs, it would be great to review, iterate on, and merge this change ASAP.

Going forward, all modules will need to use their `load.php` file accordingly just to load other files. No actual code should be added directly to the `load.php` file of a module.

## Checklist


10upsimon (approved):

felixarntz:
@mukeshpanchal27 Note that there will probably be merge conflicts between this PR and #670. We'll just have to fix them based on which PR gets merged first.

mukeshpanchal27 (approved):
Thanks @felixarntz, The changes look good to me.

PR #681, opened 2023-03-17
Update the module constant

mukeshpanchal27:
## Summary
This PR is followup work in which the `build-plugins` command updates the constant value defined in #666.

## Checklist

### DIFF
--- a/bin/plugin/commands/build-plugins.js
+++ b/bin/plugin/commands/build-plugins.js
@@ -81,6 +81,13 @@ exports.handler = async () => {
 						regex: '@package\\s{1,}performance-lab',
 						result: '@package ' + pluginSlug,
 					} );
+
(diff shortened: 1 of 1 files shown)


felixarntz (approved):
@mukeshpanchal27 Thanks, this is a solid and simple solution. Let's not merge this before we have merged #670 though. Or maybe, since we're still discussing on that PR and the change here is so small, you could add it to the other PR instead so that we only have to have one PR to review and merge - also since the two changes are so closely related.

Maybe my previous idea of splitting them didn't make so much sense after all 🙃

felixarntz:
@mukeshpanchal27 Actually never mind, I see you have the branch for #670 as target branch. In that case, we can of course go ahead and merge this one in there. I'll do that now :)

PR #683, opened 2023-03-20
Fix running tests not working unless the repository is cloned to a `performance` directory

### LINKED ISSUE #673: Fix running tests not working unless the repository is cloned to a `performance` directory
## Bug Description

When having the repository checked out to another folder name than `performance`, the `npm run test-php` and `npm run test-php-multisite` scripts will fail since they have the `performance` path hard-coded.

We should fix that, for example so that the script detects the current folder name ("base name"). Or we find another solution that doesn't require specifying the full path to the PHPUnit config file.

## Steps to reproduce

1. Clone the repository to a folder called `performance-lab` for example.
2. Install dependencies via `composer install` and `npm install`.
3. Call `npm run test-php`.
4. See the command fail and not even get to running any tests.

mukeshpanchal27:
## Summary

Fixes #673

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,9 +25,9 @@
     "prelint-php": "wp-env run composer 'install --no-interaction'",
     "lint-php": "wp-env run composer run-script lint",
     "pretest-php": "wp-env run composer 'install --no-interaction'",
-    "test-php": "wp-env run phpunit 'phpunit -c /var/www/html/wp-content/plugins/performance/phpunit.xml.dist --verbose'",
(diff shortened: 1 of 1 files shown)


10upsimon (approved):
Approving as this worked for me. I cloned to a dir named `perf-test` and config path was correct as this snippet from my terminal output shows:

```
Configuration: /var/www/html/wp-content/plugins/perf-test/phpunit.xml.dist
```

joemcgill (approved):

felixarntz (approved):
Thanks @mukeshpanchal27!

PR #686, opened 2023-03-22
GitHub workflow to build and deploy standalone plugins

### LINKED ISSUE #638: Implement GitHub workflow to build and deploy standalone plugins

mukeshpanchal27:
## Summary

Fixes #638

## Relevant technical choices
For now, this PR adds a new GitHub workflow that triggers push and pull requests, which we update after the approach is approved by other members.

The main purpose is that I have to use the [10up deploy action](https://github.com/10up/action-wordpress-plugin-deploy) so the PR reads the `plugins.json` file and sets metric values for each module that we list in that file.

I have commented out the original `SVN_USERNAME` and `SVN_PASSWORD` so we can test workflow.

Thank you, @joemcgill for all of your advice, and thank you, @costdev for all of your assistance in setting up the matrix for GHA.

## Checklist


.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -0,0 +1,59 @@
...
+    branches:
+      - trunk
+      - 'release/**'
+      - 'feature/**'
+    paths:
+      - '.github/workflows/deploy-standalone-plugins.yml'
+    types:
+      - opened
+      - reopened
+      - synchronize
```
joemcgill: I assume you have these here for testing, but I think we'll need to change this trigger to something different like a release, or the addition of a tag, or something. What do you think?

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -0,0 +1,59 @@
...
+
+jobs:
+  release:
+    name: New Release
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
```
joemcgill, line 21: Can we add a name to each of these steps so it's easier to follow?

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -0,0 +1,59 @@
...
+      - uses: actions/checkout@v3
+      - id: set-matrix
+        run: echo "matrix="$(jq -c '{include:[keys[] as $k | {name:$k,slug:.[$k].slug,version:.[$k].version }]}' plugins.json) >> $GITHUB_OUTPUT
```
joemcgill: Some in-line docs explaining what's happening here could be useful.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Deploy Standalone Plugin - ${{ matrix.slug }}
+        uses: 10up/action-wordpress-plugin-deploy@stable
+        env:
+          #SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
+          #SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          SVN_PASSWORD: SVN_PASSWORD
+          SVN_USERNAME: SVN_USERNAME
+          SLUG: ${{ matrix.slug }}
+          VERSION: ${{ matrix.version }}
+          BUILD_DIR: ./build/${{ matrix.slug }}
```
joemcgill, line 64: I noticed that the `BUILD_DIR` value that gets passed here looks incorrect. I wonder if that is a consequence of the build step not being correct or if there's some thing in this workflow that needs to be changed? See: https://github.com/WordPress/performance/actions/runs/4489075665/jobs/7894465653#step:3:52

For testing purposes, it may be better for us to fork the 10up deploy action into a separate bash script where we can add logging for everything up to the actual SVN commit to make sure things are working the way we think. What do you think?

.github/workflows/deploy-standalone-plugins.yml
costdev: @joemcgill Does this work?
```yaml
# Load the JSON file and parse from "{name: {slug, version}, ...}" to "include: [{ name, slug, version }, ...]"
# for use in the matrix.
```

joemcgill: Yup! Even having a descriptive step name would help too.

.github/workflows/deploy-standalone-plugins.yml
costdev: This makes sense.

Performance Lab uses:
```yaml
on:
  release:
    types:
      - published
```

WordPress Beta Tester uses:
```yaml
on:
  push:
    tags:
    - "*"
```

mukeshpanchal27: Yes @joemcgill, Once we have thoroughly tested the workflow, we must update it with a `release` or push `tags`.

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: Thanks, @joemcgill and @costdev! Please look at https://github.com/WordPress/performance/tree/trunk/.github/workflows, we don't use the descriptive steps for any workflow, so IMO we have to just add inline docs with explanation. WDYT?

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: The name is added for each step.

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: The inline comment added in https://github.com/WordPress/performance/pull/686/commits/4eb793bc9d4eb20f1202068250e52212b42a0fc6

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: @joemcgill i have correct the path. https://github.com/WordPress/performance/actions/runs/4511968402/jobs/7944799925?pr=686#step:6:42

I have copied `deploy.sh` from 10up action so we can test it.

joemcgill: The thing that I noticed is that there's two folders in the path called "performance": `/home/runner/work/performance/performance/build/webp-uploads`

It's long as everything is working when you are debugging the workflow, that's probably fine, it just struck me as potentially incorrect.

deploy.sh
```diff
@@ -0,0 +1,170 @@
...
+    svn propset svn:mime-type "image/svg+xml" "$SVN_DIR/assets/"*.svg || true
+fi
+
+#Resolves => SVN commit failed: Directory out of date
+svn update
+
+svn status
+
+#echo "➤ Committing files..."
+#svn commit -m "Update to version $VERSION from GitHub" --no-auth-cache --non-interactive  --username "$SVN_USERNAME" --password "$SVN_PASSWORD"
```
joemcgill, line 156: Good idea to comment this out. 👍🏻

joemcgill:
Thanks, @mukeshpanchal27, this looks like it's headed in the right direction. What are the remaining blockers, and can I help you resolve them in someway? At minimum it seems like we're waiting on the first plug-in to be available on the.org repo, but we also need to decide on how this is getting triggered. Is that something that you want to figure out, or are you looking for direction on? Any other concerns?

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: Yes, initially I also thought that it was wrong, but after debugging I found that it was working as expected.

mukeshpanchal27:
Thanks, @joemcgill for your review.
- The deploy process look pretty good but still we are waiting for .org repo once we get it, how do we test it? Do we need to check a different version for the plugin to be deployed?
- We need to figure out when the workflow is triggered so we don't get blocked on this.

costdev:
If Performance Lab will have releases that shouldn't trigger the workflow, `on: release: types: published` doesn't seem appropriate.

Some combination of these might work:
```yaml
on:
  push:
    branches:
      - 'trunk'
    paths:
      - 'modules/**'
    tags:
      - '*'
```
Needs some thought though.

joemcgill:
> ...we are waiting for .org repo once we get it, how do we test it? Do we need to check a different version for the plugin to be deployed?

This is a tricky one. We should be confident that the script is going to work even before we have a released plugin to use this with. You can already see that the environment variables are being set correctly when you look at the failed task now, so I think we're good here. Once we have an official plugin SVN repo to test with, we can keep things as is and manually execute the workflow with this forked version of the script so we can see the whole process without pushing a new commit.

> We need to figure out when the workflow is triggered so we don't get blocked on this.

I agree. For now, I still think it makes sense for us to execute these anytime a new release is published, since that is most likely the way we will want to use this task. If there are exceptions where we need to push out a manual release of either the performance lab plugin or one of the standalone plugins, we can do that manually without using this automated process. That said, I'm happy for us to come up with a better way of doing this in an automated fashion, and @costdev's suggestion is one option we should consider.

costdev:
> Do we need to check a different version for the plugin to be deployed?

From the issue:

> It should then check out the individual SVN repositories for those plugins from the plugins.json file, and for each plugin replace the current trunk folder content with the newly built plugin.
>
> From there on, it should only proceed for each plugin if the version in the plugin's main file and the plugin's readme.txt file matches and is greater than the version that was previously in those files. (The only exception is if this is the initial release, i.e. there was no version before.) If so, the next steps for each plugin are:
>
>Copy the trunk folder to a new tags/$version folder.
>
>SVN-add / SVN-delete any files as needed based on svn status.
>
> Commit the new code.

:point_up: This seems to provide good guidance here.

> We need to figure out when the workflow is triggered so we don't get blocked on this.

From the issue:

> For triggering this workflow, as a starting point we can probably set it to only run when a new Performance Lab plugin release is made, i.e. this would also trigger a new release of any standalone plugins if their versions were intentionally increased. We can always enhance this later if we e.g. also need to trigger manual releases for those plugins. Note that in the future, once the standalone plugins are no longer part of the PL plugin, the trigger will have to be changed anyway, so this is just a temporary measure.

For now, if all of the above sections are met, I think we could use this:
```yaml
on:
  release:
    types: [published]
  workflow_dispatch:
    inputs:
      slug:
        type: string
        description: 'The slug of the plugin to deploy'
```
From there, updated plugins will be deployed when:
- Performance Lab is released, OR
- We manually trigger the deployment if the standalone update needs to deploy sooner.

`inputs.slug` can be checked and if not empty, can be used to only run for the specific plugin. This fine-grained control might be preferable if multiple plugins have been updated, but only one of them should be deployed before the Performance Lab release (e.g. fixing a Fatal Error)

mukeshpanchal27:
Thanks both for your feedback.

@joemcgill We got official approval for `WebP Uploads`. Do we need to upload the initial version to SVN before checking GHA? What is the next step on that? Once we did all the testing, we had to decide when the workflow should trigger.

joemcgill:
That’s great news that we have approval. As long as we have an SVN repo we should be able to test the workflow, we just don’t want to actually commit the new files to that repo until we’ve finished the whole build process.

mukeshpanchal27:
@joemcgill The workflow is now https://github.com/WordPress/performance/actions/runs/4582789944/jobs/8093233503?pr=686 ✅ and deploy process working look good. Can you please review it. Thanks!

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -0,0 +1,63 @@
...
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+          cache: npm
+      - name: Install npm dependencies
+        run: npm ci
+      - name: Building standalone plugins
+        run: npm run build-plugins
+      - name: Deploy Standalone Plugin - ${{ matrix.slug }}
+        #uses: 10up/action-wordpress-plugin-deploy@stable
```
mukeshpanchal27, line 54: I used same `deploy.sh` that used in `10up/action-wordpress-plugin-deploy@stable` action.

.github/workflows/deploy-standalone-plugins.yml
felixarntz: @mukeshpanchal27 Can you please add both versions of the code here so that we can proceed with the full code review? You can comment out the version which we should eventually use (on release or manual dispatch), and then add a `TODO` comment to get rid of the current triggers here before we merge this.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -0,0 +1,63 @@
...
+        env:
+          #SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
+          #SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          SVN_PASSWORD: SVN_PASSWORD
+          SVN_USERNAME: SVN_USERNAME
```
felixarntz: Why are these two lines commented out?

.github/workflows/deploy-standalone-plugins.yml
felixarntz: Can you please add a code comment explaining that inline so that it's captured for any future reviewers of this code?
```suggestion
        # The `deploy.sh` script called below is almost a copy of the one used in `10up/action-wordpress-plugin-deploy@stable`.
```

felixarntz: Or are we able to actually use the original `10up/action-wordpress-plugin-deploy@stable` action directly? Is this file just a temporary change, or is it meant as the actual solution to merge?

deploy.sh
```diff
@@ -0,0 +1,170 @@
...
+	echo "ℹ︎ Copying files from build directory..."
+	rsync -rc "$BUILD_DIR/" trunk/ --delete --delete-excluded
+fi
+
+# Copy dotorg assets to /assets
+if [[ -d "$GITHUB_WORKSPACE/$ASSETS_DIR/" ]]; then
+	rsync -rc "$GITHUB_WORKSPACE/$ASSETS_DIR/" assets/ --delete
+else
+	echo "ℹ︎ No assets directory found; skipping asset copy"
+fi
```
felixarntz, line 119: I think we should remove this code for, since we still have not made a decision on which assets to include with the standalone plugins. With this code, we would automatically put the same assets as the Performance Lab plugin uses for every standalone plugin as well.

Maybe that's what we want to do, maybe not. This decision should not block the very initial release of the standalone plugin, so let's remove for now and we can figure that out separately.

felixarntz (changes requested):
@mukeshpanchal27 This looks great for the most part.

In a few places, it would be great if you could explain which changes are meant to be temporary (before the merge of this PR) and which changes are meant to be actually merged. Please use code comments to make that clear from the code itself - there are a few places where I have questions about what you intend to get merged vs what is just temporary.

joemcgill:
@felixarntz The `deploy.sh` code here is a direct fork of the `10up/action-wordpress-plugin-deploy` action and was just added to this PR so we could test out the integration of our deployment workflow with this action without actually pushing the commit to the SVN repo. We should disregard any changes to that file in this PR since it will be removed once we have completed a dry run.

The next steps here, as I see it, are to update the `.github/workflows/deploy-standalone-plugins.yml` file so the workflow is triggered correctly, and so it makes use of the SVN secrets from this repo. It would be best if we could also trigger this workflow manually in order to do one last dry run before swapping out the bash script for the `10up/action-wordpress-plugin-deploy` action.

deploy.sh
joemcgill: No need to change this file, as it will be replaced by the `10up/action-wordpress-plugin-deploy` action.

felixarntz: @joemcgill Makes sense from a technical perspective, see my above note though on how we should handle the assets. While we can do that without a change to this file, this is still a decision that needs to be sorted as part of this PR, at least for the short-term (initial launch of WebP Uploads).

felixarntz:
Thanks for clarifying @joemcgill. In that case, some of my feedback above does not need to be actioned.

That said, two most critical things to work out from my perspective:
* Finalize the trigger logic: Let's add this in commented out code already so that we can review it. Other than when a release is published on GitHub, we should also allow triggering manually via `workflow_dispatch`.
* Decide how to handle assets: Right now, the way the code works, it'll use the same assets as PL. We may want to change that to point to another `ASSETS_DIR` to not use any assets at all for the launch. Let's discuss that separately.

joemcgill:
Thanks @felixarntz.

> Finalize the trigger logic: Let's add this in commented out code already so that we can review it. Other than when a release is published on GitHub, we should also allow triggering manually via workflow_dispatch.

In that case, I suggest we go ahead with what @costdev proposed for a manual workflow:

```
on:
  release:
    types: [published]
  workflow_dispatch:
    inputs:
      slug:
        type: string
        description: 'The slug of the plugin to deploy'
```

Since `workflow_dispatch` actions can only be run if the workflow is in the default branch, I don't think we need to comment that out while we review.

deploy.sh
```diff
@@ -0,0 +1,170 @@
+#!/bin/bash
+
+# Note that this does not use pipefail
+# because if the grep later doesn't match any deleted files,
+# which is likely the majority case,
+# it does not exit with a 0, and I only care about the final exit.
```
10upsimon, line 6:
```suggestion
# Note that this does not use pipefail
# because if the grep later doesn't match any deleted files,
# which is likely to be the case the majority of the time,
# it does not exit with 0, as we're interested in the final exit.
```

@mukeshpanchal27 this is a _very_ nit pick comment, but we should not be talking in a _first person_ style within comments.

10upsimon (changes requested):
@mukeshpanchal27 I have no other feedback aside from what @joemcgill and @felixarntz have provided, other than a nit pick suggestion to clean up a comment, most specifically the first person speech aspect, which I think we should always avoid in comments.

deploy.sh
joemcgill: Thanks @10upsimon. This is not something we will need to handle since this is upstream code from the 10up deploy action and this file will be removed. That said, I agree with your comment and something that we could address upstream.

felixarntz:
Thanks @joemcgill, SGTM!

> Since `workflow_dispatch` actions can only be run if the workflow is in the default branch, I don't think we need to comment that out while we review.

Makes sense to me. It would still be good to annotate in the PR code with comments which blocks of code are temporary vs which code should eventually be merged.

.github/workflows/deploy-standalone-plugins.yml
felixarntz: Never mind, @joemcgill clarified that we want to eventually use the actual 10up GitHub workflow.

Something we need to make sure here is that we remove the temporary workflow triggers _before_ we switch to that action, to avoid accidentally publishing the WebP Uploads plugin from this PR.

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: Because if somehow the workflow starts working, then it will not commit the code in SVN because of wrong access. I'll update it once everything is finalised. Thanks.

deploy.sh
mukeshpanchal27: Thank you, @10upsimon and @joemcgill, for the comment. I have an open PR in upstream: https://github.com/10up/action-wordpress-plugin-deploy/pull/120.

mukeshpanchal27:
> * Decide how to handle assets: Right now, the way the code works, it'll use the same assets as PL. We may want to change that to point to another `ASSETS_DIR` to not use any assets at all for the launch. Let's discuss that separately.

@felixarntz In the https://github.com/WordPress/performance/pull/686/commits/5d4e5c070db2784410fbe0f5fb30481490bca9ee i set `ASSETS_DIR` variable which find the `.wordpress-org` folder in module directory. Now it skip assets copy https://github.com/WordPress/performance/actions/runs/4604169975/jobs/8134862053#step:6:53 because we don't have that folder in module directory. In future we can manage the assets from that folder.

felixarntz (commented):
@mukeshpanchal27 Thank you for the updates, this looks great now.

Consider the code approved from my side, of course pending the final changes following the TODO comments.

We also still have to make a decision outside of this PR on the assets to use, but conceptually setting `ASSETS_DIR` to a module-specific assets directory certainly makes more sense than pointing to the same directory that the PL plugin uses.

mukeshpanchal27:
@joemcgill @felixarntz In [748df3b](https://github.com/WordPress/performance/pull/686/commits/748df3b0f66ad2f7d770043ab6444096dc186d95) commit, I update PR so it will set the matrix for the release workflow conditional base.

**Note:** For manual triggers, we don't use plugin names, so I have not included them.

10upsimon (approved):
@mukeshpanchal27 this is looking good to me now.

costdev (approved):
Think this looks good to me too @mukeshpanchal27! Thanks! 👍

joemcgill (approved):
This is a clever way to use the inputs of the manual workflow. Nice work @mukeshpanchal27. I say we go with this for our dry run, and we can decide if we want to improve the way the inputs are working before finalizing the whole feature.

deploy.sh
```diff
@@ -0,0 +1,170 @@
...
+#  echo "Generating zip file..."
+
+  # use a symbolic link so the directory in the zip matches the slug
+#  ln -s "${SVN_DIR}/trunk" "${SVN_DIR}/${SLUG}"
+#  zip -r "${GITHUB_WORKSPACE}/${SLUG}.zip" "$SLUG"
+#  unlink "${SVN_DIR}/${SLUG}"
+
+#  echo "zip-path=${GITHUB_WORKSPACE}/${SLUG}.zip" >> "${GITHUB_OUTPUT}"
+#  echo "✓ Zip file generated!"
+#fi
```
felixarntz: Should this be commented out? As far as I understand this code (not too familiar with GH Actions tbh), it creates a ZIP file for the built plugin (as an artifact)? If that's the case, shouldn't we have this part of the code run already so that we can test that the ZIP file works as expected?

felixarntz (approved):
@mukeshpanchal27 One last question here, but this is looking basically ready to merge.

cc @joemcgill

deploy.sh
joemcgill: This part of the code creates a ZIP file from the release which can then be passed to additional workflow steps. See: https://github.com/10up/action-wordpress-plugin-deploy#inputs. By default, this will be set as `false`, so it it's fine to leave this comment it out. The reason that this probably needed to be commented out, is that our workflow does not set the environment variable, which might make this line result in an error.

mukeshpanchal27: Great catch! Remove comment in https://github.com/WordPress/performance/pull/686/commits/d315bda0cac856b3c900654dffb4a03a9505f543

joemcgill: Never mind, just saw the updated commit that uncomments that code, and it seems fine.

felixarntz: Thank you both!

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -0,0 +1,79 @@
...
+    branches:
+      - trunk
+      - 'release/**'
+      - 'feature/**'
+    paths:
+      - '.github/workflows/deploy-standalone-plugins.yml'
+    types:
+      - opened
+      - reopened
+      - synchronize
```
felixarntz: This bit should preferably be removed before we merge this PR.
```suggestion
```

felixarntz (commented):
One last thing here before we merge this pull request (thanks @joemcgill).

PR #690, opened 2023-03-27
Sanitize target param before using it

### LINKED ISSUE #689: Add missing sanitization to image editing requests
## Bug Description

The `$_REQUEST['target']` parameter used in the image editing compatibility logic for WebP is not being sanitized. Since it is only used in string comparison, this is not actual a security flaw, but we should still sanitize it for good measure.

I would suggest we use [`sanitize_key()`](https://developer.wordpress.org/reference/functions/sanitize_key/) on those occurrences, since the parameter is essentially just used with lowercase "identifiers" to represent a certain target.

felixarntz:
## Summary

Fixes #689

## Checklist

### DIFF
--- a/modules/images/webp-uploads/image-edit.php
+++ b/modules/images/webp-uploads/image-edit.php
@@ -120,7 +120,7 @@ function webp_uploads_update_image_onchange( $override, $file_path, $editor, $mi
 
 			$old_metadata = wp_get_attachment_metadata( $post_id );
 			$resize_sizes = array();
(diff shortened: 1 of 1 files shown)


joemcgill (approved):

mukeshpanchal27 (approved):
Thanks @felixarntz, LGTM!

spacedmonkey (commented):
For what is worth, I would have approved this!

PR #695, opened 2023-04-03
Include `WordPress-Extra` rules in PHPCS configuration and fix resulting problems

### LINKED ISSUE #691: Include `WordPress-Extra` rules in PHPCS configuration and fix resulting problems
Our `phpcs.xml.dist` config file currently does not include the `WordPress-Extra` rules, which are recommended to use for plugins. We should add it and fix the resulting validation errors.

The latter will require manual assessment: Errors will either need to be actually fixed, or they may be invalid due to the automated detection being limited, in which case a code comment to exclude the check for the "offending" line should be added.

mukeshpanchal27:
## Summary

Fixes #691 

The PR add Include `WordPress-Extra` rules in PHPCS configuration and fix resulting problems.

For now i have exclude `sqlite` module files as it shows many files to change. 

## Checklist

### DIFF
--- a/admin/load.php
+++ b/admin/load.php
@@ -138,18 +138,13 @@ function perflab_render_modules_page_field( $module_slug, $module_data, $module_
 			<?php if ( perflab_can_load_module( $module_slug ) ) { ?>
 				<input type="checkbox" id="<?php echo esc_attr( "{$base_id}_enabled" ); ?>" name="<?php echo esc_attr( "{$base_name}[enabled]" ); ?>" aria-describedby="<?php echo esc_attr( "{$base_id}_description" ); ?>" value="1"<?php checked( $enabled ); ?>>
 				<?php
+					printf(
+						/* translators: %s: module name */
+						esc_html__( 'Enable %s', 'performance-lab' ),
+						esc_html( $module_data['name'] )
+					);
 				if ( $module_data['experimental'] ) {
-					printf(
-						/* translators: %s: module name */
-						__( 'Enable %s <strong>(experimental)</strong>', 'performance-lab' ),
-						esc_html( $module_data['name'] )
-					);
-				} else {
-					printf(
-						/* translators: %s: module name */
-						__( 'Enable %s', 'performance-lab' ),
-						esc_html( $module_data['name'] )
-					);
+					echo '<strong>' . esc_html__( '(experimental)', 'performance-lab' ) . '</strong>';
 				}
 				?>
 			<?php } else { ?>
@@ -173,7 +168,7 @@ function perflab_render_modules_page_field( $module_slug, $module_data, $module_
 				} else {
 					printf(
 						/* translators: %s: module name */
-						__( '%s is already part of your WordPress version and therefore cannot be loaded as part of the plugin.', 'performance-lab' ),
+						esc_html__( '%s is already part of your WordPress version and therefore cannot be loaded as part of the plugin.', 'performance-lab' ),
 						esc_html( $module_data['name'] )
 					);
 				}
@@ -250,7 +245,7 @@ function perflab_get_modules( $modules_root = null ) {
 
 	$modules      = array();
 	$module_files = array();
-	$modules_dir  = @opendir( $modules_root );
+	$modules_dir  = @opendir( $modules_root ); // phpcs:disable WordPress.PHP.NoSilencedErrors.Discouraged
 
 	// Modules are organized as {focus}/{module-slug} in the modules folder.
 	if ( $modules_dir ) {
@@ -265,7 +260,7 @@ function perflab_get_modules( $modules_root = null ) {
 				continue;
 			}
 
-			$focus_dir = @opendir( $modules_root . '/' . $focus );
+			$focus_dir = @opendir( $modules_root . '/' . $focus ); // phpcs:disable WordPress.PHP.NoSilencedErrors.Discouraged
 			if ( $focus_dir ) {
 				// phpcs:ignore WordPress.CodeAnalysis.AssignmentInCondition.FoundInWhileCondition
 				while ( ( $file = readdir( $focus_dir ) ) !== false ) {
@@ -274,7 +269,7 @@ function perflab_get_modules( $modules_root = null ) {
 						continue;
 					}
 
-					$module_dir = @opendir( $modules_root . '/' . $focus . '/' . $file );
+					$module_dir = @opendir( $modules_root . '/' . $focus . '/' . $file ); // phpcs:disable WordPress.PHP.NoSilencedErrors.Discouraged
 					if ( $module_dir ) {
 						// phpcs:ignore WordPress.CodeAnalysis.AssignmentInCondition.FoundInWhileCondition
 						while ( ( $subfile = readdir( $module_dir ) ) !== false ) {
--- a/admin/load.php
+++ b/admin/load.php
@@ -144,7 +144,7 @@ function perflab_render_modules_page_field( $module_slug, $module_data, $module_
 						esc_html( $module_data['name'] )
 					);
 				if ( $module_data['experimental'] ) {
-					echo '<strong>' . esc_html__( '(experimental)', 'performance-lab' ) . '</strong>';
+					echo ' <strong>' . esc_html__( '(experimental)', 'performance-lab' ) . '</strong>';
 				}
 				?>
 			<?php } else { ?>
--- a/admin/load.php
+++ b/admin/load.php
@@ -254,7 +254,8 @@ function perflab_get_modules( $modules_root = null ) {
 	$modules      = array();
 	$module_files = array();
 	// PHPCS ignore reason: A modules directory is always present.
-	$modules_dir = @opendir( $modules_root ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+	// phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+	$modules_dir = @opendir( $modules_root );
 
 	// Modules are organized as {focus}/{module-slug} in the modules folder.
 	if ( $modules_dir ) {
@@ -270,7 +271,8 @@ function perflab_get_modules( $modules_root = null ) {
 			}
 
 			// PHPCS ignore reason: Only the focus area directory is allowed.
-			$focus_dir = @opendir( $modules_root . '/' . $focus ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+			// phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+			$focus_dir = @opendir( $modules_root . '/' . $focus );
 			if ( $focus_dir ) {
 				// phpcs:ignore WordPress.CodeAnalysis.AssignmentInCondition.FoundInWhileCondition
 				while ( ( $file = readdir( $focus_dir ) ) !== false ) {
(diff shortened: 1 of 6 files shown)


phpcs.xml.dist
```diff
@@ -68,4 +71,9 @@
 	<rule ref="PHPCompatibility.FunctionDeclarations.NewReturnTypeDeclarations.boolFound">
 		<exclude-pattern>tests/utils/*</exclude-pattern>
 	</rule>
+	<!--
+	Prevent errors caused by WordPress Coding Standards not supporting PHP 8.0+.
+	See https://github.com/WordPress/WordPress-Coding-Standards/issues/2035
+	-->
+	<ini name="error_reporting" value="E_ALL &#38; ~E_DEPRECATED" />
```
felixarntz, line 79: 👍

phpcs.xml.dist
```diff
@@ -5,8 +5,11 @@
 	<rule ref="PHPCompatibility"/>
 	<config name="testVersion" value="5.6-"/>
 
-	<rule ref="WordPress-Core"/>
 	<rule ref="WordPress-Docs"/>
+	<rule ref="WordPress-Extra">
+		<exclude-pattern>tests/*</exclude-pattern>
+		<exclude-pattern>modules/database/sqlite/*</exclude-pattern>
```
felixarntz, line 12: Why is this module specifically excluded? Can you please add a code comment to explain, as this isn't exactly clear to me?

admin/load.php
```diff
@@ -138,18 +138,13 @@ function perflab_render_modules_page_field( $module_slug, $module_data, $module_
...
-				} else {
-					printf(
-						/* translators: %s: module name */
-						__( 'Enable %s', 'performance-lab' ),
+						esc_html__( 'Enable %s', 'performance-lab' ),
 						esc_html( $module_data['name'] )
 					);
+				if ( $module_data['experimental'] ) {
+					echo ' <strong>' . esc_html__( '(experimental)', 'performance-lab' ) . '</strong>';
 				}
```
felixarntz: This fix is not right, translation strings should not be concatenated. Please use the string as it was before.

What we need to change is how it is being escaped. We cannot use `esc_html()` here because it contains HTML, so we need to escape it so that only that specific HTML is allowed. Please use `wp_kses( $translated_string, array( 'strong' => array() ) )`. This way the string is safe, the only allowed HTML will be the `<strong>` tag.

modules/images/webp-uploads/image-edit.php
```diff
@@ -120,7 +120,7 @@ function ( $metadata, $post_meta_id ) use ( $post_id, $file_path, $mime_type, $e
 
 			$old_metadata = wp_get_attachment_metadata( $post_id );
 			$resize_sizes = array();
-			$target       = isset( $_REQUEST['target'] ) ? sanitize_key( $_REQUEST['target'] ) : 'all';
+			$target       = isset( $_REQUEST['target'] ) ? sanitize_key( $_REQUEST['target'] ) : 'all'; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
felixarntz: It may be worth adding a comment to explain why this error is being silenced here, especially because this is security related:
```suggestion
			// A nonce check is not necessary here as this logic directly ties in with WordPress core logic which already has one.
			$target       = isset( $_REQUEST['target'] ) ? sanitize_key( $_REQUEST['target'] ) : 'all'; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```

felixarntz (commented):
@mukeshpanchal27 A few points of feedback here, but mostly looks good.

modules/database/sqlite/wp-includes/sqlite/class-perflab-sqlite-pdo-engine.php
```diff
@@ -326,7 +326,7 @@ private function prepare_directory() {
 		global $wpdb;
 		$u = umask( 0000 );
 		if ( ! is_dir( FQDBDIR ) ) {
-			if ( ! @mkdir( FQDBDIR, 0704, true ) ) {
+			if ( ! @mkdir( FQDBDIR, 0704, true ) ) { // phpcs:disable WordPress.PHP.NoSilencedErrors.Discouraged
```
joemcgill: If we're ignoring WordPress-Extra checks for the SQLite module in our `phpcs.xml.dist`, then let's not modify any files in that module.

phpcs.xml.dist
joemcgill: Nice find, and great use of inline commenting here ✨

joemcgill (changes requested):
This is on the right track. Thanks @mukeshpanchal27! Besides the translation fix that Felix already noted, the only other thing I would like to change is the way we're handling lines that we want to ignore.

First, you should always use `phpcs:ignore` instead of `phpcs:disable` to ignore a specific line, because `phpcs:disable` will disable that sniff for the rest of the file, unless paired with a `phpcs:enable` rule later in the file. 

See: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-parts-of-a-file

The other thing I'd like is for us to always include an inline comment explaining why something is being ignored, as @felixarntz suggested with the one nonce example. This makes it very clear why the line was added and helps determine if/when it can be removed in the future.

For readability, it would be easier to put these `phpcs:ignore` statements on the line above the code it is ignoring with the explanation right above. [Here is an example from WP Core](https://github.com/WordPress/wordpress-develop/commit/ab9aee8af474ac512b31b012f3c7c44fab31a990) that shows what this could look like

phpcs.xml.dist
mukeshpanchal27: In https://github.com/WordPress/performance/pull/695/commits/071ffacf0181bf7c3370d3d11a666e5a1ec54a87, I have added `TODO` comment for sqlite exclusion rule.

admin/load.php
mukeshpanchal27: Thanks for pointing it out. I have address this in https://github.com/WordPress/performance/pull/695/commits/071ffacf0181bf7c3370d3d11a666e5a1ec54a87

modules/database/sqlite/wp-includes/sqlite/class-perflab-sqlite-pdo-engine.php
mukeshpanchal27: 👍

mukeshpanchal27:
Thanks, @felixarntz and @joemcgill for the feedback. I have addressed that feedback, and PR is ready for review. Thanks!

modules/images/dominant-color/hooks.php
```diff
@@ -236,7 +236,9 @@ function dominant_color_filter_content_tags( $content, $context = null ) {
  */
 function dominant_color_add_inline_style() {
 	$handle = 'dominant-color-styles';
-	wp_register_style( $handle, false );
+	// PHPCS ignore reason: Dominant color add single line inline style that does not change frequently,
+	// so we need to avoid specifying a version at all.
```
felixarntz:
```suggestion
	// so the version number is not used.
```

felixarntz (approved):
Thanks @mukeshpanchal27! LGTM, just a tiny not-blocking note.

Can you please open a follow up issue to address the relevant problems in the SQLite module so that we don't forget about that?

admin/load.php
```diff
@@ -245,7 +253,8 @@ function perflab_get_modules( $modules_root = null ) {
 
 	$modules      = array();
 	$module_files = array();
-	$modules_dir  = @opendir( $modules_root ); // phpcs:disable WordPress.PHP.NoSilencedErrors.Discouraged
+	// PHPCS ignore reason: A modules directory is always present.
```
joemcgill, line 256: Based on all of these reasons listed for silencing `opendir`, it seems like we should be able to remove the `@` silencing warnings throughout this file. Do you know if there is a reason that these were silenced initially?

modules/images/dominant-color/hooks.php
```diff
@@ -236,7 +236,9 @@ function dominant_color_filter_content_tags( $content, $context = null ) {
  */
 function dominant_color_add_inline_style() {
 	$handle = 'dominant-color-styles';
-	wp_register_style( $handle, false );
+	// PHPCS ignore reason: Dominant color add single line inline style that does not change frequently,
+	// so the version number is not used.
```
joemcgill: We can simplify this to one line. 
```suggestion
	// PHPCS ignore reason: version not used since this handle is only registered for adding an inline style.
```

modules/images/webp-uploads/hooks.php
```diff
@@ -242,7 +242,7 @@ function webp_uploads_wp_get_missing_image_subsizes( $missing_sizes, $image_meta
 	 * @see wp_update_image_subsizes()
 	 * @see wp_get_missing_image_subsizes()
 	 */
-	$trace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS, 10 );
+	$trace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS, 10 ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_debug_backtrace
```
joemcgill: This needs a in-line comment explaining the reason the sniff is ignored.

modules/images/webp-uploads/image-edit.php
```diff
@@ -120,7 +120,8 @@ function ( $metadata, $post_meta_id ) use ( $post_id, $file_path, $mime_type, $e
 
 			$old_metadata = wp_get_attachment_metadata( $post_id );
 			$resize_sizes = array();
-			$target       = isset( $_REQUEST['target'] ) ? sanitize_key( $_REQUEST['target'] ) : 'all';
+			// PHPCS ignore reason: A nonce check is not necessary here as this logic directly ties in with WordPress core logic which already has one.
```
joemcgill: We are modifying the behavior of what gets saved to the database based on the value of this `$_REQUEST` property, so it would be best if we could identify where the nonce check is being handled.

modules/images/webp-uploads/image-edit.php
```diff
@@ -237,7 +238,7 @@ function ( $metadata, $post_meta_id ) use ( $post_id, $file_path, $mime_type, $e
  * @return array The updated metadata for the attachment to be stored in the meta table.
  */
 function webp_uploads_update_attachment_metadata( $data, $attachment_id ) {
-	$trace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS, 10 );
+	$trace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS, 10 ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_debug_backtrace
```
joemcgill: This needs an in-line common explaining why were ignoring this sniff.

modules/images/webp-uploads/image-edit.php
```diff
@@ -272,7 +273,7 @@ function webp_uploads_update_attachment_metadata( $data, $attachment_id ) {
  * @return array The updated metadata for the attachment.
  */
 function webp_uploads_backup_sources( $attachment_id, $data ) {
-	$target = isset( $_REQUEST['target'] ) ? sanitize_key( $_REQUEST['target'] ) : 'all';
+	$target = isset( $_REQUEST['target'] ) ? sanitize_key( $_REQUEST['target'] ) : 'all'; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
joemcgill: This needs an in-line comment, and preferably a reference to where nonce verification is being handled.

joemcgill (commented):
This is much better. I have a bit more feedback based on your changes that I have left inline. I still think that the in-line comments explaining why we are ignoring some PHPCS sniffs would be more readable if the ignore statement was right below the in-line comment on the line above the line being ignored, like this:

```
// PHPCS ignore reason: explanation of why this is ignored
// phpcs:ignore WordPress.PHP.SniffRule.Reason
$var = example()
```

admin/load.php
mukeshpanchal27: Thanks @joemcgill for asking. This change was introduced from the initial version, so may be @felixarntz will know the reason behind this?

If it is not blocker can we open a separate issue and discuss it there?

mukeshpanchal27:
Thanks @joemcgill for the review. In https://github.com/WordPress/performance/pull/695/commits/1aab8fad21b06dfd1cf97c67ca8bb9cd907e50d4 i have address all the feedback. PR is ready for another round of feedback.

joemcgill (approved):
Thanks, @mukeshpanchal27

admin/load.php
mukeshpanchal27: Open at https://github.com/WordPress/performance/issues/709

PR #696, opened 2023-04-04
Generator meta tag for WebP Uploads standalone plugin

### LINKED ISSUE #692: Support standalone plugins in PL generator meta tag
The Performance Lab plugin has had a generator meta tag since shortly after its launch (see #321). Now that we are publishing the modules as standalone plugins and eventually will break them out, we need to also iterate on how that generator tag includes the relevant information. We would still want to make sure the generator tag includes which of the modules or equivalent standalone plugins the site uses.

I can see two alternative approaches for this:
1. Either we include the module name in the generator tag like we do today, even if it's actually the module's equivalent standalone plugin that is being used.
2. Or we keep the existing generator tag as is, only for modules, and introduce a second generator tag that captures any of the standalone plugins including their versions.

On the one hand, approach 1. is simpler and ties in seamlessly with how the generator tag works today. Any tooling that makes use of the data to measure adoption (using the public HTTP Archive dataset) will continue to function without modifications.

On the other hand, approach 2. would clearly separate which features are used as modules vs which features are used as plugins, and it could also include each standalone plugin's version number. This provides more information, but is also a bit more effort and might be unnecessarily verbose.

Do we actually need the additional information that approach 2. provides? If yes, I think we should go with it. If not, I think we should go with approach 1.

mukeshpanchal27:
## Summary

Fixes #692 

This PR adds a generator meta tag for the WebP Uploads module. We can use the same function for other modules once the WebP Uploads plugin is released.

## Checklist

### DIFF
--- a/modules/images/webp-uploads/hooks.php
+++ b/modules/images/webp-uploads/hooks.php
@@ -773,7 +773,7 @@ add_filter( 'wp_editor_set_quality', 'webp_uploads_modify_webp_quality', 10, 2 )
 function webp_uploads_render_generator() {
 	if (
 		defined( 'WEBP_UPLOADS_VERSION' ) &&
-		! str_starts_with( constant( 'WEBP_UPLOADS_VERSION' ), 'Performance Lab ' )
+		! str_starts_with( WEBP_UPLOADS_VERSION, 'Performance Lab ' )
 	) {
 		echo '<meta name="generator" content="WebP Uploads ' . esc_attr( WEBP_UPLOADS_VERSION ) . '">' . "\n";
 	}
--- a/modules/images/webp-uploads/hooks.php
+++ b/modules/images/webp-uploads/hooks.php
@@ -775,7 +775,7 @@ function webp_uploads_render_generator() {
 		defined( 'WEBP_UPLOADS_VERSION' ) &&
 		! str_starts_with( constant( 'WEBP_UPLOADS_VERSION' ), 'Performance Lab ' )
 	) {
-		echo '<meta name="generator" content="Performance Lab plugin: webp-uploads:' . WEBP_UPLOADS_VERSION . '" />' . "\n";
+		echo '<meta name="generator" content="WebP Uploads ' . esc_attr( WEBP_UPLOADS_VERSION ) . '">' . "\n";
 	}
 }
 add_action( 'wp_head', 'webp_uploads_render_generator' );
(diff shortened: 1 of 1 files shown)


modules/images/webp-uploads/hooks.php
```diff
@@ -762,3 +762,20 @@ function webp_uploads_modify_webp_quality( $quality, $mime_type ) {
...
+ * See {@see 'wp_head'}.
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_render_generator() {
+	if (
+		defined( 'WEBP_UPLOADS_VERSION' ) &&
+		! str_starts_with( constant( 'WEBP_UPLOADS_VERSION' ), 'Performance Lab ' )
+	) {
+		echo '<meta name="generator" content="Performance Lab plugin: webp-uploads:' . WEBP_UPLOADS_VERSION . '" />' . "\n";
```
felixarntz: No need to mention Performance Lab here, particularly as this will only be added in the standalone plugin. Also maybe we should use `esc_attr()` for the dynamic bit and avoid the old HTML syntax with a self-closing slash?
```suggestion
		echo '<meta name="generator" content="WebP Uploads ' . esc_attr( WEBP_UPLOADS_VERSION ) . '">' . "\n";
```

felixarntz (approved):
Thanks @mukeshpanchal27! This is almost good to go, a few small suggestions on the output to standardize it more with e.g. how the PL plugin and WP core output it.

joemcgill (approved):
This looks good to me. I think we may want to consider a future enhancement where we can move a lot of the shared library code to a central place that gets built into the individual plugins. I'll create a new issue for that after we finish this process.

modules/images/webp-uploads/hooks.php
```diff
@@ -762,3 +762,20 @@ function webp_uploads_modify_webp_quality( $quality, $mime_type ) {
...
+ * Displays the HTML generator tag for the WebP Uploads plugin.
+ *
+ * See {@see 'wp_head'}.
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_render_generator() {
+	if (
+		defined( 'WEBP_UPLOADS_VERSION' ) &&
+		! str_starts_with( constant( 'WEBP_UPLOADS_VERSION' ), 'Performance Lab ' )
```
felixarntz:
```suggestion
		! str_starts_with( WEBP_UPLOADS_VERSION, 'Performance Lab ' )
```

felixarntz: Unnecessary use of `constant()` here.

felixarntz (approved):
@mukeshpanchal27 LGTM! One tiny non-blocking thing, I think I can quickly fix that.

PR #704, opened 2023-04-06
Configure `Dominant Color` and `Fetchpriority` modules for their standalone plugins

### LINKED ISSUE #640: Set up additional standalone plugins for Dominant Color and Fetchpriority
## Feature Description

As a last step of creating standalone plugins for (some of) the modules in the Performance Lab plugin, we should finalize the config list from #635 to include all modules that should be published as standalone plugins. As part of this, we will need to set up their directories in the WordPress.org plugin repository, then build and deploy them using the tooling that we built in the issues #635, #636, #638, and #639.

This should likely be the last step of that work, to complete at least the first milestone of unbundling the Performance Lab plugin. Which modules to publish as standalone, and how to proceed after that is pending additional discussion in #618.

### Requirements

**Before working on this issue, let's wait until #638 is completed, and until a first version of the WebP Uploads plugin was successfully released.**

* `readme.txt` files should be added to the module directories `images/dominant-color` and `images/fetchpriority` for their upcoming standalone plugins, following the conventions from #636.
* The `plugins.json` config file from #635 should be expanded to include the following modules and their standalone plugin information:
    * `images/dominant-color`: plugin slug `dominant-color-images` (a plugin with slug `dominant-color` already exists), plugin version: `1.0.0`
    * `images/fetchpriority`: plugin slug `fetchpriority`, plugin version: `1.0.0`

felixarntz:
## Summary

Fixes #640.

**Do not merge this pull request until the two plugin repositories for these two standalone plugins have been created and approved.** Likely this will only be as part of the 2.3.0 milestone.

## Checklist


modules/images/fetchpriority/readme.txt
```diff
@@ -0,0 +1,52 @@
+=== Fetchpriority ===
+
+Contributors:      wordpressdotorg
+Requires at least: 6.1
+Tested up to:      6.2
+Requires PHP:      5.6
+Stable tag:        1.0.0
+License:           GPLv2 or later
+License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+Tags:              performance, images, dominant color
```
joemcgill: Need to update the tags here?

```suggestion
Tags:              performance, images, fetchpriority
```

joemcgill (changes requested):
Found one issue that needs to be addressed before this is merged. 

Also, since we're duplicating a lot of code between all the modules in their respective `hooks.php` and `load.php` files to handle things like avoiding conflicts with the Performance Lab plugin and outputting generator meta tags. I think there is an opportunity to consolidate these types of things to a shared file that got included in each plugin during the build process. There are tradeoffs to consider, for sure, but something we may want to consider.

modules/images/fetchpriority/readme.txt
felixarntz: Good catch, updated!

felixarntz:
@joemcgill 

> Also, since we're duplicating a lot of code between all the modules in their respective `hooks.php` and `load.php` files to handle things like avoiding conflicts with the Performance Lab plugin and outputting generator meta tags. I think there is an opportunity to consolidate these types of things to a shared file that got included in each plugin during the build process. There are tradeoffs to consider, for sure, but something we may want to consider.

Agreed. Let's open an issue to think about that further. While coming up with centralized logic for this is probably trivial, we'll likely need to spend some time considering different approaches for how we handle this central logic that still needs to be somehow copied and bundled into all of the standalone plugins.

joemcgill (approved):
Sounds good to me. 👍🏻

mukeshpanchal27 (approved):
Thanks @felixarntz! PR look good to me.

modules/images/dominant-color/hooks.php
```diff
@@ -244,3 +244,20 @@ function dominant_color_add_inline_style() {
...
+ * Displays the HTML generator tag for the Dominant Color Images plugin.
+ *
+ * See {@see 'wp_head'}.
+ *
+ * @since n.e.x.t
+ */
+function dominant_color_render_generator() {
+	if (
+		defined( 'DOMINANT_COLOR_IMAGES_VERSION' ) &&
+		! str_starts_with( DOMINANT_COLOR_IMAGES_VERSION, 'Performance Lab ' )
```
10upsimon: @felixarntz checking if this is supposed to be a false conditional check? I see further on in `load.php` you _do_ define`DOMINANT_COLOR_IMAGES_VERSION ` starting with `'Performance Lab '`, so I'm struggling to see why there is a false check against this. I feel I ma be missing some context here though.

10upsimon: It seems to be that it will always be prefixed with `Performance Lab `.

modules/images/dominant-color/readme.txt
```diff
@@ -0,0 +1,52 @@
...
+Contributors:      wordpressdotorg
+Requires at least: 6.1
+Tested up to:      6.2
+Requires PHP:      5.6
+Stable tag:        1.0.0
+License:           GPLv2 or later
+License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+Tags:              performance, images, dominant color
+
+Adds support to store dominant color for an image and create a placeholder background with that color.
```
10upsimon:
```suggestion
Adds support to store the dominant color of an image and create a placeholder background of that color.
```

modules/images/dominant-color/readme.txt
```diff
@@ -0,0 +1,52 @@
...
+
+== Description ==
+
+This plugin determines and stores the dominant color for media uploads within WordPress and then uses it to create a placeholder background with that color in the frontend while the image is loading.
```
10upsimon:
```suggestion
This plugin determines and stores the dominant color for images uploaded via the media library within WordPress and then uses it to create a placeholder background of that color in the frontend, visible until the image is loaded.
```

Nit pick comment, mostly just to have things read a tad smoother. I also suggested that we explicitly say _images_ as opposed to _media_ uploaded, as this does not apply to other media such as PDF etc, correct?

modules/images/fetchpriority/readme.txt
```diff
@@ -0,0 +1,52 @@
...
+Stable tag:        1.0.0
+License:           GPLv2 or later
+License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+Tags:              performance, images, fetchpriority
+
+Adds a fetchpriority hint for the primary content image on the page to load faster.
+
+== Description ==
+
+This plugin determines adds the `fetchpriority="high"` attribute to the image that is most likely the LCP image for the current response, improving LCP performance by telling the browser to prioritize that image. The LCP image detection directly relies on the existing WordPress core heuristics that determine whether to not lazy-load an image. The only difference is that, while multiple images may not be lazy-loaded, only a single image will be annotated with `fetchpriority="high"`.
```
10upsimon:
```suggestion
This plugin adds the `fetchpriority="high"` attribute to the image that is most likely the LCP image for the current response, improving LCP performance by telling the browser to prioritize that image. The LCP image detection directly relies on the existing WordPress core heuristics that determine whether to not lazy-load an image. The only difference is that, while multiple images may not be lazy-loaded, only a single image will be annotated with `fetchpriority="high"`.
```

modules/images/dominant-color/hooks.php
mukeshpanchal27: @10upsimon We set this meta tag for standalone plugin. When workflow run `npm run build-plugins` [script will changes](https://github.com/WordPress/performance/blob/trunk/bin/plugin/commands/build-plugins.js#L89-L94) the constant value `define( 'DOMINANT_COLOR_IMAGES_VERSION', 'Performance Lab ' . PERFLAB_VERSION );` to `define( 'DOMINANT_COLOR_IMAGES_VERSION', '1.0.0' );` which we define in `plugins.json` file.

10upsimon: Thank you @mukeshpanchal27 this now makes total sense. Resolving this comment thread.

modules/images/dominant-color/readme.txt
felixarntz: Thanks @10upsimon! I'll update this throughout as this is also the module description that should therefore be updated the same way.

felixarntz: Covered by #708.

modules/images/dominant-color/readme.txt
felixarntz: Thanks, updating this now. I'll include the "newly uploaded" bit here too, similar to how we are changing the module description, to clarify that it only applies to those images.

felixarntz:
@mukeshpanchal27 @10upsimon Since there was a merge conflict and since the two changes are so closely related, I've updated this PR to include the changes in the branch from #708 as well.

You can ignore those changes in here for now - let's complete and merge #708 first, and after that we can review the changes _actually_ coming from this PR here. It should only be merged after the 2.2.0 release anyway.

mukeshpanchal27 (approved):

PR #705, opened 2023-04-06
Rename `Dominant Color` module to `Dominant Color images`

felixarntz:
## Summary

Related to #640, however can be merged independently.

## Relevant technical choices

* Per https://github.com/WordPress/performance/issues/638#issuecomment-1480343450, the `Dominant Color` module should be renamed to `Dominant Color Images`, since the `dominant-color` slug is already taken by another plugin in the WordPress plugin repository. It was decided to go with that name in the linked discussion.
* While we are only going to publish the standalone plugin later, we can already rename the module independently of that.
* I opened this separate PR to do just that, also to make it more visible, as this way it'll independently appear as an entry in the PL plugin's changelog.
* Note that the name is not yet translated in the `readme.txt` and the `module-i18n.php` files, which is not necessary as that will automatically happen as part of the release preparation scripts (see https://make.wordpress.org/performance/handbook/performance-lab/releasing-the-plugin/#preparing-the-release).

## Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -49,7 +49,7 @@
 /tests/modules/js-and-css/audit-enqueued-assets                             @manuelRod
 /tests/testdata/modules/js-and-css/audit-enqueued-assets                    @manuelRod
 
-# Module: Dominant Color
+# Module: Dominant Color Images
 /modules/images/dominant-color                                              @pbearne @spacedmonkey
 /tests/modules/images/dominant-color                                        @pbearne @spacedmonkey
 /tests/testdata/modules/images/dominant-color                               @pbearne @spacedmonkey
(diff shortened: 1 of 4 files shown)


joemcgill (approved):
This is only changing inline doc references, from what I can tell, but seems fine to me, assuming that is the intent.

felixarntz:
@joemcgill Yes, the only relevant change outside of inline docs is the module name in the module header - which is still in an inline doc, but will be displayed to users, that is the most relevant change in this PR.

mukeshpanchal27 (approved):
Thanks @felixarntz, The changes look good to me.

Do we need to update module slug `images/dominant-color` to `images/dominant-color-images` as we already implemented [legacy_module_slugs feature](https://github.com/WordPress/performance/blob/trunk/load.php#L123) in plugin

felixarntz:
@mukeshpanchal27 Thanks, that's a great idea! I'll implement a follow up PR.

PR #710, opened 2023-04-13
Use dynamic version from `plugins.json` for manual workflow

### LINKED ISSUE #703: Improve the UI for manual plugin releases
## Feature Description

In https://github.com/WordPress/performance/pull/686 we added a new GitHub workflow for deploying modules as standalone plugins (see: https://github.com/WordPress/performance/issues/638). This includes a `workflow_dispatch` trigger to do manual releases of a standalone plugin. The current manual workflow supports two text inputs, one for the plugin slug and one for the version number (see screenshot below)

image

While this works, it creates an opportunity for human error if someone mistypes the slug or the version. I would suggest two changes:

1. Replace the text input for the slug with a dropdown to select the specific modules we support. This can use the `choice` input type for the workflow ([see docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs)).
2. Remove the input for setting the version number and instead use the version number from the `plugins.json` file. This means we would need to make sure that file is updated with the correct version number before a manual plugin release is completed, but would be more consistent, as the version number will match the version included in the built plugin.

mukeshpanchal27:
## Summary
This PR mainly improves the manual workflow for standalone plugin releases.

- Replaced the text input for the slug with a dropdown to select the specific modules so we don't get any human error if someone mistypes the slug.
- Remove version input
- How does the version work for automated and manual workflows?
   - **Automated workflow:** It will read the `plugins.json` file and based on json config it setup the standalone plugin.
   - **Manual workflow:** Based on the input slug the script will find the version for it, If the selected plugin slug is not present in `plugins.json` then trigger a failure and exit the workflow.

Also use test SVN credential so we can't release the plugin when doing testing.

I already open PR https://github.com/10up/action-wordpress-plugin-deploy/pull/122 that support the debug mode in deploy process. Once it will approve we can use `DRY_RUN` ENV variable for testing.

Fixes #703

## Checklist


.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -6,12 +6,17 @@ on:
...
+        default: 'webp-uploads'
+        options:
+        - audit-autoloaded-options
+        - audit-enqueued-assets
+        - dominant-color-images
+        - fetchpriority
+        - sqlite
+        - webp-support
+        - webp-uploads
+        description: 'Select a slug of the plugin to deploy'
```
felixarntz: The slugs in here are not all correct:
* Some of them, like `audit-autoloaded-options`, will not be published as a standalone plugin.
* Others, like `sqlite`, are using the wrong slug, as we need to use the plugin slug, not the module slug.

Also, it's a bit odd to have a default, I don't think any of the plugins should be provided as a default, as they are all equal.

I think having a hard-coded list here in general is not desirable. If it is possible to get the `options` from `plugins.json` here, we should do that. Otherwise I think it's probably best to keep this a regular text field like it was before.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -21,16 +26,61 @@ jobs:
...
+
+              if ( pluginVersion && pluginSlug && ( "${{ inputs.slug }}" == pluginSlug ) ) {
+                return pluginVersion;
+              }
+            }
+
+            console.log( `The "${{ inputs.slug }}" module slug is missing in the file "${ pluginsFile }".` );
+
+            // Return with exit code 1 to trigger a failure in the deploy standalone workflow pipeline.
+            process.exit( 1 );
```
felixarntz: This is a lot of manual JS code within the GitHub workflow.

It would be cleaner to implement this code in a new command (e.g. `get-plugin-version`) similar to our other commands like `build-plugins` and `test-plugins`, and then call that command from here.

You could implement the new command with a single `-s` / `--slug` parameter and then put this logic there so that it prints the version number so that it can be used here.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -53,8 +103,10 @@ jobs:
       - name: Deploy Standalone Plugin - ${{ matrix.slug }}
         uses: 10up/action-wordpress-plugin-deploy@stable
         env:
-          SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
-          SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          #SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
+          #SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          SVN_PASSWORD: SVN_PASSWORD
+          SVN_USERNAME: SVN_USERNAME
```
felixarntz: Maybe add a `TODO` here just to change this back prior to merge.

felixarntz (changes requested):
@mukeshpanchal27 Logic-wise this looks great. I have two comments for things to improve though, particularly the hard-coded options are not desirable as they are duplicated from what should be in `plugins.json` instead.

.github/workflows/deploy-standalone-plugins.yml
costdev: Thanks for drawing my attention to this @mukeshpanchal27!

I don't believe that `inputs.{name}.options` supports dynamic assignment from a file, and even then we'd need to go further to only get the `slug` values.

However, we could add another workflow that modifies this one.

1. Run on every `push` to `trunk` that touches `plugins.json`.
2. Remove existing options from the deployment workflow file. (i.e. the one in this PR)
3. Fetch, format and add the updated slugs to the deployment workflow file.
4. Commit to `trunk`.
5. Push to remote.

Something like this (Tested locally via ACT except for Commit and Push to remote):
```yaml
on:
  push:
    branches: trunk
    paths:
      - 'plugins.json'

jobs:
  update_slugs:
    name: Update slugs for deployment options
    runs-on: ubuntu-latest
    steps:
      - name: Checkout repository
        uses: actions/checkout@v3

      - name: Install jq
        run: |
          apt update -y
          apt install jq -y

      - name: Remove existing options
        run: sed -in '/options/, /description/{//!d;};' ./.github/workflows/deploy-standalone-plugins.yml

      - name: Fetch, format and add the updated slugs to the deployment workflow file
        run: |
          SLUGS=$(jq '"          - " + .[].slug' ./plugins.json | sed 's/"//g' )
          awk -v OPTIONS="options:\n$SLUGS" '{ sub( "options:", OPTIONS);print;}' ./.github/workflows/deploy-standalone-plugins.yml > tmp
          mv tmp ./.github/workflows/deploy-standalone-plugins.yml
          cat ./.github/workflows/deploy-standalone-plugins.yml

      - name: Commit to trunk
        run: |
          git add .
          git commit -m "Update standalone plugin slugs for deployment options."

      - name: Push to remote
        run: |
          git fetch origin trunk
          git push origin HEAD:trunk
  
```

Before:
```yaml
name: Manual workflow

on:
  workflow_dispatch:
    inputs:
      slug:
        type: choice
        options: 
          - test-plugin1
          - test-plugin3
        description: 'Select a slug of the plugin to deploy'

jobs:
  test:
    name: Run Test
    runs-on: ubuntu-latest
    steps:
      - name: A Test
        id: a-test
        run: echo "Howdy, admin!"
```

After:
```yaml
name: Manual workflow

on:
   workflow_dispatch:
     inputs:
       slug:
         type: choice
         options:
           - dominant-color-images
           - fetchpriority
           - webp-uploads 
         description: 'Select a slug of the plugin to deploy'

jobs:
  test:
    name: Run Test
    runs-on: ubuntu-latest
    steps:
      - name: A Test
        id: a-test
        run: echo "Howdy, admin!"
```

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: Updated in https://github.com/WordPress/performance/pull/710/commits/6a6af58297d146aaa4d4fcfabb4ca6f1e1807713

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: Thanks @felixarntz for the feedback. I have added new `npm run get-plugin-version` command to get the plugin version from the plugins.json file with single argument.

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: Thanks both for the feedback. I don't think updating the workflow file is good idea 🤔 so i go with input filed option as @felixarntz suggested because GH did no allow dynamic value for inputs.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -21,12 +17,29 @@ jobs:
...
+        run: |
+          echo "version=$(node ./bin/plugin/cli.js get-plugin-version --slug=${{ inputs.slug }})" >> $GITHUB_OUTPUT
       - name: Set matrix
         id: set-matrix
         run: |
           if ${{ github.event_name == 'workflow_dispatch' }}; then
-            # Set the manual input values in JSON format for use in the matrix.
-            echo "matrix={\"include\":[{\"slug\":\"${{ inputs.slug }}\",\"version\":\"${{ inputs.version }}\"}]}" >> $GITHUB_OUTPUT
+            if ${{ steps.get-version.outputs.version == '' }}; then
+              echo "The ${{ inputs.slug }} module slug is missing in the file plugins.json."
```
felixarntz: Should we validate that the string is a valid version number?

That may be needed if e.g. something was wrong with the CLI script and it resulted in incorrect output. I would assume the above step would already fail and terminate with an error in that case? But if not, or maybe even just to be safe it may be good to add a regex to validate here.

For example, this regex would work: `^(\*|\d+(\.\d+){0,2}(-[a-zA-Z0-9\.]+)?)$` (covers stable versions as well as beta and RC naming conventions)

felixarntz (approved):
@mukeshpanchal27 Thanks, this looks great now!

One additional comment for consideration, though not necessarily a blocker.

package.json
```diff
@@ -22,6 +22,7 @@
     "build-plugins": "./bin/plugin/cli.js build-plugins",
     "test-plugins": "./bin/plugin/cli.js test-plugins",
     "test-plugins-multisite": "./bin/plugin/cli.js test-plugins --sitetype=multi",
+    "get-plugin-version": "./bin/plugin/cli.js get-plugin-version",
```
joemcgill: This script is fine to add for testing purposes, but is not necessary since you're not actually using it during the workflow and are calling the CLI script directly. Note that if you wanted to test this output, you need to add an extra `--` before the options that you want to pass to the CLI, like:

```
npm run get-plugin-version -- --slug="webp-uploads"
```

joemcgill (dismissed):
This is working for me. 👍🏻

package.json
mukeshpanchal27: Thanks @joemcgill for the feedback. If i use above mention command in workflow then it also return other details which we don't needed. 

```
> performance@ get-plugin-version /Users/abc/performance
> ./bin/plugin/cli.js get-plugin-version "--slug=webp-uploads"

1.0.0
```

joemcgill: Yeah, the way you're doing it now is fine. I just noticed that the shortcut to the script in the `package.json` file here isn't being used directly.

bin/plugin/commands/get-plugin-version.js
```diff
@@ -0,0 +1,79 @@
...
+		pluginsFileContent = fs.readFileSync( pluginsFile, 'utf-8' );
+	} catch ( e ) {
+		throw Error( `Error reading file at "${ pluginsFile }": ${ e }` );
+	}
+
+	// Validate that the plugins JSON file contains content before proceeding.
+	if (
+		'' === pluginsFileContent ||
+		! pluginsFileContent
+	) {
```
costdev:
```suggestion
	if ( ! pluginsFileContent ) {
```
`''` is a falsy value, so `! pluginsFileContent` will catch this.

felixarntz:
@mukeshpanchal27 Note that the `DRY_RUN` PR has now been merged upstream. So it would be great to incorporate that here.

bin/plugin/commands/get-plugin-version.js
mukeshpanchal27: Thanks @costdev for the feedback.

package.json
mukeshpanchal27: Thanks @joemcgill. In https://github.com/WordPress/performance/pull/710/commits/092a37fabade1740c135ee8c8902fdc81fc82da2 remove command from `package.json`

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: Thanks @felixarntz for the feedback. In https://github.com/WordPress/performance/pull/710/commits/4995c95b38eeacb3104e2a15e46afa72f8bdf4d5 i have address it.

mukeshpanchal27:
> @mukeshpanchal27 Note that the `DRY_RUN` PR has now been merged upstream. So it would be great to incorporate that here.

@felixarntz The `DRY_RUN` PR has been merged into `develop` branch, not the `stable` branch, so we can't use it. If we have to use `DRY_RUN` then we have to use the `develop` branch, or we will use it once it is merged into the `stable` branch. cc. @joemcgill

felixarntz:
Thanks @mukeshpanchal27, in that case let's do that separately. Could you please open a tracking issue for us to set up `DRY_RUN` in the future once it becomes available in a stable version of the action?

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -53,8 +67,12 @@ jobs:
...
         uses: 10up/action-wordpress-plugin-deploy@stable
         env:
-          SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
-          SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          # TODO Once the workflow is tested, we will remove the comment and use the secret SVN access.
+          #SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
+          #SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          # TODO Once the workflow is tested, we will remove this test credential.
+          SVN_PASSWORD: SVN_PASSWORD
+          SVN_USERNAME: SVN_USERNAME
```
felixarntz: Before merging, please make sure to resolve these comments and bring back the usage of the `secrets`.

felixarntz (approved):
@mukeshpanchal27 Thanks, LGTM!

Just a reminder to update the below prior to merging.

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27:
```suggestion
          SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
          SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
```

mukeshpanchal27: Updated in https://github.com/WordPress/performance/pull/710/commits/52c6eb1084b33da32d8ff2fdd720132e221b317c

mukeshpanchal27:
> Thanks @mukeshpanchal27, in that case let's do that separately. Could you please open a tracking issue for us to set up `DRY_RUN` in the future once it becomes available in a stable version of the action?

@felixarntz, should we consider opening a separate issue to set up the `DRY_RUN` option? By doing so, the changes made will not be uploaded to SVN. This approach could be useful for testing the process before committing any actual changes. What are your thoughts on this?

felixarntz:
@mukeshpanchal27 Yes that's what I meant. We could somehow configure it I think so that for example in pull requests it is always set to have `DRY_RUN` enabled.

joemcgill (approved):
The latest updates look right to me. Thanks @mukeshpanchal27!

PR #701, opened 2023-04-05
Prepare the deployment workflow for standalone plugins

joemcgill:
## Summary
This swaps out the temporary bash file we were using for testing with the official `10up/action-wordpress-plugin-deploy` GitHub action and uses the `SVN_PASSWORD` and `SVN_USERNAME`  secrets that are saved to the repository. It also improves the names of the workflow jobs, for better readability.

Fixes #700 

## Checklist

- [x] PR has either `[Focus]` or `Infrastructure` label.
- [x] PR has a `[Type]` label.
- [x] PR has a milestone or the `no milestone` label.

PR #711, opened 2023-04-17
Remove `deploy.sh` file

mukeshpanchal27:
## Summary
In #701, we prepared the deployment workflow for standalone plugins by using the actual 10up deploy action instead of the custom `deploy.sh` copy, so now we should remove that file so it will not be uploaded to SVN.

## Checklist


felixarntz (approved):
Good catch @mukeshpanchal27!

10upsimon (approved):

spacedmonkey (approved):

PR #716, opened 2023-04-20
Add dominant color styling before any existing inline style attributes

### LINKED ISSUE #658: Check for pre existing inline styling rules before inserting plugin related styling
## Bug Description

As raised by [one user in the support forums](https://wordpress.org/support/topic/missing-a-semicolon-between-the-properties/), in cases whereby third party plugins or WordPress core have existing inline styling added to elements of their site, without closing these rules (via a `;` suffix), the Performance Lab plugin will insert it's own styling (ie. dominant color styling: `--dominant-color: #736e5e;`) merged in place with existing rules. This results in the dominant color styling not applying (along with the third party plugins inline styling). 

While this isn't specifically a Performance Lab issue, to improve the likelihood of this happening, consider checking for pre-existing inline styling rules when inserting the dominant color and any other inline styling. If there is inline styling added via other means, check whether they are closed. If not, include a prefix `;` to the dominant color styling. 

## Steps to reproduce

1. Add a [cover block](https://wordpress.org/documentation/article/cover-block/) to a WordPress post
2. To ensure there is inline styling added, change the "Focal Point Picker", [as per this recording](https://recordit.co/OPfbyKao4F). In this case, the inline styling will appear similar to: style="object-position:22% 59%"
3. Check your sites source code, notice the inline styling isn't closed (doesn't have a semi colon)
4. Enable the "Dominants Color module" within the Performance Lab configurations
5. Notice that the styling added via the dominant color is merged with the Cover block styling, rendering both redundant 

## Screenshots

_**Below**: Example of inline styling with the Cover block - notice there is no closing semi colon:_

 
<kbd> image

_**Below**: Example of inline styling with the Cover block, with another plugin that also adds it's own inline styling to the same element - there is no distinction between the rules, they're merged._

<kbd> image

## Additional Context

 - We [do add a closing ; ](https://github.com/WordPress/performance/blob/134221b349f1277c27ca68d6584ff010a58d95dd/modules/images/dominant-color/load.php#L13)to the inline styling added via the Dominant Color option
 - Tested on with WordPress 6.1.1 and newer Beta versions
 - Separate GitHub issue raised on this over on the Gutenberg repo: https://github.com/WordPress/gutenberg/issues/48258

mukeshpanchal27:
## Summary

To replicate the issue follow below steps:
- Go to Admin Panel then add new post
- Upload featured image for the post
- Add Cover block in which use "Use featured image" for image
- Select "FOCAL POINT PICKER"
- Check frontend.

**Before the patch:**
Screenshot 2023-04-20 at 2 37 17 PM

**After the patch:**
Screenshot 2023-04-20 at 2 36 49 PM

Initially i also think about to use `str_ends_with` but it will create problem when someone add style with space i.e.`display: block; ` then it give false result.

Fixes #658

## Checklist

### DIFF
--- a/modules/images/dominant-color-images/hooks.php
+++ b/modules/images/dominant-color-images/hooks.php
@@ -62,7 +62,9 @@ function dominant_color_update_attachment_image_attributes( $attr, $attachment )
 		if ( empty( $attr['style'] ) ) {
 			$attr['style'] = '';
 		}
-		$attr['style'] .= '--dominant-color: #' . esc_attr( $image_meta['dominant_color'] ) . ';';
+		$style_attribute = $attr['style'];
+		$attr['style']   = '';
+		$attr['style']  .= '--dominant-color: #' . esc_attr( $image_meta['dominant_color'] ) . ';' . $style_attribute;
 	}
 
 	return $attr;
--- a/modules/images/dominant-color-images/hooks.php
+++ b/modules/images/dominant-color-images/hooks.php
@@ -59,12 +59,8 @@ function dominant_color_update_attachment_image_attributes( $attr, $attachment )
 
 	if ( ! empty( $image_meta['dominant_color'] ) ) {
 		$attr['data-dominant-color'] = esc_attr( $image_meta['dominant_color'] );
-		if ( empty( $attr['style'] ) ) {
-			$attr['style'] = '';
-		}
-		$style_attribute = $attr['style'];
-		$attr['style']   = '';
-		$attr['style']  .= '--dominant-color: #' . esc_attr( $image_meta['dominant_color'] ) . ';' . $style_attribute;
+		$style_attribute = empty( $attr['style'] ) ? '' : $attr['style'];
+		$attr['style']   = '--dominant-color: #' . esc_attr( $image_meta['dominant_color'] ) . ';' . $style_attribute;
 	}
 
 	return $attr;
--- a/modules/images/dominant-color-images/hooks.php
+++ b/modules/images/dominant-color-images/hooks.php
@@ -59,8 +59,8 @@ function dominant_color_update_attachment_image_attributes( $attr, $attachment )
 
 	if ( ! empty( $image_meta['dominant_color'] ) ) {
 		$attr['data-dominant-color'] = esc_attr( $image_meta['dominant_color'] );
-		$style_attribute = empty( $attr['style'] ) ? '' : $attr['style'];
-		$attr['style']   = '--dominant-color: #' . esc_attr( $image_meta['dominant_color'] ) . ';' . $style_attribute;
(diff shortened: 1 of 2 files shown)


modules/images/dominant-color-images/hooks.php
```diff
@@ -62,7 +62,9 @@ function dominant_color_update_attachment_image_attributes( $attr, $attachment )
 		if ( empty( $attr['style'] ) ) {
 			$attr['style'] = '';
 		}
```
spacedmonkey: Maybe we should get rid of this..

modules/images/dominant-color-images/hooks.php
```diff
@@ -62,7 +62,9 @@ function dominant_color_update_attachment_image_attributes( $attr, $attachment )
...
-		$attr['style'] .= '--dominant-color: #' . esc_attr( $image_meta['dominant_color'] ) . ';';
+		$style_attribute = $attr['style'];
```
spacedmonkey:
```suggestion
		$style_attribute = empty( $attr['style'] ) ? '' : $attr['style'];
```

modules/images/dominant-color-images/hooks.php
spacedmonkey:
```suggestion
```

modules/images/dominant-color-images/hooks.php
```diff
@@ -62,7 +62,9 @@ function dominant_color_update_attachment_image_attributes( $attr, $attachment )
...
+		$attr['style']   = '';
+		$attr['style']  .= '--dominant-color: #' . esc_attr( $image_meta['dominant_color'] ) . ';' . $style_attribute;
```
spacedmonkey:
```suggestion
		$attr['style']   = '--dominant-color: #' . esc_attr( $image_meta['dominant_color'] ) . ';' . $style_attribute;
```

Would this not work?

spacedmonkey (changes requested):
Some simple feedback, but otherwise, looking good.

felixarntz (commented):
@mukeshpanchal27 Almost LGTM, except for what @spacedmonkey already pointed out.

felixarntz:
@mukeshpanchal27 Since this is a bug fix that has product side implications, we should add a milestone to the issue. `no milestone` should be reserved only for PRs which don't affect the end user-facing product (e.g. non-functional changes).

joemcgill:
Nothing additional from me. Thanks for adding a new unit test for this as well!

mukeshpanchal27:
Thanks to all for the feedback. PR is ready for the second round of review.

pbearne (approved):

felixarntz (approved):
Thanks @mukeshpanchal27!

spacedmonkey (approved):

PR #729, opened 2023-05-15
Use static closures for minor performance improvement whenever instance access is not needed

### LINKED ISSUE #728: Static closures should be employed whenever `$this` is not used
## Bug Description

There closures being used in the codebase which do not use `$this`. Granted, in most cases clases aren't even involved, but in those that are, using `static` is better for memory usage since the `$this` reference can be garbage collected. See [SO answer](https://stackoverflow.com/a/64481894/93579).

I suggest that we do the following:

1. Add `static` to all closures that don't use `$this`.
2. Incorporate the [`SlevomatCodingStandard.Functions.StaticClosure`](https://github.com/slevomat/coding-standard) sniff to catch this automatically.

westonruter:
## Summary

Fixes #728

## Relevant technical choices

* Added the `SlevomatCodingStandard.Functions.StaticClosure` sniff to the PHPCS ruleset.
* Ran PHPCBF to add `static` where relevant.

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7",
     "phpcompatibility/php-compatibility": "^9.3",
     "phpunit/phpunit": "^4|^5|^6|^7|^8|^9",
+    "slevomat/coding-standard": "^8.0",
     "squizlabs/php_codesniffer": "^3.5",
     "wp-coding-standards/wpcs": "^2.2",
     "wp-phpunit/wp-phpunit": "^5.8",
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -72,6 +72,11 @@
 	<rule ref="PHPCompatibility.FunctionDeclarations.NewReturnTypeDeclarations.boolFound">
 		<exclude-pattern>tests/utils/*</exclude-pattern>
 	</rule>
+
+	<rule ref="SlevomatCodingStandard.Functions.StaticClosure">
+		<exclude-pattern>tests/*</exclude-pattern>
+	</rule>
+
 	<!--
 	Prevent errors caused by WordPress Coding Standards not supporting PHP 8.0+.
 	See https://github.com/WordPress/WordPress-Coding-Standards/issues/2035
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -73,9 +73,7 @@
 		<exclude-pattern>tests/utils/*</exclude-pattern>
 	</rule>
 
-	<rule ref="SlevomatCodingStandard.Functions.StaticClosure">
-		<exclude-pattern>tests/*</exclude-pattern>
-	</rule>
+	<rule ref="SlevomatCodingStandard.Functions.StaticClosure" />
 
 	<!--
 	Prevent errors caused by WordPress Coding Standards not supporting PHP 8.0+.
--- a/tests/admin/load-tests.php
+++ b/tests/admin/load-tests.php
@@ -81,7 +81,7 @@ class Admin_Load_Tests extends WP_UnitTestCase {
 		remove_all_filters( 'plugin_action_links_' . plugin_basename( PERFLAB_MAIN_FILE ) );
 
 		// Does not register the page if the perflab_active_modules filter is used.
-		add_filter( 'perflab_active_modules', function() {} );
+		add_filter( 'perflab_active_modules', static function() {} );
 		$hook_suffix = perflab_add_modules_page();
 		$this->assertFalse( $hook_suffix );
 		$this->assertFalse( isset( $_wp_submenu_nopriv['options-general.php'][ PERFLAB_MODULES_SCREEN ] ) );
--- a/tests/admin/load-tests.php
+++ b/tests/admin/load-tests.php
@@ -81,7 +81,7 @@ class Admin_Load_Tests extends WP_UnitTestCase {
 		remove_all_filters( 'plugin_action_links_' . plugin_basename( PERFLAB_MAIN_FILE ) );
 
 		// Does not register the page if the perflab_active_modules filter is used.
-		add_filter( 'perflab_active_modules', static function() {} );
+		add_filter( 'perflab_active_modules', '__return_null' );
 		$hook_suffix = perflab_add_modules_page();
 		$this->assertFalse( $hook_suffix );
 		$this->assertFalse( isset( $_wp_submenu_nopriv['options-general.php'][ PERFLAB_MODULES_SCREEN ] ) );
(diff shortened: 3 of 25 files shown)


phpcs.xml.dist
```diff
@@ -72,6 +72,11 @@
 	<rule ref="PHPCompatibility.FunctionDeclarations.NewReturnTypeDeclarations.boolFound">
 		<exclude-pattern>tests/utils/*</exclude-pattern>
 	</rule>
+
+	<rule ref="SlevomatCodingStandard.Functions.StaticClosure">
+		<exclude-pattern>tests/*</exclude-pattern>
```
felixarntz: Is there a reason to exclude tests? Not a big deal either way, but this seems like a rule easy enough to follow that we could also apply it to tests. Or did you exclude because it would result in a ton more to fix in this first PR?

felixarntz (approved):
@westonruter LGTM, just one non-blocking question.

phpcs.xml.dist
westonruter: Yeah, the latter. There are a lot more instances in tests. Happy to apply the changes there too if you like.

felixarntz:
@westonruter I've updated the label since this is more of an enhancement than a bug (nothing broken without this), plus changed the title slightly since that is used to auto-generate the eventual changelog for the release that this will be part of.

phpcs.xml.dist
felixarntz: Either is fine, maybe we can do that in a follow up PR instead.

westonruter: Went ahead and did it: f8b62ef

tests/admin/load-tests.php
```diff
@@ -81,7 +81,7 @@ public function test_perflab_add_modules_page() {
 		remove_all_filters( 'plugin_action_links_' . plugin_basename( PERFLAB_MAIN_FILE ) );
 
 		// Does not register the page if the perflab_active_modules filter is used.
-		add_filter( 'perflab_active_modules', function() {} );
+		add_filter( 'perflab_active_modules', static function() {} );
```
aristath: Looks like we can simplify this a bit...
```suggestion
		add_filter( 'perflab_active_modules', '__return_null' );
```

aristath (approved):
LGTM 👍 
I had a small nitpick and added a comment, but definitely not a blocker

composer.json
```diff
@@ -15,6 +15,7 @@
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7",
     "phpcompatibility/php-compatibility": "^9.3",
     "phpunit/phpunit": "^4|^5|^6|^7|^8|^9",
+    "slevomat/coding-standard": "^8.0",
```
spacedmonkey: This coding standard requires php 7.2. Will would make this plugin break in 7.2- no. 

https://github.com/slevomat/coding-standard/blob/f69e2524e8770efb9b3e5ac4a0ebc0d54eb446d7/composer.json#L19

swissspidy: It would not break the plugin, because the composer autoloader is only used in tests.

It would just break tests running on PHP 5.6. However, right now tests are not running on this version on CI, so... 🤷 

Edit: looks like there is #399 for updating tests.

felixarntz (changes requested):
Marking this as needing changes due to https://github.com/WordPress/performance/pull/730#discussion_r1195057672 which applies here too.

composer.json
westonruter: <details><summary>Please continue discussion at <a href="https://github.com/WordPress/performance/pull/730#discussion_r1195408400">#730 (comment)</a></summary>

As I understand, there are three possible ways to about this:

1. Have a separate composer file for PHP 7.2+
2. In GHA workflow, just-in-time `composer require` this package before doing `composer install` before running PHPCS
3. In GHA workflow, just-in-time `composer remove` this package before doing `composer install` prior to running PHPUnit in PHP<7.2.

I think the third option is ideal, since all normal development environments would be running PHP 7.2+. It's only the exception case, specifically on GHA, where PHP 5.6 is needed. This would allow a developer to have all packages installed normally. 

The first option doesn't seem ideal since there would be a separate composer file to maintain, and Dependabot wouldn't automatically be aware of it.

</details>

westonruter: How about we address this issue as part of #399?

westonruter: We may not want this sniff after all. See https://github.com/WordPress/wordpress-develop/pull/4457#discussion_r1194463282

felixarntz: Noting that we could still make a separate decision from core here, there are several plugins that have stricter requirements than core to enforce better coding practices. I don't feel strongly about this one, but I feel like if we can find a way that we can integrate it that remains compatible with PHP 5.6, I'd be happy to include it.

spacedmonkey: I am not against using this rule set, but php 5.6 compat is the key here.

westonruter: Can the 5.6 compat be handled as part of #399? Without CI actually running anything in 5.6, it makes it a bit difficult to verify the compat is working, or at least it seems a bit premature.

tests/server-timing/perflab-server-timing-tests.php
```diff
@@ -17,7 +17,7 @@ class Perflab_Server_Timing_Tests extends WP_UnitTestCase {
 
 	public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
 		self::$dummy_args = array(
-			'measure_callback' => function() {},
+			'measure_callback' => static function() {},
```
mukeshpanchal27:
```suggestion
			'measure_callback' => '__return_null',
```

tests/server-timing/perflab-server-timing-tests.php
```diff
@@ -106,7 +106,7 @@ public function test_register_metric_requires_access_cap() {
 
 		$this->server_timing->register_metric(
 			'metric-without-access-cap',
-			array( 'measure_callback' => function() {} )
+			array( 'measure_callback' => static function() {} )
```
mukeshpanchal27:
```suggestion
			array( 'measure_callback' => '__return_null' )
```

mukeshpanchal27 (commented):
Thanks, @westonruter. I left non-blocking feedback other than looks good to me.

westonruter:
> Marking this as needing changes due to [#730 (comment)](https://github.com/WordPress/performance/pull/730#discussion_r1195057672) which applies here too.

@felixarntz Considering your https://github.com/WordPress/performance/pull/730#issuecomment-1569019104 I think this can move forward.

felixarntz:
@westonruter Can you refresh this PR please so we can hopefully merge it by tomorrow for the 2.4.0 release?

westonruter:
> @westonruter Can you refresh this PR please so we can hopefully merge it by tomorrow for the 2.4.0 release?

@felixarntz Done ✅

felixarntz (approved):
Thanks @westonruter!

PR #730, opened 2023-05-16
Enhance code quality by adding PHPStan and fixing level 0 issues

### LINKED ISSUE #727: Incorporate PHPStan for Static Analysis
## Feature Description

Issues with static analysis were reported in #622 and fixed in #645 without actually incorporating PHPStan. Is there a reason why this wasn't done? 

I suggest we incorporate [szepeviktor/phpstan-wordpress](https://github.com/szepeviktor/phpstan-wordpress) into the tooling alongside PHPCS. This will ensure we catch more static analysis issues. While the plugin currently passes level 0 well enough, there are much more issues to find in the [higher levels](https://phpstan.org/user-guide/rule-levels).

I'm happy to work on a PR if desired.

westonruter:
## Summary

Fixes #727

## Relevant technical choices

* Add a `phpstan` composer script. This name is chosen instead of `analyze` since also spelled `analyse`.
* Run PHPStan in the `php-lint` workflow after running PHPCS. This runs whenever a PHP-related file has changed.
* Also run PHPStan in lint-staged whenever a PHP file is modified.
* Add PHPStan config with `szepeviktor/phpstan-wordpress` and bleeding-edge, starting at level 0. Additional PRs will raise the level and fix more issues.
* Fix issues uncovered by level 0.

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,8 @@
     "wp-coding-standards/wpcs": "^2.2",
     "wp-phpunit/wp-phpunit": "^5.8",
     "yoast/phpunit-polyfills": "^1.0",
-    "phpstan/phpstan-phpunit": "^1.3"
+    "phpstan/phpstan-phpunit": "^1.3",
+    "phpstan/phpstan-deprecation-rules": "^1.1"
   },
   "require": {
     "composer/installers": "~1.0",
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,8 @@
   "config": {
     "allow-plugins": {
       "dealerdirect/phpcodesniffer-composer-installer": true,
-      "composer/installers": true
+      "composer/installers": true,
+      "phpstan/extension-installer": true
     }
   },
   "autoload-dev": {
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
     "szepeviktor/phpstan-wordpress": "^1.3.0",
     "wp-coding-standards/wpcs": "^2.2",
     "wp-phpunit/wp-phpunit": "^5.8",
-    "yoast/phpunit-polyfills": "^1.0"
+    "yoast/phpunit-polyfills": "^1.0",
+    "phpstan/phpstan-phpunit": "^1.3"
   },
   "require": {
     "composer/installers": "~1.0",
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
     "php": ">=5.6|^7|^8"
   },
   "scripts": {
+    "phpstan": "phpstan analyze --ansi",
     "format": "phpcbf --standard=phpcs.xml.dist --report-summary --report-source",
     "lint": "phpcs --standard=phpcs.xml.dist",
     "test": "phpunit -c phpunit.xml.dist --verbose",
--- a/composer.json
+++ b/composer.json
@@ -14,8 +14,11 @@
   "require-dev": {
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7",
     "phpcompatibility/php-compatibility": "^9.3",
+    "phpstan/extension-installer": "^1.3.0",
+    "phpstan/phpstan": "^1.10",
     "phpunit/phpunit": "^4|^5|^6|^7|^8|^9",
     "squizlabs/php_codesniffer": "^3.5",
+    "szepeviktor/phpstan-wordpress": "^1.3.0",
     "wp-coding-standards/wpcs": "^2.2",
     "wp-phpunit/wp-phpunit": "^5.8",
     "yoast/phpunit-polyfills": "^1.0"
--- a/tests/modules/images/webp-uploads/helper-tests.php
+++ b/tests/modules/images/webp-uploads/helper-tests.php
@@ -364,12 +364,7 @@ class WebP_Uploads_Helper_Tests extends ImagesTestCase {
 	 * @test
 	 */
 	public function it_should_return_default_transforms_when_filter_returns_non_array_type() {
-		add_filter(
-			'webp_uploads_upload_image_mime_transforms',
-			function () {
-				return null;
-			}
-		);
+		add_filter( 'webp_uploads_upload_image_mime_transforms', '__return_null' );
 
 		$default_transforms = array(
 			'image/jpeg' => array( 'image/webp' ),
--- a/tests/modules/images/webp-uploads/helper-tests.php
+++ b/tests/modules/images/webp-uploads/helper-tests.php
@@ -367,7 +367,7 @@ class WebP_Uploads_Helper_Tests extends ImagesTestCase {
 		add_filter(
 			'webp_uploads_upload_image_mime_transforms',
 			function () {
-				return;
+				return null;
 			}
 		);
 
--- a/.github/workflows/php-lint.yml
+++ b/.github/workflows/php-lint.yml
@@ -56,3 +56,5 @@ jobs:
         run: composer install --no-interaction --no-progress
       - name: PHP Lint
         run: composer lint
+      - name: PHPStan
+        run: composer phpstan
--- a/.github/workflows/php-lint.yml
+++ b/.github/workflows/php-lint.yml
@@ -22,6 +22,7 @@ on:
       - '.github/workflows/php-lint.yml'
       - '**.php'
       - 'phpcs.xml.dist'
+      - 'phpstan.neon.dist'
       - 'composer.json'
       - 'composer.lock'
     types:
--- a/.github/workflows/php-lint.yml
+++ b/.github/workflows/php-lint.yml
@@ -10,6 +10,7 @@ on:
       - '.github/workflows/php-lint.yml'
       - '**.php'
       - 'phpcs.xml.dist'
+      - 'phpstan.neon.dist'
       - 'composer.json'
       - 'composer.lock'
   pull_request:
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "800fb1ee62ef74aa3e6c6cd409939cff",
+    "content-hash": "f11a3661fdc88ae05a344d5e20725279",
     "packages": [
         {
             "name": "composer/installers",
@@ -531,6 +531,50 @@
             "time": "2022-02-21T01:04:05+00:00"
         },
         {
+            "name": "php-stubs/wordpress-stubs",
+            "version": "v6.2.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/php-stubs/wordpress-stubs.git",
+                "reference": "b73fe99eadf9fb56363619dac0343b6d19907dce"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/php-stubs/wordpress-stubs/zipball/b73fe99eadf9fb56363619dac0343b6d19907dce",
+                "reference": "b73fe99eadf9fb56363619dac0343b6d19907dce",
+                "shasum": ""
+            },
+            "require-dev": {
+                "nikic/php-parser": "< 4.12.0",
+                "php": "~7.3 || ~8.0",
+                "php-stubs/generator": "^0.8.3",
+                "phpdocumentor/reflection-docblock": "^5.3",
+                "phpstan/phpstan": "^1.9"
+            },
+            "suggest": {
+                "paragonie/sodium_compat": "Pure PHP implementation of libsodium",
+                "symfony/polyfill-php73": "Symfony polyfill backporting some PHP 7.3+ features to lower PHP versions",
+                "szepeviktor/phpstan-wordpress": "WordPress extensions for PHPStan"
+            },
+            "type": "library",
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "MIT"
+            ],
+            "description": "WordPress function and class declaration stubs for static analysis.",
+            "homepage": "https://github.com/php-stubs/wordpress-stubs",
+            "keywords": [
+                "PHPStan",
+                "static analysis",
+                "wordpress"
+            ],
+            "support": {
+                "issues": "https://github.com/php-stubs/wordpress-stubs/issues",
+                "source": "https://github.com/php-stubs/wordpress-stubs/tree/v6.2.0"
+            },
+            "time": "2023-03-31T09:48:52+00:00"
+        },
+        {
             "name": "phpcompatibility/php-compatibility",
             "version": "9.3.5",
             "source": {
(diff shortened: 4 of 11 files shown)


tests/modules/images/webp-uploads/helper-tests.php
```diff
@@ -367,7 +367,7 @@ public function it_should_return_default_transforms_when_filter_returns_non_arra
 		add_filter(
 			'webp_uploads_upload_image_mime_transforms',
 			function () {
-				return;
+				return null;
 			}
 		);
```
aristath: Can't we use `__return_null` here instead of adding this function?
```php
add_filter( 'webp_uploads_upload_image_mime_transforms', '__return_null' );
```

szepeviktor:
Please consider reaching level 4.
Under level 4 there could be critical bugs and bad design decisions.

spacedmonkey:
@westonruter Do this work in PHP 5.6? As I remember, phpstan requires PHP 7.2+

szepeviktor:
PHPStan does static analysis, so it does not run the code.
Runtime PHP version can be set! https://phpstan.org/config-reference#phpversion

spacedmonkey:
> PHPStan does static analysis, so it does not run the code. Runtime PHP version can be set! https://phpstan.org/config-reference#phpversion

I am talking about composer installing and requires.

composer.json
```diff
@@ -14,8 +14,11 @@
   "require-dev": {
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7",
     "phpcompatibility/php-compatibility": "^9.3",
+    "phpstan/extension-installer": "^1.3.0",
+    "phpstan/phpstan": "^1.10",
```
swissspidy, line 18: Technically not needed because it's a dependency of `szepeviktor/phpstan-wordpress` already

composer.json
```diff
@@ -14,8 +14,11 @@
...
     "phpunit/phpunit": "^4|^5|^6|^7|^8|^9",
     "squizlabs/php_codesniffer": "^3.5",
+    "szepeviktor/phpstan-wordpress": "^1.3.0",
```
swissspidy, line 21: Not that this package and PHPStan itself require PHP 7.2.

So when we start running tests against PHP 5.6 (see #399), `composer install` would break.

The plugin itself wouldn't break, as it doesn't use the autoloader.

Same discussion as in #729 really.

Check out https://github.com/10up/plugin-check/pull/166 for a workaround by moving these deps to a separate `composer.json` file.

swissspidy, line 21: I can also highly recommend https://packagist.org/packages/swissspidy/phpstan-no-private to catch calls of seemingly private functions.

Plus https://packagist.org/packages/phpstan/phpstan-phpunit and https://packagist.org/packages/php-stubs/wordpress-tests-stubs to run PHPStan against tests as well. Believe me, it really improves the quality of the tests!

https://packagist.org/packages/phpstan/phpstan-deprecation-rules is also nice to prevent usage of deprecated functions.

felixarntz: Great point @swissspidy, let's apply the same approach here as we want to support PHP 5.6 too for our test suite as we should preferably run the tests on that version too.

westonruter:
> Please consider reaching level 4. Under level 4 there could be critical bugs and bad design decisions.

@szepeviktor yes, that's the plan. Just for the initial PR level 0 is the first target.

composer.json
westonruter: Yes, I thought about that. But I thought this might get us PHPStan updates more quickly by being an explicit dependency. Otherwise we'd need to wait for the package to update, AFAIK.

composer.json
westonruter: As I understand, there are three possible ways to about this:

1. Have a separate composer file for PHP 7.2+. (Just seeing https://github.com/10up/plugin-check/pull/166 now.)
2. In GHA workflow, just-in-time `composer require` this package before doing `composer install` before running PHPCS
3. In GHA workflow, just-in-time `composer remove` this package before doing `composer install` prior to running PHPUnit in PHP<7.2.

I think the third option is ideal, since all normal development environments would be running PHP 7.2+. It's only the exception case, specifically on GHA, where PHP 5.6 is needed. This would allow a developer to have all packages installed normally. 

The first option doesn't seem ideal since there would be a separate composer file to maintain, and Dependabot wouldn't automatically be aware of it.

westonruter: How about we address back-compat with PHP 5.6 as part of #399?

swissspidy: Downsides of (2) and (3) are that it's a less reproducible setup for developers, and that it's cumbersome to do for other packages as well (think all the PHPStan extensions or the Slevomat CS package). I did something like that in the past for downgrading/upgrading PHPUnit in CI, but it's not more maintainable than (1).

I wish (1) wasn't needed, but when dealing with such version requirements it's a necessary evil.

> Dependabot wouldn't automatically be aware of it.

FWIW this is easily solved with another entry in the Dependabot config. Example:

```yml
  - package-ecosystem: "composer"
    directory: "/"
    schedule:
      interval: "daily"

  - package-ecosystem: "composer"
    directory: "/build-cs"
    schedule:
      interval: "daily"
```

westonruter: > I can also highly recommend https://packagist.org/packages/swissspidy/phpstan-no-private to catch calls of seemingly private functions.

I love this package, but I wonder if this is the best for Performance Lab considering that the code in question will end up in core and thus it is expected that many calls to internal/private functions will be made? As it stands right now...

<details><summary>There are 73 instances of this</summary>
<p>

```
 ------ ------------------------------------------------------- 
  Line   modules/database/sqlite/activate.php                   
 ------ ------------------------------------------------------- 
  71     Call to private/internal function wp_set_wpdb_vars().  
  111    Call to private/internal function wp_set_wpdb_vars().  
 ------ ------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------------ 
  Line   modules/database/sqlite/wp-includes/sqlite/class-perflab-sqlite-alter-query.php                 
 ------ ------------------------------------------------------------------------------------------------ 
  53     Call to private/internal method command_tokenizer() of class Perflab_SQLite_Alter_Query.        
  65     Call to private/internal method handle_single_command() of class Perflab_SQLite_Alter_Query.    
  69     Call to private/internal method handle_add_primary_key() of class Perflab_SQLite_Alter_Query.   
  73     Call to private/internal method handle_drop_primary_key() of class Perflab_SQLite_Alter_Query.  
  77     Call to private/internal method handle_modify_command() of class Perflab_SQLite_Alter_Query.    
  81     Call to private/internal method handle_change_command() of class Perflab_SQLite_Alter_Query.    
  85     Call to private/internal method handle_alter_command() of class Perflab_SQLite_Alter_Query.     
  261    Call to private/internal method convert_field_types() of class Perflab_SQLite_Alter_Query.      
  367    Call to private/internal method convert_field_types() of class Perflab_SQLite_Alter_Query.      
  421    Call to private/internal method convert_field_types() of class Perflab_SQLite_Alter_Query.      
  483    Call to private/internal method convert_field_types() of class Perflab_SQLite_Alter_Query.      
  500    Call to private/internal method convert_field_types() of class Perflab_SQLite_Alter_Query.      
  513    Call to private/internal method convert_field_types() of class Perflab_SQLite_Alter_Query.      
 ------ ------------------------------------------------------------------------------------------------ 

 ------ ----------------------------------------------------------------------------------------------- 
  Line   modules/database/sqlite/wp-includes/sqlite/class-perflab-sqlite-create-query.php               
 ------ ----------------------------------------------------------------------------------------------- 
  76     Call to private/internal method strip_backticks() of class Perflab_SQLite_Create_Query.        
  77     Call to private/internal method quote_illegal_field() of class Perflab_SQLite_Create_Query.    
  78     Call to private/internal method get_table_name() of class Perflab_SQLite_Create_Query.         
  79     Call to private/internal method rewrite_comments() of class Perflab_SQLite_Create_Query.       
  80     Call to private/internal method rewrite_field_types() of class Perflab_SQLite_Create_Query.    
  81     Call to private/internal method rewrite_character_set() of class Perflab_SQLite_Create_Query.  
  82     Call to private/internal method rewrite_engine_info() of class Perflab_SQLite_Create_Query.    
  83     Call to private/internal method rewrite_unsigned() of class Perflab_SQLite_Create_Query.       
  84     Call to private/internal method rewrite_autoincrement() of class Perflab_SQLite_Create_Query.  
  85     Call to private/internal method rewrite_primary_key() of class Perflab_SQLite_Create_Query.    
  86     Call to private/internal method rewrite_foreign_key() of class Perflab_SQLite_Create_Query.    
  87     Call to private/internal method rewrite_unique_key() of class Perflab_SQLite_Create_Query.     
  88     Call to private/internal method rewrite_enum() of class Perflab_SQLite_Create_Query.           
  89     Call to private/internal method rewrite_set() of class Perflab_SQLite_Create_Query.            
  90     Call to private/internal method rewrite_key() of class Perflab_SQLite_Create_Query.            
  91     Call to private/internal method add_if_not_exists() of class Perflab_SQLite_Create_Query.      
  93     Call to private/internal method post_process() of class Perflab_SQLite_Create_Query.           
 ------ ----------------------------------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------------------------------------- 
  Line   modules/database/sqlite/wp-includes/sqlite/class-perflab-sqlite-pdo-driver.php                      
 ------ ---------------------------------------------------------------------------------------------------- 
  81     Call to private/internal method parse_query() of class Perflab_SQLite_PDO_Driver.                   
  84     Call to private/internal method handle_truncate_query() of class Perflab_SQLite_PDO_Driver.         
  88     Call to private/internal method handle_alter_query() of class Perflab_SQLite_PDO_Driver.            
  92     Call to private/internal method handle_create_query() of class Perflab_SQLite_PDO_Driver.           
  97     Call to private/internal method handle_describe_query() of class Perflab_SQLite_PDO_Driver.         
  101    Call to private/internal method handle_show_query() of class Perflab_SQLite_PDO_Driver.             
  105    Call to private/internal method handle_show_columns_query() of class Perflab_SQLite_PDO_Driver.     
  109    Call to private/internal method handle_show_index() of class Perflab_SQLite_PDO_Driver.             
  115    Call to private/internal method rewrite_date_sub() of class Perflab_SQLite_PDO_Driver.              
  116    Call to private/internal method delete_index_hints() of class Perflab_SQLite_PDO_Driver.            
  117    Call to private/internal method rewrite_regexp() of class Perflab_SQLite_PDO_Driver.                
  119    Call to private/internal method fix_date_quoting() of class Perflab_SQLite_PDO_Driver.              
  120    Call to private/internal method rewrite_between() of class Perflab_SQLite_PDO_Driver.               
  121    Call to private/internal method handle_orderby_field() of class Perflab_SQLite_PDO_Driver.          
  126    Call to private/internal method execute_duplicate_key_update() of class Perflab_SQLite_PDO_Driver.  
  127    Call to private/internal method rewrite_insert_ignore() of class Perflab_SQLite_PDO_Driver.         
  128    Call to private/internal method rewrite_regexp() of class Perflab_SQLite_PDO_Driver.                
  129    Call to private/internal method fix_date_quoting() of class Perflab_SQLite_PDO_Driver.              
  134    Call to private/internal method rewrite_update_ignore() of class Perflab_SQLite_PDO_Driver.         
  136    Call to private/internal method rewrite_limit_usage() of class Perflab_SQLite_PDO_Driver.           
  137    Call to private/internal method rewrite_order_by_usage() of class Perflab_SQLite_PDO_Driver.        
  138    Call to private/internal method rewrite_regexp() of class Perflab_SQLite_PDO_Driver.                
  139    Call to private/internal method rewrite_between() of class Perflab_SQLite_PDO_Driver.               
  144    Call to private/internal method rewrite_limit_usage() of class Perflab_SQLite_PDO_Driver.           
  145    Call to private/internal method rewrite_order_by_usage() of class Perflab_SQLite_PDO_Driver.        
  146    Call to private/internal method rewrite_date_sub() of class Perflab_SQLite_PDO_Driver.              
  147    Call to private/internal method rewrite_regexp() of class Perflab_SQLite_PDO_Driver.                
  148    Call to private/internal method delete_workaround() of class Perflab_SQLite_PDO_Driver.             
  153    Call to private/internal method rewrite_date_sub() of class Perflab_SQLite_PDO_Driver.              
  154    Call to private/internal method rewrite_regexp() of class Perflab_SQLite_PDO_Driver.                
  158    Call to private/internal method rewrite_optimize() of class Perflab_SQLite_PDO_Driver.              
  168    Call to private/internal method return_true() of class Perflab_SQLite_PDO_Driver.                   
 ------ ---------------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------------------- 
  Line   modules/database/sqlite/wp-includes/sqlite/class-perflab-sqlite-pdo-engine.php                         
 ------ ------------------------------------------------------------------------------------------------------- 
  1260   Call to private/internal method convert_to_columns_object() of class Perflab_SQLite_PDO_Engine.        
  1262   Call to private/internal method convert_to_index_object() of class Perflab_SQLite_PDO_Engine.          
  1264   Call to private/internal method convert_result_check_or_analyze() of class Perflab_SQLite_PDO_Engine.  
 ------ ------------------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------------------- 
  Line   modules/database/sqlite/wp-includes/sqlite/class-perflab-sqlite-pdo-user-defined-functions.php         
 ------ ------------------------------------------------------------------------------------------------------- 
  324    Call to private/internal method derive_interval() of class Perflab_SQLite_PDO_User_Defined_Functions.  
  356    Call to private/internal method derive_interval() of class Perflab_SQLite_PDO_User_Defined_Functions.  
 ------ ------------------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------ 
  Line   modules/images/dominant-color-images/hooks.php                                
 ------ ------------------------------------------------------------------------------ 
  19     Call to private/internal function _dominant_color_get_dominant_color_data().  
 ------ ------------------------------------------------------------------------------ 

 ------ ------------------------------------------------------------------------------------ 
  Line   modules/images/webp-uploads/helper.php                                              
 ------ ------------------------------------------------------------------------------------ 
  222    Call to private/internal function webp_uploads_generate_additional_image_source().  
 ------ ------------------------------------------------------------------------------------ 

 ------ ------------------------------------------------------------------------------------ 
  Line   modules/images/webp-uploads/hooks.php                                               
 ------ ------------------------------------------------------------------------------------ 
  82     Call to private/internal function webp_uploads_generate_additional_image_source().  
  126    Call to private/internal function _wp_image_meta_replace_original().                
 ------ ------------------------------------------------------------------------------------ 
```

</p>
</details>

westonruter: > Plus https://packagist.org/packages/phpstan/phpstan-phpunit

Thanks for that. Added in 916dd25484ea42fe16356e80816e018fa1fad8ac, fixing 2 found issues in 5e498d1b28b9f74ab25e9b8037032c67b85f64c5.

westonruter: > and https://packagist.org/packages/php-stubs/wordpress-tests-stubs

Do we need the test subs since we're already installing `wp-phpunit/wp-phpunit`? Would this essentially be an alternative to:

https://github.com/WordPress/performance/blob/5e498d1b28b9f74ab25e9b8037032c67b85f64c5/phpstan.neon.dist#L14-L15

westonruter: > https://packagist.org/packages/phpstan/phpstan-deprecation-rules is also nice to prevent usage of deprecated functions.

Added in 4554f6af2b7aa9343d57eb16b8e33fc214833a14, issues fixed in bd1c3783570b23fd3213d25b189415bd7ac3936c.

swissspidy: > I love this package, but I wonder if this is the best for for Performance Lab considering that the code in question will end up in core and thus it is expected that many calls to internal/private functions will be made? As it stands right now...

Good point! 😃 

I think it can't hurt to use it anyway, since not everything will land in core. Places where usage is OK can be ignored in the PHPStan config accordingly. But maybe in another PR :-)

All the `Perflab_SQLite_` reports are actually false positives! I'll get this fixed in a new release. (Edit: see https://github.com/swissspidy/phpstan-no-private/issues/2)
They are not properly documented though. A `private function ...()` method does not need `@access private` annotations. That's against the coding standards.

> Do we need the test subs since we're already installing `wp-phpunit/wp-phpunit`? Would this essentially be an alternative to:

Probably not in that case. Counter question: why is `wp-phpunit` even used here when all tests are running in wp-env anyway?

felixarntz: Agreed, also if our code here follows PHPStan, I don't think it would violate anything that core requires - we'd effectively have stricter requirements than core, which I think should work fine as long as it's in that direction and not the opposite :)

> why is `wp-phpunit` even used here when all tests are running in wp-env anyway?

I believe that is for additional flexibility. We don't want to _dictate_ anyone to use `wp-env`, it's just the easy way to go. If folks want to e.g. run a DB on their own machine, the use of `wp-phpunit` makes that cleaner and decouples from having to have the entire WP core dev environment.

tests/modules/images/webp-uploads/helper-tests.php
westonruter: Done in 93ec3dce1038874f539ff6e0f15c636304ebb3d2

composer.json
westonruter: > 3\. In GHA workflow, just-in-time `composer remove` this package before doing `composer install` prior to running PHPUnit in PHP<7.2.

I just noticed that Gutenberg actually [takes this approach](https://github.com/WordPress/gutenberg/blob/064f7319b8f7a8860ef4d295e37a53e46bcebae6/.github/workflows/unit-test.yml#L118-L140) instead of maintaining a separate `composer.json`.

westonruter: How is this less maintainable than maintaining a separate `composer.json`?

felixarntz:
@westonruter @swissspidy Could you please follow up on the discussions in this PR to figure out how to move it forward? Not critical, but since it's milestoned for 2.4.0 and stale for 2 weeks now, it would be great to keep this moving.

westonruter:
@felixarntz I think the only current holdup is https://github.com/WordPress/performance/pull/730#discussion_r1195057672 and how we deal with Composer dependencies that aren't compatible with 5.6. However, since tests don't currently run in 5.6, I think that issue should be punted for when #399 is addressed.

felixarntz:
@westonruter That works for me, given #399 is not currently milestoned and it seems to me this PR here is closer to completion than #544 (which is from September 2022 so likely stale at least to some degree).

westonruter:
I think this just needs an approval than so it can be merged.

felixarntz (approved):
Thanks @westonruter, LGTM!

felixarntz:
@westonruter 
> I think this just needs an approval than so it can be merged.

I've approved it, but every PR needs 2+ approvals, so let's wait for @swissspidy to give it another look next week.

westonruter:
Once this is merged we should open a PR that bumps PHPStan to level 1 and higher.

swissspidy (approved):

swissspidy:
@westonruter Mind creating the follow-up tickets for this?

westonruter:
> Once this is merged we should open a PR that bumps PHPStan to level 1 and higher.

See https://github.com/WordPress/performance/issues/775

> I can also highly recommend https://packagist.org/packages/swissspidy/phpstan-no-private to catch calls of seemingly private functions.

See https://github.com/WordPress/performance/issues/776

PR #732, opened 2023-05-18
Update npm dependencies

thelovekesh:
## Summary

This PR updates npm dependencies to their latest versions, excluding `chalk` which needs to be updated to the latest ESM module. 

Additionally, it updates the wp-env container for running tests and composer, as the latest wp-env package no longer supports standalone `phpunit` and `composer` containers.

## Checklist


thelovekesh:
Why https://github.com/WordPress/performance/pull/732/commits/937daeda13c9aab75b77a04e1fff209646fed79c?

Despite being a peer dependency, prettier was installed with the prior lock file(npm doesn't install peer dependecies now), which led to the use of prettier as a formatter. We must explicitly install prettier if we wish to utilize it as a formatter.

@mukeshpanchal27 @felixarntz Please let me know if you think we can use prettier as the formatter.

westonruter:
We should also configure Dependabot, yeah? If not in this PR, than another.

thelovekesh:
Closing this since it's long outdated.

PR #733, opened 2023-05-22
Implement new module to use Speculation Rules API for prerendering documents on hover

### LINKED ISSUE #897: Module Proposal: Speculation prerendering & prefetching with the Speculation Rules API
## Overview
- Proposed module name: Speculation Rules
- Proposed module slug: speculation-rules
- Proposed module owner GitHub username(s): felixarntz

### Purpose
The [Speculation Rules API](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API) is a relatively new browser API which allows "speculative" prefetching and prerendering. Historically, prefetching and prerendering has only been possible for specific "hard-coded" resources, which in reality is often difficult or even unrealistic to get right across an ever-changing website. The Speculation Rules API allows sites to define certain rules to dynamically prefetch and prerender certain resources, e.g. when the user hovers over or clicks on a link to another URL.

This proposed module enables prefetching and prerendering with the new API. The necessary API capabilities will only become publicly available in Chrome 121 (likely January 2024), and right now they can be used with an origin trial (see #733). As such, the plan is to develop this module in a feature branch and only launch it once the feature is rolled out publicly and once the minimum requirements agreed for the module on have been implemented.

### Scope
* The module should prerender frontend URLs on hover by default.
* Certain URLs should be excluded from this behavior (e.g. `wp-login.php`, or anything in `wp-admin`). This list should be customizable so that plugins that implement more dynamic capabilities on certain URLs can exclude those as well.
* UI controls should be provided in WP Admin to change the behavior to prefetch URLs instead of prerendering, and another one to change the behavior from prefetching/prerendering on hover to only do it on click.
    * Both alternatives are less beneficial from a performance perspective, but potentially preferable for other reasons: For example, prerendering can be problematic with dynamic content, and relying on hover can result in wastefully loaded pages, if the user doesn't actually end up navigating there.
    * As this is a Performance Lab module, it makes sense with the more performant setting by default. For a potential future WordPress core proposal, the default would have to be determined based on user feedback, and UI controls would most likely not be part of the WordPress core implementation. For early feedback, they are crucial though to make it easy to experiment with different configurations.
* It should be possible for certain URLs to only be excluded from prerendering, but not from prefetching. In other words, if the site is configured to prerender, those URLs would fall back to prefetching. If the site is configured to prefetch, those URLs would just work like any other.

The first two points above are already being implemented in #733. The remaining points would be implement in follow up pull requests against the feature branch `feature/speculationrules`.

### Rationale
WordPress has supported prefetching and prerendering via `link` tags for several years, however using those tags correctly is difficult. It is furthermore challenging from an ecosystem perspective: For instance, if every plugin was going to add their resources to be prefetched or prerendered, it would eventually defeat the purpose (if everything is a priority, nothing is a priority). A more objective, rule based approach like one the Speculation Rules API offers makes more sense. As described above, such a dynamic approach is also easier to use and has the potential to unlock prefetching and prerendering capabilities for a large part of the WordPress ecosystem and the overall web.

### Other

Learn more about the Speculation Rules API:

* https://developer.chrome.com/blog/prerender-pages/#the-speculation-rules-api
* https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API
* https://docs.google.com/presentation/d/1GK92nCORW5vKd7LgGtTsgy35eqTV7P71l05pHsni8ok/edit

felixarntz:
## Summary

Fixes #897.

ThisPR is for a new module module to use the recently introduced Speculation Rules API to prerender URLs in WordPress as soon as the user hovers over them.

## Learn more about the Speculation Rules API

* https://developer.chrome.com/blog/prerender-pages/#the-speculation-rules-api
* https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API
* https://docs.google.com/presentation/d/1GK92nCORW5vKd7LgGtTsgy35eqTV7P71l05pHsni8ok/edit

Related: [URL pattern testing tool](https://urlpattern.glitch.me/)

## How to use

* The Speculation Rules API is available in Chrome 108 or newer.
* The full set of features however, which are needed for this module to work, are only available via origin trial in Chrome 113 or newer. You need to sign up for the origin trial with the origin of the site where you would like to test it. [Sign up for the origin trial](https://github.com/WICG/nav-speculation/blob/main/chrome-2023q1-experiment-overview.md)
    * Once you have signed up and received your origin trial token, you can set it in a `PLSR_ORIGIN_TRIAL_TOKEN` constant (e.g. in your `wp-config.php` file) and the module will take care of including its opt-in tag.
* The API is expected to become fully available without origin trial in Chrome 121.
* For debugging in Chrome Developer Tools, you can enable a new section by going to _Settings > Experiments_ (within the developer toolbar) and check the box "Enable Preloading Status Panel in Application panel" (towards the very end of the list). This will enable a "Preloading & Prerendering" section in the "Application" panel (see screenshot below).
Preloading & Prerendering section in Chrome Developer Tools

## Relevant technical choices

* Currently, the list of speculation rules is hard-coded, which is just a starting point. Eventually, the module should be enhanced to include an API to manage the speculation rules programmatically in WordPress.

## Checklist


modules/js-and-css/speculation-rules/hooks.php
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	$script = <<<JS
+( function() {
+	if ( ! HTMLScriptElement.supports || ! HTMLScriptElement.supports( 'speculationrules' ) ) {
+		console.log( 'speculation rules not supported' );
+		return;
+	}
+
+	var specScript = document.createElement( 'script' );
```
westonruter: Nit: Might as well use ES6

```suggestion
	const specScript = document.createElement( 'script' );
```

modules/js-and-css/speculation-rules/hooks.php
```diff
@@ -0,0 +1,66 @@
...
+	specScript.textContent = '%s';
+	console.log( 'speculation rules added' );
+	document.body.append( specScript );
+} )();
+JS;
+
+	wp_print_inline_script_tag(
+		sprintf(
+			$script,
+			wp_json_encode( $rules )
```
westonruter: Currently this requires that there be no single quotes in the `$rules` or else a syntax error will occur. This can be guarded against with:

```suggestion
			esc_js( wp_json_encode( $rules ) )
```

However, why not just print the `script[type="speculationrules"]` element directly without adding it dynamically?

modules/js-and-css/speculation-rules/helper.php
```diff
@@ -0,0 +1,51 @@
...
+							'href_matches' => array(
+								'/wp-login.php\\\\?*#*',
+								'/wp-admin/*\\\\?*#*',
+							),
+						),
+					),
+					// And except for any links marked with a class to not prerender.
+					array(
+						'not' => array(
+							'selector_matches' => '.no-prerender',
```
adamsilverstein, line 81: Nice! I like that there is a standard method to "opt out". Curious if you chose this class name because it is already used by existing plugins?

modules/js-and-css/speculation-rules/helper.php
```diff
@@ -0,0 +1,51 @@
...
+						),
+					),
+				),
+			),
+			'eagerness' => 'moderate',
+		),
+	);
+
+	return array( 'prerender' => $prerender_rules );
```
westonruter, line 90: I suppose eventually this should be filterable?

felixarntz: Potentially, though a declarative API using a class to manage the speculation rules may be even better.

I think that's for future exploration, for now I'd like to focus just on getting a first prototype to work.

modules/js-and-css/speculation-rules/hooks.php
felixarntz: The JS approach allows to check for browser support first before uncondtionally adding the script tag. It's also referenced in the Chrome Developer blog article.

modules/js-and-css/speculation-rules/helper.php
felixarntz: This is just based on the examples I've seen in existing learning resources about the new API.

modules/js-and-css/speculation-rules/hooks.php
westonruter: Curious why that approach was taken in the article. If the browser doesn't support it, it would just be ignored, right? It would seem preferable to just do:

```php
wp_print_inline_script_tag(
    wp_json_encode( $rules ),
    array( 'type' => 'speculationrules' )
);
```

felixarntz: Potentially yes. Let's sort out whether it doesn't have any side effects on browsers that don't support it - if so, I agree it's simplest to just include like you're saying.

modules/js-and-css/speculation-rules/hooks.php
felixarntz: @westonruter Is it safe to use ES6 here? I know it's widely supported at this point, but as far as I can tell this code otherwise works with older browsers as well. If that is indeed the case, then using `const` over `var` is not enough benefit IMO to make the code _less_ compatible.

westonruter: Since there will be zero browsers that support Speculation Rules API that don't also support ES6 (as it is only new browsers that support it), it seems fine to not require ES5.  But this is moot if https://github.com/WordPress/performance/pull/733#discussion_r1203016080 is implemented.

felixarntz: Right, but the reason for this script is specifically the case where browsers _don't_ support Speculation Rules API, hence my point :)

westonruter: Nevertheless, I think those specific browsers are so few and far between that we shouldn't worry about them. I just checked and usage of all versions of IE adds up to 0.47%, which is well below the 1% threshold for [WordPress browser support](https://make.wordpress.org/core/handbook/best-practices/browser-support/). In fact, IE11 hasn't been supported [since WP 5.8](https://make.wordpress.org/core/2021/04/22/ie-11-support-phase-out-plan/). I've been inquiring about this in relation to optimizing emoji-loader which would be facilitated by not having to support IE11. I raised this as a [question in #core Slack](https://wordpress.slack.com/archives/C02RQBWTW/p1686243321666769).

In any case, [`const` would be supported by IE11 here](https://caniuse.com/const). But you're right, `var` works equivalently here.

modules/js-and-css/speculation-rules/hooks.php
```diff
@@ -0,0 +1,66 @@
...
+ * Hook callbacks used for Speculation Rules.
+ *
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+/**
+ * Prints the speculation rules in a cross-browser compatible way.
+ *
+ * For browsers that do not support speculation rules yet, the rules will not be loaded.
```
domenic: I think it would be simpler to just always output the `<script type=speculationrules>`, instead of using JavaScript to dynamically insert it. Browsers that do not support `<script type=speculationrules>` will do nothing, which is the same as with the JS-based approach, but with less script execution cost.

westonruter: +1 cf. https://github.com/WordPress/performance/pull/733#discussion_r1204705839

felixarntz:
@westonruter @domenic I've addressed your feedback.

Note that the unit test failures are due to an unrelated problem in `trunk`, see #881 for a fix for that.

domenic:
This looks great to me!

Note that if you want to only target Chrome 121+, you could simplify a couple things:

- No origin trial token should be necessary. (Pending approvals...)
- The `?*` and `?*#*` parts are not necessary anymore. (Already shipped, https://chromium-review.googlesource.com/c/chromium/src/+/4975595 .)

modules/js-and-css/speculation-rules/helper.php
adamsilverstein: +1 I agree we could use a more declarative API for both inclusion and exclusion lists, or maybe even a way of tagging elements with rules (plus filters of course).

This came up talking to optimization plugins that already implement prefetch/render either manually or with scripts like instant.page or quicklinks. These plugins need to make sure they exclude specific routes that might affect state, like clicking a signup link. They have to both discover and maintain these links, usually introduced by other plugins.

Ideally core would offer a way for other plugins to register routes to exclude from prefetch/render. Similarly we could offer a way to register routes you want to make sure to prefetch/render.

felixarntz:
@domenic 
> Note that if you want to only target Chrome 121+, you could simplify a couple things:
> 
> * No origin trial token should be necessary. (Pending approvals...)
> * The `?*` and `?*#*` parts are not necessary anymore. (Already shipped, https://chromium-review.googlesource.com/c/chromium/src/+/4975595 .)

For now, I'll keep the Origin-Trial. Eventually this should be removed for sure, before the module ships in a Performance Lab version. But now for early testing IMO worth keeping.

For your second point, probably the same. Unless this is already shipped or will ship in an earlier Chrome version? Is there any feature detection available for that change?

modules/js-and-css/speculation-rules/helper.php
felixarntz: I think for now the simple exclude paths filter allows the most critical integration, that should work for an MVP.

Eventually it's worth building out a more bespoke API, but the nesting of these rules can also be quite complex, and I would want to make sure we keep the complexity contained so that external developers have to worry about it as little as possible.

It's also worth thinking further about what _should_ be possible to be modified vs what would be discouraged from the core side.

I'd say let's go with the simple version and explore a more advanced API in a future issue, some time after this has shipped.

domenic:
> For your second point, probably the same. Unless this is already shipped or will ship in an earlier Chrome version? Is there any feature detection available for that change?

This change is also Chromium 121+. It's harmless to keep the extra parts, so if you're targeting such browsers then I'd suggest keeping it.

You can feature-detect this using the following:

```js
const hasNewSemantics = (new URLPattern("/*", "https://example.com/a?b&c")).test("https://example.com/d?e");
```

modules/js-and-css/speculation-rules/helper.php
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+/**
+ * Prints the speculation rules in a cross-browser compatible way.
+ *
+ * For browsers that do not support speculation rules yet, the rules will not be loaded.
+ *
```
westonruter: Is this relevant here?

```suggestion
```

modules/js-and-css/speculation-rules/helper.php
```diff
@@ -0,0 +1,82 @@
...
+ * @since n.e.x.t
+ *
+ * @return array Associative array of speculation rules by type.
+ */
+function plsr_get_speculation_rules() {
+	$base_href_exclude_paths = array(
+		'/wp-login.php',
+		'/wp-admin/*',
```
westonruter: I think it's also important to include any URL that includes a `_wpnonce` query parameter. Typically such nonce URLs point to the admin, but there are key examples where this is not the case: https://wpdirectory.net/search/01HFCME02F3HYMR93WDWV6KSX4

For example, BuddyPress has frontend nonce links to favorite/unfavorite a link.

modules/js-and-css/speculation-rules/helper.php
```diff
@@ -0,0 +1,82 @@
...
+
+	// Ensure that there are no duplicates and that the base paths cannot be removed.
+	$href_exclude_paths = array_map(
+		static function ( $exclude_path ) {
+			if ( ! str_starts_with( $exclude_path, '/' ) ) {
+				$exclude_path = '/' . $exclude_path;
+			}
+			return $exclude_path . '\\?*#*';
+		},
+		array_unique(
```
westonruter: Shouldn't the `array_map()` and `array_unique()` calls be swapped? In other words, it's possible that the map callback could result in a new duplicate after the `array_unique()` call reduplicated the initial paths.

jeremyroman:
I noticed this unconditionally chooses `prerender`. Could/should this be an administrator-controlled setting? There are reasons a developer might prefer `prefetch` to `prerender` under some circumstances (for example, a script which is not ready for prerender runs, or a prefetch could succeed but the site would be too heavy for prerender to be practical).

modules/js-and-css/speculation-rules/helper.php
```diff
@@ -0,0 +1,82 @@
...
+
+	$prerender_rules = array(
+		array(
+			'source'    => 'document',
+			'where'     => array(
+				'and' => array(
+					// Prerender any URLs within the same site.
+					array(
+						'href_matches' => '/*\\?*',
+						'relative_to'  => 'document',
```
jeremyroman: `relative_to` is unnecessary if the rules are inline in the document (but if you do have some reason for this to be `relative_to`, the below exclusion probably should be too.

Also, if WordPress is installed at a path other than `/`, does that need to be accounted for?

modules/js-and-css/speculation-rules/helper.php
felixarntz: Right, this is a tricky one though. Nonces can also be sent through any other kind of parameter. I'm not sure how far we should go here by default, as what you're suggesting would just be a "best guess", but not actually reliable. So potentially it's better to support just what WordPress core does and require plugins to take care of this themselves. I don't think core uses nonces in the frontend.

I realize this may not work either (as adoption is hard), but also questioning whether it's reasonable to start "patching" for what plugins may do here.

modules/js-and-css/speculation-rules/helper.php
felixarntz: Added support for that in https://github.com/WordPress/performance/pull/733/commits/ce0e56bb7adf52f5b2141b167226c3382099e955

modules/js-and-css/speculation-rules/helper.php
westonruter: It's true that nonces can be sent via any parameter, but the `wp_nonce_url()` function uses the `_wpnonce` parameter by default. So I think it makes sense to omit them by default.

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
```diff
@@ -0,0 +1,91 @@
...
+	 * The given URL path pattern is only prefixed if it does not already include the expected prefix.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $path_pattern URL pattern starting with the path segment.
+	 * @param string $context      Optional. Either 'home' (any frontend content) or 'site' (content relative to the
+	 *                             directory that WordPress is installed in). Default 'home'.
+	 * @return string URL pattern, prefixed as necessary.
+	 */
+	public function prefix_path_pattern( string $path_pattern, string $context = 'home' ) {
```
westonruter:
```suggestion
	public function prefix_path_pattern( string $path_pattern, string $context = 'home' ): string {
```

modules/js-and-css/speculation-rules/helper.php
```diff
@@ -0,0 +1,92 @@
...
+ * @since n.e.x.t
+ *
+ * @return array Associative array of speculation rules by type.
+ */
+function plsr_get_speculation_rules() {
+	$prefixer = new PLSR_URL_Pattern_Prefixer();
+
+	$base_href_exclude_paths = array(
+		$prefixer->prefix_path_pattern( '/wp-login.php', 'site' ),
+		$prefixer->prefix_path_pattern( '/wp-admin/*', 'site' ),
```
westonruter, line 24: I wonder, is the `PLSR_URL_Pattern_Prefixer` class really needed? Wouldn't this be simpler just doing:

```suggestion
	$base_href_exclude_paths = array(
		site_url( 'wp-login.php', 'login' ),
		admin_url() . '*',
```

modules/js-and-css/speculation-rules/helper.php
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * @since n.e.x.t
+ *
+ * @return array Associative array of speculation rules by type.
+ */
+function plsr_get_speculation_rules() {
+	$prefixer = new PLSR_URL_Pattern_Prefixer();
+
+	$base_href_exclude_paths = array(
+		$prefixer->prefix_path_pattern( '/wp-login.php', 'site' ),
```
westonruter, line 23: I see that `wp_login_url()` makes use of the `login` scheme rather than `site`: `site_url( 'wp-login.php', 'login' )`

modules/js-and-css/speculation-rules/helper.php
```diff
@@ -0,0 +1,92 @@
...
+		$prefixer->prefix_path_pattern( '/wp-admin/*', 'site' ),
+	);
```
westonruter, line 25: What about the other PHP files outside of the normal frontend:

* `wp-activate.php`
* `wp-signup.php`

modules/js-and-css/speculation-rules/helper.php
felixarntz: That's not the same, as these functions return full URLs. The current code only gets the paths.

We could do that in a single function too, which is actually where I started, but that approach is not properly testable, so a class like this is more suitable.

modules/js-and-css/speculation-rules/helper.php
felixarntz: See above, this shouldn't be relevant as it's paths only.

modules/js-and-css/speculation-rules/helper.php
felixarntz: I was thinking about those too, but left them out for now. We can probably add them, but I was thinking that it's not critical and could be done in its own issue once we have a first version of the module merged (i.e. this PR).

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
felixarntz: Updated in https://github.com/WordPress/performance/pull/733/commits/cd536d5ef082203e1c1f36f3945927f710798878

modules/js-and-css/speculation-rules/helper.php
westonruter: OK, I see it needs to be paths. So then why not:

```suggestion
	$base_href_exclude_paths = array(
		wp_parse_url( site_url( 'wp-login.php', 'login' ), PHP_URL_PATH ),
		wp_parse_url( admin_url(), PHP_URL_PATH ),
```

domenic: Full URLs are also fine; there's no need to restrict yourself to relative URLs.

felixarntz: @westonruter We need to prefix all results from the filter in the same way anyway, so it's IMO a bit pointless to use the WP functions directly here, when we use the prefixer class below.

Of course we could use the WP functions in all cases, but that would be slower, since for every URL it would have to "recalculate" those base URLs, parse the paths etc. The class only does that once, when instantiated. That approach also makes it easier to cover the functionality by unit tests.

westonruter: Yes, I suppose it would be a tiny bit slower, but these functions are being used all over the place in WordPress already without any kind of caching so adding two more function calls isn't really going to make a big impact. IMHO, `PLSR_URL_Pattern_Prefixer` seems a bit overkill. Also, what about when a site filters the login URL to be something different using the `login_url` filter? For example, a site that I help maintain has some plugin which filters the login URL to be `.../webadmin/` (for some reason) with `.../wp-login.php` returning a 404. The uses of the prefixer blow could be adapted to use `home_url()`, right?

felixarntz: I think this is a bit of a personal preference. Not having this class is more pragmatic, but less testable, and the latter makes it more error-prone (because of less test coverage). The class is simple and decoupled in a way that it can be used easily alongside WordPress's function-driven approach.

Good point regarding the `login_url` filter, it's probably safer to use the `wp_login_url()` function rather than hard-coding `wp-login.php`. Similarly, we should probably use the `admin_url()` function instead of hard-coding `wp-admin`. I'll make that change.

adamsilverstein, line 24: Why exclude all `wp-admin` paths, wouldn't you want to enable prefetching on those to speed up `wp-admin`? or is this intended strictly for front end so far?

felixarntz: @adamsilverstein Yes, for now this is only frontend. The whole speculation rules tag is also printed in the frontend only.

Doing it in WP Admin has its own set of complexities, and is less beneficial IMO. Those pages are also a whole lot more dynamic, which makes prerendering difficult.

felixarntz:
This PR is now ready for full review and, once approved, can be merged into the new feature branch. Please see the new module proposal #897 for additional context, including follow up work.

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
```diff
@@ -0,0 +1,91 @@
...
+	public function prefix_path_pattern( string $path_pattern, string $context = 'home' ): string {
+		// If context path does not exist, the context is invalid.
+		if ( ! isset( $this->contexts[ $context ] ) ) {
+			_doing_it_wrong(
+				__FUNCTION__,
+				sprintf(
+					/* translators: %s: context string */
+					esc_html__( 'Invalid context %s.', 'performance-lab' ),
+					esc_html( $context )
+				),
```
westonruter: Nit, I think generally it's better to escape the overall output rather than escape each part. Easier to prevent unescaped content from sneaking in, and easier to review for security.

```suggestion
				esc_html(
					sprintf(
						/* translators: %s: context string */
						__( 'Invalid context %s.', 'performance-lab' ),
						$context
					)
				),
```

westonruter (approved):

modules/js-and-css/speculation-rules/helper.php
swissspidy: Preloading the editor when hovering over the link in the admin bar would be interesting in the future.

swissspidy (approved):

mukeshpanchal27 (approved):

PR #736, opened 2023-05-30
Make `Server-Timing` header output more robust

swissspidy:
## Summary

I activated the plugin on my local WP core Docker environment and filtered `perflab_server_timing_use_output_buffer` to enable output buffering, yet I did not see the `Server-Timing` header.

The `shutdown` hook is too late to send the header, but using the `ob_start()` callback for this works.

## Checklist

### DIFF
--- a/server-timing/class-perflab-server-timing.php
+++ b/server-timing/class-perflab-server-timing.php
@@ -227,16 +227,11 @@ class Perflab_Server_Timing {
 			return $passthrough;
 		}
 
-		ob_start();
-		add_action(
-			'shutdown',
-			function() {
-				$output = ob_get_clean();
+		ob_start(
+			function( $output ) {
 				$this->send_header();
-				echo $output;
-			},
-			// phpcs:ignore PHPCompatibility.Constants.NewConstants
-			defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : -1000
+				return $output;
+			}
 		);
 		return $passthrough;
 	}


felixarntz:
@swissspidy Can you share a bit more context? What leads to it being "too late"? I have never encountered the problem, so would like to understand better.

swissspidy:
When I run this in WP core‘s Docker environment, the callback in the shutdown hook is never fired, so it‘s not sending the header. I assume it‘s because the shutdown handler is called long after the page has been served.
Weirdly it‘s not an issue when using this with LocalWP, just with the core env.

felixarntz:
@swissspidy That's odd. I have been using this in the WP core Docker env several times without problems 🤔  Maybe it's something more specific than that?

swissspidy:
IDK 🤷 Maybe I can debug at contributor day or so

westonruter (approved):
This is the approach that the output buffering approach that the AMP plugin. It has been robust.

The AMP plugin also sends `Server-Timing` response headers.

server-timing/class-perflab-server-timing.php
```diff
@@ -227,16 +227,11 @@ public function on_template_include( $passthrough = null ) {
...
-			'shutdown',
-			function() {
-				$output = ob_get_clean();
+		ob_start(
+			function( $output ) {
 				$this->send_header();
-				echo $output;
-			},
-			// phpcs:ignore PHPCompatibility.Constants.NewConstants
-			defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : -1000
```
westonruter, line 239: Curious why this action was configured to run early. Wouldn't it be better to run as late as possible?

felixarntz:
Will the removal of the `ob_get_clean()` call in this PR lead to issues? Not very familiar with how output buffers work under the hood, but I always assumed you should also "end" an output buffer that you start.

server-timing/class-perflab-server-timing.php
felixarntz: It is run early based on the idea that the template output was already completed when `shutdown` runs. If `shutdown` did a bunch of other things (extremely unlikely but possible), then it would inflate the time in the `wp-template` metric.

westonruter:
Any open buffers are automatically flushed when the page response is finished, per [docs](https://www.php.net/manual/en/function.ob-start.php#:~:text=If%20output%20buffering%20is%20still%20active%20when%20the%20script%20ends%2C%20PHP%20outputs%20the%20contents%20automatically.):

> If output buffering is still active when the script ends, PHP outputs the contents automatically.

felixarntz (approved):
Tested this in my core dev environment, and it works as expected, also no change in outcomes. LGTM, thanks for the fix!

PR #739, opened 2023-05-31
Implement migration prompt to migrate from SQLite module to standalone plugin due to removal in the following release

### LINKED ISSUE #737: Implement notice to prompt administrator to update to SQLite plugin instead of SQLite module
As discussed in the [Performance chat on April 25](https://make.wordpress.org/core/2023/04/25/performance-chat-summary-25-april-2023/), the SQLite feature will going forward no longer be directly associated with the Core Performance Team. This is mostly because it is not itself a performance focused feature. Its effects on performance are certainly considerable, but to assume that the primary reason for a site to use SQLite is performance would be wrong.

The Core Performance Team still intends to support the project and help assess its performance in the future, but it will no longer be directly tied to the team in terms of responsibility and maintenance.

The [SQLite standalone plugin](https://wordpress.org/plugins/sqlite-database-integration/) has already existed for several months at this point, and the duplicate codebases between the [standalone plugin's GitHub repository](https://github.com/WordPress/sqlite-database-integration) and the [Performance Lab module code](https://github.com/WordPress/performance/tree/trunk/modules/database/sqlite) has been confusing and problematic to handle. This issue therefore focuses on introducing a notice to administrators on sites where the Performance Lab module is active to migrate to the standalone plugin instead.

This notice should be launched as part of the upcoming Performance Lab 2.4 release (June 19), so that the module can then be completely removed from Performance Lab in the 2.5 release (July 17).

Note that this effort is decoupled from #656, since the SQLite module is not being removed as part of that effort, but rather simply due to the above decision. So it shouldn't need to be blocked by the efforts in the greater unbundling effort.

## Requirements
* Implement a WordPress admin notice (in "warning" style) that conditionally shows as long as the SQLite module is active:
    * The notice should be displayed to all users that can `activate_plugins`, and it should explain that the SQLite module will be removed in the upcoming 2.5 release of Performance Lab.
    * If the SQLite standalone plugin is not installed, the notice should prompt the user to install and activate it, and then disable the SQLite module in Performance Lab.
    * Else if the SQLite standalone plugin is not active, the notice should prompt the user to activate it and then disable the SQLite module in Performance Lab.
    * Else the notice should prompt the user to disable the SQLite module in Performance Lab.
* Implement a WordPress admin pointer pointing to the Performance Lab's submenu item (like the existing pointer) which should show under the same conditions as the existing pointer, but only if the SQLite module is active:
    * It should simply inform the user that the SQLite module will be removed and that an action is necessary to keep its functionality.
    * The admin notice should be the source of truth, so the pointer content should be kept brief. It is just added to put more attention to it, since admin notices are so heavily used that they may go unnoticed (which defeats their purpose).
    * In the case that both admin pointers would be relevant to show, the new pointer should take precedence as it contains more urgent information.

felixarntz:
## Summary

Fixes #737

## Relevant technical choices

* An admin notice is added with 3 steps to migrate, each of which is conditionally marked already completed depending where the site owner is at in the process.
* An admin pointer is added as an additional way to make site owners aware that they need to take an action, since admin notices can easily be overlooked across _other_ admin notices.
* A few minor interoperability fixes we added (in the `modules/database/sqlite` files).
* No new functions are being introduced since they would need to be removed again in the following release. Therefore the new **temporary** code relies heavily on closures.
* Also see the requirements on the issue.

## Screenshots
**Admin notice awaiting step 1**
Screenshot 2023-05-31 at 3 57 35 PM

**Admin notice awaiting step 2**
Screenshot 2023-05-31 at 3 59 29 PM

**Admin notice awaiting step 3**
Screenshot 2023-05-31 at 4 05 01 PM

**Admin pointer**
Screenshot 2023-05-31 at 3 06 46 PM

## Checklist

### DIFF
--- a/admin/load.php
+++ b/admin/load.php
@@ -647,14 +647,16 @@ add_action(
 		$step3_before      = $todo_before;
 		$step3_after       = $todo_after;
 		$step3_placeholder = 'SQLite';
-		if ( 'index.php' === $hook_suffix ) { // Link to Performance Lab settings.
+		if ( 'index.php' === $hook_suffix && defined( 'SQLITE_MAIN_FILE' ) ) {
+			// Link to Performance Lab settings.
 			$screen_url = add_query_arg(
 				'page',
 				PERFLAB_MODULES_SCREEN,
 				admin_url( 'options-general.php' )
 			);
 
-			$step3_placeholder = '<a href="' . esc_url( $screen_url ) . '">' . $step3_placeholder . '</a>';
+			$step3_before .= '<a href="' . esc_url( $screen_url ) . '">';
+			$step3_after   = '</a>' . $step3_after;
 		}
 
 		/*
--- a/admin/load.php
+++ b/admin/load.php
@@ -201,7 +201,10 @@ function perflab_render_modules_page_field( $module_slug, $module_data, $module_
 			<?php else : ?>
 				<?php // Don't use the WP notice classes here, as that makes them move to the top of the page. ?>
 				<p class="notice notice-warning" style="padding:1em;max-width:50em;">
-					<?php esc_html_e( 'Enabling this module will switch to a separate database and install WordPress in it. You will need to reconfigure your site, and start with a fresh site. Disabling the module you will get back to your previous MySQL database, with all your previous data intact.', 'performance-lab' ); ?>
+					<?php esc_html_e( 'It is not advised to activate this module since it will be removed in the upcoming Performance Lab release.', 'performance-lab' ); ?>
+					<a href="https://wordpress.org/plugins/sqlite-database-integration">
+						<?php esc_html_e( 'Please use the standalone plugin instead.', 'performance-lab' ); ?>
+					</a>
 				</p>
 			<?php endif; ?>
 		<?php endif; ?>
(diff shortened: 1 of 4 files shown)


admin/load.php
```diff
@@ -201,7 +201,10 @@ function perflab_render_modules_page_field( $module_slug, $module_data, $module_
 			<?php else : ?>
 				<?php // Don't use the WP notice classes here, as that makes them move to the top of the page. ?>
 				<p class="notice notice-warning" style="padding:1em;max-width:50em;">
-					<?php esc_html_e( 'Enabling this module will switch to a separate database and install WordPress in it. You will need to reconfigure your site, and start with a fresh site. Disabling the module you will get back to your previous MySQL database, with all your previous data intact.', 'performance-lab' ); ?>
+					<?php esc_html_e( 'It is not advised to activate this module since it will be removed in the upcoming Performance Lab release.', 'performance-lab' ); ?>
+					<a href="https://wordpress.org/plugins/sqlite-database-integration">
```
mukeshpanchal27, line 205: Could we open URL in new tab?

admin/load.php
```diff
@@ -512,10 +565,155 @@ function perflab_plugin_action_links_add_settings( $links ) {
...
+
+		$todo_before = '<span>';
+		$todo_after  = '</span>';
+		$done_before = '<span style="text-decoration: line-through;">';
+		$done_after  = '</span> ✅';
+
+		if ( file_exists( WP_PLUGIN_DIR . '/sqlite-database-integration/load.php' ) ) {
+			$step1_before      = $done_before;
+			$step1_after       = $done_after;
+			$step1_placeholder = 'SQLite Database Integration';
```
mukeshpanchal27, line 619: Do we need to use translation function?

mukeshpanchal27 (commented):
Thanks, @felixarntz. It looks good to me. I have some nitpick feedback and a few questions.

Sometimes, when we activate the SQLite Database Integration through step 2 and then return to the Performance Lab settings page, deactivating the module also deactivates the SQLite Database Integration plugin. is it intended behaviour?

admin/load.php
felixarntz: For accessibility reasons, opening in a new tab by default is usually discouraged. So I think it's better to leave this as is.

admin/load.php
felixarntz: Plugin names should never be translated, so I think we can avoid that here for that reason.

felixarntz:
@mukeshpanchal27 
> Sometimes, when we activate the SQLite Database Integration through step 2 and then return to the Performance Lab settings page, deactivating the module also deactivates the SQLite Database Integration plugin. is it intended behaviour?

I noticed that too, that's okay. It is due to the migration routine that the standalone plugin comes with. In the cases where the plugin takes care of that automatically, even better, one less step for the user.

mukeshpanchal27 (approved):
Thanks @felixarntz, LGTM!

felixarntz:
@aristath @joemcgill Would one of you be able to review this PR please?

aristath (approved):
LGTM! 👍

PR #764, opened 2023-06-22
Remove SQLite module

### LINKED ISSUE #738: Remove SQLite module from Performance Lab plugin
Follow-up to #737: The SQLite module should be removed from the Performance Lab plugin _after_ the upcoming Performance Lab 2.4 release (June 19), i.e. as part of the following 2.5 release (July 17).

Note that this effort is decoupled from https://github.com/WordPress/performance/issues/656, since the SQLite module is not being removed as part of that effort, but rather simply due to the above decision. So it shouldn't need to be blocked by the efforts in the greater unbundling effort.

## Requirements
* Remove `database/sqlite` module entirely, as well as any references to it in the files `phpcs.xml.dist`, `admin/load.php`.
* Remove the migration code that was added as part of #737.
* Add an upgrade notice informing about the removal of the module to the end of `readme.txt` to accompany the 2.5 release:
```
== Upgrade Notice ==

= 2.5.0 =

The SQLite module is no longer present starting with this release. If you still use it, please migrate to the standalone plugin before updating.
```

felixarntz:
## Summary

Fixes #738

## Checklist


mukeshpanchal27 (commented):
Thanks @felixarntz for the PR! It looks solid.

Can we also consider removing certain functions and text associated with SQLite?

- Activate and deactivated module code that introduce through module integration that we did not use for any other module.
  - `perflab_run_module_activation_deactivation()`
  - `perflab_activate_module()`
  - `perflab_deactivate_module()`
  - `'update_option_' . PERFLAB_MODULES_SETTING` action
  - `'add_option_' . PERFLAB_MODULES_SETTING` action
- Remove `SQLite Integration` from [readme.txt](https://github.com/WordPress/performance/blob/trunk/readme.txt#L26)

aristath (approved):

felixarntz:
@mukeshpanchal27 While the functions you reference were added as part of the SQLite PR, they are generally applicable and part of the infrastructure itself, as they could be useful for any module in the future. So I think we should keep them.

Regarding the `readme.txt` update, this will be taken care of automatically as part of the release process, via `npm run readme`.

mukeshpanchal27 (approved):

PR #769, opened 2023-06-28
Check for fetchpriority feature being available in WordPress core before loading the module

felixarntz:
## Summary

This PR checks for existence of the fetchpriority feature in WordPress core, which was merged earlier this week and will become widely available with the WordPress 6.3 launch in August (see https://core.trac.wordpress.org/ticket/58235). We should therefore ensure _before_ that release that Performance Lab caters for that situation.

## Relevant technical choices

* For the module, a `can-load.php` file is added, the default mechanism to check whether a module can be loaded, with the closure returning `false` when the WP core version on the site already provides the feature.
* For the standalone plugin version of the module, the `can-load.php` file is irrelevant, therefore the PR adds an admin notice closure which is only shown for the standalone plugin (based on the specific constant check).
* The standalone plugin's `readme.txt` has been updated and the standalone plugin version is bumped to 1.1.0. This will automatically be published together with the next Performance Lab plugin release on July 17.

## Screenshots

**Performance Lab module checkbox when feature already in core**
Screenshot 2023-06-28 at 10 54 13 AM

**Admin notice for standalone plugin when feature already in core**
Screenshot 2023-06-28 at 11 03 17 AM

## Checklist

### DIFF
--- a/modules/images/fetchpriority/can-load.php
+++ b/modules/images/fetchpriority/can-load.php
@@ -1,0 +1,11 @@
+<?php
+/**
+ * Can load function to determine if the Fetchpriority feature is already available in WordPress core.
+ *
+ * @since   n.e.x.t
+ * @package performance-lab
+ */
+
+return static function() {
+	return ! function_exists( 'wp_get_loading_optimization_attributes' );
+};
--- a/modules/images/fetchpriority/hooks.php
+++ b/modules/images/fetchpriority/hooks.php
@@ -67,3 +67,19 @@ function fetchpriority_render_generator() {
 	}
 }
 add_action( 'wp_head', 'fetchpriority_render_generator' );
+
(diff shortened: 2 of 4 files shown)


kt-12 (approved):
Thank you @felixarntz. LGTM!. 

Just an observation, since we are not forcefully disabling `fetchpriority` for standalone plugin, there can be a chance of having 2 `fetchpriority` in a page - If the post thumbnail is less than the `50000px` threshold, but  one of the first 3 content image is larger than the threshold ( a very rare edge case )

felixarntz:
@kt-12 Fair point, thanks for raising. I think that's okay though, also given that the standalone plugin currently only has 20+ installs.

mukeshpanchal27 (approved):
Thanks @felixarntz for the PR, Look good to me.

PR #778, opened 2023-07-13
Fix whitespace not matching in `since` command.

joemcgill:
## Summary

This fixes an issue where the since command was failing to replace instances of `n.e.x.t` when more than once space occurred between the `@since` and the placeholder version number, e.g., for docblock alignment purposes.

## Relevant technical choices

While prepping for the 2.5.0 release, I noticed that running `npm run since -- -r 2.5.0` missed one case in [modules/images/fetchpriority/can-load.php](https://github.com/WordPress/performance/blob/release/2.5.0/modules/images/fetchpriority/can-load.php#L5) that was added during this release. It looks like this is due to the alignment spacing in the docblock, which is causing the regex pattern to not match. Updating the regex pattern to look for multiple spaces and apply the same number of spaces in the replacement fixes the issue.

## Checklist

### DIFF
--- a/bin/plugin/commands/since.js
+++ b/bin/plugin/commands/since.js
@@ -47,13 +47,14 @@ exports.handler = async ( opt ) => {
 		ignore: [ __filename, '**/node_modules', '**/vendor' ],
 	} );
 
-	const regexp = new RegExp( '@since n.e.x.t', 'g' );
+	const regexp = new RegExp( '@since(\\s+)n.e.x.t', 'g' );
+
 	files.forEach( ( file ) => {
 		const content = fs.readFileSync( file, 'utf-8' );
 		if ( regexp.test( content ) ) {
 			fs.writeFileSync(
 				file,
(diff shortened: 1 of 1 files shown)


felixarntz (approved):
@joemcgill Good catch, LGTM!

felixarntz:
@joemcgill Maybe adjust the base branch to `release/2.5.0` so it already applies in that one?

mukeshpanchal27 (approved):
Thanks @joemcgill for the PR. LGTM.

- I have updated the base branch to `release/2.5.0`.
- In https://github.com/WordPress/performance/commit/08d517644060273cc6b0ab0dab23b602b7e90612 we have similar string with spaces for since annotation in `modules/images/webp-uploads/can-load.php` file but it working fine for that release don't know what has changed 🤔

joemcgill:
Thanks for adjusting the branch, @mukeshpanchal27. I noticed the other place where this would have applied and thought the same thing! I assume that one just got adjusted manually.

PR #784, opened 2023-07-18
Implement logic to measure specific hook execution time with Server-Timing controlled by a WP Admin screen

felixarntz:
## Summary

This PR implements a WP Admin screen _Tools > Server-Timing_ which allows providing action and filter names. Any hook names provided will be measured and exposed in the `Server-Timing` header in the frontend.

While the Server-Timing API from #553 has been a powerful foundation, it can sometimes be cumbersome to measure specific hooks performance, since it always requires writing custom code. [Gists like this one](https://gist.github.com/felixarntz/63c05392dbf7d51cc7f8f4a424b1ff39) can be used for that, but even with such thing it may be easier to just add the action or filter to a UI instead of having to modify a plugin's code. This PR makes it really easy and gives site owners control about specific hooks to measure.

Admin screen to manage hooks to measure with Server-Timing

_The new admin screen with example input_

## Relevant technical choices

* Adds a new WordPress option/setting to store Server-Timing configuration.
* Adds a new WP Admin screen with textareas to provide action and filter names.
* Uses the data from the new option/setting to measure the duration of those hooks, using similar logic as previously implemented in [the aforementioned Gist](https://gist.github.com/felixarntz/63c05392dbf7d51cc7f8f4a424b1ff39).

## Testing
1. Go to _Tools > Server-Timing_.
2. Add some hook names (actions and/or filters), one per line. For example those from the screenshot above.
3. Save and verify the hook names are still shown in the textareas, in a sanitized way.
    * For example, try to add spaces or (other) invalid characters and check that they're stripped after submitting.
4. Load the frontend.
5. Type `console.table( performance.getEntries()[0].serverTiming )` in the browser's dev tools console.
6. Verify that you see measurements for the actions/filters you provided, via `wp-action-{$action}` / `wp-filter-{$filter}` metrics.
    * Note that metrics will only show up if the respective action/filter was actually run during the request.
    * If you provide hook names that are only run after the template is being parsed (e.g. `wp_enqueue_scripts` action), they won't show up by default. You'll need to activate output buffering (e.g. via [this Gist plugin](https://gist.github.com/felixarntz/9c3d7150c74082e69bb426393b68b12e)) so that those can be exposed as well.

## Checklist

### DIFF
--- a/admin/server-timing.php
+++ b/admin/server-timing.php
@@ -5,6 +5,10 @@
  * @package performance-lab
  */
 
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
 /**
  * Adds the Server-Timing page to the Tools menu.
  *
--- a/admin/server-timing.php
+++ b/admin/server-timing.php
@@ -1,0 +1,131 @@
+<?php
+/**
+ * Server-Timing API admin integration file.
+ *
+ * @package performance-lab
+ */
+
+/**
+ * Adds the Server-Timing page to the Tools menu.
+ *
+ * @since n.e.x.t
+ */
+function perflab_add_server_timing_page() {
+	$hook_suffix = add_management_page(
+		__( 'Server-Timing', 'performance-lab' ),
+		__( 'Server-Timing', 'performance-lab' ),
+		'manage_options',
+		PERFLAB_SERVER_TIMING_SCREEN,
+		'perflab_render_server_timing_page'
+	);
+
+	// Add the following hooks only if the screen was successfully added.
+	if ( false !== $hook_suffix ) {
+		add_action( "load-{$hook_suffix}", 'perflab_load_server_timing_page' );
+	}
+
+	return $hook_suffix;
+}
+add_action( 'admin_menu', 'perflab_add_server_timing_page' );
+
+/**
+ * Initializes settings sections and fields for the Server-Timing page.
+ *
+ * @since n.e.x.t
+ */
+function perflab_load_server_timing_page() {
+	add_settings_section(
+		'benchmarking',
+		__( 'Benchmarking', 'performance-lab' ),
+		static function() {
+			?>
+			<p>
+				<?php esc_html_e( 'In this section, you can provide hook names to include measurements for them in the Server-Timing header.', 'performance-lab' ); ?>
+			</p>
+			<?php
+		},
+		PERFLAB_SERVER_TIMING_SCREEN
+	);
+
+	/*
+	 * For all settings fields, the field slug, option sub key, and label
+	 * suffix have to match for the rendering in the callback to be
+	 * semantically correct.
+	 */
+	add_settings_field(
+		'benchmarking_actions',
+		__( 'Actions', 'performance-lab' ),
+		static function() {
+			perflab_render_server_timing_page_field( 'benchmarking_actions' );
+		},
+		PERFLAB_SERVER_TIMING_SCREEN,
+		'benchmarking',
+		array( 'label_for' => 'server_timing_benchmarking_actions' )
+	);
+	add_settings_field(
+		'benchmarking_filters',
+		__( 'Filters', 'performance-lab' ),
+		static function() {
+			perflab_render_server_timing_page_field( 'benchmarking_filters' );
+		},
+		PERFLAB_SERVER_TIMING_SCREEN,
+		'benchmarking',
+		array( 'label_for' => 'server_timing_benchmarking_filters' )
+	);
+}
+
+/**
+ * Renders the Server-Timing page.
+ *
+ * @since n.e.x.t
+ */
+function perflab_render_server_timing_page() {
+	?>
+	<div class="wrap">
+		<?php settings_errors(); ?>
+		<h1>
+			<?php esc_html_e( 'Server-Timing', 'performance-lab' ); ?>
+		</h1>
+
+		<form action="options.php" method="post" novalidate="novalidate">
+			<?php settings_fields( PERFLAB_SERVER_TIMING_SCREEN ); ?>
+			<?php do_settings_sections( PERFLAB_SERVER_TIMING_SCREEN ); ?>
+			<?php submit_button(); ?>
+		</form>
+	</div>
+	<?php
+}
+
+/**
+ * Renders a field for the given Server-Timing option.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $slug Slug of the field and sub-key in the Server-Timing option.
+ */
+function perflab_render_server_timing_page_field( $slug ) {
+	$options = (array) get_option( PERFLAB_SERVER_TIMING_SETTING, array() );
+
+	// Value for the sub-key is an array of hook names.
+	$value = '';
+	if ( isset( $options[ $slug ] ) ) {
+		$value = implode( "\n", $options[ $slug ] );
+	}
+
+	$field_id       = "server_timing_{$slug}";
+	$field_name     = PERFLAB_SERVER_TIMING_SETTING . '[' . $slug . ']';
+	$description_id = "{$field_id}_description";
+
+	?>
+	<textarea
+		id="<?php echo esc_attr( $field_id ); ?>"
+		name="<?php echo esc_attr( $field_name ); ?>"
+		aria-describedby="<?php echo esc_attr( $description_id ); ?>"
+		class="large-text code"
+		rows="8"
+	><?php echo esc_textarea( $value ); ?></textarea>
+	<p id="<?php echo esc_attr( $description_id ); ?>" class="description">
+		<?php esc_html_e( 'Enter a single hook name per line.', 'performance-lab' ); ?>
+	</p>
+	<?php
+}
--- a/admin/server-timing.php
+++ b/admin/server-timing.php
@@ -94,7 +94,7 @@ function perflab_render_server_timing_page() {
 			<?php esc_html_e( 'Server-Timing', 'performance-lab' ); ?>
 		</h1>
 
-		<form action="options.php" method="post" novalidate>
+		<form action="options.php" method="post">
 			<?php settings_fields( PERFLAB_SERVER_TIMING_SCREEN ); ?>
 			<?php do_settings_sections( PERFLAB_SERVER_TIMING_SCREEN ); ?>
 			<?php submit_button(); ?>
--- a/admin/server-timing.php
+++ b/admin/server-timing.php
@@ -47,6 +47,15 @@ function perflab_load_server_timing_page() {
 					__( 'For any hook name provided, the <strong>cumulative duration between all callbacks</strong> attached to the hook is measured, in milliseconds.', 'performance-lab' ),
 					array( 'strong' => array() )
 				);
+				if ( ! perflab_server_timing_use_output_buffer() ) {
+					?>
+					<br>
+					<?php
+					echo wp_kses(
+						__( 'Since the Server-Timing header is sent before the template is loaded, only hooks before the <code>template_include</code> filter can be measured.', 'performance-lab' ),
+						array( 'code' => array() )
+					);
+				}
 				?>
 			</p>
 			<?php
--- a/admin/server-timing.php
+++ b/admin/server-timing.php
@@ -40,7 +40,16 @@ function perflab_load_server_timing_page() {
 		static function() {
 			?>
 			<p>
-				<?php esc_html_e( 'In this section, you can provide hook names to include measurements for them in the Server-Timing header.', 'performance-lab' ); ?>
+				<?php
+				echo wp_kses(
+					sprintf(
+						/* translators: %s: Server-Timing */
+						__( 'In this section, you can provide hook names to include measurements for them in the %s header.', 'performance-lab' ),
+						'<code>Server-Timing</code>'
+					),
+					array( 'code' => array() )
+				);
+				?>
 				<br>
 				<?php
 				echo wp_kses(
@@ -52,7 +61,12 @@ function perflab_load_server_timing_page() {
 					<br>
 					<?php
 					echo wp_kses(
-						__( 'Since the Server-Timing header is sent before the template is loaded, only hooks before the <code>template_include</code> filter can be measured.', 'performance-lab' ),
+						sprintf(
+							/* translators: 1: Server-Timing, 2: template_include */
+							__( 'Since the %1$s header is sent before the template is loaded, only hooks before the %2$s filter can be measured.', 'performance-lab' ),
+							'<code>Server-Timing</code>',
+							'<code>template_include</code>'
+						),
 						array( 'code' => array() )
 					);
 				}
--- a/admin/server-timing.php
+++ b/admin/server-timing.php
@@ -41,6 +41,13 @@ function perflab_load_server_timing_page() {
 			?>
 			<p>
 				<?php esc_html_e( 'In this section, you can provide hook names to include measurements for them in the Server-Timing header.', 'performance-lab' ); ?>
+				<br>
+				<?php
+				echo wp_kses(
+					__( 'For any hook name provided, the <strong>cumulative duration between all callbacks</strong> attached to the hook is measured, in milliseconds.', 'performance-lab' ),
+					array( 'strong' => array() )
+				);
+				?>
 			</p>
 			<?php
 		},
--- a/admin/server-timing.php
+++ b/admin/server-timing.php
@@ -94,7 +94,7 @@ function perflab_render_server_timing_page() {
 			<?php esc_html_e( 'Server-Timing', 'performance-lab' ); ?>
 		</h1>
 
-		<form action="options.php" method="post" novalidate="novalidate">
+		<form action="options.php" method="post" novalidate>
 			<?php settings_fields( PERFLAB_SERVER_TIMING_SCREEN ); ?>
 			<?php do_settings_sections( PERFLAB_SERVER_TIMING_SCREEN ); ?>
 			<?php submit_button(); ?>
--- a/server-timing/defaults.php
+++ b/server-timing/defaults.php
@@ -223,63 +223,62 @@ add_action( 'wp_loaded', 'perflab_register_default_server_timing_template_metric
 function perflab_register_additional_server_timing_metrics_from_setting() {
 	$options = (array) get_option( PERFLAB_SERVER_TIMING_SETTING, array() );
 
+	/*
+	 * This closure measures performance of a hook (action or filter), as follows:
+	 *
+	 * 1. Add a hook callback at the minimum (i.e. earliest) priority possible.
+	 * 2. In that callback, register the metric for the hook, with a prefix of either "action" or "filter".
+	 * 3. Provide a measuring callback which captures the time span between beginning to end of the hook:
+	 *     1. Capture the current time immediately, i.e. at the earliest hook priority.
+	 *     2. Add another hook callback at the maximum (i.e. latest) priority possible.
+	 *     3. In that callback, capture the current time, leading the Server-Timing API to calculate the difference.
+	 *
+	 * Parameters to this closure are the hook name, what type of hook it is (either "action" or "filter", used as a
+	 * prefix for the registered metric), and the callback function to add the hook (either "add_action" or
+	 * "add_filter").
+	 */
+	$measure_hook = static function( $hook_name, $hook_type, $add_hook_func ) {
+		$metric_slug      = "{$hook_type}-{$hook_name}";
+		$measure_callback = static function( $metric ) use ( $hook_name, $add_hook_func ) {
+			$metric->measure_before();
+			call_user_func(
+				$add_hook_func,
+				$hook_name,
+				static function( $passthrough = null ) use ( $metric ) {
+					$metric->measure_after();
+					return $passthrough;
+				},
+				PHP_INT_MAX
+			);
+		};
+
+		call_user_func(
+			$add_hook_func,
+			$hook_name,
+			static function( $passthrough = null ) use ( $metric_slug, $measure_callback ) {
+				perflab_server_timing_register_metric(
+					$metric_slug,
+					array(
+						'measure_callback' => $measure_callback,
+						'access_cap'       => 'exist',
+					)
+				);
+				return $passthrough;
+			},
+			// phpcs:ignore PHPCompatibility.Constants.NewConstants.php_int_minFound
+			defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : ~PHP_INT_MAX
+		);
+	};
+
 	if ( isset( $options['benchmarking_actions'] ) ) {
 		foreach ( $options['benchmarking_actions'] as $action ) {
-			$metric_slug = 'action-' . $action;
-
-			$measure_callback = static function( $metric ) use ( $action ) {
-				$metric->measure_before();
-				add_action( $action, array( $metric, 'measure_after' ), PHP_INT_MAX, 0 );
-			};
-
-			add_action(
-				$action,
-				static function() use ( $metric_slug, $measure_callback ) {
-					perflab_server_timing_register_metric(
-						$metric_slug,
-						array(
-							'measure_callback' => $measure_callback,
-							'access_cap'       => 'exist',
-						)
-					);
-				},
-				// phpcs:ignore PHPCompatibility.Constants.NewConstants.php_int_minFound
-				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : ~PHP_INT_MAX
-			);
+			$measure_hook( $action, 'action', 'add_action' );
 		}
 	}
 
 	if ( isset( $options['benchmarking_filters'] ) ) {
 		foreach ( $options['benchmarking_filters'] as $filter ) {
-			$metric_slug = 'filter-' . $filter;
-
-			$measure_callback = static function( $metric ) use ( $filter ) {
-				$metric->measure_before();
-				add_filter(
-					$filter,
-					static function( $passthrough ) use ( $metric ) {
-						$metric->measure_after();
-						return $passthrough;
-					},
-					PHP_INT_MAX
-				);
-			};
-
-			add_filter(
-				$filter,
-				static function( $passthrough ) use ( $metric_slug, $measure_callback ) {
-					perflab_server_timing_register_metric(
-						$metric_slug,
-						array(
-							'measure_callback' => $measure_callback,
-							'access_cap'       => 'exist',
-						)
-					);
-					return $passthrough;
-				},
-				// phpcs:ignore PHPCompatibility.Constants.NewConstants.php_int_minFound
-				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : ~PHP_INT_MAX
-			);
+			$measure_hook( $filter, 'filter', 'add_filter' );
 		}
 	}
 }
--- a/server-timing/defaults.php
+++ b/server-timing/defaults.php
@@ -240,7 +240,7 @@ function perflab_register_additional_server_timing_metrics_from_setting() {
 					);
 				},
 				// phpcs:ignore PHPCompatibility.Constants.NewConstants.php_int_minFound
-				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : -9999
+				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : ~PHP_INT_MAX
 			);
 		}
 	}
@@ -274,7 +274,7 @@ function perflab_register_additional_server_timing_metrics_from_setting() {
 					return $passthrough;
 				},
 				// phpcs:ignore PHPCompatibility.Constants.NewConstants.php_int_minFound
-				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : -9999
+				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : ~PHP_INT_MAX
 			);
 		}
 	}
--- a/server-timing/defaults.php
+++ b/server-timing/defaults.php
@@ -223,7 +223,7 @@ function perflab_register_additional_server_timing_metrics_from_setting() {
 		foreach ( $options['benchmarking_actions'] as $action ) {
 			$metric_slug = 'action-' . $action;
 
-			$measure_callback = function( $metric ) use ( $action ) {
+			$measure_callback = static function( $metric ) use ( $action ) {
 				$metric->measure_before();
 				add_action( $action, array( $metric, 'measure_after' ), PHP_INT_MAX, 0 );
 			};
@@ -239,6 +239,7 @@ function perflab_register_additional_server_timing_metrics_from_setting() {
 						)
 					);
 				},
+				// phpcs:ignore PHPCompatibility.Constants.NewConstants.php_int_minFound
 				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : -9999
 			);
 		}
@@ -248,7 +249,7 @@ function perflab_register_additional_server_timing_metrics_from_setting() {
 		foreach ( $options['benchmarking_filters'] as $filter ) {
 			$metric_slug = 'filter-' . $filter;
 
-			$measure_callback = function( $metric ) use ( $filter ) {
+			$measure_callback = static function( $metric ) use ( $filter ) {
 				$metric->measure_before();
 				add_filter(
 					$filter,
@@ -272,6 +273,7 @@ function perflab_register_additional_server_timing_metrics_from_setting() {
 					);
 					return $passthrough;
 				},
+				// phpcs:ignore PHPCompatibility.Constants.NewConstants.php_int_minFound
 				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : -9999
 			);
 		}
--- a/server-timing/defaults.php
+++ b/server-timing/defaults.php
@@ -207,3 +207,84 @@ function perflab_register_default_server_timing_template_metrics() {
 	}
 }
 add_action( 'wp_loaded', 'perflab_register_default_server_timing_template_metrics' );
+
+/**
+ * Registers additional Server-Timing metrics as configured in the setting.
+ *
+ * These metrics should be registered as soon as possible. They can be added
+ * and modified in the "Tools > Server-Timing" screen.
+ *
+ * @since n.e.x.t
+ */
+function perflab_register_additional_server_timing_metrics_from_setting() {
+	$options = (array) get_option( PERFLAB_SERVER_TIMING_SETTING, array() );
+
+	if ( isset( $options['benchmarking_actions'] ) ) {
+		foreach ( $options['benchmarking_actions'] as $action ) {
+			$metric_slug = 'action-' . $action;
+
+			$measure_callback = function( $metric ) use ( $action ) {
+				$metric->measure_before();
+				add_action( $action, array( $metric, 'measure_after' ), PHP_INT_MAX, 0 );
+			};
+
+			add_action(
+				$action,
+				static function() use ( $metric_slug, $measure_callback ) {
+					perflab_server_timing_register_metric(
+						$metric_slug,
+						array(
+							'measure_callback' => $measure_callback,
+							'access_cap'       => 'exist',
+						)
+					);
+				},
+				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : -9999
+			);
+		}
+	}
+
+	if ( isset( $options['benchmarking_filters'] ) ) {
+		foreach ( $options['benchmarking_filters'] as $filter ) {
+			$metric_slug = 'filter-' . $filter;
+
+			$measure_callback = function( $metric ) use ( $filter ) {
+				$metric->measure_before();
+				add_filter(
+					$filter,
+					static function( $passthrough ) use ( $metric ) {
+						$metric->measure_after();
+						return $passthrough;
+					},
+					PHP_INT_MAX
+				);
+			};
+
+			add_filter(
+				$filter,
+				static function( $passthrough ) use ( $metric_slug, $measure_callback ) {
+					perflab_server_timing_register_metric(
+						$metric_slug,
+						array(
+							'measure_callback' => $measure_callback,
+							'access_cap'       => 'exist',
+						)
+					);
+					return $passthrough;
+				},
+				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : -9999
+			);
+		}
+	}
+}
+
+/*
+ * If this file is loaded from the Server-Timing logic in the object-cache.php
+ * drop-in, it must not call this function right away since otherwise the cache
+ * will not be loaded yet.
+ */
+if ( ! did_action( 'muplugins_loaded' ) ) {
+	add_action( 'muplugins_loaded', 'perflab_register_additional_server_timing_metrics_from_setting' );
+} else {
+	perflab_register_additional_server_timing_metrics_from_setting();
+}
--- a/server-timing/defaults.php
+++ b/server-timing/defaults.php
@@ -276,21 +276,18 @@ function perflab_register_additional_server_timing_metrics_from_setting() {
 				$metric->measure_before();
 
 				if ( 'action' === $hook_type ) {
-					add_action(
-						$hook_name,
-						array( $metric, 'measure_after' ),
-						PHP_INT_MAX,
-						0
-					);
+					$cb = static function() use ( $metric, $hook_name, &$cb ) {
+						$metric->measure_after();
+						remove_action( $hook_name, $cb, PHP_INT_MAX );
+					};
+					add_action( $hook_name, $cb, PHP_INT_MAX );
 				} else {
-					add_filter(
-						$hook_name,
-						static function( $passthrough ) use ( $metric ) {
-							$metric->measure_after();
-							return $passthrough;
-						},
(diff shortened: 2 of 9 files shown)


admin/server-timing.php
```diff
@@ -0,0 +1,131 @@
...
+		PERFLAB_SERVER_TIMING_SCREEN,
+		'perflab_render_server_timing_page'
+	);
+
+	// Add the following hooks only if the screen was successfully added.
+	if ( false !== $hook_suffix ) {
+		add_action( "load-{$hook_suffix}", 'perflab_load_server_timing_page' );
+	}
+
+	return $hook_suffix;
```
spacedmonkey, line 31: What are you returning this for? This is hooked into an action right?

server-timing/load.php
```diff
@@ -6,6 +6,54 @@
...
+			static function( $hooks ) {
+				if ( ! is_array( $hooks ) ) {
+					$hooks = explode( "\n", $hooks );
+				}
+				return array_filter( array_map( 'sanitize_key', $hooks ) );
+			},
+			$value
+		)
+	);
+}
```
spacedmonkey: This could do with a unit test IMO.

server-timing/load.php
```diff
@@ -6,6 +6,54 @@
...
+function perflab_register_server_timing_setting() {
+	register_setting(
+		PERFLAB_SERVER_TIMING_SCREEN,
+		PERFLAB_SERVER_TIMING_SETTING,
+		array(
+			'type'              => 'object',
+			'sanitize_callback' => 'perflab_sanitize_server_timing_setting',
+			'default'           => array(),
+		)
+	);
```
spacedmonkey: Not required, but I like to make settings available in the REST API if I am registering the settings anyway. 
```
'show_in_rest' => true,
```

server-timing/defaults.php
```diff
@@ -207,3 +207,84 @@ function( $acc, $query ) {
...
+						array(
+							'measure_callback' => $measure_callback,
+							'access_cap'       => 'exist',
+						)
+					);
+					return $passthrough;
+				},
+				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : -9999
+			);
+		}
```
spacedmonkey: This section is a little hard to follow, some comments might be useful here.

server-timing/defaults.php
```diff
@@ -207,3 +207,84 @@ function( $acc, $query ) {
...
+	}
+
+	if ( isset( $options['benchmarking_filters'] ) ) {
+		foreach ( $options['benchmarking_filters'] as $filter ) {
+			$metric_slug = 'filter-' . $filter;
+
+			$measure_callback = function( $metric ) use ( $filter ) {
```
spacedmonkey: This is a callback to finish the measurements right? Maybe naming the variable something now clear would make this more readable.

server-timing/defaults.php
```diff
@@ -207,3 +207,84 @@ function( $acc, $query ) {
...
+					perflab_server_timing_register_metric(
+						$metric_slug,
+						array(
+							'measure_callback' => $measure_callback,
+							'access_cap'       => 'exist',
+						)
+					);
+					return $passthrough;
+				},
+				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : -9999
```
spacedmonkey:
```suggestion
				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : -9223372036854775808
```

Should this be like this?

spacedmonkey: Also a comment saying this can be removed once we change to only support PHP 7.0+

server-timing/defaults.php
```diff
@@ -207,3 +207,84 @@ function( $acc, $query ) {
...
+function perflab_register_additional_server_timing_metrics_from_setting() {
+	$options = (array) get_option( PERFLAB_SERVER_TIMING_SETTING, array() );
+
+	if ( isset( $options['benchmarking_actions'] ) ) {
+		foreach ( $options['benchmarking_actions'] as $action ) {
+			$metric_slug = 'action-' . $action;
+
+			$measure_callback = function( $metric ) use ( $action ) {
+				$metric->measure_before();
+				add_action( $action, array( $metric, 'measure_after' ), PHP_INT_MAX, 0 );
```
spacedmonkey: As add_action is just a wrapper around add_filter, can we just reuse the logic below.

server-timing/defaults.php
westonruter: I don't think the min integer should be hard coded since it can differ from system to system (32-bit vs 64-bit).

Actually, this can be achieved with `~PHP_INT_MAX` which is the same as `PHP_INT_MIN`:

```suggestion
				defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : ~PHP_INT_MAX
```

admin/server-timing.php
felixarntz: Right, this would only be useful for test coverage, which I'm planning to add.

server-timing/load.php
felixarntz: Yeah, I haven't added unit tests yet, but definitely planning to do so.

server-timing/load.php
felixarntz: Not opposed, but I think if we do that we also should do it for the other setting that the PL plugin registers.

server-timing/defaults.php
felixarntz: I've updated this to go with `~PHP_INT_MAX` which in fact I even found used like that in one place in WP core.

swissspidy:
Just from looking at the screenshot alone it is not clear what exactly would be measured. Is it the time since the last hook (i.e. `do_action(wp_loaded) *start* ... *stop* do_action(init)`, or the duration of all the functions hooked into it (i.e. `*start* ... do_action( xyz ); *stop*`). Looking at https://gist.github.com/felixarntz/63c05392dbf7d51cc7f8f4a424b1ff39, it seems to be the latter, but would be nice to clarify that in the UI with a clearer description.

mukeshpanchal27 (commented):
Thanks @felixarntz for the PR, If anyone add duplicate value in Actions or Filters then it shows error.

For ex.
```
init
wp_loaded
init
```

server-timing/defaults.php
felixarntz: Thanks @spacedmonkey, I've made the logic reusable and added a long inline comment to document what it does, see https://github.com/WordPress/performance/pull/784/commits/694a6d1d4e3c968102ddce960ebbb252f468ed37.

server-timing/defaults.php
felixarntz: This is now simplified, via https://github.com/WordPress/performance/pull/784/commits/694a6d1d4e3c968102ddce960ebbb252f468ed37.

server-timing/defaults.php
felixarntz: Good idea, I've made all that logic reusable now. It looks a bit more complicated now, but I've also provided additional inline documentation to explain, see https://github.com/WordPress/performance/pull/784/commits/694a6d1d4e3c968102ddce960ebbb252f468ed37.

felixarntz:
Thanks @swissspidy @mukeshpanchal27 for the feedback.
* In https://github.com/WordPress/performance/pull/784/commits/6b5d12ab1f62175042db3624428ba114dcafc8f5, I've pushed a clarification for how the hooks are measured.
* In https://github.com/WordPress/performance/pull/784/commits/61f5ccc6a391831ea886235d41c566c5fa73dc3b, I've pushed a fix to ensure no duplicate hooks can be provided.

I'll work on some test coverage now, mostly relevant for the WP Admin UI.

felixarntz:
In https://github.com/WordPress/performance/pull/784/commits/585911dbe454315298e39b2f4be1e21008966667 and https://github.com/WordPress/performance/pull/784/commits/6db1e85c4e7fc2af233ab4871b9126af8e2fb31f I've added test coverage, so this PR is now ready for a full review.

westonruter:
> * If you provide hook names that are only run after the template is being parsed (e.g. `wp_enqueue_scripts` action), they won't show up by default. You'll need to activate output buffering (e.g. via [this Gist plugin](https://gist.github.com/felixarntz/9c3d7150c74082e69bb426393b68b12e)) so that those can be exposed as well.

Would this be an opportunity to also add a UI toggle to enable output buffering by default?

admin/server-timing.php
```diff
@@ -0,0 +1,138 @@
...
+ */
+function perflab_render_server_timing_page() {
+	?>
+	<div class="wrap">
+		<?php settings_errors(); ?>
+		<h1>
+			<?php esc_html_e( 'Server-Timing', 'performance-lab' ); ?>
+		</h1>
+
+		<form action="options.php" method="post" novalidate="novalidate">
```
westonruter: No need to use XHTML syntax as the admin is HTML5.

```suggestion
		<form action="options.php" method="post" novalidate>
```

But why `novalidate` in the first place? The UI could listen for `change` events on the `textarea` and then check for valid input to then call `setCustomValidity()` with `null` if it is valid or else an error message string which would then show up in the UI when the user tries to submit.

server-timing/defaults.php
```diff
@@ -211,3 +211,85 @@ static function( $acc, $query ) {
...
+					$metric_slug,
+					array(
+						'measure_callback' => $measure_callback,
+						'access_cap'       => 'exist',
+					)
+				);
+				return $passthrough;
+			},
+			// phpcs:ignore PHPCompatibility.Constants.NewConstants.php_int_minFound
+			defined( 'PHP_INT_MIN' ) ? PHP_INT_MIN : ~PHP_INT_MAX
```
westonruter: This may not catch an action which is running at `PHP_INT_MIN`. There is an option to actually go even earlier, and that is to use the `all` hook. This could also solve the problem of catching hooks that already run at `PHP_INT_MAX`, since `perflab_register_additional_server_timing_metrics_from_setting()` is running at or near `muplugins_loaded` and thus any actions added later at `PHP_INT_MAX` will run afterward. The callback can be added in the `all` hook _just in time_ so it is assured to be the very last last `PHP_INT_MAX` callback priority since `WP_Hook::add_filter()` will pushes onto the array.

Here is an example of what I understand can be problematic with the current approach:

```php
<?php
// Begin Performance code.
$hook_start_time = 0.0;

add_action(
	'foo',
	 function ()  {
		global $hook_start_time;
		echo "Starting capturing foo action timing\n";
		$hook_start_time = microtime( true );
	},
	PHP_INT_MIN
);

add_action(
	'foo',
	 function ()  {
		global $hook_start_time;
		echo "Ending capturing foo action timing\n";
		printf( "Hook took %f microseconds.\n", microtime( true ) - $hook_start_time );
	},
	PHP_INT_MAX
);

// Begin userland code.
add_action( 'foo', function () {
	echo "Sleeping for 100 microseconds\n";
	usleep( 100 );
} );

// Note: This action callback will not get measured!
add_action( 'foo', function () {
	echo "Sleeping for 300 microseconds\n";
	usleep( 300 );
}, PHP_INT_MAX );

echo "Before do_action foo\n";
do_action( 'foo' );
echo "After do_action foo\n";
```

This outputs:

```
Before do_action foo
Starting capturing foo action timing
Sleeping for 100 microseconds
Ending capturing foo action timing
Hook took 0.000165 microseconds.
Sleeping for 300 microseconds
After do_action foo
```

Note that the 300 microsecond callback does _not_ get included in the measurement.

Now consider this alternative that uses the `all` hook:

```php
<?php
// Begin Performance code.
$hook_start_time = 0.0;

add_filter(
	'all',
	function ( $hook_name ) {
		if ( 'foo' !== $hook_name ) {
			return;
		}

		$hook_start_time = microtime( true );
		echo "Starting capturing foo action timing\n";

		$callback = function ( $passthrough ) use ( $hook_name, $hook_start_time, &$callback ) {
			remove_filter( $hook_name, $callback, PHP_INT_MAX ); // Remove myself.

			echo "Ending capturing foo action timing\n";
			printf( "Hook took %f microseconds.\n", microtime( true ) - $hook_start_time );
			return $passthrough;
		};

		add_filter( $hook_name, $callback, PHP_INT_MAX ); // Remove myself.
	},
	PHP_INT_MIN
);

// Begin userland code.
add_action( 'foo', function () {
	echo "Sleeping for 100 microseconds\n";
	usleep( 100 );
} );

add_action( 'foo', function () {
	echo "Sleeping for 300 microseconds\n";
	usleep( 300 );
}, PHP_INT_MAX );

echo "Before do_action foo\n";
do_action( 'foo' );
echo "After do_action foo\n";
```

The output here is as follows:

```
Before do_action foo
Starting capturing foo action timing
Sleeping for 100 microseconds
Sleeping for 300 microseconds
Ending capturing foo action timing
Hook took 0.000539 microseconds.
After do_action foo
```

Note that even the later-added user callback to the `foo` action which had a `PHP_INT_MAX` priority is also included in the measured hooks.

server-timing/load.php
```diff
@@ -10,6 +10,9 @@
 	exit; // Exit if accessed directly.
 }
 
+define( 'PERFLAB_SERVER_TIMING_SETTING', 'perflab_server_timing_settings' );
+define( 'PERFLAB_SERVER_TIMING_SCREEN', 'perflab-server-timing' );
```
westonruter, line 14: Would it make sense to start using `const`? We get better typing.

```suggestion
const PERFLAB_SERVER_TIMING_SETTING = 'perflab_server_timing_settings';
const PERFLAB_SERVER_TIMING_SCREEN = 'perflab-server-timing';
```

PhpStorm complains about the use of `define()`. Using `const` was introduced in PHP 5.3.

server-timing/load.php
```diff
@@ -113,3 +116,57 @@ function perflab_wrap_server_timing( $callback, $metric_slug, $access_cap ) {
...
+	return array_filter(
+		array_map(
+			static function( $hooks ) {
+				if ( ! is_array( $hooks ) ) {
+					$hooks = explode( "\n", $hooks );
+				}
+				return array_values(
+					array_unique(
+						array_filter(
+							array_map( 'sanitize_key', $hooks )
```
westonruter: I don't think that `sanitize_key` will work here because hooks can contain non-key characters, even arbitrary characters since hook names often can include variables. Here are some examples from Human Made repos:

[`do_action( 'mercator.mapping.made_primary', $this );`](https://github.com/humanmade/Mercator/blob/9240e7cce9882a9fe452ced8a5cf7845d60c5a2e/class-mapping.php#L138)
[`do_action( 'hm_parse_request_' . $this->get_regex(), $wp );`](https://github.com/humanmade/hm-rewrite/blob/75e6bf7a81f56c8d64a84ff07e4da334ccb4350f/hm-rewrites.php#L187)
[`do_action( 'amf/loaded' );`](https://github.com/humanmade/asset-manager-framework/blob/58e37f495638a40ce6ddab10b400d702cb4883bc/inc/namespace.php#L70)

admin/server-timing.php
```diff
@@ -0,0 +1,138 @@
...
+	$field_name     = PERFLAB_SERVER_TIMING_SETTING . '[' . $slug . ']';
+	$description_id = "{$field_id}_description";
+
+	?>
+	<textarea
+		id="<?php echo esc_attr( $field_id ); ?>"
+		name="<?php echo esc_attr( $field_name ); ?>"
+		aria-describedby="<?php echo esc_attr( $description_id ); ?>"
+		class="large-text code"
+		rows="8"
```
westonruter, line 159: Without `novalidate`, custom validation could be added, for example:

```suggestion
		rows="8"
		onchange="this.setCustomValidity( /^\S+((\r?\n)+\S+)*\s*$/s.test( this.value ) ? '' : <?php echo esc_attr( wp_json_encode( __( 'Please enter one hook per line.', 'performance-lab' ) ) ); ?> )"
```

Nevertheless, as noted in the comment about `sanitize_key()`, I don't know actually if there is any validation that can be done since hook names can be basically anything.

westonruter (changes requested):

server-timing/defaults.php
westonruter: > The callback can be added in the `all` hook _just in time_ so it is assured to be the very last last `PHP_INT_MAX` callback priority since `WP_Hook::add_filter()` will pushes onto the array.

Actually, it seems this is not relevant since the `PHP_INT_MAX`-priority hook is being added in the `PHP_INT_MIN` callback. Nevertheless, what remains is the possibility that another existing `PHP_INT_MIN` callback may not get measured?

server-timing/defaults.php
```diff
@@ -211,3 +211,85 @@ static function( $acc, $query ) {
...
+		$metric_slug      = "{$hook_type}-{$hook_name}";
+		$measure_callback = static function( $metric ) use ( $hook_name, $add_hook_func ) {
+			$metric->measure_before();
+			call_user_func(
+				$add_hook_func,
+				$hook_name,
+				static function( $passthrough = null ) use ( $metric ) {
+					$metric->measure_after();
+					return $passthrough;
+				},
```
westonruter: Doesn't this callback need to remove itself here? Otherwise, if a hook gets called multiple times it seems there will end up being leftover callbacks at `PHP_INT_MAX` that will try to keep calling `measure_after` on the previous `$metric`.

admin/server-timing.php
felixarntz: I don't think the validation here is critical since it's sanitized later anyway, so implementing some validation would open a new topic of discussion on what it should be like. I don't think core's settings typically provide client-side validation, so I don't think we need to here either.

Regarding the XHTML syntax, I'll remove that.

server-timing/defaults.php
felixarntz: Thanks @westonruter, the `all` approach looks a bit more reliable indeed. In any case, the described situations are edge-cases, so I don't think they are super critical to cover. As in, I'll look into implementing the `all` approach but the fact that even that still has some limitations IMO is perfectly fine.

server-timing/load.php
felixarntz: Hmm, not really common in WP land I guess? I don't feel strongly about it, but we use `define` everywhere else in PL, so I think this should rather be holistically discussed (and potentially changed) in another issue.

server-timing/load.php
felixarntz: That's a good point. I'm unsure how far we should go with validation to satisfy all sorts of custom cases, arguably the `sanitize_key()` function does work at least for how core uses hooks. I'm thinking maybe I'll change it to a custom regex that also allows for some other "common" hook characters like "/", "." for example. I don't think we need to cover everything here, validation can always be loosened later if needed. Better to be strict and loosen up later rather than the other way around.

admin/server-timing.php
felixarntz: See my reply on that comment https://github.com/WordPress/performance/pull/784#discussion_r1286036347

admin/server-timing.php
felixarntz: Done in https://github.com/WordPress/performance/pull/784/commits/a8d141f594672d3d1aa03e9b2e79b41d9cabdd0b

server-timing/defaults.php
felixarntz: I modified the logic to use `all` in https://github.com/WordPress/performance/pull/784/commits/400c0a152b95dca374b9581e15ac4264e953ff9a.

Note that for now only hooks that run once are measured, or for hooks that run multiple times only the first occurrence is measured. That is something we can change in the future, but it would also require a few enhancements to even the underlying Server-Timing API, so it's out of scope for this. I have included logic though to ensure that hooks that run multiple times do not cause errors (e.g. due to duplicate metric registration), so it was also a good callout from that perspective.

server-timing/load.php
felixarntz: I implemented a custom regex in https://github.com/WordPress/performance/pull/784/commits/c5fe384ece17f73a7e62837a2b8db3a290ca8a68. As per my previous comment, I acknowledge this probably doesn't encompass _all_ possible hook names, but given there seem to be no real technical limitations, I'm not even sure supporting _all_ possible hook names is a good idea. I _did_ add support for the special characters from your examples, as I believe they are somewhat common in the plugin ecosystem. We can always add more later if needed.

admin/server-timing.php
westonruter: Since there is no HTML validation constraints, then there's no need for the `novalidate` attribute at all. It's only applicable if you want to bypass the browser from doing client-side validation. But since none is present, it doesn't do anything.

server-timing/load.php
westonruter: Sounds good

server-timing/defaults.php
felixarntz: Done in https://github.com/WordPress/performance/pull/784/commits/447f078523412aa5184b75f04b28dfe6da9dc5a3

server-timing/load.php
```diff
@@ -113,3 +116,66 @@ function perflab_wrap_server_timing( $callback, $metric_slug, $access_cap ) {
...
+				}
+				return array_values(
+					array_unique(
+						array_filter(
+							array_map(
+								static function( $hookname ) {
+									return preg_replace(
+										'/[^A-Za-z0-9_\-\/\.]/',
+										'',
+										strtolower( $hookname )
```
westonruter: Hooks can involve upper-case letters: https://wpdirectory.net/search/01H78FM13RS16AKX0G7K06N2JE

```suggestion
										$hookname
```

westonruter: Also, probably good to add `trim()`:

```suggestion
										trim( $hookname )
```

admin/server-timing.php
westonruter: If the sanitization is removing all `[^A-Za-z0-9_\-\/\.]` characters, then adding this client-side validation would be important. Otherwise, the user could add such a custom hook and it would get silently munged when saving. Client-side validation would prevent that from happening. This should do it, along with removal of `novalidate`:

```suggestion
		rows="8"
		onchange="this.setCustomValidity( /^\s*([A-Za-z0-9_\-\/.]+\s*)*\s*$/s.test( this.value ) ? '' : <?php echo esc_attr( wp_json_encode( __( 'Please enter one hook per line and only use these characters: [A-Za-z0-9_\\-/\\.].', 'performance-lab' ) ) ); ?> )"
```

server-timing/load.php
```diff
@@ -113,3 +116,66 @@ function perflab_wrap_server_timing( $callback, $metric_slug, $access_cap ) {
...
+									);
+								},
+								$hooks
+							)
```
westonruter: Would this be simpler and less prone to munge complex hook names?

```suggestion
							array_map( 'sanitize_text_field', $hooks )
```

(Note: It also does `trim()` automatically.)

admin/server-timing.php
felixarntz: Per my reply in https://github.com/WordPress/performance/pull/784#discussion_r1286036347, I'd prefer not to get hung up on that here, since it's not common in core either. There's tons of sanitization in core and little to no validation in most places.

Not saying this is something we can't improve, but client-side validation is by no means critical, the server-side sanitization however is. I'd prefer if we considered/discussed this separately.

admin/server-timing.php
felixarntz: You're probably right, but aren't there some browser defaults that could still trigger client-side validation? I wonder why for example WP core then uses `novalidate` in `options-general.php`.

server-timing/defaults.php
westonruter: Humm, yeah. Being able to capture server timing for multiple hook runs seems important, doesn't it? 

Here are the hooks which are called more than once in a logged-out visit to the homepage on a vanilla site (via [Gist](https://gist.github.com/westonruter/fbf22063a74c7844c85ee2a8013b934e)):

<details><summary>Table of Hook Counts</summary>

Hook | Count
-- | --:
gettext | 1,507
gettext_default | 1,503
gettext_with_context | 1,231
gettext_with_context_default | 1,142
pre_wp_load_alloptions | 795
alloptions | 795
pre_option | 793
pre_option_stylesheet | 393
option_stylesheet | 393
stylesheet | 389
set_url_scheme | 293
should_load_separate_core_block_assets | 236
theme_root | 206
stylesheet_directory | 185
attribute_escape | 170
pre_option_siteurl | 162
option_siteurl | 162
site_url | 147
includes_url | 122
plugins_url | 113
sanitize_key | 110
block_type_metadata | 92
register_block_type_args | 92
block_type_metadata_settings | 91
gettext_with_context_twentytwentythree | 89
block_parser_class | 61
esc_html | 56
render_block | 51
safe_style_css | 51
pre_render_block | 48
render_block_data | 48
render_block_context | 48
pre_option_timezone_string | 42
option_timezone_string | 42
safecss_filter_attr_allow_css | 42
default_template_types | 41
sanitize_title | 35
clean_url | 33
register_post_type_args | 28
registered_post_type | 28
query | 25
pre_option_home | 22
option_home | 22
wp_register_sidebar_widget | 22
parse_tax_query | 22
home_url | 21
pre_option_gmt_offset | 21
option_gmt_offset | 21
pre_option_template | 19
option_template | 19
template | 19
is_post_type_viewable | 19
pre_option_permalink_structure | 15
option_permalink_structure | 15
register_taxonomy_args | 14
registered_taxonomy | 14
get_term | 14
block_default_classname | 14
no_texturize_tags | 14
no_texturize_shortcodes | 14
locale | 13
pre_option_sticky_posts | 13
option_sticky_posts | 13
kses_allowed_protocols | 12
sanitize_text_field | 12
content_url | 12
theme_root_uri | 12
stylesheet_directory_uri | 12
wp_lazy_loading_enabled | 12
template_directory | 11
parse_query | 11
pre_get_posts | 11
wp_allow_query_attachment_by_filename | 11
pre_option_comments_per_page | 11
option_comments_per_page | 11
posts_search | 11
get_terms_defaults | 11
parse_term_query | 11
pre_get_terms | 11
get_terms_args | 11
get_terms_orderby | 11
list_terms_exclusions | 11
get_meta_sql | 11
get_terms_fields | 11
terms_clauses | 11
terms_pre_query | 11
posts_where | 11
posts_join | 11
posts_where_paged | 11
posts_groupby | 11
posts_join_paged | 11
posts_orderby | 11
posts_distinct | 11
post_limits | 11
posts_fields | 11
posts_clauses | 11
posts_selection | 11
posts_where_request | 11
posts_groupby_request | 11
posts_join_request | 11
posts_orderby_request | 11
posts_distinct_request | 11
posts_fields_request | 11
post_limits_request | 11
posts_clauses_request | 11
posts_request | 11
posts_pre_query | 11
posts_results | 11
the_posts | 11
theme_file_uri | 11
theme_file_path | 10
admin_url | 10
wp_parse_str | 10
pre_kses | 10
pre_term_name | 9
pre_wp_theme_name | 9
get_wp_theme | 9
split_the_query | 9
pre_option_show_on_front | 9
option_show_on_front | 9
render_block_ | 9
pre_option_cron | 8
option_cron | 8
default_wp_template_part_areas | 8
get_post_metadata | 8
default_post_metadata | 8
get_the_terms | 8
pre_get_scheduled_event | 7
current_theme_supports-block-templates | 7
map_meta_cap | 6
user_has_cap | 6
pre_option_use_smilies | 6
option_use_smilies | 6
update_post_metadata_cache | 6
post_thumbnail_id | 6
get_user_metadata | 6
content_pagination | 6
the_title | 6
post_password_required | 6
pre_determine_locale | 5
sanitize_locale_name | 5
determine_locale | 5
pre_get_block_templates | 5
get_block_templates | 5
posts_request_ids | 5
get_category | 5
pre_option_blogname | 5
option_blogname | 5
render_block_core/group | 5
pre_option_blog_charset | 4
option_blog_charset | 4
rest_url | 4
ngettext | 4
ngettext_default | 4
pre_option_date_format | 4
option_date_format | 4
pre_site_option_can_compress_scripts | 4
default_site_option_can_compress_scripts | 4
pre_option_can_compress_scripts | 4
option_can_compress_scripts | 4
site_option_can_compress_scripts | 4
rest_url_prefix | 4
gettext_twentytwentythree | 4
perflab_server_timing_use_output_buffer | 4
show_admin_bar | 4
the_post | 4
pre_post_link | 4
post_link | 4
wp_date | 4
get_post_time | 4
get_the_date | 4
sanitize_html_class | 4
wp_resource_hints | 4
default_feed | 4
script_loader_src | 4
pre_option_theme_mods_twentytwentythree | 3
option_theme_mods_twentytwentythree | 3
nonce_user_logged_out | 3
nonce_life | 3
salt | 3
pre_option_blog_public | 3
option_blog_public | 3
wp_sitemaps_add_provider | 3
wp_theme_json_data_theme | 3
register_setting_args | 3
register_setting | 3
pre_option_html_type | 3
option_html_type | 3
wp_should_replace_insecure_home_url | 3
render_block_core/spacer | 3
render_block_core/paragraph | 3
bloginfo | 3
register_category_taxonomy_args | 2
taxonomy_labels_category | 2
registered_taxonomy_category | 2
register_post_tag_taxonomy_args | 2
taxonomy_labels_post_tag | 2
registered_taxonomy_post_tag | 2
register_nav_menu_taxonomy_args | 2
taxonomy_labels_nav_menu | 2
registered_taxonomy_nav_menu | 2
register_link_category_taxonomy_args | 2
taxonomy_labels_link_category | 2
registered_taxonomy_link_category | 2
register_post_format_taxonomy_args | 2
taxonomy_labels_post_format | 2
registered_taxonomy_post_format | 2
register_wp_theme_taxonomy_args | 2
taxonomy_labels_wp_theme | 2
registered_taxonomy_wp_theme | 2
register_wp_template_part_area_taxonomy_args | 2
taxonomy_labels_wp_template_part_area | 2
registered_taxonomy_wp_template_part_area | 2
register_post_post_type_args | 2
post_type_labels_post | 2
registered_post_type_post | 2
register_page_post_type_args | 2
post_type_labels_page | 2
registered_post_type_page | 2
register_attachment_post_type_args | 2
post_type_labels_attachment | 2
registered_post_type_attachment | 2
register_revision_post_type_args | 2
post_type_labels_revision | 2
registered_post_type_revision | 2
register_nav_menu_item_post_type_args | 2
post_type_labels_nav_menu_item | 2
registered_post_type_nav_menu_item | 2
register_custom_css_post_type_args | 2
post_type_labels_custom_css | 2
registered_post_type_custom_css | 2
register_customize_changeset_post_type_args | 2
post_type_labels_customize_changeset | 2
registered_post_type_customize_changeset | 2
register_oembed_cache_post_type_args | 2
post_type_labels_oembed_cache | 2
registered_post_type_oembed_cache | 2
register_user_request_post_type_args | 2
post_type_labels_user_request | 2
registered_post_type_user_request | 2
register_wp_block_post_type_args | 2
post_type_labels_wp_block | 2
registered_post_type_wp_block | 2
register_wp_template_post_type_args | 2
post_type_labels_wp_template | 2
registered_post_type_wp_template | 2
register_wp_template_part_post_type_args | 2
post_type_labels_wp_template_part | 2
registered_post_type_wp_template_part | 2
register_wp_global_styles_post_type_args | 2
post_type_labels_wp_global_styles | 2
registered_post_type_wp_global_styles | 2
register_wp_navigation_post_type_args | 2
post_type_labels_wp_navigation | 2
registered_post_type_wp_navigation | 2
plugin_loaded | 2
wp_video_extensions | 2
pre_option_perflab_modules_settings | 2
option_perflab_modules_settings | 2
perflab_active_modules | 2
pre_option_category_base | 2
option_category_base | 2
pre_option_tag_base | 2
option_tag_base | 2
pre_option_start_of_week | 2
option_start_of_week | 2
pre_option_time_format | 2
option_time_format | 2
wp_theme_json_data_default | 2
wp_theme_json_data_blocks | 2
wp_theme_json_data_user | 2
pre_option_rewrite_rules | 2
option_rewrite_rules | 2
get_terms | 2
wp_using_themes | 2
rest_queried_resource_route | 2
pre_get_shortlink | 2
get_shortlink | 2
loop_no_results | 2
render_block_core_template_part_file | 2
render_block_core/site-title | 2
render_block_core/template-part | 2
update_user_metadata_cache | 2
loop_start | 2
post_thumbnail_size | 2
post_thumbnail_html | 2
render_block_core/post-featured-image | 2
render_block_core/post-title | 2
get_the_excerpt | 2
excerpt_allowed_wrapper_blocks | 2
excerpt_allowed_blocks | 2
the_content | 2
excerpt_length | 2
excerpt_more | 2
wp_trim_words | 2
wp_trim_excerpt | 2
render_block_core/post-excerpt | 2
render_block_core/post-date | 2
render_block_core/null | 2
current_theme_supports-post-thumbnails | 2
has_post_thumbnail | 2
is_sticky | 2
post_class_taxonomies | 2
post_class | 2
render_block_core/column | 2
render_block_core/pattern | 2
wp_theme_json_get_style_nodes | 2
styles_inline_size_limit | 2
feed_content_type | 2
feed_link | 2
print_styles_array | 2
print_scripts_array | 2
script_loader_tag | 2
pre_option_wp_page_for_privacy_policy | 2
option_wp_page_for_privacy_policy | 2

</details>

server-timing/load.php
felixarntz: `sanitize_text_field` allows spaces, which I really don't think we should allow here. How about we use `preg_replace( '/\s/', '', sanitize_text_field( $value ) )`?

server-timing/load.php
felixarntz: See my reply in https://github.com/WordPress/performance/pull/784#discussion_r1286215594.

admin/server-timing.php
westonruter: It's most likely stems from the time (over a decade ago) that browsers implemented poor logic for validating email addresses: https://core.trac.wordpress.org/ticket/15080#comment:2

Since there is an `input[type=email]` on the general settings screen, it's probably why `novalidate` is added there. Nevertheless, the `novalidate` attribute is _absent_ from other settings screens, including: Writing, Reading, Discussion, Media, Permalinks, and Privacy. 

I also proposed removing `novalidate` from the comments form: https://core.trac.wordpress.org/ticket/47595

server-timing/defaults.php
felixarntz: Sure, but we can always iterate. No need to implement everything we need in a single PR, that makes things extremely complex to review. FWIW, out of all the benchmarking we've been doing over the past few months I don't recall us ever having needed to benchmark a specific hook that runs multiple times. Having something that works for "single use" hooks available is most critical, and we can follow up with further enhancements after that.

admin/server-timing.php
westonruter: Just checked and I see `novalidate` was only added to the forms that involve `input` fields of type `email` and `url`: https://github.com/WordPress/wordpress-develop/commit/0a4e8b2b7ee84efb28acc946a9a54cdd03414c28

server-timing/load.php
westonruter: Spaces could appear in hooks though, right? Humm, but it seems they're mostly typos: https://wpdirectory.net/search/01H78JM9J0FXXMR2BS0JC4T0TN

So I'm fine with stripping all whitespace from the hook names.

felixarntz: Updated in https://github.com/WordPress/performance/pull/784/commits/b85959e844956d00997748d7b2ee321b71364c8d

admin/server-timing.php
felixarntz: Updated in https://github.com/WordPress/performance/pull/784/commits/c48d206320a233ba5ef95a1c17780b1f8fbc0d9b

server-timing/load.php
```diff
@@ -113,3 +116,72 @@ function perflab_wrap_server_timing( $callback, $metric_slug, $access_cap ) {
...
+	if ( ! is_array( $value ) ) {
+		return array();
+	}
+
+	/*
+	 * Ensure that every element is an indexed array of hook names.
+	 * Any duplicates across a group of hooks are removed.
+	 */
+	return array_filter(
+		array_map(
```
westonruter: Should this be checking whether the keys are those which are allowed, i.e. `benchmarking_actions` and `benchmarking_filters`?

server-timing/load.php
```diff
@@ -113,3 +116,72 @@ function perflab_wrap_server_timing( $callback, $metric_slug, $access_cap ) {
...
+									);
+								},
+								$hooks
+							)
+						)
+					)
+				);
+			},
+			$value
+		)
```
westonruter: There is a lot of nesting happening here. Not a problem, but just makes it a bit hard on the eyes. Not a blocker.

server-timing/load.php
felixarntz: Good idea, updated in https://github.com/WordPress/performance/pull/784/commits/7f99021dd4afe5420dc126a96b6d26a8ea0bf16c

server-timing/load.php
felixarntz: Reduced the nesting a bit in https://github.com/WordPress/performance/pull/784/commits/7f99021dd4afe5420dc126a96b6d26a8ea0bf16c

westonruter:
For some reason when I add a filter to get Server-Timing for, e.g. `the_content`, I'm getting a PHP notice:

<blockquote>Notice: Function Perflab_Server_Timing::register_metric was called <strong>incorrectly</strong>. The method must be called before or during the perflab_server_timing_send_header action. Please see <a href="https://wordpress.org/documentation/article/debugging-in-wordpress/">Debugging in WordPress</a> for more information.  in /var/www/html/wp-includes/functions.php on line 5865
</blockquote>

I don't get it for when I add actions, however. 😕

felixarntz:
@westonruter That's because the `the_content` hook runs after the header is output. So hooks like that can only be used if output buffering is enabled. Maybe we open a separate PR to add a checkbox to allow controlling that as well? That would make that bit easier as well, without the need for another one-liner plugin.

westonruter:
> @westonruter That's because the `the_content` hook runs after the header is output. So hooks like that can only be used if output buffering is enabled. Maybe we open a separate PR to add a checkbox to allow controlling that as well? That would make that bit easier as well, without the need for another one-liner plugin.

Oh, of course. Same as mentioned in https://github.com/WordPress/performance/pull/784#issuecomment-1666298437.

What do you think about the presence of any Server-Timing hooks to automatically cause output buffering to be enabled automatically? That would prevent others from getting tripped up the same way I did.

felixarntz:
@westonruter I think it would be better to have an explicit setting for it. To clarify the limitation when output buffering is not enabled, I have added a message for it in https://github.com/WordPress/performance/pull/784/commits/fd41db1ebfbafe0545722dfafcf75de4ed83c565 (see also the updated screenshot in the PR description).

Once we add a setting control for it, we can consider toggling that with JS based on the checkbox.

westonruter (approved):

admin/server-timing.php
```diff
@@ -0,0 +1,147 @@
...
+ * @since n.e.x.t
+ */
+function perflab_load_server_timing_page() {
+	add_settings_section(
+		'benchmarking',
+		__( 'Benchmarking', 'performance-lab' ),
+		static function() {
+			?>
+			<p>
+				<?php esc_html_e( 'In this section, you can provide hook names to include measurements for them in the Server-Timing header.', 'performance-lab' ); ?>
```
swissspidy: Since the `Server-Timing` header is not something that should be translated, I'd recommend using `sprintf` with a placeholder here (+ a translator comment of course).

admin/server-timing.php
```diff
@@ -0,0 +1,147 @@
...
+				echo wp_kses(
+					__( 'For any hook name provided, the <strong>cumulative duration between all callbacks</strong> attached to the hook is measured, in milliseconds.', 'performance-lab' ),
+					array( 'strong' => array() )
+				);
+				if ( ! perflab_server_timing_use_output_buffer() ) {
+					?>
+					<br>
+					<?php
+					echo wp_kses(
+						__( 'Since the Server-Timing header is sent before the template is loaded, only hooks before the <code>template_include</code> filter can be measured.', 'performance-lab' ),
```
swissspidy: I'd use sprintf with a placeholder for `template_include` so that it does not get translated

swissspidy (approved):
LGTM, just two i18n notes

mukeshpanchal27 (commented):
Thanks @felixarntz for the updates.

I still reproduce https://github.com/WordPress/performance/pull/784#pullrequestreview-1549628503 error. Could you please take a look. Thanks.

felixarntz:
@mukeshpanchal27 
> I still reproduce https://github.com/WordPress/performance/pull/784#pullrequestreview-1549628503 error. Could you please take a look. Thanks.

Maybe it was because you had those settings saved from before? The actual behavior from having duplicate hooks was not changed, but I added sanitization to strip duplicate values. So can you please try to go to the Tools screen again and re-save? It should result in the duplicates being stripped.

westonruter:
> Once we add a setting control for it, we can consider toggling that with JS based on the checkbox.

I'd like to propose Output Buffering as a new module. This module would be to explore implementation of Core-43258/Core-58285 while also providing an easy way to implement the toggle. It will give us an area to explore use of output buffering to make Performance enhancements as well, such as applying LCP enhancements to arbitrary elements not rendered using the normal WordPress mechanisms (e.g. hard-coded images and page builders). cc @joemcgill @kt-12

mukeshpanchal27 (approved):
Thanks @felixarntz, I test in fresh installation and it working for me.

westonruter:
> Maybe we open a separate PR to add a checkbox to allow controlling that as well? That would make that bit easier as well, without the need for another one-liner plugin.

I'll work on this.

westonruter:
> > Maybe we open a separate PR to add a checkbox to allow controlling that as well? That would make that bit easier as well, without the need for another one-liner plugin.
> 
> I'll work on this.

See https://github.com/WordPress/performance/pull/801

PR #795, opened 2023-08-08
Allow disabling Server-Timing entirely using `PERFLAB_DISABLE_SERVER_TIMING` constant

### LINKED ISSUE #794: Server-Timing with wp-before-template is always sent
## Bug Description

Sending of this HTTP header cannot be disabled in the plugin.
It is sent unconditionally.

https://github.com/WordPress/performance/blob/b09405b885d485248cdb38a59f28e2007aed15ce/server-timing/defaults.php#L122

What do you recommend? How to stop sending this header?

## Steps to reproduce

1. Just activate the plugin

felixarntz:
## Summary

Fixes #794

Note that the branch is based on the branch from #784 since it touches a lot of related code. Once that PR is merged, this one can have its base updated to `trunk`.

## Testing

1. Set constant `PERFLAB_DISABLE_SERVER_TIMING` to `true` in your `wp-config.php`.
2. Ensure that Server-Timing header is not present in response.
3. Ensure that _Tools > Server-Timing_ WP Admin screen is not available.
4. Ensure that, on a fresh install, the PL plugin's version of the `object-cache.php` drop-in is not placed.

## Checklist

### DIFF
--- a/load.php
+++ b/load.php
@@ -316,8 +316,9 @@ add_action( 'plugins_loaded', 'perflab_load_active_and_valid_modules' );
  * This only runs in WP Admin to not have any potential performance impact on
  * the frontend.
  *
- * This function will short-circuit if the constant
- * 'PERFLAB_DISABLE_OBJECT_CACHE_DROPIN' is set as true.
+ * This function will short-circuit if at least one of the constants
+ * 'PERFLAB_DISABLE_SERVER_TIMING' or 'PERFLAB_DISABLE_OBJECT_CACHE_DROPIN' is
+ * set as true.
  *
  * @since 1.8.0
  * @since 2.1.0 No longer attempts to use two of the drop-ins together.
@@ -326,6 +327,11 @@ add_action( 'plugins_loaded', 'perflab_load_active_and_valid_modules' );
  */
 function perflab_maybe_set_object_cache_dropin() {
 	global $wp_filesystem;
+
+	// Bail if Server-Timing is disabled entirely.
+	if ( defined( 'PERFLAB_DISABLE_SERVER_TIMING' ) && PERFLAB_DISABLE_SERVER_TIMING ) {
+		return;
+	}
 
 	// Bail if disabled via constant.
 	if ( defined( 'PERFLAB_DISABLE_OBJECT_CACHE_DROPIN' ) && PERFLAB_DISABLE_OBJECT_CACHE_DROPIN ) {
--- a/admin/server-timing.php
+++ b/admin/server-timing.php
@@ -5,6 +5,15 @@
  * @package performance-lab
  */
 
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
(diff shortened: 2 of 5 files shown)


szepeviktor (approved):

load.php
```diff
@@ -327,6 +328,11 @@ function perflab_load_active_and_valid_modules() {
 function perflab_maybe_set_object_cache_dropin() {
 	global $wp_filesystem;
 
+	// Bail if Server-Timing is disabled entirely.
+	if ( defined( 'PERFLAB_DISABLE_SERVER_TIMING' ) && PERFLAB_DISABLE_SERVER_TIMING ) {
+		return;
+	}
+
```
westonruter, line 335: Why does Server-Timing being disabled also disable the object cache drop-in?

szepeviktor: The object cache drop-in measures server timing?

westonruter: I see:

https://github.com/WordPress/performance/blob/b09405b885d485248cdb38a59f28e2007aed15ce/server-timing/object-cache.copy.php#L14-L16

That being the case, couldn't `Server-Timing` just be disabled with the existing `PERFLAB_DISABLE_OBJECT_CACHE_DROPIN` constant?

swissspidy: Server-Timing can be measured with or without the drop-in though. With the drop-in we can just measure more things.
In my project I actually rely on being able to measure it without the drop-in.

westonruter (approved):

swissspidy (approved):

mukeshpanchal27 (changes requested):
Thanks for the PR @felixarntz. The Server-Timing API remains accessible through the handy mini plugin you've crafted: https://gist.github.com/felixarntz/63c05392dbf7d51cc7f8f4a424b1ff39.

For a comprehensive review, here's a rundown of the steps to reproduce the issue:

1. Introduce `define( 'PERFLAB_DISABLE_SERVER_TIMING', true );` in your config.php file.
2. Verify the deactivation of Server-Timing API.
3. Implement the Server Timing Extra mini plugin from https://gist.github.com/felixarntz/63c05392dbf7d51cc7f8f4a424b1ff39.
4. Confirm the successful reactivation of Server-Timing API.

Correct me if i miss anything here. Thanks!

felixarntz:
Thanks @mukeshpanchal27, good catch! In https://github.com/WordPress/performance/pull/795/commits/93afb037f3ee2f0253834e177ef12dff5fa0e614 I implemented a fix that ensures that, even when another plugin uses the Server-Timing API, at least the header output is still being suppressed if it is generally disabled.

This way it is ensured there are no fatal errors, but the header won't be sent.

felixarntz:
Noting that there are some unit test failures, however these are also present in `trunk` and stem from changes in the WordPress 6.3 release. See #796 for a PR to address them.

mukeshpanchal27 (approved):
Thanks @felixarntz for the changes, LGTM!

PR #796, opened 2023-08-09
Fix WebP handling when editing images based on WordPress 6.3 change

felixarntz:
## Summary

Unit tests currently fail in `trunk`, based on the 6.3 release from yesterday. There are two problems that lead to those failures:
* WordPress 6.3 makes the Fetchpriority module useless. The module is already not being loaded with that version, so we need to ensure we don't run tests in that case either.
* WordPress 6.3 actually comes with a breaking change in a low-level function which we use in the WebP Uploads module: Per https://core.trac.wordpress.org/ticket/57685 / https://core.trac.wordpress.org/changeset/55935, editing thumbnails (or all image sizes other than thumbnails) is no longer enabled by default, which is technically a breaking change 😞 

This PR brings parity with those changes:
* Handle the image editing target values `thumbnail` and `nothumb` only when editing thumbnails separately is enabled (if using WP 6.3 or greater, per the new filter).
* Skip tests that aren't relevant based on the related new WP 6.3 functionality.

## Checklist


tests/modules/images/fetchpriority/fetchpriority-tests.php
```diff
@@ -38,6 +38,10 @@ public static function tear_down_after_class() {
 	}
 
 	public function test_fetchpriority_img_tag_add_attr_based_on_context_and_loading_lazy() {
+		if ( version_compare( get_bloginfo( 'version' ), '6.3', '>=' ) ) {
```
joemcgill: These would be better placed in a `setUp()` method in this class so all tests are skipped if not applicable. See: https://docs.phpunit.de/en/9.6/incomplete-and-skipped-tests.html#skipping-tests

I also am curious if we can make use of the logic we use to determine if the module can be enabled in `can-load.php` rather than creating a specific version check?

joemcgill (changes requested):

felixarntz:
@joemcgill Turns out there is a breaking change in WordPress 6.3 😞 

Not sure if it's worth fixing since it's a very low-level function, but in any case I opened https://core.trac.wordpress.org/ticket/59040 for consideration.

modules/images/webp-uploads/image-edit.php
```diff
@@ -420,3 +420,26 @@ function webp_uploads_restore_image( $attachment_id, $data ) {
...
+function webp_uploads_image_edit_thumbnails_separately() {
+	if ( version_compare( get_bloginfo( 'version' ), '6.3', '<' ) ) {
+		return true;
+	}
+
+	/*
+	 * This filter was introduced in WordPress 6.3 with default value `false`,
+	 * thus changing the behavior of several image editing functions.
+	 */
+	return apply_filters( 'image_edit_thumbnails_separately', false );
```
mukeshpanchal27:
```suggestion
	return (bool) apply_filters( 'image_edit_thumbnails_separately', false );
```
We should cast bool like WordPress core. WDYT?

mukeshpanchal27 (commented):
Thanks @felixarntz for the PR. Left one question.

With the changes we have to update WebP Uploads plugin.

modules/images/webp-uploads/image-edit.php
```diff
@@ -420,3 +420,26 @@ function webp_uploads_restore_image( $attachment_id, $data ) {
...
+ *
+ * @return bool True if editing image thumbnails is enabled, false otherwise.
+ */
+function webp_uploads_image_edit_thumbnails_separately() {
+	if ( version_compare( get_bloginfo( 'version' ), '6.3', '<' ) ) {
+		return true;
+	}
+
+	/*
+	 * This filter was introduced in WordPress 6.3 with default value `false`,
```
swissspidy: This is already stated in the function docblock above, so a bit redundant. We should rather follow the coding standards to say „This filter was documented in xyz…“ (don‘t know the exact wording right now)

felixarntz:
@mukeshpanchal27 
> With the changes we have to update WebP Uploads plugin.

Good catch! I've prepared the WebP Uploads standalone plugin for the update in https://github.com/WordPress/performance/pull/796/commits/65b5fdc6cb7ef93e4b219b08407525f81fa807b6.

swissspidy (approved):

mukeshpanchal27 (approved):
Thanks @felixarntz for the changes. LGTM!

mukeshpanchal27:
@joemcgill Could you please take a look so we can merge this PR. Thanks!

joemcgill (approved):
The updates look good to me. I've not gotten a chance to review the breaking change in WP yet, but thanks for reporting it.

PR #801, opened 2023-08-12
Add output buffering checkbox to Server-Timing screen

westonruter:
## Summary

Implements output buffering checkbox as discussed at https://github.com/WordPress/performance/pull/784#issuecomment-1673494785.

## Relevant technical choices

A new boolean `output_buffering` array item is added to the `perflab_server_timing_settings` option. This value, defaulting to `false`, will be used if there is no `perflab_server_timing_use_output_buffer` filter present.

## Screenshots

When the option is not enabled, the note at the top of the screen remains:

image

When the option is enabled, the note at the top is removed:

image

When a `perflab_server_timing_use_output_buffer` filter is added which returns `true`, the checkbox is checked and disabled with a note about the filter being present:

image

The same is true when the filter returns `false`:

image

## Checklist


admin/server-timing.php
```diff
@@ -70,13 +70,20 @@ static function() {
...
-							/* translators: 1: Server-Timing, 2: template_include */
-							__( 'Since the %1$s header is sent before the template is loaded, only hooks before the %2$s filter can be measured.', 'performance-lab' ),
-							'<code>Server-Timing</code>',
-							'<code>template_include</code>'
+						str_replace(
+							'<a>',
+							'<a href="#server_timing_output_buffering">',
+							sprintf(
+								/* translators: 1: Server-Timing, 2: template_include */
+								__( 'Since the %1$s header is sent before the template is loaded, only hooks before the %2$s filter can be measured. Enable <a>Output Buffering</a> to measure hooks during template rendering.', 'performance-lab' ),
```
felixarntz: I don't think we need the `str_replace()` here, that's a bit hacky. You can put the regular `<a href...` into the translation string, I'd recommend to make the actual `href` value a `%3$s` placeholder. That way we can get rid of the string replacement.

admin/server-timing.php
```diff
@@ -105,12 +112,20 @@ static function() {
...
 		PERFLAB_SERVER_TIMING_SCREEN,
 		'benchmarking',
 		array( 'label_for' => 'server_timing_benchmarking_filters' )
 	);
+	add_settings_field(
+		'output_buffering',
+		__( 'Output Buffering', 'performance-lab' ),
+		'perflab_render_server_timing_page_output_buffer_checkbox',
+		PERFLAB_SERVER_TIMING_SCREEN,
+		'benchmarking',
```
felixarntz: Instead of this, I would advise to put this field into a new, standalone "Output Buffering" settings section on the screen that is displayed on top (since that puts more attention to the field). You could handle that in the same way and appearance as what the "Uploading Files" section in _Settings > Media_ looks like: The settings section heading and below the checkbox, aligned to the left of the screen.

This feedback is primarily based on looking at the screenshot in the PR description. I think bringing this field to the top, in a dedicated section, would ensure it's not missed, and a dedicated section also makes sense since it is not directly tied to the benchmarking hooks fields.

admin/server-timing.php
```diff
@@ -168,3 +183,56 @@ class="large-text code"
...
+	$field_name     = PERFLAB_SERVER_TIMING_SETTING . '[' . $slug . ']';
+	$description_id = "{$field_id}_description";
+	$has_filter     = has_filter( 'perflab_server_timing_use_output_buffer' );
+	$is_enabled     = perflab_server_timing_use_output_buffer();
+
+	?>
+	<fieldset>
+		<legend class="screen-reader-text">
+			<?php esc_html_e( 'Output Buffering', 'performance-lab' ); ?>
+		</legend>
```
felixarntz: See my feedback above. You can simplify how this field is rendered by using a dedicated section and heading. Additionally, a `fieldset` isn't really useful around a single checkbox, it makes more sense when it's a group of checkboxes that somehow work together.

admin/server-timing.php
```diff
@@ -168,3 +183,56 @@ class="large-text code"
...
+				<?php else : ?>
+					<?php
+					echo wp_kses(
+						__( 'Output buffering has been forcibly disabled via the <code>perflab_server_timing_use_output_buffer</code> filter.', 'performance-lab' ),
+						array( 'code' => array() )
+					);
+					?>
+				<?php endif; ?>
+			<?php endif; ?>
+			<?php esc_html_e( 'This is needed to capture metrics after headers have been sent and the template is being rendered. Without output buffering, adding a hook that fires during template rendering may result in a PHP notice: "Function Perflab_Server_Timing::register_metric was called incorrectly. The method must be called before or during the perflab_server_timing_send_header action." Note that output buffering may possibly cause an increase in TTFB if the response would be flushed multiple times.', 'performance-lab' ); ?>
```
felixarntz: I think this is too detailed and too long for a field description. The error messages that it leads to IMO shouldn't be mentioned here. (Last but not least, if they are, they should be using a placeholder instead of being part of the translation string.)
```suggestion
			<?php esc_html_e( 'This is needed to capture metrics after headers have been sent and while the template is being rendered. Note that output buffering may possibly cause an increase in TTFB if the response would be flushed multiple times.', 'performance-lab' ); ?>
```

Another thought: The "after headers have been sent and the template is being rendered" may be easy to misunderstand? Would it be better to say "after headers have been sent and while the template is being rendered"

server-timing/class-perflab-server-timing.php
```diff
@@ -195,18 +195,25 @@ static function( $value ) {
...
+			 *
+			 * @param bool $use_output_buffer Whether to use an output buffer.
+			 */
+			return apply_filters( 'perflab_server_timing_use_output_buffer', false );
+		}
+
+		// Use the stored value from the Server-Timing screen.
+		$slug    = 'output_buffering';
+		$options = (array) get_option( PERFLAB_SERVER_TIMING_SETTING, array() );
+		return ! empty( $options[ $slug ] );
```
felixarntz: Here I would advise to not call the filter conditionally but rather filter the value of the option at the end. That's a bit more typical in WP, rather than using a "short-circuit" filter.

felixarntz (changes requested):
Thanks @westonruter, this looks great so far. I left some feedback, most importantly I'm thinking we could display the field in a slightly more prominent way.

felixarntz:
@westonruter Given that this applies to the newly added screen, I've added this to the 2.6.0 milestone and updated the base branch accordingly, let's try to merge it in time.

admin/server-timing.php
westonruter: Sure. I was thinking it felt more closely to [`createInterpolateElement`](https://make.wordpress.org/core/2020/07/17/introducing-createinterpolateelement/) but I'll change it to inline the `href` attribute.

westonruter: Done in 4868823

admin/server-timing.php
westonruter: I would agree, but the PerfLab plugin uses the same `fieldset` approach on its settings screen:

image

```html
<fieldset>
			Dominant Color Images		</legend>
		<label for="module_images/dominant-color-images_enabled">
							<input type="checkbox" id="module_images/dominant-color-images_enabled" name="perflab_modules_settings[images/dominant-color-images][enabled]" aria-describedby="module_images/dominant-color-images_description" value="1" checked="checked">
				Enable Dominant Color Images					</label>
		<p id="module_images/dominant-color-images_description" class="description">
			Adds support to store the dominant color of newly uploaded images and create a placeholder background of that color.		</p>
	</fieldset>
```


This is the same markup used  for other admin screens, like on General settings:

image

```html
<fieldset><legend class="screen-reader-text"><span>
	Membership</span></legend><label for="users_can_register">
<input name="users_can_register" type="checkbox" id="users_can_register" value="1">
	Anyone can register</label>
</fieldset>
```

And on Reading settings:

image

```html
<fieldset>
	<legend class="screen-reader-text"><span>
		Search engine visibility	</span></legend>
	<label for="blog_public"><input name="blog_public" type="checkbox" id="blog_public" value="0">
	Discourage search engines from indexing this site</label>
	<p class="description">It is up to search engines to honor this request.</p>
</fieldset>
```

That being said, there are other solitary checkboxes that lack the `fieldset`, including on Discussion settings:

image

```html
<td>
	<label for="show_avatars">
		<input type="checkbox" id="show_avatars" name="show_avatars" value="1" checked="checked">
		Show Avatars	</label>
</td>
```

westonruter: There is also a bit of styling present that only applies when the `fieldset` is present:

image

Without `fieldset`:

image

With `fieldset`:

image

admin/server-timing.php
westonruter: I edited your suggestion to add "while"

admin/server-timing.php
westonruter: I like the idea to move it up, but unfortunately it doesn't look like the Uploading Files section can be used as a model. It's actually a hard-coded settings screen and isn't using the settings API: https://github.com/WordPress/wordpress-develop/blob/7cd891845e5a3f461c3dbc1487467ab5e91c1827/src/wp-admin/options-media.php#L125-L169

Setting fields in sections always include that left table heading cell: https://github.com/WordPress/wordpress-develop/blob/7cd891845e5a3f461c3dbc1487467ab5e91c1827/src/wp-admin/includes/template.php#L1801-L1828

westonruter: Looks like we could do so by not adding the field via `add_settings_field()`

westonruter: Nevertheless, by not using a table then we don't get these styles:

image

image

Here you can see the field with and without the `table`:

image

I suppose we could just inline those styles to target just this fieldset.

westonruter: Well, I did this in 671c001246bf75ac125523213cf417a4ac732ad5, although duplicating the styles is not great. It looks good though:

image

server-timing/class-perflab-server-timing.php
westonruter: I was also wrestling with this. The reason why I went with the current approach is that _any_ filter present will cause the Output Buffering checkbox to be `disabled` on the settings screen. Therefore, it didn't seem to make sense to use the stored option. But it wouldn't really make much difference the other way around. Done in 8b415eb

admin/server-timing.php
felixarntz: Fair point, there are clearly some inconsistencies. I was primarily referring to it because we don't need the fieldset approach anymore if the section has a heading anyway. The fieldset + legend approach doesn't help at all in this case because the "Output Buffering" text is already used as heading anyway. That would mean a screen reader would have "Output Buffering" twice, from the heading and then from the legend, which would be pointless.

I think we should remove the `legend` for that reason, and then using `fieldset` doesn't really provide a value either, other than the default styling tweak, which is very minor though, and arguably irrelevant if we don't _display_ the legend anyway. I would suggest to simply use a `div.output-buffering` if we need a wrapper element to target everything within.

admin/server-timing.php
felixarntz: Thanks, I think that looks solid. While it's not great we have to duplicate those styles, I think it's an acceptable trade-off, particularly as you've already added a comment outlining where those come from, so 👍

admin/server-timing.php
```diff
@@ -168,3 +179,77 @@ class="large-text code"
...
+	$description_id = "{$field_id}_description";
+	$has_filter     = has_filter( 'perflab_server_timing_use_output_buffer' );
+	$is_enabled     = perflab_server_timing_use_output_buffer();
+
+	?>
+	<style>
+	/* The following styles are copied from core where they normally include `.form-table td` in the selectors. */
+	fieldset.output-buffering {
+		margin-bottom: 9px;
+		padding: 15px 10px 15px 0;
```
felixarntz: Let's remove this padding as it's pointless here and makes it look worse. Without this padding, the spacing overall is more consistent, while right now there is too much top and bottom spacing around this field.

admin/server-timing.php
```diff
@@ -168,3 +179,77 @@ class="large-text code"
...
+		>
+		<label for="<?php echo esc_attr( $field_id ); ?>">
+			<?php esc_html_e( 'Enable output buffering of template rendering.', 'performance-lab' ); ?>
+		</label>
+		<p id="<?php echo esc_attr( $description_id ); ?>" class="description">
+			<?php if ( $has_filter ) : ?>
+				<?php if ( $is_enabled ) : ?>
+					<?php
+					echo wp_kses(
+						__( 'Output buffering has been forcibly enabled via the <code>perflab_server_timing_use_output_buffer</code> filter.', 'performance-lab' ),
```
felixarntz: Please use a placeholder for the `<code>perflab_server_timing_use_output_buffer</code>` bit, similar to e.g. how it's handled above in the benchmarking section description.

admin/server-timing.php
```diff
@@ -168,3 +179,77 @@ class="large-text code"
...
+						array( 'code' => array() )
+					);
+					?>
+				<?php else : ?>
+					<?php
+					echo wp_kses(
+						__( 'Output buffering has been forcibly disabled via the <code>perflab_server_timing_use_output_buffer</code> filter.', 'performance-lab' ),
```
felixarntz: See above.

admin/server-timing.php
```diff
@@ -168,3 +179,77 @@ class="large-text code"
...
+						array( 'code' => array() )
+					);
+					?>
+				<?php endif; ?>
+			<?php endif; ?>
+			<?php esc_html_e( 'This is needed to capture metrics after headers have been sent and while the template is being rendered. Note that output buffering may possibly cause an increase in TTFB if the response would be flushed multiple times.', 'performance-lab' ); ?>
```
felixarntz: I would suggest to put this sentence into an `else`, as I think displaying all this explanation is less crucial if the user cannot tweak it anyway.

More importantly, can you split those two sentences apart and use a `<br>` in between? Right now, this string is so long that it becomes hard to read on large viewports due to very long lines.

admin/server-timing.php
```diff
@@ -168,3 +179,77 @@ class="large-text code"
...
+						array( 'code' => array() )
+					);
+					?>
+				<?php else : ?>
+					<?php
+					echo wp_kses(
+						__( 'Output buffering has been forcibly disabled via the <code>perflab_server_timing_use_output_buffer</code> filter.', 'performance-lab' ),
+						array( 'code' => array() )
+					);
+					?>
```
felixarntz: Related to my other comment, _if_ we don't put the other text below into an `else`, we should add a `<br>` here to avoid overly long lines.

admin/server-timing.php
westonruter: Done in c9a4d2137c17c0ef85c3f387ee9347e73d588dd2

server-timing/class-perflab-server-timing.php
```diff
@@ -195,6 +195,10 @@ static function( $value ) {
 	 * @return bool True if an output buffer should be used, false otherwise.
 	 */
 	public function use_output_buffer() {
+		$slug    = 'output_buffering';
+		$options = (array) get_option( PERFLAB_SERVER_TIMING_SETTING, array() );
+		$enabled = ! empty( $options[ $slug ] );
```
felixarntz: Nit-pick: No need for the `$slug` variable as it's only used once.
```suggestion
		$options = (array) get_option( PERFLAB_SERVER_TIMING_SETTING, array() );
		$enabled = ! empty( $options['output_buffering'] );
```

admin/server-timing.php
```diff
@@ -71,12 +78,16 @@ static function() {
 					<?php
 					echo wp_kses(
 						sprintf(
-							/* translators: 1: Server-Timing, 2: template_include */
-							__( 'Since the %1$s header is sent before the template is loaded, only hooks before the %2$s filter can be measured.', 'performance-lab' ),
+							/* translators: 1: Server-Timing, 2: template_include, 3: anchor link */
+							__( 'Since the %1$s header is sent before the template is loaded, only hooks before the %2$s filter can be measured. Enable <a href="%3$s">Output Buffering</a> to measure hooks during template rendering.', 'performance-lab' ),
```
felixarntz, line 102: Similar to my similar comments, can you split those two sentences and use a `<br>` between them to avoid long lines?

felixarntz (commented):
@westonruter Some mostly very minor follow-up feedback.

admin/server-timing.php
westonruter: How about we keep the bottom padding:

image

IMO, removing all padding makes it look worse.

This keeps whitespace similar to other settings screens:

image

admin/server-timing.php
westonruter: Done in 1df8a9794f6f59c000662fe897dbb4ec676561d1

admin/server-timing.php
westonruter: Done in 1df8a9794f6f59c000662fe897dbb4ec676561d1

admin/server-timing.php
westonruter: How about instead of adding line breaks, we instead add a max-width to the container? Adding line breaks can make it look bad on small viewports too:

image

westonruter: See 37d775aa86582a470b2fb1ca9abb44e3213b5353:

image

image

westonruter: Note a 800px max-width is also the approach that the Privacy Policy settings screen takes.

admin/server-timing.php
felixarntz: Good point! Keeping the bottom padding LGTM!

admin/server-timing.php
felixarntz: Thanks, I like that idea! 👍

admin/server-timing.php
```diff
@@ -58,25 +99,23 @@ static function() {
...
-				?>
-				<br>
-				<?php
 				echo wp_kses(
 					__( 'For any hook name provided, the <strong>cumulative duration between all callbacks</strong> attached to the hook is measured, in milliseconds.', 'performance-lab' ),
 					array( 'strong' => array() )
 				);
 				if ( ! perflab_server_timing_use_output_buffer() ) {
-					?>
-					<br>
```
felixarntz, line 70: I think this line break in particular makes sense to keep, not because of long lines but because of the underlying content of the two parts split by it.

admin/server-timing.php
```diff
@@ -58,25 +99,23 @@ static function() {
 					),
 					array( 'code' => array() )
 				);
-				?>
-				<br>
-				<?php
```
felixarntz, line 63: +1 to remove this, but it needs to be replaced with a space (`echo ' ';`), otherwise the text formatting is off.

admin/server-timing.php
felixarntz: The first part of this comment is still relevant, curious what you think.

felixarntz (commented):
This is almost there. A few last points of feedback.

admin/server-timing.php
westonruter: Ah yes, I tried that but then I thought that it's still actually useful to have the description. Otherwise, a user could think, "who cares that this option is disabled; I probably didn't need it anyway."

admin/server-timing.php
westonruter: Oh yes, good catch. Fixed in 60005c4

admin/server-timing.php
westonruter: How about a paragraph break instead? I think that is better:

image

felixarntz: Makes sense, let's go with that!

admin/server-timing.php
```diff
@@ -168,3 +212,61 @@ class="large-text code"
...
+							/* translators: %s: perflab_server_timing_use_output_buffer */
+							__( 'Output buffering has been forcibly disabled via the %s filter.', 'performance-lab' ),
+							'<code>perflab_server_timing_use_output_buffer</code>'
+						),
+						array( 'code' => array() )
+					);
+					?>
+				<?php endif; ?>
+			<?php endif; ?>
+			<?php esc_html_e( 'This is needed to capture metrics after headers have been sent and while the template is being rendered. Note that output buffering may possibly cause an increase in TTFB if the response would be flushed multiple times.', 'performance-lab' ); ?>
```
felixarntz: Maybe we should change the "This is needed" to another wording that works regardless of whether the setting is force-controlled or not? When the filter is force-disabled, it doesn't work well with the previous sentence. It sounds like disabling output buffering is needed...
Screenshot 2023-08-17 at 12 35 20 PM

This scenario is probably an edge-case, but the description reads a bit odd then.

felixarntz (approved):
@westonruter One final nit-pick on wording, but otherwise this looks great! 🎉

admin/server-timing.php
westonruter: How about:

```suggestion
			<?php esc_html_e( 'Output buffering is needed to capture metrics after headers have been sent and while the template is being rendered. Note that output buffering may possibly cause an increase in TTFB if the response would be flushed multiple times.', 'performance-lab' ); ?>
```

felixarntz: SGTM!

admin/server-timing.php
```diff
@@ -43,6 +43,47 @@ function perflab_add_server_timing_page() {
...
+					line-height: 1.3;
+					vertical-align: middle;
+					font-size: 14px;
+				}
+				.output-buffering-field label,
+				.output-buffering-field p {
+					line-height: 1.4;
+				}
+				.output-buffering-field label {
+					margin: 0.35em 0 0.5em !important;
```
mukeshpanchal27:
```suggestion
					margin: 0.35em 0 0.5em;
```
Important is not required here as it apply cleanly.

admin/server-timing.php
```diff
@@ -43,6 +43,47 @@ function perflab_add_server_timing_page() {
...
+					display: inline-block;
+					vertical-align: middle;
```
mukeshpanchal27:
```suggestion
```

Label set middle alignment in form css.
```
label, #your-profile label + a {
    vertical-align: middle;
}
```

admin/server-timing.php
```diff
@@ -43,6 +43,47 @@ function perflab_add_server_timing_page() {
...
+				}
+				.output-buffering-field p {
+					font-size: 14px;
+				}
```
mukeshpanchal27: CSS not applied to provided selector. It was override by below CSS so the text use 13px font size. If we have to use 14px font then we have to update selector `.output-buffering-field p.description`

```
p.help, p.description, span.description, .form-wrap p {
    font-size: 13px;
}
```

tests/server-timing/perflab-server-timing-tests.php
```diff
@@ -187,10 +187,56 @@ public function data_get_header() {
...
+				'expected' => false,
+			),
+		);
+	}
+
+	/**
+	 * @covers Perflab_Server_Timing::use_output_buffer
+	 * @dataProvider get_data_to_test_use_output_buffer
+	 * @param callable $set_up   Set up.
+	 * @param bool     $expected Expected value.
```
mukeshpanchal27:
```suggestion
	 * @covers Perflab_Server_Timing::use_output_buffer
	 *
	 * @dataProvider get_data_to_test_use_output_buffer
	 *
	 * @param callable $set_up   Set up.
	 * @param bool     $expected Expected value.
```

mukeshpanchal27 (approved):
Thanks @westonruter for the PR. LGTM! I left some non blocking feedback.

admin/server-timing.php
felixarntz: Good catch! Not sure why we would want to set it to `14px` actually, why not use the regular size core would use?

felixarntz:
@westonruter Could you take a look at @mukeshpanchal27's feedback please so that we can try to merge this today?

admin/server-timing.php
westonruter: I'm OoO today. So please make whatever change you need. But I largely copied the styles from core as I recall.

admin/server-timing.php
westonruter: This is copied from core. Please make the change as you see fit.

admin/server-timing.php
westonruter: This is copied from core. Please make the change as you see fit.

admin/server-timing.php
felixarntz: Sorry for the noise then :) Will do!

felixarntz:
@westonruter @mukeshpanchal27 In light of the CSS feedback, after taking a closer look I've pushed an update in https://github.com/WordPress/performance/pull/801/commits/7e8c9c501eebb3ad178152941ec3cb2cd4ca83fb that changes the markup to match what is used on WP core's own settings screens, specifically for the `uploads_use_yearmonth_folders` UI control (which for example still uses a `.form-table`, which was previously missing here and therefore required several style tweaks to look consistent with other core settings UI). This change also means that almost all of the custom CSS was removed while maintaining a consistent appearance.

Issue #652, opened 2023-02-16
Implement migration logic and UI from old PL modules to new standalone plugins in PL settings screen


felixarntz:
## Feature Description

Follow-up to #651: For any site for which the old module slugs are still set in the option for active modules, an additional UI component should be added to the updated PL settings screen to make it easy, yet explicitly intentional, for the admin user to install and activate the respective standalone plugins with a single click.

This will require defining a mapping between the old module slugs and their standalone plugins, so that the plugin can determine where to point the user.

### Requirements

* If any of the module slugs that are already available as standalone plugins are still remaining in the option for active modules, some sort of prompt (e.g. an admin notice) should be displayed on the _Settings > Performance_ screen.
    * The check for whether this prompt needs to be shown should be a reusable function (important for #653).
    * Which exact modules need to be replaced with their plugins is less important to be reusable, though it could certainly be part of a reusable function too if it makes sense.
* The prompt shown on the screen should include:
    * A brief explanation (1-2 sentences) to the user that they have modules activated which will soon be removed from the Performance Lab plugin, and that they need to instead install/activate the relevant standalone plugins
    * A list of which modules are affected (if it's a single module, we could alternatively consider putting it into the floating text, since a list of 1 item is a bit odd)
    * One of the following:
        * _If the current user has the `install_plugins` and `activate_plugins` capabilities:_ A primary CTA button (see below) to migrate the active legacy modules to their respective standalone plugins
        * _If the current user lacks any of the two capabilities:_ A CTA message prompting the user to contact an administrator with access to installing/activating plugins
* When the primary CTA button is clicked, logic (could be implemented with a full page reload/form submission, or AJAX, whatever works best) should be run that:
    * Installs and activates the respective standalone plugins as needed.
    * Deactivates the respective legacy modules.

felixarntz:
@10upsimon The requirements here are extremely detailed, and I would argue go too far into detail, to the degree it is very challenging to review them as it's almost like a code review without code :)

I took a look at your draft PR and left some feedback (https://github.com/WordPress/performance/pull/806#pullrequestreview-1583243568), which is similar to what I would have as feedback on the requirements: Rather than the detailed list of what to implement, I think the description here would benefit from a much higher-level outline of requirements, while instead providing context for _why_ the requirements are like that. For example, I am questioning why we need to introduce REST endpoints when WP core already provides AJAX functions and REST endpoints to manage plugin installation, activation, etc.

[...]

PR #806, opened 2023-08-17
DO NOT MERGE | POC | Enhancement / 652 Migration UI From Old Modules to Standalone Plugins

10upsimon:
## Summary

Addresses: #652

## Relevant technical choices

- TBA

## Checklist


load.php
```diff
@@ -25,6 +25,11 @@
 define( 'PERFLAB_MODULES_SETTING', 'perflab_modules_settings' );
 define( 'PERFLAB_MODULES_SCREEN', 'perflab-modules' );
 
+// Load Composer dependencies if applicable.
+if ( file_exists( PERFLAB_PLUGIN_DIR_PATH . '/vendor/autoload.php' ) ) {
+	require_once PERFLAB_PLUGIN_DIR_PATH . '/vendor/autoload.php';
+}
```
felixarntz, line 31: While Composer is generally great, using it in a WordPress plugin comes with additional quirks and interoperability problems that I would prefer to avoid unless truly beneficial. Let's `require_once` any class files instead, as there shouldn't be any meaningful performance benefit from autoloading those few files anyway.

includes/classes/rest/class-rest-plugins-controller.php
```diff
@@ -0,0 +1,131 @@
...
+use WP_REST_Server;
+
+/**
+ * Class for handling standalone performance plugin actions via REST endpoints.
+ *
+ * @since n.e.x.t
+ * @access private
+ * @ignore
+ */
+class REST_Plugins_Controller {
```
felixarntz, line 28: Do we really need those REST routes? While I agree the REST API is generally the recommended approach over the oldschool AJAX request, since we are working on a WP core feature plugin here and since this functionality is supplemental, we should keep our maintenance burden as small as possible and use what core already provides. Is there a good reason not to use `wp_ajax_install_plugin()` and `WP_REST_Plugins_Controller`?

composer.json
```diff
@@ -48,5 +48,8 @@
     "psr-4": {
       "PerformanceLab\\Tests\\": "tests/utils"
     }
+  },
+  "autoload": {
+    "classmap": ["includes/"]
```
felixarntz, line 53: See above, let's avoid the usage of Composer _within_ the actual plugin.

includes/classes/class-plugin-manager.php
```diff
@@ -0,0 +1,568 @@
...
+			),
+			'sqlite-database-integration' => array(
+				'Name'        => esc_html__( 'SQLite Database Integration', 'performance-lab' ),
+				'Description' => esc_html__( 'Allows testing an SQLite integration with WordPress and gather feedback, with the goal of eventually landing it in WordPress core.', 'performance-lab' ),
+				'Author'      => esc_html__( 'WordPress Performance Team', 'performance-lab' ),
+				'AuthorURI'   => esc_url( 'https://make.wordpress.org/performance/' ),
+				'PluginURI'   => esc_url( 'https://wordpress.org/plugins/sqlite-database-integration/' ),
+				'Download'    => 'wporg',
+			),
+		);
```
felixarntz, line 51: All this shouldn't be hard-coded. Instead, we should rely on the `plugin.json` file for the slugs of which plugins to cover here, and all other data should come from the WordPress.org API, which we can query in the same way that it's done on the _Plugins > Add New_ screen.

includes/classes/class-plugin-manager.php
```diff
@@ -0,0 +1,568 @@
...
+			'Status'      => '',
+		);
+
+		// Add plugin status info and fill in defaults.
+		foreach ( $managed_standalone_plugins as $plugin_slug => $managed_standalone_plugin ) {
+			$status = self::get_managed_standalone_plugin_status( $plugin_slug );
+
+			$managed_standalone_plugins[ $plugin_slug ]['Status']      = $status;
+			$managed_standalone_plugins[ $plugin_slug ]['Slug']        = $plugin_slug;
+			$managed_standalone_plugins[ $plugin_slug ]['HandoffLink'] = isset( $managed_standalone_plugins[ $plugin_slug ]['EditPath'] ) ? admin_url( $managed_standalone_plugins[ $plugin_slug ]['EditPath'] ) : null;
```
felixarntz, line 72: What does "HandoffLink" refer to? What is it for?

includes/classes/class-plugin-manager.php
```diff
@@ -0,0 +1,568 @@
...
+		if ( ! is_plugin_active( $plugin_file ) ) {
+			return new WP_Error( 'wpp_plugin_not_active', __( 'The plugin is not active.', 'performance-lab' ) );
+		}
+
+		deactivate_plugins( $plugin_file );
+		if ( is_plugin_active( $plugin_file ) ) {
+			return new WP_Error( 'wpp_plugin_failed_deactivation', __( 'Failed to deactivate plugin.', 'performance-lab' ) );
+		}
+		return true;
+	}
```
felixarntz, line 186: Similarly to my comment on the REST classes, why do we need methods to activate and deactivate? Can't we use what core already provides for that purpose?

felixarntz (commented):
@10upsimon I left a bit of early feedback, mostly related to the overall approach outlined on #652.

On another note, please create a feature branch for this work, and then this PR should be based on that instead of `trunk`.

PR #851, opened 2023-10-10
Bump minimum required PHP version to 7.0 and minimum required WP version to 6.3

### LINKED ISSUE #606: Bump minimum required PHP version to 7.0 and minimum required WP version to 6.3
## Bug Description

As WordPress bumps the [minimum required PHP version to 7.0](https://core.trac.wordpress.org/ticket/57345), we should do the same for plugins.

mukeshpanchal27:
## Summary

Fixes #606

- Bump main plugin minimum required WP version to 6.3
- Bump standalone plugin minimum required WP version to 6.3.
- Bump main plugin minimum required PHP version to 7.0.
- Bump standalone plugin minimum required PHP version to 7.0.
- Update/remove legacy code.

## Checklist


mukeshpanchal27:
@felixarntz, should we include the update of the standalone plugin version as part of this PR, or would it be better to handle it in a separate PR?

composer.json
```diff
@@ -28,7 +28,7 @@
   },
   "require": {
     "composer/installers": "~1.0",
-    "php": ">=5.6|^7|^8"
+    "php": ">=7|^8"
```
spacedmonkey, line 30: What about `"phpunit/phpunit": "^4|^5|^6|^7|^8|^9",`?

spacedmonkey: PHPUnit is installed by `yoast/phpunit-polyfills`. This line should be removed. CC @swissspidy

swissspidy: Indeed, it should be safe to simply remove that `phpunit/phpunit` line (and run `composer update` afterwards) because it's required by `yoast/phpunit-polyfills` anyway.

modules/images/fetchpriority/readme.txt
```diff
@@ -1,9 +1,9 @@
 === Fetchpriority ===
 
 Contributors:      wordpressdotorg
-Requires at least: 6.1
+Requires at least: 6.3
```
swissspidy: Isn't this a bit pointless? Fetchpriority landed in WordPress 6.3, and this plugin does nothing when installing on 6.3. Shouldn't we just archive this plugin instead?

composer.json
mukeshpanchal27: Removed in https://github.com/WordPress/performance/pull/851/commits/b5179cfbcaba41ef327436d0513a370699be7b7f. I don't know why `composer update` did not solve the CI job. Could anyone guide me.

modules/images/fetchpriority/readme.txt
felixarntz: Good point, I agree. Let's remove `modules/images/fetchpriority` entirely. But better to do that in a separate PR, since it's a different change. @mukeshpanchal27 Would you mind opening another one simply for removal of Fetchpriority? We don't need to ship any more updates to that plugin, and it doesn't need to be part of Performance Lab anymore.

felixarntz (approved):

Happy to merge this despite the one outstanding point mentioned by @swissspidy. Let's open a separate PR for that, should be a quick one.

felixarntz:
@mukeshpanchal27 
> @felixarntz, should we include the update of the standalone plugin version as part of this PR, or would it be better to handle it in a separate PR?

Just saw this comment. How about we merge this PR as is, and then open another PR that does the following:
* remove Fetchpriority module/plugin completely
* bump patch version of WebP Uploads plugin

See the new #853.

Note that Dominant Color Images doesn't need any version bump because it's not released yet anyway.

cc @swissspidy

spacedmonkey:
Composer install is not failing.

modules/images/fetchpriority/readme.txt
mukeshpanchal27: Thanks. Open PR https://github.com/WordPress/performance/pull/854 that remove the module.

aristath (approved):
LGTM 👍 
Pre-approving, pending the fix for a failing composer issue

mukeshpanchal27:
@swissspidy or @spacedmonkey, could either of you please take care of pushing the changes needed for `composer update`? It seems I'm encountering issues when trying to perform this task.

modules/images/fetchpriority/readme.txt
joemcgill: Left a comment on the PR, but wondering if we should reconsider removing this yet.

PR #854, opened 2023-10-11
Remove Fetchpriority module as the functionality is now available in WordPress core

### LINKED ISSUE #853: Remove Fetchpriority module and plugin
See https://github.com/WordPress/performance/pull/851#discussion_r1352417111 and #606: Now that Performance Lab requires WordPress 6.3+, the Fetchpriority module is completely pointless. The plugin doesn't require any more updates either, so we should completely remove it from PL.

The PR for this should also bump the patch version of WebP Uploads, given the latest changes.

mukeshpanchal27:
## Summary

Fixes #853

## Checklist


swissspidy (approved):

joemcgill:
I understand that the rationale here is that since fetchpriority has landed in WP 6.3, that this plugin is essentially finished software and could be archived. However, we continue to have [ongoing conversations](https://wordpress.slack.com/archives/C02KGN5K076/p1696952251046129) about how to improve the heuristics and markup techniques used to accurately apply fetchpriority to the best LCP candidate.

For that reason, I wonder if we should keep this plugin in active development as a way of quickly iterating on those improvements prior to them being shipped in an official WP release?

felixarntz:
@joemcgill I thought about this as well before agreeing to remove the module, my conclusion is the following: Yes, we may want to improve the heuristics, but to me this does not fit into this existing module / plugin, as its goal was to enable fetchpriority support.

Improving the heuristics affects multiple attributes (`loading` and `fetchpriority`) and IMO would better fit into a separate module, potentially even as part of a "smart LCP detection" module. Even if that is focused on client-side detection, it could also provide functionality to improve the core server-side detection.

felixarntz (approved):

joemcgill:
> and IMO would better fit into a separate module, potentially even as part of a "smart LCP detection" module

I had the very same thought when writing my original comment. I'm onboard with that plan.

PR #866, opened 2023-10-20
Exclude `.wordpress-org` directory when deploying standalone plugins

### LINKED ISSUE #850: Exclude `.wordpress-org` directory when deploying standalone plugins
## Bug Description

When deploying standalone plugins, it looks like at the moment the `.wordpress-org` directories for them are actually deployed to WordPress.org, as part of the actual plugin. The files in that directory should only be deployed into the SVN repository's `assets` folder, which works as expected, but additionally the `.wordpress-org` directory is also added to the plugin itself. This is unnecessary and notably increases the plugin ZIP size, so it should be fixed.

I discovered this bug when doing a `DRY_RUN` for the Dominant Color Images plugin (see https://github.com/WordPress/performance/actions/runs/6425426416/job/17447991722#step:6:100).
After that I checked the existing standalone plugins and found the bug already present there too, see below:
* https://plugins.trac.wordpress.org/browser/webp-uploads/trunk/.wordpress-org
* https://plugins.trac.wordpress.org/browser/fetchpriority/trunk/.wordpress-org
* https://plugins.trac.wordpress.org/browser/dominant-color-images/trunk/.wordpress-org

**Note:** This is _only_ a problem for the standalone plugins, not for the main Performance Lab plugin.

## Solution
* We need to ensure that the `.wordpress-org` directory for each standalone plugin is not deployed as part of the plugin itself (i.e. within the plugin's SVN `trunk` directory).
* While fixing that, we must however ensure that the files from the `.wordpress-org` directory continue to be considered for the plugin's SVN `assets` directory.
* Since this bug is already in production for `webp-uploads` and `fetchpriority`, and `dominant-color-images`, we should probably as part of the PR also bump the version of those plugins as a patch version increase, so that upon the next deployment the problem is fixed.
* `fetchpriority` is no longer part of the repository, so we can ignore that part.

mukeshpanchal27:
## Summary

Fixes #850

I have manually tested the standalone workflow on the forked repository. I have generated a zip file for the plugins and attached it to the summary to verify the process. You can check the following URLs and verify the attached zip:
- [WebP Uploads](https://github.com/mukeshpanchal27/performance/actions/runs/6583740260)
- [Dominant Color Images](https://github.com/mukeshpanchal27/performance/actions/runs/6583742027)

## Checklist

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -1,10 +1,10 @@
 {
   "images/dominant-color-images": {
     "slug": "dominant-color-images",
-    "version": "1.0.0"
+    "version": "1.0.1"
   },
   "images/webp-uploads": {
     "slug": "webp-uploads",
-    "version": "1.0.4"
+    "version": "1.0.5"
   }
 }
--- a/admin/load.php
+++ b/admin/load.php
@@ -41,9 +41,9 @@ add_action( 'admin_menu', 'perflab_add_modules_page' );
 /**
  * Initializes settings sections and fields for the modules page.
  *
- * @global array $wp_settings_sections Registered WordPress settings sections.
- *
  * @since 1.0.0
+ *
+ * @global array $wp_settings_sections Registered WordPress settings sections.
  *
  * @param array|null $modules     Associative array of available module data, keyed by module slug. By default, this
  *                                will rely on {@see perflab_get_modules()}.
--- a/modules/database/audit-autoloaded-options/helper.php
+++ b/modules/database/audit-autoloaded-options/helper.php
@@ -95,6 +95,8 @@ function perflab_aao_autoloaded_options_test() {
  *
  * @since 1.0.0
  *
+ * @global wpdb $wpdb WordPress database abstraction object.
+ *
  * @return int autoloaded data in bytes.
  */
 function perflab_aao_autoloaded_options_size() {
@@ -106,6 +108,8 @@ function perflab_aao_autoloaded_options_size() {
  * Fetches autoload top list.
  *
  * @since 1.5.0
+ *
+ * @global wpdb $wpdb WordPress database abstraction object.
  *
  * @return array Autoloaded data as option names and their sizes.
  */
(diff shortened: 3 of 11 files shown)


felixarntz (commented):
@mukeshpanchal27 Could we use `.gitattributes` instead of `.distignore`, the same way we do for the main plugin?

plugins.json
```diff
@@ -1,10 +1,10 @@
...
     "slug": "dominant-color-images",
-    "version": "1.0.0"
+    "version": "1.0.1"
   },
   "images/webp-uploads": {
     "slug": "webp-uploads",
-    "version": "1.0.4"
+    "version": "1.0.5"
   }
 }
```
felixarntz, line 10: Can you please also update the readme for these two and add a changelog entry pointing to this PR?

mukeshpanchal27: Missed it 😞. Thanks for pointing. Added changes on readme in https://github.com/WordPress/performance/pull/866/commits/a77d3e97dfa64175fc83efd8377bc221b4e3dddb

mukeshpanchal27:
@felixarntz thanks for the feedback. PR is ready for another round of feedback. Thanks! cc. @joemcgill

felixarntz (approved):

felixarntz:
@mukeshpanchal27 One thing we should test here is to make sure that the module specific `.gitattributes` files are not part of the _main_ plugin ZIP. We could probably test that with a dry run of the main release workflow.

mukeshpanchal27:
Thank you, @felixarntz, for bringing this to my attention. I've conducted testing on the main plugin zip through a GitHub action workflow in the forked repository, and it appears that the plugin is exclude the `.gitattributes` and `.wordpress-org` files for the module directories.

You can review the workflow here: [GitHub Workflow](https://github.com/mukeshpanchal27/performance/actions/runs/6622545166). Once you download the attached `performance-lab` file, extract it, and subsequently extract the zip generated by the GitHub action, which is meant to be deployed on WordPress.org, you will notice that these excluded files cannot be found.

Screenshot 2023-10-24 at 2 23 24 PM

swissspidy (approved):

PR #875, opened 2023-10-25
Add Image Loading Optimization module

### LINKED ISSUE #805: Module Proposal: Alternative LCP Optimization (via Output Buffering)
## Overview
- Proposed module name: Alternative LCP Optimization
- Proposed module slug: alternative-lcp-optimization
- Proposed module owner GitHub username(s): westonruter

### Purpose

Disable core's insertion of `fetchpriority="high"` and `loading="lazy"` attributes during template rendering to instead insert them on the fully-rendered template captured via output buffering. The module would explore whether this is feasible and whether it produces better results.

### Scope

* Add output buffering of the template in a similar way as Server-Timing is doing already. A separate output buffer can be used specific for this module without needing to reuse the one from Server-Timing.
* Introduce a filter in the output buffer callback. This would be implementing Core-43258 which could be reused by plugins seeking to do optimizations or by caching plugins to store the rendered page.
* Use `WP_HTML_Tag_Processor` to iterate over the HTML document and inject the `fetchpriority="high"` and `loading="lazy"` attributes. This may involve the use of bookmarks to gather up a set of candidates, also informed by existing heuristics during template generation.

### Rationale

WordPress currently optimized LCP elements by injecting `fetchpriority="high"` and omitting `loading="lazy"` during template rendering. This has been the only option available for core because core lacks any standardized mechanism for output buffering and filtering the output of the page, even though a multitude of plugins do so (including the Performance Lab plugin to do Server-Timing). 

As noted in #793, there are currently some cases where the optimization attributes are not applied correctly, due to the complex and varied mechanisms that WordPress core and plugins use to render templates. Some themes and plugins may even hard-code `img` tags in which case it is currently impossible to apply the correct attributes.

If the rendered template were output buffered, then the optimization attributes could be applied once on the entire document before the complete HTML page is returned to the user. 

Note that block themes are already essentially output buffered, although without using `ob_start()`.

westonruter:
## Summary

Add scaffolding for new Image Loading Module with some key basic functionality, namely the output buffering of template renders and the introduction of a filter for that buffered output.

Fixes #805

## Relevant technical choices

When the module is active, the output buffering UI elements for Server Timing are omitted since the Image Loading Optimization module always does output buffering. Similarly, when the module is running, Server Timing will wait to send the header at the end of the new `perflab_template_output_buffer` filter.

## Checklist

### DIFF
--- a/modules/images/image-loading-optimization/load.php
+++ b/modules/images/image-loading-optimization/load.php
@@ -1,0 +1,17 @@
+<?php
+/**
+ * Module Name: Image Loading Optimization
+ * Description: Improves accuracy of optimizing the loading of the LCP image by leveraging client-side detection with real user metrics. Also enables output buffering of template rendering which can be filtered.
+ * Experimental: Yes
+ *
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+// Do not load the code if it is already loaded through another means.
+if ( function_exists( 'image_loading_optimization_buffer_output' ) ) {
+	return;
+}
+
+require_once __DIR__ . '/helper.php';
+require_once __DIR__ . '/hooks.php';
--- a/modules/images/image-loading-optimization/load.php
+++ b/modules/images/image-loading-optimization/load.php
@@ -8,6 +8,13 @@
  * @since n.e.x.t
  */
 
+// Define the constant.
+if ( defined( 'IMAGE_LOADING_OPTIMIZATION_VERSION' ) ) {
+	return;
+}
+
+define( 'IMAGE_LOADING_OPTIMIZATION_VERSION', 'Performance Lab ' . PERFLAB_VERSION );
+
 // Do not load the code if it is already loaded through another means.
 if ( function_exists( 'image_loading_optimization_buffer_output' ) ) {
 	return;
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -53,3 +53,8 @@
 /modules/images/dominant-color-images                                       @pbearne @spacedmonkey
 /tests/modules/images/dominant-color-images                                 @pbearne @spacedmonkey
 /tests/testdata/modules/images/dominant-color-images                        @pbearne @spacedmonkey
+
+# Module: Image Loading Optimization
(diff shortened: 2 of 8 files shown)


swissspidy (approved):

felixarntz:
@westonruter Can you please update this to be against a `feature/*` branch? I'm not sure we'd want to ship this yet just with this PR, so I think it shouldn't be against `trunk` until we're comfortable with the version we have.

modules/images/image-loading-optimization/load.php
```diff
@@ -0,0 +1,17 @@
...
+ * Experimental: Yes
+ *
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+// Do not load the code if it is already loaded through another means.
+if ( function_exists( 'image_loading_optimization_buffer_output' ) ) {
+	return;
+}
```
felixarntz, line 21: Can you replace this with an `IMAGE_LOADING_OPTIMIZATION_VERSION` constant, similar to the other modules?

Those should be used on every module for interoperability with standalone plugins. The `function_exists()` checks were added before those constants were required, but because of the constants they are now no longer needed (we could even remove them from the other modules, but not important).

felixarntz (changes requested):
@westonruter This looks good, just one point of feedback (plus let's change the base branch before merge to a new feature branch).

modules/images/image-loading-optimization/load.php
westonruter: Got it. I was using `webp-support` as my model. Done in 66284bb.

westonruter:
> plus let's change the base branch before merge to a new feature branch

@felixarntz I just now pushed up a new `feature/image-loading-optimization` branch off of `trunk` and changed the base branch of this PR to target it.

felixarntz (approved):
Thanks @westonruter!

felixarntz:
Just noted that this PR and #878 are annotated with the PL 2.8.0 release, but it's not actually milestoned for a release yet since we are working in the feature branch towards a working version before moving ahead with merging into `trunk`. Let's use `no milestone` on those PRs against the feature branch.

Issue #872, opened 2023-10-25
Apply image loading optimizations based on stored client-side detection


westonruter:
## Feature Description

Part of #869. See [document section](https://docs.google.com/document/d/16qAJ7I_ljhEdx2Cn2VlK7IkiixobY9zNn8FXxN9T9Ls/edit#heading=h.goz1o71xk9ad).

Issue #871, opened 2023-10-25
Create storage backend for image loading optimization from client-side detection


westonruter:
## Feature Description

Part of #869. See [document section](https://docs.google.com/document/d/16qAJ7I_ljhEdx2Cn2VlK7IkiixobY9zNn8FXxN9T9Ls/edit#heading=h.koav74tnylo8).

westonruter:
Done in #878

Issue #870, opened 2023-10-25
Create client-side detection logic for optimizing image loading


westonruter:
## Feature Description

Part of #869. See [document section](https://docs.google.com/document/d/16qAJ7I_ljhEdx2Cn2VlK7IkiixobY9zNn8FXxN9T9Ls/edit#heading=h.322h3vh9dtxi).

PR #864, opened 2023-10-19
Introduce UI for managing Performance Lab standalone plugins.

10upsimon:
## Summary

Addresses: #651

## Relevant technical choices
- Created new `PerformanceLab\Plugin_Manager` class
    - Addition of `get_standalone_plugins()` static method  as a means to reference the available standalone plugins. This will need to be updated in PL plugin releases as more standalone plugins are released to the WordPress plugin directory.
    - Addition of `render_plugins_ui()` static method as a catch all use-once method to handle rendering of the standalone plugins UI within the PL setting screen. Used within `admin/load.php` as the main means of rendering the view. This borrows heavily from cores implementation of the UI via `Plugins > Add New`.
    - Addition of `render_plugin_card()` static method that handles rendering of the UI for each individual plugin card. This borrows heavily from cores implementation of the UI via `Plugins > Add New`.
- Enhancements to `admin/load.php`
    - Updates to `perflab_render_modules_page()` function to include a call to `Plugin_Manager::render_plugins_ui()` in order to render the standalone plugins UI as the first UI section within the PL settings screen.
    - Definition of new callback function `perflab_handle_inline_plugin_actions()` hooked to action `admin_init` in order to successfully redirect the end user back to the PL Settings screen upon activation or deactivation of a standalone plugin from the standalone plugins UI rendered within the PL settings screen.
    - Definition of new callback function `perflab_admin_scripts` hooked to action `admin_enqueue_scripts` in order to:
        - Enqueue the core `updates` script, used to handle plugin actions within the rendered UI as a result of `PerformanceLab\Plugin_Manager::render_plugins_ui()`.
        - Localize the `updates` script via `wp_localize_script()` and provide the following data:
            - `settings` An array of settings necessary for correct functionality of `plugins` script, most notably the `totals` key which contains plugin update totals, facilitate via a call to `wp_get_update_data()`. Without this, the `updates` script does not function as intended as it utilizes this data for retrieving and updating data stores.
        - Enqueues the core `thickbox` script, necessary for correct lightbox handling from within each plugin UI card.
        - Enqueues the core `thickbox` style, necessary for correct lightbox appearance as triggered from within each plugin UI card.

## Checklist

- [x] PR has either `[Focus]` or `Infrastructure` label.
- [x] PR has a `[Type]` label.
- [x] PR has a milestone or the `no milestone` label.

PR #876, opened 2023-10-26
Add initial detection logic for Image Loading Optimization

westonruter:
## Summary

Add logic to detect the LCP element and images and elements with background images which are in the initial viewport.

See #870.

This is currently a sub-PR of #875.

## Relevant technical choices

The logic is contained in a `detect.js` script which gets imported as a module from an inline module script. This module script is currently printed to the page with every page load, but once storage (https://github.com/WordPress/performance/issues/871) is implemented, it will be conditionally added based on whether no existing metrics are available.

The logic in `detect.js` only runs if it has been invoked within 5 seconds of the page being generated. This ensures that if the page output is cached, any future conditional serving of the detection script will not end up resulting in many clients running the logic unnecessarily when page caching is involved. The 5-second window can be filtered by `perflab_image_loading_detection_time_window`.

The detection logic involves gathering up a list of the optimizable elements, namely images and elements with background images (for now). For each of these elements, breadcrumbs are obtained which is an array of objects containing the tag name and element index. For example,  the custom logo's breadcrumbs are:

```json
[
    {
        "tagName": "HTML",
        "index": 0
    },
    {
        "tagName": "BODY",
        "index": 1
    },
    {
        "tagName": "DIV",
        "index": 2
    },
    {
        "tagName": "HEADER",
        "index": 0
    },
    {
        "tagName": "DIV",
        "index": 0
    },
    {
        "tagName": "DIV",
        "index": 0
    },
    {
        "tagName": "DIV",
        "index": 0
    },
    {
        "tagName": "A",
        "index": 0
    },
    {
        "tagName": "IMG",
        "index": 0
    }
]
```

These breadcrumbs are captured as soon as the module runs so that any subsequent dynamic insertions of new elements by JavaScript won't invalidate the breadcrumbs to look up the elements on the server-rendered page at the end of output buffering.

With the list of optimizable elements in hand, an IntersectionObserver is used to keep track of when they become visible (if ever). If the initial page load is not at the top of the page then the detection logic aborts since we don't want to capture metrics for an anchor link. Similarly, the IntersectionObserver disconnects as soon as the user starts scrolling. Elements contained inside of the admin bar are ignored.

In addition to observing for visible optimizable elements, the `onLCP` function from web-vitals.js library is used to obtain the LCP element metric. Currently `onLCP` is invoked with `reportAllChanges` as `true` as otherwise it will only fire when the user clicks or presses a key. So at the very least for development purposes, `reportAllChanges` is enabled. When the window's `load` event fires, then the latest LCP candidate metric reported is considered LCP.

Finally, page metrics are gathered to send to the server, including:

* The viewport width and height
* Metrics for each optimizable element, including:
  * Whether it is LCP
  * Whether it was an LCP candidate (i.e. second most important)
  * Breadcrumbs
  * Intersection ratio
  * Intersection rectangle
  * Bounding client rectangle

For example:

```json
{
    "viewport": {
        "width": 3201,
        "height": 1644
    },
    "elements": [
        {
            "isLCP": false,
            "isLCPCandidate": false,
            "breadcrumbs": [
                {
                    "tagName": "HTML",
                    "index": 0
                },
                {
                    "tagName": "BODY",
                    "index": 1
                },
                {
                    "tagName": "DIV",
                    "index": 2
                },
                {
                    "tagName": "HEADER",
                    "index": 0
                },
                {
                    "tagName": "DIV",
                    "index": 0
                },
                {
                    "tagName": "DIV",
                    "index": 0
                },
                {
                    "tagName": "DIV",
                    "index": 0
                },
                {
                    "tagName": "A",
                    "index": 0
                },
                {
                    "tagName": "IMG",
                    "index": 0
                }
            ],
            "intersectionRatio": 1,
            "intersectionRect": {
                "x": 1000.40625,
                "y": 79.984375,
                "width": 120.015625,
                "height": 120.015625,
                "top": 79.984375,
                "right": 1120.421875,
                "bottom": 200,
                "left": 1000.40625
            },
            "boundingClientRect": {
                "x": 1000.40625,
                "y": 79.984375,
                "width": 120.015625,
                "height": 120.015625,
                "top": 79.984375,
                "right": 1120.421875,
                "bottom": 200,
                "left": 1000.40625
            }
        },
        {
            "isLCP": false,
            "isLCPCandidate": false,
            "breadcrumbs": [
                {
                    "tagName": "HTML",
                    "index": 0
                },
                {
                    "tagName": "BODY",
                    "index": 1
                },
                {
                    "tagName": "DIV",
                    "index": 2
                },
                {
                    "tagName": "MAIN",
                    "index": 1
                },
                {
                    "tagName": "DIV",
                    "index": 1
                },
                {
                    "tagName": "FIGURE",
                    "index": 0
                },
                {
                    "tagName": "IMG",
                    "index": 0
                }
            ],
            "intersectionRatio": 1,
            "intersectionRect": {
                "x": 1275.40625,
                "y": 466.328125,
                "width": 0.015625,
                "height": 0.015625,
                "top": 466.328125,
                "right": 1275.421875,
                "bottom": 466.34375,
                "left": 1275.40625
            },
            "boundingClientRect": {
                "x": 1275.40625,
                "y": 466.328125,
                "width": 0.015625,
                "height": 0.015625,
                "top": 466.328125,
                "right": 1275.421875,
                "bottom": 466.34375,
                "left": 1275.40625
            }
        },
        {
            "isLCP": true,
            "isLCPCandidate": true,
            "breadcrumbs": [
                {
                    "tagName": "HTML",
                    "index": 0
                },
                {
                    "tagName": "BODY",
                    "index": 1
                },
                {
                    "tagName": "DIV",
                    "index": 2
                },
                {
                    "tagName": "MAIN",
                    "index": 1
                },
                {
                    "tagName": "DIV",
                    "index": 1
                },
                {
                    "tagName": "FIGURE",
                    "index": 1
                },
                {
                    "tagName": "IMG",
                    "index": 0
                }
            ],
            "intersectionRatio": 1,
            "intersectionRect": {
                "x": 1275.40625,
                "y": 490.328125,
                "width": 300.015625,
                "height": 196.015625,
                "top": 490.328125,
                "right": 1575.421875,
                "bottom": 686.34375,
                "left": 1275.40625
            },
            "boundingClientRect": {
                "x": 1275.40625,
                "y": 490.328125,
                "width": 300.015625,
                "height": 196.015625,
                "top": 490.328125,
                "right": 1575.421875,
                "bottom": 686.34375,
                "left": 1275.40625
            }
        }
    ]
}
```

With these metrics in hand, the server can to store them (#871) and make use of them to optimize the loading of those elements on the page (#872).

## Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,8 +16,7 @@ const config = {
 	globals: {
 		scheduler: false,
 	},
-	// Note: The '/wp-*' pattern is to ignore symlinks which may be added for local development.
-	ignorePatterns: [ '/vendor', '/node_modules', '/wp-*' ],
+	ignorePatterns: [ '/vendor', '/node_modules' ],
 };
 
 module.exports = config;
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,11 +11,13 @@ const config = {
 		'no-console': 'off',
 	},
 	env: {
-		'browser': true,
+		browser: true,
 	},
 	globals: {
 		scheduler: false,
-	}
+	},
+	// Note: The '/wp-*' pattern is to ignore symlinks which may be added for local development.
+	ignorePatterns: [ '/vendor', '/node_modules', '/wp-*' ],
 };
 
 module.exports = config;
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,7 +8,6 @@ const config = {
 	rules: {
 		...( wpConfig?.rules || {} ),
 		'jsdoc/valid-types': 'off',
-		'no-console': 'off',
 	},
 	env: {
 		browser: true,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,7 +15,11 @@ const config = {
 	globals: {
 		scheduler: false,
 	},
-	ignorePatterns: [ '/vendor', '/node_modules' ],
+	ignorePatterns: [
+		'/vendor',
+		'/node_modules',
+		'/modules/images/webp-uploads/fallback.js', // TODO: Issues need to be fixed here.
+	],
 };
 
 module.exports = config;
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,7 +8,14 @@ const config = {
 	rules: {
 		...( wpConfig?.rules || {} ),
 		'jsdoc/valid-types': 'off',
+		'no-console': 'off',
 	},
+	env: {
+		'browser': true,
+	},
+	globals: {
+		scheduler: false,
+	}
 };
 
 module.exports = config;
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,3 @@ temp/
 ._*
 .Trashes
 .svn
-
-# Possible symlinks to wp-env install-path directories.
-/wp-*
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ temp/
 ._*
 .Trashes
 .svn
+
+# Possible symlinks to wp-env install-path directories.
+/wp-*
--- a/modules/images/image-loading-optimization/detect.js
+++ b/modules/images/image-loading-optimization/detect.js
@@ -14,26 +14,6 @@ function warn( ...message ) {
 }
 
 /**
- * Yield to the main thread.
- *
- * @see https://developer.chrome.com/blog/introducing-scheduler-yield-origin-trial/#enter-scheduleryield
- * @return {Promise<void>}
- */
-function yieldToMain() {
-	if (
-		typeof scheduler !== 'undefined' &&
-		typeof scheduler.yield === 'function'
-	) {
-		return scheduler.yield();
-	}
-
-	// Fall back to setTimeout:
-	return new Promise( ( resolve ) => {
-		setTimeout( resolve, 0 );
-	} );
-}
-
-/**
  * @typedef {Object} Breadcrumb
  * @property {number} index   - Index of element among sibling elements.
  * @property {string} tagName - Tag name.
--- a/modules/images/image-loading-optimization/detect.js
+++ b/modules/images/image-loading-optimization/detect.js
@@ -32,6 +32,36 @@ function yieldToMain() {
 }
 
 /**
+ * @typedef {Object} Breadcrumb
+ * @property {number} index
+ * @property {string} tagName
+ */
+
+/**
+ * Gets breadcrumbs for a given element.
+ *
+ * @param {Element} element
+ * @return {Breadcrumb[]} Breadcrumbs.
+ */
+function getBreadcrumbs( element ) {
+	/** @type {Breadcrumb[]} */
+	const breadcrumbs = [];
+
+	let node = element;
+	while ( node instanceof Element ) {
+		breadcrumbs.unshift( {
+			tagName: node.tagName,
+			index: node.parentElement
+				? Array.from( node.parentElement.children ).indexOf( node )
+				: 0,
+		} );
+		node = node.parentElement;
+	}
+
+	return breadcrumbs;
+}
+
+/**
  * Detect the LCP element, loaded images, client viewport and store for future optimizations.
  *
  * @param {number}  serveTime           The serve time of the page in milliseconds from PHP via `ceil( microtime( true ) * 1000 )`.
@@ -59,6 +89,14 @@ export default async function detect(
 		log( 'Proceeding with detection' );
 	}
 
+	const results = {
+		viewport: {
+			width: window.innerWidth,
+			height: window.innerHeight,
+		},
+		images: [],
+	};
+
 	// TODO: Use a local copy of web-vitals.
 	const { onLCP } = await import(
 		// eslint-disable-next-line import/no-unresolved
@@ -66,13 +104,13 @@ export default async function detect(
 	);
 
 	/** @type {LCPMetricWithAttribution[]} */
-	const lcpCandidates = [];
+	const lcpMetricCandidates = [];
 
 	// Obtain at least one LCP candidate.
 	const lcpCandidateObtained = new Promise( ( resolve ) => {
 		onLCP(
 			( metric ) => {
-				lcpCandidates.push( metric );
+				lcpMetricCandidates.push( metric );
 				resolve();
 			},
 			{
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -59,3 +59,6 @@ temp/
 ._*
 .Trashes
 .svn
+
+# Possible symlinks to wp-env install-path directories.
+/wp-*
```
westonruter: In order to have WP core files indexed in my project, I create symlinks to the directories for the wp-env. For example:

```bash
install_path=$(npm --silent run wp-env install-path)
ln -s "$install_path/WordPress" wp-core
ln -s "$install_path/WordPress-PHPUnit/tests/phpunit" wp-tests-phpunit 
```

swissspidy: If you are using IntelliJ/PhpStorm, have you tried adding those as Include Paths instead (Preferences | Languages & Frameworks | PHP)?

Otherwise maybe this change is better suited for your own global `.gitignore` file.

Interesting though that you point to files from wp-env for this, I usually just point to a local checkout of the `WordPress-develop` repo.

modules/images/image-loading-optimization/detect.js
```diff
@@ -0,0 +1,279 @@
...
+		} );
+
+		// Stop observing as soon as the page scrolls since we only want initial-viewport elements.
+		win.addEventListener( 'scroll', disconnectIntersectionObserver, {
+			once: true,
+			passive: true,
+		} );
+	}
+
+	// TODO: Use a local copy of web-vitals.
```
swissspidy, line 204: Let me know if you'd like help with that

westonruter: ~This todo isn't to be done in this PR. I'm working on this in https://github.com/WordPress/performance/pull/878.~

westonruter: Oops. I thought you commented on a different todo :)

westonruter: I was presuming that there would be a JS build process introduced to the plugin as part of #864 and so we could incorporate the use of a local copy of web-vitals as part of that. But actually it looks like it's not going to involve JS. If you think a JS build process could be incorporated as part of the PR, I'd appreciate your help. But otherwise I think this can be tackled in a subsequent PR.

modules/images/image-loading-optimization/detect.js
```diff
@@ -0,0 +1,279 @@
...
+			intersectionRatio: elementIntersection.intersectionRatio,
+			intersectionRect: elementIntersection.intersectionRect,
+			boundingClientRect: elementIntersection.boundingClientRect,
+		};
+
+		pageMetrics.elements.push( elementMetrics );
+	}
+
+	// TODO: Send data to server.
+	log( pageMetrics );
```
westonruter, line 287: This is to be done in another PR. See https://github.com/WordPress/performance/pull/878.

.gitignore
westonruter: Yeah, I've done that in the past but this has seemed a bit simpler, and it doesn't depend on having to add the include paths to the settings. Nevertheless, I'm happy to switch to using include paths this if it's only useful to me. I've reverted in 73dbfd1675d114f8989ef6e41ce1a38f914278f6. 

> Interesting though that you point to files from wp-env for this, I usually just point to a local checkout of the `WordPress-develop` repo.

But then the files aren't the actual files in the environment, right? Then you can't hack on the core files for debugging, right?

westonruter: Actually, it's much nicer to have those symlinks after trying without. I can easily navigate to `wp-content/plugins`, for example, to drop in something for testing, for example.

I'm now just putting those symlinks inside of `vendor`.

swissspidy (approved):

.eslintrc.js
```diff
@@ -8,7 +8,15 @@ const config = {
 	rules: {
 		...( wpConfig?.rules || {} ),
 		'jsdoc/valid-types': 'off',
+		'no-console': 'off',
```
joemcgill: Not sure if it's a good idea to turn this off for the whole plugin. Perhaps we can just disable it where a `console.log` is expected instead?

modules/images/image-loading-optimization/detect.js
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	/** @type {PageMetrics} */
+	const pageMetrics = {
+		viewport: {
+			width: win.innerWidth,
+			height: win.innerHeight,
+		},
+		elements: [],
+	};
```
joemcgill, line 251: Curious how you're thinking of using various viewport dimensions in the data when applying optimizations?

joemcgill (approved):
Couple of questions, but generally this looks like a good start. No reason to block this for the feature branch.

modules/images/image-loading-optimization/detect.js
westonruter: If it turns out that that there are different LCP images for different viewports, then `fetchpriority` would not be added to the `img` elements themselves. Instead the `viewport` widths captured here would then be used to construct the `media` attribute of `preload` links so that only the relevant image URL is prioritized for the given viewport. In this way, the same markup can be served to both desktop and mobile, while both still get optimized.

.eslintrc.js
westonruter: Done in 912abef

modules/images/image-loading-optimization/detect.js
felixarntz: @westonruter Just looking through these changes now, I think it would be good to open a separate issue/PR for introducing a build process for JS. This can be against the feature branch because no other module needs it so far, and I don't anticipate any new module being introduced in the near future where it would be needed. So we can do that as part of this one. Let's just make sure the approach could be used by other modules too.

modules/images/image-loading-optimization/hooks.php
```diff
@@ -46,3 +46,36 @@ static function ( $output ) {
...
+function image_loading_optimization_print_detection_script() {
+	$serve_time = ceil( microtime( true ) * 1000 );
+
+	/**
+	 * Filters the time window between serve time and run time in which loading detection is allowed to run.
+	 *
+	 * Allow this amount of milliseconds between when the page was first generated (and perhaps cached) and when the
+	 * detect function on the page is allowed to perform its detection logic and submit the request to store the results.
+	 * This avoids situations in which there is missing detection metrics in which case a site with page caching which
+	 * also has a lot of traffic could result in a cache stampede.
```
felixarntz, line 62: ❤️

Issue #869, opened 2023-10-24
Optimization Detective: Server-applied optimizations informed by client-side detection


westonruter:
## Feature Description

See exploration document: [Image Loading Optimization via Client-side Detection](https://docs.google.com/document/d/16qAJ7I_ljhEdx2Cn2VlK7IkiixobY9zNn8FXxN9T9Ls/edit?usp=sharing)

Work on this is contained in the `feature/image-loading-optimization` branch. See issues/PRs with the [[Plugin] Optimization Detective](https://github.com/WordPress/performance/labels/%5BPlugin%5D%20Optimization%20Detective) label.

For more context, see https://github.com/WordPress/performance/issues/117

What follows are the main issues to achieve the MVP. **For ongoing business-as-usual issues to pull from, see the [labelled issues](https://github.com/WordPress/performance/labels/%5BPlu*gin%5D%20Optimization%20Detective).**

## MVP Issues

- [x] #805
  - [x] #875
- [x] Detecting: #870
  - [x] #876
  - [x] https://github.com/WordPress/performance/issues/893 (TODO: Make sure web-vitals.js is included in the distribution.)
  - [x] https://github.com/WordPress/performance/issues/894
  - [ ] Improve normalization of URL or query args to avoid infinite space. See https://github.com/WordPress/performance/issues/1467.
- [x] Storing: #871
  - [x] #878
  - [x] https://github.com/WordPress/performance/issues/895
  - [ ] Deal with cache invalidation for archive pages and search results: disable for search results and include post IDs from The Loop on archives ([comment](https://github.com/WordPress/performance/pull/878#discussion_r1393132794)). See https://github.com/WordPress/performance/issues/1466.
  - [x] Revisit how URL and slug are handled ([comment](https://github.com/WordPress/performance/pull/988#discussion_r1506781135)): https://github.com/WordPress/performance/pull/1043.
- [x] Optimizing: #872
  - [x] https://github.com/WordPress/performance/pull/884
  - [x] https://github.com/WordPress/performance/pull/892
  - [x] Add `loading=lazy` to any image which is not visible across all breakpoints (cf. [comment](https://github.com/WordPress/performance/pull/884#discussion_r1412694525)). See https://github.com/WordPress/performance/issues/1077
  - [ ] Add `fetchpriority=low` to images hidden but still "in" the initial viewport in which case it should get `fetchpriority=low` (as it could be in a header mega menu or perhaps some later carousel slide): https://github.com/WordPress/performance/issues/1309
[...]

PR #878, opened 2023-11-02
Add storage for URL metrics to optimize image loading

westonruter:
## Summary

See https://github.com/WordPress/performance/issues/871.

## Relevant technical choices

A new `ilo_url_metrics` post type is introduced which stores the client-submitted URL metrics as JSON in `post_content`, similar to the `customize_changeset` and `oembed_cache` post types. The `post_name` is a MD5 hash of the query vars from the `$wp->query_vars` global after normalization. (Using the WP query vars is safer than using the URL due to random additional query vars added by 3rd parties, such as analytics.)

Collected URL metrics are segmented into viewport breakpoints. By default it attempts to collect a sample size of three URL metrics for each breakpoint. The default breakpoint is 480, which means there will be a total of 6 URL metrics collected: 3 for mobile for viewport widths from 0 to 480, and 3 for desktop for viewport widths larger. The breakpoints can be modified with the `ilo_breakpoint_max_widths` filter, and the sample size can be filtered with `ilo_url_metrics_breakpoint_sample_size`. Each URL metric is stored with the timestamp at which it was captured. When a URL metric is older than a given freshness TTL (filtered by `ilo_url_metric_freshness_ttl`), a new URL metric will be attempted to be measured for that viewport breakpoint. New URL metrics for a given breakpoint push out the oldest URL metrics for that breakpoint.

When a client sends a URL metric, that client is locked from sending another URL metric for one minute (filtered by `ilo_metrics_storage_lock_ttl`). Client-side logic prevents another request from being made via `sessionStorage`, while a server-side check with the REST API is also done using the client IP address. Similarly, if no URL metrics are needed for the current viewport (given the breakpoints), no measurement will be performed and no data will be sent; likewise, the server will also reject any submission for viewport URL metrics it doesn't need.

REST API requests are allowed by unauthenticated visitors, but the parameters of the request (namely the WP query vars that the URL metrics are for) are authenticated with an HMAC via a WP nonce.

Other points:

* The `image_loading_optimization_` prefix has been shorted to `ilo_`.
* Files have been renamed for better organization. Namely, logic in `hooks.php` has been moved to `detection.php`.
* Unit test coverage is not included in this PR. I intend to include this after the POC for optimization is developed.

To test facilitate testing, turn on `WP_DEBUG` and playing around with filters to disabling locking, zeroing out the freshness TTL, and changing the breakpoint sample size, for example:

```php
add_filter( 'ilo_url_metric_storage_lock_ttl', function () { return 0; } );
add_filter( 'ilo_url_metric_freshness_ttl', function () { return 0; } );
add_filter( 'ilo_url_metrics_breakpoint_sample_size', function () { return 5; } );
```

## Checklist


modules/images/image-loading-optimization/rest-api.php
```diff
@@ -0,0 +1,115 @@
...
+		),
+	);
+
+	register_rest_route(
+		'perflab/v1',
+		'/image-loading-optimization/metrics-storage',
+		array(
+			'methods'             => 'POST',
+			'callback'            => 'image_loading_optimization_handle_rest_request',
+			'permission_callback' => '__return_true', // Needs to be available to unauthenticated visitors.
```
swissspidy: If it's going to be public, this probably needs some other kind of protection to prevent being flooded with entries.

Years ago I built something similar where multiple requests were blocked by storing the IP address in a transient. Maybe it's useful here:

https://github.com/wearerequired/rest-likes/blob/8c522be82ca468622d287c7a8539ea1dc72e45e9/classes/Controller.php#L251-L253
https://github.com/wearerequired/rest-likes/blob/8c522be82ca468622d287c7a8539ea1dc72e45e9/classes/Controller.php#L282-L353

modules/images/image-loading-optimization/rest-api.php
```diff
@@ -0,0 +1,115 @@
...
+			'args'                => array(
```
swissspidy: Is the URL also passed as an argument?

westonruter: Good point! 😊  However, this could also obtained via the `Referer` HTTP request header. But probably better to be explicit, even though perhaps `Referer` would be a bit safer to rely on. If nothing else, a custom validation callback could be used for the `url` arg to make sure it matches the `Referer`.

modules/images/image-loading-optimization/rest-api.php
westonruter: Thank you! Yes, I was also intending to do something similar following the same pattern as WP comments.

modules/images/image-loading-optimization/rest-api.php
westonruter: Added `url` in b4e29d609647c8e23fc4db6532e50ba8692e82c5

modules/images/image-loading-optimization/detection.php
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @todo This should eventually only print the script if metrics are needed.
+ * @todo This script should not be printed if the page was requested with non-removal (non-canonical) query args.
+ */
+function ilo_print_detection_script() {
+
+	// TODO: Also abort if we don't need any new page metrics due to the sample size being full.
+	if ( ilo_is_page_metric_storage_locked() ) {
+		return;
+	}
```
felixarntz: Is this the only change in this function over #876? Just asking since it's hard to spot because the function was moved to another file.

modules/images/image-loading-optimization/detection.php
```diff
@@ -0,0 +1,59 @@
...
+		$serve_time,
+		$detection_time_window,
+		WP_DEBUG,
+		rest_url( ILO_REST_API_NAMESPACE . ILO_PAGE_METRIC_STORAGE_ROUTE ),
+		wp_create_nonce( 'wp_rest' ),
+	);
+	wp_print_inline_script_tag(
+		sprintf(
+			'import detect from %s; detect( ...%s )',
+			wp_json_encode( add_query_arg( 'ver', PERFLAB_VERSION, plugin_dir_url( __FILE__ ) . 'detection/detect.js' ) ),
```
felixarntz: This should use the constant specific to this module, as `PERFLAB_VERSION` may not be defined (as this is going to be a standalone plugin eventually).

modules/images/image-loading-optimization/detection.php
```diff
@@ -0,0 +1,59 @@
...
+	 */
+	$detection_time_window = apply_filters( 'perflab_image_loading_detection_time_window', 5000 );
+
+	$detect_args = array(
+		$serve_time,
+		$detection_time_window,
+		WP_DEBUG,
+		rest_url( ILO_REST_API_NAMESPACE . ILO_PAGE_METRIC_STORAGE_ROUTE ),
+		wp_create_nonce( 'wp_rest' ),
+	);
```
felixarntz: Nit-pick, but providing these as a list is a bit error-prone. Maybe better to use an associative array here so that we have named entries, and then in JS we can expect an object and use destructuring instead?

The way it currently is, with the order that matters, it's easy to make accidental mistakes here.

modules/images/image-loading-optimization/detection/detect.js
```diff
@@ -283,8 +299,33 @@ export default async function detect(
 		pageMetrics.elements.push( elementMetrics );
 	}
 
-	// TODO: Send data to server.
-	log( pageMetrics );
+	if ( isDebug ) {
+		log( 'Page metrics:', pageMetrics );
+	}
+
+	// TODO: Wait until idle? Yield to main?
```
felixarntz, line 406: This would be great. Maybe even the whole detection logic should have low priority and give precedence to other tasks? We'll want to make sure this script doesn't make e.g. INP worse, as this is probably one of the least important JS pieces to run for the end user.

We definitely don't have to solve that here, but could you open an issue to implement that eventually?

modules/images/image-loading-optimization/load.php
```diff
@@ -9,16 +9,17 @@
  */
 
 // Define the constant.
-if ( defined( 'IMAGE_LOADING_OPTIMIZATION_VERSION' ) ) {
+if ( defined( 'ILO_VERSION' ) ) {
 	return;
 }
 
-define( 'IMAGE_LOADING_OPTIMIZATION_VERSION', 'Performance Lab ' . PERFLAB_VERSION );
+define( 'ILO_VERSION', 'Performance Lab ' . PERFLAB_VERSION );
```
felixarntz: Maybe keep this name fully qualified, i.e. don't change it? The constants are mostly for external use so I think the full name is better here. Within the module itself, you'll probably only use it in very few places such as asset versioning (although even that could maybe be replaced once we have a build process).

modules/images/image-loading-optimization/load.php
```diff
@@ -9,16 +9,17 @@
...
 
 // Do not load the code if it is already loaded through another means.
-if ( function_exists( 'image_loading_optimization_buffer_output' ) ) {
+if ( function_exists( 'ilo_buffer_output' ) ) {
 	return;
 }
```
felixarntz: I think I had mentioned that in the initial PR, but I think this if clause can be safely removed, as the constant check on top of this file already covers for that.

modules/images/image-loading-optimization/storage.php
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+require_once __DIR__ . '/storage/lock.php';
+require_once __DIR__ . '/storage/post-type.php';
+require_once __DIR__ . '/storage/data.php';
+require_once __DIR__ . '/storage/rest-api.php';
```
felixarntz: A bit strange to have a file just to load other files. Can you move these lines to the module's `load.php` file? Also makes it easier to spot and maintain in one place where the files are loaded.

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -0,0 +1,125 @@
...
+	/**
+	 * Filters the breakpoint max widths to group page metrics for various viewports.
+	 *
+	 * @param int[] $breakpoint_max_widths Max widths for viewport breakpoints.
+	 */
+	$breakpoint_max_widths = array_map(
+		static function ( $breakpoint_max_width ) {
+			return (int) $breakpoint_max_width;
+		},
+		(array) apply_filters( 'ilo_breakpoint_max_widths', array( 480 ) )
```
felixarntz, line 205: What is the 480 based on? Is that maybe a bit small? It would be good to clarify whether it's based on some other authoritative source.

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Gets desired sample size for a breakpoint's page metrics.
+ *
+ * @return int Sample size.
+ */
+function ilo_get_page_metrics_breakpoint_sample_size() {
+	/**
+	 * Filters desired sample size for a viewport's page metrics.
+	 *
+	 * @param int $sample_size Sample size.
```
felixarntz: Can you add another explanation to the filter doc for what "sample size" is? The name itself doesn't necessarily clarify that. To be clear, I think the name is perfectly fine, but still probably needs an explanation.

Also isn't this the sample size for "breakpoint _plus URL_"?

modules/images/image-loading-optimization/storage/lock.php
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/**
+ * Gets transient key for locking page metric storage (for the current IP).
+ *
+ * @todo Should the URL be included in the key? Or should a user only be allowed to store one metric?
+ * @return string Transient key.
+ */
+function ilo_get_page_metric_storage_lock_transient_key() {
+	$ip_address = $_SERVER['HTTP_X_FORWARDED_FOR'] ?? $_SERVER['REMOTE_ADDR'];
```
felixarntz, line 48: Not too familiar with the `??` operator, is this safe to use even if `HTTP_X_FORWARDED_FOR` was not set? In other words, is this safe without an `isset()`?

modules/images/image-loading-optimization/storage/lock.php
```diff
@@ -0,0 +1,69 @@
...
+	return 'page_metrics_storage_lock_' . wp_hash( $ip_address );
+}
```
felixarntz: This will create _a lot_ of transients, so I'm a bit concerned about that. Is there another way to handle this? I realize grouping may have other issues such as race conditions. But still, this approach also makes me feel uneasy 🤔

modules/images/image-loading-optimization/storage/post-type.php
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * This the configuration for this post type is similar to the oembed_cache in core.
+ */
+function ilo_register_page_metrics_post_type() {
+	register_post_type(
+		ILO_PAGE_METRICS_POST_TYPE,
+		array(
+			'labels'           => array(
+				'name'          => __( 'Page Metrics', 'performance-lab' ),
+				'singular_name' => __( 'Page Metrics', 'performance-lab' ),
```
felixarntz: Should this be called "URL metrics", to avoid confusion with the specific WordPress post type "page"?

modules/images/image-loading-optimization/storage/rest-api.php
```diff
@@ -0,0 +1,158 @@
...
+ * @since n.e.x.t
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+const ILO_REST_API_NAMESPACE = 'image-loading-optimization/v1';
+
+const ILO_PAGE_METRIC_STORAGE_ROUTE = '/image-loading-optimization/page-metric-storage';
```
felixarntz: Two points of feedback on the route name:
* The `image-loading-optimization` is already in the namespace, so I think we should drop it from the route name.
* To better follow [resource-oriented design](https://google.aip.dev/121), I would advise we center the naming around the resource, which is the "page metrics". I realize we don't need to provide API endpoints for CRUD operations on the individual custom post type entries for now, but we could choose a name here that would allow ourselves to do that in the future: We could follow [this custom methods recommendation](https://google.aip.dev/136) and go with:
```suggestion
const ILO_PAGE_METRIC_STORAGE_ROUTE = '/page-metrics:store';
```

modules/images/image-loading-optimization/storage/rest-api.php
```diff
@@ -0,0 +1,158 @@
...
+	);
+
+	register_rest_route(
+		ILO_REST_API_NAMESPACE,
+		ILO_PAGE_METRIC_STORAGE_ROUTE,
+		array(
+			'methods'             => 'POST',
+			'callback'            => static function ( WP_REST_Request $request ) {
+				return ilo_handle_rest_request( $request );
+			},
```
felixarntz, line 62: Nit-pick:
```suggestion
			'callback'            => 'ilo_handle_rest_request',
```

felixarntz (commented):
@westonruter I realize this is still in draft, but I already took a first look and left a few comments. Looking forward to try it eventually!

modules/images/image-loading-optimization/storage/rest-api.php
westonruter: Actually I did it this way intentionally for better static analysis.

modules/images/image-loading-optimization/storage/lock.php
westonruter: I don't think it would actually create that many transients because the intention is that metrics would only be gathered when they are needed. So if no metrics are needed, then a frontend request would not create a transient. I believe in the worst case that there would be the total number of transients would be the sample size times the total number of URLs.

modules/images/image-loading-optimization/storage/lock.php
westonruter: It's the [null coalescing operator](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op), new in PHP 7. This line is equivalent to:

```php
$ip_address = isset( $_SERVER['HTTP_X_FORWARDED_FOR'] ) ? $_SERVER['HTTP_X_FORWARDED_FOR'] : $_SERVER['REMOTE_ADDR'];
```

So it is using `isset()` under the hood. 

I copied this line from @swissspidy's [implementation](https://github.com/wearerequired/rest-likes/blob/8c522be82ca468622d287c7a8539ea1dc72e45e9/classes/Controller.php#L291).

modules/images/image-loading-optimization/detection.php
westonruter: Yes, no change to this function, sorry. I just reorganized functions better into files better corresponding to their purpose.

modules/images/image-loading-optimization/detection.php
westonruter: Yes, good idea.

modules/images/image-loading-optimization/storage.php
westonruter: The reason I had it like this was that I originally had all the functions in this one file, and it was too hard to manage. So then I split them up. I could just move them to `load.php`, but I thought having them in `storage.php` may make them better organized as part of this "module". I'm happy to move them to `load.php` instead.

modules/images/image-loading-optimization/storage/data.php
westonruter: WordPress's own responsive preview in the block editor is actually even smaller at 360 for mobile. And 480 is wider than all of the widths for mobile devices in Chrome DevTools. I picked 480 from doing various searches, though I can't recall the specific resource I used. Some references for 480 pixels:

* https://www.lambdatest.com/blog/how-to-use-css-breakpoints-for-responsive-design/#:~:text=What%20are%20common%20breakpoints?,extra%20large%20screens%20like%20TV.
* https://testsigma.com/blog/css-breakpoints/#:~:text=While%20there%20is%20no%20hard%20and%20fast%20rule%20for%20what%20breakpoints%20should%20be%20used%2C%20values%20of%20common%20CSS%20breakpoints%20are%20320px%20or%20480px%20(for%20mobile%20phones)%2C%20768px%20(for%20tablets)%2C%20and%201920px%20(for%20desktop%20computers).%20Ultimately%2C%20it%20is%20up%20to%20the%20designer%20to%20decide%20what%20breakpoints%20make%20sense%20for%20their%20project.
* https://testsigma.com/blog/css-breakpoints/#:~:text=from%2C%20or%20to.-,Which%20CSS%20breakpoints%20to%20use%3F,-Although%20every%20device

modules/images/image-loading-optimization/detection.php
westonruter: Correction, there were some other changes. Here's the diff up until I moved the function: https://github.com/WordPress/performance/pull/878/files/79bafac29a0882a1f1f577aab133ee4846a5f0c1#diff-c7d8aefac2b952f0c1a6c0dd7ae0eef789df564cb76df63e5045ab7d867783f7

modules/images/image-loading-optimization/storage/rest-api.php
westonruter: Oops, missed the redundancy. I think it's because I originally had the namespace for PerfLab.

As for the route, how about just `/page-metrics`? (Or `/url-metrics`?) I don't think we need to incorporate any custom method here as the resource is `/page-metrics` and the method of `POST` is sufficiently semantic. And WP REST API routes don't typically include colons like that, as far as I'm aware.

modules/images/image-loading-optimization/storage/rest-api.php
westonruter: Namely, the `ilo_handle_rest_request` function otherwise is not detected as being used if it is just passed as a string. In PHP 8.1 this can be avoided with [first class callable syntax](https://www.php.net/manual/en/functions.first_class_callable_syntax.php) like so:

```suggestion
			'callback'            => ilo_handle_rest_request(...),
```

modules/images/image-loading-optimization/storage/post-type.php
westonruter: Yeah, I think "URL metrics" works just as well as "page metrics". And it would be less confusing to WP users.

modules/images/image-loading-optimization/storage/data.php
westonruter: My idea here is that it's the sample size for the page metrics for a given breakpoint for a given URL. So let's say that the sample size is 3 and there is one breakpoint at 480. Then for a given URL, there would be a total of 6 page metrics stored for a given URL: 3 for mobile and 3 for desktop. I will flesh that out in the phpdoc.

modules/images/image-loading-optimization/load.php
westonruter: Done in a17c9a267fe4a31d071f0e59b957ff2f51d1199d

modules/images/image-loading-optimization/load.php
westonruter: Done in a65d0be988ad313f3c851c5bb66e662a1a63d126

modules/images/image-loading-optimization/storage.php
westonruter: Done in 7f6cc2b

modules/images/image-loading-optimization/storage/rest-api.php
westonruter: Fixed in 7d5b2f5

modules/images/image-loading-optimization/detection.php
westonruter: OK, so should the `IMAGE_LOADING_OPTIMIZATION_VERSION` constant then not use the `PERFLAB_VERSION` either? And then should `IMAGE_LOADING_OPTIMIZATION_VERSION` be used here?

felixarntz: @westonruter `PERFLAB_VERSION` should only be referenced in `IMAGE_LOADING_OPTIMIZATION_VERSION` actually. It will be replaced by a script with the relevant standalone version when creating the standalone plugin build.

westonruter: So this should be modified to just use `IMAGE_LOADING_OPTIMIZATION_VERSION`?

```suggestion
			wp_json_encode( add_query_arg( 'ver', IMAGE_LOADING_OPTIMIZATION_VERSION, plugin_dir_url( __FILE__ ) . 'detection/detect.js' ) ),
```

modules/images/image-loading-optimization/detection.php
westonruter: Done in 80c6827

modules/images/image-loading-optimization/detection/detect.js
westonruter: Added _todo el_ todo to https://github.com/WordPress/performance/issues/869

modules/images/image-loading-optimization/storage/data.php
westonruter: Fleshed out phpdoc in f6208c20ba799d74b2eced8a161145d68cb5c551

modules/images/image-loading-optimization/detection/detect.js
```diff
@@ -84,22 +132,76 @@ function getBreadcrumbs( leafElement ) {
...
+ * @param {Object}                   args                             Args.
+ * @param {number}                   args.serveTime                   The serve time of the page in milliseconds from PHP via `ceil( microtime( true ) * 1000 )`.
+ * @param {number}                   args.detectionTimeWindow         The number of milliseconds between now and when the page was first generated in which detection should proceed.
+ * @param {boolean}                  args.isDebug                     Whether to show debug messages.
+ * @param {string}                   args.restApiEndpoint             URL for where to send the detection data.
+ * @param {string}                   args.restApiNonce                Nonce for writing to the REST API.
+ * @param {string}                   args.pageMetricsSlug             Slug for page metrics.
+ * @param {string}                   args.pageMetricsNonce            Nonce for page metrics storage.
+ * @param {Array<number, boolean>[]} args.neededMinimumViewportWidths Needed minimum viewport widths for page metrics.
+ * @param {number}                   args.storageLockTTL              The TTL (in seconds) for the page metric storage lock.
```
swissspidy: The comparison is done in microseconds, so should this value perhaps be provided in microseconds as well?

modules/images/image-loading-optimization/detection/detect.js
```diff
@@ -84,22 +132,76 @@ function getBreadcrumbs( leafElement ) {
...
+}
+
+/**
+ * Gets the current time in milliseconds.
+ *
+ * @return {number} Current time in milliseconds.
+ */
+function getCurrentTime() {
+	return new Date().valueOf();
+}
```
swissspidy: To prevent issues with timezones and other kinds of clock adjustments, WDYT about using `performance.now()` instead?

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -0,0 +1,259 @@
...
+	// assembled by iterating over public_query_vars.
+	$normalized_query_vars = $wp->query_vars;
+
+	// Normalize unbounded query vars.
+	if ( is_404() ) {
+		$normalized_query_vars = array(
+			'error' => 404,
+		);
+	} elseif ( array_key_exists( 's', $normalized_query_vars ) ) {
+		$normalized_query_vars['s'] = '...';
```
swissspidy: Does that mean every search query will just use the same cache? Seems a bit error prone

modules/images/image-loading-optimization/storage/lock.php
swissspidy: Yep, basically syntactic sugar for `isset` with ternary.

modules/images/image-loading-optimization/storage/rest-api.php
swissspidy: @westonruter Alternatively you could just add `@uses ilo_handle_rest_request` to the `ilo_register_endpoint` docblock

modules/images/image-loading-optimization/storage/post-type.php
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * The $validated_page_metric parameter has the following array shape:
+ *
+ * {
+ *      'viewport': array{
+ *          'width': int,
+ *          'height': int
+ *      },
+ *      'elements': array
+ * }
```
swissspidy: WordPress doc standards have a way to describe array shapes within the parameter description. See https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#1-1-parameters-that-are-arrays

modules/images/image-loading-optimization/storage/data.php
westonruter: Yes, because it's the template that we're concerned about. What's needed here is to map query vars to a consistent template layout where the image breadcrumbs/selectors will reliably match. This should be the case for any search query, as there aren't any template subtypes for the search template.

modules/images/image-loading-optimization/detection/detect.js
westonruter: AFAIK, `new Date().valueOf()` returns the milliseconds since January 1st, 1970 UTC. So it is not impacted by timezone, is it?

modules/images/image-loading-optimization/detection/detect.js
westonruter: Milliseconds not microseconds, right? As `new Date().valueOf()` returns milliseconds. The server is using `ceil( microtime( true ) * 1000 )` (which is milliseconds). (Though it is also using `time()` which I need to reconcile.)

I thought that seconds for the TTL would be the adequate granularity for developers. I can't imagine a site wanting to filter the TTL to be anything other than whole seconds, but I'm happy to change this.

modules/images/image-loading-optimization/detection/detect.js
westonruter: Also, `performance.now()` returns not the current time but the current time in milliseconds since the start of the page load.

modules/images/image-loading-optimization/storage/rest-api.php
westonruter: @swissspidy Good point. Although that seems more error prone as it doesn't specify _how_ the function is used including what parameters are passed to it.

modules/images/image-loading-optimization/detection/detect.js
swissspidy: Eh, yes, I meant milliseconds 🤦 

Fair enough 👍 Was just a thought.

modules/images/image-loading-optimization/detection/detect.js
swissspidy: > Also, `performance.now()` returns not the current time but the current time in milliseconds since the start of the page load.

Which is fine because from what I can see you only need relative time anyway for the lock, no?

modules/images/image-loading-optimization/storage/data.php
swissspidy: Just so I better understand: So it doesn't matter when a search results page sometimes shows only 1 post with no image, and sometimes it shows a post with an LCP image?

modules/images/image-loading-optimization/detection/detect.js
westonruter: The lock is captured at a specific `time()` not from the start of the page load time. When the REST API request to update page metrics is successful, it sets a lock for the IP at the current time. So we need to compare against that time.

I've refined time handling a bit in 6cd8198532b7076770d93b2003364949d0d3030f.

modules/images/image-loading-optimization/storage/data.php
westonruter: Oh yes, you're absolutely right. That is problematic. The goal behind normalization is to prevent filling up the `ilo_page_metric` posts with garbage. But it is certainly possible that different search results will result in a random assortment of posts that do and do not have featured images. 

Indeed, this will also be a problem for archive pages in general. One way to deal with this would be to include the post IDs in The Loop among the query vars so that a new post being published would effectively invalidate the cache for page metrics. But I don't think this makes sense to do for search results. Perhaps image loading optimization should just be disabled for search results. What do you think?

westonruter: I've disabled optimizing search results entirely in 99ef2996caf805cf2ca0d253374ee66e901cbf84 and added a `ilo_can_optimize_response` filter which can be used to override which responses get optimized.

modules/images/image-loading-optimization/storage/post-type.php
westonruter: Yeah, I was here originally trying to port the JSON schema over to an [array shape](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes) in PHPStan for static analysis. WordPress's convention isn't supported. Since the array shape isn't needed for documentation here and since there is no type checking, I've just removed it in favor referencing the JSON Schema which does define it: a4ffbaef191b7d2b859c0a8bf08834fd57defedf

swissspidy: That works too I suppose :-)

If it helps, what I usually do for these cases is follow WP‘s convention and then use a separate phpstan-param annotation for type checking.

modules/images/image-loading-optimization/storage/data.php
swissspidy: Sounds like a good compromise 👍

modules/images/image-loading-optimization/storage/rest-api.php
swissspidy: Right. PHPStan would help with detailed type checks.
The annotation just addresses the concern about it being marked as unused.

modules/images/image-loading-optimization/detection/detect.js
```diff
@@ -5,6 +5,43 @@ const doc = win.document;
 
 const consoleLogPrefix = '[Image Loading Optimization]';
 
+const storageLockTimeSessionKey = 'iloStorageLockTime';
+
+/**
+ * Checks whether storage is locked.
+ *
+ * @param {number} currentTime    - Current time in milliseconds.
```
adamsilverstein, line 13: the `-` on this line and the next seem unintentional?

westonruter: It's intentional, actually:

> You can add a hyphen before the description to make it more readable. Be sure to include a space before and after the hyphen.

https://jsdoc.app/tags-param

modules/images/image-loading-optimization/detection/detect.js
adamsilverstein: maybe even wait for the `pagehide` event?

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -0,0 +1,311 @@
...
+ * Gets the freshness age (TTL) for a given page metric.
+ *
+ * When a page metric expires it is eligible to be replaced by a newer one if its viewport lies within the same breakpoint.
+ *
+ * @return int Expiration TTL in seconds.
+ */
+function ilo_get_page_metric_freshness_ttl() {
+	/**
+	 * Filters the freshness age (TTL) for a given page metric.
+	 *
```
adamsilverstein, line 28: add `@since` tag for the filter (? we do in core)

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -0,0 +1,311 @@
...
+ * TODO: For non-singular requests, consider adding the post IDs from The Loop to ensure publishing a new post will invalidate the cache.
+ *
+ * @return array Normalized query vars.
+ */
+function ilo_get_normalized_query_vars() {
+	global $wp;
+
+	// Note that the order of this array is naturally normalized since it is
+	// assembled by iterating over public_query_vars.
+	$normalized_query_vars = $wp->query_vars;
```
adamsilverstein, line 76: I wonder about plugins that might alter these, maybe we should limit to a preset list? or at least check what plugins might add query_vars that would be picked up here

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -0,0 +1,311 @@
...
+		if ( count( $breakpoint_page_metrics ) > $sample_size ) {
+
+			// Sort page metrics in descending order by timestamp.
+			usort(
+				$breakpoint_page_metrics,
+				static function ( $a, $b ) {
+					if ( ! isset( $a['timestamp'] ) || ! isset( $b['timestamp'] ) ) {
+						return 0;
+					}
+					return $b['timestamp'] <=> $a['timestamp'];
```
adamsilverstein, line 166: nice!

westonruter: First time I've ever gotten to use this 😄

modules/images/image-loading-optimization/storage/data.php
westonruter: Yeah, all of the functions and filters will need `@since` tags eventually. I guess I could add `@since n.e.x.t` everywhere now.

modules/images/image-loading-optimization/storage/data.php
adamsilverstein: i had to look it up to remember what it does :)

modules/images/image-loading-optimization/storage/data.php
adamsilverstein: i was going to ask the same question as I am sure others will so it would be good to eventually have a concise answer for why the specific values were chosen (and in what instances you might want to alter them)

modules/images/image-loading-optimization/storage/data.php
westonruter: You mean beyond filtering `query_vars` to register additional public query vars? I should think that anything that is populating the `query_vars` is going to be essential for varying the cache for stored detection data. It will also automatically exclude arbitrary query vars added by analytics, such as `utm_*`. Granted, themes and plugins may also customize templates based on arbitrary query parameters being read directly from `$_GET`, so in this case the cache could be polluted. It's impossible to know which query parameters are going to impact the rendering of the page and which are just there for client-side purposes.

modules/images/image-loading-optimization/storage/post-type.php
```diff
@@ -0,0 +1,181 @@
...
+}
+add_action( 'init', 'ilo_register_page_metrics_post_type' );
+
+/**
+ * Get page metrics post.
+ *
+ * @param string $slug Page metrics slug.
+ * @return WP_Post|null Post object if exists.
+ */
+function ilo_get_page_metrics_post( string $slug ) /*: ?WP_Post (in PHP 7.1) */ {
```
adamsilverstein: why does this say "(in PHP 7.1)"?

westonruter: Because this is only syntactically valid in PHP 7.1 ([3v4l](https://3v4l.org/YGSZf)):

```php
function ilo_get_page_metrics_post( string $slug ): ?WP_Post {
    // ...
}
```

westonruter: Maybe I went over the top with adding PHP type declarations, but in the future the comments could be removed when WP is bumped to that version of PHP.

modules/images/image-loading-optimization/storage/post-type.php
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * Gets page metrics for a slug.
+ *
+ * This is a convenience abstractions for lower-level functions.
+ *
+ * @see ilo_get_page_metrics_post()
+ * @see ilo_parse_stored_page_metrics()
+ *
+ * @param string $slug Page metrics slug.
+ * @return array Page metrics data, or empty array if invalid.
```
adamsilverstein: curious why you decided to return an empty array if invalid, I would expect an error

adamsilverstein (approved):
Looking good!

modules/images/image-loading-optimization/storage/post-type.php
adamsilverstein: Ah, right - thanks.

modules/images/image-loading-optimization/storage/data.php
adamsilverstein: Right, good point, I guess you need to consider them all.

modules/images/image-loading-optimization/detection/detect.js
adamsilverstein: Ah, I guess I hadn't seen it used before so it looked like a typo 🤷🏼

modules/images/image-loading-optimization/detection.php
felixarntz: @westonruter Yes, correct.

modules/images/image-loading-optimization/storage/data.php
felixarntz: Thanks both, for now 480 is good IMO. We can always adjust later, and there's a filter anyway.

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -0,0 +1,311 @@
...
+ *
+ * @return int[] Breakpoint max widths, sorted in ascending order.
+ */
+function ilo_get_breakpoint_max_widths(): array {
+
+	/**
+	 * Filters the breakpoint max widths to group page metrics for various viewports.
+	 *
+	 * @param int[] $breakpoint_max_widths Max widths for viewport breakpoints.
+	 */
```
felixarntz: This doc block is now in the wrong place :)

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -0,0 +1,311 @@
...
+ *
+ * @param array $query_vars Normalized query vars.
+ * @return string Slug.
+ */
+function ilo_get_page_metrics_slug( array $query_vars ): string {
+	return md5( wp_json_encode( $query_vars ) );
+}
+
+/**
+ * Compute nonce for storing page metrics for a specific slug.
```
felixarntz:
```suggestion
 * Computes nonce for storing page metrics for a specific slug.
```

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -0,0 +1,311 @@
...
+ *
+ * @param string $slug Page metrics slug.
+ * @return string Nonce.
+ */
+function ilo_get_page_metrics_storage_nonce( string $slug ): string {
+	return wp_create_nonce( "store_page_metrics:{$slug}" );
+}
+
+/**
+ * Verify nonce for storing page metrics for a specific slug.
```
felixarntz:
```suggestion
 * Verifies nonce for storing page metrics for a specific slug.
```

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -0,0 +1,311 @@
...
+ * @return int 1 if the nonce is valid and generated between 0-12 hours ago,
+ *             2 if the nonce is valid and generated between 12-24 hours ago.
+ *             0 if the nonce is invalid.
+ */
+function ilo_verify_page_metrics_storage_nonce( string $nonce, string $slug ): int {
+	return (int) wp_verify_nonce( $nonce, "store_page_metrics:{$slug}" );
+}
+
+/**
+ * Unshift a new page metric onto an array of page metrics.
```
felixarntz:
```suggestion
 * Unshifts a new page metric onto an array of page metrics.
```

There may be some more functions not using third-person singular, would be good to double check.

modules/images/image-loading-optimization/detection/detect.js
```diff
@@ -5,6 +5,43 @@ const doc = win.document;
...
+ * @param {number} storageLockTTL - Storage lock TTL in seconds.
+ * @return {boolean} Whether storage is locked.
+ */
+function isStorageLocked( currentTime, storageLockTTL ) {
+	try {
```
felixarntz: Shouldn't we include a check for when the TTL is 0 here? Something like:
```suggestion
function isStorageLocked( currentTime, storageLockTTL ) {
	if ( ! storageLockTTL ) {
		return false;
	}

	try {
```

modules/images/image-loading-optimization/storage/lock.php
```diff
@@ -0,0 +1,69 @@
...
+function ilo_get_page_metric_storage_lock_ttl(): int {
+
+	/**
+	 * Filters how long a given IP is locked from submitting another metric-storage REST API request.
+	 *
+	 * Filtering the TTL to zero will disable any metric storage locking. This is useful during development.
+	 *
+	 * @param int $ttl TTL.
+	 */
+	return (int) apply_filters( 'ilo_metrics_storage_lock_ttl', MINUTE_IN_SECONDS );
```
felixarntz: Maybe we should document that this can also be `0` and that, in that case, transients shouldn't be used at all?

Two related thoughts:
* What is the rationale for allowing `0`?
* If we keep it, we need to triple-check that this value is never simply passed through to `set_transient()`. As far as I can tell, that's currently (rightfully) not the case, but it's a bit error-prone, and in case of an error it would lead to transients without expiration, which ties into my other scalability concern with lots of transients. With expiring transients, a large number can be fine, but not if they remain forever.

I think to harden against the potential problems from my second point, it would be good to make this a private function, so that nobody can call it externally (potentially without knowing these implications).

I think the 4 functions in this file could be easily refactored into e.g. a `ILO_Page_Metric_Storage_Lock` class, where the methods to get the TTL and transient key could be private. The public interface could then simply be the `lock()` and `is_locked()` methods.

Alternatively, we disallow returning `0` via the filter, avoiding the risk in a simple way. We could also reconsider later, combining with potential refactoring to keep use of this function "contained".

modules/images/image-loading-optimization/detection.php
```diff
@@ -0,0 +1,66 @@
+<?php
+/**
+ * Detection for image loading optimization.
+ *
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
```
felixarntz, line 8: There are a lot of functions in this file, many of which seem low-level and would be good to be private.

I think at a minimum, we should mark those functions private that we don't want external code to use. Alternatively, we could refactor into a simple class to actually use private methods, but if so, that should probably be done in a follow up PR to not have to focus on refactoring _and_ the actual logic at the same time.

felixarntz (commented):
@westonruter This looks great. A few additional thoughts, but I think this is almost ready for merge.

modules/images/image-loading-optimization/storage/lock.php
felixarntz: @westonruter Right, that can be _a lot_ of transients though, for sites with lots of content.

I think as long as we keep these transients with an expiration (so that they disappear over time and are not autoloaded as options), it should be okay.

modules/images/image-loading-optimization/detection.php
westonruter: Do you mean the file `storage/data.php`?

And yeah, none of these functions should need to be accessed by developers.

modules/images/image-loading-optimization/storage/data.php
westonruter: Oops! Fixed in caaef87

modules/images/image-loading-optimization/storage/data.php
westonruter: Fixed in eadec5592a8b103e3522fa9f29b75e5b8dcea2e9 and 02dbb359fa726ed889475f9cad66bb362f9b36e8

modules/images/image-loading-optimization/detection/detect.js
westonruter: Yeah, if `storageLockTTL <= 0`.

modules/images/image-loading-optimization/storage/lock.php
westonruter: Yes, it is useful to allow zero, beyond just for development purposes. For example, consider this example:

```php
add_filter( 'ilo_metrics_storage_lock_ttl', static function ( $ttl ) {
    return is_user_logged_in() ? 0 : $ttl;
} );
```

This allows authenticated users in WP to store page metrics without any locking, which could even make sense to do by default. A user who is already authenticated in WP is not going to be the same level of risk to necessitate locking. I've added this as an example in the filter doc in da0b4208b73694d1e1800cb04178140b9837b094.

Also, I'm now ensuring that the TTL is never less than zero.  

As for when the TTL is zero exactly, code is in place to make sure that no transient is set:

https://github.com/WordPress/performance/blob/da0b4208b73694d1e1800cb04178140b9837b094/modules/images/image-loading-optimization/storage/lock.php#L53-L61

So there is no possible way for a transient to be set with a zero expiration. Not really a need to make it private, though I intend to mark all of the functions as private in any case.

modules/images/image-loading-optimization/detection.php
westonruter: I've marked all functions as being private in 19ce752.

modules/images/image-loading-optimization/storage/data.php
westonruter: `@since` tags added in 19ce752

modules/images/image-loading-optimization/detection.php
westonruter: Done in 19a2c3b

modules/images/image-loading-optimization/storage/post-type.php
westonruter: I've reworked this in f566663. Now any errors are emitted from `ilo_parse_stored_page_metrics()` via `wp_trigger_error()` rather than being returned in a `WP_Error`. Since the parse error case should be truly exceptional, and the error recovery case is to just blow the data away and start again with an empty array, I think this makes more sense.

modules/images/image-loading-optimization/storage/post-type.php
```diff
@@ -0,0 +1,169 @@
...
+ * @since n.e.x.t
+ * @access private
+ */
+function ilo_register_page_metrics_post_type() /*: void (in PHP 7.1) */ {
+	register_post_type(
+		ILO_PAGE_METRICS_POST_TYPE,
+		array(
+			'labels'           => array(
+				'name'          => __( 'Page Metrics', 'performance-lab' ),
+				'singular_name' => __( 'Page Metrics', 'performance-lab' ),
```
felixarntz: I think this still needs to be updated per https://github.com/WordPress/performance/pull/878#discussion_r1387245125?

modules/images/image-loading-optimization/storage/rest-api.php
```diff
@@ -0,0 +1,193 @@
...
+ * @since n.e.x.t
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+const ILO_REST_API_NAMESPACE = 'image-loading-optimization/v1';
+
+const ILO_PAGE_METRICS_ROUTE = '/page-metrics';
```
felixarntz: See above, if we changed the post type name, this name should be changed accordingly.

Additionally though, following up on https://github.com/WordPress/performance/pull/878#discussion_r1387484165, I don't think just this name is semantically correct. `POST /page-metrics` would be intended to create a new resource of this type, which isn't really the case as this endpoint doesn't create a new post of the metrics post type, but modifies an existing one with specific data for a request. That is effectively a custom method, so it shouldn't use just the resource based identifier.

Core's endpoints indeed do not include `:` characters, though it's probably not an issue introducing one. But if we want to tie more nicely into core's naming, we should simply use another forward slash to separate the custom method from the resource, i.e. `/page-metrics/store` (or `/url-metrics/store`). There are already a few endpoints in core that use a custom method in this way.

Maybe the `store` is also not super clear, how about `store-request`?

modules/images/image-loading-optimization/storage/lock.php
```diff
@@ -0,0 +1,89 @@
...
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * Gets the TTL (in seconds) for the page metric storage lock.
+ *
+ * @since n.e.x.t
+ * @access private
+ *
+ * @return int TTL in seconds, greater than or equal to zero.
```
felixarntz: Based on https://github.com/WordPress/performance/pull/878#discussion_r1393429581:
```suggestion
 * @return int TTL in seconds, greater than or equal to zero. A value of zero means that the storage lock should be disabled and thus that transients must not be used.
```

modules/images/image-loading-optimization/storage/post-type.php
felixarntz: I'd say this is a bit overkill for now. There are a ton of places like this in core and in plugins, and I don't think it's worth adding this to all of them. So I would say let's stick with what we _can_ add at this point.

modules/images/image-loading-optimization/storage/lock.php
felixarntz: Well, making the function private ensures nobody uses it incorrectly. Obviously we can't truly make it private as long as it's a function, but if someone still uses it, we have no obligation to keep it working for them.

The rationale for this specific function to be private is that it's error-prone to _have_ to remember that you have to always add an extra condition for whether TTL is `0` since otherwise you'd create transients without expiration. That's why I think this function specifically really should be private, or extremely well documented on what should happen in the case it returns `0`. See https://github.com/WordPress/performance/pull/878#discussion_r1394643420.

modules/images/image-loading-optimization/storage/rest-api.php
felixarntz: Left a follow up comment in https://github.com/WordPress/performance/pull/878#discussion_r1394635453.

modules/images/image-loading-optimization/storage/rest-api.php
westonruter: > `POST /page-metrics` would be intended to create a new resource of this type, which isn't really the case as this endpoint doesn't create a new post of the metrics post type, but modifies an existing one with specific data for a request.

Actually, it can create a new resource of that type. If there were no page metrics gathered for the URL, then a new `ilo_page_metrics` post would be created. However, if the post does already exist, then essentially it creates a new _sub_-resource (by adding the page metric to the collection of page metrics for a given URL).

I don't feel strongly about the endpoint name, however. If you want to add a `:` then I'm fine with it. But IMO the `POST` method's broad semantics don't require it. In particular, the [HTTP spec](https://httpwg.org/specs/rfc7231.html#rfc.section.4.3.3) includes this use case for `POST`:

> Appending data to a resource's existing representation(s).

modules/images/image-loading-optimization/storage/post-type.php
westonruter: Renamed in 9876aa44911a17fe5cae2843a34f5eb5c15368ef

modules/images/image-loading-optimization/storage/rest-api.php
felixarntz: Right, it _can_ create a new resource, but it doesn't always, and that's not the purpose of the endpoint.

Regarding "Appending data to a resource's existing representation(s).", that applies to a specific resource. In other words, `POST /metrics/<id>` can append data to that resource. However `POST /metrics` does not touch any existing resource, it's typically intended to create a new resource. This is applied in most core endpoints in that way and also the best practice per https://google.aip.dev/133.

I don't feel strongly about the `:`, however I do think a custom method is the appropriate solution for what this endpoint does. It's not a distinct CRUD operation for a resource (as it can either create or update), and the parameters aren't fields on the resource.

If we consider the requests within a post of the post type a sub resource, we could go with that approach, but that would require knowing the post's identifier, which is probably too much of a hassle. So with the way things are now I think `POST /{page|url}-metrics{:|/}store-request` works best.

modules/images/image-loading-optimization/storage/post-type.php
westonruter: Done in 9876aa44911a17fe5cae2843a34f5eb5c15368ef

modules/images/image-loading-optimization/storage/rest-api.php
westonruter: I've implemented this in 698d89f. I think `store` is better than `store-request`, as it is immediately following `url-metrics`.

felixarntz: I was thinking `store-request` because the data that is processed by that endpoint is from a single specific request.  And it's storing that within a URL metrics resource, rather than the URL metrics resource itself.

With only `store`, it sounds like you're storing a resource of the URL metrics post type, which if that was the case should not be a custom method. So I think `store-request` makes that more clear.

But not a blocker if you feel strongly about it.

felixarntz (approved):
LGTM!

westonruter:
Thanks all for the great reviews!

modules/images/image-loading-optimization/storage/data.php
westonruter: I discovered that this number is indeed too small, at least on its own. The columns block, for example, changes its layout at the 786px breakpoint. Gutenberg includes a [few breakpoints defined](https://github.com/WordPress/gutenberg/blob/093d52cbfd3e2c140843d3fb91ad3d03330320a5/packages/base-styles/_breakpoints.scss#L11-L13), including for mobile (480px), small (600px), and medium (786px). These seem to be used the most for media queries on the frontend, so I'm now using them in https://github.com/WordPress/performance/pull/884

PR #881, opened 2023-11-13
Fix test failures due to WP 6.4 change

felixarntz:
## Summary

As of the WordPress 6.4 release, our unit tests are failing. This hasn't happened _yet_ in `trunk`, but only because no PRs have been opened since the release. It is however obvious in https://github.com/WordPress/performance/actions/runs/6851370708/job/18627496026?pr=864.

The failures happen simply because of deprecation warnings for `print_emoji_styles()` which per core behavior remains added as a callback for the `wp_print_styles` action but normally gets removed via another core function before it would be executed. In certain custom situations though (such as tests), it may be necessary to manually remove it. Even core itself does that for example in https://github.com/WordPress/wordpress-develop/blob/d66ef46847e2340c613d0a01c36d4fcc4b5f3a6b/src/wp-includes/block-editor.php#L363.

## Checklist

### DIFF
--- a/tests/modules/js-and-css/audit-enqueued-assets/audit-enqueued-assets-test.php
+++ b/tests/modules/js-and-css/audit-enqueued-assets/audit-enqueued-assets-test.php
@@ -93,8 +93,12 @@ class Audit_Enqueued_Assets_Tests extends WP_UnitTestCase {
 
 		Audit_Assets_Transients_Set::set_style_transient_with_data( 3 );
 
-		// Avoids echoing styles.
+		// Avoid deprecation warning due to related change in WordPress 6.4.
+		if ( has_action( 'wp_print_styles', 'print_emoji_styles' ) ) {
+			remove_action( 'wp_print_styles', 'print_emoji_styles' );
+		}
 		get_echo( 'wp_print_styles' );
+
 		perflab_aea_audit_enqueued_styles();
 		$transient = get_transient( 'aea_enqueued_front_page_styles' );
 		$this->assertIsArray( $transient );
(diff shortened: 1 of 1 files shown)


swissspidy:
I feel like we should reconsider the approach with how this function is handled in core. I don't see why the function still needs to be hooked into `wp_print_styles` just so everyone has to unhook it again. That seems like worse DX than just unhooking it by default.

See also https://core.trac.wordpress.org/ticket/59892

tests/modules/js-and-css/audit-enqueued-assets/audit-enqueued-assets-test.php
```diff
@@ -93,8 +93,12 @@ public function test_perflab_aea_audit_enqueued_styles_transient_already_set() {
 
 		Audit_Assets_Transients_Set::set_style_transient_with_data( 3 );
 
-		// Avoids echoing styles.
+		// Avoid deprecation warning due to related change in WordPress 6.4.
+		if ( has_action( 'wp_print_styles', 'print_emoji_styles' ) ) {
+			remove_action( 'wp_print_styles', 'print_emoji_styles' );
+		}
```
swissspidy: Don't really need the conditional check. Can just call `remove_action()` directly.

felixarntz:
@swissspidy 
> I feel like we should reconsider the approach with how this function is handled in core. I don't see why the function still needs to be hooked into `wp_print_styles` just so everyone has to unhook it again. That seems like worse DX than just unhooking it by default.

This was done for backward compatibility with existing code. It's a clunky DX, but it's a tradeoff between breaking BC and clunky DX. I don't feel strongly either way. cc @westonruter 

I've addressed your code feedback in https://github.com/WordPress/performance/pull/881/commits/a3bc55d63112f6cb78e040c04b813ff9214d9c1f.

westonruter (approved):

westonruter:
What is [this](https://github.com/WordPress/performance/actions/runs/6857900642/job/18647791806?pr=878#step:7:153) about:

```
Warning: file_put_contents(/var/www/html/wp-content/plugins/performance/.phpunit.result.cache): failed to open stream: Permission denied in phar:///usr/local/bin/phpunit/phpunit/Runner/DefaultTestResultCache.php on line 111
```

swissspidy (approved):

swissspidy:
> This was done for backward compatibility with existing code.

Yes I saw that in the code comments, but I wonder if it _actually_ addresses a real backward compatibility issue.

PR #882, opened 2023-11-14
Refresh feature branch with `trunk` (due to unit test failures fixed)

felixarntz:
## Summary

Mostly relevant because of the fix in #881 that was merged into `trunk`.

## Checklist

### DIFF
--- a/admin/load.php
+++ b/admin/load.php
@@ -41,9 +41,9 @@ add_action( 'admin_menu', 'perflab_add_modules_page' );
 /**
  * Initializes settings sections and fields for the modules page.
  *
- * @global array $wp_settings_sections Registered WordPress settings sections.
- *
  * @since 1.0.0
+ * 
+ * @global array $wp_settings_sections Registered WordPress settings sections.
  *
  * @param array|null $modules     Associative array of available module data, keyed by module slug. By default, this
  *                                will rely on {@see perflab_get_modules()}.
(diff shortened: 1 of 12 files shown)


swissspidy (approved):
Why not just push directly? Creating a PR for that is rather unusual.

westonruter:
> Why not just push directly? Creating a PR for that is rather unusual.

I was just going to raise this point. Currently branch protections are blocking any direct pushes to feature branches. This adds too much friction, IMO. I think feature branch pushes should be allowed.

felixarntz:
@swissspidy It's not possible because of branch protections, see Weston's comment.

@westonruter I agree it's not ideal, let's discuss how we can improve it without giving up branch protection completely. In the meantime, we can probably still merge this?

swissspidy:
> I agree it's not ideal, let's discuss how we can improve it without giving up branch protection completely

Well you can specify which branches should be protected with glob patterns. Only `trunk` and `release/*` really need to be protected.

felixarntz:
@swissspidy It's not that simple. Sure, we could do that, but branch protection is also to prevent accidental deletion, e.g. relevant for the configuration for automatically deleting PR branches.

If this branch protection of feature branches leads to more headaches than it helps, we can consider getting rid of it, but that does not only come with benefits - otherwise we would have never enabled it in the first place.

westonruter (approved):

westonruter:
Branch protection can be enabled to prevent deletion without also blocking pushes.

Issue #117, opened 2022-01-20
Identifying and Optimising the LCP Image


dainemawer:
This task is something that can be tricky to determine, but if we get it right, I think it could have huge positive effects on WordPress performance. Improving the loading performance of the Largest Contentful Paint image is one of the hallmarks of keeping LCP scores in the green. However, doing so has proven relatively tricky. 

In order to reduce LCP, we need to consider a few things:
1. The size (dimensions of the image)
2. The resource size (mb) - is the image compressed etc
3. How its served and where its served from (server vs CDN)
4. How soon the browser renders the image

There are a number of things out of our control here, especially from a server, caching and infrastructure perspective. But we can look at preloading that image if we can identify that it is in fact the LCP on the page. This spoken about here on `web.dev` in detail: https://web.dev/optimize-lcp/#preload-important-resources especially for responsive images.

sgomes:
I'm afraid this goes beyond tricky. It's simply not possible to determine with certainty what the LCP will be (and whether it will be an image). Beyond the factors you mentioned, device screen sizes are also a huge factor, as is anything else that could be conditional to a device type, user, location, etc.

And this is assuming the image even appears in markup; a background image can easily become the largest element, and would be completely unknown to WordPress if defined in a theme CSS file, for example.

Since there is no generic way we can accurately determine what the LCP will be for a particular page view, my suggestion is to not try to do this as a platform, and instead leave it to the user to manually optimise if necessary.

westonruter:
The identification could be done client-side. I believe this is the approach taken with Jetpack Boost for generating the critical CSS. The frontend could be loaded up into an iframe with some `PerformanceObserver` code added. Once the iframe has finished loading, the `PerformanceObserver` could send the LCP element information to the parent page. Assuming that the LCP element (e.g. image) remains consistent across page loads, the URL for the LCP image (or font) could be stored so that a `preload` link could be automatically added. This process could happen in the editor after saving. 

This would be similar to CLS mitigation that I've suggested in https://github.com/GoogleChromeLabs/layout-shift-terminator/issues/5 and overall gathering if metrics in https://github.com/WordPress/gutenberg/issues/33578#issuecomment-919313851.

A caveat about images: responsive images are only preloadable on Chromium-based browsers at the moment, since only they support `imagesizes` and `imagesrcset` ([ref](https://web.dev/preload-responsive-images/#responsive-images-+-preload-faster-image-loads)).

sgomes:
[...]

PR #884, opened 2023-11-16
Optimize the loading of images using stored URL metrics

### LINKED ISSUE #872: Apply image loading optimizations based on stored client-side detection

westonruter:
## Summary

Fixes #872 as part of #869.

This implements the optimization of image loading based on the client-side detection (https://github.com/WordPress/performance/pull/876) that has been stored (https://github.com/WordPress/performance/pull/878).

_This is fully functional and able to be tested in a WP install._

## Relevant technical choices

When a response is eligible for being optimized, an output buffer filter is added to apply optimizations. This involves looking up the `ilo_url_metrics` post containing the stored URL metrics, and when present groupong the URL metrics into the breakpoints (480px, 600px, 782px). If all breakpoints have URL metrics collected and all share the same image as LCP element, then `fetchpriority=high` is set on this image while being removed from all other images. If all breakpoints don't have URL metrics collected yet, then the `fetchpriority` attribute added by server-side heuristics is not removed. Certain data attributes are also added to elements so we can track what operations are being performed:

Attribute | Condition
--|--
`data-ilo-added-fetchpriority` | When the `fetchpriority` attribute is added to an element.
`data-ilo-removed-fetchpriority` | When the `fetchpriority` attribute is removed from an element.
`data-ilo-fetchpriority-already-added` | When WordPress's existing server-side heuristics have already set `fetchpriority=high` on the LCP image, then the `data-ilo-removed-fetchpriority` attribute is added.
`data-ilo-added-tag` | Present on the `preload` links which are injected at the end of the `head`.

When the `fetchpriority` attribute is added, a `data-ilo-added-fetchpriority` attribute is also added so we can keep track of what operations it is doing. 

Whenever there are URL metrics available, `preload` links are also added for the LCP image in each breakpoint. This is particularly important when different breakpoints have different LCP elements (even none at all), as in such cases the `fetchpriority` attribute must be removed from each breakpoint-specific LCP image element. This is also useful for when breakpoints don't have URL metrics collected yet, as otherwise `fetchpriority=high` could be erroneously added to the wrong image for the current viewport. 

In order to apply the `fetchpriority` optimizations to the page and to collect the attributes from `IMG` elements for use in preload links, the `WP_HTML_Tag_Processor` class is leveraged, although it is not used directly. An `ILO_HTML_Tag_Processor` class is used which incorporates `WP_HTML_Tag_Processor` (by composition instead of extension). This processor is similar to `WP_HTML_Tag_Processor` except it is more constrained. It provides a generator method called `open_tags()`; while iterating over this generator, the breadcrumbs for the current open tag can be obtained via the `get_breadcrumbs()` method. Otherwise, the following methods from the underlying `WP_HTML_Tag_Processor` are exposed: 

* `get_attribute()`
* `set_attribute()`
* `remove_attribute()`
* `get_updated_html()`

Importantly, the `next_tag()` and `seek()` methods are not exposed since in order to correctly compute breadcrumbs the processor must proceed forward through the entire document visiting all open and closing tags.

Tip: The following command deletes all `ilo_url_metrics` posts:

```bash
npm run wp-env run cli -- post delete --force $( npm run --silent wp-env run cli -- post list --post_type=ilo_url_metrics --format=ids 2> /dev/null )
```

### Other changes

* Besides the existing 480px breakpoint, additional breakpoints of 600px and 782px have been added. These three breakpoints correspond to the `mobile`, `small`, and `medium` breakpoints that are defined in Gutenberg's [`_breakpoints.css`](https://github.com/WordPress/gutenberg/blob/093d52cbfd3e2c140843d3fb91ad3d03330320a5/packages/base-styles/_breakpoints.scss#L11-L13).
* Detection and optimization by default do not run now in the Customizer preview or in response to a non-`GET` request (e.g. a `POST` request).
* Breadcrumb calculation has been updated to disregard the presence of the Admin Bar. Similarly, client-side detection logic has been updated to account for JS injection of the `.skip-link.screen-reader-text` element, which throws off generation of breadcrumbs on the client to match breadcrumbs generated on the server. In reality, the generation of breadcrumbs client-side is fragile and in a subsequent PR I intend to rely on server-side breadcrumb generation exclusively.
* Some additional `preflab_` prefixes have been replaced with `ilo_`.
* The `tagName` key has been replaced with `tag` in breadcrumbs.

## Checklist

### DIFF
--- a/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
+++ b/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
@@ -10,6 +10,7 @@
  * Subclass of WP_HTML_Tag_Processor that adds support for breadcrumbs and a visiting callback.
  *
  * @since n.e.x.t
+ * @access private
  */
 class ILO_HTML_Tag_Processor {
 
--- a/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
+++ b/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
@@ -218,14 +218,14 @@ class ILO_HTML_Tag_Processor {
 	 *
 	 * Breadcrumbs are constructed to match the format from detect.js.
 	 *
-	 * @return array<array{tagName: string, index: int}> Breadcrumbs.
+	 * @return array<array{tag: string, index: int}> Breadcrumbs.
 	 */
 	public function get_breadcrumbs(): array {
 		$breadcrumbs = array();
 		foreach ( $this->open_stack_tags as $i => $breadcrumb_tag_name ) {
 			$breadcrumbs[] = array(
-				'tagName' => $breadcrumb_tag_name, // TODO: Just 'tag'.
-				'index'   => $this->open_stack_indices[ $i ],
+				'tag'   => $breadcrumb_tag_name,
+				'index' => $this->open_stack_indices[ $i ],
 			);
 		}
 		return $breadcrumbs;
--- a/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
+++ b/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
@@ -1,0 +1,310 @@
+<?php
+/**
+ * Image Loading Optimization: ILO_HTML_Tag_Processor class
+ *
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+/**
+ * Subclass of WP_HTML_Tag_Processor that adds support for breadcrumbs and a visiting callback.
+ *
+ * @since n.e.x.t
+ */
+class ILO_HTML_Tag_Processor {
+
+	/**
+	 * HTML elements that are self-closing.
+	 *
+	 * @link https://www.w3.org/TR/html5/syntax.html#serializing-html-fragments
+	 * @link https://github.com/ampproject/amp-toolbox-php/blob/c79a0fe558a3c042aee4789bbf33376cca7a733d/src/Html/Tag.php#L206-L232
+	 *
+	 * @var string[]
+	 */
+	const SELF_CLOSING_TAGS = array(
+		'AREA',
+		'BASE',
+		'BASEFONT',
+		'BGSOUND',
+		'BR',
+		'COL',
+		'EMBED',
+		'FRAME',
+		'HR',
+		'IMG',
+		'INPUT',
+		'KEYGEN',
+		'LINK',
+		'META',
+		'PARAM',
+		'SOURCE',
+		'TRACK',
+		'WBR',
+	);
+
+	/**
+	 * The set of HTML tags whose presence will implicitly close a <p> element.
+	 * For example '<p>foo<h1>bar</h1>' should parse the same as '<p>foo</p><h1>bar</h1>'.
+	 *
+	 * @link https://www.w3.org/TR/html-markup/p.html
+	 * @link https://github.com/ampproject/amp-toolbox-php/blob/c79a0fe558a3c042aee4789bbf33376cca7a733d/src/Html/Tag.php#L262-L293
+	 */
+	const P_CLOSING_TAGS = array(
+		'ADDRESS',
+		'ARTICLE',
+		'ASIDE',
+		'BLOCKQUOTE',
+		'DIR',
+		'DL',
+		'FIELDSET',
+		'FOOTER',
+		'FORM',
+		'H1',
+		'H2',
+		'H3',
+		'H4',
+		'H5',
+		'H6',
+		'HEADER',
+		'HR',
+		'MENU',
+		'NAV',
+		'OL',
+		'P',
+		'PRE',
+		'SECTION',
+		'TABLE',
+		'UL',
+	);
+
+	/**
+	 * Open stack tags.
+	 *
+	 * @var string[]
+	 */
+	private $open_stack_tags = array();
+
+	/**
+	 * Open stag indices.
+	 *
+	 * @var int[]
+	 */
+	private $open_stack_indices = array();
+
+	/**
+	 * Processor.
+	 *
+	 * @var WP_HTML_Tag_Processor
+	 */
+	private $processor;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param string $html HTML to process.
+	 */
+	public function __construct( string $html ) {
+		$this->processor = new WP_HTML_Tag_Processor( $html );
+	}
+
+	/**
+	 * Walk over the document.
+	 *
+	 * Whenever an open tag is encountered, invoke the supplied $open_tag_callback and pass the tag name and breadcrumbs.
+	 *
+	 * @param Closure $open_tag_callback Open tag callback. The processor instance is passed as the sole argument.
+	 */
+	public function walk( Closure $open_tag_callback ) {
+		$p = $this->processor;
+
+		/*
+		 * The keys for the following two arrays correspond to each other. Given the following document:
+		 *
+		 * <html>
+		 *   <head>
+		 *   </head>
+		 *   <body>
+		 *     <p>Hello!</p>
+		 *     <img src="lcp.png">
+		 *   </body>
+		 * </html>
+		 *
+		 * The two upon processing the IMG element, the two arrays should be equal to the following:
+		 *
+		 * $open_stack_tags    = array( 'HTML', 'BODY', 'IMG' );
+		 * $open_stack_indices = array( 0, 1, 1 );
+		 */
+		$this->open_stack_tags    = array();
+		$this->open_stack_indices = array();
+		while ( $p->next_tag( array( 'tag_closers' => 'visit' ) ) ) {
+			$tag_name = $p->get_tag();
+			if ( ! $p->is_tag_closer() ) {
+
+				// Close an open P tag when a P-closing tag is encountered.
+				if ( in_array( $tag_name, self::P_CLOSING_TAGS, true ) ) {
+					$i = array_search( 'P', $this->open_stack_tags, true );
+					if ( false !== $i ) {
+						array_splice( $this->open_stack_tags, $i );
+						array_splice( $this->open_stack_indices, count( $this->open_stack_tags ) );
+					}
+				}
+
+				$level                   = count( $this->open_stack_tags );
+				$this->open_stack_tags[] = $tag_name;
+
+				if ( ! isset( $this->open_stack_indices[ $level ] ) ) {
+					$this->open_stack_indices[ $level ] = 0;
+				} else {
+					++$this->open_stack_indices[ $level ];
+				}
+
+				// TODO: We should consider not collecting metrics when the admin bar is shown and the user is logged-in.
+				// Only increment the tag index at this level only if it isn't the admin bar, since the presence of the
+				// admin bar can throw off the indices.
+				if ( 'DIV' === $tag_name && $p->get_attribute( 'id' ) === 'wpadminbar' ) {
+					--$this->open_stack_indices[ $level ];
+				}
+
+				// Invoke the callback to do processing.
+				$open_tag_callback( $tag_name, $this->get_breadcrumbs() );
+
+				// Immediately pop off self-closing tags.
+				if ( in_array( $tag_name, self::SELF_CLOSING_TAGS, true ) ) {
+					array_pop( $this->open_stack_tags );
+				}
+			} else {
+				// If the closing tag is for self-closing tag, we ignore it since it was already handled above.
+				if ( in_array( $tag_name, self::SELF_CLOSING_TAGS, true ) ) {
+					continue;
+				}
+
+				// Since SVG and MathML can have a lot more self-closing/empty tags, potentially pop off the stack until getting to the open tag.
+				$did_splice = false;
+				if ( 'SVG' === $tag_name || 'MATH' === $tag_name ) {
+					$i = array_search( $tag_name, $this->open_stack_tags, true );
+					if ( false !== $i ) {
+						array_splice( $this->open_stack_tags, $i );
+						$did_splice = true;
+					}
+				}
+
+				if ( ! $did_splice ) {
+					$popped_tag_name = array_pop( $this->open_stack_tags );
+					if ( $popped_tag_name !== $tag_name ) {
+						error_log( "Expected popped tag stack element $popped_tag_name to match the currently visited closing tag $tag_name." ); // phpcs:ignore
+					}
+				}
+				array_splice( $this->open_stack_indices, count( $this->open_stack_tags ) + 1 );
+			}
+		}
+	}
+
+	/**
+	 * Returns the uppercase name of the matched tag.
+	 *
+	 * This is a wrapper around the underlying HTML_Tag_Processor method of the same name since only a limited number of
+	 * methods can be exposed to prevent moving the pointer in such a way as the breadcrumb calculation is invalidated.
+	 *
+	 * @see WP_HTML_Tag_Processor::get_tag()
+	 *
+	 * @return string|null Name of currently matched tag in input HTML, or `null` if none found.
+	 */
+	public function get_tag() {
+		return $this->processor->get_tag();
+	}
+
+	/**
+	 * Gets breadcrumbs for the current open tag.
+	 *
+	 * Breadcrumbs are constructed to match the format from detect.js.
+	 *
+	 * @return array<array{tagName: string, index: int}> Breadcrumbs.
+	 */
+	public function get_breadcrumbs(): array {
+		$breadcrumbs = array();
+		foreach ( $this->open_stack_tags as $i => $breadcrumb_tag_name ) {
+			$breadcrumbs[] = array(
+				'tagName' => $breadcrumb_tag_name, // TODO: Just 'tag'.
+				'index'   => $this->open_stack_indices[ $i ],
+			);
+		}
+		return $breadcrumbs;
+	}
+
+	/**
+	 * Removes the fetchpriority attribute from the current node being walked over.
+	 *
+	 * Also sets an attribute to indicate that the attribute was removed.
+	 *
+	 * @return bool Whether an attribute was removed.
+	 */
+	public function remove_fetchpriority_attribute(): bool {
+		$p = $this->processor;
+		if ( $p->get_attribute( 'fetchpriority' ) ) {
+			$p->set_attribute( 'data-ilo-removed-fetchpriority', $p->get_attribute( 'fetchpriority' ) );
+			return $p->remove_attribute( 'fetchpriority' );
+		} else {
+			return false;
+		}
+	}
+
+	/**
+	 * Returns the value of a requested attribute from a matched tag opener if that attribute exists.
+	 *
+	 * This is a wrapper around the underlying HTML_Tag_Processor method of the same name since only a limited number of
+	 * methods can be exposed to prevent moving the pointer in such a way as the breadcrumb calculation is invalidated.
+	 *
+	 * @see WP_HTML_Tag_Processor::get_attribute()
+	 *
+	 * @param string $name Name of attribute whose value is requested.
+	 * @return string|true|null Value of attribute or `null` if not available. Boolean attributes return `true`.
+	 */
+	public function get_attribute( string $name ) {
+		return $this->processor->get_attribute( $name );
+	}
+
+	/**
+	 * Updates or creates a new attribute on the currently matched tag with the passed value.
+	 *
+	 * This is a wrapper around the underlying HTML_Tag_Processor method of the same name since only a limited number of
+	 * methods can be exposed to prevent moving the pointer in such a way as the breadcrumb calculation is invalidated.
+	 *
+	 * @see WP_HTML_Tag_Processor::set_attribute()
+	 *
+	 * @param string      $name  The attribute name to target.
+	 * @param string|bool $value The new attribute value.
+	 * @return bool Whether an attribute value was set.
+	 */
+	public function set_attribute( string $name, $value ): bool {
+		return $this->processor->set_attribute( $name, $value );
+	}
+
+	/**
+	 * Removes an attribute from the currently-matched tag.
+	 *
+	 * This is a wrapper around the underlying HTML_Tag_Processor method of the same name since only a limited number of
+	 * methods can be exposed to prevent moving the pointer in such a way as the breadcrumb calculation is invalidated.
+	 *
+	 * @see WP_HTML_Tag_Processor::remove_attribute()
+	 *
+	 * @param string $name The attribute name to remove.
+	 * @return bool Whether an attribute was removed.
+	 */
+	public function remove_attribute( string $name ): bool {
+		return $this->processor->remove_attribute( $name );
+	}
+
+	/**
+	 * Returns the string representation of the HTML Tag Processor.
+	 *
+	 * This is a wrapper around the underlying HTML_Tag_Processor method of the same name since only a limited number of
+	 * methods can be exposed to prevent moving the pointer in such a way as the breadcrumb calculation is invalidated.
+	 *
+	 * @see WP_HTML_Tag_Processor::get_updated_html()
+	 *
+	 * @return string The processed HTML.
+	 */
+	public function get_updated_html(): string {
+		return $this->processor->get_updated_html();
+	}
+}
--- a/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
+++ b/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
@@ -112,9 +112,9 @@ class ILO_HTML_Tag_Processor {
 	 *
 	 * Whenever an open tag is encountered, invoke the supplied $open_tag_callback and pass the tag name and breadcrumbs.
 	 *
-	 * @param Closure $open_tag_callback Open tag callback. The processor instance is passed as the sole argument.
+	 * @param callable $open_tag_callback Open tag callback. The processor instance is passed as the sole argument.
 	 */
-	public function walk( Closure $open_tag_callback ) {
+	public function walk( callable $open_tag_callback ) {
 		$p = $this->processor;
 
 		/*
--- a/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
+++ b/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
@@ -7,7 +7,9 @@
  */
 
 /**
- * Processor leveraging WP_HTML_Tag_Processor which gathers breadcrumbs which can be queried while iterating the open_tags() generator .
+ * Processor leveraging WP_HTML_Tag_Processor which gathers breadcrumbs which can be queried while iterating the open_tags() generator.
+ *
+ * Eventually this class should be made largely obsolete once `WP_HTML_Processor` is fully implemented to support all HTML tags.
  *
  * @since n.e.x.t
  * @access private
--- a/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
+++ b/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
@@ -15,29 +15,30 @@
 final class ILO_HTML_Tag_Processor {
 
 	/**
-	 * HTML elements that are self-closing.
+	 * HTML void tags (i.e. those which are self-closing).
 	 *
-	 * @link https://www.w3.org/TR/html5/syntax.html#serializing-html-fragments
-	 * @link https://github.com/ampproject/amp-toolbox-php/blob/c79a0fe558a3c042aee4789bbf33376cca7a733d/src/Html/Tag.php#L206-L232
+	 * @link https://html.spec.whatwg.org/multipage/syntax.html#void-elements
+	 * @see WP_HTML_Processor::is_void()
+	 * @todo Reuse `WP_HTML_Processor::is_void()` once WordPress 6.4 is the minimum-supported version.
 	 *
 	 * @var string[]
 	 */
-	const SELF_CLOSING_TAGS = array(
+	const VOID_TAGS = array(
 		'AREA',
 		'BASE',
-		'BASEFONT',
-		'BGSOUND',
+		'BASEFONT', // Obsolete.
+		'BGSOUND', // Obsolete.
 		'BR',
 		'COL',
 		'EMBED',
-		'FRAME',
+		'FRAME', // Deprecated.
 		'HR',
 		'IMG',
 		'INPUT',
-		'KEYGEN',
+		'KEYGEN', // Obsolete.
 		'LINK',
 		'META',
-		'PARAM',
+		'PARAM', // Deprecated.
 		'SOURCE',
 		'TRACK',
 		'WBR',
@@ -47,8 +48,7 @@ final class ILO_HTML_Tag_Processor {
 	 * The set of HTML tags whose presence will implicitly close a <p> element.
 	 * For example '<p>foo<h1>bar</h1>' should parse the same as '<p>foo</p><h1>bar</h1>'.
 	 *
-	 * @link https://www.w3.org/TR/html-markup/p.html
-	 * @link https://github.com/ampproject/amp-toolbox-php/blob/c79a0fe558a3c042aee4789bbf33376cca7a733d/src/Html/Tag.php#L262-L293
+	 * @link https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element
 	 *
 	 * @var string[]
 	 */
@@ -57,9 +57,12 @@ final class ILO_HTML_Tag_Processor {
 		'ARTICLE',
 		'ASIDE',
 		'BLOCKQUOTE',
-		'DIR',
+		'DETAILS',
+		'DIV',
 		'DL',
 		'FIELDSET',
+		'FIGCAPTION',
+		'FIGURE',
 		'FOOTER',
 		'FORM',
 		'H1',
@@ -69,12 +72,15 @@ final class ILO_HTML_Tag_Processor {
 		'H5',
 		'H6',
 		'HEADER',
+		'HGROUP',
 		'HR',
+		'MAIN',
 		'MENU',
 		'NAV',
 		'OL',
 		'P',
 		'PRE',
+		'SEARCH',
 		'SECTION',
 		'TABLE',
 		'UL',
@@ -176,12 +182,12 @@ final class ILO_HTML_Tag_Processor {
 				yield $tag_name;
 
 				// Immediately pop off self-closing tags.
-				if ( in_array( $tag_name, self::SELF_CLOSING_TAGS, true ) ) {
+				if ( in_array( $tag_name, self::VOID_TAGS, true ) ) {
 					array_pop( $this->open_stack_tags );
 				}
 			} else {
 				// If the closing tag is for self-closing tag, we ignore it since it was already handled above.
-				if ( in_array( $tag_name, self::SELF_CLOSING_TAGS, true ) ) {
+				if ( in_array( $tag_name, self::VOID_TAGS, true ) ) {
 					continue;
 				}
 
--- a/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
+++ b/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
@@ -235,23 +235,6 @@ class ILO_HTML_Tag_Processor {
 	}
 
 	/**
-	 * Removes the fetchpriority attribute from the current node being walked over.
-	 *
-	 * Also sets an attribute to indicate that the attribute was removed.
-	 *
-	 * @return bool Whether an attribute was removed.
-	 */
-	public function remove_fetchpriority_attribute(): bool {
-		$p = $this->processor;
-		if ( $p->get_attribute( 'fetchpriority' ) ) {
-			$p->set_attribute( 'data-ilo-removed-fetchpriority', $p->get_attribute( 'fetchpriority' ) );
-			return $p->remove_attribute( 'fetchpriority' );
-		} else {
-			return false;
-		}
-	}
-
-	/**
 	 * Returns the value of a requested attribute from a matched tag opener if that attribute exists.
 	 *
 	 * This is a wrapper around the underlying HTML_Tag_Processor method of the same name since only a limited number of
--- a/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
+++ b/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
@@ -7,7 +7,7 @@
  */
 
 /**
- * Processor leveraging WP_HTML_Tag_Processor which walks over a document and gathers breadcrumbs and invokes a callback for each open tag.
+ * Processor leveraging WP_HTML_Tag_Processor which gathers breadcrumbs which can be queried while iterating the open_tags() generator .
  *
  * @since n.e.x.t
  * @access private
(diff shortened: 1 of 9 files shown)


westonruter:
This is now working!

* When there is only one LCP image shared for all breakpoints, it adds `fetchpriority=high` to that element and removes it from all others.
* When breakpoints have different LCP images (or some have none), `fetchpriority` is removed from all images and preload links are added to target the breakpoints.

Elements that have been muted get annotated with `data-ilo-*` attributes to easily track what it is doing.

Items to further work on:

- [x] Create a subclass of `WP_HTML_Tag_Processor` instead of using the `ilo_walk_document()` function.
- [x] Disable detection when in the Customizer preview due to the injection of inline editing controls.
- [x] Disable detection and optimization when `$_SERVER['REQUEST_METHOD'] !== 'GET'`.

For a future PR:

- [ ] Improve handling of breadcrumb generation when there the document structure varies either due to admin bar being present or due to client-side logic. Currently breadcrumbs have to explicitly skip over the admin bar element to ignore it. Similarly, the client-side breadcrumb logic has to skip over `.skip-link.screen-reader-text` which is injected by `wp_enqueue_block_template_skip_link()`. This issue of client-side mutated elements is particularly problematic, which may require server-side tagging of elements with their breadcrumbs which would eliminate needing to also do that calculation on the client. But this would mean extraneous `data-` attributes on all images. When detection isn't needed, the tag processor could strip those attributes out before sending back the buffer. We may need to monitor this to see how much of a problem it is, such as by checking various themes to see what degree they inject elements to mess up breadcrumbs. As another alternative, we may consider relying on IDs and class names for breadcrumbs rather than element indices, although JS here too can cause problems by modifying the class names.
- [x] Go ahead and inject a preload link even when there is only one LCP image shared by all breakpoints?
- [ ] What about LCP images that are injected with JavaScript, e.g. a dynamically-constructed carousel? The element will not be in the DOM for adding `fetchpriority=high` to. However detection could capture the image attributes (e.g. `src` and `srcset`) to then add a preload link. The challenge here is that there's no way to guard against preloading an image that JS no longer is showing. This isn't a problem for server-side processing because we don't store the `src` but rather just the breadcrumbs, so we validate the LCP element is still in the page.
- [ ] Should detection be disabled when the user is logged-in (and the admin bar is shown)?

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -277,6 +288,86 @@ static function ( $breakpoint ) {
...
+			$most_common_breadcrumb_index = key( $breadcrumb_counts );
+
+			$lcp_element_by_viewport_minimum_width[ $viewport_minimum_width ] = $breadcrumb_element[ $most_common_breadcrumb_index ];
+		} elseif ( ! empty( $breakpoint_url_metrics ) ) {
+			$lcp_element_by_viewport_minimum_width[ $viewport_minimum_width ] = false; // No LCP image at this breakpoint.
+		}
+	}
+
+	// Now we need to merge the breakpoints when there is an LCP element common between them.
+	$last_lcp_element = null;
```
westonruter: Nit: `s/last_lcp_element/prev_lcp_element/`

modules/images/image-loading-optimization/optimization.php
```diff
@@ -0,0 +1,322 @@
...
+								}
+							}
+						}
+					}
+				}
+			);
+
+			$preload_links = ilo_construct_preload_links( $lcp_images_by_minimum_viewport_widths );
+
+			$buffer = str_replace( '</head>', $preload_links . '</head>', $buffer );
```
westonruter:
```suggestion
			$buffer = preg_replace( '#(?=</head>)#', $preload_links, $buffer, 1 );
```

modules/images/image-loading-optimization/optimization.php
```diff
@@ -0,0 +1,322 @@
...
+		if ( ! $p->is_tag_closer() ) {
+
+			// Close an open P tag when a P-closing tag is encountered.
+			if ( in_array( $tag_name, ILO_P_CLOSING_TAGS, true ) ) {
+				$i = array_search( 'P', $open_stack_tags, true );
+				if ( false !== $i ) {
+					array_splice( $open_stack_tags, $i );
+					array_splice( $open_stack_indices, count( $open_stack_tags ) );
+				}
+			}
```
westonruter: There are quite a few more cases of optional closing tags: https://html.spec.whatwg.org/multipage/syntax.html#optional-tags

Nevertheless, given WordPress's legacy of XHTML compatibility, the lack of closing tags may not be common enough to warrant worrying about any of them.

modules/images/image-loading-optimization/detection/detect.js
```diff
@@ -111,7 +113,21 @@ function getElementIndex( element ) {
...
-	return [ ...element.parentElement.children ].indexOf( element );
+	const children = [ ...element.parentElement.children ];
+	let index = children.indexOf( element );
+	if ( children.includes( document.getElementById( adminBarId ) ) ) {
+		// TODO: Should detection just be turned off when is_user_logged_in()?
+		--index;
+	}
+	if (
+		children.includes(
+			document.querySelector( '.skip-link.screen-reader-text' )
```
westonruter, line 125: This should be cached.

westonruter: I'm going to leave this as-is for this PR. In the next PR I intend to eliminate client-side breadcrumb construction since it is too easy for JavaScript DOM mutations to cause client-side generated breadcrumbs from being able to apply on the server.

modules/images/image-loading-optimization/detection/detect.js
```diff
@@ -238,23 +257,27 @@ export default async function detect( {
...
 
 	/** @type {Map<Element, Breadcrumb[]>} */
 	const breadcrumbedElementsMap = new Map(
-		[ ...breadcrumbedImages, ...breadcrumbedElementsWithBackgrounds ].map(
+		[
+			...breadcrumbedImages /*, ...breadcrumbedElementsWithBackgrounds*/,
+		].map(
+			// TODO: Instead of generating breadcrumbs here, rely instead on server-generated breadcrumbs that are added to a data attribute by the server.
 			( element ) => [ element, getBreadcrumbs( element ) ]
 		)
```
westonruter, line 281: These changes will be undone in subsequent PRs which will (1) move to do all breadcrumbing on the server, and (2) actually implement support for `background-image`.

modules/images/image-loading-optimization/optimization.php
```diff
@@ -0,0 +1,172 @@
...
+add_action( 'wp', 'ilo_maybe_add_template_output_buffer_filter' );
+
+/**
+ * Constructs preload links.
+ *
+ * @since n.e.x.t
+ * @access private
+ *
+ * @param array $lcp_images_by_minimum_viewport_widths LCP images keyed by minimum viewport width, amended with attributes key for the IMG attributes.
+ * @return string Markup for one or more preload link tags.
```
westonruter:
```suggestion
 * @return string Markup for zero or more preload link tags.
```

modules/images/image-loading-optimization/optimization.php
```diff
@@ -0,0 +1,172 @@
...
+		return $buffer;
+	}
+
+	$url_metrics = ilo_parse_stored_url_metrics( $post );
+
+	$breakpoint_max_widths                   = ilo_get_breakpoint_max_widths();
+	$url_metrics_grouped_by_breakpoint       = ilo_group_url_metrics_by_breakpoint( $url_metrics, $breakpoint_max_widths );
+	$lcp_elements_by_minimum_viewport_widths = ilo_get_lcp_elements_by_minimum_viewport_widths( $url_metrics_grouped_by_breakpoint );
+
+	// TODO: Handle case when the LCP element is not an image at all, but rather a background-image.
```
westonruter, line 138: To be done in a subsequent PR.

modules/images/image-loading-optimization/detection/detect.js
```diff
@@ -104,19 +106,36 @@ function error( ...message ) {
 /**
  * Gets element index among siblings.
  *
+ * @todo Eliminate this in favor of doing all breadcrumb generation exclusively on the server.
```
westonruter, line 109: To be done in subsequent PR.

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	/**
+	 * Walk over the document.
+	 *
+	 * Whenever an open tag is encountered, invoke the supplied $open_tag_callback and pass the tag name and breadcrumbs.
+	 *
+	 * @param callable $open_tag_callback Open tag callback. The processor instance is passed as the sole argument.
+	 */
+	public function walk( callable $open_tag_callback ) {
```
westonruter: Actually, I think I'll reimplement this as a generator and use the `next_tag` method name for additional parity with `WP_HTML_Tag_Processor`

westonruter:
@dmsnell FYI: The `ILO_HTML_Tag_Processor` may be of interest to you here. I didn't use `WP_HTML_Processor` because it doesn't support all HTML tags and because I thought I could make it more efficient for the use case here. Also, for breadcrumbs to reliably target a specific element they need to not only include the tag name but also the sibling index of that tag (i.e. `HTML, BODY, MAIN, FIGURE, IMG` could match many elements whereas `(HTML, 0), (BODY, 1), (MAIN, 0), (FIGURE, 10), (IMG, 0))` would exclusively target the image in the 11th image block of the page. Maybe I should have used `WP_HTML_Processor` to some extent, however.

modules/images/image-loading-optimization/detection/detect.js
```diff
@@ -104,19 +106,36 @@ function error( ...message ) {
...
  */
 function getElementIndex( element ) {
 	if ( ! element.parentElement ) {
 		return 0;
 	}
-	return [ ...element.parentElement.children ].indexOf( element );
+	const children = [ ...element.parentElement.children ];
+	let index = children.indexOf( element );
+	if ( children.includes( document.getElementById( adminBarId ) ) ) {
+		// TODO: Should detection just be turned off when is_user_logged_in()?
```
adamsilverstein: > Should detection just be turned off when is_user_logged_in()?

probably, the whole point of this exercise is to gather real user metrics.

modules/images/image-loading-optimization/detection/detect.js
adamsilverstein: Once you switch to generate breadcrumbs on the server,  maybe a shorter index or hash passed to the front end?

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -0,0 +1,296 @@
...
+		 * <html>
+		 *   <head>
+		 *   </head>
+		 *   <body>
+		 *     <p>Hello!</p>
+		 *     <img src="lcp.png">
+		 *   </body>
+		 * </html>
+		 *
+		 * The two upon processing the IMG element, the two arrays should be equal to the following:
```
adamsilverstein:
```suggestion
		 * Upon processing the IMG element, the two arrays should be equal to the following:
```

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -0,0 +1,296 @@
...
+				$level                   = count( $this->open_stack_tags );
+				$this->open_stack_tags[] = $tag_name;
+
+				if ( ! isset( $this->open_stack_indices[ $level ] ) ) {
+					$this->open_stack_indices[ $level ] = 0;
+				} else {
+					++$this->open_stack_indices[ $level ];
+				}
+
+				// TODO: We should consider not collecting metrics when the admin bar is shown and the user is logged-in.
```
adamsilverstein: +1

modules/images/image-loading-optimization/detection/detect.js
joemcgill: What about sites that require users to be logged in to see certain content or comment?

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -0,0 +1,296 @@
...
+
+	/**
+	 * HTML elements that are self-closing.
+	 *
+	 * @link https://www.w3.org/TR/html5/syntax.html#serializing-html-fragments
+	 * @link https://github.com/ampproject/amp-toolbox-php/blob/c79a0fe558a3c042aee4789bbf33376cca7a733d/src/Html/Tag.php#L206-L232
+	 *
+	 * @var string[]
+	 */
+	const SELF_CLOSING_TAGS = array(
```
dmsnell: it's just a term, but there's a distinction in HTML and these are _void_ tags, not _self-closing_. only foreign elements (SVG, custom-elements, MathML) can contain self-closing flag.

it looks like I left out the deprecated elements in the HTML Processor, but we can add them in and then you can use the `HTML_Processor::is_void()` static method without instantiating the class.

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -0,0 +1,296 @@
...
+		'HEADER',
+		'HR',
+		'MENU',
+		'NAV',
+		'OL',
+		'P',
+		'PRE',
+		'SECTION',
+		'TABLE',
+		'UL',
```
dmsnell, line 88: CENTER, DETAILS, DIALOG, DIV, FIGCAPTION, FIGURE, HEADER, HGROUP, MAIN, SEARCH, SUMMARY are additional tags supported by the HTML Processor which also close an open P

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -0,0 +1,296 @@
...
+		$this->open_stack_indices = array();
+		while ( $p->next_tag( array( 'tag_closers' => 'visit' ) ) ) {
+			$tag_name = $p->get_tag();
+			if ( ! $p->is_tag_closer() ) {
+
+				// Close an open P tag when a P-closing tag is encountered.
+				// TODO: There are quite a few more cases of optional closing tags: https://html.spec.whatwg.org/multipage/syntax.html#optional-tags
+				// Nevertheless, given WordPress's legacy of XHTML compatibility, the lack of closing tags may not be common enough to warrant worrying about any of them.
+				if ( in_array( $tag_name, self::P_CLOSING_TAGS, true ) ) {
+					$i = array_search( 'P', $this->open_stack_tags, true );
```
dmsnell, line 159: this is probably fine enough, but be aware that there are certain markers/boundaries in the stack that effectively isolate regions of HTML. for example, if we're currently inside a TEMPLATE element and encounter one of these P-closing elements, they don't close the P from the outside of the template.

```html
<p>
	This template is isolated.
	<template><section><p>Hidden</p></section></template>
	The outer P remains.
</p>
```

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -0,0 +1,296 @@
...
+					continue;
+				}
+
+				// Since SVG and MathML can have a lot more self-closing/empty tags, potentially pop off the stack until getting to the open tag.
+				$did_splice = false;
+				if ( 'SVG' === $tag_name || 'MATH' === $tag_name ) {
+					$i = array_search( $tag_name, $this->open_stack_tags, true );
+					if ( false !== $i ) {
+						array_splice( $this->open_stack_tags, $i );
+						$did_splice = true;
```
dmsnell, line 201: if encountering an element which _isn't_ an HTML _Void_ element, then regardless of the tag, if the self-closing flag is present, then it closes the tag. all self-closing flags on HTML elements are invalid and ignored, but all on HTML _foreign elements_ are authoritative and obeyed.

```html
<p>There are no empty <div/> tags. This is inside the DIV</p>.
<wp-group>Is a custom element, and so <wp-group /> is self-closing while others aren't.</wp-group>
```

```php
if (
	WP_HTML_Processor::is_void( $tag_name ) || 
	(
		! WP_HTML_Processor::is_html_element( $tag_name ) &&
		$p->has_self_closing_flag
	)
) {
	// this tag immediately closes as soon as we jump to the next tag.
}
```

this requires having a list of all HTML elements, which the HTML Processor currently doesn't do because it doesn't support any, but we'll have to add it. this _can_ lead to common parsing failures because the invalid self-closing flag has become more popular post-React where it's valid and normative in JSX.

modules/images/image-loading-optimization/optimization.php
```diff
@@ -0,0 +1,174 @@
...
+			// Map img attribute name to link attribute name.
+			if ( 'srcset' === $name || 'sizes' === $name ) {
+				$name = 'image' . $name;
+			} elseif ( 'src' === $name ) {
+				$name = 'href';
+			}
+
+			$link_tag .= sprintf( ' %s="%s"', $name, esc_attr( $value ) );
+		}
+		$link_tag .= ">\n";
```
dmsnell, line 78: this is also a reasonable place for the Tag Processor, though obviously not essential. it simply removes the need to remember to escape, and joins together the split tag/html/php syntax.

```php
$link_tag = new WP_HTML_Tag_Processor( '<link data-ilo-added-tag rel="preload" fetchpriority="high" as="image">' );
foreach ( array_filter( $img_attributes ) as $name => $value ) {
	switch ( $name ) {
		case 'srcset':
		case 'sizes':
			$name = "image{$name}";
			break;

		case 'src':
			$name = 'href';
			break;	
	}

	$link_tag->set_attribute( $name, $value );
}
$link_html = $link_tag->get_updated_html();
```

I'm hoping to have basic templating in place for 6.5, which should turn this into something potentially nicer still.

```php
$attributes = [];
foreach ( array_filter( $img_attributes ) as $name => value ) {
	switch ( $name ) {
		case 'srcset':
		case 'sizes':
			$attributes[ "image{$name}" ] = $value;
			continue;

		case 'src':
			$attributes['href'] = $value;
			continue;
	}

	$attributes[ $name ] = $value;
}

$link_tag = WP_HTML::render(
	'<link data-ilo-added-tag rel="preload" fetchpriority="high" as="image" ...attrs>',
	array( 'attrs' => $attributes )
);
```

dmsnell (commented):
@westonruter there's quite a bit in this PR! I can't say that I fully understanding it, but as far as the HTML Processor goes I'm glad you're experimenting with the lower-level tooling for your own needs. plugins are the right place to push ahead with riskier code.

it hasn't been my intention to expose which nth-child a node is compared to the parent, but I was going to build a CSS selector that queries that when prompted. maybe that's a good thing to build in from the start.

my outlook has been positive on the performance side for the eventual HTML Processor, but it's still opaque until it supports everything it needs to 🤷‍♂️

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
westonruter: Ah, very helpful. Thanks!

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
westonruter: Thank you. I copied this list from amp-toolbox-php without validating that the list matches what I `@link`'ed to in the spec.

image

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
westonruter: Very interesting. I wasn't aware of the `WP_HTML_Processor::has_self_closing_flag()` method. 

Nevertheless, maybe having a list of all HTML elements isn't needed here because: (1) if `math` or `svg` are ancestors, we can assume that all tags with self-closing flags will close the tag, and (2) custom elements always have hyphens in them, so if present we can also honor the self-closing tag.

Nevertheless, I just checked your example and it doesn't seem the second example with a custom element is actually true. I adapted your example to use a `span` instead of `div` (since a `div` closes an open `p`):

```html
<p>There are no empty <span/> tags. This is inside the SPAN</p>.
<wp-group>Is a custom element, and so <wp-group /> is self-closing while others aren't.</wp-group>
```

And Chrome renders this as:

image

So perhaps such self-closing foreign elements are limited to MathML and SVG contexts?

westonruter: For example:

```html
<svg>Is a custom element, and so <g /> is self-closing while others aren't.</svg>
```

image

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
westonruter: Resolved in 957d2c45b69ae4ed7a216c7ed50d0369bf120268

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
westonruter: Improved in 957d2c45b69ae4ed7a216c7ed50d0369bf120268

modules/images/image-loading-optimization/detection/detect.js
westonruter: Yes, exactly. Currently I'm thinking that when detection is needed that the server would add breadcrumbs via a data attribute, for example:

```html
<img data-ilo-breadcrumbs="html,0 body,1 main,2 figure,10 img,0" ...>
```

When detection is not needed, such data attributes would not be present.

modules/images/image-loading-optimization/detection/detect.js
westonruter: Yeah, good point. I'm removing this todo.

```suggestion
```

westonruter: This being said, we could segment URL metrics by logged-in users versus non-logged-in users. For example, in `ilo_get_normalized_query_vars()` we could include:

```diff
--- a/modules/images/image-loading-optimization/storage/data.php
+++ b/modules/images/image-loading-optimization/storage/data.php
@@ -89,6 +89,8 @@ function ilo_get_normalized_query_vars(): array {
 		);
 	}
 
+	$normalized_query_vars['logged_in'] = is_user_logged_in();
+
 	return $normalized_query_vars;
 }
```

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
westonruter: See https://github.com/WordPress/performance/pull/884#discussion_r1412403078

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -0,0 +1,301 @@
...
+	 * For example '<p>foo<h1>bar</h1>' should parse the same as '<p>foo</p><h1>bar</h1>'.
+	 *
+	 * @link https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element
+	 *
+	 * @var string[]
+	 */
+	const P_CLOSING_TAGS = array(
+		'ADDRESS',
+		'ARTICLE',
+		'ASIDE',
```
felixarntz, line 60: Why do we need these lists here? What specifically in regards to void and p closing tags is needed here, and why? Is there a bug in the `WP_HTML_Tag_Processor`, or crucial limitation?

It would be great if you could clarify the need for all this in relation to this module. Right now I'm wondering whether this module should need all those functionalities - like why are those special tags relevant for image optimization?

modules/images/image-loading-optimization/detection/detect.js
felixarntz: Can you add a `TODO` comment? :)

modules/images/image-loading-optimization/optimization.php
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+require_once __DIR__ . '/class-ilo-html-tag-processor.php';
```
felixarntz: Similar to my comment on the other PR, can you put this into `load.php` to have all file requirements in one place?

modules/images/image-loading-optimization/optimization.php
```diff
@@ -0,0 +1,174 @@
...
+	$lcp_elements_by_minimum_viewport_widths = ilo_get_lcp_elements_by_minimum_viewport_widths( $url_metrics_grouped_by_breakpoint );
+	$all_breakpoints_have_url_metrics        = count( array_filter( $url_metrics_grouped_by_breakpoint ) ) === count( $breakpoint_max_widths ) + 1;
+
+	// TODO: Handle case when the LCP element is not an image at all, but rather a background-image.
+	// Prepare to set fetchpriority attribute on the image when all breakpoints have the same LCP element.
+	if (
+		// All breakpoints share the same LCP element (or all have none at all).
+		1 === count( $lcp_elements_by_minimum_viewport_widths )
+		&&
+		// The breakpoints don't share a common lack of an LCP element.
```
felixarntz: Isn't this rather "lack of an LCP image"?

modules/images/image-loading-optimization/optimization.php
```diff
@@ -0,0 +1,174 @@
...
+		if ( 'IMG' !== $tag_name ) {
+			continue;
+		}
+
+		// Ensure the fetchpriority attribute is set on the element properly.
+		if ( $common_lcp_element && $processor->get_breadcrumbs() === $common_lcp_element['breadcrumbs'] ) {
+			if ( 'high' === $processor->get_attribute( 'fetchpriority' ) ) {
+				$processor->set_attribute( 'data-ilo-fetchpriority-already-added', true );
+			} else {
+				$processor->set_attribute( 'fetchpriority', 'high' );
```
felixarntz, line 167: Shouldn't we also remove a potential `loading="lazy"` attribute here? Or is this completely irrelevant in that case and `fetchpriority="high"` takes precendence?

If the latter, I think that would be good to document in the code somewhere, e.g. in this function's doc block.

modules/images/image-loading-optimization/optimization.php
```diff
@@ -0,0 +1,174 @@
...
+		} elseif ( $all_breakpoints_have_url_metrics && $processor->get_attribute( 'fetchpriority' ) ) {
+			// Note: The $all_breakpoints_have_url_metrics condition here allows for server-side heuristics to
+			// continue to apply while waiting for all breakpoints to have metrics collected for them.
+			$processor->set_attribute( 'data-ilo-removed-fetchpriority', $processor->get_attribute( 'fetchpriority' ) );
+			$processor->remove_attribute( 'fetchpriority' );
+		}
+
+		// Capture the attributes from the LCP elements to use in preload links.
+		foreach ( $lcp_elements_by_minimum_viewport_widths as &$lcp_element ) {
+			if ( $lcp_element && $lcp_element['breadcrumbs'] === $processor->get_breadcrumbs() ) {
```
felixarntz: Let's optimize this, as a `foreach` for every tag just to find the right entry isn't great and easy to resolve. We can either update `$lcp_elements_by_minimum_viewport_widths` or create a copy from it in another variable where the entries are keyed by the `breadcrumbs` value.

Alternatively, since we want to update `$lcp_elements_by_minimum_viewport_widths` here, we could make the new variable a simple lookup, where keys are the `breadcrumbs` values and values are the respective indexes in `$lcp_elements_by_minimum_viewport_widths`.

That way we can avoid this `foreach` and simply check for `isset( $new_var[ $processor->get_breadcrumbs() ] )`.

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -186,8 +193,17 @@ static function ( $a, $b ) {
  *  3. 481-576 (phablets)
  *  4. >576 (desktop)
  *
+ * The default breakpoints are reused from Gutenberg where the _breakpoints.scss file includes these variables:
+ *
+ *     $break-medium: 782px; // adminbar goes big
```
felixarntz, line 198: Curious about the admin bar comment: Is that relevant? Since we should exclude the admin bar from the considerations here, I would say if this breakpoint is only to cater for that, we would rather need to find a different solution.

In other words, if 782px is a real breakpoint also for the regular frontend content, let's use it. But if it's only relevant as the admin bar breakpoint, we shouldn't need to capture LCP image data for that.

modules/images/image-loading-optimization/optimization.php
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * @param array $lcp_images_by_minimum_viewport_widths LCP images keyed by minimum viewport width, amended with attributes key for the IMG attributes.
+ * @return string Markup for zero or more preload link tags.
+ */
+function ilo_construct_preload_links( array $lcp_images_by_minimum_viewport_widths ): string {
+	$preload_links = array();
+
+	$minimum_viewport_widths = array_keys( $lcp_images_by_minimum_viewport_widths );
+	for ( $i = 0, $len = count( $minimum_viewport_widths ); $i < $len; $i++ ) {
+		$lcp_element = $lcp_images_by_minimum_viewport_widths[ $minimum_viewport_widths[ $i ] ];
```
felixarntz, line 42: This seems unnecessarily complicated: Why do we key this data by minimum viewport width, when we then use an index here? If we want to keep the viewport width as key, then why don't we use a `foreach` here instead of the `for`? Just makes this less complicated to read.

modules/images/image-loading-optimization/optimization.php
```diff
@@ -0,0 +1,174 @@
...
+		// Prevent preloading src for browsers that don't support imagesrcset on the link element.
+		if ( isset( $img_attributes['src'], $img_attributes['srcset'] ) ) {
+			unset( $img_attributes['src'] );
+		}
+
+		// Add media query if it's going to be something other than just `min-width: 0px`.
+		if ( $minimum_viewport_widths[ $i ] > 0 || isset( $minimum_viewport_widths[ $i + 1 ] ) ) {
+			$media_query = sprintf( '( min-width: %dpx )', $minimum_viewport_widths[ $i ] );
+			if ( isset( $minimum_viewport_widths[ $i + 1 ] ) ) {
+				$media_query .= sprintf( ' and ( max-width: %dpx )', $minimum_viewport_widths[ $i + 1 ] - 1 );
```
felixarntz: Related to the above, it looks like this is the only reason why the `$i` is relevant. A simpler solution would be to capture the last key in the array before the loop, then go through the `foreach` loop and instead of `isset( $minimum_viewport_widths[ $i + 1 ] )` check for `$minimum_viewport_width !== $last_viewport_width`.

felixarntz (commented):
@westonruter This looks great so far. I left a few comments, mostly for minor optimization. My biggest concern/question is in regards to the custom tag processor class, I'm not sure why that needs to be included here.

westonruter:
> My biggest concern/question is in regards to the custom tag processor class, I'm not sure why that needs to be included here.

@felixarntz The custom tag processor is needed in order to be able to obtain breadcrumbs for any given element. Without it, we wouldn't be able to apply optimizations for elements identified during detection.

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
westonruter: These are needed to be able to iterate over tags in the document and be able to correctly keep track of the open tag stack. The `WP_HTML_Tag_Processor` is intentionally very limited in what processing it supports. Namely, it does not have any sense of how tags are related to each other (e.g. that `html` is a parent of `body`). It is a streaming parser that iterates over open tags and close tags. In order to keep track of when one tag is a child of another tag, we have to know the semantics of these tags, whether they are self-closing (void) or how they relate to tags that have optional closers (e.g. here `p`). 

The superset of functionality from `WP_HTML_Tag_Processor` is in `WP_HTML_Processor` which _does_ have capabilities regarding document structure. However, it can't be used yet because it doesn't support all HTML tags yet. Eventually this  class shouldn't be needed as `WP_HTML_Processor`. I've added a comment to reflect this in 0cca4dd2bf49afcfa9af51feaf58a83f7a5f57c8.

westonruter: And yeah, these lists of tags are all related to image optimization in that we need to know them in order to accurately construct breadcrumbs for any given tag.

modules/images/image-loading-optimization/detection/detect.js
westonruter: There is already one:

https://github.com/WordPress/performance/blob/0cca4dd2bf49afcfa9af51feaf58a83f7a5f57c8/modules/images/image-loading-optimization/detection/detect.js#L136

modules/images/image-loading-optimization/optimization.php
westonruter: Done in 83d03621

modules/images/image-loading-optimization/optimization.php
westonruter: At the moment, yes. But other LCP elements will be supported, namely elements with background images. There is a TODO for this above:

https://github.com/WordPress/performance/blob/83d0362170501b5356cea8cb403ba29913ea0ce6/modules/images/image-loading-optimization/optimization.php#L108

modules/images/image-loading-optimization/optimization.php
westonruter: You're right. It would be removed here. But I was intending to leave processing of lazy-loading optimizations to a subsequent PR. I'll add a TODO here for that.

```suggestion
				// TODO: When optimizing lazy-loading, this should also remove any `loading` attribute here.
				$processor->set_attribute( 'fetchpriority', 'high' );
```

modules/images/image-loading-optimization/storage/data.php
westonruter: I'm only including it here because these three lines are all copied from the `_breakpoints.scss` file: https://github.com/WordPress/gutenberg/blob/093d52cbfd3e2c140843d3fb91ad3d03330320a5/packages/base-styles/_breakpoints.scss#L11-L13

And yes, the 782px breakpoint is used in the Columns block. When the window is narrower than this, then the columns get stacked.

modules/images/image-loading-optimization/optimization.php
westonruter: I adopted the second approach in 74145a460146521da14bfdc2f769a5faccf22f29.

Since breadcrumbs are an array, they can't be used as an array key. So I have to convert to a string first. This string I intend to be reusing later as mentioned in https://github.com/WordPress/performance/pull/884#discussion_r1412402227

modules/images/image-loading-optimization/optimization.php
westonruter: I did it this way because for each minimum viewport width we also need to get the following one. So we're iterating over not just one viewport widths, but a window of two throughout the array.

modules/images/image-loading-optimization/optimization.php
westonruter: Humm. We need more than just the last key. We need the next key after the current key. So if the minimum widths are (0, 480, 782), then we need to iterate 3 times and have access to these values:

1. 0, 480-1
2. 480, 782-1
3. 782, null

westonruter: I've made this easier to read in 2758c0dba01a489e0b204cfcbb57d66b4d47b5ef. Let me know if you have anything else in mind to improve this.

modules/images/image-loading-optimization/optimization.php
westonruter: I went ahead and made this change: 9d8261535314ef0bfe79ba2294ab0150fa54f764.

Nevertheless, since the LCP images also get `preload` links added for them, the presence of `loading=lazy` actually won't hurt performance for the LCP image. This is based on my findings in https://github.com/WordPress/performance/issues/117#issuecomment-1773456772.

Conversely, in the `else` condition here, we actually need to _add_ `loading=lazy` to breakpoint-specific LCP images if they aren't visible on all breakpoints. This I'll leave for another PR.

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
dmsnell: by golly, you're right. I have been misinterpreting "foreign element" for a long time now! thankfully I haven't gotten to the place in the HTML Processor where that matters.

thank you very much for pointing this out.

modules/images/image-loading-optimization/optimization.php
```diff
@@ -0,0 +1,208 @@
...
+ *
+ * @param array<array{tag: string, index: int}> $breadcrumbs Breadcrumbs.
+ * @return string Breadcrumb string.
+ */
+function ilo_construct_breadcrumbs_string( array $breadcrumbs ): string {
+	return implode(
+		' ',
+		array_map(
+			static function ( $breadcrumb ) {
+				return sprintf( '%s,%s', $breadcrumb['tag'], $breadcrumb['index'] );
```
westonruter, line 97: In another PR I'm working on, I'm going to make this into an XPath constructor.

```suggestion
function ilo_construct_breadcrumbs_xpath( array $breadcrumbs ): string {
	return implode(
		'',
		array_map(
			static function ( $breadcrumb ) {
				return sprintf( '/%s[%d]', $breadcrumb['tag'], $breadcrumb['index'] );
```

westonruter: See https://github.com/WordPress/performance/pull/892

modules/images/image-loading-optimization/detection/detect.js
westonruter: Done in the next PR: https://github.com/WordPress/performance/pull/892

modules/images/image-loading-optimization/optimization.php
```diff
@@ -0,0 +1,209 @@
...
+	return implode( '', $preload_links );
+}
+
+/**
+ * Constructs a breadcrumbs string from a breadcrumbs array.
+ *
+ * @param array<array{tag: string, index: int}> $breadcrumbs Breadcrumbs.
+ * @return string Breadcrumb string.
+ */
+function ilo_construct_breadcrumbs_string( array $breadcrumbs ): string {
```
westonruter, line 92: Function moved to `ILO_HTML_Tag_Processor` method in https://github.com/WordPress/performance/pull/892

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
felixarntz: Thanks, that makes sense. I wonder how this version of the tag processor is for performance.

For the plugin we don't need to worry about this too much, but for a potential core proposal we would need to spend more time on this - particularly as I'm not sure we would want two separate versions of the tag processor in core. And we will only be able to use this one throughout if it doesn't have notable negative performance implications.

modules/images/image-loading-optimization/optimization.php
felixarntz: Nit-pick, but could you still adjust this comment to say "LCP image"? Regardless of potential background image support in the future, my point here is mostly that "lack of an LCP element" is not a thing, as every page has an LCP element.

modules/images/image-loading-optimization/optimization.php
felixarntz: Thanks, that looks great. Good point on having to transform breadcrumbs to a string.

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
westonruter: It should have about the same performance as `WP_HTML_Tag_Processor` as it is just iterating over tags and is doing much less than `WP_HTML_Processor`, which is less predictable for performance.

modules/images/image-loading-optimization/optimization.php
felixarntz: Ah, good point, I missed that. Can you maybe add a comment above this to clarify? For example:
`// This uses a for loop to be able to access the following element within the iteration, using a numeric index.`

modules/images/image-loading-optimization/optimization.php
westonruter: Done in 185ebf6c

modules/images/image-loading-optimization/optimization.php
westonruter: Done in 1d259b5a

felixarntz (approved):
@westonruter Thanks for the updates, LGTM!

westonruter:
It would be nice if pull requests that aren't targeting `trunk` or a release branch could be merged with only a single reviewer. Getting two reviews for every review regardless seems like a bottleneck.

westonruter:
@adamsilverstein or @swissspidy could either of you drop in a LGTM?

mukeshpanchal27 (approved):

adamsilverstein (approved):
Tremendous

PR #886, opened 2023-11-22
Add oEmbed optimizer module

adamsilverstein:
## Summary

Full issue - https://github.com/WordPress/performance/issues/888

Attempts to optimize oEmbeds using two techniques:
* For oEmbeds delivered as iframes, ensure the `loading=lazy` attribute is present
* For oEmbeds built using script tags:
	* move the script src to a data attribute
	* add a page level interaction observer that resets `src` before the user scrolls to the oEmbed

Related: https://core.trac.wordpress.org/ticket/58773

## Checklist


modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,99 @@
...
+ * @since n.e.x.t
+ *
+ * @param string $html The oEmbed HTML.
+ * @return string
+ */
+function plab_optimize_oembed_html( $html ) {
+	global $oembed_lazy_load_scripts;
+
+	// Find script tags in the oembed using regex.
+	preg_match_all( '/<script\b[^>]*>([\s\S]*?)<\/script>/mi', $html, $matches );
```
sgomes: Any chance we could consider using the HTML API instead of regexes, for this module? It feels like a missed opportunity to introduce new regexes, when [we know all the issues they come with](https://make.wordpress.org/core/2023/08/19/progress-report-html-api/#why-the-html-api).

Since we don't appear to be targeting older WP versions, this seems like the perfect opportunity to do it, as oEmbeds aren't as common as other hooks for which we've considered using the API (such as image processing). This means we won't be introducing unreasonable processing delays — even though the HTML API has pretty good performance at this point, to my knowledge.

CC @dmsnell

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,99 @@
...
+function plab_optimize_oembed_lazy_load_scripts() {
+	global $oembed_lazy_load_scripts;
+
+	if ( ! $oembed_lazy_load_scripts ) {
+		return;
+	}
+	?>
+	<script>
+		(function() {
+			var lazyEmbeds = document.querySelectorAll( 'script[data-lazy-embed-src]' );
```
sgomes: Lazy loading oEmbed scripts with an Intersection Observer feels like a reasonable approach, but I'm somewhat concerned with potential orchestration issues if an oEmbed makes use of multiple scripts; we should ensure that they execute in order when using this approach.

I don't have the opportunity to test things right now myself, but we should also ensure that this works correctly in all major browsers. I seem to recall that some browsers (Firefox, IIRC) don't trigger script execution when changing the `src` in an existing `<script>`, and instead require a new `<script>` tag to be inserted.

sgomes (commented):
Thank you for the PR, @adamsilverstein! I think this has the potential for some great wins! 👍

I left some non-blocking comments for your consideration. I'm not sure if I'll have the opportunity to re-review because of some upcoming AFK, so please feel free to move ahead at your discretion!

modules/js-and-css/optimize-oembeds/hooks.php
adamsilverstein: Sure, I'm happy to switch this to try using `WP_HTML_Tag_Processor`, I used these regexes reflexively :)

modules/js-and-css/optimize-oembeds/hooks.php
adamsilverstein: > Lazy loading oEmbed scripts with an Intersection Observer feels like a reasonable approach, but I'm somewhat concerned with potential orchestration issues if an oEmbed makes use of multiple scripts; we should ensure that they execute in order when using this approach.

So far in my testing of several types, there is a single script tag that loads the oEmbed. We could enforce that, only lazy load if there is a single script. I'd also like to add tests that validate the transforms, we can probably cover all the supported oembed endpoints in the tests.

> I don't have the opportunity to test things right now myself, but we should also ensure that this works correctly in all major browsers. I seem to recall that some browsers (Firefox, IIRC) don't trigger script execution when changing the src in an existing <script>, and instead require a new <script> tag to be inserted.

Ah yes, I was expecting to run into some browser compatibility issues and have not tested yet.  Hopefully we can find a single approach that works across our supported browsers - inserting a new script tag sounds reasonable. I started with this simple src approach for a POC. I will research which approach is best. I also wonder if we can increase the threshold for low bandwidth connections (so the oEmbeds start loading earlier).

modules/js-and-css/optimize-oembeds/hooks.php
adamsilverstein: switched to WP_HTML_Tag_Processor in https://github.com/WordPress/performance/pull/886/commits/21f29082121061704e84d58a185b79d60b7198c8

adamsilverstein: Created a Gist with the regex version which could be useful for optimization plugins that want to ensure compatibility with older WordPress versions - https://gist.github.com/adamsilverstein/413bc328b3ae01113085fb9edabce963

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,89 @@
...
+	}
+	$p = new WP_HTML_Tag_Processor( $html );
+	if ( $p->next_tag( 'script' ) ) {
+		$oembed_lazy_load_scripts = true;
+		$p->set_attribute( 'data-lazy-embed-src', $p->get_attribute( 'src' ) );
+		$p->set_attribute( 'src', '' );
+		$html = $p->get_updated_html();
+	}
+
+	return $html;
```
westonruter: Seems these could be combined.

```suggestion
		}
	}
	$p = new WP_HTML_Tag_Processor( $html );
	if ( $p->next_tag( 'script' ) ) {
		$oembed_lazy_load_scripts = true;
		$p->set_attribute( 'data-lazy-embed-src', $p->get_attribute( 'src' ) );
		$p->set_attribute( 'src', '' );
	}

	return $p->get_updated_html();
```

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,89 @@
...
+			$html = $p->get_updated_html();
+		}
+	}
+	$p = new WP_HTML_Tag_Processor( $html );
+	if ( $p->next_tag( 'script' ) ) {
+		$oembed_lazy_load_scripts = true;
+		$p->set_attribute( 'data-lazy-embed-src', $p->get_attribute( 'src' ) );
+		$p->set_attribute( 'src', '' );
+		$html = $p->get_updated_html();
+	}
```
dmsnell, line 62: to avoid scanning the document twice we can scan `next_tag()` and attach our behaviors based on the currently-matched tag name. the Tag Processor has to scan every tag anyway, so we're not skipping anything other than an extra function call per tag.

```php
$p = new WP_HTML_Tag_Processor( $html );

// These are only important if we expect to encounter multiple scripts
// or iframes but only want to process the first. This is here to preserve
// the behavior of the snippet in the PR, but I don't know if that was
// intentional or happenstance.
$have_seen_iframe = false;
$have_seen_script = false;

while ( ! ( $has_seen_iframe && $has_seen_script ) && $p->next_tag() ) {
	if ( 'IFRAME' === $p->get_tag() && ! $has_seen_iframe ) {
		$has_seen_iframe = true;

		$loading_value = $p->get_attribute( 'loading' );
		if ( ! is_string( $loading_value ) || empty( $loading_value ) ) {
			$p->set_attribute( 'loading', 'lazy' );
		}

		continue;
	}

	if ( 'SCRIPT' === $p->get_tag() && ! $has_seen_script ) {
		$has_seen_script = true;

		$oembed_lazy_load_scripts = true;
		$p->set_attribute( 'data-lazy-embed-src', $p->get_attribute( 'src' ) );
		$p->set_attribute( 'src', '' );
	}
}

return $p->get_updated_html();
```

modules/js-and-css/optimize-oembeds/hooks.php
westonruter: Sorry, I missed that there were two instances of `WP_HTML_Tag_Processor`. This is made obsolete anyway by https://github.com/WordPress/performance/pull/886#discussion_r1406860617

modules/js-and-css/optimize-oembeds/hooks.php
adamsilverstein: > the Tag Processor has to scan every tag anyway, so we're not skipping anything other than an extra function call per tag.

Ah, that is good to know, I will adjust accordingly. in general these HTML blocks will have either an iframe or a script tag, so definitely makes sense to only make one pass! Thanks.

modules/js-and-css/optimize-oembeds/hooks.php
adamsilverstein: good point. I'm going to combine these into a single loop anyway as @dmsnell suggested.

modules/js-and-css/optimize-oembeds/hooks.php
adamsilverstein: > ! is_string( $loading_value )

@dmsnell why is this check needed (in `if ( ! is_string( $loading_value ) || empty( $loading_value ) ) {`)

dmsnell: if it's no problem to set them all as lazy it gets simpler in the code.

```php
$p = new WP_HTML_Tag_Processor( $html );
while ( $p->next_tag() ) {
	switch ( $p->get_tag() ) {
		case 'IFRAME':
			$loading_value = $p->get_attribute( 'loading' );
			if ( ! is_string( $loading_value ) || empty( $loading_value ) ) {
				$p->set_attribute( 'loading', 'lazy' );
			}
			break;

		case 'SCRIPT':
			$oembed_lazy_load_scripts = true;
			$p->set_attribute( 'data-lazy-embed-src', $p->get_attribute( 'src' ) );
			$p->set_attribute( 'src', '' );
			break;
	}
}

return $p->get_updated_html();
```

alternatively, to set the first of either of those tags replace the `break` with `return $p->get_updated_html()` or break out of both structures with `break 2`

adamsilverstein: updated to single loop in https://github.com/WordPress/performance/pull/886/commits/9eab381384c362638f70e76d8b702863d17ef43d

adamsilverstein: removed is_string check in https://github.com/WordPress/performance/pull/886/commits/3701f2d04048ac4445bdbb3aab134f2c45a50732 - since `get_tag` wither returns a string or null and `empty` works as expected on either

dmsnell: @adamsilverstein if an attribute is a boolean attribute then in the DOM it will return an empty string but the HTML API expresses that as `true` so that code can recognize the difference. unfortunately, `empty(true) === false`, so we might think there's a value set when in fact it's empty in the DOM.

I'm not happy with this, but I've not yet seen a great solution. I've thought about adding a new method that provides the stringy value attribute, like `get_attribute_string()` or `get_attribute_value()`

dmsnell: (the only practical drawback here is that we'll miss adding the lazy attribute if we encounter something like `<script loading src="…">`, which may not seem like a big deal because it's still safe, just not 100% correct)

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,92 @@
...
+				} );
+			}, {
+				rootMargin: '0px 0px 500px 0px',
+				threshold: 0
+			} );
+			parents.forEach( function( lazyEmbed ) {
+				lazyEmbedObserver.observe( lazyEmbed );
+			} );
+		})();
+	</script>
```
westonruter: I haven't tested this, but a few improvements to suggest here:

1. We can use ES6+ syntax since `IntersectionObserver` is only supported in modern browsers anyway.
2. We can combine the two loops over embeds into one.
3. We can avoid having to query selector for the embed script which we previously already found by storing it in a `Map`.
4. The inline script can be printed as a `module` script which will prevent it from blocking rendering, as module scripts are delayed until the DOM has loaded.
5. This also implements the replacement of the script instead of just setting the `src`.

```suggestion
	<script type="module">
        const lazyEmbedsScripts = document.querySelectorAll( 'script[data-lazy-embed-src]' );
        const lazyEmbedScriptsByParents = new Map();

        const lazyEmbedObserver = new IntersectionObserver( 
            ( entries ) => {
                for ( const entry of entries ) {
                    if ( entry.isIntersecting ) {
                        const lazyEmbedParent = entry.target;
                        const lazyEmbedScript = lazyEmbedScriptsByParents.get( lazyEmbedParent );
                        const embedScript = lazyEmbedScript.cloneNode();
                        embedScript.src = lazyEmbedScript.dataset.lazyEmbedSrc;
                        lazyEmbedScript.replaceWith( embedScript );
                        lazyEmbedObserver.unobserve( lazyEmbedParent );
                    }
                }
            }, 
            {
                rootMargin: '0px 0px 500px 0px',
                threshold: 0
            } 
        );
        
        for ( const lazyEmbedScript of lazyEmbedsScripts ) {
            const lazyEmbedParent = lazyEmbedScript.parentNode;
            lazyEmbedScriptsByParents.set( lazyEmbedParent, lazyEmbedScript );
            lazyEmbedObserver.observe( lazyEmbedParent );
        }
	</script>
```

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,92 @@
...
+			}
+			return $p->get_updated_html();
+
+		} elseif ( 'SCRIPT' === $p->get_tag() ) {
+			$oembed_lazy_load_scripts = true;
+			$p->set_attribute( 'data-lazy-embed-src', $p->get_attribute( 'src' ) );
+			$p->set_attribute( 'src', '' );
+			return $p->get_updated_html();
+		}
+	}
```
westonruter: An embed may have both an `iframe` _and_ a `script`. For example, a Crowdsignal poll looks like this:

```html
<script type="text/javascript" charset="utf-8" src="https://secure.polldaddy.com/p/7012505.js"></script>
<noscript><iframe title="Which design do you prefer?" src="https://poll.fm/7012505/embed" frameborder="0" class="cs-iframe-embed"></iframe></noscript>
```

So this wouldn't be handled properly here since the script would never get modified.

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,92 @@
...
+ * @since n.e.x.t
+ */
+function perflab_optimize_oembed_lazy_load_scripts() {
+	global $oembed_lazy_load_scripts;
+
+	if ( ! $oembed_lazy_load_scripts ) {
+		return;
+	}
+	?>
+	<script>
```
westonruter: Also, this can use `wp_print_inline_script_tag()`.

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,92 @@
...
+	// Find the first iframe or script tag to act on.
+	while ( $p->next_tag() ) {
+		if ( 'IFRAME' === $p->get_tag() ) {
+			$loading_value = $p->get_attribute( 'loading' );
+			if ( empty( $loading_value ) ) {
+				$p->set_attribute( 'loading', 'lazy' );
+			}
+			return $p->get_updated_html();
+
+		} elseif ( 'SCRIPT' === $p->get_tag() ) {
```
westonruter, line 47: This needs to make sure it only targets external scripts:

```suggestion
		} elseif ( 'SCRIPT' === $p->get_tag() && $p->get_attribute( 'src' ) ) {
```

Example oEmbed for Crowdsignal survey responses:

```html
<div class="pd-embed" data-settings="{&quot;type&quot;:&quot;iframe&quot;,&quot;auto&quot;:true,&quot;domain&quot;:&quot;rydk.survey.fm&quot;,&quot;id&quot;:&quot;test-survey&quot;}"></div>
<script type="text/javascript">
(function(d,c,j){if(!document.getElementById(j)){var pd=d.createElement(c),s;pd.id=j;pd.src=('https:'==document.location.protocol)?'https://polldaddy.com/survey.js':'http://i0.poll.fm/survey.js';s=document.getElementsByTagName(c)[0];s.parentNode.insertBefore(pd,s);}}(document,'script','pd-embed'));
</script>
```

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,92 @@
...
+			$oembed_lazy_load_scripts = true;
+			$p->set_attribute( 'data-lazy-embed-src', $p->get_attribute( 'src' ) );
+			$p->set_attribute( 'src', '' );
+			return $p->get_updated_html();
+		}
+	}
+
+	return $html;
```
westonruter: Sorry, I haven't tested this, but it should:

1. Prevent making scripts lazy if there is an inline script.
2. Only make script lazy if there is a single external script (since if there are multiple they may not get loaded in the right order).
3. Ensure that both the `iframe` and the `script` are made lazy if both occur in the same embed.

```suggestion
	$iframe_count      = 0;
	$script_count      = 0;
	$has_inline_script = false;

	// Locate the iframes and scripts.
	while ( $p->next_tag() ) {
		if ( 'IFRAME' === $p->get_tag() ) {
			$loading_value = $p->get_attribute( 'loading' );
			if ( empty( $loading_value ) ) {
				++$iframe_count;
				$p->set_bookmark( 'iframe' );
			}
		} elseif ( 'SCRIPT' === $p->get_tag() ) {
			if ( ! $p->get_attribute( 'src' ) ) {
				$has_inline_script = true;
			} else {
				++$script_count;
				$p->set_bookmark( 'script' );
			}
		}
	}

	// If there was only one non-inline script, make it lazy.
	if ( 1 === $script_count && ! $has_inline_script ) {
		$oembed_lazy_load_scripts = true;
		$p->seek( 'script' );
		$p->set_attribute( 'data-lazy-embed-src', $p->get_attribute( 'src' ) );
		$p->remove_attribute( 'src' );
	}

	// If there was only one iframe, make it lazy.
	if ( 1 === $iframe_count ) {
		$p->seek( 'frame' );
		$p->set_attribute( 'loading', 'lazy' );
	}

	return $p->get_updated_html();
```

modules/js-and-css/optimize-oembeds/hooks.php
adamsilverstein: Ah, good point and thanks for the example - I wasn't sure this was used. I'll test this without the `else` with a polldaddy embed.

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,96 @@
...
+        const lazyEmbedScriptsByParents = new Map();
+
+        const lazyEmbedObserver = new IntersectionObserver( 
+            ( entries ) => {
+                for ( const entry of entries ) {
+                    if ( entry.isIntersecting ) {
+                        const lazyEmbedParent = entry.target;
+                        const lazyEmbedScript = lazyEmbedScriptsByParents.get( lazyEmbedParent );
+                        const embedScript = lazyEmbedScript.cloneNode();
+                        embedScript.src = lazyEmbedScript.dataset.lazyEmbedSrc;
```
westonruter: For some reason this is not working for me. I tested with a Tweet and a TikTok and neither of them had their scripts load. I was able to get them to load by creating a fresh `script` instead of cloning it:

```suggestion
                        const embedScript = document.createElement( 'script' );
                        for ( const attr of lazyEmbedScript.attributes ) {
                            if ( attr.nodeName === 'src' ) {
                                // Even though the src attribute is absent, the browser seems to presume it is present.
                                continue;
                            }

                            embedScript.setAttribute(
                                attr.nodeName === 'data-lazy-embed-src' ? 'src' : attr.nodeName,
                                attr.nodeValue
                            );
                        }
                        lazyEmbedScript.replaceWith( embedScript );
```

westonruter:
I just tested this and I'm seeing some curious results. With 4x CPU slowdown and Fast 3G network emulation, there is no difference in LCP on a page that has a featured image, followed by 10 paragraphs of Lorem Ipsum, and then a YouTube video, TikTok, and Tweet.

Without optimization: https://trace.cafe/t/55ylEfwjsM
With optimization: https://trace.cafe/t/SKl8g2oATn

The page load time is drastically improved, from 11 seconds to 5 seconds. But the load event isn't a CWV metric, of course.

Probably the real win would be if JS is being used to render the LCP element (e.g. a carousel) rather than using something static.

<details><summary>Sample Post Content</summary>

```html
<!-- wp:image {"id":23,"sizeSlug":"large","linkDestination":"none"} -->
<figure class="wp-block-image size-large"><img src="http://localhost:10008/wp-content/uploads/2023/05/American_bison_k5680-1-1024x668.webp" alt="" class="wp-image-23"/></figure>
<!-- /wp:image -->

<!-- wp:paragraph -->
<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Etiam commodo purus diam, ac consectetur justo dapibus non. Donec consequat erat ac ipsum finibus laoreet. Nunc accumsan pulvinar ipsum eu vulputate. Cras sodales vel purus sed tincidunt. Mauris lobortis sollicitudin risus, nec pulvinar sem interdum et. Phasellus vehicula sollicitudin urna, nec rhoncus mauris efficitur id. Nulla sed molestie lectus. Proin tincidunt, felis at eleifend laoreet, lectus est interdum arcu, ac tempor metus sapien quis nunc.</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Curabitur non pulvinar ex. In ligula velit, blandit dignissim posuere non, lacinia et lorem. Cras erat metus, molestie vel consectetur vitae, sodales eu leo. Phasellus eget vestibulum sem, ut rhoncus quam. Etiam in convallis nunc. Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos. Mauris sed aliquet felis, at ornare purus. Curabitur eu blandit quam, at gravida eros. Integer ullamcorper dictum enim, commodo venenatis est molestie sed.</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Etiam scelerisque tellus erat. Mauris finibus nunc vitae ex sollicitudin venenatis. Integer dapibus ultrices tellus, a rutrum magna pharetra quis. Donec imperdiet imperdiet dolor non semper. Vestibulum interdum bibendum lorem et sollicitudin. Proin ac dapibus sapien. Aenean non velit ultrices, ultricies sem eu, ullamcorper diam. Aenean tristique rhoncus sodales. Aenean pharetra mollis sem. Aliquam a augue vitae lorem fringilla dignissim. Donec congue finibus arcu eget bibendum.</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Ut faucibus vel nulla non sodales. Sed laoreet eros id vehicula iaculis. Vivamus semper neque est. Pellentesque ultricies, arcu sed tristique pretium, turpis purus consequat nibh, id pulvinar tellus metus ac libero. Donec commodo nec mi id congue. Nam viverra in lacus consectetur faucibus. Sed nec odio eleifend ipsum pulvinar interdum. Sed sit amet tincidunt nisl. Praesent lacus mi, fermentum et erat vel, ornare pharetra massa. Donec ac nunc sagittis, scelerisque lectus sed, tempor lectus. Aenean sed lacinia libero. Pellentesque volutpat erat tincidunt ex sollicitudin ullamcorper. Donec ut metus facilisis, tempus erat vitae, blandit est. Sed venenatis est eu dui viverra, id imperdiet magna pretium.</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Aenean eget cursus augue, ac pulvinar justo. Duis ornare vitae arcu vitae scelerisque. Ut non libero vitae lorem maximus elementum. Ut consectetur ligula non massa ultricies tempus. Maecenas ultricies lectus at justo efficitur, non pellentesque nulla congue. Aliquam vulputate luctus nisl in iaculis. Nulla facilisi.</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Cras tristique, lacus in malesuada imperdiet, neque erat rhoncus metus, non pellentesque lorem neque nec felis. Etiam gravida magna purus, et ullamcorper massa gravida et. Vivamus tincidunt ut nulla in vestibulum. In id rutrum est, nec facilisis sapien. Vivamus pharetra dapibus ipsum. Donec vehicula nulla id volutpat tincidunt. Proin sodales, ex in lobortis eleifend, tellus tortor laoreet lacus, eget interdum nisi mauris et dui. Duis augue ipsum, ultricies eget dui ut, ullamcorper mattis lorem. Pellentesque vestibulum quam quis posuere euismod. Donec suscipit ut lacus at sollicitudin. Mauris rutrum pellentesque tristique. Sed fermentum dapibus nisl sit amet gravida. Suspendisse eros mauris, viverra pretium blandit eu, rhoncus et ante. Etiam scelerisque risus nec mauris vulputate venenatis.</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Nulla tempus elementum sem vel fermentum. Praesent varius tincidunt aliquet. Donec pretium efficitur ullamcorper. Maecenas tempor feugiat eros, eget suscipit magna suscipit at. Mauris fermentum in odio sit amet dignissim. Proin euismod efficitur hendrerit. Sed eleifend pretium risus quis elementum. Quisque felis eros, condimentum non ultricies id, hendrerit eget velit.</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>In dapibus diam a sapien tempus viverra. Aenean convallis risus in elit pharetra, vel tempus turpis tincidunt. Maecenas ultricies sem urna, nec dignissim mauris consequat blandit. Maecenas eget posuere arcu, nec iaculis eros. Sed posuere ex eget sagittis finibus. Suspendisse scelerisque nisi auctor venenatis ultrices. Morbi eu massa id orci imperdiet imperdiet. Ut lobortis nibh porta, volutpat orci ac, vestibulum quam. Mauris pellentesque nisi et libero luctus commodo. Aliquam scelerisque, justo ac efficitur mollis, purus eros porttitor quam, a auctor dui ipsum ut urna.</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas. Vivamus lacus lorem, fermentum vitae augue a, gravida molestie nisi. Cras aliquam luctus lacinia. Aenean vel tellus nec nisl finibus tincidunt id id arcu. Etiam lacinia condimentum interdum. Aenean vitae congue nisl. Vivamus placerat metus pulvinar aliquet ultricies. Etiam at lacus a purus congue pulvinar vitae a metus. Nulla odio felis, varius eu dignissim et, pretium in ex. Pellentesque mattis risus vitae est elementum, ut aliquam odio iaculis. Sed vestibulum mi ac lectus molestie porta. Integer a convallis sapien. Praesent nisl ex, vulputate vitae massa non, porta aliquet erat. Phasellus a velit gravida, sollicitudin ligula sit amet, ultrices eros. Aliquam erat volutpat.</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Pellentesque finibus justo magna, eu vehicula arcu pharetra vitae. Nullam eget dictum sem. Vestibulum rutrum ultricies nibh, nec laoreet augue porta at. Aliquam ultricies ligula diam, in fringilla elit varius ut. Etiam in congue velit. Praesent posuere venenatis libero ut placerat. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec porttitor nec orci quis ultrices. Donec sollicitudin orci tellus, sed venenatis arcu tincidunt et.</p>
<!-- /wp:paragraph -->

<!-- wp:embed {"url":"https://youtu.be/3KtWfp0UopM?si=WPX-Q-BTugpSeOo-","type":"video","providerNameSlug":"youtube","responsive":true,"className":"wp-embed-aspect-16-9 wp-has-aspect-ratio"} -->
<figure class="wp-block-embed is-type-video is-provider-youtube wp-block-embed-youtube wp-embed-aspect-16-9 wp-has-aspect-ratio"><div class="wp-block-embed__wrapper">
https://youtu.be/3KtWfp0UopM?si=WPX-Q-BTugpSeOo-
</div></figure>
<!-- /wp:embed -->

<!-- wp:embed {"url":"https://www.tiktok.com/@deeptomcruise/video/6932166297996233989?is_from_webapp=1\u0026sender_device=pc\u0026web_id=7187900767931729451","type":"video","providerNameSlug":"tiktok","responsive":true} -->
<figure class="wp-block-embed is-type-video is-provider-tiktok wp-block-embed-tiktok"><div class="wp-block-embed__wrapper">
https://www.tiktok.com/@deeptomcruise/video/6932166297996233989?is_from_webapp=1&amp;sender_device=pc&amp;web_id=7187900767931729451
</div></figure>
<!-- /wp:embed -->

<!-- wp:embed {"url":"https://twitter.com/googlechrome/status/1747307793055617249","type":"rich","providerNameSlug":"twitter","responsive":true} -->
<figure class="wp-block-embed is-type-rich is-provider-twitter wp-block-embed-twitter"><div class="wp-block-embed__wrapper">
https://twitter.com/googlechrome/status/1747307793055617249
</div></figure>
<!-- /wp:embed -->
```

</details>

westonruter:
Well, I stand corrected. I was just testing in DevTools. I just tried again with `benchmark-web-vitals` and over 20 requests the LCP-TTFB is reduced (median) from 77.05ms to 66.85ms (~13% reduction).

When I throttle CPU 4x, the LCP-TTFB is reduced (median) from 396.75ms to 337.1ms (~15% reduction).

When I throttle CPU 4x and emulate Fast 3G network conditions, the LCP-TTFB is reduced (median) from 1158.55ms to 1094.65ms (~6% reduction).

modules/js-and-css/optimize-oembeds/hooks.php
westonruter: Oh, the empty `src` is found because the PHP code has `$p->set_attribute( 'src', '' );` I think it can be updated to remove the attribute altogether?

westonruter: Just seeing this is included in my suggestion: https://github.com/WordPress/performance/pull/886#discussion_r1408292217

modules/js-and-css/optimize-oembeds/hooks.php
adamsilverstein: we are now using a different pattern where a new script tag is created and inserted. I will be sure to test this across recent browsers.

adamsilverstein: we are now using a different pattern where a new script tag is created and inserted. I will be sure to test this across recent browsers.

modules/js-and-css/optimize-oembeds/hooks.php
adamsilverstein: I adopted these suggestions in [`3c03658` (#886)](https://github.com/WordPress/performance/pull/886/commits/3c0365845f38cdb9f0482c18ccf62b7f013e196d) adding a docblock explaining what the intended behavior is.

adamsilverstein:
@westonruter 🎉 Thanks for testing! The performance measurements are especially helpful. 

I'd like to do some more manual testing with different browsers, then land this so we can get wider testing and fix edge cases that we are probably missing. When/if we propose for core we can add unit tests that validate the expected transforms of the embed content.

modules/js-and-css/optimize-oembeds/hooks.php
adamsilverstein: resolving as outdated, the new logic takes this into account.

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,130 @@
...
+	// If there was only one non-inline script, make it lazy.
+	if ( 1 === $script_count && ! $has_inline_script ) {
+		$oembed_lazy_load_scripts = true;
+		$p->seek( 'script' );
+		$p->set_attribute( 'data-lazy-embed-src', $p->get_attribute( 'src' ) );
+		$p->remove_attribute( 'src' );
+	}
+	// If there was only one iframe, make it lazy.
+	if ( 1 === $iframe_count ) {
+		$p->seek( 'frame' );
```
westonruter: Just noticed PHP complaining here:

> PHP Notice:  Function WP_HTML_Tag_Processor::seek was called incorrectly. Unknown bookmark name. Please see <a>Debugging in WordPress</a> for more information. (This message was added in version 6.2.0.)

Due to typo:

```suggestion
		$p->seek( 'iframe' );
```

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * @since n.e.x.t
+ * @package performance-lab
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+$oembed_lazy_load_scripts = false;
```
westonruter:
```suggestion
```

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,130 @@
...
+ * Lazy load any script tags.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $html The oEmbed HTML.
+ * @return string
+ */
+function perflab_optimize_oembed_html( $html ) {
+	global $oembed_lazy_load_scripts;
+
```
westonruter:
```suggestion
```

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,130 @@
...
+				$has_inline_script = true;
+			} else {
+				++$script_count;
+				$p->set_bookmark( 'script' );
+			}
+		}
+	}
+	// If there was only one non-inline script, make it lazy.
+	if ( 1 === $script_count && ! $has_inline_script ) {
+		$oembed_lazy_load_scripts = true;
```
westonruter:
```suggestion
		add_action( 'wp_footer', 'perflab_optimize_oembed_lazy_load_scripts' );
```

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,130 @@
...
+ * Load the embed's scripts when they approach the viewport using an IntersectionObserver.
+ *
+ * @since n.e.x.t
+ */
+function perflab_optimize_oembed_lazy_load_scripts() {
+	global $oembed_lazy_load_scripts;
+
+	if ( ! $oembed_lazy_load_scripts ) {
+		return;
+	}
```
westonruter:
```suggestion
```

westonruter (changes requested):
I think we can avoid the new global just be calling `add_action()` in the filter callback.

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,130 @@
...
+					if ( entry.isIntersecting ) {
+						const lazyEmbedParent = entry.target;
+						const lazyEmbedScript = lazyEmbedScriptsByParents.get( lazyEmbedParent );
+						const embedScript = document.createElement( 'script' );
+						for ( const attr of lazyEmbedScript.attributes ) {
+							if ( attr.nodeName === 'src' ) {
+								// Even though the src attribute is absent, the browser seems to presume it is present.
+								continue;
+							}
+
```
westonruter: This is no longer needed now that the `src` attribute is removed.

```suggestion
```

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,130 @@
...
+		for ( const lazyEmbedScript of lazyEmbedsScripts ) {
+			const lazyEmbedParent = lazyEmbedScript.parentNode;
+			lazyEmbedScriptsByParents.set( lazyEmbedParent, lazyEmbedScript );
+			lazyEmbedObserver.observe( lazyEmbedParent );
+		}
+	</script>
+	<?php
+}
+add_action( 'wp_footer', 'perflab_optimize_oembed_lazy_load_scripts', 99 );
```
westonruter: To add dynamically in the filter callback.

```suggestion
```

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,130 @@
...
+		$p->remove_attribute( 'src' );
+	}
+	// If there was only one iframe, make it lazy.
+	if ( 1 === $iframe_count ) {
+		$p->seek( 'frame' );
+		$p->set_attribute( 'loading', 'lazy' );
+	}
+	return $p->get_updated_html();
+}
+add_filter( 'embed_oembed_html', 'perflab_optimize_oembed_html', 10 );
```
westonruter: Since 10 is the default:

```suggestion
add_filter( 'embed_oembed_html', 'perflab_optimize_oembed_html' );
```

modules/js-and-css/optimize-oembeds/hooks.php
dmsnell: woo hoo! glad the warning came in handy :)

adamsilverstein:
Thanks for all the feedback and suggestions @westonruter and @mukeshpanchal27! I'm going to work on adding tests.

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,115 @@
...
+								attr.nodeValue
+							);
+						}
+						lazyEmbedScript.replaceWith( embedScript );
+						lazyEmbedObserver.unobserve( lazyEmbedParent );
+					}
+				}
+			},
+			{
+				rootMargin: '0px 0px 500px 0px',
```
westonruter: Curious how the `500px` value was determined? I wonder if it should rather be something like `1vh`?

I recall AMP has different values for this, depending on how heavy the component is. For example, `amp-youtube` uses [`0.75vh`](https://github.com/ampproject/amphtml/blob/ef8eac730183fba961e15b77d61b4ff7ccb36c6d/extensions/amp-youtube/0.1/amp-youtube.js#L129-L136).

[By default](https://github.com/ampproject/amphtml/blob/ef8eac730183fba961e15b77d61b4ff7ccb36c6d/src/base-element.js#L506-L518) AMP would render a lazy-loaded element when scrolling within three viewports (`3vh`) of the element, for example `amp-img`.

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,115 @@
...
+	$p = new WP_HTML_Tag_Processor( $html );
+
+	/**
+	 * Determine how to lazy load the embed.
+	 *
+	 * - If there is only one iframe, set loading="lazy".
+	 * - Prevent making scripts lazy if there is an inline script.
+	 *  - Only make script lazy if there is a single external script (since if there are
+	 *    multiple they may not get loaded in the right order).
+	 *  - Ensure that both the iframe and the script are made lazy if both occur in the same embed.
```
westonruter:
```suggestion
	 * - Only make script lazy if there is a single external script (since if there are
	 *   multiple they may not get loaded in the right order).
	 * - Ensure that both the iframe and the script are made lazy if both occur in the same embed.
```

modules/js-and-css/optimize-oembeds/hooks.php
adamsilverstein: Ah good question, this was largely experimental. I believe Chrome native lazy loading is even more conservative with a value over 1000px. It is also sensitive to low bandwidth, raising the threshold. 

I like the viewport suggestion, maybe `1vh`. Do you think we should add a filter here to make it adjustable?

westonruter: Also seeing another IntersectionObserver in AMP which [uses](https://github.com/ampproject/amphtml/blob/5df810730e3fc56f119e8ee5812bf2699ac7dde8/extensions/amp-list/1.0/component.js#L124) percentages here: 

```
bufferPct = Math.floor(viewportBuffer * 100)
...
rootMargin: `0% 0% ${bufferPct}% 0%`,
```

[Also](https://github.com/ampproject/amphtml/blob/5df810730e3fc56f119e8ee5812bf2699ac7dde8/extensions/amp-a4a/0.1/within-viewport.js#L61-L64):

```js
  const observer = new win.IntersectionObserver(callback, {
    root,
    rootMargin: `${(viewportNum - 1) * 100}%`,
  });
```

I guess `1vh` would do the same? It's not immediately clear if the percentage refers to the viewport or the element size.

There's also the two patterns for `rootMargin`, of `0 0 X 0` and just `X` (i.e. `X X X X`). The latter would make sense when scrolling toward an embed in any direction other than from top to bottom, such as when a page is loaded with an anchor link I think.

As for a filter, maybe? But maybe later if there is a clear need.

westonruter: How about this:

```suggestion
				rootMargin: '100% 0% 100% 0%',
```

(I previously used `1vh` when I meant `100vh` as the unit is divided into 100.)

This will cause the element to be initialized when scrolling the element into view from the top _or_ bottom when within 1 viewport.

adamsilverstein: thanks for researching, accepting your suggestion for now - we can always fine tune later

swissspidy (approved):

modules/js-and-css/optimize-oembeds/hooks.php
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * Add loading="lazy" to any iframe tags.
+ * Lazy load any script tags.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $html The oEmbed HTML.
+ * @return string
+ */
+function perflab_optimize_oembed_html( $html ) {
```
westonruter: Nit

```suggestion
function perflab_optimize_oembed_html( string $html ): string {
```

westonruter:
Two things:

1. Shouldn't tests be added prior to merging?
2. No new modules have been merged in a while as they're waiting for migration to standalone plugins. I don't think we're supposed to be adding new modules.

Perhaps what should be done is to create a new `feature/lazy-embeds` branch off of `trunk` and then re-base this PR to target it. Then it can be merged prior to tests and we can also wait until we have the module-to-plugins plumbing finished.

adamsilverstein:
> Perhaps what should be done is to create a new feature/lazy-embeds branch off of trunk and then re-base this PR to target it. Then it can be merged prior to tests and we can also wait until we have the module-to-plugins plumbing finished.

That sounds good to me!

westonruter:
I created the `feature/lazy-embeds` branch but I can't change the base branch in your PR since it is on a fork. You'll have to do that. Once you've done it we can merge and then proceed with tests.

adamsilverstein:
>  I created the feature/lazy-embeds branch but I can't change the base branch in your PR since it is on a fork. You'll have to do that. Once you've done it we can merge and then proceed with tests.

Thanks @westonruter - changed base branch

westonruter (approved):

PR #890, opened 2023-11-30
Add support for plugin live preview in the plugin directory

### LINKED ISSUE #885: Support Plugin Previews in the .org Plugin Directory
## Feature Description

The WP.org Meta team just announced the relaunching of the plugin preview feature, whereby people can try out the plugin in a sandbox using WP Playground before installing. This is pretty cool, but requires plugin devs to opt-in. Here's the post with the details:

https://make.wordpress.org/meta/2023/11/22/plugin-directory-preview-button-revisited/

We should consider adding opt in to Performance Lab and for the standalone plugins.

mukeshpanchal27:
## Summary

Fixes #885

The PR enable live preview support for Performance lab plugin. I can't added the support for standalone plugin because they don't have any settings page.

Note: When we implemented site health related checks in modules then they it will be automatically disable in Playground environment.

## Checklist

### DIFF
--- a/.wordpress-org/blueprints/blueprint.json
+++ b/.wordpress-org/blueprints/blueprint.json
@@ -1,0 +1,24 @@
+{
+  "$schema": "https://playground.wordpress.net/blueprint-schema.json",
+  "landingPage": "\/wp-admin\/admin.php?page=perflab-modules",
+  "phpExtensionBundles": [
+    "kitchen-sink"
+  ],
+  "steps": [
+    {
+      "step": "login",
+      "username": "admin",
+      "password": "password"
+    },
+    {
+      "step": "installPlugin",
+      "pluginZipFile": {
+        "resource": "wordpress.org\/plugins",
+        "slug": "performance-lab"
+      },
+      "options": {
+        "activate": true
+      }
+    }
+  ]
+}


mukeshpanchal27:
I've added the `Infrastructure` label, so the Check Milestone workflow to pass successfully.

.wordpress-org/blueprints/blueprint.json
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "https://playground.wordpress.net/blueprint-schema.json",
+  "landingPage": "\/wp-admin\/admin.php?page=perflab-modules",
```
felixarntz, line 3: The example on https://make.wordpress.org/meta/2023/11/22/plugin-directory-preview-button-revisited/ includes a `preferredVersions` entry. Not sure we need that here, but my main questions are: Which WordPress / PHP version does Playground use by default?

Since we require WP 6.3+ and PHP 7.0+, I just want to make sure those requirements are fulfilled by the preview. If we need the `preferredVersions` entry to ensure that, let's add it, otherwise we can leave it out.

felixarntz (commented):
Thanks @mukeshpanchal27, this looks solid.

I only have one question - not sure whether this is required or not, so let's clarify that first.

.wordpress-org/blueprints/blueprint.json
swissspidy: It uses the latest versions of both, definitely not something outdated :-)

felixarntz (approved):

swissspidy (approved):

PR #892, opened 2023-12-04
Rely on server-side breadcrumbs for detection and optimization

westonruter:
## Summary

See https://github.com/WordPress/performance/issues/869.

Eliminate client-side computation of element breadcrumbs in favor of generating them server-side while processing the output buffer. Breadcrumbs are serialized as XPath and added to image elements as a `data-ilo-xpath` attribute. This data attribute is used by the detection script instead of generating breadcrumbs client-side. Lastly, the detection script is now injected as part of the output buffer processing instead of a separate `wp_footer` code path.

## Relevant technical choices

This improves the reliability of connecting elements from client-side detection with server-side optimization by exclusively computing breadcrumbs XPath on the server. This eliminates the issue encountered in #884 in which JavaScript injecting a new element. Namely it address the following todo from https://github.com/WordPress/performance/pull/884#issuecomment-1817577551:

> Improve handling of breadcrumb generation when there the document structure varies either due to admin bar being present or due to client-side logic. Currently breadcrumbs have to explicitly skip over the admin bar element to ignore it. Similarly, the client-side breadcrumb logic has to skip over `.skip-link.screen-reader-text` which is injected by `wp_enqueue_block_template_skip_link()`. This issue of client-side mutated elements is particularly problematic, which may require server-side tagging of elements with their breadcrumbs which would eliminate needing to also do that calculation on the client. But this would mean extraneous `data-` attributes on all images. When detection isn't needed, the tag processor could strip those attributes out before sending back the buffer. We may need to monitor this to see how much of a problem it is, such as by checking various themes to see what degree they inject elements to mess up breadcrumbs. As another alternative, we may consider relying on IDs and class names for breadcrumbs rather than element indices, although JS here too can cause problems by modifying the class names.

As part of this, XPath strings are used throughout the module rather than comparing breadcrumbs arrays.

Additionally, instead of injecting the detection script at `wp_footer` and then having some of the same conditions during output buffer processing for optimization, the detection script is now injected during optimization.

## Checklist

### DIFF
--- a/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
+++ b/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
@@ -226,19 +226,14 @@ final class ILO_HTML_Tag_Processor {
 	/**
 	 * Gets breadcrumbs for the current open tag.
 	 *
-	 * Breadcrumbs are constructed to match the format from detect.js.
+	 * A breadcrumb consists of a tag name and its sibling index.
 	 *
-	 * TODO: Consider rather each breadcrumb being a (tag,index) tuple.
-	 *
-	 * @return array<array{tag: string, index: int}> Breadcrumbs.
+	 * @return array<array{string, int}> Breadcrumbs.
 	 */
 	public function get_breadcrumbs(): array {
 		$breadcrumbs = array();
 		foreach ( $this->open_stack_tags as $i => $breadcrumb_tag_name ) {
-			$breadcrumbs[] = array(
-				'tag'   => $breadcrumb_tag_name,
-				'index' => $this->open_stack_indices[ $i ],
-			);
+			$breadcrumbs[] = array( $breadcrumb_tag_name, $this->open_stack_indices[ $i ] );
 		}
 		return $breadcrumbs;
 	}
--- a/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
+++ b/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
@@ -239,6 +239,9 @@ final class ILO_HTML_Tag_Processor {
 	/**
 	 * Gets XPath for the current open tag.
 	 *
+	 * It would be nicer if this were like `/html[1]/body[2]` but in XPath the position() here refers to the
+	 * index of the preceding node set. So it has to rather be written `/*[1][self::html]/*[2][self::body]`.
+	 *
 	 * @return string XPath.
 	 */
 	public function get_xpath(): string {
--- a/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
+++ b/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
@@ -89,6 +89,14 @@ final class ILO_HTML_Tag_Processor {
 	);
 
 	/**
+	 * Pattern for valid XPath subset for breadcrumb.
+	 *
+	 * @see self::get_xpath()
+	 * @var string
+	 */
+	const XPATH_PATTERN = '^(/\*\[\d+\]\[self::.+?\])+$';
+
+	/**
 	 * Open stack tags.
 	 *
 	 * @var string[]
--- a/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
+++ b/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
@@ -7,7 +7,7 @@
  */
 
 /**
- * Processor leveraging WP_HTML_Tag_Processor which gathers breadcrumbs which can be queried while iterating the open_tags() generator.
+ * Processor leveraging WP_HTML_Tag_Processor which gathers breadcrumbs which can be obtained as XPath while iterating the open_tags() generator.
  *
  * Eventually this class should be made largely obsolete once `WP_HTML_Processor` is fully implemented to support all HTML tags.
  *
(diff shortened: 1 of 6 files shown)


felixarntz:
@westonruter Would you like this PR to be reviewed and merged into the other PR's branch, or did you simply start work on this before and prefer #884 to be merged first?

westonruter:
@felixarntz I'd like #884 to be merged first so that the PRs don't get excessively large.

westonruter:
Note: I'm working on the tests for this in a sub-PR: #898. 

In the meantime, this PR can be reviewed (and merged).

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -226,21 +226,30 @@ public function open_tags(): Generator {
...
-			);
+			yield array( $breadcrumb_tag_name, $this->open_stack_indices[ $i ] );
+		}
+	}
+
+	/**
+	 * Gets XPath for the current open tag.
+	 *
+	 * It would be nicer if this were like `/html[1]/body[2]` but in XPath the position() here refers to the
+	 * index of the preceding node set. So it has to rather be written `/*[1][self::html]/*[2][self::body]`.
```
westonruter, line 251: This being said, the breadcrumbs could be updated to keep track of the index of the tag of a given type rather than the index of the tag in general. This would make breadcrumbs more in line with the natural XPath syntax, so instead `/*[1][self::html]/*[2][self::body]` we could do just `/html[1]/body[1]`.

Nevertheless, what is implemented now is what the `:nth-child()` pseudo-selector implements in CSS, as opposed to `:nth-of-type()`.

But doing this would take some extra bookkeeping, and I don't see a clear benefit.

I also just came across a Chromium doc that is looking at the same problem space: [Identifying element consistently across reloads](https://docs.google.com/document/d/18QpJB1q7eSgzG7n3SQPC61kn5UJGaS1ZTG0Y1eaCq5w/edit).

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -226,21 +226,30 @@ public function open_tags(): Generator {
...
 	 */
-	public function get_breadcrumbs(): array {
-		$breadcrumbs = array();
+	private function get_breadcrumbs(): Generator {
 		foreach ( $this->open_stack_tags as $i => $breadcrumb_tag_name ) {
-			$breadcrumbs[] = array(
-				'tag'   => $breadcrumb_tag_name,
-				'index' => $this->open_stack_indices[ $i ],
-			);
+			yield array( $breadcrumb_tag_name, $this->open_stack_indices[ $i ] );
```
felixarntz, line 243: Any particular reason for the data type change here? I find an associative array with named keys easier to understand than an indexed array that isn't really for a list of something.

modules/images/image-loading-optimization/optimization.php
```diff
@@ -114,24 +96,29 @@ function ilo_optimize_template_output_buffer( string $buffer ): string {
 	$slug = ilo_get_url_metrics_slug( ilo_get_normalized_query_vars() );
 	$post = ilo_get_url_metrics_post( $slug );
 
-	// No URL metrics are present, so there's nothing we can do.
-	if ( ! $post ) {
-		return $buffer;
-	}
+	$url_metrics = $post ? ilo_parse_stored_url_metrics( $post ) : array();
```
felixarntz, line 99: Why remove the early return here? Isn't that more efficient if no URL metrics post is found?

modules/images/image-loading-optimization/storage/rest-api.php
```diff
@@ -130,24 +130,10 @@ function ilo_register_endpoint() {
...
-											'pattern'  => '^[a-zA-Z0-9-]+$',
-										),
-										'index' => array(
-											'type'     => 'int',
-											'required' => true,
-											'minimum'  => 0,
-										),
-									),
-								),
+								'pattern'  => '^(/\*\[\d+\]\[self::.+?\])+$', // See ILO_HTML_Tag_Processor::get_xpath() for format.
```
felixarntz: Just a suggestion: Make this a `const` on `ILO_HTML_Tag_Processor` since that's where the definition really comes from.

felixarntz (approved):
This looks great to me! It's more reliable now that the source of truth is on the server - and while I don't like XPath, at least its usage here is simple enough (mostly just to compare against it) that I feel it doesn't make the code unapproachable. Also +1 to combining the logic for when to print the detection script to avoid duplicate logic.

Just a few minor points below.

modules/images/image-loading-optimization/optimization.php
westonruter: Good question. Because now if no URL metrics are found, we need to proceed and add breadcrumbs to the elements so that we can gather the URL metrics. You can see below too that the detection script is now injected in this function as well.

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
westonruter: Mainly because it makes it more compact and slightly easier to iterate over per below:

```
foreach ( $this->get_breadcrumbs() as list( $tag_name, $index ) ) {
```

Originally too I had envisioned that a breadcrumbs could have other keys, like a class name or maybe some attributes to make it more like a CSS selector. But this didn't end up being the case, as we only ever need the tag name and index. So making it a tuple seems to make sense to me.

felixarntz: I realize it's a few fewer lines of code, but it's harder to understand. Keys allow clarifying what something is, while now that's less apparent, particularly when looking at where the function is used and only finding `list`.

I'd prefer to go for clarity over compactness, but not a blocker as long as this function remains purely internal.

modules/images/image-loading-optimization/storage/rest-api.php
westonruter: Good idea. Done in 2df915d8

adamsilverstein (approved):
Splendid

PR #896, opened 2023-12-05
Fix Plugin card UI and thickbox issue

mukeshpanchal27:
## Summary

This PR fixed the plugin card UI issue and the auto resize thickbox popup issue.
- The plugin card is not equal height as per the install plugin screen.

**Before**
Screenshot 2023-12-05 at 11 09 42 AM

**After**
Screenshot 2023-12-05 at 11 10 30 AM

- The popup is not auto resize when the screen is big or small.

**Before**
Screenshot 2023-12-05 at 11 10 03 AM

**After**
Screenshot 2023-12-05 at 11 10 51 AM

## Checklist

### DIFF
--- a/admin/plugins.php
+++ b/admin/plugins.php
@@ -15,7 +15,6 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @since n.e.x.t
  *
  * @param string $plugin_slug The string identifier for the plugin in questions slug.
- *
  * @return array Array of plugin data, or empty if none/error.
  */
 function perflab_query_plugin_info( string $plugin_slug ) {
@@ -46,7 +45,7 @@ function perflab_query_plugin_info( string $plugin_slug ) {
  *
  * @since n.e.x.t
  *
- * @return array of wpp standalone plugins as slugs.
+ * @return array List of WPP standalone plugins as slugs.
  */
 function perflab_get_standalone_plugins() {
 	return array(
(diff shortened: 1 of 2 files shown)


admin/plugins.php
```diff
@@ -76,7 +75,7 @@ function perflab_render_plugins_ui() {
 		return;
 	}
 	?>
-	<div class="wrap">
+	<div class="wrap plugin-install-php">
```
felixarntz, line 78: @mukeshpanchal27 What is the purpose of this class?

swissspidy: The `.plugin-install-php` selector is targeted by some of the WP admin styles.

felixarntz: But is that the right class to use here? I don't see any application of the class in core when searching the codebase, which makes me wonder if this is the legacy class. Maybe I'm missing something. But if core doesn't use it, then why should we here?

swissspidy: Yes. This class name _is_ used by core. This is needed here so that all the plugin cards have the equal height. This is because the core CSS targets `.plugin-install-php .plugin-card` to make the cards use flexbox

felixarntz: I understand what the CSS and JS is doing. But where is it used in core? I only found the class name in CSS and JS but nowhere actually added in the HTML markup.

swissspidy: Go to `wp-admin/plugin-install.php` and you'll see it on the `<body>` element. It's a dynamic class based on the file name (`plugin-install.php` -> `plugin-install-php`)

felixarntz: Ah I see, thanks for clarifying!

felixarntz (approved):

swissspidy (approved):

admin/plugins.php
mukeshpanchal27: Thank you, @swissspidy, for responding to Felix's questions.

PR #898, opened 2023-12-06
Add PHPUnit tests for Image Loading Optimization

westonruter:
## Summary

This is part of https://github.com/WordPress/performance/issues/869.

This adds full test coverage for the functions in the Image Loading Optimization module.

## Relevant technical choices

* The `ILO_HTML_Tag_Processor::open_tags()` generator is updated to ensure accurate breadcrumbs/XPath for SVG and MathML elements.
* The phpdoc and variable names are improved in `ilo_construct_preload_links()`.
* The `integrity` attribute is no longer copied from an `IMG` element to the preload `LINK`. This attribute doesn't appear on `IMG`, but rather is only on `LINK` and `SCRIPT`.
* Rework how timestamps are added to URL metrics by adding and examining them in `ilo_unshift_url_metrics()`. Previously the timestamp was added in `ilo_store_url_metric()` before passing to `ilo_construct_preload_links()`.
* Refactor `ilo_unshift_url_metrics()` to make it easier to test by passing more arguments.
* Fix JSON Schema definition in `ilo_register_endpoint()`.
* ~Discontinue excluding `WordPress-Extra` sniffs from applying to tests, and add specific sniff exclusions. By allowing `WordPress-Extra` to apply, PHPCBF can do automatic formatting. PHPCS errors were fixed in other test files.~ See #924.
* Fix REST API endpoint to reject storing a URL metric when the sample is already fully populated for the provided viewport. Previously it would continue to accept requests when _any_ of the breakpoints lacked a full sample size of URL metrics.
* Remove debug code from being returned in REST API response.

## Checklist


modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -183,7 +183,11 @@ public function open_tags(): Generator {
...
 
 				// Immediately pop off self-closing tags.
-				if ( in_array( $tag_name, self::VOID_TAGS, true ) ) {
+				if (
+					in_array( $tag_name, self::VOID_TAGS, true )
+					||
+					( $p->has_self_closing_flag() && $this->is_foreign_element() )
+				) {
 					array_pop( $this->open_stack_tags );
 				}
```
westonruter, line 202: @dmsnell: FYI, I'm pretty happy with how `has_self_closing_flag` and the foreign element aspect of the HTML spec simplified things! Now it can properly get breadcrumbs/XPaths for each SVG/MathML element as well.

modules/images/image-loading-optimization/optimization.php
```diff
@@ -30,26 +30,27 @@ function ilo_maybe_add_template_output_buffer_filter() {
...
-function ilo_construct_preload_links( array $lcp_images_by_minimum_viewport_widths ): string {
+function ilo_construct_preload_links( array $lcp_elements_by_minimum_viewport_widths ): string {
 	$preload_links = array();
 
 	// This uses a for loop to be able to access the following element within the iteration, using a numeric index.
-	$minimum_viewport_widths = array_keys( $lcp_images_by_minimum_viewport_widths );
+	$minimum_viewport_widths = array_keys( $lcp_elements_by_minimum_viewport_widths );
 	for ( $i = 0, $len = count( $minimum_viewport_widths ); $i < $len; $i++ ) {
-		$lcp_element = $lcp_images_by_minimum_viewport_widths[ $minimum_viewport_widths[ $i ] ];
-		if ( false === $lcp_element || empty( $lcp_element['attributes'] ) ) {
```
westonruter, line 43: The `empty( $lcp_element['attributes'] )` check is removed here because it is always set.

modules/images/image-loading-optimization/storage/rest-api.php
```diff
@@ -171,6 +172,8 @@ function ilo_handle_rest_request( WP_REST_Request $request ) {
 		ilo_get_url_metrics_breakpoint_sample_size(),
 		ilo_get_url_metric_freshness_ttl()
 	);
+
+	// TODO: This is not right. It is asking if it is needed for any breakpoint, not if it is needed for the supplied breakpoint. This logic here is specific for the frontend.
```
westonruter: I should address this.

composer.json
```diff
@@ -28,6 +28,7 @@
   "require": {
     "composer/installers": "~1.0",
     "php": ">=7|^8",
+    "ext-dom": "*",
```
westonruter, line 31: Added because `DOMDocument` is used in tests now.

modules/images/image-loading-optimization/optimization.php
```diff
@@ -169,10 +170,13 @@ function ilo_optimize_template_output_buffer( string $buffer ): string {
 			$processor->remove_attribute( 'fetchpriority' );
 		}
 
+		// TODO: If the image is visible (intersectionRatio!=0) in any of the URL metrics, remove loading=lazy.
+		// TODO: Conversely, if an image is the LCP element for one breakpoint but not another, add loading=lazy. This won't hurt performance since the image is being preloaded.
+
 		// Capture the attributes from the LCP elements to use in preload links.
 		if ( isset( $lcp_element_minimum_viewport_width_by_xpath[ $xpath ] ) ) {
 			$attributes = array();
-			foreach ( array( 'src', 'srcset', 'sizes', 'crossorigin', 'integrity' ) as $attr_name ) {
```
westonruter, line 175: The `integrity` attribute doesn't exist on `IMG` so it shouldn't be copied.

tests/modules/images/image-loading-optimization/hooks-tests.php
```diff
@@ -1,14 +1,12 @@
...
+ * Tests for image-loading-optimization module hooks.php.
  *
  * @package performance-lab
  * @group   image-loading-optimization
  */
 
-use PerformanceLab\Tests\TestCase\ImagesTestCase;
-
-class Image_Loading_Optimization_Load_Tests extends ImagesTestCase {
+class Image_Loading_Optimization_Hooks_Tests extends WP_UnitTestCase {
```
westonruter:
```suggestion
class ILO_Hooks_Tests extends WP_UnitTestCase {
```

tests/modules/images/image-loading-optimization/detection-tests.php
```diff
@@ -0,0 +1,74 @@
+<?php
+/**
+ * Tests for image-loading-optimization module detection.php.
+ *
+ * @package performance-lab
+ * @group   image-loading-optimization
+ */
+
+class Image_Loading_Optimization_Detection_Tests extends WP_UnitTestCase {
```
westonruter:
```suggestion
class ILO_Detection_Tests extends WP_UnitTestCase {
```

tests/modules/images/image-loading-optimization/class-ilo-html-tag-processor-tests.php
```diff
@@ -0,0 +1,335 @@
+<?php
+/**
+ * Tests for image-loading-optimization class ILO_HTML_Tag_Processor.
+ *
+ * @package performance-lab
+ * @group   image-loading-optimization
+ *
+ * @coversDefaultClass ILO_HTML_Tag_Processor
+ */
+class Image_Loading_Optimization_ILO_HTML_Tag_Processor_Tests extends WP_UnitTestCase {
```
westonruter:
```suggestion
class ILO_HTML_Tag_Processor_Tests extends WP_UnitTestCase {
```

tests/modules/images/image-loading-optimization/optimization-tests.php
```diff
@@ -0,0 +1,467 @@
+<?php
+/**
+ * Tests for image-loading-optimization module optimization.php.
+ *
+ * @package performance-lab
+ * @group   image-loading-optimization
+ */
+
+class Image_Loading_Optimization_Optimization_Tests extends WP_UnitTestCase {
```
westonruter:
```suggestion
class ILO_Optimization_Tests extends WP_UnitTestCase {
```

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -244,6 +237,20 @@ private function get_breadcrumbs(): Generator {
 		}
 	}
 
+	/**
+	 * Determines whether currently inside a foreign element (MATH or SVG).
+	 *
```
mukeshpanchal27, line 246: This file missing the `@since n.e.x.t` annotation for functions.

🔢 Apply to other functions also.

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -244,6 +237,20 @@ private function get_breadcrumbs(): Generator {
...
+	 * @return bool In foreign element.
+	 */
+	private function is_foreign_element(): bool {
+		foreach ( $this->open_stack_tags as $open_stack_tag ) {
```
mukeshpanchal27, line 252: Could we use `in_array` as it slightly more efficient as it could potentially exit earlier upon finding the first occurrence, while `foreach` iterates through the entire array.

```
if ( in_array( 'MATH', $this->open_stack_tags, true ) || in_array( 'SVG', $this->open_stack_tags, true ) ) {
	return true;
}
```

tests/modules/images/image-loading-optimization/class-ilo-html-tag-processor-tests.php
```diff
@@ -0,0 +1,335 @@
...
+					'/*[0][self::HTML]/*[1][self::BODY]/*[0][self::UL]',
+				),
+			),
+		);
+	}
+
+	/**
+	 * Test open_tags() and get_xpath().
+	 *
+	 * @test
```
mukeshpanchal27:
```suggestion
```
For test functions to execute correctly, we should maintain consistency by using either the `test_` prefix in function names or the `@test` annotation. In the mentioned PR, both methods were employed, so I suggest removing the `@test` annotation as it is unnecessary. This adjustment should also be applied to other tests for consistency. Thank you!

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
dmsnell: @mukeshpanchal27 if you look closely you'll notice that this for loop bails as soon as it finds either a MATH or an SVG element. in contrast, with the dual `in_array()` checks we'd be iterating twice over the stack if neither tag is present whereas with the code as-written, we only iterate once.

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
westonruter: Done in df0d09ea9872f28cf135e10277f4db14cba93501

tests/modules/images/image-loading-optimization/class-ilo-html-tag-processor-tests.php
westonruter: Done in 1fa877c7

modules/images/image-loading-optimization/optimization.php
```diff
@@ -30,26 +30,27 @@ function ilo_maybe_add_template_output_buffer_filter() {
  * @since n.e.x.t
  * @access private
  *
- * @param array $lcp_images_by_minimum_viewport_widths LCP images keyed by minimum viewport width, amended with attributes key for the IMG attributes.
+ * @param array<int, array{attributes: array{src?: string, srcset?: string, sizes?: string, crossorigin?: string}}|false> $lcp_elements_by_minimum_viewport_widths LCP images keyed by minimum viewport width, amended with attributes key for the IMG attributes.
```
felixarntz, line 65: Maybe a personal preference, but I find this notation overkill. This isn't used anywhere in WordPress. If we want to describe more complex shapes like this, we should probably introduce a class for them. This would be safer in any case than using an associative array which by definition takes any arbitrary data.

I think a simple `ILO_LCP_Element` class or similar would be helpful here. Then this would become `ILO_LCP_Element[]`.

Separate minor question: Why was this renamed to "LCP elements" instead of "LCP images"? Would it ever _not_ be an LCP image? If so, the choice of fields would be a bit odd, since almost all of them are specific to images.

modules/images/image-loading-optimization/optimization.php
felixarntz: Another benefit of using a class to represent LCP elements would be that that would be truly enforced :)

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -150,33 +150,40 @@ function ilo_verify_url_metrics_storage_nonce( string $nonce, string $slug ): in
  * @since n.e.x.t
  * @access private
  *
- * @param array $url_metrics          URL metrics.
+ * @param array $url_metrics          URL metrics. Each URL metric is expected to have a timestamp key.
```
felixarntz, line 119: Not a blocker for this PR, but I think it would be useful to follow up once this is merged with another PR to introduce a class to represent a URL metric (e.g. `ILO_URL_Metric`). It would make the code much more intuitive and easier to follow as the fields would be enforced by the class definition.

modules/images/image-loading-optimization/storage/post-type.php
```diff
@@ -110,7 +110,40 @@ function ilo_parse_stored_url_metrics( WP_Post $post ): array {
...
+							__( 'Unexpected shape to JSON array in post_content of %s post type.', 'performance-lab' ),
+							ILO_URL_METRICS_POST_TYPE
+						)
+					);
+				}
+
+				return $is_valid;
+			}
+		)
+	);
```
felixarntz, line 146: This makes another good case why there should be a class to represent these objects. Validation would belong there, and probably be fewer lines of code as some of those checks would become irrelevant when enforcing proper fields via class properties.

modules/images/image-loading-optimization/storage/rest-api.php
```diff
@@ -171,16 +172,28 @@ function ilo_handle_rest_request( WP_REST_Request $request ) {
...
+	$viewport_width  = $request->get_param( 'viewport' )['width'];
+	$last_was_needed = false;
+	foreach ( $needed_minimum_viewport_widths as list( $minimum_viewport_width, $is_needed ) ) {
+		if ( $viewport_width >= $minimum_viewport_width ) {
+			$last_was_needed = $is_needed;
+		} else {
+			break;
+		}
+	}
+	if ( ! $last_was_needed ) {
```
felixarntz, line 187: Is the `ilo_needs_url_metric_for_breakpoint()` still valuable given it's no longer used here? Would it be possible to find a way to enhance the function in a way that it can be used here, rather than more inline logic in this function?

felixarntz (changes requested):
This PR encompasses _a lot_ of changes, far more than only test coverage, which makes it difficult to review.

For now, I left a few comments on the production code changes. They mostly look good to me, although I think the code of some of those functions is extremely complex and could be simplified by separating responsibilities, and using smaller functions, or classes where applicable, for distinct processes. This is by no means a blocker for this PR, but I think it is critical to make this module/plugin long-term maintainable and more accessible to other contributors.

I think the PHPCS changes that affect the entire plugin make sense too, though I think that would be great to decouple as well, so that it can already be merged. FWIW, this probably shouldn't happen in the feature branch at all, as it likely will lead to more and more merge conflicts, the longer it takes for this feature branch to be merged. Could you open a PR with the PHPCS adjustments that don't affect the module directly against `trunk`?

For the production code changes and tests, I don't feel strongly that they need to be separated, though I believe that it could help getting the production code changes merged a lot faster. Reviewing the tests takes notably more time.

modules/images/image-loading-optimization/optimization.php
westonruter: > Maybe a personal preference, but I find this notation overkill. This isn't used anywhere in WordPress. If we want to describe more complex shapes like this, we should probably introduce a class for them. This would be safer in any case than using an associative array which by definition takes any arbitrary data.

Actually, it is used in WordPress, specifically [in the `sodium_compat` code](https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop%20%2Farray%7B%2F&type=code). Granted, this is not really organic WordPress code. In any case, with precise phpdoc typing here, PHPStan will detect any anomalies with static analysis. So it's trying to gain the same benefits as TypeScript gives.

> I think a simple `ILO_LCP_Element` class or similar would be helpful here. Then this would become `ILO_LCP_Element[]`.

Yeah, I like that idea. However, believe it or not, I was wanting to limit the refactor of the existing code in this PR once the test coverage was completed. With the coverage in place, refactoring will be safer as I'll be able to detect regressions better.

> Separate minor question: Why was this renamed to "LCP elements" instead of "LCP images"? Would it ever not be an LCP image? If so, the choice of fields would be a bit odd, since almost all of them are specific to images.

Yes, for example in https://github.com/WordPress/performance/pull/914 this is extended to support elements with CSS background images.

modules/images/image-loading-optimization/storage/post-type.php
westonruter: Nevertheless, the data would be stored as JSON in the database and not as a serialized class. So this validation would still have to be done somewhere.

modules/images/image-loading-optimization/optimization.php
felixarntz: > Yeah, I like that idea. However, believe it or not, I was wanting to limit the refactor of the existing code in this PR once the test coverage was completed. With the coverage in place, refactoring will be safer as I'll be able to detect regressions better.

+1, that makes total sense to me. Maybe open an issue already?

modules/images/image-loading-optimization/storage/post-type.php
felixarntz: Right, I think it could then be done in the class. I would expect a class to e.g. receive this associative array as constructor parameter and then validate it.

modules/images/image-loading-optimization/storage/rest-api.php
westonruter: Good point! Yeah, it's not valuable. I've removed it in 3da1363c. I did try to consolidate this logic into that existing function, but the two codepaths didn't mesh well.

modules/images/image-loading-optimization/optimization.php
```diff
@@ -107,7 +108,12 @@ function ilo_optimize_template_output_buffer( string $buffer ): string {
 	);
 
 	// Whether we need to add the data-ilo-xpath attribute to elements and whether the detection script should be injected.
-	$needs_detection = ilo_needs_url_metric_for_breakpoint( $needed_minimum_viewport_widths );
+	$needs_detection = in_array(
+		true,
+		// Each array item is array{int, bool}, with the second item being whether the viewport width is needed.
+		array_column( $needed_minimum_viewport_widths, 1 ),
+		true
+	);
```
westonruter, line 148: Some cool things with PHPStan's static analysis here. If 0 is supplied as the second arg to `array_column()` it detects an error:

> phpstan: Call to function in_array() with arguments true, list<int> and true will always evaluate to false.

This is because from the array shape returned by `ilo_get_needed_minimum_viewport_widths()` it knows that the first column will all be integers (the minimum viewport widths) and thus the `in_array()` check will always be `false`. 

Conversely, if I supply a higher index like 2 in `array_column()`, I get:

> phpstan: Call to function in_array() with arguments true, array{} and true will always evaluate to false.

Here PHPStan knows that since there are only two items in each tuple, the return of `array_column()` will be an empty array, and thus `in_array()` will also always return `false`.

westonruter:
@felixarntz 

> For now, I left a few comments on the production code changes. They mostly look good to me, although I think the code of some of those functions is extremely complex and could be simplified by separating responsibilities, and using smaller functions, or classes where applicable, for distinct processes. This is by no means a blocker for this PR, but I think it is critical to make this module/plugin long-term maintainable and more accessible to other contributors. [...] For the production code changes and tests, I don't feel strongly that they need to be separated, though I believe that it could help getting the production code changes merged a lot faster. Reviewing the tests takes notably more time.

Agreed. As mentioned above, I wanted to limit further refactoring until tests were written. The code changes not related to test coverage were primarily to fix blatant issues I uncovered while writing the tests. Sorry it grew so large. Since the review cycles have been pretty long, it's hard to keep the PRs very concise.

> I think the PHPCS changes that affect the entire plugin make sense too, though I think that would be great to decouple as well, so that it can already be merged. FWIW, this probably shouldn't happen in the feature branch at all, as it likely will lead to more and more merge conflicts, the longer it takes for this feature branch to be merged. Could you open a PR with the PHPCS adjustments that don't affect the module directly against `trunk`?

Agreed. Done in https://github.com/WordPress/performance/pull/924. I've reverted the changes in this PR via 2181d884305abce6a95182fc8b037beed747bd9b.

westonruter:
Unit tests are failing due to `Audit_Enqueued_Assets_Tests`, which isn't related to the changes in this PR. ~Two of the failures are fixed in `trunk` and~ the other is fixed in #926 (which for some reason doesn't happen in `trunk`).

westonruter:
The PHPStan failures are likewise not related to code changes here. They may be resolved by merging `trunk` into the feature branch, but I will wait to do that until #926 is merged.

modules/images/image-loading-optimization/optimization.php
westonruter: I've just now added it as a todo item on the [overview issue](https://github.com/WordPress/performance/issues/869).

tests/modules/images/image-loading-optimization/storage/data-tests.php
```diff
@@ -0,0 +1,618 @@
...
+		);
+	}
+
+	/**
+	 * Test ilo_can_optimize_response().
+	 *
+	 * @covers ::ilo_can_optimize_response
+	 * @dataProvider data_provider_test_ilo_can_optimize_response
+	 */
+	public function test_ilo_can_optimize_response( Closure $set_up, bool $expected ) {
```
felixarntz: I was just wondering why this function is in the `storage/data.php` file. It's only called in `optimization.php`, and the function name to me suggests it would fit better in there.

I then noticed `storage/data.php` contains several other functions like this, that to me don't seem storage related. Is that intentional? If so, can you clarify how you see the responsibilities between these two files?

Maybe there's a more intuitive way to break apart the functions here. To be fair, some of it may sort of automatically be addressed when using classes for some of those things as applicable.

tests/modules/images/image-loading-optimization/storage/data-tests.php
```diff
@@ -0,0 +1,618 @@
...
+			},
+			10,
+			2
+		);
+
+		// Create first nonce for unauthenticated user.
+		$slug   = ilo_get_url_metrics_slug( array() );
+		$nonce1 = ilo_get_url_metrics_storage_nonce( $slug );
+		$this->assertMatchesRegularExpression( '/^[0-9a-f]{10}$/', $nonce1 );
+		$this->assertSame( 1, ilo_verify_url_metrics_storage_nonce( $nonce1, $slug ) );
```
felixarntz: Nit-pick: I find the return of 0, 1, or 2 an odd choice. I realize `wp_verify_nonce()` returns that, but it's extremely unintuitive, and I don't think many people know, except from reading the docs.

There's no value for us (and probably any other use-cases) to make a difference between when the nonce was generated, so I'd prefer to just return a boolean (i.e. cast the result to a bool). You aren't testing the `2` return value here either as far as I can tell (which makes sense IMO!), so that confirms it's useless to us :)

felixarntz (approved):
Thanks @westonruter, LGTM! I think this is good to merge once the PHPStan problems have been addressed (here or via the other PR).

Let's separately follow up on simplifying some of the logic with classes, and also file structure (see my comment below).

tests/modules/images/image-loading-optimization/storage/data-tests.php
westonruter: Yeah, sounds good. Done in d865bacb.

tests/modules/images/image-loading-optimization/storage/data-tests.php
westonruter: Good catch! Function moved in ed305a9efeba2411f37cd1eaf69f7258d2828fb6.

The other functions are generally storage related in that they relate to how fresh the stored data is, how to determine the slug for the data, manipulating the stored data, etc. But I agree this the functions in this file are much more of a grab bag compared to the functions in other files. I'll work to organize them better with the refactoring.

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -244,12 +241,28 @@ private function get_breadcrumbs(): Generator {
 		}
 	}
 
+	/**
+	 * Determines whether currently inside a foreign element (MATH or SVG).
+	 *
```
mukeshpanchal27, line 246:
```suggestion
	 *
	 * @since n.e.x.t
	 *
```

tests/modules/images/image-loading-optimization/class-ilo-html-tag-processor-tests.php
```diff
@@ -0,0 +1,333 @@
...
+		);
+	}
+
+	/**
+	 * Test open_tags() and get_xpath().
+	 *
+	 * @covers ::open_tags
+	 * @covers ::get_xpath
+	 *
+	 * @dataProvider data_provider_sample_documents
```
mukeshpanchal27, line 287: Nice if we add `@param` that used in `dataProvider`.

tests/modules/images/image-loading-optimization/detection-tests.php
```diff
@@ -0,0 +1,73 @@
...
+				),
+			),
+		);
+	}
+
+	/**
+	 * Make sure the expected script is printed.
+	 *
+	 * @dataProvider data_provider_ilo_get_detection_script
+	 * @covers ::ilo_get_detection_script
```
mukeshpanchal27:
```suggestion
	 * @covers ::ilo_get_detection_script
	 *
	 * @dataProvider data_provider_ilo_get_detection_script
```

tests/modules/images/image-loading-optimization/optimization-tests.php
```diff
@@ -0,0 +1,528 @@
...
+				'expected' => false,
+			),
+		);
+	}
+
+	/**
+	 * Test ilo_can_optimize_response().
+	 *
+	 * @covers ::ilo_can_optimize_response
+	 * @dataProvider data_provider_test_ilo_can_optimize_response
```
mukeshpanchal27:
```suggestion
	 * @covers ::ilo_can_optimize_response
	 *
	 * @dataProvider data_provider_test_ilo_can_optimize_response
```
Nice if we add @param that used in dataProvider.

tests/modules/images/image-loading-optimization/optimization-tests.php
```diff
@@ -0,0 +1,528 @@
...
+				',
+			),
+		);
+	}
+
+	/**
+	 * Test ilo_construct_preload_links().
+	 *
+	 * @covers ::ilo_construct_preload_links
+	 * @dataProvider data_provider_test_ilo_construct_preload_links
```
mukeshpanchal27:
```suggestion
	 * @covers ::ilo_construct_preload_links
	 *
	 * @dataProvider data_provider_test_ilo_construct_preload_links
```

tests/modules/images/image-loading-optimization/optimization-tests.php
```diff
@@ -0,0 +1,528 @@
...
+			),
+
+		);
+	}
+
+	/**
+	 * Test ilo_optimize_template_output_buffer().
+	 *
+	 * @covers ::ilo_optimize_template_output_buffer
+	 * @dataProvider data_provider_test_ilo_optimize_template_output_buffer
```
mukeshpanchal27:
```suggestion
	 * @covers ::ilo_optimize_template_output_buffer
	 *
	 * @dataProvider data_provider_test_ilo_optimize_template_output_buffer
```

tests/modules/images/image-loading-optimization/storage/data-tests.php
```diff
@@ -0,0 +1,559 @@
...
+				},
+			),
+		);
+	}
+
+	/**
+	 * Test ilo_get_normalized_query_vars().
+	 *
+	 * @covers ::ilo_get_normalized_query_vars
+	 * @dataProvider data_provider_test_ilo_get_normalized_query_vars
```
mukeshpanchal27:
```suggestion
	 * @covers ::ilo_get_normalized_query_vars
	 *
	 * @dataProvider data_provider_test_ilo_get_normalized_query_vars
```

tests/modules/images/image-loading-optimization/storage/data-tests.php
```diff
@@ -0,0 +1,559 @@
...
+				'viewport_widths' => array( 400, 800 ),
+			),
+		);
+	}
+
+	/**
+	 * Test ilo_unshift_url_metrics().
+	 *
+	 * @covers ::ilo_unshift_url_metrics
+	 * @dataProvider data_provider_sample_size_and_breakpoints
```
mukeshpanchal27:
```suggestion
	 * @covers ::ilo_unshift_url_metrics
	 *
	 * @dataProvider data_provider_sample_size_and_breakpoints
```

tests/modules/images/image-loading-optimization/storage/data-tests.php
```diff
@@ -0,0 +1,559 @@
...
+				'viewport_widths' => array( 400, 600, 800, 1000 ),
+			),
+		);
+	}
+
+	/**
+	 * Test ilo_group_url_metrics_by_breakpoint().
+	 *
+	 * @covers ::ilo_group_url_metrics_by_breakpoint
+	 * @dataProvider data_provider_test_ilo_group_url_metrics_by_breakpoint
```
mukeshpanchal27:
```suggestion
	 * @covers ::ilo_group_url_metrics_by_breakpoint
	 *
	 * @dataProvider data_provider_test_ilo_group_url_metrics_by_breakpoint
```

tests/modules/images/image-loading-optimization/storage/data-tests.php
```diff
@@ -0,0 +1,559 @@
...
+				)
+			),
+		);
+	}
+
+	/**
+	 * Test ilo_get_needed_minimum_viewport_widths().
+	 *
+	 * @covers ::ilo_get_needed_minimum_viewport_widths
+	 * @dataProvider data_provider_test_ilo_get_needed_minimum_viewport_widths
```
mukeshpanchal27:
```suggestion
	 * @covers ::ilo_get_needed_minimum_viewport_widths
	 *
	 * @dataProvider data_provider_test_ilo_get_needed_minimum_viewport_widths
```

tests/modules/images/image-loading-optimization/storage/lock-tests.php
```diff
@@ -0,0 +1,128 @@
...
+				'expected' => 0,
+			),
+		);
+	}
+
+	/**
+	 * Test ilo_get_url_metric_storage_lock_ttl().
+	 *
+	 * @covers ::ilo_get_url_metric_storage_lock_ttl
+	 * @dataProvider data_provider_ilo_get_url_metric_storage_lock_ttl
```
mukeshpanchal27:
```suggestion
	 * @covers ::ilo_get_url_metric_storage_lock_ttl
	 *
	 * @dataProvider data_provider_ilo_get_url_metric_storage_lock_ttl
```

tests/modules/images/image-loading-optimization/storage/post-type-tests.php
```diff
@@ -0,0 +1,158 @@
...
+				'expected_value' => $valid_content,
+			),
+		);
+	}
+
+	/**
+	 * Test ilo_parse_stored_url_metrics().
+	 *
+	 * @covers ::ilo_parse_stored_url_metrics
+	 * @dataProvider data_provider_test_ilo_parse_stored_url_metrics
```
mukeshpanchal27:
```suggestion
	 * @covers ::ilo_parse_stored_url_metrics
	 *
	 * @dataProvider data_provider_test_ilo_parse_stored_url_metrics
```

tests/modules/images/image-loading-optimization/storage/rest-api-tests.php
```diff
@@ -0,0 +1,257 @@
...
+			)
+		);
+	}
+
+	/**
+	 * Test bad params.
+	 *
+	 * @covers ::ilo_register_endpoint
+	 * @covers ::ilo_handle_rest_request
+	 * @dataProvider data_provider_invalid_params
```
mukeshpanchal27:
```suggestion
	 * @covers ::ilo_handle_rest_request
	 *
	 * @dataProvider data_provider_invalid_params
```

mukeshpanchal27 (approved):
Thank you, @westonruter, for the updates. I've left some minor suggestions, but they don't pose any blockers.

- It would be nice to add `@param` annotations for the parameters used in the `dataProvider` for tests.

tests/modules/images/image-loading-optimization/class-ilo-html-tag-processor-tests.php
westonruter: Yeah, somewhat. But it looks like most of the other tests using data providers don't have `@param` tags in the plugin. Not that this is a good reason.

PR #899, opened 2023-12-11
Implement migration logic and UI from Performance Lab modules to their standalone plugins

### LINKED ISSUE #652: Implement migration logic and UI from old PL modules to new standalone plugins in PL settings screen

mukeshpanchal27:
## Summary

Fixes #652

## Checklist

### DIFF
--- a/admin/load.php
+++ b/admin/load.php
@@ -558,10 +558,9 @@ function perflab_enqueue_modules_page_scripts() {
 		'perflab-module-migration-notice',
 		'perflab_module_migration_notice',
 		array(
-			'ajaxurl'        => admin_url( 'admin-ajax.php' ),
-			'nonce'          => wp_create_nonce( 'perflab-install-activate-plugins' ),
-			'network_error'  => esc_html__( 'Network response was not ok.', 'performance-lab' ),
-			'prompt_message' => esc_html__( 'Are you sure you want to migrate legacy modules to standalone plugins?', 'performance-lab' ),
+			'ajaxurl'       => admin_url( 'admin-ajax.php' ),
+			'nonce'         => wp_create_nonce( 'perflab-install-activate-plugins' ),
+			'network_error' => esc_html__( 'Network response was not ok.', 'performance-lab' ),
 		)
 	);
 }
--- a/admin/load.php
+++ b/admin/load.php
@@ -59,6 +59,9 @@ function perflab_load_modules_page( $modules = null, $focus_areas = null ) {
 	// Handle style for settings page.
 	add_action( 'admin_head', 'perflab_print_modules_page_style' );
 
+	// Handle admin notices for settings page.
+	add_action( 'admin_notices', 'perflab_plugin_admin_notices' );
+
 	// Register sections for all focus areas, plus 'Other'.
 	if ( ! is_array( $focus_areas ) ) {
 		$focus_areas = perflab_get_focus_areas();
@@ -642,22 +645,6 @@ add_action( 'admin_action_perflab_deactivate_plugin', 'perflab_deactivate_plugin
  */
 function perflab_plugin_admin_notices() {
 	$standalone_plugins_with_active_modules = perflab_get_standalone_plugins_with_active_modules();
-	if ( 'settings_page_perflab-modules' !== get_current_screen()->id ) {
-		if ( ! empty( $standalone_plugins_with_active_modules ) ) {
-			echo '<div class="notice notice-warning is-dismissible"><p>';
-				printf(
-					/* translators: Link to Performance Lab Settings screen */
-					esc_html__( 'Performance Lab has detected usage of modules that have standalone plugin equivalents. Please visit the %1$s screen to see the plugins available for installation.', 'performance-lab' ),
-					sprintf(
-						'<a href="%s">%s</a>',
-						esc_html( menu_page_url( PERFLAB_MODULES_SCREEN, false ) ),
-						esc_html__( 'Performance Lab Settings', 'performance-lab' )
-					)
-				);
-			echo '</p></div>';
-			return;
-		}
-	}
 
 	if ( ! empty( $standalone_plugins_with_active_modules ) ) {
 		$module_data                 = perflab_get_modules();
@@ -729,7 +716,6 @@ function perflab_plugin_admin_notices() {
 		<?php
 	}
 }
-add_action( 'admin_notices', 'perflab_plugin_admin_notices' );
 
 /**
  * Helper function to check if a plugin is installed.
--- a/admin/load.php
+++ b/admin/load.php
@@ -645,7 +645,7 @@ function perflab_deactivate_plugin() {
 add_action( 'admin_action_perflab_deactivate_plugin', 'perflab_deactivate_plugin' );
 
 /**
- * Callback function hooked to admin_notices to render plugin activate/deactivate notices.
+ * Callback function hooked to admin_notices to render admin notices on the plugin's screen.
  *
  * @since n.e.x.t
  */
@@ -684,12 +684,12 @@ function perflab_plugin_admin_notices() {
 			$message  = '<p>';
 			$message .= sprintf(
 				/* translators: Module name */
-				esc_html__( 'Performance Lab has detected usage of "%s" module that have standalone plugin equivalents. Installing and activating a performance plugin will disable the legacy module equivalent. This will not impact functionality of the module.', 'performance-lab' ),
+				esc_html__( 'Your site is using the "%s" module which will be removed in the future in favor of its equivalent standalone plugin. Please click the following button to install and activate the relevant plugin in favor of the module. This will not impact any of the underlying functionality.', 'performance-lab' ),
 				esc_attr( $available_modules_list[0] )
 			);
 			$message .= '</p>';
 		} else {
-			$message  = '<p>' . esc_html__( 'Performance Lab has detected usage of modules that have standalone plugin equivalents. Please see the available plugins below. Installing and activating a performance plugin will disable the legacy module equivalent. This will not impact functionality of the module.', 'performance-lab' ) . '</p>';
+			$message  = '<p>' . esc_html__( 'Your site is using modules which will be removed in the future in favor of their equivalent standalone plugins. Please click the following button to install and activate the relevant plugins in favor of the modules. This will not impact any of the underlying functionality.', 'performance-lab' ) . '</p>';
 			$message .= '<strong>' . esc_html__( 'Available standalone plugins:', 'performance-lab' ) . '</strong>';
 			$message .= '<ol>';
 			foreach ( $available_modules_list as $names ) {
@@ -698,15 +698,17 @@ function perflab_plugin_admin_notices() {
 			$message .= '</ol>';
 		}
 
-		echo '<div class="notice notice-warning is-dismissible">';
-			echo wp_kses_post( $message );
-			echo '<p class="perflab-button-wrapper">';
-				echo '<button type="button" class="button button-primary perflab-install-active-plugin">';
-					echo esc_html__( 'Migrate legacy modules to standalone plugins', 'performance-lab' );
-				echo '</button>';
-				echo '<span class="dashicons dashicons-update hidden"></span>';
-			echo '</p>';
-		echo '</div>';
+		?>
+		<div class="notice notice-warning is-dismissible">
+			<?php echo wp_kses_post( $message ); ?>
+			<p class="perflab-button-wrapper">
+				<button type="button" class="button button-primary perflab-install-active-plugin">
+					<?php echo esc_html__( 'Migrate legacy modules to standalone plugins', 'performance-lab' ); ?>
+				</button>
+				<span class="dashicons dashicons-update hidden"></span>
+			</p>
+		</div>
+		<?php
 	}
 }
 
--- a/admin/load.php
+++ b/admin/load.php
@@ -57,7 +57,7 @@ function perflab_load_modules_page( $modules = null, $focus_areas = null ) {
 	add_action( 'admin_enqueue_scripts', 'perflab_enqueue_modules_page_scripts' );
 
 	// Handle style for settings page.
-	add_action( 'admin_footer', 'perflab_enqueue_modules_page_style' );
+	add_action( 'admin_head', 'perflab_print_modules_page_style' );
 
 	// Register sections for all focus areas, plus 'Other'.
 	if ( ! is_array( $focus_areas ) ) {
@@ -748,7 +748,7 @@ function is_plugin_installed( $plugin_slug ) {
  *
  * @since n.e.x.t
  */
-function perflab_enqueue_modules_page_style() {
+function perflab_print_modules_page_style() {
 	?>
 <style type="text/css">
 	.perflab-button-wrapper {
--- a/admin/load.php
+++ b/admin/load.php
@@ -616,7 +616,34 @@ add_action( 'admin_action_perflab_deactivate_plugin', 'perflab_deactivate_plugin
  */
 function perflab_plugin_admin_notices() {
 	if ( 'settings_page_perflab-modules' !== get_current_screen()->id ) {
+		if ( ! empty( perflab_get_standalone_plugins_with_active_modules() ) ) {
+			?>
+			<div class="notice notice-warning is-dismissible">
+				<p>
+					<?php
+					printf(
+						/* translators: 1: Link to Performance Lab Settings screen */
+						esc_html__( 'Performance Lab has detected usage of modules that have standalone plugin equivalents. Please visit the %1$s screen to see the plugins available for installation.', 'performance-lab' ),
+						sprintf(
+							'<a href="%s">%s</a>',
+							esc_html( menu_page_url( PERFLAB_MODULES_SCREEN, false ) ),
+							esc_html__( 'Performance Lab Settings', 'performance-lab' )
+						)
+					);
+					?>
+				</p>
+			</div>
+			<?php
+		}
 		return;
+	}
+
+	if ( ! empty( perflab_get_standalone_plugins_with_active_modules() ) ) {
+		?>
+		<div class="notice notice-warning is-dismissible">
+			<p><?php esc_html_e( 'Performance Lab has detected usage of modules that have standalone plugin equivalents. Please see the available plugins below. Installing and activating a performance plugin will disable the legacy module equivalent. This will not impact functionality of the module.', 'performance-lab' ); ?></p>
+		</div>
+		<?php
 	}
 
 	if ( isset( $_GET['activate'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
--- a/admin/load.php
+++ b/admin/load.php
@@ -681,7 +681,7 @@ function perflab_install_activate_standalone_plugins_callback() {
 
 		// Return early if plugin API return an error.
 		if ( ! $api ) {
-			$status['errorMessage'] = __( 'An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration.', 'performance-lab' );
+			$status['errorMessage'] = html_entity_decode( __( 'An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration.', 'performance-lab' ), ENT_QUOTES );
 			wp_send_json_error( $status );
 		}
 
--- a/admin/load.php
+++ b/admin/load.php
@@ -56,6 +56,9 @@ function perflab_load_modules_page( $modules = null, $focus_areas = null ) {
 	// Handle script enqueuing for settings page.
 	add_action( 'admin_enqueue_scripts', 'perflab_enqueue_modules_page_scripts' );
 
+	// Handle style for settings page.
+	add_action( 'admin_footer', 'perflab_enqueue_modules_page_style' );
+
 	// Register sections for all focus areas, plus 'Other'.
 	if ( ! is_array( $focus_areas ) ) {
 		$focus_areas = perflab_get_focus_areas();
@@ -529,6 +532,29 @@ function perflab_enqueue_modules_page_scripts() {
 
 	wp_enqueue_script( 'thickbox' );
 	wp_enqueue_style( 'thickbox' );
+
+	wp_enqueue_script(
+		'perflab-admin',
+		plugin_dir_url( __FILE__ ) . 'perflab-admin.js',
+		array(),
+		'1.0.0',
+		array(
+			'strategy' => 'defer',
+		)
+	);
+
+	wp_localize_script(
+		'perflab-admin',
+		'perflab_admin',
+		array(
+			'ajaxurl'          => admin_url( 'admin-ajax.php' ),
+			'nonce'            => wp_create_nonce( 'perflab-install-activate-plugins' ),
+			'has_permission'   => current_user_can( 'install_plugins' ) && current_user_can( 'activate_plugins' ),
+			'permission_error' => esc_html__( 'Sorry, you are not allowed to manage plugins for this site. Please contact the administrator.', 'performance-lab' ),
+			'network_error'    => esc_html__( 'Network response was not ok.', 'performance-lab' ),
+			'prompt_message'   => esc_html__( 'Are you sure you want to migrate legacy modules to standalone plugins?', 'performance-lab' ),
+		)
+	);
 }
 
 /**
@@ -611,39 +637,80 @@ add_action( 'admin_action_perflab_deactivate_plugin', 'perflab_deactivate_plugin
  * Callback function hooked to admin_notices to render plugin activate/deactivate notices.
  *
  * @since n.e.x.t
- *
- * @return void
  */
 function perflab_plugin_admin_notices() {
+	$standalone_plugins_with_active_modules = perflab_get_standalone_plugins_with_active_modules();
 	if ( 'settings_page_perflab-modules' !== get_current_screen()->id ) {
-		if ( ! empty( perflab_get_standalone_plugins_with_active_modules() ) ) {
-			?>
-			<div class="notice notice-warning is-dismissible">
-				<p>
-					<?php
-					printf(
-						/* translators: 1: Link to Performance Lab Settings screen */
-						esc_html__( 'Performance Lab has detected usage of modules that have standalone plugin equivalents. Please visit the %1$s screen to see the plugins available for installation.', 'performance-lab' ),
-						sprintf(
-							'<a href="%s">%s</a>',
-							esc_html( menu_page_url( PERFLAB_MODULES_SCREEN, false ) ),
-							esc_html__( 'Performance Lab Settings', 'performance-lab' )
-						)
-					);
-					?>
-				</p>
-			</div>
-			<?php
+		if ( ! empty( $standalone_plugins_with_active_modules ) ) {
+			echo '<div class="notice notice-warning is-dismissible"><p>';
+				printf(
+					/* translators: Link to Performance Lab Settings screen */
+					esc_html__( 'Performance Lab has detected usage of modules that have standalone plugin equivalents. Please visit the %1$s screen to see the plugins available for installation.', 'performance-lab' ),
+					sprintf(
+						'<a href="%s">%s</a>',
+						esc_html( menu_page_url( PERFLAB_MODULES_SCREEN, false ) ),
+						esc_html__( 'Performance Lab Settings', 'performance-lab' )
+					)
+				);
+			echo '</p></div>';
+			return;
 		}
-		return;
 	}
 
-	if ( ! empty( perflab_get_standalone_plugins_with_active_modules() ) ) {
-		?>
-		<div class="notice notice-warning is-dismissible">
-			<p><?php esc_html_e( 'Performance Lab has detected usage of modules that have standalone plugin equivalents. Please see the available plugins below. Installing and activating a performance plugin will disable the legacy module equivalent. This will not impact functionality of the module.', 'performance-lab' ); ?></p>
-		</div>
-		<?php
+	if ( ! empty( $standalone_plugins_with_active_modules ) ) {
+		$module_data                 = perflab_get_modules();
+		$standalone_plugins_file_map = perflab_get_standalone_plugins_file_map();
+		$modules                     = array();
+		foreach ( $standalone_plugins_with_active_modules as $standalone_plugins_slug => $module_slug ) {
+			if (
+				isset( $module_data[ $module_slug ] ) &&
+				(
+					! is_plugin_installed( $standalone_plugins_file_map[ $standalone_plugins_slug ] ) ||
+					! is_plugin_active( $standalone_plugins_file_map[ $standalone_plugins_slug ] )
+				)
+			) {
+				$modules[] = esc_attr( $module_data[ $module_slug ]['name'] );
+			}
+		}
+
+		$modules_count = count( $modules );
+		if ( $modules_count < 1 ) {
+			return;
+		} elseif ( 1 === $modules_count ) {
+			echo '<div class="notice notice-warning is-dismissible">';
+				printf(
+					'%s %s',
+					sprintf(
+						/* translators: Performance module */
+						'<p>' . esc_html__( 'Performance Lab has detected usage of "%s" module that have standalone plugin equivalents. Installing and activating a performance plugin will disable the legacy module equivalent. This will not impact functionality of the module.', 'performance-lab' ) . '</p>',
+						esc_attr( $modules[0] )
+					),
+					sprintf(
+						/* translators: Button text */
+						'<p class="perflab-button-wrapper"><button type="button" class="button button-primary perflab-install-active-plugin">%s</button><span class="dashicons dashicons-update hidden"></span></p>',
+						esc_html__( 'Migrate legacy modules to standalone plugins', 'performance-lab' )
+					)
+				);
+			echo '</div>';
+		} else {
+			$module_name = '<ol>';
+			foreach ( $modules as $module ) {
+				$module_name .= "<li>$module</li>";
+			}
+			$module_name .= '</ol>';
+			echo '<div class="notice notice-warning is-dismissible">';
+				printf(
+					'%s %s %s %s',
+					'<p>' . esc_html__( 'Performance Lab has detected usage of modules that have standalone plugin equivalents. Please see the available plugins below. Installing and activating a performance plugin will disable the legacy module equivalent. This will not impact functionality of the module.', 'performance-lab' ) . '</p>',
+					'<strong>' . esc_html__( 'Available standalone plugins:', 'performance-lab' ) . '</strong>',
+					wp_kses_post( $module_name ),
+					sprintf(
+						'<p class="perflab-button-wrapper"><button type="button" class="button button-primary perflab-install-active-plugin">%s</button><span class="dashicons dashicons-update hidden"></span></p>',
+						esc_html__( 'Migrate legacy modules to standalone plugins', 'performance-lab' )
+					)
+				);
+			echo '</div>';
+		}
 	}
 
 	if ( isset( $_GET['activate'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
@@ -661,3 +728,35 @@ function perflab_plugin_admin_notices() {
 	}
 }
 add_action( 'admin_notices', 'perflab_plugin_admin_notices' );
+
+/**
+ * Helper function to check if a plugin is installed.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $plugin_slug The plugin slug.
+ */
+function is_plugin_installed( $plugin_slug ) {
+	$plugins = get_plugins();
+	return isset( $plugins[ $plugin_slug ] );
+}
+
+/**
+ * Callback function to handle admin inline style.
+ *
+ * @since n.e.x.t
+ */
+function perflab_enqueue_modules_page_style() {
+	?>
+<style type="text/css">
+	.perflab-button-wrapper {
+		display: flex;
+		align-items: center;
+	}
+	.perflab-button-wrapper span {
+		animation: rotation 2s infinite linear;
+		margin-left: 5px;
+	}
+</style>
+	<?php
+}
--- a/admin/load.php
+++ b/admin/load.php
@@ -558,12 +558,10 @@ function perflab_enqueue_modules_page_scripts() {
 		'perflab-module-migration-notice',
 		'perflab_module_migration_notice',
 		array(
-			'ajaxurl'          => admin_url( 'admin-ajax.php' ),
-			'nonce'            => wp_create_nonce( 'perflab-install-activate-plugins' ),
-			'has_permission'   => current_user_can( 'install_plugins' ) && current_user_can( 'activate_plugins' ),
-			'permission_error' => esc_html__( 'Sorry, you are not allowed to manage plugins for this site. Please contact the administrator.', 'performance-lab' ),
-			'network_error'    => esc_html__( 'Network response was not ok.', 'performance-lab' ),
-			'prompt_message'   => esc_html__( 'Are you sure you want to migrate legacy modules to standalone plugins?', 'performance-lab' ),
+			'ajaxurl'        => admin_url( 'admin-ajax.php' ),
+			'nonce'          => wp_create_nonce( 'perflab-install-activate-plugins' ),
+			'network_error'  => esc_html__( 'Network response was not ok.', 'performance-lab' ),
+			'prompt_message' => esc_html__( 'Are you sure you want to migrate legacy modules to standalone plugins?', 'performance-lab' ),
 		)
 	);
 }
@@ -655,6 +653,11 @@ add_action( 'wp_ajax_perflab_install_activate_standalone_plugins', 'perflab_inst
 function perflab_install_activate_standalone_plugins_callback() {
 	if ( ! wp_verify_nonce( $_REQUEST['nonce'], 'perflab-install-activate-plugins' ) ) {
 		$status['errorMessage'] = esc_html__( 'Invalid nonce: Please refresh and try again.', 'performance-lab' );
+		wp_send_json_error( $status );
+	}
+
+	if ( ! current_user_can( 'install_plugins' ) || ! current_user_can( 'activate_plugins' ) ) {
+		$status['errorMessage'] = esc_html__( 'Sorry, you are not allowed to manage plugins for this site. Please contact the administrator.', 'performance-lab' );
 		wp_send_json_error( $status );
 	}
 
--- a/admin/load.php
+++ b/admin/load.php
@@ -644,6 +644,86 @@ function perflab_deactivate_plugin() {
 }
 add_action( 'admin_action_perflab_deactivate_plugin', 'perflab_deactivate_plugin' );
 
+// WordPress AJAX action to handle the button click event.
+add_action( 'wp_ajax_perflab_install_activate_standalone_plugins', 'perflab_install_activate_standalone_plugins_callback' );
+
+/**
+ * Handles the standalone plugin install and activation via AJAX.
+ *
+ * @since n.e.x.t
+ */
+function perflab_install_activate_standalone_plugins_callback() {
+	if ( ! wp_verify_nonce( $_REQUEST['nonce'], 'perflab-install-activate-plugins' ) ) {
+		$status['errorMessage'] = esc_html__( 'Invalid nonce: Please refresh and try again.', 'performance-lab' );
+		wp_send_json_error( $status );
+	}
+
+	if ( ! function_exists( 'plugins_api' ) ) {
+		require_once ABSPATH . 'wp-admin/includes/plugin-install.php';
+	}
+	require_once ABSPATH . 'wp-admin/includes/class-wp-upgrader.php';
+	require_once ABSPATH . 'wp-admin/includes/class-wp-ajax-upgrader-skin.php';
+
+	$plugins_to_activate = perflab_get_active_modules_with_standalone_plugins();
+	$modules             = perflab_get_module_settings();
+	$plugins             = get_plugins();
+	$status              = array();
+
+	foreach ( $plugins_to_activate as $module_slug ) {
+
+		// Skip checking for already activated plugin.
+		if ( perflab_is_standalone_plugin_loaded( $module_slug ) ) {
+			continue;
+		}
+
+		$plugin_slug     = basename( $module_slug );
+		$plugin_basename = $plugin_slug . '/load.php';
+		$api             = perflab_query_plugin_info( $plugin_slug );
+
+		// Return early if plugin API return an error.
+		if ( ! $api ) {
+			$status['errorMessage'] = esc_html__( 'An unexpected error occurred. Something may be wrong with WordPress.org or this servers configuration.', 'performance-lab' );
+			wp_send_json_error( $status );
+		}
+
+		if ( ! $plugin_slug ) {
+			$status['errorMessage'] = esc_html__( 'Invalid plugin.', 'performance-lab' );
+			wp_send_json_error( $status );
+		}
+
+		// Install the plugin if it is not installed yet.
+		if ( ! isset( $plugins[ $plugin_basename ] ) ) {
+			// Replace new Plugin_Installer_Skin with new Quiet_Upgrader_Skin when output needs to be suppressed.
+			$skin     = new WP_Ajax_Upgrader_Skin( array( 'api' => $api ) );
+			$upgrader = new Plugin_Upgrader( $skin );
+			$result   = $upgrader->install( $api['download_link'] );
+
+			if ( is_wp_error( $result ) ) {
+				$status['errorMessage'] = $result->get_error_message();
+				wp_send_json_error( $status );
+			} elseif ( is_wp_error( $skin->result ) ) {
+				$status['errorMessage'] = $skin->result->get_error_message();
+				wp_send_json_error( $status );
+			} elseif ( $skin->get_errors()->has_errors() ) {
+				$status['errorMessage'] = $skin->get_error_messages();
+				wp_send_json_error( $status );
+			}
+		}
+
+		$result = activate_plugin( WP_PLUGIN_DIR . '/' . $plugin_basename );
+		if ( is_wp_error( $result ) ) {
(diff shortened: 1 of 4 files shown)


admin/load.php
```diff
@@ -56,6 +56,9 @@ function perflab_load_modules_page( $modules = null, $focus_areas = null ) {
 	// Handle script enqueuing for settings page.
 	add_action( 'admin_enqueue_scripts', 'perflab_enqueue_modules_page_scripts' );
 
+	// Handle style for settings page.
+	add_action( 'admin_footer', 'perflab_enqueue_modules_page_style' );
```
felixarntz: Maybe better to use `admin_head`, as CSS is probably crucial? Nit-pick: I think it would be better to use the word "print" instead of "enqueue" in the function name, as the function doesn't enqueue any styles.

admin/load.php
```diff
@@ -613,12 +639,80 @@ function perflab_deactivate_plugin() {
  * Callback function hooked to admin_notices to render plugin activate/deactivate notices.
  *
  * @since n.e.x.t
- *
- * @return void
  */
 function perflab_plugin_admin_notices() {
+	$standalone_plugins_with_active_modules = perflab_get_standalone_plugins_with_active_modules();
 	if ( 'settings_page_perflab-modules' !== get_current_screen()->id ) {
-		return;
```
felixarntz: This shouldn't be changed. We should make sure to only ever add admin notices to our own screen. It's not a good practice to add admin notices throughout the entire WP Admin.

See also the #652 requirements, the notice should only appear on the Performance Lab screen.

admin/plugins.php
```diff
@@ -55,6 +55,54 @@ function perflab_get_standalone_plugins() {
...
+ * @since n.e.x.t
+ *
+ * @return string[]
+ */
+function perflab_get_standalone_plugins_file_map() {
+	return array(
+		'webp-uploads'          => 'webp-uploads/load.php',
+		'dominant-color-images' => 'dominant-color-images/load.php',
+	);
+}
```
felixarntz: Those functions shouldn't be needed. We already have `perflab_get_standalone_plugins_constants()` and `perflab_is_standalone_plugin_loaded()`, and we should be able to use those functions instead of introducing new ones.

The plugin slug is always the same as the module slug (minus the focus area directory), so let's just rely on that rather than hard-coding another map.

admin/plugins.php
```diff
@@ -55,6 +55,54 @@ function perflab_get_standalone_plugins() {
...
+	$modules                                = perflab_get_module_settings();
+	$standalone_plugins_with_active_modules = array_filter(
+		perflab_get_standalone_plugins_module_map(),
+		static function ( $v ) use ( $modules ) {
+			return ! empty( $modules[ $v ] ) && $modules[ $v ]['enabled'];
+		},
+		ARRAY_FILTER_USE_BOTH
+	);
+
+	return $standalone_plugins_with_active_modules;
```
felixarntz: See above. Also why do we need `ARRAY_FILTER_USE_BOTH`?

I would suggest something like this:
```suggestion
	$modules = perflab_get_module_settings();
	return array_filter(
		array_keys( perflab_get_standalone_plugins_constants() ),
		static function ( $v ) use ( $modules ) {
			return ! empty( $modules[ $v ] ) && $modules[ $v ]['enabled'];
		}
	);
```

It should be sufficient to simply return the module slugs here. Maybe we can then rename the function to `perflab_get_active_modules_with_standalone_plugins()` to clarify?

admin/plugins.php
```diff
@@ -262,8 +317,21 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+			if ( $plugin_data['slug'] === $standalone_plugin && perflab_is_standalone_plugin_loaded( $module_slug ) ) {
+				$is_standalone_plugin_loaded = true;
+			}
+		}
+
+		if ( true === $perflab_has_enabled_module && ! $is_standalone_plugin_loaded ) {
+			echo '<div class="notice inline notice-warning notice-alt"><p>';
+			echo esc_html__( 'An active performance module was detected, please install and activate this plugin instead.', 'performance-lab' );
+			echo '</p></div>';
+		}
```
felixarntz: This wasn't part of the requirements from the issue, so I would recommend to remove it. The PR is already relatively complex, so we shouldn't add more than needed.

admin/plugins.php
```diff
@@ -137,7 +191,8 @@ function perflab_render_plugin_card( array $plugin_data ) {
 	$tested_wp      = ( empty( $plugin_data['tested'] ) || version_compare( get_bloginfo( 'version' ), $plugin_data['tested'], '<=' ) );
 	$action_links   = array();
 
-	$status = install_plugin_install_status( $plugin_data );
+	$status                     = install_plugin_install_status( $plugin_data );
+	$perflab_has_enabled_module = ! empty( $plugin_data['perflab_has_enabled_module'] ) && true === $plugin_data['perflab_has_enabled_module'];
```
felixarntz: See my other comment below, this change shouldn't be needed as it's not related to any requirement.

admin/load.php
```diff
@@ -613,12 +639,80 @@ function perflab_deactivate_plugin() {
...
+			if (
+				isset( $module_data[ $module_slug ] ) &&
+				(
+					! is_plugin_installed( $standalone_plugins_file_map[ $standalone_plugins_slug ] ) ||
+					! is_plugin_active( $standalone_plugins_file_map[ $standalone_plugins_slug ] )
+				)
+			) {
+				$modules[] = esc_attr( $module_data[ $module_slug ]['name'] );
+			}
+		}
```
felixarntz: Why perform all these checks here? You're already implementing a function to get active modules with standalone plugins, so why not use that here? That would be much simpler.

Last but not least, this shouldn't be escaped. Escaping should always happen right when printing output.

Something like this:
```suggestion
		$active_modules_with_plugins = perflab_get_active_modules_with_standalone_plugins();
		$module_names                = array();
		foreach ( $active_modules_with_plugins as $module_slug ) {
			if ( isset( $module_data[ $module_slug ] ) ) {
				$module_names[] = $module_data[ $module_slug ]['name'];
			}
		}
```

admin/load.php
```diff
@@ -613,12 +639,80 @@ function perflab_deactivate_plugin() {
...
+
+		$modules_count = count( $modules );
+		if ( $modules_count < 1 ) {
+			return;
+		} elseif ( 1 === $modules_count ) {
```
felixarntz: No need for `else` after a `return` statement, we should avoid that.
```suggestion
		}

		if ( 1 === $modules_count ) {
```

admin/load.php
```diff
@@ -613,12 +639,80 @@ function perflab_deactivate_plugin() {
...
+					'%s %s %s %s',
+					'<p>' . esc_html__( 'Performance Lab has detected usage of modules that have standalone plugin equivalents. Please see the available plugins below. Installing and activating a performance plugin will disable the legacy module equivalent. This will not impact functionality of the module.', 'performance-lab' ) . '</p>',
+					'<strong>' . esc_html__( 'Available standalone plugins:', 'performance-lab' ) . '</strong>',
+					wp_kses_post( $module_name ),
+					sprintf(
+						'<p class="perflab-button-wrapper"><button type="button" class="button button-primary perflab-install-active-plugin">%s</button><span class="dashicons dashicons-update hidden"></span></p>',
+						esc_html__( 'Migrate legacy modules to standalone plugins', 'performance-lab' )
+					)
+				);
+			echo '</div>';
```
felixarntz: A few recommendations here:
* Can you combine the two conditions so that the common output is shared between them? We should only need if-else for the parts that differ.
* It would be great if you could directly print HTML where possible rather than putting it in PHP strings and echoing them. Also, the `sprintf()` calls with all the HTML are difficult to follow, we can probably avoid them.
    * It's easier to read.
    * It's more error-proof as we'll have HTML syntax highlighting.

admin/load.php
```diff
@@ -636,3 +730,35 @@ function perflab_plugin_admin_notices() {
...
+ * Helper function to check if a plugin is installed.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $plugin_slug The plugin slug.
+ */
+function is_plugin_installed( $plugin_slug ) {
+	$plugins = get_plugins();
+	return isset( $plugins[ $plugin_slug ] );
+}
```
felixarntz: Based on my suggestions, I believe this function will only be used in a single place. Since it's very simple, I would suggest to remove it and instead inline that functionality in the one place where it's needed. Every function is another piece of API that we'll have to support, so it's best to have as many as needed, but as few as possible.

Last but not least, we shouldn't have an unprefixed function like this in our codebase.

admin/plugins.php
```diff
@@ -390,3 +458,87 @@ function perflab_render_plugin_card( array $plugin_data ) {
 	</div>
 	<?php
 }
+
+// WordPress AJAX action to handle the button click event.
+add_action( 'wp_ajax_perflab_install_activate_standalone_plugins', 'perflab_install_activate_standalone_plugins_callback' );
+add_action( 'wp_ajax_nopriv_perflab_install_activate_standalone_plugins', 'perflab_install_activate_standalone_plugins_callback' );
```
felixarntz: I don't think this needs to be available for unprivileged users? They wouldn't (and shouldn't) be able to install or activate plugins anyway.

admin/plugins.php
```diff
@@ -390,3 +458,87 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+	}
+
+	if ( ! function_exists( 'plugins_api' ) ) {
+		require_once ABSPATH . 'wp-admin/includes/plugin-install.php';
+	}
+	require_once ABSPATH . 'wp-admin/includes/class-wp-upgrader.php';
+	require_once ABSPATH . 'wp-admin/includes/class-wp-ajax-upgrader-skin.php';
+
+	$plugins_to_install = $_REQUEST['data'];
+	$plugins_to_install = explode( ',', $plugins_to_install );
```
felixarntz: Nit-pick: Shouldn't this better be called `$plugins_to_activate`? The goal here is to activate those plugins. Installing those that aren't installed is only a partial step to getting there.

admin/plugins.php
```diff
@@ -390,3 +458,87 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+	$status = array();
+	foreach ( $plugins_to_install as $plugin_slug ) {
+
+		$api = perflab_query_plugin_info( $plugin_slug );
+
+		$module_slug = isset( $get_standalone_plugins_module_map[ $plugin_slug ] ) ? $get_standalone_plugins_module_map[ $plugin_slug ] : '';
+		$plugin_slug = $get_standalone_plugins_file_map[ $plugin_slug ];
+
+		// Check if the plugin is already installed.
+		if ( ! is_plugin_installed( $plugin_slug ) ) {
```
felixarntz: See my other comment, we can probably inline the check here, and have the `get_plugins()` function call before the `foreach` loop.

admin/plugins.php
```diff
@@ -390,3 +458,87 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+	$get_standalone_plugins_file_map   = perflab_get_standalone_plugins_file_map();
+	$modules                           = perflab_get_module_settings();
+
+	$status = array();
+	foreach ( $plugins_to_install as $plugin_slug ) {
+
+		$api = perflab_query_plugin_info( $plugin_slug );
+
+		$module_slug = isset( $get_standalone_plugins_module_map[ $plugin_slug ] ) ? $get_standalone_plugins_module_map[ $plugin_slug ] : '';
+		$plugin_slug = $get_standalone_plugins_file_map[ $plugin_slug ];
```
felixarntz: See my other comments, I think this can be simplified by relying on the existing function `perflab_get_standalone_plugins_constants()` and assuming that the plugin slug is the same as the module slug minus the focus area directory.

admin/plugins.php
```diff
@@ -390,3 +458,87 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+			$status['errorMessage'] = $result->get_error_message();
+			wp_send_json_error( $status );
+		}
+
+		// Deactivate legacy modules.
+		foreach ( $modules as $slug => $data ) {
+			if ( $module_slug === $slug ) {
+				unset( $modules[ $slug ] );
+			}
+		}
```
felixarntz: I think this loop is unnecessary.
```suggestion
		unset( $modules[ $module_slug ] );
```

admin/load.php
```diff
@@ -531,6 +534,29 @@ function perflab_enqueue_modules_page_scripts() {
 	wp_enqueue_style( 'thickbox' );
 
 	wp_enqueue_script( 'plugin-install' );
+
+	wp_enqueue_script(
```
felixarntz, line 547: We should only enqueue these scripts if needed, to avoid unnecessary JavaScript. It is only needed if there are any active modules with standalone plugins. For example call the new `perflab_get_active_modules_with_standalone_plugins()` function here and bail early if there aren't any.

admin/load.php
```diff
@@ -531,6 +534,29 @@ function perflab_enqueue_modules_page_scripts() {
...
+		'perflab-admin',
+		plugin_dir_url( __FILE__ ) . 'perflab-admin.js',
```
felixarntz: Let's name this script more specific to what it is: It's not a general script for the admin screen, it's only for the very specific scenario of migrating modules to their standalone plugins. Maybe `perflab-module-migration-notice`?

felixarntz (changes requested):
Thank you @mukeshpanchal27 for the PR. I think there are a few things that can be simplified here. Most importantly, let's remove the additional piece of functionality in the module cards which wasn't required and makes the PR more complicated. I left more specific feedback below. Please let me know if you have any questions for clarification.

admin/load.php
```diff
@@ -613,12 +639,80 @@ function perflab_deactivate_plugin() {
...
 	if ( 'settings_page_perflab-modules' !== get_current_screen()->id ) {
-		return;
+		if ( ! empty( $standalone_plugins_with_active_modules ) ) {
+			echo '<div class="notice notice-warning is-dismissible"><p>';
+				printf(
+					/* translators: Link to Performance Lab Settings screen */
+					esc_html__( 'Performance Lab has detected usage of modules that have standalone plugin equivalents. Please visit the %1$s screen to see the plugins available for installation.', 'performance-lab' ),
+					sprintf(
+						'<a href="%s">%s</a>',
+						esc_html( menu_page_url( PERFLAB_MODULES_SCREEN, false ) ),
```
```suggestion
						esc_url( menu_page_url( PERFLAB_MODULES_SCREEN, false ) ),
```

admin/load.php
```diff
@@ -613,12 +639,80 @@ function perflab_deactivate_plugin() {
...
+			}
+		}
+
+		$modules_count = count( $modules );
+		if ( $modules_count < 1 ) {
+			return;
+		} elseif ( 1 === $modules_count ) {
+			echo '<div class="notice notice-warning is-dismissible">';
+				printf(
+					'%s %s',
```
westonruter: Why the extra `printf()` at the top? Seems like the two `sprintf()` calls could be just made their own separate `printf()` calls.

admin/load.php
```diff
@@ -613,12 +639,80 @@ function perflab_deactivate_plugin() {
...
+						/* translators: Button text */
+						'<p class="perflab-button-wrapper"><button type="button" class="button button-primary perflab-install-active-plugin">%s</button><span class="dashicons dashicons-update hidden"></span></p>',
+						esc_html__( 'Migrate legacy modules to standalone plugins', 'performance-lab' )
+					)
+				);
+			echo '</div>';
+		} else {
+			$module_name = '<ol>';
+			foreach ( $modules as $module ) {
+				$module_name .= "<li>$module</li>";
```
```suggestion
				$module_name .= sprintf( "<li>%s</li>", esc_html( $module ) );
```

admin/load.php
```diff
@@ -613,12 +639,80 @@ function perflab_deactivate_plugin() {
...
+			}
+			$module_name .= '</ol>';
+			echo '<div class="notice notice-warning is-dismissible">';
+				printf(
+					'%s %s %s %s',
```
westonruter: Per above, I don't think this use of `printf()` is ideal. Better to just print each separately.

admin/load.php
```diff
@@ -613,12 +639,80 @@ function perflab_deactivate_plugin() {
...
+					),
+					sprintf(
+						/* translators: Button text */
+						'<p class="perflab-button-wrapper"><button type="button" class="button button-primary perflab-install-active-plugin">%s</button><span class="dashicons dashicons-update hidden"></span></p>',
+						esc_html__( 'Migrate legacy modules to standalone plugins', 'performance-lab' )
+					)
+				);
+			echo '</div>';
+		} else {
+			$module_name = '<ol>';
```
westonruter: Is this variable named properly? Should it rather be named something like `$available_modules_list`?

admin/perflab-admin.js
```diff
@@ -0,0 +1,53 @@
...
+		});
+
+		document.addEventListener( 'click', function(event) {
+			if ( event.target.classList.contains( 'perflab-install-active-plugin' ) ) {
+                if ( ! perflab_admin.has_permission ) {
+                    alert( perflab_admin.permission_error );
+                    return;
+                }
+				if ( confirm( perflab_admin.prompt_message ) ) {
+					var __this = event.target;
```
westonruter: I suggest `target` instead of `__this`

westonruter (changes requested):

admin/plugins.php
mukeshpanchal27: 😔

admin/load.php
mukeshpanchal27: Just removed the other page admin notice so not needed now.

admin/plugins.php
mukeshpanchal27: Thanks @felixarntz for the feedback.

We use `perflab_get_standalone_plugins_file_map()` to get plugin main file to check if that plugin is installed (not activated) in WP or not.

mukeshpanchal27:
Thank you, @felixarntz and @westonruter, for the review feedback. I've simplified the logic, addressed all the feedback, and responded to some open questions. The PR is now ready for review.

admin/load.php
```diff
@@ -613,14 +648,8 @@ function perflab_deactivate_plugin() {
  * Callback function hooked to admin_notices to render plugin activate/deactivate notices.
```
felixarntz:
```suggestion
 * Callback function hooked to admin_notices to render admin notices on the plugin's screen.
```

admin/plugins.php
```diff
@@ -64,10 +95,16 @@ function perflab_render_plugins_ui() {
...
+	$get_active_modules_with_standalone_plugins = perflab_get_active_modules_with_standalone_plugins();
 	foreach ( perflab_get_standalone_plugins() as $managed_standalone_plugin_slug ) {
 		$standalone_plugins[ $managed_standalone_plugin_slug ] = array(
-			'plugin_data' => perflab_query_plugin_info( $managed_standalone_plugin_slug ),
+			'plugin_data' => array_merge(
+				perflab_query_plugin_info( $managed_standalone_plugin_slug ),
+				array(
+					'perflab_has_enabled_module' => ! empty( $get_active_modules_with_standalone_plugins[ $managed_standalone_plugin_slug ] ),
+				)
+			),
```
felixarntz: Can you please revert this to how it was before? Since the `perflab_has_enabled_module` is no longer used after https://github.com/WordPress/performance/pull/899#discussion_r1424389311.

admin/load.php
```diff
@@ -634,5 +663,69 @@ function perflab_plugin_admin_notices() {
...
 	}
+
+	$active_modules_with_plugins = perflab_get_active_modules_with_standalone_plugins();
+
+	if ( ! empty( $active_modules_with_plugins ) ) {
+		$module_data            = perflab_get_modules();
+		$available_modules_list = array();
+		foreach ( $active_modules_with_plugins as $module_slug ) {
+			if ( isset( $module_data[ $module_slug ] ) && ! perflab_is_standalone_plugin_loaded( $module_slug ) ) {
+				$available_modules_list[] = $module_data[ $module_slug ]['name'];
```
felixarntz: Maybe rename this variable to something like `$available_module_names`? When using just "modules", it could be assumed to contain the module slugs, which are typically used to identify a module. But here it's the user-facing module names.

admin/load.php
```diff
@@ -634,5 +663,69 @@ function perflab_plugin_admin_notices() {
...
+		$modules_count = count( $available_modules_list );
+		if ( $modules_count < 1 ) {
+			return;
+		}
+
+		if ( 1 === $modules_count ) {
+			$message  = '<p>';
+			$message .= sprintf(
+				/* translators: Module name */
+				esc_html__( 'Performance Lab has detected usage of "%s" module that have standalone plugin equivalents. Installing and activating a performance plugin will disable the legacy module equivalent. This will not impact functionality of the module.', 'performance-lab' ),
```
felixarntz:
```suggestion
				esc_html__( 'Your site is using the "%s" module which will be removed in the future in favor of its equivalent standalone plugin. Please click the following button to install and activate the relevant plugin in favor of the module. This will not impact any of the underlying functionality.', 'performance-lab' ),
```

admin/load.php
```diff
@@ -634,5 +663,69 @@ function perflab_plugin_admin_notices() {
...
+				esc_attr( $available_modules_list[0] )
+			);
+			$message .= '</p>';
+		} else {
+			$message  = '<p>' . esc_html__( 'Performance Lab has detected usage of modules that have standalone plugin equivalents. Please see the available plugins below. Installing and activating a performance plugin will disable the legacy module equivalent. This will not impact functionality of the module.', 'performance-lab' ) . '</p>';
```
felixarntz:
```suggestion
			$message  = '<p>' . esc_html__( 'Your site is using modules which will be removed in the future in favor of their equivalent standalone plugins. Please click the following button to install and activate the relevant plugins in favor of the modules. This will not impact any of the underlying functionality.', 'performance-lab' ) . '</p>';
```

admin/load.php
```diff
@@ -634,5 +663,69 @@ function perflab_plugin_admin_notices() {
...
+
+		echo '<div class="notice notice-warning is-dismissible">';
+			echo wp_kses_post( $message );
+			echo '<p class="perflab-button-wrapper">';
+				echo '<button type="button" class="button button-primary perflab-install-active-plugin">';
+					echo esc_html__( 'Migrate legacy modules to standalone plugins', 'performance-lab' );
+				echo '</button>';
+				echo '<span class="dashicons dashicons-update hidden"></span>';
+			echo '</p>';
+		echo '</div>';
```
felixarntz: As a follow up to my comment in https://github.com/WordPress/performance/pull/899#discussion_r1424435651, here's what I meant with directly printing HTML to get syntax highlighting:
```suggestion
		?>
		<div class="notice notice-warning is-dismissible">
			<?php echo wp_kses_post( $message ); ?>
			<p class="perflab-button-wrapper">
				<button type="button" class="button button-primary perflab-install-active-plugin">
					<?php echo esc_html__( 'Migrate legacy modules to standalone plugins', 'performance-lab' ); ?>
				</button>
				<span class="dashicons dashicons-update hidden"></span>
			</p>
		</div>
		<?php
```
Other than syntax highlighting, it's also a bit more appropriate in terms of indentation, since indenting PHP just because it outputs HTML is a bit odd.

admin/plugins.php
```diff
@@ -390,3 +426,82 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+		$plugin_slug = basename( $module_slug );
+		$api         = perflab_query_plugin_info( $plugin_slug );
+		$plugin_slug = isset( $get_standalone_plugins_file_map[ $plugin_slug ] ) ? $get_standalone_plugins_file_map[ $plugin_slug ] : '';
+
+		if ( ! $plugin_slug ) {
+			$status['errorMessage'] = esc_html__( 'Invalid plugin.', 'performance-lab' );
+			wp_send_json_error( $status );
+		}
+
+		// Check if the plugin is already installed.
```
felixarntz:
```suggestion
		// Install the plugin if it is not installed yet.
```

admin/plugins.php
```diff
@@ -390,3 +426,82 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+	foreach ( $plugins_to_activate as $module_slug ) {
+
+		// Skip checking for already activated plugin.
+		if ( perflab_is_standalone_plugin_loaded( $module_slug ) ) {
+			continue;
+		}
+
+		$plugin_slug = basename( $module_slug );
+		$api         = perflab_query_plugin_info( $plugin_slug );
+		$plugin_slug = isset( $get_standalone_plugins_file_map[ $plugin_slug ] ) ? $get_standalone_plugins_file_map[ $plugin_slug ] : '';
```
felixarntz: As a follow-up to https://github.com/WordPress/performance/pull/899#discussion_r1424442316, here's what I meant: We don't need to have a file map:
* `perflab_get_active_modules_with_standalone_plugins()` only returns modules with standalone plugins, so we don't need to check that somewhere else again.
* We can rely purely on `basename( $module_slug )` for the plugin slug as you're doing below. The plugin slug is needed only to _install_ the plugin.
* We can rely on `basename( $module_slug ) . '/load.php'` for the plugin basename. The plugin basename is needed to _activate_ the plugin.

All PL modules use a `load.php` file which means their standalone plugins which get deployed from here also use the `load.php` file.
```suggestion
	$plugins_to_activate = perflab_get_active_modules_with_standalone_plugins();
	$plugins             = get_plugins();
	$status              = array();

	foreach ( $plugins_to_activate as $module_slug ) {

		// Skip checking for already activated plugin.
		if ( perflab_is_standalone_plugin_loaded( $module_slug ) ) {
			continue;
		}

		$plugin_slug     = basename( $module_slug );
		$plugin_basename = $plugin_slug . '/load.php';
		$api             = perflab_query_plugin_info( $plugin_slug );
```

Then we can remove the `perflab_get_standalone_plugins_file_map()` function.

admin/plugins.php
```diff
@@ -390,3 +426,82 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+
+	foreach ( $plugins_to_activate as $module_slug ) {
+
+		// Skip checking for already activated plugin.
+		if ( perflab_is_standalone_plugin_loaded( $module_slug ) ) {
+			continue;
+		}
+
+		$plugin_slug = basename( $module_slug );
+		$api         = perflab_query_plugin_info( $plugin_slug );
```
felixarntz: We should check here that the API returned data. If `$api` is empty, we need to return with an error here that the querying the plugin information failed.

admin/plugins.php
```diff
@@ -390,3 +426,82 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+			$status['errorMessage'] = esc_html__( 'Invalid plugin.', 'performance-lab' );
+			wp_send_json_error( $status );
+		}
+
+		// Check if the plugin is already installed.
+		if ( ! isset( $plugins[ $plugin_slug ] ) ) {
+			$plugin_path = WP_PLUGIN_DIR . '/' . $plugin_slug;
+
+			// Check if the plugin file exists.
+			if ( ! file_exists( $plugin_path ) ) {
```
felixarntz: Why is this check needed? If the file doesn't exist, the plugin is by definition not installed, which you have already checked above in line 472. I think this check is redundant and should be removed.

admin/plugins.php
```diff
@@ -390,3 +426,82 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+					$status['errorMessage'] = $skin->result->get_error_message();
+					wp_send_json_error( $status );
+				} elseif ( $skin->get_errors()->has_errors() ) {
+					$status['errorMessage'] = $skin->get_error_messages();
+					wp_send_json_error( $status );
+				}
+			}
+		}
+
+		$result = activate_plugin( $plugin_slug );
```
felixarntz: See my other comment above, here you need to use the full path to the plugin main file, as documented for the `activate_plugin()` function.
```suggestion
		$result = activate_plugin( WP_PLUGIN_DIR . '/' . $plugin_basename );
```

admin/plugins.php
```diff
@@ -55,6 +55,37 @@ function perflab_get_standalone_plugins() {
...
+ * @since n.e.x.t
+ *
+ * @return string[]
+ */
+function perflab_get_standalone_plugins_file_map() {
+	return array(
+		'webp-uploads'          => 'webp-uploads/load.php',
+		'dominant-color-images' => 'dominant-color-images/load.php',
+	);
+}
```
felixarntz: See my other comments, I think this is not needed and can be removed.

admin/plugins.php
```diff
@@ -390,3 +426,82 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+ * Handles the standalone plugin install and activation via AJAX.
+ *
+ * @since n.e.x.t
+ */
+function perflab_install_activate_standalone_plugins_callback() {
+	if ( ! wp_verify_nonce( $_REQUEST['nonce'], 'perflab-install-activate-plugins' ) ) {
+		$status['errorMessage'] = esc_html__( 'Invalid nonce: Please refresh and try again.', 'performance-lab' );
+		wp_send_json_error( $status );
+	}
+
```
felixarntz: Here we'll need to also include a capability check to make sure that the current user can both install and activate plugins.

admin/plugins.php
```diff
@@ -390,3 +426,82 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+// WordPress AJAX action to handle the button click event.
+add_action( 'wp_ajax_perflab_install_activate_standalone_plugins', 'perflab_install_activate_standalone_plugins_callback' );
+
+/**
+ * Handles the standalone plugin install and activation via AJAX.
+ *
+ * @since n.e.x.t
+ */
+function perflab_install_activate_standalone_plugins_callback() {
```
felixarntz: Maybe move this function into `admin/load.php`? That's where the similar logic functions for activating and deactivating plugins live.

felixarntz (commented):
@mukeshpanchal27 I left some follow up feedback, mostly minor things to clarify wording. For example, I think we should include in the user-facing message that the modules will be removed in the future, for a sense of urgency.

There's one larger consideration here that we may want to change, which I'm just realizing now:
* You've implemented the migration logic in an AJAX request. While this works well, it comes with the problem that then the whole UI on the page may be outdated (e.g. after other plugins were installed, modules deactivated etc.).
* It will probably be too much effort to fix all the UI throughout the page to reflect those changes. Therefore I would suggest to go with a regular full page refresh and instead of AJAX move the logic to an `admin_action_{$action}` callback, similar to how the plugin activation and deactivation currently is.
* You should be able to copy the logic pretty much as is, just replace the `wp_send_json_error()` calls with equivalent `wp_die()`, and the `wp_send_json_success()` with a redirect back to the Performance Lab plugin screen.
* Doing that will also drastically simplify this PR: You can probably remove the JS file entirely, and the button can simply be a link to e.g. `wp-admin/plugins.php?action=...&_wpnonce=...`.
* Because this is then a full page reload, we don't have to worry about all other UI on the page, as it will be automatically refreshed after the changes have been made.

Let me know what you think about this, or if you have any follow up questions. For now, it's probably best if you just address the more specific feedback below, and we can discuss the above separately afterwards (e.g. in Slack or in PR comments).

admin/plugins.php
mukeshpanchal27: @felixarntz We already checked it on JS code. I don't think we needs to check it again. WDYT?

```
if ( ! perflab_module_migration_notice.has_permission ) {
   alert( perflab_module_migration_notice.permission_error );
   return;
}
```

admin/plugins.php
mukeshpanchal27: Make sense. Thanks

admin/plugins.php
felixarntz: @mukeshpanchal27 Checking permissions on the client isn't safe. It can be useful for sure, but anybody could make a request to the server manually (i.e. not based on that JS logic). So it has to always be checked on the server.

On the client-side it's a nice to have. I'd argue you could remove it from the JS code to make it a bit simpler.

mukeshpanchal27: Got it. Updated in https://github.com/WordPress/performance/pull/899/commits/abfc2f3b1976b24e987a1ef3dcfa6fcc59068ce5

admin/plugins.php
```diff
@@ -64,7 +81,8 @@ function perflab_render_plugins_ui() {
 	require_once ABSPATH . 'wp-admin/includes/plugin-install.php';
 	require_once ABSPATH . 'wp-admin/includes/plugin.php';
 
-	$standalone_plugins = array();
+	$standalone_plugins                         = array();
+	$get_active_modules_with_standalone_plugins = perflab_get_active_modules_with_standalone_plugins();
```
felixarntz: Please revert, as this isn't used anymore.
```suggestion
	$standalone_plugins = array();
```

admin/load.php
```diff
@@ -609,18 +642,97 @@ function perflab_deactivate_plugin() {
...
+			continue;
+		}
+
+		$plugin_slug     = basename( $module_slug );
+		$plugin_basename = $plugin_slug . '/load.php';
+		$api             = perflab_query_plugin_info( $plugin_slug );
+
+		// Return early if plugin API return an error.
+		if ( ! $api ) {
+			$status['errorMessage'] = esc_html__( 'An unexpected error occurred. Something may be wrong with WordPress.org or this servers configuration.', 'performance-lab' );
```
felixarntz:
```suggestion
			$status['errorMessage'] = __( 'An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration.', 'performance-lab' );
```

admin/perflab-module-migration-notice.js
```diff
@@ -0,0 +1,40 @@
+( function( document ) {
+	document.addEventListener( 'DOMContentLoaded', function() {
+		document.addEventListener( 'click', function(event) {
+			if ( event.target.classList.contains( 'perflab-install-active-plugin' ) ) {
+
+				if ( confirm( perflab_module_migration_notice.prompt_message ) ) {
```
felixarntz: I don't think we need the confirm prompt. This isn't a destructive action, and having a confirm prompt makes it sound more alarming than it needs to be - particularly since we _want_ users to perform this action.

This will furthermore allow us to remove the `prompt_message` data.

felixarntz (commented):
@mukeshpanchal27 I just tested this again, and I think it looks almost ready to merge.

Regarding my feedback in https://github.com/WordPress/performance/pull/899#pullrequestreview-1780241614, I just now realized that you have a `window.location.reload()` in the JS, so that works perfectly fine then, and the AJAX with loading spinner still makes for a bit nicer user experience. 🎉 

My final bits of feedback are below, after that it should be good to merge from my perspective.

admin/load.php
```diff
@@ -634,5 +746,71 @@ function perflab_plugin_admin_notices() {
...
+			return;
+		}
+
+		if ( 1 === $modules_count ) {
+			$message  = '<p>';
+			$message .= sprintf(
+				/* translators: Module name */
+				esc_html__( 'Your site is using the "%s" module which will be removed in the future in favor of its equivalent standalone plugin. Please click the following button to install and activate the relevant plugin in favor of the module. This will not impact any of the underlying functionality.', 'performance-lab' ),
+				esc_attr( $available_module_names[0] )
+			);
```
felixarntz: Maybe split with a line break to avoid long lines on large screens and make this easier to read.
```suggestion
			$message .= sprintf(
				/* translators: Module name */
				esc_html__( 'Your site is using the "%s" module which will be removed in the future in favor of its equivalent standalone plugin.', 'performance-lab' ),
				esc_attr( $available_module_names[0] )
			);
			$message .= '<br>';
			$message .= esc_html__( 'Please click the following button to install and activate the relevant plugin in favor of the module. This will not impact any of the underlying functionality.', 'performance-lab' );
```

admin/load.php
```diff
@@ -634,5 +746,71 @@ function perflab_plugin_admin_notices() {
...
+			$message .= '</p>';
+		} else {
+			$message  = '<p>' . esc_html__( 'Your site is using modules which will be removed in the future in favor of their equivalent standalone plugins. Please click the following button to install and activate the relevant plugins in favor of the modules. This will not impact any of the underlying functionality.', 'performance-lab' ) . '</p>';
```
felixarntz: See above:
```suggestion
			$message  = '<p>';
			$message .= esc_html__( 'Your site is using modules which will be removed in the future in favor of their equivalent standalone plugins.', 'performance-lab' );
			$message .= '<br>';
			$message .= esc_html__( 'Please click the following button to install and activate the relevant plugins in favor of the modules. This will not impact any of the underlying functionality.', 'performance-lab' );
			$message .= '</p>';
```

mukeshpanchal27:
> Regarding my feedback in [#899 (review)](https://github.com/WordPress/performance/pull/899#pullrequestreview-1780241614), I just now realized that you have a `window.location.reload()` in the JS, so that works perfectly fine then, and the AJAX with loading spinner still makes for a bit nicer user experience. 🎉

Thanks @felixarntz. In https://github.com/WordPress/performance/pull/899/commits/388be6aa4ae4c2f2d4dd3061a85698ba03fdd8bd i introduce `perflab_replace_html_entity` function that change html entriey as in alert it show `server&#8217;s` instead of `server's`

felixarntz (approved):
Looks great, thanks @mukeshpanchal27!

admin/load.php
```diff
@@ -531,6 +537,32 @@ function perflab_enqueue_modules_page_scripts() {
...
+		)
+	);
+
+	wp_localize_script(
+		'perflab-module-migration-notice',
+		'perflab_module_migration_notice',
+		array(
+			'ajaxurl'       => admin_url( 'admin-ajax.php' ),
+			'nonce'         => wp_create_nonce( 'perflab-install-activate-plugins' ),
+			'network_error' => esc_html__( 'Network response was not ok.', 'performance-lab' ),
```
westonruter: Since the string is not being printed here, it should not get escaped. The `wp_localize_script()` function will do the escaping.

```suggestion
			'network_error' => __( 'Network response was not ok.', 'performance-lab' ),
```

admin/load.php
```diff
@@ -609,18 +641,97 @@ function perflab_deactivate_plugin() {
...
  * @since n.e.x.t
- *
- * @return void
  */
-function perflab_plugin_admin_notices() {
-	if ( 'settings_page_perflab-modules' !== get_current_screen()->id ) {
-		return;
+function perflab_install_activate_standalone_plugins_callback() {
+	if ( ! wp_verify_nonce( $_REQUEST['nonce'], 'perflab-install-activate-plugins' ) ) {
+		$status['errorMessage'] = esc_html__( 'Invalid nonce: Please refresh and try again.', 'performance-lab' );
```
westonruter: Again, escaping will be handled by `wp_json_encode()`.

```suggestion
		$status['errorMessage'] = __( 'Invalid nonce: Please refresh and try again.', 'performance-lab' );
```

admin/load.php
```diff
@@ -609,18 +641,97 @@ function perflab_deactivate_plugin() {
...
+		wp_send_json_error( $status );
+	}
+
+	if ( ! current_user_can( 'install_plugins' ) || ! current_user_can( 'activate_plugins' ) ) {
+		$status['errorMessage'] = esc_html__( 'Sorry, you are not allowed to manage plugins for this site. Please contact the administrator.', 'performance-lab' );
```
westonruter: Ditto

```suggestion
		$status['errorMessage'] = __( 'Sorry, you are not allowed to manage plugins for this site. Please contact the administrator.', 'performance-lab' );
```

admin/load.php
```diff
@@ -609,18 +641,97 @@ function perflab_deactivate_plugin() {
...
+		$api             = perflab_query_plugin_info( $plugin_slug );
+
+		// Return early if plugin API return an error.
+		if ( ! $api ) {
+			$status['errorMessage'] = __( 'An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration.', 'performance-lab' );
+			wp_send_json_error( $status );
+		}
+
+		if ( ! $plugin_slug ) {
+			$status['errorMessage'] = esc_html__( 'Invalid plugin.', 'performance-lab' );
```
westonruter: Ditto. And note in previous `if` statement that `esc_html__()` is not used.

```suggestion
			$status['errorMessage'] = __( 'Invalid plugin.', 'performance-lab' );
```

admin/load.php
```diff
@@ -609,18 +641,97 @@ function perflab_deactivate_plugin() {
...
+			continue;
+		}
+
+		$plugin_slug     = basename( $module_slug );
+		$plugin_basename = $plugin_slug . '/load.php';
+		$api             = perflab_query_plugin_info( $plugin_slug );
+
+		// Return early if plugin API return an error.
+		if ( ! $api ) {
+			$status['errorMessage'] = __( 'An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration.', 'performance-lab' );
```
westonruter: Would it not be better to not use an HTML entity in this case? Then you wouldn't have to decode it later with `perflab_replace_html_entity()` in JS. But you or @swissspidy would know better in regards to i18n best practices. 

```suggestion
			$status['errorMessage'] = __( 'An unexpected error occurred. Something may be wrong with WordPress.org or this server’s configuration.', 'performance-lab' );
```

admin/perflab-module-migration-notice.js
```diff
@@ -0,0 +1,61 @@
...
+					} );
+			}
+		} );
+
+		// Function to replace HTML entities with their corresponding characters.
+		function perflab_replace_html_entity( str ) {
+			return str.replace( /&#(\d+);/g, function ( match, dec ) {
+				return String.fromCharCode( dec );
+			} );
+		}
```
westonruter: See above. No need to send HTML entities in JSON. But @swissspidy may overrule me of course 😄 

```suggestion
								result.data.errorMessage
							);
						}
						window.location.reload();
					} )
					.catch( function ( error ) {
						alert( error.errorMessage );
						window.location.reload();
					} );
			}
		} );
```

admin/load.php
westonruter: Otherwise, the decoding of the HTML entities could be done right here in PHP:

```suggestion
			$status['errorMessage'] = html_entity_decode( __( 'An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration.', 'performance-lab' ), ENT_QUOTES );
```

admin/load.php
```diff
@@ -609,18 +641,97 @@ function perflab_deactivate_plugin() {
...
+			} elseif ( is_wp_error( $skin->result ) ) {
+				$status['errorMessage'] = $skin->result->get_error_message();
+				wp_send_json_error( $status );
+			} elseif ( $skin->get_errors()->has_errors() ) {
+				$status['errorMessage'] = $skin->get_error_messages();
+				wp_send_json_error( $status );
+			}
+		}
+
+		$result = activate_plugin( WP_PLUGIN_DIR . '/' . $plugin_basename );
```
westonruter: What if the plugin isn't located in `WP_PLUGIN_DIR` (even though this may be unlikely)? Also, the phpdoc for `activate_plugin()` says that this arg is _relative_ not absolute.

admin/load.php
```diff
@@ -609,18 +641,97 @@ function perflab_deactivate_plugin() {
...
+	}
+
+	if ( ! current_user_can( 'install_plugins' ) || ! current_user_can( 'activate_plugins' ) ) {
+		$status['errorMessage'] = esc_html__( 'Sorry, you are not allowed to manage plugins for this site. Please contact the administrator.', 'performance-lab' );
+		wp_send_json_error( $status );
+	}
+
+	if ( ! function_exists( 'plugins_api' ) ) {
+		require_once ABSPATH . 'wp-admin/includes/plugin-install.php';
+	}
```
westonruter: Couldn't this just do the same as the next two `require_once` calls?

```suggestion
	require_once ABSPATH . 'wp-admin/includes/plugin-install.php';
```

admin/load.php
```diff
@@ -634,5 +745,77 @@ function perflab_plugin_admin_notices() {
...
+		}
+
+		if ( 1 === $modules_count ) {
+			$message  = '<p>';
+			$message .= sprintf(
+				/* translators: Module name */
+				esc_html__( 'Your site is using the "%s" module which will be removed in the future in favor of its equivalent standalone plugin.', 'performance-lab' ),
+				esc_attr( $available_module_names[0] )
+			);
+			$message .= '<br>';
```
westonruter: Is the manual line break needed? On smaller screens, this could result in some unsightly line wrapping. If the issue is that the string is too wide, then a `max-width` on the container would be better.

```suggestion
			$message .= ' ';
```

admin/load.php
```diff
@@ -634,5 +745,77 @@ function perflab_plugin_admin_notices() {
...
+			$message .= esc_html__( 'Please click the following button to install and activate the relevant plugin in favor of the module. This will not impact any of the underlying functionality.', 'performance-lab' );
+			$message .= '</p>';
+		} else {
+			$message  = '<p>';
+			$message .= esc_html__( 'Your site is using modules which will be removed in the future in favor of their equivalent standalone plugins.', 'performance-lab' );
+			$message .= '<br>';
```
```suggestion
			$message .= ' ';
```

admin/load.php
```diff
@@ -634,5 +745,77 @@ function perflab_plugin_admin_notices() {
...
+			$message .= esc_html__( 'Please click the following button to install and activate the relevant plugins in favor of the modules. This will not impact any of the underlying functionality.', 'performance-lab' );
+			$message .= '</p>';
+			$message .= '<strong>' . esc_html__( 'Available standalone plugins:', 'performance-lab' ) . '</strong>';
+			$message .= '<ol>';
+			foreach ( $available_module_names as $names ) {
+				$message .= sprintf( '<li>%s</li>', esc_html( $names ) );
```
```suggestion
			foreach ( $available_module_names as $module_name ) {
				$message .= sprintf( '<li>%s</li>', esc_html( $module_name ) );
```

westonruter (dismissed):

admin/load.php
swissspidy: Well I‘d normally recommend using the wp.i18n functions in JS instead of passing strings from PHP to JS. Then you don‘t really have to deal with that.

admin/perflab-module-migration-notice.js
westonruter: Otherwise, if client-side unescaping is really needed, then I think there's a better approach that will also handle named entities and hex entities: https://stackoverflow.com/a/42182294/93579

admin/load.php
swissspidy: Usually I'd recommend adding `wp-i18n` as a script dependency and then use `wp.i18n.__` in the JS file instead of using `wp_localize_script` for localization.

Not a big deal though because this is just 1 string, but something to keep in mind

admin/plugins.php
```diff
@@ -55,6 +55,23 @@ function perflab_get_standalone_plugins() {
...
+ *
+ * @since n.e.x.t
+ *
+ * @return string[]
+ */
+function perflab_get_active_modules_with_standalone_plugins() {
+	$modules = perflab_get_module_settings();
+	return array_filter(
+		array_keys( perflab_get_standalone_plugins_constants() ),
+		static function ( $v ) use ( $modules ) {
```
swissspidy: Maybe rename `$v` to `$module` for more clarity

admin/load.php
mukeshpanchal27: The `wp-i18n` has been loaded in the setting scree, so I included the dependency for it and utilized it in JavaScript. Consequently, I removed `network_error` from the PHP.

admin/load.php
mukeshpanchal27: Fantastic catch! I completely overlooked that aspect. Thank you for pointing it out. I've made the necessary updates, and you can find them in the commit here: https://github.com/WordPress/performance/pull/899/commits/2bd591d1eb3efd7172e1c356cc250df37616ec5a.

admin/load.php
mukeshpanchal27: Thanks both for the feedback. As we have different  messages for error in JS do we need to add them conditionally then use `wp.i18n.__( 'Error message', 'performance-lab' )` function to print them in JS?

swissspidy: I see, you're sending them to JS with `wp_send_json_*`. That should be fine to keep as-is 👍

mukeshpanchal27: Just to know do we need to address this and https://github.com/WordPress/performance/pull/899#discussion_r1428388934 feedbacks? Thanks!

swissspidy: Yes, Weston's feedback. Just my comment above can be ignored :)

I suggest using the `html_entity_decode` approach suggested above.

mukeshpanchal27: Thanks. Addressed in https://github.com/WordPress/performance/pull/899/commits/3e0256c45bdcc02fc05e5563b8d4d4f6c571fa9a

swissspidy (approved):

Issue #651, opened 2023-02-16
Update PL settings screen with new UI to manage standalone plugins and built-in modules


felixarntz:
## Feature Description

Following https://github.com/WordPress/performance/issues/618#issuecomment-1432785852, this is the first of a few issues to work towards removing the published standalone plugins from the Performance Lab plugin and, as part of it, refactoring the plugin's settings screen to focus on managing those plugins as well as enabling a smooth migration for existing users of the plugin.

All this will eventually need to be published in a single release 3.0.0, which should bump the major version number due to that significant change in user-facing behavior.

It is also worth highlighting that the implementation of these additional issues should only start once the previous milestone of publishing standalone plugins is completed (i.e. after #635 #636 #637 #638 #639 #640).

This issue, likely the most time consuming one, is focused on modifying the UI of the settings screen:
* Today it shows the list of modules, with checkboxes for which one to enable.
* Going forward, it should be broken in two sections, one for managing the standalone plugins that are performance related core feature projects, and another one for managing the remaining built-in modules (likely only Site Health).
    * This will likely also involve changing the UI controls available, as installing / activating / deactivating plugins will require a more explicit action than just toggling a checkbox.
    * It will also involve including additional capability checks, since not everyone that can access this screen can install or activate plugins.

### Requirements

* TODO.

10upsimon:
@felixarntz @joemcgill I've picked this up again. Based on my notes and cross referencing [this comment in an associated Asana task](https://app.asana.com/0/0/1204439792047518/1204462865730138/f), I mocked up some screens to re-visit the conversation so that I have some pointers to work on. Based on my understanding, we needed UI mocks for:

**1) A notice to render in the admin, likely on plugin install/update. I assume a notice of this sorts (ignore the wording, just a mock) with a CTA to the performance settings page is what we are after:**

![Plugins ‹ performance — WordPress 2023-07-20 at 3 49 59 PM](https://github.com/WordPress/performance/assets/72872375/265a4989-f753-40a8-9c46-8f0e3084acf1)

**2) On the performance settings screen, a notice similar to the above that still indicates usage of legacy modules and encouragement to update, but ALSO a list of standalone plugins. If a certain plugin is detected as having the legacy module equivalent used, mark is such, in my case not so subtly (this is just an over exaggerated example), and continue to list available modules below:**

[...]

PR #900, opened 2023-12-12
Enhance experience when installing and activating a standalone plugin

### LINKED ISSUE #880: Enhance experience when installing _and_ activating a standalone plugin
This is an enhancement for the experience of installing a standalone plugin via _Settings > Performance_ and then immediately activating it, i.e. without refreshing the current screen. See https://github.com/WordPress/performance/pull/864#pullrequestreview-1690280731:

> When installing one of the plugins and then activating (without reloading the page), it will end up on the regular Plugins screen rather than redirecting back to the Performance Lab screen. This is happening because the WordPress updates JS code doesn't set the Performance Lab specific activation URL on the button, but the regular WordPress one. This is not a dealbreaker, but if we can fix it somehow that would be great (e.g. is there some event listener we could use in JS to replace the URL with our own?).

This is not a major problem, and if we cannot resolve it without rewriting/copying a lot of WordPress core's `update.js` logic, we should probably not go above and beyond to fix it. But if there's a reasonable path to implement it (e.g. maybe an event listener where we dynamically replace the activation `href` attribute to use the `perflab_activate_plugin` action), it's worth improving the user experience.

mukeshpanchal27:
## Summary

Fixes #880

- When #899 will merge we will move the custom JS code in `admin/perflab-admin.js` file.
- `update.js` updates plugin activation link in 1 sec so in our custom script we will update it after 1.5 sec.

## Checklist

### DIFF
--- a/admin/load.php
+++ b/admin/load.php
@@ -551,6 +551,7 @@ function perflab_activate_plugin() {
 
 	check_admin_referer( "perflab_activate_plugin_{$plugin_slug}" );
 
+	// Since we know the plugin folder name, for efficiency we can simply look for the plugin main file there.
 	$plugins = get_plugins( '/' . $plugin_slug );
 
 	if ( empty( $plugins ) ) {
@@ -664,10 +665,10 @@ function perflab_print_plugin_activation_script() {
 				return;
 			}
 			/*
-				* WordPress core uses a 1s timeout for updating the activation link,
-				* so we set a 1.5 timeout here to ensure our changes get updated after
-				* the core changes have taken place.
-				*/
+			* WordPress core uses a 1s timeout for updating the activation link,
+			* so we set a 1.5 timeout here to ensure our changes get updated after
+			* the core changes have taken place.
+			*/
 			setTimeout( function() {
 				var plugin_url = target_element.attr( 'href' );
 				if ( ! plugin_url ) {
--- a/admin/load.php
+++ b/admin/load.php
@@ -670,7 +670,7 @@ function perflab_print_plugin_activation_script() {
 				return;
 			}
 
-			var target_element = $( `.wpp-standalone-plugins a[data-slug="${slug}"]` );
+			var target_element = $( '.wpp-standalone-plugins a[data-slug="' + slug + '"]' );
 			if ( ! target_element ) {
 				return;
 			}
--- a/admin/load.php
+++ b/admin/load.php
@@ -57,7 +57,7 @@ function perflab_load_modules_page( $modules = null, $focus_areas = null ) {
 	add_action( 'admin_enqueue_scripts', 'perflab_enqueue_modules_page_scripts' );
 
 	// Handle script for settings page.
-	add_action( 'admin_footer', 'perflab_enqueue_custom_script' );
+	add_action( 'admin_footer', 'perflab_print_plugin_activation_script' );
 
 	// Register sections for all focus areas, plus 'Other'.
 	if ( ! is_array( $focus_areas ) ) {
@@ -641,31 +641,38 @@ function perflab_plugin_admin_notices() {
 add_action( 'admin_notices', 'perflab_plugin_admin_notices' );
 
 /**
- * Callback function to handle admin custom script.
+ * Callback function that print plugin activation script.
  *
  * @since n.e.x.t
  */
-function perflab_enqueue_custom_script() {
+function perflab_print_plugin_activation_script() {
+	ob_start();
 	?>
-<script>
-( function( $ ) {
-	if ( window.pagenow === 'settings_page_perflab-modules' ) {
-		// Only listen for ajaxComplete if this is the performance setting page.
+	( function( $ ) {
 		$( document ).ajaxComplete( function( event, xhr, settings ) {
 			// Check if this is the 'install-plugin' request.
 			if ( settings.data && typeof settings.data === 'string' && settings.data.includes( 'action=install-plugin' ) ) {
+				var target_element = $( event.target.activeElement );
+				// Bail early if the element is not a[href].
+				if ( ! target_element.is( 'a[href]' ) ) {
+					return;
+				}
+				/*
+				 * WordPress core uses a 1s timeout for updating the activation link,
+				 * so we set a 1.5 timeout here to ensure our changes get updated after
+				 * the core changes have taken place.
+				 */
 				setTimeout( function() {
-					var plugin_url = $( event.target.activeElement ).attr( 'href' );
-					var nonce = $( event.target.activeElement ).parents( '.plugin-card' ).attr( 'data-plugin' );
+					var plugin_url = target_element.attr( 'href' );
+					var nonce = target_element.attr( 'data-plugin-activation-nonce' );
 					var url = new URL( plugin_url );
 					url.searchParams.set( 'action', 'perflab_activate_plugin' );
 					url.searchParams.set( '_wpnonce', nonce );
-					$( event.target.activeElement ).attr( 'href', url.href );
+					target_element.attr( 'href', url.href );
 				}, 1500 );
 			}
 		} );
-	}
-} )( jQuery );
-</script>
+	} )( jQuery );
 	<?php
+	wp_print_inline_script_tag( ob_get_clean() );
 }
--- a/admin/load.php
+++ b/admin/load.php
@@ -653,8 +653,7 @@ function perflab_print_plugin_activation_script() {
 			// Check if this is the 'install-plugin' request.
 			if ( settings.data && typeof settings.data === 'string' && settings.data.includes( 'action=install-plugin' ) ) {
 				var target_element = $( event.target.activeElement );
-				// Bail early if the element is not a[href].
-				if ( ! target_element.is( 'a[href]' ) ) {
+				if ( ! target_element ) {
 					return;
 				}
 				/*
@@ -664,6 +663,9 @@ function perflab_print_plugin_activation_script() {
 				 */
 				setTimeout( function() {
 					var plugin_url = target_element.attr( 'href' );
+					if ( ! plugin_url ) {
+						return;
+					}
 					var nonce = target_element.attr( 'data-plugin-activation-nonce' );
 					var url = new URL( plugin_url );
 					url.searchParams.set( 'action', 'perflab_activate_plugin' );
--- a/admin/load.php
+++ b/admin/load.php
@@ -662,25 +662,15 @@ function perflab_print_plugin_activation_script() {
 	$( document ).ajaxComplete( function( event, xhr, settings ) {
 		// Check if this is the 'install-plugin' request.
 		if ( settings.data && typeof settings.data === 'string' && settings.data.includes( 'action=install-plugin' ) ) {
-			// Split the data string by '&' to separate parameters.
-			var params = settings.data.split( '&' );
-
-			// Loop through the parameters to find the 'slug' parameter.
-			var slug = '';
-			for ( var i = 0; i < params.length; i++ ) {
-				var param = params[ i ].split( '=' );
-				if ( param[0] === 'slug') {
-					slug = param[1];
-					break;
-				}
-			}
+			var params = new URLSearchParams( settings.data );
+			var slug = params.get('slug');
 
 			// Check if 'slug' was found and output the value.
 			if ( ! slug ) {
 				return;
 			}
 
-			var target_element = $( '.wpp-standalone-plugins a[data-slug=' + slug + ']' );
+			var target_element = $( `.wpp-standalone-plugins a[data-slug="${slug}"]` );
 			if ( ! target_element ) {
 				return;
 			}
--- a/admin/load.php
+++ b/admin/load.php
@@ -56,6 +56,9 @@ function perflab_load_modules_page( $modules = null, $focus_areas = null ) {
 	// Handle script enqueuing for settings page.
 	add_action( 'admin_enqueue_scripts', 'perflab_enqueue_modules_page_scripts' );
 
+	// Handle script for settings page.
+	add_action( 'admin_footer', 'perflab_enqueue_custom_script' );
+
 	// Register sections for all focus areas, plus 'Other'.
 	if ( ! is_array( $focus_areas ) ) {
 		$focus_areas = perflab_get_focus_areas();
@@ -636,3 +639,33 @@ function perflab_plugin_admin_notices() {
 	}
 }
 add_action( 'admin_notices', 'perflab_plugin_admin_notices' );
+
+/**
+ * Callback function to handle admin custom script.
+ *
+ * @since n.e.x.t
+ */
+function perflab_enqueue_custom_script() {
+	?>
+<script>
+( function( $ ) {
+	if ( window.pagenow === 'settings_page_perflab-modules' ) {
+		// Only listen for ajaxComplete if this is the performance setting page.
+		$( document ).ajaxComplete( function( event, xhr, settings ) {
+			// Check if this is the 'install-plugin' request.
+			if ( settings.data && typeof settings.data === 'string' && settings.data.includes( 'action=install-plugin' ) ) {
+				setTimeout( function() {
+					var plugin_url = $( event.target.activeElement ).attr( 'href' );
+					var nonce = $( event.target.activeElement ).parents( '.plugin-card' ).attr( 'data-plugin' );
+					var url = new URL( plugin_url );
+					url.searchParams.set( 'action', 'perflab_activate_plugin' );
+					url.searchParams.set( '_wpnonce', nonce );
+					$( event.target.activeElement ).attr( 'href', url.href );
+				}, 1500 );
+			}
+		} );
+	}
+} )( jQuery );
+</script>
+	<?php
+}
--- a/admin/load.php
+++ b/admin/load.php
@@ -59,7 +59,7 @@ function perflab_load_modules_page( $modules = null, $focus_areas = null ) {
 	// Handle style for settings page.
 	add_action( 'admin_head', 'perflab_print_modules_page_style' );
 
-  // Handle script for settings page.
+	// Handle script for settings page.
 	add_action( 'admin_footer', 'perflab_print_plugin_activation_script' );
 
 	// Handle admin notices for settings page.
--- a/admin/load.php
+++ b/admin/load.php
@@ -543,29 +543,34 @@ function perflab_enqueue_modules_page_scripts() {
  */
 function perflab_activate_plugin() {
 	// Do not proceed if plugin query arg is not present.
-	if ( empty( $_GET['plugin_slug'] ) ) {
+	if ( empty( $_GET['plugin'] ) ) {
 		return;
 	}
 
-	$plugin_slug = sanitize_text_field( wp_unslash( $_GET['plugin_slug'] ) );
+	$plugin = sanitize_text_field( wp_unslash( $_GET['plugin'] ) );
 
-	check_admin_referer( "perflab_activate_plugin_{$plugin_slug}" );
+	// Check whether "plugin" is a plugin slug or a plugin basename.
+	if ( str_contains( $plugin, '/' ) ) {
+		check_admin_referer( "perflab_activate_plugin_{$plugin}" );
+	} else {
+		check_admin_referer( "perflab_activate_plugin_{$plugin}" );
 
-	$plugins = get_plugins( '/' . $plugin_slug );
+		$plugins = get_plugins( '/' . $plugin );
 
-	if ( empty( $plugins ) ) {
-		return;
+		if ( empty( $plugins ) ) {
+			wp_die( esc_html__( 'Plugin not found.', 'default' ) );
+		}
+
+		$plugin_file_names = array_keys( $plugins );
+		$plugin            = $plugin . '/' . $plugin_file_names[0];
 	}
 
-	$keys        = array_keys( $plugins );
-	$plugin_file = $plugin_slug . '/' . $keys[0];
-
-	if ( ! current_user_can( 'activate_plugin', $plugin_file ) ) {
+	if ( ! current_user_can( 'activate_plugin', $plugin ) ) {
 		wp_die( esc_html__( 'Sorry, you are not allowed to activate this plugin.', 'default' ) );
 	}
 
 	// Activate the plugin in question and return to prior screen.
-	$do_plugin_activation = activate_plugins( $plugin_file );
+	$do_plugin_activation = activate_plugins( $plugin );
 	$referer              = wp_get_referer();
 	if ( ! is_wp_error( $do_plugin_activation ) ) {
 		$referer = add_query_arg(
@@ -664,10 +669,10 @@ function perflab_print_plugin_activation_script() {
 				return;
 			}
 			/*
-				* WordPress core uses a 1s timeout for updating the activation link,
-				* so we set a 1.5 timeout here to ensure our changes get updated after
-				* the core changes have taken place.
-				*/
+			 * WordPress core uses a 1s timeout for updating the activation link,
+			 * so we set a 1.5 timeout here to ensure our changes get updated after
+			 * the core changes have taken place.
+			 */
 			setTimeout( function() {
 				var plugin_url = target_element.attr( 'href' );
 				if ( ! plugin_url ) {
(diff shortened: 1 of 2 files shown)


admin/load.php
```diff
@@ -636,3 +639,33 @@ function perflab_plugin_admin_notices() {
...
+/**
+ * Callback function to handle admin custom script.
+ *
+ * @since n.e.x.t
+ */
+function perflab_enqueue_custom_script() {
+	?>
+<script>
+( function( $ ) {
+	if ( window.pagenow === 'settings_page_perflab-modules' ) {
```
swissspidy: Can we do this check in PHP instead? No need to add JS to pages that aren't ours.

admin/plugins.php
```diff
@@ -261,8 +275,11 @@ function perflab_render_plugin_card( array $plugin_data ) {
 	$action_links = apply_filters( 'plugin_install_action_links', $action_links, $plugin_data );
 
 	$last_updated_timestamp = strtotime( $plugin_data['last_updated'] );
+
+	$standalone_plugins_file_map = perflab_get_standalone_plugins_file_map();
+	$plugin_file                 = isset( $standalone_plugins_file_map[ $plugin_data['slug'] ] ) ? wp_create_nonce( 'perflab_activate_plugin_' . $standalone_plugins_file_map[ $plugin_data['slug'] ] ) : '';
 	?>
-	<div class="plugin-card plugin-card-<?php echo sanitize_html_class( $plugin_data['slug'] ); ?>">
+	<div class="plugin-card plugin-card-<?php echo sanitize_html_class( $plugin_data['slug'] ); ?>" data-module="<?php echo esc_attr( $plugin_data['slug'] ); ?>" data-plugin="<?php echo esc_attr( $plugin_file ); ?>">
```
felixarntz: Two points of feedback here:
* The `data-module` isn't used as far as I can see.
* Couldn't we put these attributes on the Install link `a` tag? That would be closer to where it applies and would allow us to avoid unnecessary parent lookups in the jQuery code.

admin/plugins.php
```diff
@@ -55,6 +55,20 @@ function perflab_get_standalone_plugins() {
...
+ * @since n.e.x.t
+ *
+ * @return string[]
+ */
+function perflab_get_standalone_plugins_file_map() {
+	return array(
+		'webp-uploads'          => 'webp-uploads/load.php',
+		'dominant-color-images' => 'dominant-color-images/load.php',
+	);
+}
```
felixarntz: I don't think this function is needed. We could probably just use `{slug}/load.php`?

Alternatively we could add support to activate just via the slug to the AJAX PHP functions. That way we wouldn't need to guess the plugin main file name at all, and it could be detected properly via the PHP function after the plugin was already installed.

The latter is maybe a bit too much effort for the benefit, so maybe we just go with the assumption of `{slug}/load.php`?

admin/load.php
```diff
@@ -636,3 +639,33 @@ function perflab_plugin_admin_notices() {
...
+		// Only listen for ajaxComplete if this is the performance setting page.
+		$( document ).ajaxComplete( function( event, xhr, settings ) {
+			// Check if this is the 'install-plugin' request.
+			if ( settings.data && typeof settings.data === 'string' && settings.data.includes( 'action=install-plugin' ) ) {
+				setTimeout( function() {
+					var plugin_url = $( event.target.activeElement ).attr( 'href' );
```
felixarntz: Just to be safe, we should probably add a check here for if there's actually a value. It could be that the target element is _not_ the `a[href]` tag, in which case we need to bail early and do nothing.

admin/load.php
```diff
@@ -636,3 +639,33 @@ function perflab_plugin_admin_notices() {
...
+					var nonce = $( event.target.activeElement ).parents( '.plugin-card' ).attr( 'data-plugin' );
```
felixarntz: As the attribute contains a nonce, `data-plugin` isn't a good name really. How about `data-activation-nonce` or `data-plugin-activation-nonce`?

Also see my other comment on trying to avoid the `parents()` lookup here.

admin/load.php
```diff
@@ -636,3 +639,33 @@ function perflab_plugin_admin_notices() {
...
+					var url = new URL( plugin_url );
+					url.searchParams.set( 'action', 'perflab_activate_plugin' );
+					url.searchParams.set( '_wpnonce', nonce );
+					$( event.target.activeElement ).attr( 'href', url.href );
+				}, 1500 );
```
felixarntz: Can you add a comment here to explain why there is the 1.5s timeout?

felixarntz (changes requested):
Thanks @mukeshpanchal27, it's great you found a solution for this problem. I left a bit of feedback below, but the overall approach looks solid.

admin/plugins.php
swissspidy: I think the WP updates.js script expects `data-plugin` here and not on the link.

admin/load.php
```diff
@@ -636,3 +639,33 @@ function perflab_plugin_admin_notices() {
...
 add_action( 'admin_notices', 'perflab_plugin_admin_notices' );
+
+/**
+ * Callback function to handle admin custom script.
+ *
+ * @since n.e.x.t
+ */
+function perflab_enqueue_custom_script() {
+	?>
+<script>
```
westonruter: Ideally this would use `wp_print_inline_script_tag()`

admin/plugins.php
mukeshpanchal27: When initially working on another PR, I considered using `{slug}/load.php`. However, upon examining the main file for the 'Performant Translations' plugin, I found it as `performant-translations/performant-translations.php`. That's why I included the mapping function. Happy to update if it safe?

swissspidy: Performant Translations is not an existing module, so is that even relevant here?

mukeshpanchal27: Nope. But in future if someone add new module with different file name then it create bug. `Performant Translations` is just example.

admin/plugins.php
mukeshpanchal27: Thanks for the feedback. The custom attribute is moved to href.

admin/load.php
mukeshpanchal27: Thanks for raising. Added check for a href in https://github.com/WordPress/performance/pull/900/commits/35a69fcc2831d5f4ac0e1e69ece748691c8ab5d2

admin/plugins.php
swissspidy: > Nope. But in future if someone add new module with different file name then it create bug.

Do you mean a new module or a new standalone plugin? For modules `load.php` is the standard, so we wouldn't use a different file name there. And for standalone plugins this function here seems not relevant? Or what am I missing?

felixarntz: @mukeshpanchal27 @swissspidy Actually I think for this functionality Performant Translations _does_ matter. This PR is about installing and activating a standalone plugin, regardless of whether or not it is for an existing module. Performant Translations is a plugin that we want to link to here, so that needs to be covered. This is different in #899, which is only about plugins that _have_ an equivalent module.

That said, I still don't think we should need this function. How about the following instead:
* We enhance the logic for the `perflab_activate_plugin` action.
* Normally it is passed the plugin basename (e.g. `webp-uploads/load.php`). We could add support for alternatively receiving just the plugin slug (e.g. `webp-uploads`).
* Since at that point the plugin would already have been installed, we can simply look up the full basename. For example, call `get_plugins()` and find the one that is `webp-uploads/{mainfile.php}`. Then we have the basename dynamically, without a need to hardcode it.
* It also simplifies the behavior on the client, basically we don't need to know the basename on the client in JS, as the `perflab_activate_plugin` action on the server just figures it out automatically.

admin/plugins.php
felixarntz: I thought this was just a custom data attribute. @swissspidy What is the purpose of `data-plugin` on the card, and is that relevant here or not?

admin/load.php
```diff
@@ -636,3 +639,40 @@ function perflab_plugin_admin_notices() {
...
+				if ( ! target_element.is( 'a[href]' ) ) {
+					return;
+				}
+				/*
+				 * WordPress core uses a 1s timeout for updating the activation link,
+				 * so we set a 1.5 timeout here to ensure our changes get updated after
+				 * the core changes have taken place.
+				 */
+				setTimeout( function() {
+					var plugin_url = target_element.attr( 'href' );
```
felixarntz: That check here is not what I meant in https://github.com/WordPress/performance/pull/900#discussion_r1424303670. It's safer to check that `plugin_url` is actually set. For example:
```suggestion
				if ( ! target_element ) {
					return;
				}
				/*
				 * WordPress core uses a 1s timeout for updating the activation link,
				 * so we set a 1.5 timeout here to ensure our changes get updated after
				 * the core changes have taken place.
				 */
				setTimeout( function() {
					var plugin_url = target_element.attr( 'href' );
					if ( ! plugin_url ) {
						return;
					}
```

admin/plugins.php
```diff
@@ -143,13 +157,17 @@ function perflab_render_plugin_card( array $plugin_data ) {
 		case 'install':
 			if ( $status['url'] ) {
 				if ( $compatible_php && $compatible_wp && current_user_can( 'install_plugins' ) ) {
+					$standalone_plugins_file_map = perflab_get_standalone_plugins_file_map();
+					$plugin_file                 = isset( $standalone_plugins_file_map[ $plugin_data['slug'] ] ) ? wp_create_nonce( 'perflab_activate_plugin_' . $standalone_plugins_file_map[ $plugin_data['slug'] ] ) : '';
```
felixarntz: See my comment in https://github.com/WordPress/performance/pull/900/files#r1425820667: We could use the plugin slug here and use that same value to validate the nonce later. No need to use the plugin basename.

admin/load.php
```diff
@@ -636,3 +639,40 @@ function perflab_plugin_admin_notices() {
...
+					var url = new URL( plugin_url );
+					url.searchParams.set( 'action', 'perflab_activate_plugin' );
+					url.searchParams.set( '_wpnonce', nonce );
+					target_element.attr( 'href', url.href );
+				}, 1500 );
+			}
+		} );
+	} )( jQuery );
+	<?php
+	wp_print_inline_script_tag( ob_get_clean() );
```
westonruter: This will retain syntax highlighting in most IDEs. See https://core.trac.wordpress.org/ticket/59444

```suggestion
	$js = <<<JS
	?>
	( function( $ ) {
		$( document ).ajaxComplete( function( event, xhr, settings ) {
			// Check if this is the 'install-plugin' request.
			if ( settings.data && typeof settings.data === 'string' && settings.data.includes( 'action=install-plugin' ) ) {
				var target_element = $( event.target.activeElement );
				// Bail early if the element is not a[href].
				if ( ! target_element.is( 'a[href]' ) ) {
					return;
				}
				/*
				 * WordPress core uses a 1s timeout for updating the activation link,
				 * so we set a 1.5 timeout here to ensure our changes get updated after
				 * the core changes have taken place.
				 */
				setTimeout( function() {
					var plugin_url = target_element.attr( 'href' );
					var nonce = target_element.attr( 'data-plugin-activation-nonce' );
					var url = new URL( plugin_url );
					url.searchParams.set( 'action', 'perflab_activate_plugin' );
					url.searchParams.set( '_wpnonce', nonce );
					target_element.attr( 'href', url.href );
				}, 1500 );
			}
		} );
	} )( jQuery );
JS;
	wp_print_inline_script_tag( $js );
```

admin/plugins.php
```diff
@@ -143,13 +157,17 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+
 					$action_links[] = sprintf(
-						'<a class="install-now button" data-slug="%s" href="%s" aria-label="%s" data-name="%s">%s</a>',
+						'<a class="install-now button" data-slug="%s" href="%s" aria-label="%s" data-name="%s" data-plugin-activation-nonce="%s">%s</a>',
 						esc_attr( $plugin_data['slug'] ),
 						esc_url( $status['url'] ),
 						/* translators: %s: Plugin name and version. */
 						esc_attr( sprintf( _x( 'Install %s now', 'plugin', 'default' ), $name ) ),
 						esc_attr( $name ),
+						esc_attr( $plugin_file ),
```
westonruter: Variable name should rather be mention `nonce`, no?

admin/plugins.php
mukeshpanchal27: Yes, that is a custom field that I use to retrieve the plugin slug in JS. However, I'm uncertain about how the JavaScript utilizes that particular field in the code located here: [link to the code on GitHub](https://github.com/WordPress/wordpress-develop/blob/trunk/src/js/_enqueues/wp/updates.js#L2167) 😞

admin/load.php
mukeshpanchal27: Thanks for sharing.

admin/plugins.php
mukeshpanchal27: @felixarntz @swissspidy i have added plugin activation support using the module slug as suggested.

mukeshpanchal27:
Thank you, @felixarntz, @swissspidy, and @westonruter, for your feedback. I have addressed it, and the PR is now ready for review.

admin/load.php
```diff
@@ -636,3 +647,44 @@ function perflab_plugin_admin_notices() {
...
+				var plugin_slug = target_element.attr( 'data-slug' );
+				var url = new URL( plugin_url );
+				url.searchParams.set( 'action', 'perflab_activate_plugin' );
+				url.searchParams.set( '_wpnonce', nonce );
+				url.searchParams.append( 'plugin_slug', plugin_slug );
+				target_element.attr( 'href', url.href );
+			}, 1500 );
+		}
+	} );
+} )( jQuery );
```
westonruter: Reformatting with ~Prettier~:

```suggestion
( function( $ ) {
	$( document ).ajaxComplete( function( event, xhr, settings ) {
		// Check if this is the 'install-plugin' request.
		if ( settings.data && typeof settings.data === 'string' && settings.data.includes( 'action=install-plugin' ) ) {
			var target_element = $( event.target.activeElement );
			if ( ! target_element ) {
				return;
			}
			/*
			* WordPress core uses a 1s timeout for updating the activation link,
			* so we set a 1.5 timeout here to ensure our changes get updated after
			* the core changes have taken place.
			*/
			setTimeout( function() {
				var plugin_url = target_element.attr( 'href' );
				if ( ! plugin_url ) {
					return;
				}
				var nonce = target_element.attr( 'data-plugin-activation-nonce' );
				var plugin_slug = target_element.attr( 'data-slug' );
				var url = new URL( plugin_url );
				url.searchParams.set( 'action', 'perflab_activate_plugin' );
				url.searchParams.set( '_wpnonce', nonce );
				url.searchParams.append( 'plugin_slug', plugin_slug );
				target_element.attr( 'href', url.href );
			}, 1500 );
		}
	} );
} )( jQuery );
```

admin/load.php
```diff
@@ -540,21 +543,29 @@ function perflab_enqueue_modules_page_scripts() {
...
+	check_admin_referer( "perflab_activate_plugin_{$plugin_slug}" );
 
-	if ( ! current_user_can( 'activate_plugin', $plugin ) ) {
+	$plugins = get_plugins( '/' . $plugin_slug );
+
+	if ( empty( $plugins ) ) {
+		return;
+	}
+
+	$keys        = array_keys( $plugins );
```
westonruter: I suggest something more descriptive than `$keys`. What does this array actually contain?

westonruter: It's the "plugin file name" per `get_plugins()` so probably better to name this `$plugin_file_names`

admin/load.php
```diff
@@ -540,21 +543,29 @@ function perflab_enqueue_modules_page_scripts() {
...
+	$plugin_file = $plugin_slug . '/' . $keys[0];
```
westonruter: How is it that the first key is the one we know needs to be used?

admin/load.php
```diff
@@ -540,21 +543,29 @@ function perflab_enqueue_modules_page_scripts() {
...
 
-	// The plugin being activated.
-	$plugin = sanitize_text_field( wp_unslash( $_GET['plugin'] ) );
+	$plugin_slug = sanitize_text_field( wp_unslash( $_GET['plugin_slug'] ) );
 
-	check_admin_referer( "perflab_activate_plugin_{$plugin}" );
+	check_admin_referer( "perflab_activate_plugin_{$plugin_slug}" );
 
-	if ( ! current_user_can( 'activate_plugin', $plugin ) ) {
+	$plugins = get_plugins( '/' . $plugin_slug );
```
westonruter: I'm somewhat confused by the `'/' . $plugin_slug` argument. I probably just missed this, but it's getting plugins inside of the specified plugin? Shouldn't this rather be getting plugins that are installed since they are standalone?

admin/load.php
felixarntz: I'm not sure we'll want to do that. Did you use regular Prettier instead of the WordPress-flavored version? It doesn't follow the WordPress JS coding standards I believe (see e.g. all the missing spaces).

admin/load.php
```diff
@@ -540,21 +543,29 @@ function perflab_enqueue_modules_page_scripts() {
  */
 function perflab_activate_plugin() {
 	// Do not proceed if plugin query arg is not present.
-	if ( empty( $_GET['plugin'] ) ) {
+	if ( empty( $_GET['plugin_slug'] ) ) {
```
felixarntz: Why change the field name? I believe this will break the other plugin activation functionality that already existed via #651.

Let's implement this to be compatible with both ways:
* We should keep using the existing `plugin` name, to not have to update unrelated code.
* We should check whether `plugin` is a plugin slug or a plugin basename and handle accordingly:
    * If it's a basename, everything should work as it did before.
    * If it's a plugin slug (and _only_ then), we should run the logic to find the plugin basename (i.e. the `get_plugins()` call etc. you already have).

admin/load.php
felixarntz: I was confused by that at first, but it's actually a very smart solution. Since this callback is by definition about a single plugin, in order to find its plugin basename (i.e. plugin slug + main file name), we need to call `get_plugins()`.

Looking for the plugin within only the folder we already know is much more efficient than calling `get_plugins()` for _all_ available plugins and searching through those.

Maybe worth clarifying in a comment:
```suggestion
	// Since we know the plugin folder name, for efficiency we can simply look for the plugin main file there.
	$plugins = get_plugins( '/' . $plugin_slug );
```

admin/load.php
```diff
@@ -540,21 +543,29 @@ function perflab_enqueue_modules_page_scripts() {
...
+	$plugin_slug = sanitize_text_field( wp_unslash( $_GET['plugin_slug'] ) );
 
-	check_admin_referer( "perflab_activate_plugin_{$plugin}" );
+	check_admin_referer( "perflab_activate_plugin_{$plugin_slug}" );
 
-	if ( ! current_user_can( 'activate_plugin', $plugin ) ) {
+	$plugins = get_plugins( '/' . $plugin_slug );
+
+	if ( empty( $plugins ) ) {
+		return;
```
felixarntz: This shouldn't just return but rather `wp_die()` with an error message, e.g. `Plugin not found.` or `Plugin main file not found.`.

admin/load.php
felixarntz: See above, it'll always be just the one plugin (or none if it's not found for some reason).

admin/load.php
```diff
@@ -636,3 +647,44 @@ function perflab_plugin_admin_notices() {
...
+				var plugin_url = target_element.attr( 'href' );
+				if ( ! plugin_url ) {
+					return;
+				}
+				var nonce = target_element.attr( 'data-plugin-activation-nonce' );
+				var plugin_slug = target_element.attr( 'data-slug' );
+				var url = new URL( plugin_url );
+				url.searchParams.set( 'action', 'perflab_activate_plugin' );
+				url.searchParams.set( '_wpnonce', nonce );
+				url.searchParams.append( 'plugin_slug', plugin_slug );
```
felixarntz: See above, let's use `plugin` instead of `plugin_slug` for interoperability with the existing code.

admin/load.php
westonruter: Right. I used regular Prettier. I've updated the suggestion to just fix the comment indentation.

admin/load.php
westonruter: Ah, I see. So it's acting as if the plugin which is in a directory is actually a single PHP file plugin.

mukeshpanchal27: WP core use this core pattern many places https://github.com/WordPress/wordpress-develop/blob/f2c78fe642b307acbb77f21abf85a45ef35b1ca4/src/wp-admin/includes/class-plugin-upgrader.php#L541 so we reuse that to find the plugin file.

mukeshpanchal27:
@felixarntz @westonruter, I identified an edge case while installing multiple plugins and try to activate first plugin where we encountered an incorrect selector in `event.target.activeElement`. Consequently, In https://github.com/WordPress/performance/pull/900/commits/b936064451d66e02b3af8ff1996b39cc666a2e82 i have updated the selector in the logic, and the issue has been resolved. Could you please review it? Thank you.

admin/load.php
```diff
@@ -544,10 +547,23 @@ function perflab_activate_plugin() {
...
-	// The plugin being activated.
 	$plugin = sanitize_text_field( wp_unslash( $_GET['plugin'] ) );
 
-	check_admin_referer( "perflab_activate_plugin_{$plugin}" );
+	// Check whether "plugin" is a plugin slug or a plugin basename.
+	if ( str_contains( $plugin, '/' ) ) {
+		check_admin_referer( "perflab_activate_plugin_{$plugin}" );
+	} else {
+		check_admin_referer( "perflab_activate_plugin_{$plugin}" );
+
```
felixarntz: This can be simplified as the admin referer check remains the same regardless of whether it's a plugin slug or basename. The respective logic dealing with the nonce handles this for both approaches.
```suggestion
	check_admin_referer( "perflab_activate_plugin_{$plugin}" );

	// If `$plugin` is a plugin slug rather than a plugin basename, determine the full plugin basename.
	if ( ! str_contains( $plugin, '/' ) ) {
```

felixarntz (approved):
Great work @mukeshpanchal27!

Only one minor thing that would be good to address, but not a blocker.

admin/plugins.php
```diff
@@ -144,12 +144,13 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
 				if ( $compatible_php && $compatible_wp && current_user_can( 'install_plugins' ) ) {
 					$action_links[] = sprintf(
-						'<a class="install-now button" data-slug="%s" href="%s" aria-label="%s" data-name="%s">%s</a>',
+						'<a class="install-now button" data-slug="%s" href="%s" aria-label="%s" data-name="%s" data-plugin-activation-nonce="%s">%s</a>',
 						esc_attr( $plugin_data['slug'] ),
 						esc_url( $status['url'] ),
 						/* translators: %s: Plugin name and version. */
 						esc_attr( sprintf( _x( 'Install %s now', 'plugin', 'default' ), $name ) ),
 						esc_attr( $name ),
+						wp_create_nonce( 'perflab_activate_plugin_' . $plugin_data['slug'] ),
```
westonruter: Perhaps overkill, but better to be consistent. Also, `wp_create_nonce()` is a pluggable function so in theory it could return anything.

```suggestion
						esc_attr( wp_create_nonce( 'perflab_activate_plugin_' . $plugin_data['slug'] ) ),
```

admin/load.php
```diff
@@ -636,3 +652,63 @@ function perflab_plugin_admin_notices() {
...
+			// Loop through the parameters to find the 'slug' parameter.
+			var slug = '';
+			for ( var i = 0; i < params.length; i++ ) {
+				var param = params[ i ].split( '=' );
+				if ( param[0] === 'slug') {
+					slug = param[1];
+					break;
+				}
+			}
```
westonruter: If I understand the `settings.data` correctly, this can be simplified by using `URLSearchParams`:

```suggestion
			var params = new URLSearchParams( settings.data );
			var slug = params.get('slug');
```

admin/load.php
```diff
@@ -636,3 +652,63 @@ function perflab_plugin_admin_notices() {
...
+
+			// Check if 'slug' was found and output the value.
+			if ( ! slug ) {
+				return;
+			}
+
+			var target_element = $( '.wpp-standalone-plugins a[data-slug=' + slug + ']' );
```
```suggestion
			var target_element = $( '.wpp-standalone-plugins a[data-slug="' + slug + '"]' );
```

westonruter (dismissed):

admin/load.php
westonruter: Good to know!

admin/load.php
swissspidy: Or maybe a template literal? Reduces chance of mistakes

westonruter: Good idea.

```suggestion
			var target_element = $( `.wpp-standalone-plugins a[data-slug="${slug}"]` );
```

mukeshpanchal27:
@swissspidy Could you kindly take a moment to review the recent changes?

swissspidy (approved):

Issue #654, opened 2023-02-16
Remove the modules that were published as standalone plugins from the Performance Lab plugin


felixarntz:
## Feature Description

Follow up to #651 #652 #653: The final part of this workstream will be to actually remove the modules that are published as standalone plugins from the PL plugin. This will not actually involve deleting the folders in the repository, as they will still be developed here (i.e. as a monorepo), instead we will need to ensure they are excluded from the plugin, maybe move to a new `plugins` directory in this repo, and at that point we can also change their file headers to actual plugin headers instead of PL module headers.

### Requirements

**This work should be implemented against a new `feature/modules-to-plugins` feature branch.**

After #934 has been completed:
* Move the folders for the modules referenced in `plugins.json` into the new `plugins` folder (without their focus area folders).
* Keep those modules/plugins unchanged, _except_ for:
    * Update the respective `load.php` file to have the plugin header instead of the module header.
    * The plugin header should have the same content as if it had been built using our current "module to plugin" build workflow.
* Migrate the existing `plugins.json` entries from the `modules` property to the `plugins` property.
* Update the references to the migrated modules in `CODEOWNERS` to point to the respective plugin folders.
* Remove references to the migrated modules in the `perflab_get_standalone_plugins_constants()` function.

paaljoachim:
Why remove features from the Performance Lab plugin? Can we have the various modules still included in the PL plugin so the user can choose to either use the PL modules on/off plugin or individual plugins?
(Makes me think of a Jetpack light with the opportunity to turn a feature on or off.)

I really would like to continue using the Performance Lab plugin that also includes the various modules so that I just need to install one plugin that includes a lot of really helpful features.

joemcgill:
Thanks for the feedback @paaljoachim. Ultimately, the decision to publish these features as standalone plugins rather than as a bundled plugin came from the WordPress project leadership. Our goal is to make clear distinctions between the standalone plugins and the main performance lab plugin to not create confusion by having the same features available in two separate places. We're going to do our best to make the user experience for activating performance features as plugins as seamless as we can.

paaljoachim:
Hey Joe.

This is how I understand it.
Creating separate plugins these will in a sense become the "standard" feature plugin. So that one can campaign the single feature plugins to be included into WordPress core.

[...]

PR #904, opened 2023-12-14
Add: Auto Sizes for Lazy-loaded Images

### LINKED ISSUE #791: Module Proposal: auto-sizes for lazy-loaded images
## Overview
- Image Auto-sizes
- Proposed module slug: image-auto-sizes
- Proposed module owner GitHub username(s): @joemcgill (others welcome!)

## About the module
Recently, the HTML spec was updated to support [sizes="auto" for lazy-loaded images](https://github.com/whatwg/html/pull/8008). The purpose of this module would be to automatically add `auto` to the beginning of the `sizes` attribute for any image that is being lazy loaded.

### Purpose

While [implementation is still ongoing](https://chromium-review.googlesource.com/c/chromium/src/+/4520266/20#message-661c7b2341e041dc6692d16dfbd30d680071ca01), this could be a pretty big performance improvement for WordPress sites, since many rely on the default `sizes` attribute rather than filtering the value to a more accurate one based on the intended layout of the site's theme.

### Scope

This would only implement auto sizes for lazy-loaded images as an experiment until browser adoption for this feature is far enough along to propose for core, and should be a base implementation that a future core proposal would be based on.

This does not take the place of #760, which would still be useful to pursue, but could make that effort less important by getting the same performance benefits for all lazy-loaded images.

At minimum, the implementation should filter `wp_calculate_image_sizes` to prepend "auto, " to the beginning of the string. Given that this is a progressive enhancement, any browser that doesn't recognize the "auto" value in the sizes string, will ignore it and process the rest of the string like normal.

It is [currently unclear](https://github.com/whatwg/html/pull/9493#issuecomment-1646236877) if this syntax is used on an image that is not being lazy loaded, if the original sizes value will be respected or if the default `100vw` fallback for the parser will be used instead, so experimentation against a real implementation will be important. If the default is used, then the filter would need to also check to see if lazy loading is enabled and likely post-filter images to remove auto sizing for any images not being lazy loaded, since `srcset` and `sizes` is calculated prior to lazy loading in `wp_filter_content_tags`.

### Rationale

See above.

### Other

Additional info:
* [Current implementation considerations](https://github.com/whatwg/html/issues/9448)
* [Test suite for the auto sizes syntax](http://wpt.live/html/semantics/embedded-content/the-img-element/sizes/sizes-auto.html)

joemcgill:
## Summary

**Blocked for merge by #934.**

This adds a new module that will automatically enhance the `sizes` attribute of lazy-loaded images to support the new `auto` syntax.

See: https://github.com/whatwg/html/issues/4654

Fixes: #791.

## Relevant technical choices

This adds just the basic functionality needed to implement auto sizes for images created by WP `wp_get_attachment_image` function or which get responsive attributes added by the `wp_filter_content_tags` callback. In both cases, `auto` is prepended to the sizes attribute only if the image already has `loading="lazy"` applied, which is a requirement to avoid introducing regressions. For more details about compatibility with browser implementations, see [this comment](https://github.com/WordPress/performance/issues/791#issuecomment-1856330050).

## Checklist


joemcgill:
Still need to add some test coverage, but wanted to share an initial implementation for testing/feedback.

modules/images/auto-sizes/load.php
```diff
@@ -0,0 +1,18 @@
...
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+// Define the constant.
+if ( defined( 'IMAGE_AUTO_SIZES' ) ) {
+	return;
+}
+
+define( 'IMAGE_AUTO_SIZES', 'Performance Lab ' . PERFLAB_VERSION );
```
felixarntz: Maybe worth clarifying in the name that it is a version.
```suggestion
if ( defined( 'IMAGE_AUTO_SIZES_VERSION' ) ) {
	return;
}

define( 'IMAGE_AUTO_SIZES_VERSION', 'Performance Lab ' . PERFLAB_VERSION );
```

modules/images/auto-sizes/load.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/**
```
felixarntz: Can you also add entries for this module to `.github/CODEOWNERS` like there are for the other modules?

felixarntz (commented):
Thanks @joemcgill, code looks great so far.

One consideration here unrelated to the code itself is how we are going to publish this. A few thoughts:
* We're right now in the last steps of unbundling modules (to be released mid-January). So we probably don't want to add any modules now.
* For that reason I'd recommend to change this PR to go against a new feature branch for this module for now. That way development can continue, and once the repository has shifted to the full standalone plugins monorepo (likely early February) it can be merged into `trunk`.
* Separately, we may want to already kick off the plugin review process for this? Even if the code as of now isn't necessarily final, it's probably solid enough to submit for review. You could use another experimental or local branch based on this where you add this new module to `plugins.json`. With that, you can then run `npm run build-plugins` and get the plugin in a version ready to submit to .org for review.
* The only other thing needed for .org submission is a `readme.txt` (which we'll need later anyway). Could you maybe add one to this PR? You can probably copy one of the other standalone plugins' readme files and adjust just the few necessary bits.

joemcgill:
Thanks @felixarntz. Great questions about how we would publish this. This is already essentially the feature branch for this effort, so happy to just keep iterating here and/or submitting additional PRs into this branch. I'll address the rest of your suggestions over the coming week.

The bigger question you raise is how we make modules like this available for people to test before they're published to the wp.org repository once the unbundling effort is complete, given that size of the review backlog at the moment.

I wonder if we can keep some modules in an experimental state where they can be activated from the main performance lab plugin prior to being either promoted to a standalone plugin or as a direct implementation into Core. Right now, you can install the plugin from this repo and checkout this feature branch to test the functionality, but that wont work for modules added after the unbundling effort is complete.

Otherwise, I think we need to update the release script so that zips of pre-release plugins can be downloaded from the https://github.com/WordPress/performance/releases page so we're not completely dependent on the WP.org repo for distributing standalone plugins from this mono-repo.

modules/images/auto-sizes/load.php
```diff
@@ -0,0 +1,18 @@
...
+ * Module Name: Auto Sizes for Lazy-loaded Images.
+ * Description: Implements auto sizes for lazy loaded images.
```
swissspidy: Does `module-i18n.php` need to be updated to include these strings?

joemcgill: Good question. I'm not sure if that is something that needs to happen as part of a new module PR or is automated as part of the release process.

felixarntz:
@joemcgill 
> This is already essentially the feature branch for this effort, so happy to just keep iterating here and/or submitting additional PRs into this branch. I'll address the rest of your suggestions over the coming week.

Happy for you to keep working in here, though the main reason I'd suggest a feature branch is to avoid accidental merge into `trunk`. As in: Even once this has 2 approvals, we need to determine the best path forward before "just merging". Alternatively, one of us can just keep a "requesting changes" on this to avoid merging "by accident".

> I wonder if we can keep some modules in an experimental state where they can be activated from the main performance lab plugin prior to being either promoted to a standalone plugin or as a direct implementation into Core. Right now, you can install the plugin from this repo and checkout this feature branch to test the functionality, but that wont work for modules added after the unbundling effort is complete.

Those are good ideas. We could always consider keeping experimental modules in Performance Lab in the beginning, and once they're approved on .org, we'd phase them out. In that case, maybe the migration logic that @10upsimon and @mukeshpanchal27 have been working on should remain in the plugin even indefinitely, as it would continue to be useful in the future, not only this initial migration push as previously assumed.

modules/images/auto-sizes/load.php
felixarntz: No need to update, this is part of the release process (see https://make.wordpress.org/performance/handbook/performance-lab/releasing-the-plugin/#update-translation-strings).

joemcgill:
> Even once this has 2 approvals, we need to determine the best path forward before "just merging". Alternatively, one of us can just keep a "requesting changes" on this to avoid merging "by accident".

For the time being, I've converted this to a `Draft` PR to avoid accidental merge.

I think the conversation about how to handle new modules once we've completed the transition to standalone plugins #656 is worth it's own conversation. I'll open a new issue for us to discuss. I am also going to try to spend some time seeing how difficult it would be to post standalone releases to this repo, since I think that would be useful regardless of whichever future strategy we choose.

joemcgill:
@felixarntz I forgot that the branch protections on feature branches would block me from continuing to update this branch directly, so I've opened #913 to address the code review feedback and to add initial unit tests.

modules/images/auto-sizes/load.php
joemcgill: [a7dffc5](https://github.com/WordPress/performance/pull/904/commits/a7dffc584250ddaa67666cb6617eedb173ba1f37)

modules/images/auto-sizes/load.php
joemcgill: [a5e4afa](https://github.com/WordPress/performance/pull/904/commits/a5e4afa644c085fc9367b8596b9c17cae7f91d7b)

modules/images/auto-sizes/hooks.php
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Adds auto to the sizes attribute to the image, if applicable.
+ *
+ * @since n.e.x.t
+ *
+ * @param array $attr Attributes for the image markup.
+ * @return array The filtered attributes for the image markup.
+ */
+function performance_lab_auto_sizes_attr( $attr ) {
```
mukeshpanchal27: @joemcgill, would it be possible to remove the `performance_lab_` prefix from the function name since it isn't utilized for any module? Perhaps we could rename it to something like `auto_sizes_image_attr`, aligning with our format of `modulename_`.

joemcgill: Handled this is PR 922

joemcgill:
@felixarntz I've opened [a new PR](https://github.com/WordPress/performance/pull/922) against this feature branch that I believe prepares this to be submitted to the the plugin repo for review. I've build the plugin locally by modifying the `plugins.json` file and have it successfully running [on my personal site](https://joemcgill.net/).

I'm happy to submit this to the Plugin Team for review, but want to make sure that the repo they create will be accessible using the same credentials as the other plugins built from this repo and would appreciate your guidance.

felixarntz:
@joemcgill Repos will need to be submitted using the `performanceteam` account. I'm happy to take care of that, but also happy for you to do so. We can chat about the process and credentials via DM.

joemcgill:
Now that #922 has been merged, this module is ready for final review. 

**Merging is currently blocked while we await branching and releasing PL v2.8.0.**

tests/modules/images/auto-sizes/auto-sizes-test.php
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * @package performance-lab
+ * @group   auto-sizes
+ */
+
+class AutoSizesTests extends WP_UnitTestCase {
+
+	/**
+	 * Attachment ID.
+     *
```
westonruter: I realize you can't (shouldn't?) apply this suggestion here, but this will fix the linting error I believe.

```suggestion
	 *
```

joemcgill: Yep! Handled in https://github.com/WordPress/performance/pull/931, which also requires some extra hoops to add the correct labels, etc.

Interestingly our `npm run lint-php` script does not catch this, while `composer run lint` does. I'm going to investigate why that is at a later date. I'm not sure what changed upstream that caused the error to show up here in the first place but it's good that it did.

westonruter (approved):

felixarntz:
@joemcgill Just sharing a thought here: You may want to move this module into a new `plugins/{pluginSlug}` folder in the repository root and update the module header to become a plugin header.

Once #934 has been completed (which I don't think requires a lot of time), then this could be merged safely without further changes. The automation processes for deploying would require #935 to be completed as well, but in any case the plugin could already live in `trunk` after #934 is complete without any downsides.

Just to facilitate this PR getting merged sooner than later, in case it blocks anything. I may do the same for the `feature/speculationrules` branch.

joemcgill:
Good idea, @felixarntz. I'll track the progress of those two issues and prepare to make the switch.

felixarntz:
@joemcgill See https://github.com/WordPress/performance/pull/946#issuecomment-1897105844 for a related problem that you may face here too, given this feature has test coverage too.

mukeshpanchal27:
@joemcgill, as per Felix's comment https://github.com/WordPress/performance/issues/654#issuecomment-1898808673, could you please update the branch to `feature/modules-to-plugins` instead of `trunk` and make the necessary changes so that we can check the plugin's unit test suite? cc. @swissspidy

mukeshpanchal27:
@joemcgill, if you haven't started yet, may I proceed with the necessary changes as per the comments above?

mukeshpanchal27:
@joemcgill @swissspidy, this PR is ready for merge. The test suite will be handled in #972. WDYT?

swissspidy (approved):
Good to get the ball rolling 👍

mukeshpanchal27 (approved):

joemcgill (commented):
I'm happy to have you merge this into the `feature/modules-to-plugins` branch ahead of the changes being proposed in #972, I think I'm just a bit confused about the sequencing of these changes, but I'm sure it's just me being dense.

mukeshpanchal27 (approved):
Approve again for merge.

PR #910, opened 2023-12-18
Implement admin pointer to indicate to the user they need to migrate modules to their standalone plugins

### LINKED ISSUE #652: Implement migration logic and UI from old PL modules to new standalone plugins in PL settings screen

### LINKED ISSUE #653: Implement admin pointer to indicate to the user they need to activate the new standalone plugins
## Feature Description

Follow-up to #652: For sites where the conditions in #652 apply, an admin pointer should also show up to ensure the user sees that they need to do something on the PL settings screen (specifically, install/activate the standalone plugins for the modules that they used to have active).

### Requirements

* TODO.

mukeshpanchal27:
## Summary

Fixes #653

Prior to merging this PR, we'll need to ensure that #899 is merged.

## Checklist

### DIFF
--- a/admin/load.php
+++ b/admin/load.php
@@ -420,10 +420,9 @@ function perflab_admin_pointer( $hook_suffix ) {
 		add_action(
 			'admin_print_footer_scripts',
 			static function () {
-				$content  = __( 'Your site is using modules which will be removed in the future in favor of their equivalent standalone plugins.', 'performance-lab' );
-				$content .= ' ' . sprintf(
+				$content = sprintf(
 					/* translators: %s: settings page link */
-					__( 'Open %s to learn more about next steps to keep the functionality available.', 'performance-lab' ),
+					__( 'Your site is using modules which will be removed in the future in favor of their equivalent standalone plugins. Open %s to learn more about next steps to keep the functionality available.', 'performance-lab' ),
 					'<a href="' . esc_url( add_query_arg( 'page', PERFLAB_MODULES_SCREEN, admin_url( 'options-general.php' ) ) ) . '">' . __( 'Settings > Performance', 'performance-lab' ) . '</a>'
 				);
 				perflab_render_pointer(
--- a/admin/load.php
+++ b/admin/load.php
@@ -403,6 +403,41 @@ function perflab_admin_pointer( $hook_suffix ) {
 	$current_user = get_current_user_id();
 	$dismissed    = explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) );
 
+	/*
+	 * Show an admin pointer if module is active to prompt
+	 * for an action to use the standalone plugin instead.
+	 */
+	$available_module_names = perflab_get_modules_compatible_with_standalone_plugin();
+	if (
+		! empty( $available_module_names )
+		&& current_user_can( 'install_plugins' )
+		&& current_user_can( 'activate_plugins' )
+		&& ! in_array( 'perflab-module-migration-pointer', $dismissed, true )
+	) {
+		// Enqueue the pointer logic and return early.
+		wp_enqueue_style( 'wp-pointer' );
+		wp_enqueue_script( 'wp-pointer' );
+		add_action(
+			'admin_print_footer_scripts',
+			static function () {
+				$content  = __( 'Your site is using modules which will be removed in the future in favor of their equivalent standalone plugins.', 'performance-lab' );
+				$content .= ' ' . sprintf(
+					/* translators: %s: settings page link */
+					__( 'Open %s to learn more about next steps to keep the functionality available.', 'performance-lab' ),
+					'<a href="' . esc_url( add_query_arg( 'page', PERFLAB_MODULES_SCREEN, admin_url( 'options-general.php' ) ) ) . '">' . __( 'Settings > Performance', 'performance-lab' ) . '</a>'
+				);
+				perflab_render_pointer(
+					'perflab-module-migration-pointer',
+					array(
+						'heading' => __( 'Action required', 'performance-lab' ),
+						'content' => $content,
+					)
+				);
+			}
+		);
+		return;
+	}
+
 	if ( in_array( 'perflab-admin-pointer', $dismissed, true ) ) {
 		return;
 	}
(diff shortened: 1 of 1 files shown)


admin/load.php
```diff
@@ -403,6 +403,41 @@ function perflab_admin_pointer( $hook_suffix ) {
...
+				$content  = __( 'Your site is using modules which will be removed in the future in favor of their equivalent standalone plugins.', 'performance-lab' );
+				$content .= ' ' . sprintf(
+					/* translators: %s: settings page link */
+					__( 'Open %s to learn more about next steps to keep the functionality available.', 'performance-lab' ),
+					'<a href="' . esc_url( add_query_arg( 'page', PERFLAB_MODULES_SCREEN, admin_url( 'options-general.php' ) ) ) . '">' . __( 'Settings > Performance', 'performance-lab' ) . '</a>'
+				);
+				perflab_render_pointer(
+					'perflab-module-migration-pointer',
+					array(
+						'heading' => __( 'Action required', 'performance-lab' ),
```
swissspidy: Maybe use `Performance Lab` (or `Performance Lab: Action required` or so) as the heading? Makes it more clear which plugin the pointer is coming from.

admin/load.php
```diff
@@ -403,6 +403,41 @@ function perflab_admin_pointer( $hook_suffix ) {
...
+		&& ! in_array( 'perflab-module-migration-pointer', $dismissed, true )
+	) {
+		// Enqueue the pointer logic and return early.
+		wp_enqueue_style( 'wp-pointer' );
+		wp_enqueue_script( 'wp-pointer' );
+		add_action(
+			'admin_print_footer_scripts',
+			static function () {
+				$content  = __( 'Your site is using modules which will be removed in the future in favor of their equivalent standalone plugins.', 'performance-lab' );
+				$content .= ' ' . sprintf(
```
swissspidy: Possible to do this in one string instead of doing string concatenation?

admin/load.php
mukeshpanchal27: Thank you for the swift review.

I've made an update to the pointer heading. While we implemented a similar pointer previously in #739 for the SQLite module and omitted it, considering the overall migration, it seems sensible to include it now.

admin/load.php
mukeshpanchal27: Updated in https://github.com/WordPress/performance/pull/910/commits/c8ac7e0698bf0a3cfb53d6585b46fa1c2d3b437b

swissspidy (approved):

admin/load.php
```diff
@@ -757,58 +791,74 @@ function perflab_plugin_admin_notices() {
...
-		</div>
-		<?php
+/**
+ * Returns an array of standalone plugins name which currently active modules.
+ *
+ * @since n.e.x.t
+ *
+ * @return string[]
+ */
+function perflab_get_modules_compatible_with_standalone_plugin() {
```
felixarntz: The behavior and naming of this function is a bit confusing since it pretty much does the same as `perflab_get_active_modules_with_standalone_plugins()`, only that it returns the module names and only if the relevant standalone plugin isn't active.

Returning the module _names_ is a UI-specific detail that IMO shouldn't be baked into the function. If we need the module names to display them, we should rather just return the entire module data.

How about:
* renaming this to `perflab_get_active_module_data_with_inactive_standalone_plugins()`
    * very verbose, but this is still better than an unclear name
* returning the full module data below. Then where we need the names we just extract them specifically.
* we then have to update the `@return` type doc as well (please make sure to add a description to clarify what it returns)

admin/load.php
```diff
@@ -757,58 +791,74 @@ function perflab_plugin_admin_notices() {
...
+	$active_modules_with_plugins = perflab_get_active_modules_with_standalone_plugins();
+	if ( empty( $active_modules_with_plugins ) ) {
+		return;
```
felixarntz: This returns `null` right now which isn't correct given the documented return type. Makes more sense to return an empty array.
```suggestion
		return array();
```

admin/load.php
```diff
@@ -757,58 +791,74 @@ function perflab_plugin_admin_notices() {
...
+	}
+
+	$module_data            = perflab_get_modules();
+	$available_module_names = array();
+	foreach ( $active_modules_with_plugins as $module_slug ) {
+		if ( isset( $module_data[ $module_slug ] ) && ! perflab_is_standalone_plugin_loaded( $module_slug ) ) {
+			$available_module_names[] = $module_data[ $module_slug ]['name'];
+		}
 	}
+	return $available_module_names;
```
felixarntz: See above. Also a small nit-pick, the `$module_slug` here is typically referred to as `$module_dir` elsewhere in the code as it contains both the focus directory and the module directory (the latter of which is the module's slug).
```suggestion
	$available_modules_data = array();
	foreach ( $active_modules_with_plugins as $module_dir ) {
		if ( isset( $module_data[ $module_dir ] ) && ! perflab_is_standalone_plugin_loaded( $module_dir ) ) {
			$available_modules_data[] = $module_data[ $module_dir ];
		}
	}
	return $available_modules_data;
```

admin/load.php
```diff
@@ -757,58 +791,74 @@ function perflab_plugin_admin_notices() {
 		<?php
 	}
 
-	$active_modules_with_plugins = perflab_get_active_modules_with_standalone_plugins();
+	$available_module_names = perflab_get_modules_compatible_with_standalone_plugin();
+	if ( empty( $available_module_names ) ) {
+		return;
+	}
```
felixarntz: See my other comment: Showing the names is a UI specific detail that shouldn't be baked into the API functions. If we change the function to return the full module data, we can simply change this to e.g.:
```suggestion
	$available_modules_data = perflab_get_active_module_data_with_inactive_standalone_plugins();
	if ( empty( $available_modules_data ) ) {
		return;
	}

	$available_module_names = wp_list_pluck( $available_modules_data, 'name' );
```

admin/load.php
```diff
@@ -757,58 +791,74 @@ function perflab_plugin_admin_notices() {
...
-		$available_module_names = array();
-		foreach ( $active_modules_with_plugins as $module_slug ) {
-			if ( isset( $module_data[ $module_slug ] ) && ! perflab_is_standalone_plugin_loaded( $module_slug ) ) {
-				$available_module_names[] = $module_data[ $module_slug ]['name'];
-			}
-		}
+	$modules_count = count( $available_module_names );
+	if ( $modules_count < 1 ) {
+		return;
+	}
```
felixarntz: This is useless now: We already check `empty` above, and `count( ... ) < 1` will always result in the same for an array.

Since we need the module count below, we should probably just remove the `empty` check above.

admin/load.php
```diff
@@ -406,6 +406,40 @@ function perflab_admin_pointer( $hook_suffix ) {
 	$current_user = get_current_user_id();
 	$dismissed    = explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) );
 
+	/*
+	 * Show an admin pointer if module is active to prompt
+	 * for an action to use the standalone plugin instead.
+	 */
+	$available_module_names = perflab_get_modules_compatible_with_standalone_plugin();
+	if (
+		! empty( $available_module_names )
```
felixarntz: If we change the function, then this can become something like the following which is less confusing:
```suggestion
	/*
	 * If there are any active modules with inactive standalone plugins,
	 * show an admin pointer to prompt the user to migrate.
	 */
	$active_modules_with_inactive_plugins = perflab_get_active_module_data_with_inactive_standalone_plugins();
	if (
		! empty( $active_modules_with_inactive_plugins )
```

felixarntz (changes requested):
Thanks @mukeshpanchal27. This looks good, though I left a few suggestions for improving the new function which is a bit too specifically tied to a UI consideration only relevant in one place.

felixarntz:
Another thought on this (which if we go for it, we could implement in a separate PR): We may want to "un-dismiss" the migration pointer when the user activates a module? This would ensure that, if the user brings the site back into a "problematic" state after initially having fixed the problem, the pointer shows again. I think that would be helpful, especially if the logic here remains in the plugin indefinitely (see related https://github.com/WordPress/performance/pull/904#issuecomment-1861314039). We could just always remove the pointer from being dismissed whenever a module is activated, as the pointer would only actually show anyway if relevant for the site's current state.

WDYT @mukeshpanchal27 @swissspidy?

admin/load.php
```diff
@@ -406,6 +406,40 @@ function perflab_admin_pointer( $hook_suffix ) {
...
+		// Enqueue the pointer logic and return early.
+		wp_enqueue_style( 'wp-pointer' );
+		wp_enqueue_script( 'wp-pointer' );
+		add_action(
+			'admin_print_footer_scripts',
+			static function () {
+				$content = sprintf(
+					/* translators: %s: settings page link */
+					__( 'Your site is using modules which will be removed in the future in favor of their equivalent standalone plugins. Open %s to learn more about next steps to keep the functionality available.', 'performance-lab' ),
+					'<a href="' . esc_url( add_query_arg( 'page', PERFLAB_MODULES_SCREEN, admin_url( 'options-general.php' ) ) ) . '">' . __( 'Settings > Performance', 'performance-lab' ) . '</a>'
```
```suggestion
					'<a href="' . esc_url( add_query_arg( 'page', PERFLAB_MODULES_SCREEN, admin_url( 'options-general.php' ) ) ) . '">' . esc_html__( 'Settings > Performance', 'performance-lab' ) . '</a>'
```

admin/load.php
```diff
@@ -406,6 +406,40 @@ function perflab_admin_pointer( $hook_suffix ) {
...
+	) {
+		// Enqueue the pointer logic and return early.
+		wp_enqueue_style( 'wp-pointer' );
+		wp_enqueue_script( 'wp-pointer' );
+		add_action(
+			'admin_print_footer_scripts',
+			static function () {
+				$content = sprintf(
+					/* translators: %s: settings page link */
+					__( 'Your site is using modules which will be removed in the future in favor of their equivalent standalone plugins. Open %s to learn more about next steps to keep the functionality available.', 'performance-lab' ),
```
westonruter: Perhaps overkill given that `wp_kses_post()` is being used when output in `perflab_render_pointer()`. However, since no markup is expected in this string, it seems like escaping here and below would make sense. The change could also be made to the default content in `perflab_render_pointer()` as well.

```suggestion
					esc_html__( 'Your site is using modules which will be removed in the future in favor of their equivalent standalone plugins. Open %s to learn more about next steps to keep the functionality available.', 'performance-lab' ),
```

admin/load.php
mukeshpanchal27: Thanks for feedback. Address in https://github.com/WordPress/performance/pull/910/commits/d855b9e67bcc8ec5a8dc22e24f325752724d421f

admin/load.php
mukeshpanchal27: Thanks for the feedback.

I have removed the `$modules_count < 1` check as it is duplicate of empty check.

mukeshpanchal27:
> Another thought on this (which if we go for it, we could implement in a separate PR): We may want to "un-dismiss" the migration pointer when the user activates a module?

Yes. IMO we should do that.

admin/load.php
```diff
@@ -757,58 +791,72 @@ function perflab_plugin_admin_notices() {
...
-				<span class="dashicons dashicons-update hidden"></span>
-			</p>
-		</div>
-		<?php
+/**
+ * Returns an array of active module data with inactive standalone plugins.
+ *
+ * @since n.e.x.t
+ *
+ * @return array Returns an array of active module data with inactive standalone plugins, otherwise an empty array.
```
felixarntz:
```suggestion
 * @return array Array of active module data with inactive standalone plugins, otherwise an empty array.
```

felixarntz (approved):

felixarntz:
> > Another thought on this (which if we go for it, we could implement in a separate PR): We may want to "un-dismiss" the migration pointer when the user activates a module?
> 
> Yes. IMO we should do that.

Thanks @mukeshpanchal27. Can you open a follow up PR for this when you get a chance? Not a blocker for the release, but a nice to have enhancement.

PR #914, opened 2023-12-20
Preload image for LCP element with background-image

westonruter:
## Summary

Add image `preload` links for the `background-image` on the LCP element.

Doing a quick test with the parallax Cover block, the changes in this PR reduce median LCP-TTFB from 71.1ms to 64.8ms over 25 requests respectively: almost a 9% reduction.

## Relevant technical choices

Adding `preload` links was already being done for LCP `img` elements. When all breakpoints had the same `img` as the LCP element, it also adds `fetchpriority=high` to that `img`. Since a CSS `background-image` cannot get `fetchpriority=high`, this is only added via the `preload` link.

Both the `background-image` property and `background` shorthand property are supported. A `url()` containing a data: URL is skipped.

Note that CSS allows for a `background`/`background-image` to have multiple `url()` CSS functions, resulting in multiple background images being layered on top of each other. This ability is not employed in core. Here is a regex to search WPDirectory for instances of this: `background(-image)?:[^;}]+?url\([^;}]+?[^_]url\(`. It is used in Jetpack with the second background image being a gradient. To support multiple background images, the logic would need to be modified to make `$background_image` an array and to have a more robust parser of the `url()` functions from the property value.

The `preload` links are improved to add a `screen` media type and to omit the needless `min-width:0px` media feature.

This PR also adds server timing for `image-loading-optimization`. When checking with `benchmark-server-timing` over 100 requests, enabling Image Loading Optimization only increases the response time by ~2ms.

Also, this removes the workaround for browsers that don't support `imagesrcset` since >95% do. See https://github.com/WordPress/performance/issues/117#issuecomment-1854489717.

## Checklist


modules/images/image-loading-optimization/optimization.php
swissspidy, line 1: There's now an outdated `TODO: Handle case when the LCP element is not an image at all, but rather a background-image.` comment on line 170

westonruter: Good catch! Fixed in d1250e6

swissspidy (approved):

modules/images/image-loading-optimization/optimization.php
```diff
@@ -180,44 +185,71 @@ function ilo_optimize_template_output_buffer( string $buffer ): string {
...
+		 * Note that CSS allows for a `background`/`background-image` to have multiple `url()` CSS functions, resulting
+		 * in multiple background images being layered on top of each other. This ability is not employed in core. Here
+		 * is a regex to search WPDirectory for instances of this: /background(-image)?:[^;}]+?url\([^;}]+?[^_]url\(/.
+		 * It is used in Jetpack with the second background image being a gradient. To support multiple background
+		 * images, this logic would need to be modified to make $background_image an array and to have a more robust
+		 * parser of the `url()` functions from the property value.
+		 */
+		$background_image = null;
+		if ( $style && preg_match( '/background(-image)?\s*:[^;]*?url\(\s*[\'"]?(?!data:)(?<background_image>.+?)[\'"]?\s*\)/', $style, $matches ) ) {
+			$background_image = $matches['background_image'];
```
adamsilverstein, line 201: nice use of a named sub-pattern, I forgot that was possible.

modules/images/image-loading-optimization/optimization.php
```diff
@@ -180,44 +185,71 @@ function ilo_optimize_template_output_buffer( string $buffer ): string {
...
+					$processor->set_attribute( 'data-ilo-added-fetchpriority', true );
+				}
 
-			// Never include loading=lazy on the LCP image common across all breakpoints.
-			if ( 'lazy' === $processor->get_attribute( 'loading' ) ) {
-				$processor->set_attribute( 'data-ilo-removed-loading', $processor->get_attribute( 'loading' ) );
-				$processor->remove_attribute( 'loading' );
+				// Never include loading=lazy on the LCP image common across all breakpoints.
+				if ( 'lazy' === $processor->get_attribute( 'loading' ) ) {
+					$processor->set_attribute( 'data-ilo-removed-loading', $processor->get_attribute( 'loading' ) );
```
adamsilverstein, line 222: nice, I like how the change is marked with a data attribute

modules/images/image-loading-optimization/optimization.php
```diff
@@ -180,44 +185,71 @@ function ilo_optimize_template_output_buffer( string $buffer ): string {
...
+					$processor->remove_attribute( 'loading' );
+				}
+			} elseif ( $all_breakpoints_have_url_metrics && $processor->get_attribute( 'fetchpriority' ) ) {
```
adamsilverstein, line 225: this bit is unclear to me. if all breakpoints have data, you are falling back to the server heuristics?

tests/modules/images/image-loading-optimization/optimization-tests.php
```diff
@@ -286,6 +358,124 @@ public function data_provider_test_ilo_optimize_template_output_buffer(): array
...
+				'buffer'   => '
+					<html lang="en">
+						<head>
+							<meta charset="utf-8">
+							<title>...</title>
+							<style>/* responsive styles to show only one div.header at a time... */</style>
+						</head>
+						<body>
+							<div class="header desktop" style="background: red no-repeat center/80% url(\'https://example.com/desktop-bg.jpg\'); width:100%; height: 200px;">This is the desktop background!</div>
+							<div class="header tablet" style=\'background-image:url( "https://example.com/tablet-bg.jpg" ); width:100%; height: 200px;\'>This is the tablet background!</div>
```
adamsilverstein, line 455: the quoting on this line is reversed from all of the others which makes it stand out. is that required for the test to pass?

tests/modules/images/image-loading-optimization/optimization-tests.php
```diff
@@ -286,6 +358,124 @@ public function data_provider_test_ilo_optimize_template_output_buffer(): array
...
+							<meta charset="utf-8">
+							<title>...</title>
+							<style>/* responsive styles to show only one div.header at a time... */</style>
+							<link as="image" data-ilo-added-tag="" fetchpriority="high" href="https://example.com/mobile-bg.jpg" media="screen and (max-width: 480px)" rel="preload">
+							<link as="image" data-ilo-added-tag="" fetchpriority="high" href="https://example.com/tablet-bg.jpg" media="screen and (min-width: 481px) and (max-width: 600px)" rel="preload">
+							<link as="image" data-ilo-added-tag="" fetchpriority="high" href="https://example.com/desktop-bg.jpg" media="screen and (min-width: 601px)" rel="preload">
+						</head>
+						<body>
+							<div class="header desktop" style="background: red no-repeat center/80% url(\'https://example.com/desktop-bg.jpg\'); width:100%; height: 200px;">This is the desktop background!</div>
+							<div class="header tablet" style=\'background-image:url( "https://example.com/tablet-bg.jpg" ); width:100%; height: 200px;\'>This is the tablet background!</div>
```
adamsilverstein, line 472: same Q this line

adamsilverstein (commented):
Looks great, left some small comments/questions

adamsilverstein (approved):

modules/images/image-loading-optimization/optimization.php
westonruter: If all breakpoints have data, then we can reliably use the `preload` links to prioritize loading of the LCP image. In that case, we can remove the `fetchpriority` attribute from the LCP image here.

tests/modules/images/image-loading-optimization/optimization-tests.php
westonruter: It's just testing that single-quoted and double-quoted strings both work in the CSS `url()`.

tests/modules/images/image-loading-optimization/optimization-tests.php
westonruter: Ditto. Just to make sure that the regex works with single quotes as well as double quotes.

modules/images/image-loading-optimization/optimization.php
adamsilverstein: I see, so you are favoring preload over fetchpriority? maybe we discussed this before, couldn't you use both of them? or is high priority already implied when using preload?

westonruter: The `preload` links also include `fetchpriority=high`, so having it on the element is redundant. Also, it can be counter productive to be on the `img` in the case where there are different LCP images for different viewports. If there is an image with `fetchpriority=high` but it is only for the desktop viewport then it will get loaded with high priority even for mobile users. The preload links also include `media` attribute with media queries so they enable `fetchpriority=high` to be conditional on the viewport size.

PR #915, opened 2023-12-20
Reset admin pointer dismissal for module migration when the user activates a module

mukeshpanchal27:
## Summary

Follow-up to https://github.com/WordPress/performance/pull/910#issuecomment-1861395246

The PR introduced the capability to "un-dismiss" a pointer when a user activates a module. Within the code, the `perflab_run_module_activation_deactivation` function is utilized, triggered upon saving settings for the Performance Lab.

## Checklist

### DIFF
--- a/load.php
+++ b/load.php
@@ -458,25 +458,30 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
 
 	// Get the list of modules that were activated, and load the activate.php files if they exist.
 	if ( ! empty( $value ) ) {
+		$enable_module_migration_pointer = false;
 		foreach ( $value as $module => $module_settings ) {
 			if ( ! empty( $module_settings['enabled'] ) && ( empty( $old_value[ $module ] ) || empty( $old_value[ $module ]['enabled'] ) ) ) {
 				perflab_activate_module( PERFLAB_PLUGIN_DIR_PATH . 'modules/' . $module );
+				$enable_module_migration_pointer = true;
 			}
+		}
+		if ( $enable_module_migration_pointer ) {
+			// Retrieve a list of active modules with associated standalone plugins.
+			$active_modules_with_plugins = perflab_get_active_modules_with_standalone_plugins();
 
-			// Check whether it is necessary to display an admin pointer to prompt the user to migrate.
-			$plugin_slug        = basename( $module );
-			$standalone_plugins = perflab_get_standalone_plugins();
-
-			// Plugin slug should be in standalone plugins list.
-			if ( ! in_array( $plugin_slug, $standalone_plugins, true ) ) {
-				continue;
+			/*
+			 * Check if there are any active modules with compatible standalone plugins.
+			 * If no such modules are found bail early.
+			 */
+			if ( empty( $active_modules_with_plugins ) ) {
+				return;
 			}
 
 			$current_user = get_current_user_id();
 			$dismissed    = array_filter( explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) ) );
 
 			if ( ! in_array( 'perflab-module-migration-pointer', $dismissed, true ) ) {
-				continue;
+				return;
 			}
 
 			unset( $dismissed[ array_search( 'perflab-module-migration-pointer', $dismissed, true ) ] );
--- a/load.php
+++ b/load.php
@@ -462,6 +462,27 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
 			if ( ! empty( $module_settings['enabled'] ) && ( empty( $old_value[ $module ] ) || empty( $old_value[ $module ]['enabled'] ) ) ) {
 				perflab_activate_module( PERFLAB_PLUGIN_DIR_PATH . 'modules/' . $module );
 			}
+
+			// Check whether it is necessary to display an admin pointer to prompt the user to migrate.
+			$plugin_slug        = basename( $module );
+			$standalone_plugins = perflab_get_standalone_plugins();
+
+			// Plugin slug should be in standalone plugins list.
+			if ( ! in_array( $plugin_slug, $standalone_plugins, true ) ) {
+				continue;
+			}
+
+			$current_user = get_current_user_id();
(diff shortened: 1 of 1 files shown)


swissspidy (approved):

joemcgill (approved):
Seems sensible to me.

load.php
```diff
@@ -462,6 +462,27 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
...
+			$dismissed    = array_filter( explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) ) );
+
+			if ( ! in_array( 'perflab-module-migration-pointer', $dismissed, true ) ) {
+				continue;
+			}
+
+			unset( $dismissed[ array_search( 'perflab-module-migration-pointer', $dismissed, true ) ] );
+			$dismissed = implode( ',', $dismissed );
+
+			update_user_meta( $current_user, 'dismissed_wp_pointers', $dismissed );
```
felixarntz: This shouldn't be in the `foreach`, otherwise it'll unnecessarily run several times.

What we should probably do is put this logic after the `foreach`, and only run it if _any_ module was activated. We could set a flag around line 463 to indicate that a module was activated in this function.

load.php
```diff
@@ -462,6 +462,27 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
...
+			// Check whether it is necessary to display an admin pointer to prompt the user to migrate.
+			$plugin_slug        = basename( $module );
+			$standalone_plugins = perflab_get_standalone_plugins();
+
+			// Plugin slug should be in standalone plugins list.
+			if ( ! in_array( $plugin_slug, $standalone_plugins, true ) ) {
+				continue;
+			}
+
+			$current_user = get_current_user_id();
```
felixarntz, line 480: This will only work for the current user, but modules are active for the site. In other words, it's not working for all relevant users. We would need to use `WP_User_Query` for that, with a meta query for the `dismissed_wp_pointers` key (and possibly even with a `LIKE` for the `perflab-module-migration-pointer` value. That way we would unset the value for all relevant users.

Realistically, this is only a problem for sites that have more than 1 admin, which speaking at scale is not a ton. So I think we could also make this a future enhancement PR to unblock this. But I would suggest we implement it sooner or later.

felixarntz (changes requested):
Thank you for the PR @mukeshpanchal27. Looks close, I have one performance thing worth addressing here and another functional concern that could however also be addressed in a separate follow up PR.

load.php
mukeshpanchal27: Thank you for your feedback. I have incorporated the suggested changes in https://github.com/WordPress/performance/pull/915/commits/f56d80427584626df652d2e5c21c17b89d800827 based on your input.

load.php
mukeshpanchal27: While I appreciate the proposed solution, I'd like to highlight a potential challenge we might encounter. Consider a scenario where a site has two administrators, one catering to the non-technical team and the other to the technical team.

If we implement suggested changes. The behaviour with the migration pointer is as follows:

- Upon plugin release, the migration pointer becomes visible on the site.
- When the non-technical team logs in, they see the migration pointer.
- If they dismiss the pointer, it won't appear for them again.
- However, the technical team, when logging in, never encounters this migration pointer.

The concern with implementing the suggested solution is that it would remove the pointer for all users across the site, affecting both the non-technical and technical teams. This could potentially lead to the technical team missing out on crucial information conveyed by the pointer.

Looking forward to further discussion on this matter.

load.php
```diff
@@ -458,11 +458,37 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
...
 	// Get the list of modules that were activated, and load the activate.php files if they exist.
 	if ( ! empty( $value ) ) {
+		$enable_module_migration_pointer = false;
 		foreach ( $value as $module => $module_settings ) {
 			if ( ! empty( $module_settings['enabled'] ) && ( empty( $old_value[ $module ] ) || empty( $old_value[ $module ]['enabled'] ) ) ) {
 				perflab_activate_module( PERFLAB_PLUGIN_DIR_PATH . 'modules/' . $module );
+				$enable_module_migration_pointer = true;
 			}
 		}
+		if ( $enable_module_migration_pointer ) {
```
felixarntz: Maybe a more descriptive name would be `$reset_module_migration_pointer_dismissals` or (to shorten) `$reset_migration_pointer_dismissals`?
```suggestion
		$reset_migration_pointer_dismissals = false;
		foreach ( $value as $module => $module_settings ) {
			if ( ! empty( $module_settings['enabled'] ) && ( empty( $old_value[ $module ] ) || empty( $old_value[ $module ]['enabled'] ) ) ) {
				perflab_activate_module( PERFLAB_PLUGIN_DIR_PATH . 'modules/' . $module );
				$reset_migration_pointer_dismissals = true;
			}
		}
		if ( $reset_migration_pointer_dismissals ) {
```

load.php
felixarntz: > If we implement suggested changes. The behaviour with the migration pointer is as follows:
> 
> * Upon plugin release, the migration pointer becomes visible on the site.
> * When the non-technical team logs in, they see the migration pointer.
> * If they dismiss the pointer, it won't appear for them again.
> * However, the technical team, when logging in, never encounters this migration pointer.

That wouldn't be the case, maybe there was a misunderstanding or I didn't explain it well. The pointer wouldn't be dismissed for every user, it would only be "un-dismissed" for every user.

In other words:
* User A dismisses the pointer.
* User B still sees the pointer.
* User B performs the migration.
* Later, user B for some reason deactivates a standalone plugin and reactivates the corresponding module again.
* Now there's the difference:
    * Based on the current PR code, only user B would see the pointer again, but the problem applies to the entire site, so I think user A should see the pointer again as well.
    * In my proposal, both user B and user A would see the pointer again as a result of the change from user B, so user A would be aware of the problem and then could fix it (e.g. if user B goes on holidays).

felixarntz (approved):
@mukeshpanchal27 LGTM, but let's follow up on the user conversation, potentially with a follow up PR once we have agreed on the approach.

load.php
mukeshpanchal27: Thanks @felixarntz for explaining. Opened followup PR https://github.com/WordPress/performance/pull/917.

PR #917, opened 2023-12-22
Revise migration pointer un-dismissal to work for all admins

mukeshpanchal27:
## Summary

Follow-up to https://github.com/WordPress/performance/pull/915#discussion_r1432982498

Scenario that resolved by this PR:

> If we implement suggested changes. The behaviour with the migration pointer is as follows:
> 
> * Upon plugin release, the migration pointer becomes visible on the site.
> * When the non-technical team logs in, they see the migration pointer.
> * If they dismiss the pointer, it won't appear for them again.
> * However, the technical team, when logging in, never encounters this migration pointer.

That wouldn't be the case, maybe there was a misunderstanding or I didn't explain it well. The pointer wouldn't be dismissed for every user, it would only be "un-dismissed" for every user.

In other words:
* User A dismisses the pointer.
* User B still sees the pointer.
* User B performs the migration.
* Later, user B for some reason deactivates a standalone plugin and reactivates the corresponding module again.
* Now there's the difference:
    * Based on the current PR code, only user B would see the pointer again, but the problem applies to the entire site, so I think user A should see the pointer again as well.
    * In my proposal, both user B and user A would see the pointer again as a result of the change from user B, so user A would be aware of the problem and then could fix it (e.g. if user B goes on holidays).

## Checklist

### DIFF
--- a/load.php
+++ b/load.php
@@ -477,17 +477,37 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
 				return;
 			}
 
-			$current_user = get_current_user_id();
-			$dismissed    = array_filter( explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) ) );
+			$current_user       = wp_get_current_user();
+			$current_user_roles = array_shift( $current_user->roles );
 
-			if ( ! in_array( 'perflab-module-migration-pointer', $dismissed, true ) ) {
-				return;
+			$args = array(
+				'role'       => $current_user_roles,
+				'meta_query' => array(
+					array(
+						'key'     => 'dismissed_wp_pointers',
+						'value'   => 'perflab-module-migration-pointer',
+						'compare' => 'LIKE',
+					),
+				),
+			);
+
+			$user_query = new WP_User_Query( $args );
+			$get_users  = $user_query->get_results();
+
+			if ( ! empty( $get_users ) ) {
+				foreach ( $get_users as $user ) {
+					$dismissed = array_filter( explode( ',', (string) get_user_meta( $user->ID, 'dismissed_wp_pointers', true ) ) );
+
+					if ( ! in_array( 'perflab-module-migration-pointer', $dismissed, true ) ) {
+						return;
+					}
+
+					unset( $dismissed[ array_search( 'perflab-module-migration-pointer', $dismissed, true ) ] );
+					$dismissed = implode( ',', $dismissed );
+
+					update_user_meta( $user->ID, 'dismissed_wp_pointers', $dismissed );
+				}
 			}
-
-			unset( $dismissed[ array_search( 'perflab-module-migration-pointer', $dismissed, true ) ] );
-			$dismissed = implode( ',', $dismissed );
-
-			update_user_meta( $current_user, 'dismissed_wp_pointers', $dismissed );
 		}
 	}
 
--- a/load.php
+++ b/load.php
@@ -489,7 +489,8 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
 			if ( ! wp_is_large_user_count() ) {
 				perflab_dismissed_wp_pointers( $current_user );
 			} else {
-				$current_user_roles = array_shift( $current_user->roles );
+				$current_user_roles = $current_user->roles;
+				$current_user_role  = array_shift( $current_user_roles );
 
 				$args = array(
 					'role'       => $current_user_roles,
@@ -524,20 +525,21 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
 }
 
 /**
- * Handles dismissing a WordPress pointer.
+ * Reverts the module migration pointer dismissal for the given user.
  *
  * @since n.e.x.t
  *
  * @param WP_User $user The WP_User object.
  */
-function perflab_dismissed_wp_pointers( $user ) {
+function perflab_undismiss_module_migration_pointer( $user ) {
 	$dismissed = array_filter( explode( ',', (string) get_user_meta( $user->ID, 'dismissed_wp_pointers', true ) ) );
 
-	if ( ! in_array( 'perflab-module-migration-pointer', $dismissed, true ) ) {
+	$pointer_index = array_search( 'perflab-module-migration-pointer', $dismissed, true );
+	if ( false === $pointer_index ) {
 		return;
 	}
 
-	unset( $dismissed[ array_search( 'perflab-module-migration-pointer', $dismissed, true ) ] );
+	unset( $dismissed[ $pointer_index ] );
 	$dismissed = implode( ',', $dismissed );
 
 	update_user_meta( $user->ID, 'dismissed_wp_pointers', $dismissed );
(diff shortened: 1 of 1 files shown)


load.php
```diff
@@ -477,17 +477,37 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
...
 
-			$current_user = get_current_user_id();
-			$dismissed    = array_filter( explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) ) );
-
-			if ( ! in_array( 'perflab-module-migration-pointer', $dismissed, true ) ) {
-				return;
+			$current_user       = wp_get_current_user();
+			$current_user_roles = array_shift( $current_user->roles );
+
+			$args = array(
```
swissspidy: That seems like a potentially very slow query to me

mukeshpanchal27: Thank you for asking this. Yes, using the LIKE comparison method slows down the query due to the serialized storage of that meta value. Is there an alternative approach available to access that value? I'm eager to explore any potential options. Thank you!

felixarntz: I don't think there's a faster way to query this using WordPress methods. IMO it's okay to use this, but we should definitely limit the results using the `number` argument, to avoid problems at scale. Unlikely, but if a site had 100s of administrators, we'd probably run into memory limit issues.

I think we could just limit to something like 20, I think it's unlikely for the vast majority of sites to have more than 20 administrators.

load.php
```diff
@@ -477,17 +477,37 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
...
+					array(
+						'key'     => 'dismissed_wp_pointers',
+						'value'   => 'perflab-module-migration-pointer',
+						'compare' => 'LIKE',
+					),
+				),
+			);
+
+			$user_query = new WP_User_Query( $args );
+			$get_users  = $user_query->get_results();
```
felixarntz: Better name: `$users`

The variable contains the users, it's a bit of an odd convention to put a verb before it.

load.php
```diff
@@ -477,17 +477,37 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
...
+
+			if ( ! empty( $get_users ) ) {
```
felixarntz: Not needed, we can simply use the `foreach`, which will just do nothing if the list is empty.

load.php
```diff
@@ -477,17 +477,37 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
...
+
+			$args = array(
+				'role'       => $current_user_roles,
+				'meta_query' => array(
+					array(
+						'key'     => 'dismissed_wp_pointers',
+						'value'   => 'perflab-module-migration-pointer',
+						'compare' => 'LIKE',
+					),
+				),
```
felixarntz: We may want to use `get_users()` instead of manually relying on `WP_User_Query`. It makes the code a bit simpler and also automatically sets `count_total` to `false`, which makes the query faster.

load.php
felixarntz: The only alternative I see would be to run a completely custom query against `usermeta`. But then we'd still have to manually clear caches, and it becomes a bit more error-prone. I'm not sure there's a real risk here with a potentially slow query, as it would only happen explicitly on the module activation action, so it's probably overkill.

How about we only run the `WP_User_Query` if `! wp_is_large_user_count()`? That would at least avoid problems for sites with tons of users. For those, we could fall back to just dealing with the current user.

felixarntz (changes requested):
@mukeshpanchal27 Mostly looks great! A few points of feedback, most importantly regarding the query approach.

load.php
mukeshpanchal27: Updated in https://github.com/WordPress/performance/pull/917/commits/f4fcc72de9bf6475098be8920f942e8ef7c21872

load.php
mukeshpanchal27: Thank you for sharing your thoughts. I have updated the query approach based on the feedback received. The PR is now ready for review.

load.php
```diff
@@ -503,6 +523,26 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
...
 }
 
+/**
+ * Handles dismissing a WordPress pointer.
+ *
+ * @since n.e.x.t
+ *
+ * @param WP_User $user The WP_User object.
+ */
+function perflab_dismissed_wp_pointers( $user ) {
```
felixarntz: I think the function name and description aren't entirely accurate:
* The function doesn't dismiss a WordPress pointer, it "un-dismisses" a previously dismissed WordPress pointer.
* It doesn't just handle _any_ pointer, but specifically the module migration pointer.

I would suggest something like:
```suggestion
/**
 * Reverts the module migration pointer dismissal for the given user.
 *
 * @since n.e.x.t
 *
 * @param WP_User $user The WP_User object.
 */
function perflab_undismiss_module_migration_pointer( $user ) {
```

load.php
```diff
@@ -503,6 +523,26 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
...
+	$dismissed = array_filter( explode( ',', (string) get_user_meta( $user->ID, 'dismissed_wp_pointers', true ) ) );
+
+	if ( ! in_array( 'perflab-module-migration-pointer', $dismissed, true ) ) {
+		return;
+	}
+
+	unset( $dismissed[ array_search( 'perflab-module-migration-pointer', $dismissed, true ) ] );
```
felixarntz: Slightly cleaner would be to only call `array_search()` which can be used alone to get the same information.
```suggestion
	$pointer_index = array_search( 'perflab-module-migration-pointer', $dismissed, true );
	if ( false === $pointer_index ) {
		return;
	}

	unset( $dismissed[ $pointer_index ] );
```

load.php
```diff
@@ -477,17 +477,37 @@ function perflab_run_module_activation_deactivation( $old_value, $value ) {
...
+			 * Disable WordPress pointers for specific users based on conditions.
+			 *
+			 * Checks if there is a large user count on the site. If true,
+			 * disables pointers for the current user only. Otherwise, disables
+			 * pointers for users with the same role as the current user.
+			 */
+			if ( ! wp_is_large_user_count() ) {
+				perflab_dismissed_wp_pointers( $current_user );
+			} else {
+				$current_user_roles = array_shift( $current_user->roles );
```
felixarntz: The `array_shift()` is problematic as it modifies the actual `$current_user` object, which we need to avoid. On another note, the variable should probably be renamed as it only contains a single role, not multiple. Here's an alternative solution that doesn't modify the original object property.
```suggestion
				$current_user_roles = $current_user->roles;
				$current_user_role  = array_shift( $current_user_roles );
```

felixarntz (commented):
@mukeshpanchal27 The approach looks good to me now, there are only a few last quirks to iron out.

mukeshpanchal27:
@swissspidy @felixarntz PR is ready for review.

felixarntz (approved):

felixarntz:
@swissspidy Could you give this another look when you get a chance?

swissspidy (approved):

PR #920, opened 2024-01-05
Load migration script in plugin setting page

mukeshpanchal27:
## Summary

In #899, we incorporated logic to load the migration script only when any module activated with the associated standalone plugins.

However, in #900, we further enhanced the experience when installing and activating standalone plugins. There's a need to ensure that this enhancement still functions properly if a module is deactivated and someone tries to install and activate the standalone plugins from the settings page.

### To replicate the issue:

1. Log in to the admin panel.
2. Go to Settings > Performance and deactivate all modules.
3. Install any module from the 'standalone performance plugins' list and activate the plugin.
4. Observe if it redirects to the plugin list page instead of settings page.

## Checklist


swissspidy (approved):

felixarntz:
Thanks @mukeshpanchal27, good catch!

Regarding the fix though, I think there would be a cleaner solution:
* The only reason we would need to load that script regardless of the migration is for the plugin installation enhancement. That part however has nothing to do with module migration anyway. So it's arguably not in the right JS file, or the JS file has a misleading name.
* Therefore, I would suggest we move the logic related to the plugin installation/activation into another JS file (e.g. something like `perflab-plugin-management.js`). Then we can load that JS file unconditionally, while continuing to load the `perflab-module-migration-notice.js` file only when relevant.

mukeshpanchal27:
Thank you for the feedback! In the commit https://github.com/WordPress/performance/pull/920/commits/abc39aacc235f21dc2b4ad46832d48722b8cfc8c, I've incorporated the new JS as per your suggestion and relocated both JS file into a designated `JS` folder.

swissspidy (approved):

felixarntz (approved):
This looks great, thank you @mukeshpanchal27!

PR #922, opened 2024-01-05
Prepare the auto-sizes module as a standalone plugin

joemcgill:
## Summary

This updates the auto-sizes module feature to prepare for merging into trunk. Included in this change:

- Updates the hook function names
- Adds a plugin readme.txt file to the module folder
- Fixes the Plugin header name

See: #904 

## Checklist

### DIFF
--- a/modules/images/auto-sizes/load.php
+++ b/modules/images/auto-sizes/load.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Module Name: Auto Sizes for Lazy-loaded Images.
+ * Module Name: Auto Sizes for Lazy-loaded Images
  * Description: Implements auto sizes for lazy loaded images.
  * Experimental: Yes
  *
--- a/modules/images/auto-sizes/load.php
+++ b/modules/images/auto-sizes/load.php
@@ -1,7 +1,7 @@
 <?php
 /**
- * Module Name: Auto Sizes for Lazy-loaded Images
- * Description: Implements auto sizes for lazy loaded images.
+ * Module Name: Auto-sizes for Lazy-loaded Images
+ * Description: This plugin implements the HTML spec for adding `sizes="auto"` to lazy-loaded images.
  * Experimental: Yes
  *
  * @package performance-lab
--- a/modules/images/auto-sizes/readme.txt
+++ b/modules/images/auto-sizes/readme.txt
@@ -1,4 +1,4 @@
-=== Auto Sizes for Lazy-loaded Images ===
+=== Auto-sizes for Lazy-loaded Images ===
 
 Contributors:      wordpressdotorg
 Requires at least: 6.4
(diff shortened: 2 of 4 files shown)


modules/images/auto-sizes/readme.txt
```diff
@@ -0,0 +1,53 @@
...
+Stable tag:        1.0.0
+License:           GPLv2 or later
+License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+Tags:              performance, images, auto sizes
+
+Adds support for automatically calculating the sizes attribute for lazy-loaded images.
+
+== Description ==
+
+This plugin implements the HTML spec for auto sizes for lazy-loaded images.
```
felixarntz: Should we maybe include a mention of `sizes="auto"` here or elsewhere? While "Auto sizes" is a decent name, there's a good chance a developer will have a better idea of what this does when seeing `sizes="auto"`, as only that clarifies it's about the "sizes" _attribute_. "Auto sizes" could also have to do with e.g. the generated image sizes or other things.

I don't want to get hung up on that, but I think at least one mention of the attribute would be helpful in clarifying. For instance:
```suggestion
This plugin implements the HTML spec for `sizes="auto"` for lazy-loaded images.
```

felixarntz (approved):
@joemcgill LGTM, just one readme docs recommendation.

modules/images/auto-sizes/readme.txt
joemcgill: I'm happy to go with this. Thanks!

modules/images/auto-sizes/load.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Module Name: Auto Sizes for Lazy-loaded Images.
+ * Module Name: Auto Sizes for Lazy-loaded Images
```
mukeshpanchal27: It's not a blocker, but in the [CODEOWNERS file](https://github.com/WordPress/performance/blob/feature/auto-sizes/.github/CODEOWNERS#L57), we have `Auto-sizes for Lazy-Loaded Images` specified, and the module name differs from this designation. Should we consider correcting this discrepancy for consistency?

mukeshpanchal27 (approved):
Thank you, @joemcgill, for the PR! I've provided one non-blocking feedback. Aside from that, everything appears to be in good from my perspective.

modules/images/auto-sizes/load.php
joemcgill: Good observation! Looking at the official spec and various references from early browser implementers, I see this both hyphenated and not, but it's good for us to be consistent, so I've updated all references to be hyphenated in [6870baf](https://github.com/WordPress/performance/pull/922/commits/6870baf52050e5e1240db347cd7d55670b4d58bb).

PR #927, opened 2024-01-10
Fix plugin deactivation link color

felixarntz:
## Summary

The plugin deactivation link is currently bright red (via a simple `color: red` hard-coded in a `style` attribute). This is not a good color choice, as it is very bright and doesn't follow the WordPress admin color scheme. It should instead use `#b32d2e` which is used for other deactivation links on a white background (e.g. in the plugins list table).

This PR fixes that, and as part of it also avoids the `style` attribute and instead colocates the style with the other styles for the admin page.

This PR uses `no milestone` to not include it in changelogs, since for end users this bug is not relevant - it only will be if we don't fix it now :)

## Checklist

### DIFF
--- a/admin/plugins.php
+++ b/admin/plugins.php
@@ -194,13 +194,15 @@ function perflab_render_plugin_card( array $plugin_data ) {
 
 					$action_links[] = sprintf(
 						'<a href="%s" id="deactivate-%s" aria-label="%s">%s</a>',
-						add_query_arg(
-							array(
-								'_wpnonce' => wp_create_nonce( 'perflab_deactivate_plugin_' . $status['file'] ),
-								'action'   => 'perflab_deactivate_plugin',
-								'plugin'   => $status['file'],
-							),
-							network_admin_url( 'plugins.php' )
+						esc_url(
+							add_query_arg(
+								array(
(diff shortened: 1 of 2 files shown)


swissspidy (approved):

admin/plugins.php
```diff
@@ -193,7 +193,7 @@ function perflab_render_plugin_card( array $plugin_data ) {
 					$context = $status['status'];
 
 					$action_links[] = sprintf(
-						'<a href="%s" id="deactivate-%s" aria-label="%s" style="color:red;text-decoration: underline;">%s</a>',
+						'<a href="%s" id="deactivate-%s" aria-label="%s">%s</a>',
 						add_query_arg(
```
westonruter: This should be wrapped in `esc_url()` while you're at it.

westonruter (changes requested):

admin/plugins.php
felixarntz: Addressed in https://github.com/WordPress/performance/pull/927/commits/a36f6b157cc6f4a01e2ecb7f246cad1f759ceb74

westonruter (approved):

PR #930, opened 2024-01-10
Fix JS linting

westonruter:
## Summary

This cherry-picks changes from `feature/image-loading-optimization` to fix linting of JS files that aren't inside of `bin`.

## Relevant technical choices

* Include `browser` in `env` for ESLint.
* Include JS files in `lint-staged`.
* Run `format-js` on the JS files.
* Include all JS files when running `lint-js` not just `bin`.
* Disable certain ESLint rules which do not need to be fixed.
* Fix `no-redeclare` ESLint error by using `let` in `for` loops.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,8 +27,8 @@
     "test-plugins": "./bin/plugin/cli.js test-plugins",
     "test-plugins-multisite": "./bin/plugin/cli.js test-plugins --sitetype=multi",
     "enabled-modules": "./bin/plugin/cli.js enabled-modules",
-    "format-js": "wp-scripts format ./bin",
-    "lint-js": "wp-scripts lint-js ./bin",
+    "format-js": "wp-scripts format",
+    "lint-js": "wp-scripts lint-js",
     "format-php": "wp-env run composer run-script format",
     "phpstan": "wp-env run composer run-script phpstan",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -27,8 +27,8 @@
     "test-plugins": "./bin/plugin/cli.js test-plugins",
     "test-plugins-multisite": "./bin/plugin/cli.js test-plugins --sitetype=multi",
     "enabled-modules": "./bin/plugin/cli.js enabled-modules",
-    "format-js": "wp-scripts format ./bin",
-    "lint-js": "wp-scripts lint-js ./bin",
+    "format-js": "wp-scripts format",
```
joemcgill, line 30: This is a good catch. Interesting that these were previously only running against the `./bin` directory. I wonder if there was a reason for that limitation previously?

joemcgill (approved):
This looks like a good improvement to me.

package.json
westonruter: I think it's because there were no JS files elsewhere in the project before.

felixarntz: Yes, for a long time the only JS in the repo were the workflow files. Makes sense to loosen that now.

felixarntz (approved):

PR #938, opened 2024-01-15
Add `readme.txt` for Speculation Rules plugin

### LINKED ISSUE #905: Add `readme.txt` for Speculation Rules
A `readme.txt` file is needed for the new module (soon to be plugin).

felixarntz:
## Summary

Fixes #905

## Relevant technical choices

* The readme already includes references to the UI added in #939.
* Includes #937, just to prevent irrelevant lint failure.

## Checklist

### DIFF
--- a/modules/js-and-css/speculation-rules/readme.txt
+++ b/modules/js-and-css/speculation-rules/readme.txt
@@ -15,9 +15,9 @@ Uses the Speculation Rules API to prerender linked URLs upon hover.
 
 This plugin adds support for the [Speculation Rules API](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API), which allows defining rules by which certain URLs are dynamically prefetched or prerendered based on user interaction.
 
-By default, the plugin is configured to prerender WordPress frontend URLs when the user hovers over a relevant link.
+By default, the plugin is configured to prerender WordPress frontend URLs when the user hovers over a relevant link. This can be customized via the "Speculation Rules" section under _Settings > Reading_.
 
-A filter can be used to exclude certain URL paths from being eligible for prefetching and prerendering (see FAQ section).
+A filter can be used to exclude certain URL paths from being eligible for prefetching and prerendering (see FAQ section). Alternatively, you can add the 'no-prerender' CSS class to any link (`<a>` tag) that should not be prerendered.
 
 = Browser support =
 
--- a/modules/js-and-css/speculation-rules/readme.txt
+++ b/modules/js-and-css/speculation-rules/readme.txt
@@ -1,0 +1,82 @@
+=== Speculation Rules ===
+
+Contributors:      wordpressdotorg
+Requires at least: 6.3
+Tested up to:      6.4
+Requires PHP:      7.0
+Stable tag:        1.0.0
+License:           GPLv2 or later
+License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+Tags:              performance, javascript, speculation rules, prerender, prefetch
+
+Uses the Speculation Rules API to prerender linked URLs upon hover.
+
+== Description ==
+
+This plugin adds support for the [Speculation Rules API](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API), which allows defining rules by which certain URLs are dynamically prefetched or prerendered based on user interaction.
+
+By default, the plugin is configured to prerender WordPress frontend URLs when the user hovers over a relevant link.
+
+A filter can be used to exclude certain URL paths from being eligible for prefetching and prerendering (see FAQ section).
+
+= Browser support =
+
+The Speculation Rules API is a new web API and currently requires using Chrome 121+.
+
+Other browsers will not see any adverse effects, however the feature will not work for those clients.
+
+* [Browser support for the Speculation Rules API in general](https://caniuse.com/mdn-html_elements_script_type_speculationrules)
+* [Information on document rules syntax support used by the plugin](https://developer.chrome.com/blog/chrome-121-beta#speculation_rules_api)
+
+== Installation ==
+
+= Installation from within WordPress =
+
+1. Visit **Plugins > Add New**.
+2. Search for **Speculation Rules**.
+3. Install and activate the **Speculation Rules** plugin.
+
+= Manual installation =
+
+1. Upload the entire `speculation-rules` folder to the `/wp-content/plugins/` directory.
+2. Visit **Plugins**.
+3. Activate the **Speculation Rules** plugin.
+
+== Frequently Asked Questions ==
+
+= How can I prevent certain URLs from being prefetched and prerendered? =
+
+Not every URL can be reasonably prerendered. Prerendering static content is typically reliable, however prerendering dynamic content, such as a shopping cart or a checkout page, can lead to issues. You can prevent certain URL paths from being prefetched or prerendered by using the `plsr_speculation_rules_href_exclude_paths` filter.
+
+This example would ensure that URLs like `https://example.com/cart/` or `https://example.com/cart/foo` would be excluded from prefetching and prerendering.
+`
+<?php
+
+add_filter(
+	'plsr_speculation_rules_href_exclude_paths',
+	function ( $exclude_paths ) {
+		$exclude_paths[] = '/cart/*';
+		return $exclude_paths;
+	}
+);
+`
+
+= Where can I submit my plugin feedback? =
+
+Feedback is encouraged and much appreciated, especially since this plugin may contain future WordPress core features. If you have suggestions or requests for new features, you can [submit them as an issue in the WordPress Performance Team's GitHub repository](https://github.com/WordPress/performance/issues/new/choose). If you need help with troubleshooting or have a question about the plugin, please [create a new topic on our support forum](https://wordpress.org/support/plugin/speculation-rules/#new-topic-0).
+
+= Where can I report security bugs? =
+
+The Performance team and WordPress community take security bugs seriously. We appreciate your efforts to responsibly disclose your findings, and will make every effort to acknowledge your contributions.
+
+To report a security issue, please visit the [WordPress HackerOne](https://hackerone.com/wordpress) program.
+
+= How can I contribute to the plugin? =
+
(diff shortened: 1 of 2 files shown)


swissspidy (approved):

modules/js-and-css/speculation-rules/readme.txt
```diff
@@ -0,0 +1,84 @@
...
+A filter can be used to exclude certain URL paths from being eligible for prefetching and prerendering (see FAQ section). Alternatively, you can add the 'no-prerender' CSS class to any link (`<a>` tag) that should not be prerendered.
+
+= Browser support =
+
+The Speculation Rules API is a new web API and currently requires using Chrome 121+.
+
+Other browsers will not see any adverse effects, however the feature will not work for those clients.
+
+* [Browser support for the Speculation Rules API in general](https://caniuse.com/mdn-html_elements_script_type_speculationrules)
+* [Information on document rules syntax support used by the plugin](https://developer.chrome.com/blog/chrome-121-beta#speculation_rules_api)
```
felixarntz: @domenic @jeremyroman Would be great to get a quick check from you on the wording and explanations here.

modules/js-and-css/speculation-rules/readme.txt
```diff
@@ -0,0 +1,84 @@
...
+Contributors:      wordpressdotorg
+Requires at least: 6.3
+Tested up to:      6.4
+Requires PHP:      7.0
+Stable tag:        1.0.0
+License:           GPLv2 or later
+License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+Tags:              performance, javascript, speculation rules, prerender, prefetch
+
+Uses the Speculation Rules API to prerender linked URLs upon hover.
```
jeremyroman: It says "prerender" here but "prefetch or prerender" (or words to that effect) everywhere else. Should we be consistent?

modules/js-and-css/speculation-rules/readme.txt
```diff
@@ -0,0 +1,84 @@
...
+
+1. Upload the entire `speculation-rules` folder to the `/wp-content/plugins/` directory.
+2. Visit **Plugins**.
+3. Activate the **Speculation Rules** plugin.
+
+== Frequently Asked Questions ==
+
+= How can I prevent certain URLs from being prefetched and prerendered? =
+
+Not every URL can be reasonably prerendered. Prerendering static content is typically reliable, however prerendering dynamic content, such as a shopping cart or a checkout page, can lead to issues. You can prevent certain URL paths from being prefetched or prerendered by using the `plsr_speculation_rules_href_exclude_paths` filter.
```
jeremyroman: This example feels slightly confusing to me.

Content that is server-side dynamic can cause surprises with either prefetching or prerendering, if the user is likely to noticeably change the state between preloading and navigation.

Prerendering additionally causes client-side code to run before navigation, which might cause surprises if there is JavaScript that does things that shouldn't happen yet.

Stuff like a shopping cart seems more likely to be an example of the former. Right?

modules/js-and-css/speculation-rules/readme.txt
felixarntz: The "prerender on hover" is the default behavior, that's why I chose to go with that here.

modules/js-and-css/speculation-rules/readme.txt
felixarntz: Makes sense. Do you have another suggestion for an example that could be used here?

modules/js-and-css/speculation-rules/readme.txt
tunetheweb: Could we add "by default (with configuration options)" then to make that super obvious?

modules/js-and-css/speculation-rules/readme.txt
```diff
@@ -0,0 +1,84 @@
...
+
+[See the Speculation Rules WICG specification draft.](https://wicg.github.io/nav-speculation/speculation-rules.html)
+
+By default, the plugin is configured to prerender WordPress frontend URLs when the user hovers over a relevant link. This can be customized via the "Speculation Rules" section under _Settings > Reading_.
+
+A filter can be used to exclude certain URL paths from being eligible for prefetching and prerendering (see FAQ section). Alternatively, you can add the 'no-prerender' CSS class to any link (`<a>` tag) that should not be prerendered.
+
+= Browser support =
+
+The Speculation Rules API is a new web API and currently requires using Chrome 121+.
```
tunetheweb:
```suggestion
The Speculation Rules API is a relatively new web API and currently requires using Chrome 121+ (or equivalent Chromium-based browser) for the features of this that we are using.
```

The Speculation Rules API was available well before 121 but not for this feature. So suggest this to make it clearer.

Also it will be available in other browsers (not sure if from 121 version of those - @jeremyroman when you say "enabled by default from 121" do you mean no longer in Finch or still controlled by Finch so not yet available in other Chromium browsers based off of 121?

modules/js-and-css/speculation-rules/readme.txt
tunetheweb: Logout link?

modules/js-and-css/speculation-rules/readme.txt
felixarntz: Adding "by default" makes sense to me, will update. I think the "with configuration options" bit is too long, and is explained below, so IMO that's sufficient.

modules/js-and-css/speculation-rules/readme.txt
felixarntz: Thanks, I'll update this and try to clarify.

modules/js-and-css/speculation-rules/readme.txt
felixarntz: While the API was available before, the relevant functionality of it which this plugin is based upon (document rules) are only available in Chrome 121+ as far as I'm aware (before they required origin trial, so not really live). So while older versions still have the API, they wouldn't be able to use this plugin.

I'll try to clarify that the specific functionality requires Chrome 121+.

felixarntz:
@jeremyroman @tunetheweb I've pushed some content updates in https://github.com/WordPress/performance/pull/938/commits/671eb6b15b66aec0adadc380a64306cd4757f1ed to address your feedback, would be great if you could take another look.

tunetheweb (approved):
LGTM. But would be good to get clarification from @jeremyroman whether it's Chrome 121+ or Chromium 121+

felixarntz:
@tunetheweb With the typically less tech savvy audience of WordPress, I'd prefer to refer to common browsers, so if it's Chromium, maybe we should say something like "Chrome 121+ or Edge X+" (I assume this will be in a new Edge release too)?

tunetheweb:
> @tunetheweb With the typically less tech savvy audience of WordPress, I'd prefer to refer to common browsers, so if it's Chromium, maybe we should say something like "Chrome 121+ or Edge X+"

Yeah that's fair.

> (I assume this will be in a new Edge release too)?

That's my question! I don't think it's a safe assumption as it might still technical be a Chrome experiment with 100% of Chrome users (which therefore won't be available in Edge), rather than a Chromium default (which would be available in Edge). Typically we go 100% in Chrome first and then roll out to Chromium later by making it the actual default. But not 100% sure hence why I'm asking Jeremy.

modules/js-and-css/speculation-rules/readme.txt
```diff
@@ -0,0 +1,84 @@
...
+License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+Tags:              performance, javascript, speculation rules, prerender, prefetch
+
+Uses the Speculation Rules API to prerender linked URLs upon hover by default.
+
+== Description ==
+
+This plugin adds support for the [Speculation Rules API](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API), which allows defining rules by which certain URLs are dynamically prefetched or prerendered based on user interaction.
+
+[See the Speculation Rules WICG specification draft.](https://wicg.github.io/nav-speculation/speculation-rules.html)
```
westonruter:
```suggestion
See the [Speculation Rules WICG specification draft](https://wicg.github.io/nav-speculation/speculation-rules.html).
```

modules/js-and-css/speculation-rules/readme.txt
jeremyroman: The functionality mentioned here is out of origin trial and enabled by default (not only by Finch) in Chrome 121. And the code substantially lives in content/.

So other Chromium-based browsers would get it unless they disabled it. For instance, it works in Microsoft Edge 121 (beta) right now (just checked on my Windows VM), so unless Microsoft disables it between now and when that ships to stable, it will be available in Edge as well.

Some tweaks to the heuristics might ship outside Chromium proper and so be Google Chrome-specific, but not the availability of the API itself.

tunetheweb: Good to know. Whether you wanna mention this is up to you @felixarntz or if "Chrome 121" is good enough for your user base (and those that know will know that also means things like Edge soon enough after).

jeremyroman:
> > @tunetheweb With the typically less tech savvy audience of WordPress, I'd prefer to refer to common browsers, so if it's Chromium, maybe we should say something like "Chrome 121+ or Edge X+"
> 
> Yeah that's fair.
> 
> > (I assume this will be in a new Edge release too)?
> 
> That's my question! I don't think it's a safe assumption as it might still technical be a Chrome experiment with 100% of Chrome users (which therefore won't be available in Edge), rather than a Chromium default (which would be available in Edge). Typically we go 100% in Chrome first and then roll out to Chromium later by making it the actual default. But not 100% sure hence why I'm asking Jeremy.

AFAIK most web platform features in Chrome enable by default rather than launching via Finch (with some notable exceptions), and this is going that route. It works in Edge 121 beta, so it seems likely it will work in Edge 121 stable too.

modules/js-and-css/speculation-rules/readme.txt
felixarntz: Maybe we go with just "Chrome 121" for now (i.e. keep this as is) and modify if we're certain Edge support is present when that version launches? That'll most likely still be before this plugin is published anyway, so we can just fix it in another PR.

westonruter (approved):

PR #939, opened 2024-01-16
Add Speculation Rules settings and UI

### LINKED ISSUE #906: Add settings and UI to control Speculation Rules prerender vs prefetch and eagerness
See #897:
* UI controls should be provided in WP Admin to change the behavior to prefetch URLs instead of prerendering, and another one to change the behavior from prefetching/prerendering on hover to only do it on click.
    * Both alternatives are less beneficial from a performance perspective, but potentially preferable for other reasons: For example, prerendering can be problematic with dynamic content, and relying on hover can result in wastefully loaded pages, if the user doesn't actually end up navigating there.
    * As this is a Performance Lab module, it makes sense with the more performant setting by default. For a potential future WordPress core proposal, the default would have to be determined based on user feedback, and UI controls would most likely not be part of the WordPress core implementation. For early feedback, they are crucial though to make it easy to experiment with different configurations.

felixarntz:
## Summary

Fixes #906

## Relevant technical choices

* Setting uses an associative array, to avoid using multiple options in the database.
* UI to control the settings is added to _Settings > Reading_.
* See [the spec](https://wicg.github.io/nav-speculation/speculation-rules.html#valid-eagerness-strings) for additional context on the available choices.
* Test coverage is included.
* Includes #937, just to prevent irrelevant lint failure.

## Checklist

### DIFF
--- a/modules/js-and-css/speculation-rules/helper.php
+++ b/modules/js-and-css/speculation-rules/helper.php
@@ -60,7 +60,11 @@ function plsr_get_speculation_rules() {
 		)
 	);
 
-	$prerender_rules = array(
+	$option    = get_option( 'plsr_speculation_rules' );
+	$mode      = $option['mode'];
+	$eagerness = $option['eagerness'];
+
+	$rules = array(
 		array(
 			'source'    => 'document',
 			'where'     => array(
@@ -83,9 +87,9 @@ function plsr_get_speculation_rules() {
 					),
 				),
 			),
-			'eagerness' => 'moderate',
+			'eagerness' => $eagerness,
 		),
 	);
 
-	return array( 'prerender' => $prerender_rules );
+	return array( $mode => $rules );
 }
--- a/modules/js-and-css/speculation-rules/helper.php
+++ b/modules/js-and-css/speculation-rules/helper.php
@@ -83,17 +83,20 @@ function plsr_get_speculation_rules() {
 							'href_matches' => $href_exclude_paths,
 						),
 					),
-					// And except for any links marked with a class to not prerender.
-					array(
-						'not' => array(
-							'selector_matches' => '.no-prerender',
-						),
-					),
 				),
 			),
 			'eagerness' => $eagerness,
 		),
 	);
+
+	// Allow adding a class on any links to prevent prerendering.
+	if ( 'prerender' === $mode ) {
+		$rules[0]['where']['and'][] = array(
+			'not' => array(
+				'selector_matches' => '.no-prerender',
+			),
+		);
+	}
 
 	return array( $mode => $rules );
 }
--- a/modules/js-and-css/speculation-rules/helper.php
+++ b/modules/js-and-css/speculation-rules/helper.php
@@ -60,7 +60,11 @@ function plsr_get_speculation_rules() {
 		)
 	);
 
-	$option    = get_option( 'plsr_speculation_rules' );
+	$option = get_option( 'plsr_speculation_rules' );
+	if ( ! $option ) {
+		$option = plsr_get_setting_default();
+	}
+
 	$mode      = $option['mode'];
 	$eagerness = $option['eagerness'];
 
--- a/modules/js-and-css/speculation-rules/helper.php
+++ b/modules/js-and-css/speculation-rules/helper.php
@@ -51,8 +51,15 @@ function plsr_get_speculation_rules() {
 	);
 
 	$option = get_option( 'plsr_speculation_rules' );
-	if ( ! $option ) {
+
+	/*
+	 * This logic is only relevant for edge-cases where the setting may not be registered,
+	 * a.k.a. defensive coding.
+	 */
+	if ( ! $option || ! is_array( $option ) ) {
 		$option = plsr_get_setting_default();
+	} else {
+		$option = array_merge( plsr_get_setting_default(), $option );
 	}
 
 	$mode      = $option['mode'];
--- a/modules/js-and-css/speculation-rules/settings.php
+++ b/modules/js-and-css/speculation-rules/settings.php
@@ -11,7 +11,7 @@
  *
  * @since n.e.x.t
  *
- * @return array Associative array of `$mode => $label` pairs.
+ * @return array<string, string> Associative array of `$mode => $label` pairs.
  */
 function plsr_get_mode_labels() {
 	return array(
@@ -25,7 +25,7 @@ function plsr_get_mode_labels() {
  *
  * @since n.e.x.t
  *
- * @return array Associative array of `$eagerness => $label` pairs.
+ * @return array<string, string> Associative array of `$eagerness => $label` pairs.
  */
 function plsr_get_eagerness_labels() {
 	return array(
@@ -55,7 +55,12 @@ function plsr_get_setting_default() {
  * @since n.e.x.t
  *
  * @param mixed $input Setting to sanitize.
- * @return array Sanitized setting, an associative array with 'mode' and 'eagerness' keys.
+ * @return array<string, string> {
+ *     Sanitized setting.
+ *
+ *     @type string $mode      Mode.
+ *     @type string $eagerness Eagerness.
+ * }
  */
 function plsr_sanitize_setting( $input ) {
 	$default_value = plsr_get_setting_default();
--- a/modules/js-and-css/speculation-rules/settings.php
+++ b/modules/js-and-css/speculation-rules/settings.php
@@ -29,8 +29,8 @@ function plsr_get_mode_labels() {
  */
 function plsr_get_eagerness_labels() {
 	return array(
-		'conservative' => _x( 'Conservative (on click)', 'setting label', 'performance-lab' ),
-		'moderate'     => _x( 'Moderate (on hover)', 'setting label', 'performance-lab' ),
+		'conservative' => _x( 'Conservative (typically on click)', 'setting label', 'performance-lab' ),
+		'moderate'     => _x( 'Moderate (typically on hover)', 'setting label', 'performance-lab' ),
 		'eager'        => _x( 'Eager (on slightest suggestion)', 'setting label', 'performance-lab' ),
 	);
 }
--- a/modules/js-and-css/speculation-rules/settings.php
+++ b/modules/js-and-css/speculation-rules/settings.php
@@ -140,7 +140,7 @@ function plsr_add_setting_ui() {
 	$fields = array(
 		'mode'      => array(
 			'title'       => __( 'Mode', 'performance-lab' ),
-			'description' => __( 'Prerendering will lead to faster load times than prefetching, however in case of interactive content prefetching may be a safer choice.', 'performance-lab' ),
+			'description' => __( 'Prerendering will lead to faster load times than prefetching. However, in case of interactive content, prefetching may be a safer choice.', 'performance-lab' ),
 		),
 		'eagerness' => array(
 			'title'       => __( 'Eagerness', 'performance-lab' ),
--- a/modules/js-and-css/speculation-rules/settings.php
+++ b/modules/js-and-css/speculation-rules/settings.php
@@ -40,7 +40,12 @@ function plsr_get_eagerness_labels() {
  *
  * @since n.e.x.t
  *
- * @return array Default value, an associative array with 'mode' and 'eagerness' keys.
+ * @return array<string, string> {
+ *     Default setting value.
+ *
+ *     @type string $mode      Mode.
+ *     @type string $eagerness Eagerness.
+ * }
  */
 function plsr_get_setting_default() {
 	return array(
@@ -175,7 +180,13 @@ add_action( 'load-options-reading.php', 'plsr_add_setting_ui' );
  * @since n.e.x.t
  * @access private
  *
- * @param array $args Associative array with 'field', 'title', and optional 'description' keys.
+ * @param array<string, string> $args {
+ *     Associative array of arguments.
+ *
+ *     @type string $field       The slug of the sub setting controlled by the field.
+ *     @type string $title       The title for the field.
+ *     @type string $description Optional. A description to show for the field.
+ * }
  */
 function plsr_render_settings_field( array $args ) {
 	if ( empty( $args['field'] ) || empty( $args['title'] ) ) { // Invalid.
--- a/modules/js-and-css/speculation-rules/settings.php
+++ b/modules/js-and-css/speculation-rules/settings.php
@@ -139,13 +139,13 @@ function plsr_add_setting_ui() {
 
 	$fields = array(
 		'mode'      => array(
-			'title'       => __( 'Mode', 'performance-lab' ),
+			'title'       => __( 'Speculation Mode', 'performance-lab' ),
 			'description' => __( 'Prerendering will lead to faster load times than prefetching. However, in case of interactive content, prefetching may be a safer choice.', 'performance-lab' ),
 		),
 		'eagerness' => array(
 			'title'       => __( 'Eagerness', 'performance-lab' ),
 			'description' => __( 'The eagerness setting defines the heuristics based on which the loading is triggered.', 'performance-lab' )
-				. '<br>' . __( '"Eager" will have the minimum delay to start loading, "Conservative" increases the chance that only URLs the user actually navigates to are loaded.', 'performance-lab' ),
+				. '<br>' . __( '"Eager" will have the minimum delay to start speculative loads, "Conservative" increases the chance that only URLs the user actually navigates to are loaded.', 'performance-lab' ),
 		),
 	);
 	foreach ( $fields as $slug => $args ) {
--- a/modules/js-and-css/speculation-rules/settings.php
+++ b/modules/js-and-css/speculation-rules/settings.php
@@ -115,7 +115,7 @@ function plsr_register_setting() {
 							'enum'        => array_keys( plsr_get_mode_labels() ),
 						),
 						'eagerness' => array(
-							'description' => __( 'Whether to trigger on click (conservative), on hover (moderate), or on even the slight suggestion (eager) that the user may navigate to the URL.', 'performance-lab' ),
+							'description' => __( 'The eagerness setting defines the heuristics based on which the loading is triggered. "Eager" will have the minimum delay to start speculative loads, "Conservative" increases the chance that only URLs the user actually navigates to are loaded.', 'performance-lab' ),
 							'type'        => 'string',
 							'enum'        => array_keys( plsr_get_eagerness_labels() ),
 						),
--- a/modules/js-and-css/speculation-rules/settings.php
+++ b/modules/js-and-css/speculation-rules/settings.php
@@ -1,0 +1,218 @@
+<?php
+/**
+ * Settings functions used for Speculation Rules.
+ *
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+/**
+ * Returns the available options for the Speculation Rules mode and their labels.
+ *
+ * @since n.e.x.t
+ *
+ * @return array Associative array of `$mode => $label` pairs.
+ */
+function plsr_get_mode_labels() {
+	return array(
+		'prefetch'  => _x( 'Prefetch', 'setting label', 'performance-lab' ),
+		'prerender' => _x( 'Prerender', 'setting label', 'performance-lab' ),
+	);
+}
+
+/**
+ * Returns the available options for the Speculation Rules eagerness and their labels.
+ *
+ * @since n.e.x.t
+ *
+ * @return array Associative array of `$eagerness => $label` pairs.
+ */
+function plsr_get_eagerness_labels() {
+	return array(
+		'conservative' => _x( 'Conservative (on click)', 'setting label', 'performance-lab' ),
+		'moderate'     => _x( 'Moderate (on hover)', 'setting label', 'performance-lab' ),
+		'eager'        => _x( 'Eager (on slightest suggestion)', 'setting label', 'performance-lab' ),
+	);
+}
+
+/**
+ * Returns the default setting value for Speculation Rules configuration.
+ *
+ * @since n.e.x.t
+ *
+ * @return array Default value, an associative array with 'mode' and 'eagerness' keys.
+ */
+function plsr_get_setting_default() {
+	return array(
+		'mode'      => 'prerender',
+		'eagerness' => 'moderate',
+	);
+}
+
+/**
+ * Sanitizes the setting for Speculation Rules configuration.
+ *
+ * @since n.e.x.t
+ *
+ * @param mixed $input Setting to sanitize.
+ * @return array Sanitized setting, an associative array with 'mode' and 'eagerness' keys.
+ */
+function plsr_sanitize_setting( $input ) {
+	$default_value = plsr_get_setting_default();
+
+	if ( ! is_array( $input ) ) {
+		return $default_value;
+	}
+
+	$mode_labels      = plsr_get_mode_labels();
+	$eagerness_labels = plsr_get_eagerness_labels();
+
+	// Ensure only valid keys are present.
+	$value = array_intersect_key( $input, $default_value );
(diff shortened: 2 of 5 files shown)


modules/js-and-css/speculation-rules/settings.php
```diff
@@ -0,0 +1,218 @@
...
+		'mode'      => array(
+			'title'       => __( 'Mode', 'performance-lab' ),
+			'description' => __( 'Prerendering will lead to faster load times than prefetching, however in case of interactive content prefetching may be a safer choice.', 'performance-lab' ),
+		),
+		'eagerness' => array(
+			'title'       => __( 'Eagerness', 'performance-lab' ),
+			'description' => __( 'The eagerness setting defines the heuristics based on which the loading is triggered.', 'performance-lab' )
+				. '<br>' . __( '"Eager" is best from a performance perspective, "Conservative" increases the chance that only URLs the user actually navigates to are loaded.', 'performance-lab' ),
+		),
+	);
```
felixarntz: @domenic @jeremyroman Would be great to get a quick check from you on the explanations in the UI (mostly here, also in a few other places in this file).

modules/js-and-css/speculation-rules/settings.php
```diff
@@ -0,0 +1,218 @@
...
+ * Returns the available options for the Speculation Rules eagerness and their labels.
+ *
+ * @since n.e.x.t
+ *
+ * @return array Associative array of `$eagerness => $label` pairs.
+ */
+function plsr_get_eagerness_labels() {
+	return array(
+		'conservative' => _x( 'Conservative (on click)', 'setting label', 'performance-lab' ),
+		'moderate'     => _x( 'Moderate (on hover)', 'setting label', 'performance-lab' ),
```
jeremyroman: These are the heuristics that are landed in Chrome today. We might tweak/improve these to not be exactly this in the future (for example, "hover" isn't really a thing on mobile but we're looking at other heuristics there). And other browsers might use other heuristics especially on platforms with other kinds of things that indicate an upcoming click (e.g., if the user generally selects links with the keyboard rather than a pointing device).

It might still be reasonable to describe in this way to get an intution, but that's the full context here.

modules/js-and-css/speculation-rules/settings.php
```diff
@@ -0,0 +1,218 @@
...
+			'show_in_rest'      => array(
+				'schema' => array(
+					'properties' => array(
+						'mode'      => array(
+							'description' => __( 'Whether to prefetch or prerender URLs.', 'performance-lab' ),
+							'type'        => 'string',
+							'enum'        => array_keys( plsr_get_mode_labels() ),
+						),
+						'eagerness' => array(
+							'description' => __( 'Whether to trigger on click (conservative), on hover (moderate), or on even the slight suggestion (eager) that the user may navigate to the URL.', 'performance-lab' ),
```
jeremyroman: if any change is made above it should be reflected here

modules/js-and-css/speculation-rules/settings.php
```diff
@@ -0,0 +1,218 @@
...
+			</p>
+			<?php
+		},
+		'reading'
+	);
+
+	$fields = array(
+		'mode'      => array(
+			'title'       => __( 'Mode', 'performance-lab' ),
+			'description' => __( 'Prerendering will lead to faster load times than prefetching, however in case of interactive content prefetching may be a safer choice.', 'performance-lab' ),
```
jeremyroman: grammar nit

```suggestion
			'description' => __( 'Prerendering will lead to faster load times than prefetching. However, in case of interactive content, prefetching may be a safer choice.', 'performance-lab' ),
```

There are other aspects to the tradeoff (resource utilization, so you can prefetch more links than you can prerender, etc). But we can only fit so much nuance here, so this might suffice.

modules/js-and-css/speculation-rules/settings.php
```diff
@@ -0,0 +1,218 @@
...
+ *
+ * @since n.e.x.t
+ *
+ * @return array Associative array of `$eagerness => $label` pairs.
+ */
+function plsr_get_eagerness_labels() {
+	return array(
+		'conservative' => _x( 'Conservative (on click)', 'setting label', 'performance-lab' ),
+		'moderate'     => _x( 'Moderate (on hover)', 'setting label', 'performance-lab' ),
+		'eager'        => _x( 'Eager (on slightest suggestion)', 'setting label', 'performance-lab' ),
```
jeremyroman, line 34: BTW, at the last moment we split this into `"eager"` and `"immediate"`. Currently they're synonyms in Chrome, but the expectation is that `"immediate"` will continue to mean "right away, like on page load" and `"eager"` will be like this describes, "on the slightest suggestion".

This is probably still fine ("all links immediately" is probably not a very useful setting for a WP site), just mentioning it for context.

modules/js-and-css/speculation-rules/settings.php
felixarntz: Thanks! Good point, hover doesn't make sense for mobile. Still, I feel this may be the shortest way to describe it to give people a better idea.

Should I maybe add a "typically", e.g. "typically on hover"? That would better suggest that the behavior is not set in stone and may differ between browsers.

modules/js-and-css/speculation-rules/settings.php
felixarntz: Is there a good way for me to stay in the loop on this? If not, it would be great if you could ping me should there be related discussions.

modules/js-and-css/speculation-rules/settings.php
felixarntz: I saw the `immediate` in the spec, but when reading came to the same conclusion, that it would seem too excessive to be realistically of value in the WordPress context - a bit too "aggressive" :)

modules/js-and-css/speculation-rules/settings.php
tunetheweb: "best" seems a loaded term. Yes it gives most chance of the prerendering being completed before use, but also risks causing other performance problems (e.g. wasted resources/bandwidth) that may have other impacts.

WDYT about something like this:

> "Eager" will have the minimum delay to start prerendering. "Conservative"...

felixarntz: Thanks, that sounds good. Except I don't like to use "prerendering", since it's actually "prefetching or prerendering". I'm broadly using the term "loading" to cover both, since repeating these two terms everywhere is very verbose and long. I realize "loading" may not be the most adequate term, please let me know if there's a better single word to describe these two alternatives.

felixarntz: Updated in https://github.com/WordPress/performance/pull/939/commits/b1399613364dcaa8df75f12582b37d840ac3da3b

modules/js-and-css/speculation-rules/settings.php
felixarntz: Updated in https://github.com/WordPress/performance/pull/939/commits/9a6c3a99cdeeebc0277769cdf3ba9e607ac2b39d

modules/js-and-css/speculation-rules/settings.php
tunetheweb: "Speculative loads" or "Speculations" is what DevTools uses (it used to use "Preloading" in the past but [we got complaints](https://x.com/tunetheweb/status/1718899373168341106?s=20)).

felixarntz: Updated in https://github.com/WordPress/performance/pull/939/commits/24c3bb09c40e75bf8aa8927c6f9cd98dcda13a05.

I renamed "Mode" to "Speculation Mode" in the UI to clarify that relationship better.

modules/js-and-css/speculation-rules/helper.php
```diff
@@ -60,7 +60,15 @@ static function ( $exclude_path ) use ( $prefixer ) {
 		)
 	);
 
-	$prerender_rules = array(
+	$option = get_option( 'plsr_speculation_rules' );
+	if ( ! $option ) {
+		$option = plsr_get_setting_default();
+	}
```
westonruter: Alternative which would should the same result:

```suggestion
	$option = get_option( 'plsr_speculation_rules', plsr_get_setting_default() );
```

modules/js-and-css/speculation-rules/settings.php
```diff
@@ -0,0 +1,218 @@
...
+							'type'        => 'string',
+							'enum'        => array_keys( plsr_get_eagerness_labels() ),
+						),
+					),
+				),
+			),
+		)
+	);
+}
+add_action( 'init', 'plsr_register_setting' );
```
westonruter, line 128: Shoudn't this this rather be at `admin_init` and `rest_api_init`?

```suggestion
add_action( 'admin_init', 'plsr_register_setting' );
add_action( 'rest_api_init', 'plsr_register_setting' );
```

modules/js-and-css/speculation-rules/helper.php
westonruter: Also, what about declaring the option with `add_option()`? The default value and the desired autoload behavior can then be specified explicitly.

modules/js-and-css/speculation-rules/settings.php
```diff
@@ -0,0 +1,218 @@
...
+
+	$fields = array(
+		'mode'      => array(
+			'title'       => __( 'Speculation Mode', 'performance-lab' ),
+			'description' => __( 'Prerendering will lead to faster load times than prefetching. However, in case of interactive content, prefetching may be a safer choice.', 'performance-lab' ),
+		),
+		'eagerness' => array(
+			'title'       => __( 'Eagerness', 'performance-lab' ),
+			'description' => __( 'The eagerness setting defines the heuristics based on which the loading is triggered.', 'performance-lab' )
+				. '<br>' . __( '"Eager" will have the minimum delay to start speculative loads, "Conservative" increases the chance that only URLs the user actually navigates to are loaded.', 'performance-lab' ),
```
westonruter: Is the `<br>` necessary? It seems preferable to me to set a `max-width` on the container if the line is getting too long. This will prevent strange line wrapping on all viewports, including mobile.

modules/js-and-css/speculation-rules/settings.php
```diff
@@ -0,0 +1,218 @@
...
+					<?php echo esc_html( $label ); ?>
+				</label>
+			</p>
+			<?php
+		}
+
+		if ( ! empty( $args['description'] ) ) {
+			?>
+			<p class="description">
+				<?php echo wp_kses( $args['description'], array( 'br' => array() ) ); ?>
```
westonruter: If the `<br>` is removed per above, this can then be:

```suggestion
				<?php echo esc_html( $args['description'] ); ?>
```

modules/js-and-css/speculation-rules/settings.php
```diff
@@ -0,0 +1,218 @@
...
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+/**
+ * Returns the available options for the Speculation Rules mode and their labels.
+ *
+ * @since n.e.x.t
+ *
+ * @return array Associative array of `$mode => $label` pairs.
```
westonruter:
```suggestion
 * @return array<string, string> Associative array of `$mode => $label` pairs.
```

modules/js-and-css/speculation-rules/settings.php
```diff
@@ -0,0 +1,218 @@
...
+		'prerender' => _x( 'Prerender', 'setting label', 'performance-lab' ),
+	);
+}
+
+/**
+ * Returns the available options for the Speculation Rules eagerness and their labels.
+ *
+ * @since n.e.x.t
+ *
+ * @return array Associative array of `$eagerness => $label` pairs.
```
westonruter:
```suggestion
 * @return array<string, string> Associative array of `$eagerness => $label` pairs.
```

modules/js-and-css/speculation-rules/settings.php
```diff
@@ -0,0 +1,218 @@
...
+		'eager'        => _x( 'Eager (on slightest suggestion)', 'setting label', 'performance-lab' ),
+	);
+}
+
+/**
+ * Returns the default setting value for Speculation Rules configuration.
+ *
+ * @since n.e.x.t
+ *
+ * @return array Default value, an associative array with 'mode' and 'eagerness' keys.
```
westonruter:
```suggestion
 * @return array<string, string> Default value, an associative array with 'mode' and 'eagerness' keys.
```

modules/js-and-css/speculation-rules/settings.php
```diff
@@ -0,0 +1,218 @@
...
+	);
+}
+
+/**
+ * Sanitizes the setting for Speculation Rules configuration.
+ *
+ * @since n.e.x.t
+ *
+ * @param mixed $input Setting to sanitize.
+ * @return array Sanitized setting, an associative array with 'mode' and 'eagerness' keys.
```
westonruter: Could be more specific, if not with an array shape, then:

```suggestion
 * @return array<string, string> {
 *     Sanitized setting.
 *
 *     @type string $mode      Mode.
 *     @type string $eagerness Eagerness.
 * }
```

westonruter (changes requested):

modules/js-and-css/speculation-rules/helper.php
felixarntz: @westonruter Normally core will handle the default itself since this is a registered setting, i.e. passing the default isn't necessary. The fallback is only here for weird edge-cases (as far as I can tell only relevant for tests) where for some reason the option isn't registered. In other words, it's purely "very defensive coding".

> Also, what about declaring the option with `add_option()`? The default value and the desired autoload behavior can then be specified explicitly.

IMO not necessary. If a site uses the default, we don't have to store it in the DB. More importantly though, I prefer to avoid this as it would require some sort of "on activation" routine, which is generally something unreliable in WordPress plugins that doesn't scale well e.g. when using multisite. So IMO relying on a default is easier, also since we would want this option to be autoloaded anyway as it is small and used on pretty much every frontend page load.

westonruter: ~Is an activation routine needed? Isn't `add_option()` just run at `init`?~

Apparently it _is_ run at the activation hook.

modules/js-and-css/speculation-rules/settings.php
felixarntz: Options in core are registered on `init` I believe, which I think is right because options can in principle be set in any use-case, not just limited to WP Admin and REST API.

modules/js-and-css/speculation-rules/settings.php
felixarntz: I don't feel strongly about this. Breaking up sentences may be slightly more readable, but a max width would work too. I feel both are a bit cumbersome to do, I wish core's UI would account for these things by default.

modules/js-and-css/speculation-rules/settings.php
westonruter: But this is registering the setting, not the option, right? AFAIK, the setting only needs to be registered in a context where it can be modified. I'm only seeing `register_setting()` being [added](https://github.com/WordPress/wordpress-develop/blob/35add7378e093c8808ef9c08ee4b07f71b14963d/src/wp-includes/blocks/site-logo.php#L61-L95) at `rest_api_init` and Twenty Fourteen [registers](https://github.com/WordPress/wordpress-develop/blob/35add7378e093c8808ef9c08ee4b07f71b14963d/src/wp-content/themes/twentyfourteen/inc/featured-content.php#L82) a setting at `admin_init`.

modules/js-and-css/speculation-rules/settings.php
jeremyroman: There's an internal mailing list where Chrome stuff spanning this area is discussed, and sometimes that includes status updates (though it also has other discussion that may be noise to you). I'm in the process of refreshing an evergreen status doc, too. I'll try to remember to send it to you when done.

modules/js-and-css/speculation-rules/settings.php
felixarntz: > But this is registering the setting, not the option, right?

Setting and option are pretty much the same thing. There will never be a `register_option()` as `register_setting()` covers everything that it could possibly do. There are historic reasons for the weird naming. And FWIW some core contributors have previously said in discussions that they wish `register_setting()` was called `register_option()`.

> AFAIK, the setting only needs to be registered in a context where it can be modified. I'm only seeing `register_setting()` being [added](https://github.com/WordPress/wordpress-develop/blob/35add7378e093c8808ef9c08ee4b07f71b14963d/src/wp-includes/blocks/site-logo.php#L61-L95) at `rest_api_init` and Twenty Fourteen [registers](https://github.com/WordPress/wordpress-develop/blob/35add7378e093c8808ef9c08ee4b07f71b14963d/src/wp-content/themes/twentyfourteen/inc/featured-content.php#L82) a setting at `admin_init`.

Settings can be modified anywhere. Where you register a setting can depend on where you want to have its registered behavior supported, but it doesn't change where a setting can be modified. For example, the `default` parameter that can be registered is relevant literally anywhere that `get_option()` is called, so this is a good example of why _general_ registration (on `init`) makes most sense. A lot of the confusion is most likely for historic reasons:
* Many years ago, before the REST API, the only purpose of registering settings was to use the Settings API, which is the specific API to render UI for options in WP Admin (for whatever reason the "admin options API" was called the "Settings API"). That's why it only was relevant on `admin_init`.
* When the REST API came, registering settings was expanded to do a lot more than just registering for that admin Settings API. The decision was made to build on top of the existing function rather than introducing a more generic `register_option()` function. Myself and others don't necessarily think that was the right call, but we have to live with it now. Initially, some code then started to also add setting registration to `rest_api_init` when it was relevant there.
* However, both of those are just two specific use-cases for settings. There is nothing in WordPress that ties setting options to those contexts though. For the most reliable behavior, settings need to be registered for use anywhere, i.e. on `init` (unless it's legacy usage that only registers specifically for the old school Settings API, i.e. without all the new stuff like `type`, `description` etc.). Otherwise you run into weird situations where `get_option()` sometimes returns a useful default by default and other times not, which I'm not sure is useful in any case.

modules/js-and-css/speculation-rules/helper.php
felixarntz: Yeah, since `add_option()` writes to the database, it shouldn't be called in any regularly run action.

modules/js-and-css/speculation-rules/settings.php
felixarntz: This is one of my (least?) favorite annoyances and weirdnesses in WordPress, so I just thought I'd start a poll: https://twitter.com/felixarntz/status/1747349683947880931

modules/js-and-css/speculation-rules/settings.php
domenic: I think the use of "typically" here is good. The below sentence,

> Whether to trigger on click (conservative), on hover (moderate), or on even the slight suggestion (eager) that the user may navigate to the URL.

seems more problematic, but maybe adding "typically" could help there too?

Or you could try replacing that sentence with a more vague one, e.g. below you have

> The eagerness setting defines the heuristics based on which the loading is triggered.

and

> "Eager" will have the minimum delay to start speculative loads, "Conservative" increases the chance that only URLs the user actually navigates to are loaded.

which are both quite good IMO.

modules/js-and-css/speculation-rules/settings.php
swissspidy: Init is also important for things like WP-CLI, FWIW

modules/js-and-css/speculation-rules/helper.php
mukeshpanchal27: When reviewing this PR, I had similar thoughts to what Weston proposed. Thanks for clarifying, Felix.

modules/js-and-css/speculation-rules/settings.php
felixarntz: Thanks, updated in https://github.com/WordPress/performance/pull/939/commits/da0b883ba57a08456019aa21289e453cbb6173c7

modules/js-and-css/speculation-rules/settings.php
felixarntz: Updated in https://github.com/WordPress/performance/pull/939/commits/e24b01014919208f7dae98c33d26b50a75f53de0

modules/js-and-css/speculation-rules/settings.php
felixarntz: I used the more detailed syntax including `@type` args here and in one more place, see https://github.com/WordPress/performance/pull/939/commits/3c6aec22f29bf71845139d82c39c50a9660efeba.

modules/js-and-css/speculation-rules/helper.php
```diff
@@ -50,7 +50,15 @@ function plsr_get_speculation_rules() {
...
+
+	$mode      = $option['mode'];
+	$eagerness = $option['eagerness'];
```
westonruter: While unlikely, couldn't it be that perhaps `$option['mode']` isn't set here? If someone put an array into the `plsr_speculation_rules` without setting that key, then either `$option['mode']` or `$option['eagerness']` may not be set. So for extra defensive coding, I think this would be even more robust:

```suggestion
	$option = get_option( 'plsr_speculation_rules' );
	if ( ! is_array( $option ) ) {
		$option = array();
	}
	$option = array_merge( plsr_get_setting_default(), $option );

	$mode      = $option['mode'];
	$eagerness = $option['eagerness'];
```

modules/js-and-css/speculation-rules/settings.php
```diff
@@ -0,0 +1,233 @@
...
+		),
+		'eagerness' => array(
+			'title'       => __( 'Eagerness', 'performance-lab' ),
+			'description' => __( 'The eagerness setting defines the heuristics based on which the loading is triggered. "Eager" will have the minimum delay to start speculative loads, "Conservative" increases the chance that only URLs the user actually navigates to are loaded.', 'performance-lab' ),
+		),
+	);
+	foreach ( $fields as $slug => $args ) {
+		add_settings_field(
+			"plsr_speculation_rules_{$slug}",
+			$args['title'],
```
westonruter, line 163: Not sure if my suggestion here is warranted as I don't recall how this argument is even used versus what is passed to `plsr_render_settings_field()`

```suggestion
			esc_html( $args['title'] ),
```

swissspidy (approved):

westonruter (changes requested):
Almost there!

modules/js-and-css/speculation-rules/helper.php
felixarntz: This is extremely unlikely as in the scenario you're describing the value would still always be sanitized at the lowest level (within `update_option()`) to include both values, per the setting registration.

Still, fair enough to do this from a defensive coding perspective. Updated in https://github.com/WordPress/performance/pull/939/commits/d490f9318807937723f92f8e0e005df8266f46f5

adamsilverstein:
I tested this in Chrome Canary and verified the speculation rules are working. 

Checking the page source, I see the json object:
image

Also, I noticed the plugin generator tag includes the `js-and-css/speculation-rules` string.

Checking DevTools -> Application->Speculative Loads->Speculations I can see that all of the links on my (local test) homepage are included. When I hover over a link, I see that the prerender is triggered and the "Ready":
image

Finally, even on my local machine clicking these prerendered links is notably faster!

modules/js-and-css/speculation-rules/settings.php
felixarntz: I don't think that's needed, as the output of this is handled by core and thus responsibility of core. This isn't something WPCS complains about, so I don't think it's needed here.

adamsilverstein (approved):
Fantastic!

adamsilverstein:
Not sure what this message is about, I only added my review and I still see them as reviewers so 🤷🏼 

image

westonruter (approved):

PR #944, opened 2024-01-17
Fix weird user experience with module to plugin migration where filesystem changes are disabled

### LINKED ISSUE #943: Weird user experience with module to plugin migration where filesystem changes are disabled
When testing the new module to standalone plugin migration in the 2.8.0 release on a site where any filesystem changes (most importantly for this, installing plugins) are disabled, a weird user experience can be observed.

To test, use a site where you set the `DISALLOW_FILE_MODS` constant to `true`.

First of all, the good thing is all permissions are respected, which is the most critical bit. There aren't any security flaws in the implementation as far as I can tell. 👍 

However, here's what happens on such a site:
* The admin pointer isn't shown. (This may be expected because installing plugins isn't possible anyway.)
* The admin notice however is shown. But when clicking the CTA button to migrate, it (rightfully) fails with an error regarding the missing capabilities.

I would suggest the following to improve this strange experience:
* In case the user is missing one of the `install_plugins` or `activate_plugins` capabilities, don't show the CTA button.
* Instead, adjust the text slightly to keep it purely informative. For example, the last line could be simply changed from:
> Please click the following button to install and activate the relevant plugin/s in favor of the modules. This will not impact any of the underlying functionality.

to

> Please install and activate the relevant plugin/s in favor of the modules. This will not impact any of the underlying functionality.

I think that would be good enough for a solution.

We could be more granular and still allow migration if all plugins are already installed, but I think that would be too much effort for a problem that only affects a very small niche of sites (if the plugins are already installed, the user is probably already aware and will manually migrate anyway). So I don't think we should do that.

mukeshpanchal27:
## Summary

Fixes #943 

## Checklist

### DIFF
--- a/admin/load.php
+++ b/admin/load.php
@@ -815,14 +815,14 @@ function perflab_plugin_admin_notices() {
 	if ( current_user_can( 'install_plugins' ) && current_user_can( 'activate_plugins' ) ) {
 		$user_have_capability = true;
 		if ( 1 === $modules_count ) {
-			$additional_message = esc_html__( 'Please click the following button to install and activate the relevant plugin in favor of the module.', 'performance-lab' );
+			$additional_message = __( 'Please click the following button to install and activate the relevant plugin in favor of the module.', 'performance-lab' );
 		} else {
-			$additional_message = esc_html__( 'Please click the following button to install and activate the relevant plugins in favor of the modules.', 'performance-lab' );
+			$additional_message = __( 'Please click the following button to install and activate the relevant plugins in favor of the modules.', 'performance-lab' );
 		}
 	} elseif ( 1 === $modules_count ) {
-		$additional_message = esc_html__( 'Please install and activate the relevant plugin in favor of the module.', 'performance-lab' );
+		$additional_message = __( 'Please install and activate the relevant plugin in favor of the module.', 'performance-lab' );
 	} else {
-		$additional_message = esc_html__( 'Please install and activate the relevant plugins in favor of the modules.', 'performance-lab' );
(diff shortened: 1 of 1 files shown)


admin/load.php
```diff
@@ -820,13 +833,17 @@ function perflab_plugin_admin_notices() {
 			esc_attr( $available_module_names[0] )
 		);
 		$message .= ' ';
-		$message .= esc_html__( 'Please click the following button to install and activate the relevant plugin in favor of the module. This will not impact any of the underlying functionality.', 'performance-lab' );
+		$message .= $additional_message;
```
westonruter: Better to do the escaping here, as late as possible:

```suggestion
		$message .= esc_html( $additional_message );
```

The `esc_html__()` instances can be replaced with `__()`.

mukeshpanchal27: Thanks for quick feedback. Addressed in https://github.com/WordPress/performance/pull/944/commits/e694fc091a243f1235af257700a302a45be2b42c

admin/load.php
```diff
@@ -812,6 +811,20 @@ function perflab_plugin_admin_notices() {
 	$available_module_names = wp_list_pluck( $active_modules_with_inactive_plugins, 'name' );
 	$modules_count          = count( $available_module_names );
 
+	$user_have_capability = false;
```
swissspidy: Let's rename this to `$user_has_capability` or maybe simply `$has_cap`

Also, we can simplify this to:

```php
$has_cap = current_user_can( 'install_plugins' ) && current_user_can( 'activate_plugins' );
```

swissspidy (approved):

felixarntz (approved):

Issue #934, opened 2024-01-12
Introduce `plugins` folder for standalone plugins


felixarntz:
This issue is part 1/2 for setting up the new infrastructure for developing standalone plugins.

## Requirements

**This work should be implemented against a new `feature/modules-to-plugins` feature branch.**

* Create new `plugins` folder in the repository root.
* Add a `.gitkeep` to it.
* Add the `plugins` folder to be ignored in `.gitattributes` so that it won't be part of the Performance Lab plugin build.
* Restructure the `plugins.json` file:
    * Going forward, there should be always two top level properties: `modules` and `plugins`.
    * `modules` should have the map of "module => plugin data (slug and version)", i.e. the same thing that the file contains today.
    * `plugins` should have an array of plugin slugs.
* Update the existing workflows that use the `plugins.json` file to look for modules within the new `modules` property in the file.
* Add references to the new `plugins` folder in any dev tooling that currently lints/covers the `modules` folder (e.g. in `phpcs.xml.dist` etc.).

Note: Additional changes to build/publish the new plugins should be implemented in a separate follow-up issue.

felixarntz:
@mukeshpanchal27 @joemcgill @swissspidy See https://github.com/WordPress/performance/pull/946#issuecomment-1897105844: This brought up another thing that needs to be addressed which is the actual test coverage.

I wonder how we can address this in the most efficient way. `phpunit.xml.dist` seems to support multiple `testsuites`, so maybe we can somehow have individual test suites per plugin in there? One for Performance Lab, and one for each standalone plugin in the `plugins` folder? The tests for any plugins in `plugins` folder could live in a separate folder then too, e.g. `tests/plugins`, so then the PL test suite would ignore that folder, while the individual plugins' test suites would only consider tests from their relevant directory.

Maybe the test suites could then also set a constant or environment variable which the `bootstrap.php` file can use to determine which plugin to force-activate.

Not sure whether this makes sense, I'm curious on your feedback or any alternative ideas you may have to make this possible but also keep it as simple as possible.

swissspidy:
Right, with this becoming more like a monorepo of individual plugins we need to run tests and static analysis separately for each plugin.

Right now the PHPUnit `bootstrap.php` just loads _everything_, which is certainly not ideal. Separate test suites and then some way to only load what's needed sounds good. I'd say let's just give it a try in a PR and see what works and what doesn't. These are just tests, we can quickly iterate and tweak the setup as we see fit.

[...]

PR #946, opened 2024-01-17
Move Speculation Rules to new location as purely a standalone plugin

felixarntz:
## Summary

See https://github.com/WordPress/performance/pull/904#issuecomment-1896444353: Since the Speculation Rules feature will not be published as a module (given the unbundling of Performance Lab) and thus only as a plugin, it should be moved to the new location for standalone plugins (see #934).

This PR is still against the feature branch and prepares that feature branch for a clean merge after #934 has been implemented. There's no blocker for merging this into the feature branch though.

**No functional changes are included as part of this PR.**

## Checklist


westonruter (approved):

swissspidy (approved):
I take it the lint errors are known?

felixarntz:
I realize there's a PHPStan error here, which may be addressed by including the `plugins/` directory in `phpstan.neon.dist`, however it pointed me to another problem: We should probably move the test coverage to another place as well.

This also relates to unit testing. We'll have to make a few changes to the infrastructure, either change `bootstrap.php` somehow or use different PHPUnit config files based on which plugin is tested. There's probably some way to do this with the least code duplication possible, maybe using environment variables or something. I think it makes sense to discuss that further as part of #934.

mukeshpanchal27 (approved):

plugins/speculation-rules/load.php
```diff
@@ -0,0 +1,28 @@
+<?php
+/**
+ * Plugin Name: Speculation Rules
+ * Plugin URI: https://github.com/WordPress/performance/tree/trunk/modules/js-and-css/speculation-rules
+ * Description: Uses the Speculation Rules API to prerender linked URLs upon hover by default.
+ * Requires at least: 6.3
+ * Requires PHP: 7.0
+ * Version: 1.0.0
```
westonruter: After https://github.com/WordPress/performance/pull/993 this will need to be:

```suggestion
 * Version: 1.0.1
```

plugins/speculation-rules/load.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @package speculation-rules
+ */
+
+// Define the constant.
+if ( defined( 'SPECULATION_RULES_VERSION' ) ) {
+	return;
+}
+
+define( 'SPECULATION_RULES_VERSION', '1.0.0' );
```
westonruter: Ditto:

```suggestion
define( 'SPECULATION_RULES_VERSION', '1.0.1' );
```

felixarntz:
@westonruter @swissspidy @mukeshpanchal27 Since this is the only remaining PR at the moment for Speculation Rules and we're working on the standalone plugins infrastructure in https://github.com/WordPress/performance/tree/feature/modules-to-plugins (which already includes the `auto-sizes` plugin), I will change the base branch of this PR to be that one.

The `feature/speculationrules` branch will then be obsolete and can be deleted, all further development of the Speculation Rules plugin can happen against `feature/modules-to-plugins` and, once that is merged, against `trunk`.

.github/workflows/php-test-plugins.yml
```diff
@@ -76,3 +76,7 @@ jobs:
         run: npm run test-php -- -- -- --testsuite auto-sizes
       - name: Running multisite unit tests for Auto-sizes for Lazy-Loaded Images Plugin
         run: npm run test-php-multisite -- -- -- --testsuite auto-sizes
+      - name: Running single site unit tests for Speculation Rules Plugin
+        run: npm run test-php -- -- -- --testsuite speculation-rules
+      - name: Running multisite unit tests for Speculation Rules Plugin
+        run: npm run test-php-multisite -- -- -- --testsuite speculation-rules
```
felixarntz: This is cumbersome so we should probably add shortcuts via scripts in `composer.json` and/or `package.json`.

There should also be a shortcut that runs the single site and multisite tests respectively for _all_ of these plugins at once, so that we don't have to add every single plugin to this GitHub workflow file.

Let's handle that in a separate PR though, not directly related to the Speculation Rules work.

phpcs.xml.dist
```diff
@@ -8,7 +8,7 @@
 	<rule ref="WordPress-Docs"/>
 	<rule ref="WordPress-Extra" />
 	<rule ref="WordPress.WP.I18n"/>
-	<config name="text_domain" value="performance-lab,default,auto-sizes"/>
+	<config name="text_domain" value="performance-lab,default,auto-sizes,speculation-rules"/>
```
felixarntz: This will need to be adjusted once #1002 is merged, instead a `plugins/speculation-rules/phpcs.xml.dist` file should be added then.

felixarntz (commented):
@westonruter @swissspidy @mukeshpanchal27 This is pretty much ready now and refreshed for the way things work now in `feature/modules-to-plugins`.

We should probably merge #1002 first, then do another quick refresh of this one, then merge this one.

After that we'll have both `auto-sizes` and `speculation-rules` in the `feature/modules-to-plugins` branch and can continue development against that branch until it's merged into `trunk`.

westonruter:
> We should probably merge #1002 first, then do another quick refresh of this one, then merge this one.

Let's merge it then. 🚀

.github/workflows/php-test-plugins.yml
felixarntz: I opened #1017 for that.

westonruter:
> Let's merge it then. 🚀

Merged!

felixarntz:
@westonruter @swissspidy @mukeshpanchal27 Should be ready for merge now!

You had already approved prior, but since I now changed the base branch, it would be great if you could do another quick sanity check.

Note that all the actual PHP code for Speculation Rules is unchanged and was already approved in the `feature/speculationrules` branch, so you can ignore the PHP code for the review.

felixarntz:
@westonruter It looks like a test related to the CDATA fix you implemented fails, see https://github.com/WordPress/performance/actions/runs/8071362067/job/22050829990?pr=946. Maybe it is because the test code isn't limited to the specific WordPress versions?

Since the CDATA issue is only a problem on WordPress 6.4, we probably need to limit the test accordingly. Either we make those assertions conditional, or we mark the entire test to skip unless it's WordPress version 6.4.x. WDYT?

On another note: I think the code comment linking to the Trac ticket points to the wrong ticket? Shouldn't it be https://core.trac.wordpress.org/ticket/60320?

tests/plugins/speculation-rules/plsr-url-pattern-prefixer-test.php
```diff
@@ -0,0 +1,76 @@
+<?php
+/**
+ * Tests for PLSR_URL_Pattern_Prefixer class.
+ *
+ * @package performance-lab
+ * @group speculation-rules
```
westonruter:
```suggestion
 * @package speculation-rules
```

tests/plugins/speculation-rules/speculation-rules-helper-test.php
```diff
@@ -0,0 +1,93 @@
+<?php
+/**
+ * Tests for speculation-rules helper file.
+ *
+ * @package performance-lab
+ * @group speculation-rules
```
westonruter:
```suggestion
 * @package speculation-rules
```

tests/plugins/speculation-rules/speculation-rules-settings-test.php
```diff
@@ -0,0 +1,96 @@
+<?php
+/**
+ * Tests for speculation-rules settings file.
+ *
+ * @package performance-lab
+ * @group speculation-rules
```
westonruter:
```suggestion
 * @package speculation-rules
```

tests/plugins/speculation-rules/speculation-rules-test.php
```diff
@@ -0,0 +1,60 @@
+<?php
+/**
+ * Tests for speculation-rules module.
+ *
+ * @package performance-lab
+ * @group speculation-rules
```
westonruter:
```suggestion
 * @package speculation-rules
```

westonruter:
@felixarntz I opted to keep the test but add a condition for the WP version. Tests are passing now.

westonruter (approved):
Just some trivial phpdoc changes for the test files, but otherwise I'm pre-approving.

plugins/speculation-rules/class-plsr-url-pattern-prefixer.php
```diff
@@ -0,0 +1,122 @@
+<?php
+/**
+ * Class 'PLSR_URL_Pattern_Prefixer'.
+ *
+ * @package speculation-rules
+ * @since n.e.x.t
```
mukeshpanchal27: What about version? When we change it from `n.e.x.t` to actual version `1.0.0`

plugins/speculation-rules/load.php
```diff
@@ -0,0 +1,28 @@
+<?php
+/**
+ * Plugin Name: Speculation Rules
+ * Plugin URI: https://github.com/WordPress/performance/tree/trunk/modules/js-and-css/speculation-rules
```
mukeshpanchal27:
```suggestion
 * Plugin URI: https://github.com/WordPress/performance/tree/trunk/plugins/speculation-rules
```

plugins/speculation-rules/class-plsr-url-pattern-prefixer.php
swissspidy: Since the plugin is already published, this should be already 1.0.0, no?

plugins/speculation-rules/load.php
felixarntz: Good catch! I fixed this for `auto-sizes` as well.

felixarntz:
@swissspidy @mukeshpanchal27 I've addressed your feedback! This is probably good to merge now.

PR #951, opened 2024-01-20
Escape path prefix and restrict it to be a pathname in Speculation Rules

### LINKED ISSUE #950: Speculation rules: URL patterns need escaping to work correctly when special characters are present
## Bug Description

If WordPress is deployed with an unusual path prefix (notably WordPress Playground uses something like /scope:0.123456/), this isn't properly escaped and the resulting patterns fail to parse.

## Steps to reproduce

1. Go to https://playground.wordpress.net/?plugin=speculation-rules.
2. Open one of the pages at the top level.
3. Observe that there are errors in the developer console resulting pages.

## Screenshots

n/a

## Additional Context

 - PHP Version: 7.4
 - OS: Linux
 - Browser: Chrome (121 beta)
 - Plugin Version: trunk
 - Device: Linux desktop

jeremyroman:
On unusual deployment setups (but notably WordPress Playground), the path prefix might contain one of the URL pattern special characters, such as :.

Escape these characters in the path prefix, and since some of these also confuse the shorthand constructor string syntax, use the longhand syntax which specifies that the presence of a : here does not indicate a protocol/scheme.

## Summary

Fixes #950.

## Relevant technical choices

* Prefixes are escaped in `PLSR_URL_Pattern_Prefixer::get_default_contexts()`, so that pattern syntax works in the particular excluded patterns.
* All replacements are made, consistent with https://urlpattern.spec.whatwg.org/#escape-a-pattern-string
* If the prefix contains a character that causes a component change out of path (`:`, `?` or `#`, though the latter two should be percent-encoded anyway since they'd have a problem in plain URLs, too), it's enclosed in braces to ensure it's treated literally by the URL pattern parser.

## Checklist

### DIFF
--- a/modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
+++ b/modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
@@ -31,7 +31,12 @@ class PLSR_URL_Pattern_Prefixer {
 	 */
 	public function __construct( array $contexts = array() ) {
 		if ( $contexts ) {
-			$this->contexts = array_map( 'trailingslashit', $contexts );
+			$this->contexts = array_map(
+				static function ( string $str ): string {
+					return self::escape_pattern_string( trailingslashit( $str ) );
+				},
+				$contexts
+			);
 		} else {
 			$this->contexts = self::get_default_contexts();
 		}
@@ -69,12 +74,23 @@ class PLSR_URL_Pattern_Prefixer {
 			return $path_pattern;
 		}
 
-		// If the path already starts with the context path (including '/'), there is nothing to prefix.
-		if ( str_starts_with( $path_pattern, $this->contexts[ $context ] ) ) {
-			return $path_pattern;
+		// In the event that the context path contains a :, ? or # (which can cause the URL pattern parser to
+		// switch to another state, though only the latter two should be percent encoded anyway), we need to
+		// additionally enclose it in grouping braces. The final forward slash (trailingslashit ensures there is
+		// one) affects the meaning of the * wildcard, so is left outside the braces.
+		$context_path         = $this->contexts[ $context ];
+		$escaped_context_path = $context_path;
+		if ( strcspn( $context_path, ':?#' ) !== strlen( $context_path ) ) {
+			$escaped_context_path = '{' . substr( $context_path, 0, -1 ) . '}/';
 		}
 
-		return $this->contexts[ $context ] . ltrim( $path_pattern, '/' );
+		// If the path already starts with the context path (including '/'), remove it first
+		// since it is about to be added back.
+		if ( str_starts_with( $path_pattern, $context_path ) ) {
+			$path_pattern = substr( $path_pattern, strlen( $context_path ) );
+		}
+
+		return $escaped_context_path . ltrim( $path_pattern, '/' );
 	}
 
 	/**
@@ -100,7 +116,7 @@ class PLSR_URL_Pattern_Prefixer {
 	 * @param string $str String to be escaped.
 	 * @return string String with backslashes added where required.
 	 */
-	public static function escape_pattern_string( string $str ): string {
+	private static function escape_pattern_string( string $str ): string {
 		$replacements = array(
 			'+'  => '\\+',
 			'*'  => '\\*',
--- a/modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
+++ b/modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
@@ -109,8 +109,8 @@ class PLSR_URL_Pattern_Prefixer {
 
 	/**
 	 * Escapes a string for use in a URL pattern component.
-	 * https://urlpattern.spec.whatwg.org/#escape-a-pattern-string
 	 *
+	 * @link https://urlpattern.spec.whatwg.org/#escape-a-pattern-string
 	 * @since n.e.x.t
 	 *
 	 * @param string $str String to be escaped.
--- a/modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
+++ b/modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
@@ -117,17 +117,6 @@ class PLSR_URL_Pattern_Prefixer {
 	 * @return string String with backslashes added where required.
 	 */
 	private static function escape_pattern_string( string $str ): string {
-		$replacements = array(
-			'+'  => '\\+',
-			'*'  => '\\*',
-			'?'  => '\\?',
-			':'  => '\\:',
-			'{'  => '\\{',
-			'}'  => '\\}',
-			'('  => '\\(',
-			')'  => '\\)',
-			'\\' => '\\\\',
-		);
-		return strtr( $str, $replacements );
+		return addcslashes( $str, '+*?:{}()\\' );
 	}
 }
--- a/modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
+++ b/modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
@@ -86,8 +86,32 @@ class PLSR_URL_Pattern_Prefixer {
 	 */
 	public static function get_default_contexts(): array {
 		return array(
-			'home' => trailingslashit( wp_parse_url( home_url( '/' ), PHP_URL_PATH ) ),
-			'site' => trailingslashit( wp_parse_url( site_url( '/' ), PHP_URL_PATH ) ),
+			'home' => self::escape_pattern_string( trailingslashit( wp_parse_url( home_url( '/' ), PHP_URL_PATH ) ) ),
+			'site' => self::escape_pattern_string( trailingslashit( wp_parse_url( site_url( '/' ), PHP_URL_PATH ) ) ),
 		);
+	}
+
+	/**
+	 * Escapes a string for use in a URL pattern component.
+	 * https://urlpattern.spec.whatwg.org/#escape-a-pattern-string
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $str String to be escaped.
+	 * @return string String with backslashes added where required.
+	 */
+	public static function escape_pattern_string( string $str ): string {
+		$replacements = array(
+			'+'  => '\\+',
+			'*'  => '\\*',
+			'?'  => '\\?',
+			':'  => '\\:',
+			'{'  => '\\{',
+			'}'  => '\\}',
+			'('  => '\\(',
+			')'  => '\\)',
+			'\\' => '\\\\',
+		);
+		return strtr( $str, $replacements );
 	}
 }
--- a/modules/js-and-css/speculation-rules/helper.php
+++ b/modules/js-and-css/speculation-rules/helper.php
@@ -41,10 +41,10 @@ function plsr_get_speculation_rules() {
 
 	// Ensure that there are no duplicates and that the base paths cannot be removed.
 	$href_exclude_paths = array_unique(
-		array_merge(
-			$base_href_exclude_paths,
-			array_map(
-				array( $prefixer, 'prefix_path_pattern' ),
+		array_map(
+			array( $prefixer, 'prefix_path_pattern' ),
+			array_merge(
+				$base_href_exclude_paths,
 				$href_exclude_paths
 			)
 		)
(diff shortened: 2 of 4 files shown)


jeremyroman:
@felixarntz Noticed this as part of the internal email thread talking about this plugin. It's been at least a decade since I wrote PHP, but this seems to work for me locally and I think I managed to match WordPress coding style. PTAL?

modules/js-and-css/speculation-rules/helper.php
```diff
@@ -41,10 +41,10 @@ function plsr_get_speculation_rules() {
...
-			array( $prefixer, 'prefix_path_pattern' ),
-			array_merge(
-				$base_href_exclude_paths,
+		array_merge(
+			$base_href_exclude_paths,
+			array_map(
+				array( $prefixer, 'prefix_path_pattern' ),
 				$href_exclude_paths
 			)
 		)
```
felixarntz: Can you clarify the rationale for this change? I'm not sure I understand.

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
```diff
@@ -86,8 +86,32 @@ public function prefix_path_pattern( string $path_pattern, string $context = 'ho
 	 */
 	public static function get_default_contexts(): array {
 		return array(
-			'home' => trailingslashit( wp_parse_url( home_url( '/' ), PHP_URL_PATH ) ),
-			'site' => trailingslashit( wp_parse_url( site_url( '/' ), PHP_URL_PATH ) ),
+			'home' => self::escape_pattern_string( trailingslashit( wp_parse_url( home_url( '/' ), PHP_URL_PATH ) ) ),
+			'site' => self::escape_pattern_string( trailingslashit( wp_parse_url( site_url( '/' ), PHP_URL_PATH ) ) ),
```
felixarntz, line 106: This escapes the "path prefixes", which makes sense. But wouldn't we also want to escape any other parts of the path string in the same way?

Any string that a 3P developer may provide via the `plsr_speculation_rules_href_exclude_paths` filter below could also contain one of those characters that would need to be escaped as far as I can tell.

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
```diff
@@ -86,8 +86,32 @@ public function prefix_path_pattern( string $path_pattern, string $context = 'ho
...
+			'*'  => '\\*',
+			'?'  => '\\?',
+			':'  => '\\:',
+			'{'  => '\\{',
+			'}'  => '\\}',
+			'('  => '\\(',
+			')'  => '\\)',
+			'\\' => '\\\\',
+		);
+		return strtr( $str, $replacements );
```
felixarntz: I wonder whether there's a native PHP function that does this 🤔 

But if not, I think we could simplify this to be a regular expression, since the replacement pattern is always the same (prepend with a single backslash, i.e. `\\`):
```suggestion
		// Escape any of these individual characters by prefixing with a single backslash.
		return preg_replace( '/[+*?:{}()\\\\]{1}/', '\\\\$0', $str );
```

modules/js-and-css/speculation-rules/helper.php
```diff
@@ -72,12 +72,17 @@ function plsr_get_speculation_rules() {
...
 					// Except for WP login and admin URLs.
 					array(
 						'not' => array(
-							'href_matches' => $href_exclude_paths,
+							'href_matches' => array_map(
+								static function ( string $path ): array {
+									return array( 'pathname' => $path );
+								},
+								$href_exclude_paths
+							),
```
felixarntz: As you're saying in the PR description, this is potentially controversial. I'd like to understand whether this is truly needed? Wouldn't the escaping alone solve the problem (potentially being expanded to the full path string, rather than just the "path prefix")?

Maybe limiting this to paths makes sense, but at the moment this would prevent limiting by query parameters for instance, which is relevant for some WordPress sites (most importantly for the small subset of sites that doesn't use "pretty permalinks").

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
```diff
@@ -86,8 +86,32 @@ public function prefix_path_pattern( string $path_pattern, string $context = 'ho
...
+	/**
+	 * Escapes a string for use in a URL pattern component.
+	 * https://urlpattern.spec.whatwg.org/#escape-a-pattern-string
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $str String to be escaped.
+	 * @return string String with backslashes added where required.
+	 */
+	public static function escape_pattern_string( string $str ): string {
```
felixarntz: Right now this method is only used within this class. If it remains like that, it should probably become private.

felixarntz (commented):
Thanks @jeremyroman. PHP code wise this looks good indeed, though I have a few minor suggestions and questions to better understand the changes.

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
jeremyroman: Well, such strings may in fact be patterns and not literal paths, so we probably do want the special characters to have their special meanings there.

For example, the `readme.txt` example is `'/cart/*'`, and in this case the developer probably did mean "all paths within `/cart/`", not "the literal path `/cart/*`".

However, if WordPress is serving from a directory named `/*` (however odd a choice that may be), then we need to escape the asterisk because it's _not_ intended as a wildcard.

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
jeremyroman: Done. Maybe it would be useful to some other plugin that did want to do this escaping too, but that's not immediately foreseen.

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
jeremyroman: Not that I could find. `addslashes` and `addcslashes` both do it for different sets of special characters, but `strtr` seemed to be the most common approach here.

I can use `preg_replace` (the `{1}` is implied I think but otherwise I'd come up with the same line of code) if you'd prefer.

I'd avoided it because the PHP manual and the results of my Stack Overflow searches both advised not using it where something weaker than regular expressions would suffice, for performance reasons (presumably compiling the pattern has a cost).

I think both are similarly readable but happy to defer to your expertise if `preg_replace` is preferred.

modules/js-and-css/speculation-rules/helper.php
jeremyroman: Ah, sorry, this is simply because the previous code applies `$prefixer->prefix_path_pattern` to the base exclusions twice -- once at lines 22 and 23, and again here. This kinda works when the `home_url` is a prefix of `site_url`, because of the `str_starts_with` check within, though even that would fall apart in some funny cases, like where the user installed WP in a directory itself named `/wp-admin/` or something like that.

However because these paths are _also_ sent by default to the `plsr_speculation_rules_href_exclude_paths` filter, this diff isn't really sufficient to resolve that issue, and regardless this is mostly orthogonal to the goal of this PR so I'll revert it and let that question be potentially resolved elsewhere in the future.

modules/js-and-css/speculation-rules/helper.php
jeremyroman: Yeah, that's a good point. Unfortunately there are two different kinds of escaping needed here (at least, unless we fix the URL pattern parsing logic to be a little smarter).

The backslash escaping gets us out of the wrong tokenizer behavior (treating `:` as the start of a [name token](https://urlpattern.spec.whatwg.org/#token-type-name)), but we still end up with the higher level parser seeing the : and thinking "hey! this must be an absolute URL and the stuff to the left of this is a protocol". So it still ends up thinking we have a scheme of "/scope" when we pass it "/scope:0.123/". The URL parser is...smarter about this.

I was trying to force it to be a pathname in this way, but upon a second look I can work around it in a different way, by using a group to force it to not consider this a component boundary.

jeremyroman: https://github.com/whatwg/urlpattern/issues/210 has some more context if you're curious, but even if we do make this syntax legal it won't be accepted by Chromium for at least two milestones, so still worth this workaround.

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
westonruter: The `addcslashes()` function allows you to specify the characters to escape, too. So I think this is better:

```php
return addcslashes( $str, '+*?:{}()\\' );
```

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
```diff
@@ -86,8 +102,32 @@ public function prefix_path_pattern( string $path_pattern, string $context = 'ho
...
+			'*'  => '\\*',
+			'?'  => '\\?',
+			':'  => '\\:',
+			'{'  => '\\{',
+			'}'  => '\\}',
+			'('  => '\\(',
+			')'  => '\\)',
+			'\\' => '\\\\',
 		);
+		return strtr( $str, $replacements );
```
westonruter:
```suggestion
		return addcslashes( $str, '+*?:{}()\\' );
```

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
jeremyroman: You're right! When I first read the manual page I misunderstood the section about how characters that have special C meaning, like \n and \r, and thought it _only_ applied to characters with meaningful C escape sequences. But that's not the case, so `addcslashes` is unambiguously the best solution.

jeremyroman:
Anything else I need to change for this to be mergeable?

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
```diff
@@ -69,12 +74,23 @@ public function prefix_path_pattern( string $path_pattern, string $context = 'ho
...
-		// If the path already starts with the context path (including '/'), there is nothing to prefix.
-		if ( str_starts_with( $path_pattern, $this->contexts[ $context ] ) ) {
-			return $path_pattern;
+		// In the event that the context path contains a :, ? or # (which can cause the URL pattern parser to
+		// switch to another state, though only the latter two should be percent encoded anyway), we need to
+		// additionally enclose it in grouping braces. The final forward slash (trailingslashit ensures there is
+		// one) affects the meaning of the * wildcard, so is left outside the braces.
+		$context_path         = $this->contexts[ $context ];
+		$escaped_context_path = $context_path;
+		if ( strcspn( $context_path, ':?#' ) !== strlen( $context_path ) ) {
```
westonruter, line 83: Clever. I would have done `str_contains( $context_path, ':' ) || str_contains( $context_path, '?' ) || str_contains( $context_path, '#' )` but clearly what you've done is faster.

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
```diff
@@ -86,8 +102,21 @@ public function prefix_path_pattern( string $path_pattern, string $context = 'ho
...
+			'home' => self::escape_pattern_string( trailingslashit( wp_parse_url( home_url( '/' ), PHP_URL_PATH ) ) ),
+			'site' => self::escape_pattern_string( trailingslashit( wp_parse_url( site_url( '/' ), PHP_URL_PATH ) ) ),
 		);
 	}
+
+	/**
+	 * Escapes a string for use in a URL pattern component.
+	 * https://urlpattern.spec.whatwg.org/#escape-a-pattern-string
+	 *
+	 * @since n.e.x.t
```
westonruter: nit

```suggestion
	 *
	 * @link https://urlpattern.spec.whatwg.org/#escape-a-pattern-string
	 * @since n.e.x.t
```

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
westonruter: Or rather:

```suggestion
		if ( preg_match( '/[:?#]/', $context_path ) ) {
```

This would be the more common way to do it (and perhaps more readable?) There are 31 uses of `strcspn()` in core, whereas there are 635 of `preg_replace()`.

I don't feel strongly either way.

westonruter (changes requested):
Just a nit and a comment.

adamsilverstein (approved):
Looks good!

modules/js-and-css/speculation-rules/class-plsr-url-pattern-prefixer.php
jeremyroman: It's probably being a C++ developer by day, but my mild preference is for the most efficient option if the options are similar in readability (though obviously neither is going to have a massive effect on your WordPress performance).

Happy to change if you felt more strongly; it seems not, so leaving this as-is for now.

westonruter (approved):

PR #952, opened 2024-01-22
Force HTML5 script theme support when printing JSON script

westonruter:
## Summary

When a theme doesn't declare theme support for HTML5 scripts, the JSON `script` is currently output with some XHTML compatibility wrappers:

```html
<script type="speculationrules">
/* <![CDATA[ */
{"prerender":[{"source":"document","where":{"and":[{"href_matches":"\/*"},{"not":{"href_matches":["\/wp-login.php","\/wp-admin\/*"]}},{"not":{"selector_matches":".no-prerender"}}]},"eagerness":"moderate"}]}
/* ]]> */
</script>
```

This breaks the parsing of the JSON. The `/* <![CDATA[ */` and `/* ]]> */` need to be omitted from JSON output. 

## Relevant technical choices

This needs to be fixed in core (ticket and PR pending), but a quick workaround is to temporarily override the `$_wp_theme_features` global with the theme support added.

**Update:** See core ticket: https://core.trac.wordpress.org/ticket/60320

## Checklist

### DIFF
--- a/modules/js-and-css/speculation-rules/hooks.php
+++ b/modules/js-and-css/speculation-rules/hooks.php
@@ -20,7 +20,11 @@ function plsr_print_speculation_rules() {
 	}
 
 	// This workaround is needed for WP 6.4. See <https://core.trac.wordpress.org/ticket/56313>.
-	$needs_html5_workaround = ! current_theme_supports( 'html5', 'script' );
+	$needs_html5_workaround = (
+		! current_theme_supports( 'html5', 'script' ) &&
+		version_compare( get_bloginfo( 'version' ), '6.4', '>=' ) &&
+		version_compare( get_bloginfo( 'version' ), '6.5', '<' )
+	);
 	if ( $needs_html5_workaround ) {
 		$backup_wp_theme_features = $GLOBALS['_wp_theme_features'];
 		add_theme_support( 'html5', array( 'script' ) );
--- a/modules/js-and-css/speculation-rules/hooks.php
+++ b/modules/js-and-css/speculation-rules/hooks.php
@@ -19,6 +19,7 @@ function plsr_print_speculation_rules() {
 		return;
 	}
 
+	// This workaround is needed for WP 6.4. See <https://core.trac.wordpress.org/ticket/56313>.
 	$needs_html5_workaround = ! current_theme_supports( 'html5', 'script' );
 	if ( $needs_html5_workaround ) {
 		$backup_wp_theme_features = $GLOBALS['_wp_theme_features'];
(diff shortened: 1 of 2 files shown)


felixarntz:
@westonruter Can you open a Trac ticket for this? I think we should prioritize this, and it would be great to link to that in a code comment here.

If the fix in core is straightforward (which I would expect it to be), then we can maybe ship this in 6.5 already and thus limit or remove this workaround sooner than later.

westonruter:
> Can you open a Trac ticket for this? I think we should prioritize this, and it would be great to link to that in a code comment here.

Yes, I started doing that right after opening this PR (ref: "(ticket and PR pending)").

~Here it is: https://core.trac.wordpress.org/ticket/56313~

Oops. I mean here: https://core.trac.wordpress.org/ticket/60320

modules/js-and-css/speculation-rules/hooks.php
```diff
@@ -19,9 +19,24 @@ function plsr_print_speculation_rules() {
 		return;
 	}
 
+	// This workaround is needed for WP 6.4. See <https://core.trac.wordpress.org/ticket/56313>.
+	$needs_html5_workaround = (
+		! current_theme_supports( 'html5', 'script' ) &&
+		version_compare( get_bloginfo( 'version' ), '6.4', '>=' ) &&
+		version_compare( get_bloginfo( 'version' ), '6.5', '<' )
```
westonruter: Guard against `-alpha`, `-beta`, and `-RC` from messing up comparisons:

```suggestion
		version_compare( strtok( get_bloginfo( 'version' ), '-' ), '6.4', '>=' ) &&
		version_compare( strtok( get_bloginfo( 'version' ), '-' ), '6.5', '<' )
```

swissspidy (approved):

modules/js-and-css/speculation-rules/hooks.php
westonruter:
```php
var_dump( version_compare( '6.5-beta', '6.5', '<' ) ); // => true ❌
var_dump( version_compare( strtok( '6.5-beta', '-' ), '6.5', '<' ) ); // => false ✅
```

tests/modules/js-and-css/speculation-rules/speculation-rules-test.php
```diff
@@ -8,10 +8,40 @@
...
+	}
+
+	/**
+	 * @dataProvider data_provider_to_test_print_speculation_rules
+	 */
+	public function test_plsr_print_speculation_rules_without_html5_support( bool $html5_support ) {
+		if ( $html5_support ) {
+			add_theme_support( 'html5', array( 'script' ) );
+		} else {
+			remove_theme_support( 'html5' );
```
joemcgill, line 41: I don't remember if theme support get reset between tests. Any chance of this causing side effects for other tests?

joemcgill (approved):
Looks good to me. One question about the tests, but pre-approving for merge.

tests/modules/js-and-css/speculation-rules/speculation-rules-test.php
westonruter: Right. Good point. Fixed in a1037f80

PR #953, opened 2024-01-23
Add Webpack config

### LINKED ISSUE #226: Add Asset Bundling to Plugin
## Feature Description
At some point in the near future I think we're going to need some form of static asset (JS/CSS) bundling in the plugin. This can be achieved by adding in a relatively simple `webpack` configuration. 

There is `@wordpress/scripts` which can handle asset bundling and transpiling from ES6 to ES5. Im not sure if its as versatile as a custom webpack configuration. Mainly because it seems to be centred around Gutenberg Block development.

Im putting this ticket out there so we can get some ideas going about how to handle this moving forward.

### LINKED ISSUE #893: Add JS build step to utilize local copy of web-vitals.js
## Feature Description

As noted in https://github.com/WordPress/performance/pull/876#discussion_r1380553929, the Image Loading Optimization module (see #869) is currently loading web-vitals.js directly from unpkg.com:

https://github.com/WordPress/performance/blob/03f32520241e666b3a1234b181f5f3459425fe46/modules/images/image-loading-optimization/detection/detect.js#L317-L321

This is not ideal from a performance perspective (due to the additional DNS lookup) nor from a privacy perspective (potentially). In any case, for any future consideration for merging this module into WordPress core, a local copy off the JS file will certainly be required.

A build process is also needed for the proposed Partytown module (https://github.com/WordPress/performance/pull/556#discussion_r1409657945).

In short, I believe this would entail having a new command added to the npm scripts which would copy the `web-vitals.js` module script into the plugin, while keeping this copy `.gitignore`'ed. This file would nevertheless need to be included in any PerfLab plugin ZIP/deploy or when eventually when the standalone plugin is built.

thelovekesh:
## Summary
Fixes #893 
Fixes #226 

## Relevant technical choices
- Add Webpack config with CopyWebpackPlugin to copy vendor libs from source to destination.
- Add logic to extract the package version from the `package.json` file and return the version string using the `.asset.php` file at the Webpack runtime.

## Checklist


thelovekesh:
@westonruter I think I should also add a step in release workflow to install deps and run this command.

.gitignore
```diff
@@ -59,3 +59,8 @@ temp/
 ._*
 .Trashes
 .svn
+
+############
+## Vendor Libraries
+############
+modules/images/image-loading-optimization/web-vitals/
```
westonruter: I was thinking that we would copy the `dist/modules` directory, and if we do this then the version would have to be included as part of the directory name since adding a `ver` query arg to the initially-imported module wouldn't cache-bust the other modules it depends on. 

```suggestion
modules/images/image-loading-optimization/web-vitals-*/
```

Nevertheless, after re-looking at the `dist/modules` I see that they are not minified. Therefore, it probably is just better to import the standard build as you are doing in this PR.

bin/plugin/commands/copy-vendor-libs.js
```diff
@@ -0,0 +1,147 @@
...
+				)
+			);
+			process.exit( 0 );
+		}
+
+		modules.forEach( ( module ) => {
+			const { from, to } = config[ module ];
+			patterns.set( module, {
+				from,
+				to,
```
westonruter: If we're using modules, the module version will need to be appended here. But since the `web-vitals.js` standard build is being copied, then this isn't needed.

modules/images/image-loading-optimization/detection.php
```diff
@@ -35,6 +35,9 @@ function ilo_get_detection_script( string $slug, array $needed_minimum_viewport_
 	 */
 	$detection_time_window = apply_filters( 'ilo_detection_time_window', 5000 );
 
+	$web_vitals_version = json_decode( file_get_contents( __DIR__ . '/../../../package.json' ), true )['dependencies']['web-vitals']; // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents -- It's a local file, not a remote URL.
+	$web_vitals_lib_src = add_query_arg( 'ver', $web_vitals_version, plugin_dir_url( __FILE__ ) . '/web-vitals/web-vitals.js' );
```
westonruter: This is loading the non-module version, so adding a `ver` will work fine.

Nevertheless, for the development version of the plugin should we not load the file directly from `node_modules`? Once this is built for distribution, there won't be a `package.json` either, so we wouldn't be able to read from it. So perhaps for development we should have the following:

```suggestion
	$web_vitals_version = json_decode( file_get_contents( __DIR__ . '/../../../package.json' ), true )['dependencies']['web-vitals']; // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents -- It's a local file, not a remote URL.
	$web_vitals_lib_src = add_query_arg( 'ver', $web_vitals_version, plugin_dir_url( __FILE__ ) . '/../../../../node_modules/web-vitals/dist/web-vitals.js' );
```

Then as part of the plugin build process, these get rewritten as after `copy-vendor-libs` is run?

I think there should be some plugin-specific logic here:

https://github.com/WordPress/performance/blob/fd23971345680bf13d8e697f68bca5ad92eb855c/bin/plugin/commands/build-plugins.js#L67-L94

package.json
```diff
@@ -38,7 +41,8 @@
     "pretest-php-multisite": "wp-env run composer 'install --no-interaction'",
     "test-php-multisite": "wp-env run phpunit 'WP_MULTISITE=1 phpunit -c /var/www/html/wp-content/plugins/$(basename $(pwd))/tests/multisite.xml --verbose'",
     "wp-env": "wp-env",
-    "prepare": "husky install"
+    "prepare": "husky install && npm run copy-vendor-libs",
```
westonruter: If we don't need the vendor libs for development, then this wouldn't be needed here. It could rather be run as part of `build-plugins` above.

```suggestion
    "prepare": "husky install",
```

bin/plugin/commands/copy-vendor-libs.js
westonruter: Nevertheless, Partytown does have multiple files, so in that case there would be the need to have a versioned directory name and not rely on a `ver` query parameter, right?

westonruter (changes requested):

bin/plugin/commands/copy-vendor-libs.js
westonruter: Humm, also relying on `build-plugins` wouldn't help modules that may rely on vendor libs. So maybe my suggestion to link to `node_modules` directly isn't best.

modules/images/image-loading-optimization/detection.php
westonruter: Or not per https://github.com/WordPress/performance/pull/953#discussion_r1464035683

But still there is the question of where the package version would be obtained if the `package.json` isn't present.

.gitignore
thelovekesh: Marking as resolved since updated the implementation.

modules/images/image-loading-optimization/detection.php
```diff
@@ -35,6 +35,9 @@ function ilo_get_detection_script( string $slug, array $needed_minimum_viewport_
 	 */
 	$detection_time_window = apply_filters( 'ilo_detection_time_window', 5000 );
 
+	$web_vitals_version = require_once __DIR__ . '/detection/web-vitals/index.asset.php';
```
thelovekesh: @westonruter Do we need to add some notice in case this file does not exist? given it will throw a fatal error if doesn't exist and the user will be required to run `npm run build`.

westonruter: Yes, the plugin should short-circuit with an error if the file doesn't exist. Similar to AMP plugin or PWA plugin.

webpack.config.js
```diff
@@ -0,0 +1,73 @@
...
+	);
+
+	return {
+		...sharedConfig,
+		plugins: [
+			new CopyWebpackPlugin( {
+				patterns: [
+					{
+						from: `${ source }/dist/web-vitals.js`,
+						to: `${ destination }/web-vitals/index.js`,
```
westonruter: Since we're only copying one file, should this be copied to `detection/web-vitals.js` without the additional directory? And keeping the same `web-vitals.js` name is probably preferable.

.gitignore
```diff
@@ -59,3 +59,8 @@ temp/
 ._*
 .Trashes
 .svn
+
+############
+## Vendor Libraries
+############
+modules/images/image-loading-optimization/detection/web-vitals/*
```
westonruter:
```suggestion
modules/images/image-loading-optimization/detection/web-vitals*
```

webpack.config.js
westonruter:
```suggestion
						to: `${ destination }/web-vitals.js`,
```

webpack.config.js
```diff
@@ -0,0 +1,73 @@
...
+					},
+					{
+						from: `${ source }/package.json`,
+						to: `${ destination }/web-vitals/index.asset.php`,
```
westonruter:
```suggestion
						to: `${ destination }/web-vitals.asset.php`,
```

modules/images/image-loading-optimization/detection.php
```diff
@@ -35,6 +35,9 @@ function ilo_get_detection_script( string $slug, array $needed_minimum_viewport_
...
+	$web_vitals_lib_src = add_query_arg( 'ver', $web_vitals_version, plugin_dir_url( __FILE__ ) . '/detection/web-vitals/index.js' );
```
westonruter:
```suggestion
	$web_vitals_lib_src = add_query_arg( 'ver', $web_vitals_version, plugin_dir_url( __FILE__ ) . '/detection/web-vitals.js' );
```

modules/images/image-loading-optimization/detection.php
westonruter:
```suggestion
	$web_vitals_version = require_once __DIR__ . '/detection/web-vitals.asset.php';
```

modules/images/image-loading-optimization/detection.php
```diff
@@ -35,6 +35,9 @@ function ilo_get_detection_script( string $slug, array $needed_minimum_viewport_
 	 */
 	$detection_time_window = apply_filters( 'ilo_detection_time_window', 5000 );
 
+	$web_vitals_lib_data = require_once __DIR__ . '/detection/web-vitals.asset.php';
```
westonruter: This shouldn't be `require_once` because then if this function gets called again, it won't return anything.

See https://github.com/WordPress/gutenberg/blob/970488826c004014862b3f92458e102db2201d88/lib/client-assets.php#L198

```suggestion
	$web_vitals_lib_data = require __DIR__ . '/detection/web-vitals.asset.php';
```

webpack.config.js
```diff
@@ -0,0 +1,73 @@
...
+	output: {},
+};
+
+/**
+ * Transformer to get version from package.json and return it as a PHP file.
+ *
+ * @param {Buffer} content      The content as a Buffer of the file being transformed.
+ * @param {string} absoluteFrom The absolute path to the file being transformed.
+ *
+ * @return {string} The transformed content.
```
westonruter:
```suggestion
 * @return {Buffer|string} The transformed content.
```

westonruter (changes requested):
Just two tiny things left to improve!

.github/workflows/php-test.yml
```diff
@@ -52,6 +52,8 @@ jobs:
           cache: npm
       - name: npm install
         run: npm ci
+      - name: Assets build
+        run: npm run build
```
westonruter: Oh, I believe this should also be added to the `deploy-dotork.yml` and `deploy-standalone-plugins.yml` GHA workflows, right?

webpack.config.js
```diff
@@ -0,0 +1,73 @@
...
+					},
+				],
+			} ),
+			new WebpackBar( {
+				name: 'Web Vitals',
+				color: '#f5a623',
+			} ),
+		],
+	};
+};
```
westonruter, line 71: This is not necessary now, but would it be preferable in the future to have this code actually located in `modules/images/image-loading-optimization` (or for in the relevant subdirectory `modules` or `plugins` for each module/plugin)? Then the module/plugin-specific build information could be kept with the code it's related to. Clearly not important at present. Just an idea for the future.

thelovekesh: Yes, plugin-specific config can be moved, and `webpack.config.js` can act as a base config for all.

.github/workflows/php-test.yml
thelovekesh: Yes.

westonruter (approved):
Great!

westonruter:
@mukeshpanchal27 please give this a review as well, noting the GHA workflow changes, as this is closely related to the efforts elsewhere for standalone plugins and the build process.

modules/images/image-loading-optimization/can-load.php
```diff
@@ -0,0 +1,25 @@
...
+}
+
+return static function () {
+	if (
+		! file_exists( __DIR__ . '/detection/web-vitals.asset.php' ) ||
+		! file_exists( __DIR__ . '/detection/web-vitals.js' )
+	) {
+		return new WP_Error(
+			'perflab_missing_web_vitals_library',
+			__( 'The Web Vitals library is missing. Please do "npm install &amp;&amp; npm run build" to finish installing the plugin.', 'performance-lab' )
```
mukeshpanchal27:
```suggestion
			sprintf(
				/* translators: npm command. */
				esc_html__( 'The Web Vitals library is missing. Please do "%s" to finish installing the plugin.', 'performance-lab' ),
				'npm install &amp;&amp; npm run build'
			)
```

phpcs.xml.dist
```diff
@@ -103,5 +103,10 @@
 		</properties>
 	</rule>
 
+	<!-- Check up to 20 files simultaneously. -->
+	<arg name="parallel" value="20"/>
+
+	<exclude-pattern>*.asset.php</exclude-pattern>
```
mukeshpanchal27, line 109: Could we also add the directory path? Otherwise, it will exclude all files, and in the future, if anyone adds a similar file, it will also be excluded.

phpcs.xml.dist
```diff
@@ -103,5 +103,10 @@
 		</properties>
 	</rule>
 
+	<!-- Check up to 20 files simultaneously. -->
+	<arg name="parallel" value="20"/>
```
mukeshpanchal27, line 107: Is there any specific reason for changing this?

.gitignore
```diff
@@ -59,3 +59,8 @@ temp/
 ._*
 .Trashes
 .svn
+
+############
+## Vendor Libraries
+############
+modules/images/image-loading-optimization/detection/web-vitals*
```
mukeshpanchal27: Could we add these in `Vendor` above in line no 28?

modules/images/image-loading-optimization/detection.php
```diff
@@ -45,7 +48,9 @@ function ilo_get_detection_script( string $slug, array $needed_minimum_viewport_
 		'urlMetricsNonce'             => ilo_get_url_metrics_storage_nonce( $slug ),
 		'neededMinimumViewportWidths' => $needed_minimum_viewport_widths,
 		'storageLockTTL'              => ilo_get_url_metric_storage_lock_ttl(),
+		'webVitalsLibrarySrc'         => $web_vitals_lib_src,
```
mukeshpanchal27, line 51:
```suggestion
		'webVitalsLibrarySrc'         => esc_url( $web_vitals_lib_src ),
```

Escaping needed here?

mukeshpanchal27 (commented):
Thanks @thelovekesh for the PR and @westonruter for the ping. Left some feedbacks.

modules/images/image-loading-optimization/detection.php
thelovekesh: yes, given https://github.com/WordPress/wordpress-develop/blob/89c53f4db76bf4a0f541516f8032dac690af8a3a/src/wp-includes/functions.php#L1119-L1121

phpcs.xml.dist
thelovekesh: Just to speed up linting.

<details><summary>Local Machine</summary>

Before:
```sh
$ ./vendor/bin/phpcs
............................................................ 60 / 86 (70%)
..........................                                   86 / 86 (100%)


Time: 2.55 secs; Memory: 38MB
```

After:
```sh
$ ./vendor/bin/phpcs                   
.................. 18 / 18 (100%)


Time: 595ms; Memory: 14MB
```

</details> 

<details><summary>CI</summary>

Before:
```sh
> phpcs --standard=phpcs.xml.dist
............................................................ 60 / 71 (85%)
...........                                                  71 / 71 (100%)


Time: 12.85 secs; Memory: 36MB
```

After:
```sh
> phpcs --standard=phpcs.xml.dist
.................. 18 / 18 (100%)


Time: 7.48 secs; Memory: 14MB
```

</details>

phpcs.xml.dist
thelovekesh: `.asset.php` is commonly generated assets manifest file by the Webpack dependency extraction plugin. Since we don't commit dist files, can we keep it ignored? Instead, if someone intentionally includes such a file, can it be included manually in the git ignored list?

.gitignore
thelovekesh: https://github.com/WordPress/performance/pull/953/commits/c9bf7f93c70af036b8c12a98809bdd8abfde9037

modules/images/image-loading-optimization/detection.php
thelovekesh: https://github.com/WordPress/performance/pull/953/commits/70851eb33681b858249b0a0f07a4d5181daee7c2

thelovekesh:
Thanks @mukeshpanchal27. I think it's ready for your re-review.

phpcs.xml.dist
westonruter: Yeah, such files should always be generated and thus never be applicable for PHPCS.

westonruter: @thelovekesh Could you then also add `*.asset.php` to the `.gitignore`? That would be an additional safeguard and would likely be needed in the future anyway.

modules/images/image-loading-optimization/detection.php
```diff
@@ -45,7 +48,9 @@ function ilo_get_detection_script( string $slug, array $needed_minimum_viewport_
 		'urlMetricsNonce'             => ilo_get_url_metrics_storage_nonce( $slug ),
 		'neededMinimumViewportWidths' => $needed_minimum_viewport_widths,
 		'storageLockTTL'              => ilo_get_url_metric_storage_lock_ttl(),
+		'webVitalsLibrarySrc'         => esc_url( $web_vitals_lib_src ),
```
westonruter: Actually, this should not be escaped. This is not where it is rendered. It will be escaped by `json_encode()` below. At most it could be sanitized with `esc_url_raw()` but I think this would be overkill since it is essentially a string literal defined above.

```suggestion
		'webVitalsLibrarySrc'         => $web_vitals_lib_src,
```

Importantly, `esc_url()` is for _HTML attribute_ escaping, so it will output `&` as `&amp;` which will not work properly in a script context.

modules/images/image-loading-optimization/can-load.php
```diff
@@ -0,0 +1,29 @@
...
+	if (
+		! file_exists( __DIR__ . '/detection/web-vitals.asset.php' ) ||
+		! file_exists( __DIR__ . '/detection/web-vitals.js' )
+	) {
+		return new WP_Error(
+			'perflab_missing_web_vitals_library',
+			sprintf(
+				/* translators: npm command. */
+				esc_html__( 'The Web Vitals library is missing. Please do "%s" to finish installing the plugin.', 'performance-lab' ),
+				'npm install &amp;&amp; npm run build'
```
westonruter: This shouldn't be escaped here because it is escaped later at rendering:

https://github.com/WordPress/performance/blob/0e7f28516e452e107fbaf66a3a548c274f230878/admin/load.php#L182

```suggestion
				__( 'The Web Vitals library is missing. Please do "%s" to finish installing the plugin.', 'performance-lab' ),
				'npm install && npm run build'
```

phpcs.xml.dist
adamsilverstein: Nice!

mukeshpanchal27 (approved):

PR #957, opened 2024-01-29
Sanitize metric name for `Server-Timing` header

### LINKED ISSUE #955: Server-Timing: Slashes in metric name not working
Hello.

Currently, when using a hook name such as `render_block_core/site-logo` only callbacks attached to the `render_block_core` are measured.

swissspidy:
## Summary

Ensures metric name does not contain invalid characters such as "/".

Fixes #955

## Checklist

### DIFF
--- a/server-timing/class-perflab-server-timing.php
+++ b/server-timing/class-perflab-server-timing.php
@@ -260,7 +260,7 @@ class Perflab_Server_Timing {
 		}
 
 		// See https://github.com/WordPress/performance/issues/955.
-		$name = str_replace( '/', '-', $metric->get_slug() );
+		$name = preg_replace( '/[^!#$%&\'*+\-.^_`|~0-9a-zA-Z]/', '-', $metric->get_slug() );
 
 		return sprintf( 'wp-%1$s;dur=%2$s', $name, $value );
 	}
--- a/server-timing/class-perflab-server-timing.php
+++ b/server-timing/class-perflab-server-timing.php
@@ -95,7 +95,7 @@ class Perflab_Server_Timing {
 			return;
 		}
 
-		// See https://github.com/WordPress/performance/issues/955
+		// See https://github.com/WordPress/performance/issues/955.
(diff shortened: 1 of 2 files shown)


server-timing/class-perflab-server-timing.php
```diff
@@ -258,6 +258,10 @@ private function format_metric_header_value( Perflab_Server_Timing_Metric $metri
 		if ( is_float( $value ) ) {
 			$value = round( $value, 2 );
 		}
-		return sprintf( 'wp-%1$s;dur=%2$s', $metric->get_slug(), $value );
+
+		// See https://github.com/WordPress/performance/issues/955.
+		$name = str_replace( '/', '-', $metric->get_slug() );
```
westonruter: Looking up the `Server-Timing` field definition, it is [comprised](https://www.w3.org/TR/server-timing/#the-server-timing-header-field) as follows:

```
  Server-Timing             = #server-timing-metric
  server-timing-metric      = metric-name *( OWS ";" OWS server-timing-param )
  metric-name               = token
  server-timing-param       = server-timing-param-name OWS "=" OWS server-timing-param-value
  server-timing-param-name  = token
  server-timing-param-value = token / quoted-string
```

So `$name` here should be `metric-name` which is a `token`. The spec then links off to RFC 7230 which [defines](https://httpwg.org/specs/rfc7230.html#rfc.section.3.2.6) `token` as:

```
  token          = 1*tchar

  tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                 / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~" 
                 / DIGIT / ALPHA
                 ; any VCHAR, except delimiters
```

So should more characters be replaced with `-` than just `/`? Maybe something like:

```suggestion
		$name = preg_replace( '/[^!#$%&\'*+\-.^_`|~0-9a-zA-Z]/', '-', $metric->get_slug() );
```

westonruter (approved):

adamsilverstein (approved):
Nice work!

PR #962, opened 2024-01-30
Run build-cs composer install/update via post-install/update-cmd scripts

### LINKED ISSUE #961: Lint-staged fails if composer not installed in build-cs
## Bug Description

I just tried merging `trunk` into `feature/image-loading-optimization`. After I resolved conflicts, I attempted to commit but the pre-commit check fails:

```
✔ Preparing lint-staged...
❯ Running tasks for staged files...
  ❯ package.json — 32 files
    ❯ *.php — 16 files
      ✖ composer run-script lint [FAILED]
      ◼ composer run-script phpstan
    ❯ *.js — 1 file
      ✖ npm run lint-js
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up temporary files...

✖ composer run-script lint:
> composer --working-dir=build-cs update --no-interaction 
...
Cannot update only a partial set of packages without a lock file present. Run `composer update` to generate a lock file.
Script composer --working-dir=build-cs update --no-interaction handling the lint event returned with error code 3
✔ Preparing lint-staged...
❯ Running tasks for staged files...
  ❯ package.json — 32 files
    ❯ *.php — 16 files
      ✖ composer run-script lint [FAILED]
      ◼ composer run-script phpstan
    ❯ *.js — 1 file
      ✖ npm run lint-js [KILLED]
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up temporary files...
husky - pre-commit hook exited with code 1 (error)
```

PHPCS apparently this is because the `build-cs` directory has never had `composer install` run on it before.

If I modify `composer.json` as follows to `composer install` instead of `composer update`:

```diff
--- a/composer.json
+++ b/composer.json
@@ -21,15 +21,15 @@
   },
   "scripts": {
     "phpstan": [
-      "composer --working-dir=build-cs update --no-interaction",
+      "composer --working-dir=build-cs install --no-interaction",
       "build-cs/vendor/bin/phpstan analyse --memory-limit=2048M -c phpstan.neon.dist"
     ],
     "format": [
-      "composer --working-dir=build-cs update --no-interaction",
+      "composer --working-dir=build-cs install --no-interaction",
       "build-cs/vendor/bin/phpcbf --standard=phpcs.xml.dist --report-summary --report-source"
     ],
     "lint": [
-      "composer --working-dir=build-cs update --no-interaction",
+      "composer --working-dir=build-cs install --no-interaction",
       "build-cs/vendor/bin/phpcs --standard=phpcs.xml.dist"
     ],
     "test": "phpunit -c phpunit.xml.dist --verbose",
```

I then get a different error:

```
✔ Preparing lint-staged...
❯ Running tasks for staged files...
  ❯ package.json — 1 file
    ❯ *.php — 1 file
      ✖ composer run-script lint [FAILED]
      ◼ composer run-script phpstan
    ↓ *.js — no files [SKIPPED]
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up temporary files...

✖ composer run-script lint:
> composer --working-dir=build-cs install --no-interaction '/home/westonruter/repos/performance/load.php'
Invalid argument /home/westonruter/repos/performance/load.php. Use "composer require /home/westonruter/repos/performance/load.php" instead to add packages to your composer.json.
Script composer --working-dir=build-cs install --no-interaction handling the lint event returned with error code 1
husky - pre-commit hook exited with code 1 (error)
```

This is apparently because lint-staged is passing the modified file (here `/home/westonruter/repos/performance/load.php`) to _both_ subcommands of the `lint` command:

```
    "lint": [
      "composer --working-dir=build-cs install --no-interaction",
      "build-cs/vendor/bin/phpcs --standard=phpcs.xml.dist"
    ],
```

I suggest that we not run `composer update` in the subcommands for `lint`, `format`, and `phpstan` because (1) the lock file may not be present so it will fail and (2) it slows down running those commands. Instead, I suggest that we do `composer install` and `composer update` for the `build-cs` directory via top-level `post-install-cmd` and `post-update-cmd` scripts, respectively.

## Steps to reproduce

1. Modify a PHP file and stage the change.
2. Delete the `build-cs/composer.lock` file and `build-cs/vendor` directory (if they even exist yet)
3. Try to commit the change to the PHP file.

westonruter:
## Summary

Fixes #961

This fixes the composer scripts for `lint`, `phpstan`, and `format` if `composer install` has not been run in the `build-cs` directory before. It also speeds up running those scripts by avoiding running `composer update` with each call.

## Relevant technical choices

Adds a `post-install-cmd` in the root `composer.json` which runs `composer install` in the `build-cs` directory if the PHP version is at least 8.1. Likewise adds a `post-update-cmd` to call run `composer update` in that directory. Additionally, when running PHPUnit tests on the various versions of PHP in the GHA workflow, instead of calling `composer update` directly it instead calls `composer update` via `wp-env`.

This fixes an issue which was introduced in #544.

## Checklist

### DIFF
--- a/.github/workflows/php-test.yml
+++ b/.github/workflows/php-test.yml
@@ -67,6 +67,9 @@ jobs:
         run: npm ci
       - name: Install WordPress
         run: npm run wp-env start
+        # Note that `composer update` is required instead of `composer install`
+        # for the sake of PHP versions older than 8.1, which is the version of
+        # PHP that the composer.lock was created for.
       - name: Composer update
         run: npm run wp-env run tests-cli -- --env-cwd="wp-content/plugins/$(basename $(pwd))" composer update --no-interaction
       - name: Running single site unit tests
--- a/.github/workflows/php-test.yml
+++ b/.github/workflows/php-test.yml
@@ -58,13 +58,22 @@ jobs:
     steps:
       - uses: styfle/cancel-workflow-action@0.11.0
       - uses: actions/checkout@v3
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
       - name: Setup Node.js (.nvmrc)
         uses: actions/setup-node@v3
         with:
           node-version-file: '.nvmrc'
           cache: npm
-      - name: Composer update
-        run: composer update
+      - name: General debug information
+        run: |
+          npm --version
+          node --version
+          composer --version
+          php -v
+      - name: Composer install
+        run: composer install
       - name: npm install
         run: npm ci
       - name: Install WordPress
--- a/.github/workflows/php-test.yml
+++ b/.github/workflows/php-test.yml
@@ -58,22 +58,13 @@ jobs:
     steps:
       - uses: styfle/cancel-workflow-action@0.11.0
       - uses: actions/checkout@v3
-      - uses: shivammathur/setup-php@v2
-        with:
-          php-version: ${{ matrix.php }}
       - name: Setup Node.js (.nvmrc)
         uses: actions/setup-node@v3
         with:
           node-version-file: '.nvmrc'
           cache: npm
-      - name: General debug information
-        run: |
-          npm --version
-          node --version
-          composer --version
-          php -v
-      - name: Composer install
-        run: composer install
+      - name: Composer update
+        run: composer update
       - name: npm install
         run: npm ci
       - name: Install WordPress
--- a/.github/workflows/php-test.yml
+++ b/.github/workflows/php-test.yml
@@ -68,7 +68,7 @@ jobs:
       - name: Install WordPress
         run: npm run wp-env start
       - name: Composer install
-        run: npm run wp-env run tests-cli -- --env-cwd='wp-content/plugins/$(basename $(pwd))' composer install --no-interaction
+        run: npm run wp-env run tests-cli -- --env-cwd="wp-content/plugins/$(basename $(pwd))" composer install --no-interaction
       - name: Running single site unit tests
         run: npm run test-php
       - name: Running multisite unit tests
--- a/.github/workflows/php-test.yml
+++ b/.github/workflows/php-test.yml
@@ -68,7 +68,7 @@ jobs:
       - name: Install WordPress
         run: npm run wp-env start
       - name: Composer install
-        run: npm run wp-env run tests-cli --env-cwd='wp-content/plugins/$(basename $(pwd))' composer install --no-interaction
+        run: npm run wp-env run tests-cli -- --env-cwd='wp-content/plugins/$(basename $(pwd))' composer install --no-interaction
       - name: Running single site unit tests
         run: npm run test-php
       - name: Running multisite unit tests
--- a/.github/workflows/php-test.yml
+++ b/.github/workflows/php-test.yml
@@ -63,12 +63,12 @@ jobs:
         with:
           node-version-file: '.nvmrc'
           cache: npm
-      - name: Composer update
-        run: composer update
       - name: npm install
         run: npm ci
       - name: Install WordPress
         run: npm run wp-env start
+      - name: Composer install
+        run: npm run wp-env run tests-cli --env-cwd='wp-content/plugins/$(basename $(pwd))' composer install --no-interaction
       - name: Running single site unit tests
         run: npm run test-php
       - name: Running multisite unit tests
--- a/.github/workflows/php-test.yml
+++ b/.github/workflows/php-test.yml
@@ -67,8 +67,8 @@ jobs:
         run: npm ci
       - name: Install WordPress
         run: npm run wp-env start
-      - name: Composer install
-        run: npm run wp-env run tests-cli -- --env-cwd="wp-content/plugins/$(basename $(pwd))" composer install --no-interaction
+      - name: Composer update
+        run: npm run wp-env run tests-cli -- --env-cwd="wp-content/plugins/$(basename $(pwd))" composer update --no-interaction
       - name: Running single site unit tests
         run: npm run test-php
       - name: Running multisite unit tests
--- a/.github/workflows/php-test-standalone-plugins.yml
+++ b/.github/workflows/php-test-standalone-plugins.yml
@@ -66,14 +66,14 @@ jobs:
         with:
           node-version-file: '.nvmrc'
           cache: npm
+      - name: npm install
+        run: npm ci
       - name: General debug information
         run: |
           npm --version
           node --version
           composer --version
           php -v
-      - name: npm install
-        run: npm ci
       - name: Building standalone plugins
         run: npm run build-plugins
       - name: Running single site standalone plugin integration tests
--- a/.github/workflows/php-test-standalone-plugins.yml
+++ b/.github/workflows/php-test-standalone-plugins.yml
@@ -66,14 +66,14 @@ jobs:
         with:
           node-version-file: '.nvmrc'
           cache: npm
-      - name: npm install
-        run: npm ci
       - name: General debug information
         run: |
           npm --version
           node --version
           composer --version
           php -v
+      - name: npm install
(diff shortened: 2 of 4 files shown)


swissspidy (approved):

westonruter:
Naturally running `composer install` fails on the older PHP versions, for example PHP 7.0: https://github.com/WordPress/performance/actions/runs/7716621796/job/21033857878

So in 3df9c5d I made the install/update conditional if the system is on PHP 8+. That made the jobs pass, but for some reason it still seems like `composer update` is running:

https://github.com/WordPress/performance/actions/runs/7716757155/job/21034333302#step:5:78

I was expecting to see:

> Skipping composer update for build-cs since not on PHP 8.

mukeshpanchal27 (approved):

mukeshpanchal27:
@westonruter @swissspidy not sure if it's related issue but `npm run lint-php` is not working.

Error:
```
> prelint-php
> wp-env run composer 'install --no-interaction'

wp-env run <container> [command...]

Runs an arbitrary command in one of the underlying Docker containers. A double
dash can be used to pass arguments to the container without parsing them. This
is necessary if you are using an option that is defined below. You can use
`bash` to open a shell session and both `composer` and `phpunit` are available
in all WordPress and CLI containers. WP-CLI is also available in the CLI
containers.

Positionals:
  container  The underlying Docker service to run the command on.
              [string] [required] [choices: "mysql", "tests-mysql", "wordpress",
                                          "tests-wordpress", "cli", "tests-cli"]
  command    The command to run.                           [array] [default: []]

Options:
  --help     Show help                                                 [boolean]
  --debug    Enable debug output.                     [boolean] [default: false]
  --version  Show version number                                       [boolean]
  --env-cwd  The command's working directory inside of the container. Paths
             without a leading slash are relative to the WordPress root.
                                                         [string] [default: "."]

The 'composer' container has been removed. Please use 'wp-env run cli --env-cwd=wp-content/path/to/plugin composer' instead.
```

westonruter:
@mukeshpanchal27 ah, well it seems that is obsolete now anyway. I think we should delete this:

https://github.com/WordPress/performance/blob/3df9c5d9390240bb84b841438d8852e3a4c0159c/package.json#L34

That was slowing down linting anyway. No need to re-install with every call. Does it work for you with that removed?

westonruter:
@mukeshpanchal27 I've removed those pre scripts in 2591052. The [Prerequisites](https://make.wordpress.org/performance/handbook/performance-lab/#prerequisites) section of the handbook page indicates that Composer is optional but the [Set up the development environment](https://make.wordpress.org/performance/handbook/performance-lab/#set-up-the-development-environment) section does not. I think contributors should have to always do `npm install` and `composer install` to contribute.

.github/workflows/php-test.yml
```diff
@@ -58,13 +58,22 @@ jobs:
     steps:
       - uses: styfle/cancel-workflow-action@0.11.0
       - uses: actions/checkout@v3
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
```
westonruter: Oh, I guess this isn't needed because it's running unit tests in the container.

westonruter: But it would allow `composer install` to work properly without having to run it in the container?

westonruter:
I don't get it. The [`--env-cwd`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env/#:~:text=need%20to%20pass-,%2D%2Denv%2Dcwd,-to%20make%20sure) arg is supposed to allow changing the working directory for running a command with `wp-env run`. However, it doesn't seem to be working.

On my local machine, I'm supposed to be able to do:

```
npm run wp-env run tests-cli --env-cwd='wp-content/plugins/$(basename $(pwd))' pwd
```

And it should output:

```
/var/www/html/wp-content/plugins/performance
```

But it isn't. It's outputting instead:

```
/var/www/html
```

As if the `--env-cwd` wasn't even supplied.

@mukeshpanchal27 As for the error you're seeing in https://github.com/WordPress/performance/pull/962#issuecomment-1918371416, that makes sense because the `composer` container was [removed in v7.0.0](https://github.com/WordPress/gutenberg/blob/trunk/packages/env/CHANGELOG.md#700-2023-05-10):

<blockquote>Remove the <code>composer</code> and <code>phpunit</code> Docker containers. If you are currently using the <code>run composer</code> or <code>run phpunit</code> command you can migrate to <code>run cli composer</code> or <code>run tests-cli phpunit</code> respectively. Note that with <code>composer</code>, you will need to use the <code>--env-cwd</code> option to navigate to your plugin's directory as it is no longer the default working directory.</blockquote>

So it makes sense that this wouldn't have worked anymore and should be removed (or at least updated):

https://github.com/WordPress/performance/blob/4396a841b28a6184594907c11e18458f5cfe6343/package.json#L34

But again I don't get why `--env-cwd` isn't working.

**Update:** An additional `--` needed to be inserted before the argument to ensure it was passed properly. Props @joemcgill in 9e926dd.

westonruter:
Converting to draft because this needs some more work, namely to fix running `composer install` in the `tests-cli` container.

westonruter:
OK, I see why `composer update` was needed instead of `composer install`. Composer would fail to install on PHP versions older than 8.1 which is the version the `composer.lock` was created for:

```
> wp-env
> wp-env run tests-cli --env-cwd=wp-content/plugins/performance composer install --no-interaction

ℹ Starting 'composer install --no-interaction' on the tests-cli container. 

Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - doctrine/instantiator is locked to version 2.0.0 and an update of this package was not requested.
    - doctrine/instantiator 2.0.0 requires php ^8.1 -> your php version (7.4.33) does not satisfy that requirement.
  Problem 2
    - doctrine/instantiator 2.0.0 requires php ^8.1 -> your php version (7.4.33) does not satisfy that requirement.
    - phpunit/phpunit [9](https://github.com/WordPress/performance/actions/runs/7733447799/job/21085541195?pr=962#step:7:10).6.[16](https://github.com/WordPress/performance/actions/runs/7733447799/job/21085541195?pr=962#step:7:17) requires doctrine/instantiator ^1.3.1 || ^2 -> satisfiable by doctrine/instantiator[2.0.0].
    - phpunit/phpunit is locked to version 9.6.16 and an update of this package was not requested.

✖ Command failed with exit code 2
Command failed with exit code 2
Error: Process completed with exit code 1.
```

This being said, I wonder if we should even be committing the `composer.lock` file? I guess it is still beneficial for local development, but it seems unfortunate to have to run it in GHA for older versions of PHP since (1) it takes longer and (2) the package versions won't always be consistent.

westonruter:
> I was expecting to see:
> 
> > Skipping composer update for build-cs since not on PHP 8.

OK, I'm seeing it now: https://github.com/WordPress/performance/actions/runs/7733545501/job/21085843025#step:7:99

westonruter:
Alright, I think it's in a good state now. As can be seen in the PHP 7.0 job for running PHPUnit, if you run `composer update` you now get the [following](https://github.com/WordPress/performance/actions/runs/7733725872/job/21086363124?pr=962#step:7:104) in the terminal:

```
Skipping composer update for build-cs since not on PHP 8.1+. You are running: 
PHP 7.0.33 (cli) (built: Dec 21 2018 03:17:53) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.33, Copyright (c) 1999-2017, by Zend Technologies
```

PR #963, opened 2024-01-31
Fix tag processor for compatibility with WP 6.5

westonruter:
## Summary

WordPress 6.5 has a back-compat breakage in the `WP_HTML_Tag_Processor`. It will no longer visit closing tags for open tags that have raw text contents. See Core-60392.

## Relevant technical choices

This adds a new `RAW_TEXT_TAGS` class constant which is used in the same way as the `VOID_TAGS` constant. A unit test is updated to check the tags that WP 6.5 does not explicitly visit the closers for.

## Checklist

### DIFF
--- a/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
+++ b/modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
@@ -47,6 +47,38 @@ final class ILO_HTML_Tag_Processor {
 	);
 
 	/**
+	 * Raw text tags.
+	 *
+	 * These are treated like void tags for the purposes of walking over the document since we do not process any text
+	 * nodes. To cite the docblock for WP_HTML_Tag_Processor:
+	 *
+	 * > Some HTML elements are handled in a special way; their start and end tags
+	 * > act like a void tag. These are special because their contents can't contain
+	 * > HTML markup. Everything inside these elements is handled in a special way
+	 * > and content that _appears_ like HTML tags inside of them isn't. There can
+	 * > be no nesting in these elements.
+	 * >
+	 * > In the following list, "raw text" means that all of the content in the HTML
+	 * > until the matching closing tag is treated verbatim without any replacements
+	 * > and without any parsing.
+	 *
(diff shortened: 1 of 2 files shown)


westonruter:
[Verbal approval](https://core.trac.wordpress.org/ticket/60392#comment:7) by @dmsnell in Core-60392.

modules/images/image-loading-optimization/class-ilo-html-tag-processor.php
```diff
@@ -46,6 +46,38 @@ final class ILO_HTML_Tag_Processor {
...
+	 * > In the following list, "raw text" means that all of the content in the HTML
+	 * > until the matching closing tag is treated verbatim without any replacements
+	 * > and without any parsing.
+	 *
+	 * @link https://github.com/WordPress/wordpress-develop/blob/6dd00b1ffac54c20c1c1c7721aeebbcd82d0e378/src/wp-includes/html-api/class-wp-html-tag-processor.php#L136-L155
+	 * @link https://core.trac.wordpress.org/ticket/60392#comment:2
+	 *
+	 * @var string[]
+	 */
+	const RAW_TEXT_TAGS = array(
```
dmsnell, line 70: "raw" isn't technically correct here for all the tags but I think it's fine. I just wanted to point out that the real qualifier here is whether the tags can contain HTML, or "DATA" content. some of these are RAW TEXT and some are RCDATA content. the distinction is whether character references are decoded, but it's not important to this check.

westonruter: I picked raw for the name since it was used to [describe the list](https://github.com/WordPress/wordpress-develop/blob/6dd00b1ffac54c20c1c1c7721aeebbcd82d0e378/src/wp-includes/html-api/class-wp-html-tag-processor.php#L144-L155) of tags in code comment, but I realize there are nuances to [how they are actually processed](https://github.com/WordPress/wordpress-develop/blob/6dd00b1ffac54c20c1c1c7721aeebbcd82d0e378/src/wp-includes/html-api/class-wp-html-tag-processor.php#L949-L975).

westonruter: Ah, but I see that it is not intending to describe all of the tags, but just the use of "raw text" in each tag's comment.

dmsnell (approved):

swissspidy (approved):

PR #967, opened 2024-02-02
Update test to use non-AVIF format since now supported in 6.5-alpha

westonruter:
Now that AVIF support just landed in WP trunk via Core-51228, we need to use a different image format to check for an invalid MIME type. This fixes a unit test failure on 6.5-alpha.

## Checklist


adamsilverstein (approved):
Looks good! I've never seen a  .pcx file (but that is the point I guess)

westonruter:
> Looks good! I've never seen a .pcx file (but that is the point I guess)

I used the [PCX format](https://en.wikipedia.org/wiki/PCX) a lot with [PC Paintbrush](https://en.wikipedia.org/wiki/PC_Paintbrush) as a kid 😄

joemcgill (approved):
This works.

PR #972, opened 2024-02-06
Move `auto-sizes` to plugins folder

mukeshpanchal27:
## Summary

Follow-up work https://github.com/WordPress/performance/issues/654#issuecomment-1898808673

## Checklist


.github/workflows/php-test-auto-sizes-plugin.yml
```diff
@@ -0,0 +1,78 @@
+name: Unit Testing for Auto-sizes for Lazy-Loaded Images Plugin
+
+on:
+  push:
+    branches:
+      - trunk
+      - 'release/**'
+    # Only run if PHP-related files changed.
```
swissspidy: Probably should also include plugins/auto-sizes here, no?

.github/workflows/php-test-auto-sizes-plugin.yml
```diff
@@ -0,0 +1,78 @@
+name: Unit Testing for Auto-sizes for Lazy-Loaded Images Plugin
```
swissspidy: We'll end up with _a lot_ of jobs if we have a separate workflow for each plugin, and then each workflow running a matrix of like 10 items.

Should we just have 1 file for running all these test suites?

phpunit.xml.dist
```diff
@@ -12,6 +12,9 @@
 			<exclude>./tests/utils</exclude>
 			<exclude>./tests/plugins</exclude>
 		</testsuite>
+		<testsuite name="auto-sizes">
+			<directory suffix=".php">./tests/plugins/auto-sizes</directory>
+    </testsuite>
```
joemcgill: Whitespace issue.

```suggestion
		<testsuite name="auto-sizes">
			<directory suffix=".php">./tests/plugins/auto-sizes</directory>
		</testsuite>
```

phpcs.xml.dist
```diff
@@ -8,7 +8,7 @@
 	<rule ref="WordPress-Docs"/>
 	<rule ref="WordPress-Extra" />
 	<rule ref="WordPress.WP.I18n"/>
-	<config name="text_domain" value="performance-lab,default"/>
+	<config name="text_domain" value="performance-lab,default,auto-sizes"/>
```
joemcgill, line 11: We may want to have separate `phpcs.xml.dist` files in each plugin rather than putting all the text domains in the same file. This would help us catch situations where someone accidentally used the wrong text domain from one plugin in another because of a sloppy copy/paste. Not a blocker for now though.

plugins/auto-sizes/load.php
```diff
@@ -0,0 +1,30 @@
+<?php
```
joemcgill: I question the use of "load.php" as the main plugin file once we've moved these from modules. I think a better filename would be  to use the common slug convention `plugins/auto-sizes/auto-sizes.php` or `plugins/auto-sizes/plugin.php`. Not a technical blocker, but is something that came up in the plugin review when it was first submitted. In the SVN repo, it's currently [using the former](https://plugins.trac.wordpress.org/browser/auto-sizes/trunk), I just hadn't updated the PR yet.

tests/bootstrap.php
```diff
@@ -55,5 +55,28 @@ static function () {
...
+
+	foreach ( $plugin_folders as $plugin_folder ) {
+		$plugin_name      = basename( rtrim( $plugin_folder, '/' ) );
+		$plugin_test_path = TESTS_PLUGIN_DIR . '/plugins/' . $plugin_name;
+
+		if ( file_exists( $plugin_test_path ) ) {
+			tests_add_filter(
+				'plugins_loaded',
+				static function () use ( $plugin_test_path ) {
+					require_once $plugin_test_path . '/load.php';
```
joemcgill: If we change the main plugin file names, this would also need to be updated.

tests/bootstrap.php
```diff
@@ -55,5 +55,28 @@ static function () {
 	1
 );
 
+
+$testsuite_count = array_count_values( $_SERVER['argv'] )['--testsuite'];
```
joemcgill: Rather than inventing our own CLI args, I think we could just use the `--group` or `--filter` options for PHPUnit to run specific suites and load all of the plugins like do for modules above.

joemcgill (commented):
I've left a few comments, but nothing I see as a real blocker. I'm mostly curious about why we wouldn't make these changes in #904 instead?

.github/workflows/php-test-auto-sizes-plugin.yml
mukeshpanchal27: In https://github.com/WordPress/performance/pull/972/commits/88040f2ed751b6ecb8d9108e2c471759e36e29e3 added single file that track `plugins/**.php` files.

.github/workflows/php-test-auto-sizes-plugin.yml
mukeshpanchal27: In https://github.com/WordPress/performance/pull/972/commits/88040f2ed751b6ecb8d9108e2c471759e36e29e3 added single file.

plugins/auto-sizes/load.php
mukeshpanchal27: Got it. Adjust in https://github.com/WordPress/performance/pull/972/commits/87f3a04c351ef37f9c3b411176a39f6399f48cb1

mukeshpanchal27:
Once #975 is merged, we need to rebase this PR.

tests/bootstrap.php
swissspidy: testsuite is a supported arg in PHPUnit

phpcs.xml.dist
mukeshpanchal27: Can we open issue for that and work on follow-up PR?

mukeshpanchal27:
@swissspidy @joemcgill @westonruter Could you please take another look when you have a moment?

phpunit.xml.dist
```diff
@@ -7,9 +7,16 @@
...
 			<directory suffix=".php">./tests</directory>
-    		<exclude>./tests/utils</exclude>
+			<exclude>./tests/utils</exclude>
+			<exclude>./tests/plugins</exclude>
+		</testsuite>
+		<testsuite name="auto-sizes">
+			<directory suffix=".php">./tests/plugins/auto-sizes</directory>
+		</testsuite>
+		<testsuite name="test-plugin">
+			<directory suffix=".php">./tests/plugins/test-plugin</directory>
```
swissspidy: Do we still need that test-plugin?

plugins/.gitkeep
swissspidy, line 1: No need for a gitkeep file if there are files in here anyway.

tests/plugins/test-plugin/test-plugin.php
swissspidy, line 1: Do we still need this?

tests/bootstrap.php
```diff
@@ -55,5 +55,37 @@ static function () {
...
+
+	foreach ( $plugin_folders as $plugin_folder ) {
+		$plugin_name      = basename( rtrim( $plugin_folder, '/' ) );
+		$plugin_test_path = TESTS_PLUGIN_DIR . '/plugins/' . $plugin_name;
+
+		if ( file_exists( $plugin_test_path ) ) {
+			tests_add_filter(
+				'plugins_loaded',
+				static function () use ( $plugin_test_path, $plugin_name ) {
+					// Check if plugin have "plugin/plugin.php" file.
```
swissspidy:
```suggestion
					// Check if plugin has a "plugin/plugin.php" file.
```

tests/bootstrap.php
```diff
@@ -55,5 +55,37 @@ static function () {
...
+					if ( file_exists( $plugin_test_path . '/' . $plugin_name . '.php' ) ) {
+						require_once $plugin_test_path . '/' . $plugin_name . '.php';
+						return;
+					}
+
+					// Check if plugin have "plugin/load.php" file.
```
swissspidy:
```suggestion
					// Check if plugin has a "plugin/load.php" file.
```

tests/bootstrap.php
swissspidy, line 1: Can you elaborate a bit on the changes here?

If I understand correctly, as soon as `--testsuite` is provided (regardless with which value), we load _all_ plugins?

mukeshpanchal27: In https://github.com/WordPress/performance/pull/972/commits/bc2884fc51cb81a577d869f567fff2d2eca68e3a i adjust the code so that it will load specific plugin only.

plugins.json
```diff
@@ -1,10 +1,13 @@
...
+    "images/webp-uploads": {
+      "slug": "webp-uploads",
+      "version": "1.0.5"
+    }
   },
-  "images/webp-uploads": {
-    "slug": "webp-uploads",
-    "version": "1.0.5"
-  }
+  "plugins": []
```
swissspidy: auto-sizes does not need to be in this?

swissspidy (approved):

mukeshpanchal27:
@joemcgill @westonruter Could you please take another look when you have a moment?

.github/workflows/php-test-plugins.yml
```diff
@@ -0,0 +1,78 @@
...
+      - 'package-lock.json'
+      - 'phpunit.xml.dist'
+      - 'tests/multisite.xml'
+      - 'composer.json'
+      - 'composer.lock'
+  pull_request:
+    branches:
+      - trunk
+      - 'release/**'
+      - 'feature/**'
```
westonruter, line 23: Why only these branches?

.github/workflows/php-test-plugins.yml
```diff
@@ -0,0 +1,78 @@
...
+        php: ['8.2', '8.1', '8.0', '7.4', '7.3', '7.2', '7.1', '7.0']
+        wp: [ 'latest' ]
+        include:
+          - php: '7.4'
+            wp: '6.3'
+          - php: '8.3'
+            wp: 'trunk'
+    env:
+      WP_ENV_PHP_VERSION: ${{ matrix.php }}
+      WP_ENV_CORE: ${{ matrix.wp == 'trunk' && 'WordPress/WordPress' || format( 'https://wordpress.org/wordpress-{0}.zip', matrix.wp ) }}
```
westonruter, line 57: Question: Why not use ternary expression here? Is it allowed?

```suggestion
      WP_ENV_CORE: ${{ matrix.wp == 'trunk' ? 'WordPress/WordPress' : format( 'https://wordpress.org/wordpress-{0}.zip', matrix.wp ) }}
```

composer.json
```diff
@@ -25,8 +25,8 @@
     "phpstan": "build-cs/vendor/bin/phpstan analyse --memory-limit=2048M -c phpstan.neon.dist",
     "format": "build-cs/vendor/bin/phpcbf --standard=phpcs.xml.dist --report-summary --report-source",
     "lint": "build-cs/vendor/bin/phpcs --standard=phpcs.xml.dist",
-    "test": "phpunit -c phpunit.xml.dist --verbose",
-    "test-multisite": "phpunit -c tests/multisite.xml --verbose"
+    "test": "phpunit -c phpunit.xml.dist --verbose --testsuite performance-lab",
+    "test-multisite": "phpunit -c tests/multisite.xml --verbose --testsuite performance-lab"
```
westonruter, line 29: How would the test suites for the plugins be run locally? Is the `--testsuite` really needed here? Shouldn't it rather be supplied when calling `npm run test-php`?

tests/bootstrap.php
```diff
@@ -32,28 +32,69 @@
...
-		require_once TESTS_PLUGIN_DIR . '/admin/server-timing.php';
-		require_once TESTS_PLUGIN_DIR . '/admin/plugins.php';
-		$module_files = glob( TESTS_PLUGIN_DIR . '/modules/*/*/load.php' );
-		if ( $module_files ) {
-			foreach ( $module_files as $module_file ) {
-				require_once $module_file;
-			}
+
+// Check if we use the plugin's test suite. If so, disable the PL plugin and only load the requested plugin.
+$testsuite_count = array_count_values( $_SERVER['argv'] )['--testsuite'];
```
westonruter: Question: Why `$_SERVER['argv']` and not `$argv`?

tests/bootstrap.php
```diff
@@ -32,28 +32,69 @@
...
+					// Check if plugin has a "plugin/plugin.php" file.
+					if ( file_exists( $plugin_test_path . '/' . $plugin_name . '.php' ) ) {
+						require_once $plugin_test_path . '/' . $plugin_name . '.php';
+						return;
+					}
+
+					// Check if plugin has a "plugin/load.php" file.
+					if ( file_exists( $plugin_test_path . '/load.php' ) ) {
+						require_once $plugin_test_path . '/load.php';
+					}
```
westonruter: Shouldn't this emit an error it if can't find the bootstrap file?

```suggestion
					if ( file_exists( $plugin_test_path . '/' . $plugin_name . '.php' ) ) {
						require_once $plugin_test_path . '/' . $plugin_name . '.php';
					} else if ( file_exists( $plugin_test_path . '/load.php' ) ) {
						// Check if plugin has a "plugin/load.php" file.
						require_once $plugin_test_path . '/load.php';
					} else {
						echo "Unable to locate standalone plugin bootstrap file in $plugin_test_path.";
						exit( 1 );
					}
```

tests/bootstrap.php
```diff
@@ -32,28 +32,69 @@
...
+
+	$plugin_name = '';
+	foreach ( $_SERVER['argv'] as $index => $arg ) {
+		if (
+			'--testsuite' === $arg &&
+			isset( $_SERVER['argv'][ $index + 1 ] ) &&
+			'performance-lab' !== $_SERVER['argv'][ $index + 1 ]
+		) {
+			$plugin_name = $_SERVER['argv'][ $index + 1 ];
+			break;
```
westonruter: I thought that multiple `--testsuite` parameters could be supplied, but it looks like only the last one is respected. That being the case, I think the `break` should be removed here in case multiple `--testsuite` args are provided.

```suggestion
```

tests/bootstrap.php
```diff
@@ -32,28 +32,69 @@
...
+// Check if we use the plugin's test suite. If so, disable the PL plugin and only load the requested plugin.
+$testsuite_count = array_count_values( $_SERVER['argv'] )['--testsuite'];
+if ( $testsuite_count > 1 ) {
+
+	$plugin_name = '';
+	foreach ( $_SERVER['argv'] as $index => $arg ) {
+		if (
+			'--testsuite' === $arg &&
+			isset( $_SERVER['argv'][ $index + 1 ] ) &&
+			'performance-lab' !== $_SERVER['argv'][ $index + 1 ]
```
westonruter: This could check here if the provided test suite is a valid plugin name.

```suggestion
			'performance-lab' !== $_SERVER['argv'][ $index + 1 ] &&
			file_exists( TESTS_PLUGIN_DIR . '/plugins/' . $_SERVER['argv'][ $index + 1 ] )
```

tests/bootstrap.php
```diff
@@ -32,28 +32,69 @@
...
+		) {
+			$plugin_name = $_SERVER['argv'][ $index + 1 ];
+			break;
+		}
+	}
+
+	if ( $plugin_name ) {
+		$plugin_test_path = TESTS_PLUGIN_DIR . '/plugins/' . $plugin_name;
+
+		if ( file_exists( $plugin_test_path ) ) {
```
westonruter: This could be simplified if the check is added to the condition above.

```suggestion
	if ( $plugin_name ) {
```

Naturally also to remove the `}` further below.

composer.json
westonruter: I guess this would be solved just by adding another `--testsuite` arg to the end so that it overrides whatever is set here initially? So if I wanted to run `auto-sizes` tests locally, I could just do:

```sh
npm run test-php -- -- --  --testsuite auto-sizes
```

This would result in somewhat unexpected output:

```
> phpunit -c phpunit.xml.dist --verbose --testsuite performance-lab '--testsuite' 'auto-sizes'
```

But it would successfully only run the `auto-sizes` test suite (as long as the `break` is removed as I suggested below).

westonruter (changes requested):

.github/workflows/php-test-plugins.yml
mukeshpanchal27: The majority of workflows run on these branches, so I have added them to the list. I believe these three branches are what we expect to be created for any work against any feature. If there's anything else you'd like to add that makes sense, I'm happy to include it.

.github/workflows/php-test-plugins.yml
mukeshpanchal27: This was added as part of #544. If we need to update this, then we also have to update similar settings for other workflows (php-test.yml and php-test-standalone-plugins.yml).

Shall we open a follow-up PR to make these changes?

.github/workflows/php-test-plugins.yml
westonruter: I think this workflow (and linting, for example) should run on all branches. Otherwise, when there is a sub-PR it won't get checks run. For example, I see now that checks are not getting run for these two sub-PRs I opened today: https://github.com/WordPress/performance/pull/984 and https://github.com/WordPress/performance/pull/983.

.github/workflows/php-test-plugins.yml
westonruter: Follow-up PR is fine. Just seems hard to read right now.

.github/workflows/php-test-plugins.yml
westonruter: This can be done in a follow-up PR.

composer.json
mukeshpanchal27: Do we need to make any updates based on this feedback?

westonruter: I think this is resolved, yes.

tests/bootstrap.php
```diff
@@ -32,28 +32,66 @@
...
-		require_once TESTS_PLUGIN_DIR . '/admin/server-timing.php';
-		require_once TESTS_PLUGIN_DIR . '/admin/plugins.php';
-		$module_files = glob( TESTS_PLUGIN_DIR . '/modules/*/*/load.php' );
-		if ( $module_files ) {
-			foreach ( $module_files as $module_file ) {
-				require_once $module_file;
-			}
+
+// Check if we use the plugin's test suite. If so, disable the PL plugin and only load the requested plugin.
+$testsuite_count = array_count_values( $_SERVER['argv'] )['--testsuite'];
```
westonruter: What if someone invokes `phpunit` without providing a `--testsuite` arg? Won't this cause an error due to the array index not being defined. I don't think this check is needed since it is looping over the args below anyway. Also, shouldn't passing `--testsuite performance-lab` behave the same as if no `--testsuite` was provided at all? This doesn't seem to be happening right now, as the `else` condition won't be run. As I understand, the structure should rather be as follows:

```php
// Check if we use the plugin's test suite. If so, disable the PL plugin and only load the requested plugin.
$plugin_name = '';
foreach ( $_SERVER['argv'] as $index => $arg ) {
	if (
		'--testsuite' === $arg &&
		isset( $_SERVER['argv'][ $index + 1 ] ) &&
		'performance-lab' !== $_SERVER['argv'][ $index + 1 ] &&
		file_exists( TESTS_PLUGIN_DIR . '/plugins/' . $_SERVER['argv'][ $index + 1 ] )
	) {
		$plugin_name = $_SERVER['argv'][ $index + 1 ];
	}
}

if ( $plugin_name ) {
	$plugin_test_path = TESTS_PLUGIN_DIR . '/plugins/' . $plugin_name;
	tests_add_filter(
		'plugins_loaded',
		static function () use ( $plugin_test_path, $plugin_name ) {
			// Check if plugin has a "plugin/plugin.php" file.
			if ( file_exists( $plugin_test_path . '/' . $plugin_name . '.php' ) ) {
				require_once $plugin_test_path . '/' . $plugin_name . '.php';
			} elseif ( file_exists( $plugin_test_path . '/load.php' ) ) {
				// Check if plugin has a "plugin/load.php" file.
				require_once $plugin_test_path . '/load.php';
			} else {
				echo "Unable to locate standalone plugin bootstrap file in $plugin_test_path.";
				exit( 1 );
			}
		},
		1
	);
} else {
	// Force plugin to be active.
	$GLOBALS['wp_tests_options'] = array(
		'active_plugins' => array( basename( TESTS_PLUGIN_DIR ) . '/load.php' ),
	);

	// Add filter to ensure the plugin's admin integration and all modules are loaded for tests.
	tests_add_filter(
		'plugins_loaded',
		static function () {
			require_once TESTS_PLUGIN_DIR . '/admin/load.php';
			require_once TESTS_PLUGIN_DIR . '/admin/server-timing.php';
			require_once TESTS_PLUGIN_DIR . '/admin/plugins.php';
			$module_files = glob( TESTS_PLUGIN_DIR . '/modules/*/*/load.php' );
			if ( $module_files ) {
				foreach ( $module_files as $module_file ) {
					require_once $module_file;
				}
			}
		},
		1
	);
}
```

mukeshpanchal27: Thanks. Updated in https://github.com/WordPress/performance/pull/972/commits/7f3b13b321ea1608412d8c4a83c7c73286ebc2fe

tests/bootstrap.php
westonruter: Maybe `$_SERVER['argv']` is actually safer because `$argv` can actually be disabled by default when `register_argc_argv` is off.

westonruter (approved):

phpcs.xml.dist
mukeshpanchal27: Open a separate issue for further work: https://github.com/WordPress/performance/issues/997

.github/workflows/php-test-plugins.yml
mukeshpanchal27: Open a separate issue for further work: #998

.github/workflows/php-test-plugins.yml
mukeshpanchal27: @westonruter, GitHub workflow does not allow `${{ x ? 'ifTrue' : 'ifFalse' }}` ternary operator. Instead, it uses `${{ x && 'ifTrue' || 'ifFalse' }}`. Therefore, we don't need to open a follow-up PR for this work.

PR #974, opened 2024-02-07
Remove admin bar skipping logic in favor of varying URL metrics by user being logged-in

westonruter:
The original aim was for the XPaths gathered for URL metrics when a user is logged-in to be re-used for optimizing responses for logged-out users, and vice-versa. In order to accomplish this, it was necessary to skip over the admin bar element so that the indices in the XPaths are not affected by its presence. Nevertheless, I found that this is not sufficient as there are other elements which may be conditionally printed at `wp_body_open` when the user is logged-in, namely `wp_customize_support_script()` (for example). 

Therefore, I've decided to no longer try skipping over the admin bar when calculating XPaths. Instead, the URL Metrics are now varied by whether or not the user is logged-in. This should be more robust to properly optimize pages that have additional elements that appear for logged-in users on e-commerce and membership sites. 

Lastly, this PR opts to disable optimization entirely when the user is an administrator (whether they can `customize` which is the cap check used to print `wp_customize_support_script()`). When a user is an administrator it is likely that there may be additional elements added inline on the page (e.g. edit post links or edit comment links) which would cause the XPaths captured in URL Metrics to not apply properly to non-admin users. Additionally, the purpose of the optimization is to optimize the experience of the majority of visitors to the site, and not to optimize the loading experience of admin users (who likely already have the assets cached anyway).

## Checklist

### DIFF
--- a/modules/images/image-loading-optimization/storage/data.php
+++ b/modules/images/image-loading-optimization/storage/data.php
@@ -57,7 +57,7 @@ function ilo_get_normalized_query_vars(): array {
 		);
 	}
 
-	// Vary URL metrics by whether the user is logged-in since additional elements may be present.
+	// Vary URL metrics by whether the user is logged in since additional elements may be present.
 	if ( is_user_logged_in() ) {
 		$normalized_query_vars['user_logged_in'] = true;
 	}
--- a/modules/images/image-loading-optimization/storage/data.php
+++ b/modules/images/image-loading-optimization/storage/data.php
@@ -57,6 +57,11 @@ function ilo_get_normalized_query_vars(): array {
 		);
 	}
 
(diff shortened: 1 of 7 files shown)


modules/images/image-loading-optimization/storage/data.php
```diff
@@ -57,6 +57,11 @@ function ilo_get_normalized_query_vars(): array {
 		);
 	}
 
+	// Vary URL metrics by whether the user is logged-in since additional elements may be present.
```
adamsilverstein:
```suggestion
	// Vary URL metrics by whether the user is logged in since additional elements may be present.
```

adamsilverstein (approved):
Nice! Left one doc block comment.

PR #976, opened 2024-02-07
Prevent optimizing images with data: URLs or no srcs

westonruter:
## Summary

The optimization logic was already accounting for skipping background images that use a data: URL but it didn't account for regular `img` elements with a data: URL for their `src`. Such elements are irrelevant for optimizing the loading since they are embedded in the document.

This also makes sure that `img` elements with an empty `src` are also omitted from processing.

## Checklist


joemcgill (approved):
Looks good to me

PR #977, opened 2024-02-07
Fix optimizing pages with stale URL metrics

westonruter:
## Summary

When testing I sometimes discovered an unexpected warning in the console:

> ⚠️ `<link rel=preload>` has an invalid `href` value

It turns out that this is closely related to #974. What was happening is that URL Metrics had been recorded for a page in which the admin bar was present, but then I opened the same URL in an incognito window where the admin bar was not present. At this point, the XPaths in the stored URL Metrics are invalid because the indices are offset by the absence of the admin bar. These LCP elements from the stale URL Metrics were being passed into `ilo_construct_preload_links()` without the the `img_attributes` or `background_image` which are obtained from the document while iterating over the tags. So then it would construct a preload `link` tag without those required attributes.

## Relevant technical choices

To rectify this, we need to keep track of the LCP elements from the stored URL Metrics which are actually discovered while iterating over the document. If any is absent, then it must be set to `false` as if there had been no LCP image detected at all. This will in turn cause `ilo_construct_preload_links()` to skip over it. Additionally, `ilo_construct_preload_links()` is amended to also check for attempting to construct a preload `link` without an `href` (or `imagesrcset`) attribute, in which case it emits a `_doing_it_wrong()` error.

## Checklist

### DIFF
--- a/modules/images/image-loading-optimization/optimization.php
+++ b/modules/images/image-loading-optimization/optimization.php
@@ -101,6 +101,13 @@ function ilo_construct_preload_links( array $lcp_elements_by_minimum_viewport_wi
 
 		// Skip constructing a link if it is missing required attributes.
 		if ( empty( $link_attributes['href'] ) && empty( $link_attributes['imagesrcset'] ) ) {
+			_doing_it_wrong(
+				__FUNCTION__,
+				esc_html(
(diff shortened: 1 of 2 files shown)


adamsilverstein (approved):
Splendid!

PR #978, opened 2024-02-09
Wait until idle before starting detection

### LINKED ISSUE #894: Guard against detection logic from negatively impacting INP
## Feature Description

As discussed in https://github.com/WordPress/performance/pull/878#discussion_r1387235016, the Image Loading Optimization module (see #869) has detection logic to discover the LCP image and other images which are in the initial viewport. This logic is currently running at the `load` event after which it immediately sends the data to the REST API endpoint:

https://github.com/WordPress/performance/blob/03f32520241e666b3a1234b181f5f3459425fe46/modules/images/image-loading-optimization/detection/detect.js#L342-L433

However, given that the metrics collection is a low priority task which is intended to improve the user experience, we must guard against it from possibly impacting a visitor in a negative way when URL metrics need to gathered. For this reason there is the following todo:

https://github.com/WordPress/performance/blob/03f32520241e666b3a1234b181f5f3459425fe46/modules/images/image-loading-optimization/detection/detect.js#L406

We should leverage `requestIdleCallback()` and/or `scheduler.yield()` (or else `setTimeout`)--whatever is available--to make sure that the detection task is broken up into small tasks that do not contribute to a long task.

westonruter:
## Summary

Fixes #894

This rearranges the detection logic to defer doing any DOM operations until after the page has fully loaded _and_ the CPU is idle. Additionally, it now yields to main before initiating the fetch request to submit the data to the server, additionally breaking up tasks.

## Performance Profiles

With 6X CPU throttling, the plugin active with the change applied shows a marginal improvement where tasks are further broken up.

### [Trace without ILO](https://trace.cafe/t/MQaZChmr2W)

Screenshot 2024-02-09 10 55 19

### [Trace with ILO before change](https://trace.cafe/t/yPGKS2Q4jq)

Screenshot 2024-02-09 10 56 49

### [Trace with ILO after change](https://trace.cafe/t/ueXDnq73Dl)

Screenshot 2024-02-09 10 57 57

## Checklist

### DIFF
--- a/modules/images/image-loading-optimization/detection/detect.js
+++ b/modules/images/image-loading-optimization/detection/detect.js
@@ -200,7 +200,7 @@ export default async function detect( {
 
 	// As an alternative to this, the ilo_print_detection_script() function can short-circuit if the
 	// ilo_is_url_metric_storage_locked() function returns true. However, the downside with that is page caching could
-	// result in metrics being missed being gathered when a user navigates around a site and primes the page cache.
+	// result in metrics missed from being gathered when a user navigates around a site and primes the page cache.
 	if ( isStorageLocked( currentTime, storageLockTTL ) ) {
 		if ( isDebug ) {
 			warn( 'Aborted detection due to storage being locked.' );
--- a/modules/images/image-loading-optimization/detection/detect.js
+++ b/modules/images/image-loading-optimization/detection/detect.js
@@ -155,16 +155,6 @@ export default async function detect( {
 } ) {
 	const currentTime = getCurrentTime();
 
-	// As an alternative to this, the ilo_print_detection_script() function can short-circuit if the
-	// ilo_is_url_metric_storage_locked() function returns true. However, the downside with that is page caching could
-	// result in metrics being missed being gathered when a user navigates around a site and primes the page cache.
-	if ( isStorageLocked( currentTime, storageLockTTL ) ) {
-		if ( isDebug ) {
-			warn( 'Aborted detection due to storage being locked.' );
-		}
-		return;
-	}
-
 	// Abort running detection logic if it was served in a cached page.
 	if ( currentTime - serveTime > detectionTimeWindow ) {
 		if ( isDebug ) {
@@ -206,6 +196,16 @@ export default async function detect( {
 		await new Promise( ( resolve ) => {
 			requestIdleCallback( resolve );
 		} );
+	}
+
+	// As an alternative to this, the ilo_print_detection_script() function can short-circuit if the
+	// ilo_is_url_metric_storage_locked() function returns true. However, the downside with that is page caching could
+	// result in metrics being missed being gathered when a user navigates around a site and primes the page cache.
+	if ( isStorageLocked( currentTime, storageLockTTL ) ) {
+		if ( isDebug ) {
+			warn( 'Aborted detection due to storage being locked.' );
+		}
+		return;
 	}
 
 	// Prevent detection when page is not scrolled to the initial viewport.
--- a/modules/images/image-loading-optimization/detection/detect.js
+++ b/modules/images/image-loading-optimization/detection/detect.js
@@ -175,8 +175,40 @@ export default async function detect( {
 		return;
 	}
 
+	// Abort if the current viewport is not among those which need URL metrics.
+	if ( ! isViewportNeeded( win.innerWidth, neededMinimumViewportWidths ) ) {
+		if ( isDebug ) {
+			log( 'No need for URL metrics from the current viewport.' );
+		}
+		return;
+	}
+
+	// Ensure the DOM is loaded (although it surely already is since we're executing in a module).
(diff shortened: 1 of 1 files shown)


adamsilverstein:
> Additionally, it now yields to main before initiating the fetch request to submit the data to the server, additionally breaking up tasks.

I wonder if you could avoid sending here entirely with a Beacon API [approach](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon)?

modules/images/image-loading-optimization/detection/detect.js
```diff
@@ -175,20 +165,55 @@ export default async function detect( {
...
+	// Wait yet further until idle.
+	if ( typeof requestIdleCallback === 'function' ) {
+		await new Promise( ( resolve ) => {
+			requestIdleCallback( resolve );
+		} );
+	}
+
+	// As an alternative to this, the ilo_print_detection_script() function can short-circuit if the
+	// ilo_is_url_metric_storage_locked() function returns true. However, the downside with that is page caching could
+	// result in metrics being missed being gathered when a user navigates around a site and primes the page cache.
```
adamsilverstein: this line "result in metrics being missed being gathered" is confusing, can you rephrase?

modules/images/image-loading-optimization/detection/detect.js
```diff
@@ -175,20 +165,55 @@ export default async function detect( {
...
+	if ( isStorageLocked( currentTime, storageLockTTL ) ) {
 		if ( isDebug ) {
-			log( 'No need for URL metrics from the current viewport.' );
+			warn( 'Aborted detection due to storage being locked.' );
+		}
+		return;
+	}
+
+	// Prevent detection when page is not scrolled to the initial viewport.
```
adamsilverstein, line 211: this makes me wonder if on the server side we should potentially avoid using the captured data when the link is apparently to a non scrolled to top position - ie. when it contains an anchor hash. Do you do that already or what do you think?

modules/images/image-loading-optimization/detection/detect.js
westonruter:
```suggestion
	// result in metrics missed from being gathered when a user navigates around a site and primes the page cache.
```

How's this?

adamsilverstein (commented):
Looks good! Couple of small questions.

modules/images/image-loading-optimization/detection/detect.js
adamsilverstein: much better, thanks

adamsilverstein (approved):
Superb!

modules/images/image-loading-optimization/detection/detect.js
westonruter: That effectively is what this is doing, yes. 

As for server-side detection of this, the URL fragment is not sent along in the HTTP request for the page, so the server won't be able to see it. (Which is why annoyingly URL fragments can be lost when having to authenticate.) It would have to be checked client-side. And often a URL fragment may not refer to an element on the page so that would require checking the DOM to see if it exists. But we only care in the end if the user is not in the first viewport, so that's what checking the scroll position does, I think in a more reliable way.

westonruter:
> > Additionally, it now yields to main before initiating the fetch request to submit the data to the server, additionally breaking up tasks.
> 
> I wonder if you could avoid sending here entirely with a Beacon API [approach](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon)?

@adamsilverstein That's a great question! I don't have a good reason why this shouldn't be used instead, other than initially it didn't seem it could be used to send a `Content-Type` header, but I can see [this is not the case](https://stackoverflow.com/a/41729668/93579). I'm going to open that in a separate PR.

westonruter:
> I don't have a good reason why this shouldn't be used instead, other than initially it didn't seem it could be used to send a `Content-Type` header, but I can see [this is not the case](https://stackoverflow.com/a/41729668/93579). I'm going to open that in a separate PR.

@adamsilverstein Oh, I remember the more important reason. It is setting a client-side lock to prevent attempting to send more URL metrics when the REST API response is successful:

https://github.com/WordPress/performance/blob/7761a9b3a029c3ae545cff3a3c46ee70cc00a071/modules/images/image-loading-optimization/detection/detect.js#L363-L374

This isn't possible with the Beacon API because the response is not available. In any case, I did work out how to replace the [`fetch` code](https://github.com/WordPress/performance/blob/7761a9b3a029c3ae545cff3a3c46ee70cc00a071/modules/images/image-loading-optimization/detection/detect.js#L362-L388) with the Beacon API if that didn't end up being needed:

```js
	const url = new URL( restApiEndpoint );
	url.searchParams.append( '_wpnonce', restApiNonce );
	navigator.sendBeacon(
		url,
		new Blob( [ JSON.stringify( urlMetrics ) ], {
			type: 'application/json',
		} )
	);
```

Issue #973, opened 2024-02-06
Optimize workflows; Add Dependabot and CodeQL


thelovekesh:
Currently, there are a few areas where workflows can be optimized to reduce CI times, maintenance, and security in the runners. Also, consider adding tools like @dependabot that can keep the dependency up-to-date unless some dependency requires it, for example: `chalk` since it's pure ESM now and the plugin CLI is CJS. 

## Tasks

- [ ] Optimize workflows for lower CI times and reduce files for maintenance(i.e. lint workflows can remain in one file).
- [ ] Remove dependency from third-party workflows for tasks that can be handled natively.
- [ ] Update permissions in the workflows to read-only and manually provide any other permission at the job level.
- [ ] Update **wp-env** based PHPUnit setup with MySQL + SVN which is easy to set up for any version and has very less or no compatibility maintenance.
- [x] Add @dependabot config to update `composer`, `npm`, and `github-actions` at least once a month. 
- [x] Add CodeQL for JS code analysis for any security vulnerabilities.

westonruter:
I presume you'll be opening separate PRs for each tasks for review?

thelovekesh:
Yes, that would be like:
- First three tasks in one PR since they are optimization changes.
- Fourth task in standalone PR since it will refactor PHPUnit setup.
- Last two can be included in single PR since their config files are plain and simple.

swissspidy:
Some quick thoughts on https://github.com/thelovekesh/performance/pull/2 after taking another look:

* `.github/release.yml` is not needed. We don't currently use GitHub's auto-generated release notes feature.
* The whole `determine-file-counts` logic seems a bit overkill and adds complexity. Do we really need that?
* What's the cache busting for. Never had to use something like that in a project before. Why is it needed here?

thelovekesh:
Thanks, @swissspidy for taking a look. Here are my few observations:

> .github/release.yml is not needed. We don't currently use GitHub's auto-generated release notes feature.

Since we are going to add @dependabot && if we ever require to use the GitHub notes release feature, this config can come in handy. We can skip adding it, but I don't see any problem in keeping it too.

> The whole determine-file-counts logic seems a bit overkill and adds complexity. Do we really need that?

[...]

PR #981, opened 2024-02-10
Add codeql analysis workflow for JS

thelovekesh:
## Summary
Add CodeQL analysis for JS language.

Fixes task 6 of #973

## Relevant technical choices

- Add CodeQL official workflows.
- Add concurrency to cancel any in-progress workflow runs.
- Allow only `security:write` permissions in workflow.

## Checklist


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  push:
+    # Only run if JS files changed.
+    paths:
+      - "**.js"
+    branches:
+      - trunk
+      - "release/**"
+  pull_request:
+    # Only run if JS or Python files changed.
```
westonruter:
```suggestion
    # Only run if JS files changed.
```

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,48 @@
...
+      - trunk
+      - "release/**"
+  pull_request:
+    # Only run if JS files changed.
+    paths:
+      - "**.js"
+    branches:
+      - trunk
+      - "release/**"
+      - "feature/**"
```
westonruter: I think this should apply to branches for all PRs. 

```suggestion
```

But it can be done in another PR along with the other workflows. See https://github.com/WordPress/performance/pull/972#discussion_r1487151134

westonruter (approved):

.github/workflows/codeql-analysis.yml
thelovekesh: Fixed in https://github.com/WordPress/performance/pull/981/commits/f4e103c9fda1f5c92e9695a25910a333a3cdf859

thelovekesh:
@joemcgill @felixarntz can you folks please take a look?

.github/CODEOWNERS
```diff
@@ -2,7 +2,7 @@
 /admin                                      @mukeshpanchal27 @felixarntz
 /tests/admin                                @mukeshpanchal27 @felixarntz
 /bin                                        @mukeshpanchal27 @felixarntz
-/.github                                    @mukeshpanchal27 @felixarntz
+/.github                                    @mukeshpanchal27 @felixarntz @thelovekesh
```
joemcgill, line 5: Not opposed, but am wondering if we should specify code owners specifically for workflows, rather than the whole directory?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,44 @@
...
+# Cancel previous workflow run groups that have not completed.
+concurrency:
+  # Group workflow runs by workflow name, along with the head branch ref of the pull request
+  # or otherwise the branch or tag ref.
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+# Disable permissions for all available scopes.
+# Enable permissions for specific scopes as needed on job level.
+permissions: {}
```
joemcgill, line 13: I'm assuming these are copied from the props-bot workflow? Seems sensible to me, just noting that most of our workflows don't include concurrency groups or a removal of permissions scopes. I assume the concurrency group here is meant to cancel any in-progress jobs on PRs whenever a new commit is pushed to that branch or if the current workflow is restarted for some reason?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,44 @@
...
+    # Only run if JS files changed.
+    paths:
+      - '**.js'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
```
joemcgill, line 33: Why is this needed?

joemcgill (commented):
Thanks @thelovekesh. I've left a couple of comments, but don't see any blockers.

.github/CODEOWNERS
thelovekesh: @joemcgill Not sure what would be more appropriate but I am okay to review any workflow file changes.

.github/workflows/codeql-analysis.yml
thelovekesh: @joemcgill Most of the workflows rely on a third-party action to control the concurrency which is okay but I am not in favor of depending on third-party dependencies while something can be achieved natively, and yes it's here for cancelling the previous workflow run.

Also restricting the permission in the runner is a good practice to keep the metadata of a repo intact in case some action is compromised.

thelovekesh: Referenced from: https://github.com/ampproject/amp-wp/blob/develop/.github/workflows/build-test-measure.yml

.github/CODEOWNERS
thelovekesh: I think this would be more appropriate as I will be working on optimizing most of the workflows under https://github.com/WordPress/performance/issues/973

.github/workflows/codeql-analysis.yml
thelovekesh: Since CodeQL requires permission to write to security events as per https://github.com/github/codeql-action?tab=readme-ov-file#permissions

More info on how it's add to security events - https://github.com/WordPress/performance/security/code-scanning?query=pr%3A981+is%3Aopen

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,44 @@
...
+
+# Disable permissions for all available scopes.
+# Enable permissions for specific scopes as needed on job level.
+permissions: {}
+
+on:
+  push:
+    # Only run if JS files changed.
+    paths:
+      - '**.js'
```
felixarntz, line 19: Does CodeQL only cover first-party JS? What about dependencies? Just curious, I'm not familiar with CodeQL.

I'm only asking since, if it also covers dependencies, we may also want to include `package.json` and `package-lock.json` here.

.github/workflows/codeql-analysis.yml
felixarntz: @thelovekesh Your rationale makes sense to me, however I would be in favor of making such decisions consistently. I think for this PR we should stick with how we have been handling concurrency and permissions so far, as changing that is not the purpose of the PR. And then we can open a separate issue and/or PR to consider changing it holistically.

There is no good reason why this workflow should behave differently from the other ones in those points.

.github/workflows/codeql-analysis.yml
thelovekesh: @felixarntz For dependencies, @dependabot creates security alerts that can be configured from -  https://github.com/WordPress/performance/security/dependabot

.github/workflows/codeql-analysis.yml
thelovekesh: @felixarntz  TBH, I'm not sure if we should be prevented from writing optimized workflows by default just because existing workflows don't include those optimizations. In any case, as part of https://github.com/WordPress/performance/issues/973, I will be updating them.

swissspidy (approved):

PR #982, opened 2024-02-12
Add failure checks to Embed Optimizer

westonruter:
## Summary

This adds some sanity checks for the HTML Tag Processor to catch a rare case where it would fail to set a bookmark or seek to a bookmark. (I was experiencing such a failure but it seemed due to running a broken alpha state of 6.5.)

It also tidies up the tests by:

* fixing the method name for the data provider
* add some PHP typing
* add `@covers`

## Relevant technical choices

A wrapper for `wp_trigger_error()` is used since it was introduced in WP 6.4.

## Checklist

- [x] PR has either `[Focus]` or `Infrastructure` label.
- [x] PR has a `[Type]` label.
- [x] PR has a milestone or the `no milestone` label.

PR #979, opened 2024-02-09
Add tests for Optimize Embeds

adamsilverstein:
## Summary

Adds some tests for Optimize Embeds; also complete rename. Using Optimize Embeds because this could broaden scope beyond just "lazy".

Follow up to https://github.com/WordPress/performance/pull/886/files

## Checklist

- [x] PR has either `[Focus]` or `Infrastructure` label.
- [x] PR has a `[Type]` label.
- [x] PR has a milestone or the `no milestone` label.

PR #983, opened 2024-02-12
Use custom script type for lazy-loaded scripts

westonruter:
## Summary

~_This is a sub-PR of #982. Review/merge it first._~

This implements the following https://github.com/WordPress/performance/pull/979#discussion_r1484780298:

> What if instead of converting `script[src]` to `script[data-lazy-embed-src]`, it instead set the `type` attribute of `text/x.lazy-loaded-javascript`? Clearly out of scope for this PR, but just thought of this alternative. It probably doesn't make any difference, oh except actually wouldn't it be better in that it would prevent the browser from treating this `script` as an inline script? While the browser hopefully is smart enough to skip parsing an inline script that has no contents, in the worst case it could. That would be bad because inline scripts are blocking. If the `src` is left-as is and the `type` is used to mark such lazy-loaded scripts instead, this would be avoided. This is also what Partytown does to mark its scripts via `text/partytown`.

Another important reason so do this change is that plugins that concatenate scripts may erroneously attempt to concatenate the empty `script[data-lazy-embed-src]` scripts. By using a custom type, this will be avoided.

I also just checked with WP Rocket, and [they also use a custom type](https://github.com/wp-media/wp-rocket/blob/704f03428f286a9a84e32378626848444986534a/inc/Engine/Optimization/DelayJS/HTML.php#L197-L240). The logic also looks like it would by default try to delay `script[data-lazy-embed-src]` scripts which would break Embed Optimizer. Also, the WP Rocket `Combine` logic specifically [skips over any `script` that has a `type` that isn't JavaScript](https://github.com/wp-media/wp-rocket/blob/704f03428f286a9a84e32378626848444986534a/inc/Engine/Optimization/Minify/JS/Combine.php#L239-L248). So it seems we'd be safer in both cases to use a custom `type`.

## Checklist


adamsilverstein:
Nice catch on the WP-Rocket usage, that makes a stronger argument for this approach. Good idea to ensure this is compatible with existing optimizers.

modules/js-and-css/embed-optimizer/hooks.php
```diff
@@ -63,8 +63,10 @@ function embed_optimizer_filter_oembed_html( string $html ): string {
 	if ( 1 === $script_count && ! $has_inline_script && $p->has_bookmark( 'script' ) ) {
 		add_action( 'wp_footer', 'embed_optimizer_lazy_load_scripts' );
 		if ( $p->seek( 'script' ) ) {
-			$p->set_attribute( 'data-lazy-embed-src', $p->get_attribute( 'src' ) );
-			$p->remove_attribute( 'src' );
+			if ( $p->get_attribute( 'type' ) ) {
+				$p->set_attribute( 'data-original-type', $p->get_attribute( 'type' ) );
+			}
+			$p->set_attribute( 'type', 'text/x.lazy-loaded-javascript' );
```
adamsilverstein: Should this match the plugin name? maybe something like `text/embed-optimizer`?

adamsilverstein (commented):
Looks good, left one suggestion for the type name

modules/js-and-css/embed-optimizer/hooks.php
westonruter: I tried to follow the rules outlined in https://en.wikipedia.org/wiki/Media_type

That said, what about `application/vnd.embed-optimizer.javascript`?

westonruter: Committed this in 39d44880

adamsilverstein (approved):
Fine!

Issue #969, opened 2024-02-05
Add plugin banner and icon assets for Speculation Rules and Auto Sizes


westonruter:
## Feature Description

As was done in other modules in #815, we need to add the banner and icon assets so that the plugin looks right in the directory.

westonruter:
PRs opened for new assets:

* https://github.com/WordPress/performance/pull/986
* https://github.com/WordPress/performance/pull/987

PR #987, opened 2024-02-14
Add plugin directory assets for Speculation Rules

westonruter:
See #969

I compressed the Figma exports with Squoosh and OMGSVG.

## Checklist


domenic:
I might be missing something, but these assets seem to say "SP"; should they say "SR" instead?

westonruter:
@domenic the convention for the plugins has been to use the first two letters and not the initials, though I agree it's not really clear. There has been some initial discussions about changing that convention. cc @eclarke1

swissspidy (approved):

Issue #933, opened 2024-01-11
Use class instances for common objects handled in image loading optimization logic


felixarntz:
Currently all logic is happening in functions, which in a few places makes the code complicated to follow. Specifically the handling of data arrays could be simplified, and potentially in a few places be made more efficient (e.g. by storing things in a class property rather than recalculating again). See [various](https://github.com/WordPress/performance/pull/898#discussion_r1446605520) [comments](https://github.com/WordPress/performance/pull/898#discussion_r1446603559) [on](https://github.com/WordPress/performance/pull/898#discussion_r1446600817) [unit test PR](https://github.com/WordPress/performance/pull/898).

Here's an initial proposal:
* Introduce class to represent a single URL metric (e.g. `ILO_URL_Metric`).
    * This would mostly be a value object with properties like `url`, `slug`, `viewport`, `elements`.
    * The class could potentially have private logic to take care of validation.
    * The `elements` entry may deserve some further thought, particularly around how it could work considering additional use-cases than LCP elements in the future.
* Introduce class to represent a grouped URL metrics collection by breakpoint (e.g. `ILO_Grouped_URL_Metric_Collection`).
    * The class could receive a list (indexed array) of `ILO_URL_Metric` instances and then take care of grouping them accordingly. The grouping internally could just be a map of `breakpoint => ILO_URL_Metric[]`.
    * The class should probably implement `Traversable`, so that it can be iterated just like an array (over the breakpoints).
    * The class could have a method `unshift( ILO_URL_Metric $metric )` that replaces the `ilo_unshift_url_metrics()` function and integrates the given metric properly, while discarding excess metrics beyond the configured sample size at the same time.
    * Some of the grouping "configuration" (e.g. breakpoints, sample size) may also be useful to provide via constructor.
* The `ilo_group_url_metrics_by_breakpoint()` function could then use these two classes.
    * It seems the module currently _always_ uses the breakpoints from `ilo_get_breakpoint_max_widths()` and the sample size from `ilo_get_url_metrics_breakpoint_sample_size()`. If so, we could probably avoid those function calls elsewhere and _only_ call them in `ilo_group_url_metrics_by_breakpoint()`. In other words, only the group class constructor would take them as parameters (for flexibility and testing), but the wrapper function to instantiate the group class wouldn't, it would just use the default ones, to avoid having to pass those parameters around in several places.
[...]

PR #988, opened 2024-02-15
Introduce ILO_URL_Metric to encapsulate core data structure

westonruter:
## Summary

This implements these aspects of #933:

> Introduce class to represent a single URL metric (e.g. `ILO_URL_Metric`).
> * This would mostly be a value object with properties like `url`, `slug`, `viewport`, `elements`.
> * The class could potentially have private logic to take care of validation.
> * The `elements` entry may deserve some further thought, particularly around how it could work considering additional use-cases than LCP elements in the future.

> * Functions that today return an array of URL metric associative arrays should going forward return an array of URL metric class instances (e.g. `ilo_parse_stored_url_metrics()`).

Subsequent points will be addressed in follow-up PRs.

## Relevant technical choices

* A new class called `ILO_URL_Metric` is used instead of passing around an array of the same data.
* The `ILO_URL_Metric` class contains the validation logic to ensure that the data is as expected.
* The REST API route re-uses the JSON schema defined in the `ILO_URL_Metric` class to validate requests.
* The `ILO_URL_Metric` class now requires that a `timestamp` be provided at construction time, which removes this concern from `ilo_unshift_url_metrics()`.
* Validation logic is applied to the URL metrics pulled from the `ilo_url_metrics` post type. This ensures that if the data format changes, any prior entries will just be omitted.
* Static analysis is further improved with typing and addressing various inspection issues flagged by PhpStorm.

## Checklist

### DIFF
--- a/modules/images/image-loading-optimization/class-ilo-url-metric.php
+++ b/modules/images/image-loading-optimization/class-ilo-url-metric.php
@@ -9,12 +9,10 @@
 /**
  * Representation of the measurements taken from a single client's visit to a specific URL.
  *
- * @implements ArrayAccess<string, mixed>
- *
  * @since n.e.x.t
  * @access private
  */
-final class ILO_URL_Metric implements ArrayAccess, JsonSerializable {
+final class ILO_URL_Metric implements JsonSerializable {
 
 	/**
 	 * Gets JSON schema for URL Metric.
@@ -103,11 +101,11 @@ final class ILO_URL_Metric implements ArrayAccess, JsonSerializable {
 	}
 
 	/**
-	 * Validated data.
+	 * Data.
 	 *
 	 * @var array
 	 */
-	private $data;
+	private $data = array();
 
 	/**
 	 * Constructor.
@@ -121,68 +119,37 @@ final class ILO_URL_Metric implements ArrayAccess, JsonSerializable {
 		if ( ! $validated ) {
 			$valid = rest_validate_object_value_from_schema( $data, self::get_json_schema(), self::class );
 			if ( is_wp_error( $valid ) ) {
-				throw new Exception( $valid->get_error_message() );
+				throw new Exception( esc_html( $valid->get_error_message() ) );
 			}
 		}
 		$this->data = $data;
 	}
 
+	/**
+	 * Gets viewport width.
+	 *
+	 * @return int
+	 */
 	public function get_viewport_width(): int {
 		return $this->data['viewport']['width'];
 	}
 
+	/**
+	 * Gets timestamp.
+	 *
+	 * @return float
+	 */
 	public function get_timestamp(): float {
 		return $this->data['timestamp'];
 	}
 
+	/**
+	 * Gets elements.
+	 *
+	 * @return array
+	 */
 	public function get_elements(): array {
 		return $this->data['elements'];
-	}
-
-	/**
-	 * Checks if the offset exists.
-	 *
-	 * @param string $offset Offset.
-	 * @return bool Whether exists.
-	 */
-	public function offsetExists( $offset ): bool {
-		return isset( $this->data[ $offset ] );
-	}
-
-	/**
-	 * Gets offset.
-	 *
-	 * @throws Exception If the offset does not exist.
-	 *
-	 * @param string $offset Offset.
-	 * @return mixed Value.
-	 */
-	public function offsetGet( $offset ) {
-		if ( ! $this->offsetExists( $offset ) ) {
-			throw new Exception( sprintf( __( 'Unknown property %s on ILO_URL_Metric.', 'performance-lab' ), $offset ) ); // phpcs:ignore WordPress.Security.EscapeOutput.ExceptionNotEscaped
-		}
-		return $this->data[ $offset ];
-	}
-
-	/**
-	 * Sets offset (disabled).
-	 *
-	 * @param string $offset Offset.
-	 * @param mixed  $value  Value.
-	 * @throws Exception
-	 */
-	public function offsetSet( $offset, $value ) {
-		throw new Exception( __( 'Cannot set properties on ILO_URL_Metric.', 'performance-lab' ) ); // phpcs:ignore WordPress.Security.EscapeOutput.ExceptionNotEscaped
-	}
-
-	/**
-	 * Unsets offset (disabled).
-	 *
-	 * @param string $offset Offset.
-	 * @throws Exception
-	 */
-	public function offsetUnset( $offset ) {
-		throw new Exception( __( 'Cannot unset properties on ILO_URL_Metric.', 'performance-lab' ) );
 	}
 
 	/**
--- a/modules/images/image-loading-optimization/class-ilo-url-metric.php
+++ b/modules/images/image-loading-optimization/class-ilo-url-metric.php
@@ -36,16 +36,13 @@ final class ILO_URL_Metric implements JsonSerializable {
 	 * Constructor.
 	 *
 	 * @param array $data      URL metric data.
-	 * @param bool  $validated Whether the data was already validated.
 	 *
-	 * @throws Exception When the input is invalid.
+	 * @throws ILO_Data_Validation_Exception When the input is invalid.
 	 */
-	public function __construct( array $data, bool $validated = false ) {
-		if ( ! $validated ) {
-			$valid = rest_validate_object_value_from_schema( $data, self::get_json_schema(), self::class );
-			if ( is_wp_error( $valid ) ) {
-				throw new Exception( esc_html( $valid->get_error_message() ) );
-			}
+	public function __construct( array $data ) {
+		$valid = rest_validate_object_value_from_schema( $data, self::get_json_schema(), self::class );
+		if ( is_wp_error( $valid ) ) {
+			throw new ILO_Data_Validation_Exception( esc_html( $valid->get_error_message() ) );
 		}
 		$this->data = $data;
 	}
--- a/modules/images/image-loading-optimization/class-ilo-url-metric.php
+++ b/modules/images/image-loading-optimization/class-ilo-url-metric.php
@@ -35,7 +35,7 @@ final class ILO_URL_Metric implements JsonSerializable {
 	/**
 	 * Constructor.
 	 *
-	 * @param array $data      URL metric data.
+	 * @param array $data URL metric data.
 	 *
 	 * @throws ILO_Data_Validation_Exception When the input is invalid.
 	 */
--- a/modules/images/image-loading-optimization/class-ilo-url-metric.php
+++ b/modules/images/image-loading-optimization/class-ilo-url-metric.php
@@ -53,7 +53,7 @@ final class ILO_URL_Metric implements JsonSerializable {
 	/**
 	 * Gets JSON schema for URL Metric.
 	 *
-	 * @return array
+	 * @return array Schema.
 	 */
 	public static function get_json_schema(): array {
 		$dom_rect_schema = array(
--- a/modules/images/image-loading-optimization/class-ilo-url-metric.php
+++ b/modules/images/image-loading-optimization/class-ilo-url-metric.php
@@ -38,10 +38,10 @@ final class ILO_URL_Metric implements ArrayAccess, JsonSerializable {
 		);
 
 		return array(
-			'$schema'    => 'http://json-schema.org/draft-04/schema#',
-			'title'      => 'ilo-url-metric',
-			'type'       => 'object',
-			'properties' => array(
+			'$schema'              => 'http://json-schema.org/draft-04/schema#',
+			'title'                => 'ilo-url-metric',
+			'type'                 => 'object',
+			'properties'           => array(
 				'viewport'  => array(
 					'description' => __( 'Viewport dimensions', 'performance-lab' ),
 					'type'        => 'object',
@@ -98,6 +98,7 @@ final class ILO_URL_Metric implements ArrayAccess, JsonSerializable {
 					),
 				),
 			),
+			'additionalProperties' => false,
 		);
 	}
 
@@ -113,12 +114,29 @@ final class ILO_URL_Metric implements ArrayAccess, JsonSerializable {
 	 *
 	 * @param array $data      URL metric data.
 	 * @param bool  $validated Whether the data was already validated.
+	 *
+	 * @throws Exception When the input is invalid.
 	 */
 	public function __construct( array $data, bool $validated = false ) {
 		if ( ! $validated ) {
-			// TODO: Validate.
+			$valid = rest_validate_object_value_from_schema( $data, self::get_json_schema(), self::class );
+			if ( is_wp_error( $valid ) ) {
+				throw new Exception( $valid->get_error_message() );
+			}
 		}
 		$this->data = $data;
+	}
+
+	public function get_viewport_width(): int {
+		return $this->data['viewport']['width'];
+	}
+
+	public function get_timestamp(): float {
+		return $this->data['timestamp'];
+	}
+
+	public function get_elements(): array {
+		return $this->data['elements'];
 	}
 
 	/**
--- a/modules/images/image-loading-optimization/class-ilo-url-metric.php
+++ b/modules/images/image-loading-optimization/class-ilo-url-metric.php
@@ -33,6 +33,24 @@ final class ILO_URL_Metric implements JsonSerializable {
 	private $data;
 
 	/**
+	 * Constructor.
+	 *
+	 * @param array $data      URL metric data.
+	 * @param bool  $validated Whether the data was already validated.
+	 *
+	 * @throws Exception When the input is invalid.
+	 */
+	public function __construct( array $data, bool $validated = false ) {
+		if ( ! $validated ) {
+			$valid = rest_validate_object_value_from_schema( $data, self::get_json_schema(), self::class );
+			if ( is_wp_error( $valid ) ) {
+				throw new Exception( esc_html( $valid->get_error_message() ) );
+			}
+		}
+		$this->data = $data;
+	}
+
+	/**
 	 * Gets JSON schema for URL Metric.
 	 *
 	 * @return array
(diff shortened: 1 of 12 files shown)


modules/images/image-loading-optimization/class-ilo-url-metric.php
```diff
@@ -0,0 +1,175 @@
...
+	 *              isLCP: bool,
+	 *              isLCPCandidate: bool,
+	 *              xpath: string,
+	 *              intersectionRatio: float,
+	 *              intersectionRect: array{ width: int, height: int },
+	 *              boundingClientRect: array{ width: int, height: int },
+	 *          }>
+	 *      }
+	 */
+	private $data;
```
thelovekesh: I think we can move the properties to the above?

westonruter: I'm not sure I understand.

thelovekesh: I mean keep the properties on the top in the class.

mukeshpanchal27: Same for Constructor?

thelovekesh: Yup, sounds good.

westonruter: Oh, I see. Just move the member variable above the static method. Sure.

westonruter: Done in 09637532597d2dd867d32d6193a75fcc93ecfecb and cc9e9562651ade093d5f769784b1667957a66a63

tests/modules/images/image-loading-optimization/storage/post-type-tests.php
```diff
@@ -138,22 +148,11 @@ public function test_ilo_store_url_metric() {
 
 		$url_metrics = ilo_parse_stored_url_metrics( $post );
 		$this->assertCount( 1, $url_metrics );
-		$this->assertArrayHasKey( 'timestamp', $url_metrics[0] );
-		$this->assertIsFloat( $url_metrics[0]['timestamp'] );
-		$this->assertLessThanOrEqual( microtime( true ), $url_metrics[0]['timestamp'] );
```
westonruter, line 143: Irrelevant to test now that the `timestamp` is required at the time of instantiation of `ILO_URL_Metric`. Previously it would get added by `ilo_unshift_url_metrics()`.

Same goes with the deleted assertions below.

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -140,11 +140,8 @@ static function ( $breakpoint_url_metrics ) use ( $sample_size ) {
 				// Sort URL metrics in descending order by timestamp.
 				usort(
 					$breakpoint_url_metrics,
-					static function ( $a, $b ) {
-						if ( ! isset( $a['timestamp'] ) || ! isset( $b['timestamp'] ) ) {
-							return 0;
-						}
```
westonruter, line 146: No need for this now that the timestamp is guaranteed to be set.

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -121,15 +121,15 @@ function ilo_verify_url_metrics_storage_nonce( string $nonce, string $slug ): bo
...
- * @return array Updated URL metrics, with timestamp key added.
+ * @param ILO_URL_Metric[] $url_metrics    Existing URL metrics. Each URL metric is expected to have a timestamp key.
+ * @param ILO_URL_Metric   $new_url_metric Validated URL metric. See JSON Schema defined in ilo_register_endpoint().
+ * @param int[]            $breakpoints    Breakpoint max widths.
+ * @param int              $sample_size    Sample size for URL metrics at a given breakpoint.
+ *
+ * @return ILO_URL_Metric[] Updated URL metrics.
  */
-function ilo_unshift_url_metrics( array $url_metrics, array $validated_url_metric, array $breakpoints, int $sample_size ): array {
-	$validated_url_metric['timestamp'] = microtime( true );
```
westonruter, line 131: This is now set more naturally in `ilo_handle_rest_request()`

modules/images/image-loading-optimization/storage/data.php
```diff
@@ -254,13 +251,9 @@ static function ( $breakpoint ) {
 	$grouped = array_fill_keys( array_merge( array( 0 ), $minimum_viewport_widths ), array() );
 
 	foreach ( $url_metrics as $url_metric ) {
-		if ( ! isset( $url_metric['viewport']['width'] ) ) {
-			continue;
-		}
-
```
westonruter, line 260: No need for this anymore since the property is guaranteed to exist.

modules/images/image-loading-optimization/storage/post-type.php
```diff
@@ -99,49 +99,42 @@ function ilo_parse_stored_url_metrics( WP_Post $post ): array {
...
 		);
-		$url_metrics = array();
+		$url_metrics_data = array();
 	}
 
 	return array_values(
 		array_filter(
-			$url_metrics,
-			static function ( $url_metric ) use ( $trigger_error ) {
-				// TODO: If we wanted, we could use the JSON Schema to validate the stored metrics.
```
westonruter, line 118: This TODO is now done

modules/images/image-loading-optimization/storage/rest-api.php
```diff
@@ -71,83 +101,6 @@ function ilo_register_endpoint() {
...
-					'required'          => true,
-					'pattern'           => '^[0-9a-f]+$',
-					'validate_callback' => static function ( $nonce, WP_REST_Request $request ) {
-						if ( ! ilo_verify_url_metrics_storage_nonce( $nonce, $request->get_param( 'slug' ) ) ) {
-							return new WP_Error( 'invalid_nonce', __( 'URL metrics nonce verification failure.', 'performance-lab' ) );
-						}
-						return true;
-					},
-				),
-				'viewport' => array(
```
westonruter, line 102: Schema definition is moved to `ILO_URL_Metrics`.

modules/images/image-loading-optimization/class-ilo-url-metric.php
```diff
@@ -0,0 +1,182 @@
...
+	 *             intersectionRect: array{ width: int, height: int },
+	 *             boundingClientRect: array{ width: int, height: int },
+	 *         }>
+	 */
+	public function get_elements(): array {
+		return $this->data['elements'];
+	}
+
+	/**
+	 * Gets the JSON representation of the object.
```
adamsilverstein, line 174: this function returns the object directly. is this doc line right? why  does it say it gets a "JSON representation"?

modules/images/image-loading-optimization/storage/post-type.php
```diff
@@ -99,49 +99,42 @@ function ilo_parse_stored_url_metrics( WP_Post $post ): array {
...
-				return $is_valid;
-			}
+			array_map(
+				static function ( $url_metric_data ) use ( $trigger_error ) {
+					if ( ! is_array( $url_metric_data ) ) {
+						return null;
+					}
+
+					try {
+						// TODO: This is re-validating the data which has been stored in the post type. This ensures it remains valid, but is it overkill?
```
adamsilverstein: feels fine to me, defensive coding

adamsilverstein (approved):
Looks good, left one comment

modules/images/image-loading-optimization/class-ilo-url-metric.php
westonruter: You're right. I've actually updated this in #1010 so it now reads as:

https://github.com/WordPress/performance/blob/f32a0e5f889e31f5af8a30701c3a5a55a1b37504/modules/images/image-loading-optimization/class-ilo-url-metric.php#L175

modules/images/image-loading-optimization/storage/post-type.php
westonruter: My only hesitation is that this additional validation adds latency, so it negatively impacts TTFB. However, I haven't profiled the JSON Schema validation logic, so this may not be a big deal.

modules/images/image-loading-optimization/class-ilo-url-metric.php
```diff
@@ -0,0 +1,182 @@
...
+
+	/**
+	 * Constructor.
+	 *
+	 * @param array $data      URL metric data.
+	 * @param bool  $validated Whether the data was already validated.
+	 *
+	 * @throws Exception When the input is invalid.
+	 */
+	public function __construct( array $data, bool $validated = false ) {
```
felixarntz: Boolean parameters to change what a method does are an antipattern, as methods should have a single responsibility.

I think we should consider alternatives:
* Introduce a `validate()` method that is called after the constructor?
* Or we use a `ILO_Validated_URL_Metric` class that extends this class but includes validation?
* Or handle validation elsewhere, such as a `ILO_URL_Metric_Validator` class?

But taking a step back, all of these including the current option isn't great in the way that it requires the code to dictate whether validation happens. Is it really that bad to validate at all times? Is it too costly? I wonder whether we should just always validate the data for a consistent behavior, and without requiring the code to be responsible for whether validation happens. That also seems safer in a way.

modules/images/image-loading-optimization/storage/post-type.php
```diff
@@ -99,49 +99,42 @@ function ilo_parse_stored_url_metrics( WP_Post $post ): array {
...
+						return new ILO_URL_Metric( $url_metric_data );
+					} catch ( Exception $e ) {
+						$trigger_error(
+							sprintf(
+								/* translators: %s is post type slug */
+								__( 'Unexpected shape to JSON array in post_content of %s post type.', 'performance-lab' ),
+								ILO_URL_METRICS_POST_TYPE
+							)
```
felixarntz: Maybe a nice little enhancement would be to also display the original exception message after the main message?

modules/images/image-loading-optimization/storage/rest-api.php
```diff
@@ -37,26 +37,56 @@
...
+				return true;
+			},
 		),
 	);
 
+	$schema = ILO_URL_Metric::get_json_schema();
+
+	// Make timestamp not required since it is forcibly-provided in ilo_handle_rest_request().
+	$schema['properties']['timestamp']['required'] = false;
+	$schema['properties']['timestamp']['readonly'] = true;
```
felixarntz: This feels a bit strange. Why not make this the actual schema in the first place? Or maybe we can set a `default`? Not a big deal, but there may be a cleaner solution.

felixarntz (commented):
@westonruter This mostly looks great. Just a few small points of feedback.

modules/images/image-loading-optimization/storage/rest-api.php
westonruter: I didn't want the client to be able to set the timestamp. I wanted it to always be server-supplied. I did actually set a default originally but I pulled it back out (but can't immediately recall why). I'll look at this again.

modules/images/image-loading-optimization/class-ilo-url-metric.php
westonruter: Yeah, this is the part I felt the most icky about. I did go down the route of having a subclass that doesn't require validation, but then thought it was maybe overkill to have a class just for this. I'm probably being too cautious here to avoid re-validating the data. I'll just make it require validation all the time.

westonruter: Done in c19c371e05d3c47eba2c39d0510cc1d1bb930a9a

modules/images/image-loading-optimization/storage/post-type.php
westonruter: Done in c5765e2d1e5f2c6d7af51de47f5070d65af152fe

modules/images/image-loading-optimization/storage/rest-api.php
westonruter: OK, the reason here is that the `timestamp` is required for instantiating a `ILO_URL_Metric` but it must be omitted in the REST API request. So in the former it is `required` but in the latter it is not. If the schema says it is `readonly` then it gets ignored entirely in the REST API, and the `default` value isn't used. I've updated the logic to supply the `default` in the handler instead. I think it's marginally better now: afc13ff3.

modules/images/image-loading-optimization/class-ilo-url-metric.php
```diff
@@ -0,0 +1,181 @@
...
+	 *              boundingClientRect: array{ width: int, height: int },
+	 *          }>
+	 *      }
+	 */
+	private $data;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param array $data      URL metric data.
```
felixarntz:
```suggestion
	 * @param array $data URL metric data.
```

modules/images/image-loading-optimization/storage/rest-api.php
felixarntz: Yeah, it's weird. But I like your new solution 👍

modules/images/image-loading-optimization/class-ilo-url-metric.php
```diff
@@ -0,0 +1,181 @@
...
+	 *              isLCP: bool,
+	 *              isLCPCandidate: bool,
+	 *              xpath: string,
+	 *              intersectionRatio: float,
+	 *              intersectionRect: array{ width: int, height: int },
+	 *              boundingClientRect: array{ width: int, height: int },
+	 *          }>
+	 *      }
+	 */
+	private $data;
```
felixarntz, line 33: Wouldn't it be valuable to store the `$url` and `$slug` in this class as well? Maybe they could be separate properties from `$data`, but I find it a bit odd to have a class for a URL metric exclude those identifying variables.

I think it could simplify some things by putting the responsibility for this information purely into the class representing the entity. We could still use associative arrays keyed by URL or slug if we need it, but I think it would be cleaner to have it covered by this class in its entirety - unless I'm missing something.

Alternatively, if there are scenarios where URL and slug should be decoupled from the actual _metric_ data, maybe we could separately introduce a class that is a value object representing URL and slug, and it could also provide the methods to create and validate the nonces etc. Just an idea.

modules/images/image-loading-optimization/storage/post-type.php
```diff
@@ -152,12 +146,12 @@ static function ( $url_metric ) use ( $trigger_error ) {
...
- * @param string $url                  URL for the URL metrics. This is used purely as metadata.
- * @param string $slug                 URL metrics slug (computed from query vars).
- * @param array  $validated_url_metric Validated URL metric. See JSON Schema defined in ilo_register_endpoint().
+ * @param string         $url            URL for the URL metrics. This is used purely as metadata.
+ * @param string         $slug           URL metrics slug (computed from query vars).
+ * @param ILO_URL_Metric $new_url_metric New URL metric.
  * @return int|WP_Error Post ID or WP_Error otherwise.
  */
-function ilo_store_url_metric( string $url, string $slug, array $validated_url_metric ) {
+function ilo_store_url_metric( string $url, string $slug, ILO_URL_Metric $new_url_metric ) {
```
felixarntz, line 154: See my above comment: I think a function like this could be made more intuitive if `ILO_URL_Metric` covered the URL and slug as well. This way you'd only pass around the object, not the other two values in addition.

modules/images/image-loading-optimization/storage/rest-api.php
```diff
@@ -37,18 +37,39 @@
...
+			},
+		),
+		'slug'  => array(
+			'type'        => 'string',
+			'description' => __( 'An MD5 hash of the query args.', 'performance-lab' ),
+			'required'    => true,
+			'pattern'     => '^[0-9a-f]{32}$',
+			// This is validated via the nonce validate_callback, as it is provided as input to create the nonce by the server
+			// which then is verified to match in the REST API request.
+		),
```
felixarntz, line 61: See above, this could be tied to the `ILO_URL_Metric` class.

felixarntz (commented):
@westonruter A few additional thoughts for consideration.

modules/images/image-loading-optimization/class-ilo-url-metric.php
westonruter: Well, the thing is that there are multiple URL metrics stored for a given response. The post type stores this list of URL metrics, based on the number of breakpoints and the sample size. So if we store the URL and slug with each `ILO_URL_Metric` we're storing multiple copies of the same data. The URL/slug is more higher level metadata which is only needed for the post type. We don't need the URL or slug except to get/set the post. I'd prefer to not include the URL or slug in this class.

westonruter: Well, let me think about this some more.

westonruter: Maybe we merge this as-is and I propose the change in another PR? I want to revisit how the URL is handled because in reality it is the request args really that matter, not the end URL. The slug is computed from the request args, not the URL.

westonruter: I've added this as a todo item on #869

felixarntz: That makes sense. How about a separate class though to represent the URL + slug and related logic (e.g. `ILO_URL` or something like that)? Then an instance of that class could be passed to `ILO_URL_Metric`.

Indeed the URL is not _identifying_ data as it is used for several metrics, but it'll be very intuitive from an architecture perspective that via a `ILO_URL_Metric` instance you can find out which URL it is for. This is similar to any parent entity of another entity. From the child entity you can find out which the parent entity is, which is common in REST representations just as in databases - for example in WordPress the parent post of another post, or the network of a site in a multisite.

Additionally, by using a class instance there's no duplication in terms of storage. You can pass the same `ILO_URL` instance to multiple `ILO_URL_Metric` instances, and since objects are passed by reference, they'd all reference the same data by definition. I think that would be intuitive, and it would unlock a few places where then you could pass only an `ILO_URL_Metric` instance. And then code that needs URL data could do something like `$metric->get_url()->get_url()`. Or maybe the class or at least the method could receive a different name to avoid this weirdness, such as `$metric->get_url_entity()->get_url()` which looks more intuitive.

westonruter: Ultimately, I think it would actually make sense to store a `url` with each URL Metric because in actuality the URL can vary while the request args change. This could be important for debugging, for example if it turns out that a template is modified by direct access to `$_GET` and not for a registered query var.

But I want to address this in a subsequent PR.

felixarntz: Oh I see you added more comments in the meantime. Let me know what you think about the above proposal.

But in any case, this would be okay to do in a follow up PR.

westonruter: Yeah, let's do in a separate PR. It would actually be useful to duplicate the URL and even the request args for each URL metric since it reveals the underlying data for the response that created the URL metric. The `slug` used in the post type is computed from a normalized set of request args, so having the original request args would still be useful to have on hand.

felixarntz: Sounds great!

felixarntz (approved):

modules/images/image-loading-optimization/class-ilo-url-metric.php
westonruter: See https://github.com/WordPress/performance/pull/1043

PR #989, opened 2024-02-15
Remove hard-coded PHPStan config

westonruter:
## Summary

Removing the hard-coded path to `phpstan.neon.dist` allows for PHPStan to pick the config file, allowing for local overrides. For example, I am using a stricter PHPStan config than is currently committed to the project. I have this config in `phpstan.neon` and [according to the PHPStan docs](https://phpstan.org/config-reference#:~:text=If%20you%20do%20not%20provide%20a%20config%20file%20explicitly), this should be chosen by PHPStan over the `phpstan.neon.dist`: 

<blockquote>
<p>If you do not provide a config file explicitly, PHPStan will look for files named <code>phpstan.neon</code>, <code>phpstan.neon.dist</code>, or <code>phpstan.dist.neon</code> in the current directory.</p>
<p>The resolution priority is as such:</p>
<ol> <li>If config file is provided as a command line option, it will be used.</li> <li>Otherwise, if <code>phpstan.neon</code> exists in the current working directory, it will be used.</li> <li>Otherwise, if <code>phpstan.neon.dist</code> exists in the current working directory, it will be used.</li> <li>Otherwise, if <code>phpstan.dist.neon</code> exists in the current working directory, it will be used.</li> <li>If none of the above is true, no config will be used.</li> </ol>
</blockquote>

This is not working at present, however, because the `phpstan.neon.dist` is the hard-coded config path.

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
   "scripts": {
     "post-install-cmd": "if php -r 'exit( version_compare( phpversion(), \"8.1\", \">=\" ) ? 0 : 1 );'; then composer --working-dir=build-cs install --no-interaction; else echo 'Skipping composer install for build-cs since not on PHP 8.1+. You are running: '; php -v;  fi",
     "post-update-cmd": "if php -r 'exit( version_compare( phpversion(), \"8.1\", \">=\" ) ? 0 : 1 );'; then composer --working-dir=build-cs update --no-interaction; else echo 'Skipping composer update for build-cs since not on PHP 8.1+. You are running: '; php -v; fi",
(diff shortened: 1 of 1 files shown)


swissspidy (approved):

thelovekesh (approved):

thelovekesh:
@westonruter I think the same can be done for PHPUnit as well, in case someone wants to override the file locally. Also, would it be better if we could include `phpstan.neon` in `.gitignore`?

westonruter:
@thelovekesh oh, good points! I merged before I saw them.

thelovekesh:
@westonruter No worries, I will raise a PR for that.

PR #995, opened 2024-02-16
Remove PHPUnit and PHPCS config file path from composer scripts

thelovekesh:
## Summary

Previously: https://github.com/WordPress/performance/pull/989

Remove hard-coded config file paths to let the tools decide which config to load on the user machine. With this, the following tools will be able to:
- PHPCS will be able to prioritize `phpcs.xml` over `phpcs.xml.dist` if present locally.
- PHPUnit will be able to prioritize `phpunit.xml` over `phpunit.xml.dist` if present locally.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,7 +43,7 @@ vendor/
 .DS_store?
 
 ############
-## CS and tools
+## CS tools
 ############
 phpcs.xml
 phpunit.xml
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -42,6 +42,13 @@ vendor/
 *.DS_store
 .DS_store?
 
+############
+## CS tools
```
westonruter:
```suggestion
## Config overrides for CS tools
```

westonruter (approved):

swissspidy (approved):

PR #1000, opened 2024-02-20
Implement publishing workflow for standalone plugins that aren't modules

### LINKED ISSUE #935: Implement publishing workflow for standalone plugins that aren't modules
This issue is part 2/2 for setting up the new infrastructure for developing standalone plugins.

## Requirements

**This work should be implemented against a new `feature/modules-to-plugins` feature branch.**

After #934 has been completed:
* The  `php-test-standalone-plugins.yml` and `deploy-standalone-plugins.yml` workflows should be enhanced to also cover plugins found in the `plugins` directory (additionally to plugins in the `modules` directory, which need to be built first).
* Some key differences to look out for:
    * Plugins in the `plugins` folder can be assumed to truly be plugins (not modules). So they don't require a build process (at least not for now).
    * Modules still need to be built, the same way they are now.
    * In other words, the ready-to-deploy versions of modules continue to live in a folder within `build`, while plugins can be deployed directly from their folder within `plugins`.
* Both workflows should only act for plugins that are part of `plugins.json`.
* The `npm run test-plugins` command will likely need to be expanded to also cover plugins in the `plugins` folder (but only those that are in `plugins.json`).

mukeshpanchal27:
## Summary

Fixes #935

The PR contains the following changes:
- Updated the `plugins.json` file to support the plugins config
- Introduced the `get-plugin-dir` command to retrieve the plugin directory, either build or plugins
- Updated `get-plugin-version` to allow returning plugin versions
- Updated `test-plugins` command to also test the plugins
- Updated `deploy-standalone-plugins.yml` workflow to release the plugins

Some notes:
- Added `pull_request` for checking automated workflow
- Set both auto and manual workflow DRY Run to true for security
- Updated SVN credentials for double security

## Checklist


mukeshpanchal27:
Automated deploy working fine: https://github.com/WordPress/performance/actions/runs/7969138800/job/21754473267?pr=1000

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,96 @@
...
+
+	// Check for valid and not empty object resulting from plugins JSON file parse.
+	if (
+		'object' !== typeof pluginsConfig ||
+		0 === Object.keys( pluginsConfig ).length
+	) {
+		throw Error(
+			`File at "${ pluginsFile }" parsed, but detected empty/non valid JSON object.`
+		);
+	}
```
thelovekesh: Can we try
```js
const { plugins, modules } = require( './plugins.json' );
```
considering this is a CLI script and we need the instance of `plugins.json` only once?

plugins.json
```diff
@@ -9,5 +9,10 @@
       "version": "1.0.5"
     }
   },
-  "plugins": [ "auto-sizes" ]
+  "plugins": {
+    "auto-sizes": {
+      "slug": "auto-sizes",
+      "version": "1.0.1"
```
westonruter: Should this be necessary if the `version` is already present here:

https://github.com/WordPress/performance/blob/36fe0dce276a81db16a7bd0dc13c453d0d207314/plugins/auto-sizes/auto-sizes.php#L8

Should this not rather parse the version out of the plugin file? Or we also have it present in the `readme.txt`:

https://github.com/WordPress/performance/blob/36fe0dce276a81db16a7bd0dc13c453d0d207314/plugins/auto-sizes/readme.txt#L7

I think we should avoid having this version in three different places.

westonruter: If we just store the version in Stable Tag of `readme.txt` then the plugin file could have `n.e.x.t` as the version which gets replaced during deployment, or vice-versa.

westonruter: Or the version could be stored in `plugins.json` like you have here, but if so, then `n.e.x.t` should be the version in `auto-sizes.php` and `readme.txt` to avoid having to manually keep them in sync.

thelovekesh: Like the idea of storing the version in a single place only. As mentioned `n.e.x.t` strategy can be used to populate the version on the fly.

bin/plugin/commands/get-plugin-dir.js
mukeshpanchal27: Thanks for raising. Updated in https://github.com/WordPress/performance/pull/1000/commits/d6e79c85fb91525a63a93d1f314ea5aaf069eda3

plugins.json
mukeshpanchal27: Thanks, @westonruter, for bringing this up. The plugin versions were added for the deploy process per https://github.com/WordPress/performance/issues/935#issuecomment-1952027981. Since there is no build project for `plugins`, in my opinion, we should discuss this first and address it in a follow-up issue. What do you think? cc. @joemcgill

bin/plugin/commands/test-plugins.js
```diff
@@ -445,8 +447,8 @@ function doRunStandalonePluginTests( settings ) {
 		process.exit( 1 );
 	}
 
-	const plugins = pluginsConfig.modules;
-	if ( ! plugins ) {
+	const stPlugins = pluginsConfig.modules;
```
westonruter: What does the `st` prefix mean? Oh, standalone? I think that should then be spelled out to be more clear.

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,86 @@
...
+ */
+exports.handler = async ( opt ) => {
+	doRunGetPluginDir( {
+		pluginsJsonFile: 'plugins.json', // Path to plugins.json file.
+		slug: opt.slug, // Plugin slug.
+	} );
+};
+
+/**
+ * Returns directory for plugin or module based on the slug.
```
westonruter:
```suggestion
 * Prints directory root for plugin or module based on the slug.
```

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,86 @@
...
+		if ( modules || Object.keys( modules ).length !== 0 ) {
+			for ( const moduleDir in modules ) {
+				const pluginVersion = modules[ moduleDir ]?.version;
+				const pluginSlug = modules[ moduleDir ]?.slug;
+				if (
+					pluginVersion &&
+					pluginSlug &&
+					settings.slug === pluginSlug
+				) {
+					return log( 'build' );
```
westonruter: The `console.log` function doesn't return anything, right?

```suggestion
					log( 'build' );
					return;
```

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,86 @@
...
+		if ( plugins || Object.keys( plugins ).length !== 0 ) {
+			for ( const pluginDir in plugins ) {
+				const pluginVersion = plugins[ pluginDir ]?.version;
+				const pluginSlug = plugins[ pluginDir ]?.slug;
+				if (
+					pluginVersion &&
+					pluginSlug &&
+					settings.slug === pluginSlug
+				) {
+					return log( 'plugins' );
```
westonruter:
```suggestion
					log( 'plugins' );
					return;
```

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,86 @@
...
+
+		// Validate that the modules object is not empty.
+		if ( modules || Object.keys( modules ).length !== 0 ) {
+			for ( const moduleDir in modules ) {
+				const pluginVersion = modules[ moduleDir ]?.version;
+				const pluginSlug = modules[ moduleDir ]?.slug;
+				if (
+					pluginVersion &&
+					pluginSlug &&
+					settings.slug === pluginSlug
```
westonruter: Seems this can be simplified:

```suggestion
			for ( const module of Object.values( modules ) ) {
				if (
					module.version &&
					settings.slug === module.slug
```

Also, why even check for `module.version`? It's not being used. If wanting to validate that it exists, perhaps it should rather throw an exception if absent since this indicates a malformed JSON file?

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,86 @@
...
+
+		// Validate that the plugins object is not empty.
+		if ( plugins || Object.keys( plugins ).length !== 0 ) {
+			for ( const pluginDir in plugins ) {
+				const pluginVersion = plugins[ pluginDir ]?.version;
+				const pluginSlug = plugins[ pluginDir ]?.slug;
+				if (
+					pluginVersion &&
+					pluginSlug &&
+					settings.slug === pluginSlug
```
westonruter:
```suggestion
			for ( const plugin of Object.values( plugins ) ) {
				if (
					plugin.version &&
					settings.slug === plugin.slug
```

Again, why even check for `plugin.version`? It's not being used.

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Internal dependencies
+ */
+const { log } = require( '../lib/logger' );
+
+exports.options = [
+	{
+		argname: '-s, --slug <slug>',
+		description: 'Plugin or Standalone module/plugin slug to get directory',
```
westonruter:
```suggestion
		description: 'Slug to search out whether it is a plugin or module.',
```

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Command to get directory for plugin/module based on the slug.
+ *
+ * @param {Object} opt Command options.
+ */
+exports.handler = async ( opt ) => {
+	doRunGetPluginDir( {
+		pluginsJsonFile: 'plugins.json', // Path to plugins.json file.
+		slug: opt.slug, // Plugin slug.
```
westonruter:
```suggestion
		slug: opt.slug, // Plugin/module slug.
```

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,86 @@
...
+		__dirname,
+		'../../../' + settings.pluginsJsonFile
+	);
+
+	try {
+		// Read the plugins.json file synchronously.
+		const { modules, plugins } = require( pluginsFile );
+
+		// Validate that the modules object is not empty.
+		if ( modules || Object.keys( modules ).length !== 0 ) {
```
westonruter: If this is empty, should it not rather throw an exception?

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,86 @@
...
+					pluginSlug &&
+					settings.slug === pluginSlug
+				) {
+					return log( 'build' );
+				}
+			}
+		}
+
+		// Validate that the plugins object is not empty.
+		if ( plugins || Object.keys( plugins ).length !== 0 ) {
```
westonruter: If this is empty, should it not rather throw an exception?

bin/plugin/commands/get-plugin-version.js
```diff
@@ -38,52 +37,50 @@ function doRunGetPluginVersion( settings ) {
...
+		// Validate that the modules object is not empty.
+		if ( modules || Object.keys( modules ).length !== 0 ) {
+			for ( const moduleDir in modules ) {
+				const pluginVersion = modules[ moduleDir ]?.version;
+				const pluginSlug = modules[ moduleDir ]?.slug;
+				if (
+					pluginVersion &&
+					pluginSlug &&
+					settings.slug === pluginSlug
+				) {
```
westonruter: Ditto with above. This can be simplified with a for/of loop. If either the version or slug are empty, it should throw an error because the JSON file is bad.

bin/plugin/commands/get-plugin-version.js
```diff
@@ -38,52 +37,50 @@ function doRunGetPluginVersion( settings ) {
...
-		if ( pluginVersion && pluginSlug && settings.slug === pluginSlug ) {
-			return log( pluginVersion );
+		// Validate that the plugins object is not empty.
+		if ( plugins || Object.keys( plugins ).length !== 0 ) {
+			for ( const pluginDir in plugins ) {
+				const pluginVersion = plugins[ pluginDir ]?.version;
+				const pluginSlug = plugins[ pluginDir ]?.slug;
+				if (
+					pluginVersion &&
+					pluginSlug &&
```
westonruter: Ditto

bin/plugin/commands/test-plugins.js
```diff
@@ -458,23 +460,67 @@ function doRunStandalonePluginTests( settings ) {
...
+					// Handle the error appropriately
+					log(
+						formats.error(
+							`Error copying plugin "${ plugins[ item ].slug }": ${ e.message }`
+						)
+					);
+					return false;
+				}
+			} )
+			.map( ( item ) => plugins[ item ].slug )
```
westonruter: As elsewhere, this can be simplified since the object keys are not used:

```suggestion
		Object.values( plugins )
			.filter( ( plugin ) => {
				try {
					fs.copySync(
						`${ settings.pluginsDir }${ plugin.slug }/`,
						`${ settings.builtPluginsDir }${ plugin.slug }/`,
						{
							overwrite: true,
						}
					);
					log(
						formats.success(
							`Copied plugin "${ plugin.slug }".\n`
						)
					);
					return true;
				} catch ( e ) {
					// Handle the error appropriately
					log(
						formats.error(
							`Error copying plugin "${ plugin.slug }": ${ e.message }`
						)
					);
					return false;
				}
			} )
			.map( ( plugin ) => plugin.slug )
```

plugins.json
westonruter: Yes, I saw that comment. Still, not ideal to have the version in triplicate. If you want to address that in another issue, sure.

westonruter (changes requested):

bin/plugin/commands/test-plugins.js
swissspidy: Just `plugins` seems to suffice, no?

westonruter: That conflicts with another declaration below:

https://github.com/WordPress/performance/blob/e802a71dc43a22816384c3cfc0905b095b5c0648/bin/plugin/commands/test-plugins.js#L482

mukeshpanchal27: In https://github.com/WordPress/performance/pull/1000/commits/8a3c690ce87654522b610979c7d77deb135e982c i updated the variable name to `standalonePlugins`

plugins.json
mukeshpanchal27: Open a separate issue for further work: #1006

mukeshpanchal27:
Thanks, @westonruter, for the feedback. I've addressed all of it. The PR is now ready for review.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -69,15 +86,20 @@ jobs:
...
           dry-run: ${{ matrix.dry-run }}
         env:
-          SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
-          SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          # TODO Once the workflow is tested, we will remove the comment and use the secret SVN access.
+          #SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
+          #SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          # TODO Once the workflow is tested, we will remove this test credential.
+          SVN_PASSWORD: SVN_PASSWORD
+          SVN_USERNAME: SVN_USERNAME
```
swissspidy, line 111: Duplicate

bin/plugin/commands/get-plugin-version.js
```diff
@@ -38,52 +37,40 @@ function doRunGetPluginVersion( settings ) {
...
-	}
-
-	const plugins = pluginsConfig.modules;
-	if ( ! plugins ) {
-		throw Error(
-			`File at "${ pluginsFile }" parsed, but the modules are missing, or they are misspelled.`
-		);
-	}
+		// Validate that the modules object is not empty.
+		if ( modules && Object.keys( modules ).length !== 0 ) {
```
swissspidy: IMHO this safeguard is overly defensive and thus overkill.

As long as there's always at the minimum `modules: {}` in the `plugins.json` file, this is not needed. Especially the `Object.keys()` length is unnecessary, because the for loop will just do nothing on an empty array.

Thus I suggest we remove this unnecessary`if` here.

bin/plugin/commands/get-plugin-version.js
```diff
@@ -38,52 +37,40 @@ function doRunGetPluginVersion( settings ) {
...
+			}
+		}
 
-	for ( const moduleDir in plugins ) {
-		const pluginVersion = plugins[ moduleDir ]?.version;
-		const pluginSlug = plugins[ moduleDir ]?.slug;
-		if ( pluginVersion && pluginSlug && settings.slug === pluginSlug ) {
-			return log( pluginVersion );
+		// Validate that the plugins object is not empty.
+		if ( plugins && Object.keys( plugins ).length !== 0 ) {
```
swissspidy: IMHO this safeguard is overly defensive and thus overkill.

As long as there's always at the minimum `plugins: {}` in the `plugins.json` file, this is not needed. Especially the `Object.keys()` length is unnecessary, because the for loop will just do nothing on an empty array.

Thus I suggest we remove this unnecessary`if` here.

bin/plugin/commands/get-plugin-version.js
```diff
@@ -38,52 +37,40 @@ function doRunGetPluginVersion( settings ) {
...
-	const plugins = pluginsConfig.modules;
-	if ( ! plugins ) {
-		throw Error(
-			`File at "${ pluginsFile }" parsed, but the modules are missing, or they are misspelled.`
-		);
-	}
+		// Validate that the modules object is not empty.
+		if ( modules && Object.keys( modules ).length !== 0 ) {
+			for ( const module of Object.values( modules ) ) {
+				if ( module.version && settings.slug === module.slug ) {
```
swissspidy: When does a module _not_ have a version? 🤔

bin/plugin/commands/get-plugin-version.js
```diff
@@ -38,52 +37,40 @@ function doRunGetPluginVersion( settings ) {
...
 
-	for ( const moduleDir in plugins ) {
-		const pluginVersion = plugins[ moduleDir ]?.version;
-		const pluginSlug = plugins[ moduleDir ]?.slug;
-		if ( pluginVersion && pluginSlug && settings.slug === pluginSlug ) {
-			return log( pluginVersion );
+		// Validate that the plugins object is not empty.
+		if ( plugins && Object.keys( plugins ).length !== 0 ) {
+			for ( const plugin of Object.values( plugins ) ) {
+				if ( plugin.version && settings.slug === plugin.slug ) {
```
swissspidy: When does a plugin _not_ have a version?

bin/plugin/commands/get-plugin-version.js
westonruter: Yeah, it should always. So if absent it should throw an exception. Maybe there should be a helper function that loads the JSON file and runs through the validation checks so that it doesn't have to be done in the modules each time it's read.

swissspidy: Do we really need such validation? We have code reviews etc. I don't see how this file ever becomes invalid. And if we're worried about that, then such a validation check could be easily done in a GitHub Action.

bin/plugin/commands/get-plugin-version.js
```diff
@@ -38,52 +37,40 @@ function doRunGetPluginVersion( settings ) {
...
-	}
+		// Validate that the modules object is not empty.
+		if ( modules && Object.keys( modules ).length !== 0 ) {
+			for ( const module of Object.values( modules ) ) {
+				if ( module.version && settings.slug === module.slug ) {
+					log( module.version );
+					return;
+				}
+			}
+		}
```
westonruter:
```suggestion
		for ( const module of Object.values( modules ) ) {
			if ( settings.slug === module.slug ) {
				log( module.version );
				return;
			}
		}
```

bin/plugin/commands/get-plugin-version.js
```diff
@@ -38,52 +37,40 @@ function doRunGetPluginVersion( settings ) {
...
-			return log( pluginVersion );
+		// Validate that the plugins object is not empty.
+		if ( plugins && Object.keys( plugins ).length !== 0 ) {
+			for ( const plugin of Object.values( plugins ) ) {
+				if ( plugin.version && settings.slug === plugin.slug ) {
+					log( plugin.version );
+					return;
+				}
+			}
 		}
```
westonruter:
```suggestion
		// Validate that the plugins object is not empty.
		for ( const plugin of Object.values( plugins ) ) {
			if ( settings.slug === plugin.slug ) {
				log( plugin.version );
				return;
			}
		}
```

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,76 @@
...
+
+		// Validate that the plugins object is not empty.
+		if ( plugins && Object.keys( plugins ).length !== 0 ) {
+			for ( const plugin of Object.values( plugins ) ) {
+				if ( plugin.version && settings.slug === plugin.slug ) {
+					log( 'plugins' );
+					return;
+				}
+			}
+		}
```
westonruter:
```suggestion
for ( const module of Object.values( modules ) ) {
	if ( settings.slug === module.slug ) {
		log( 'build' );
		return;
	}
}

for ( const plugin of Object.values( plugins ) ) {
	if ( settings.slug === plugin.slug ) {
		log( 'plugins' );
		return;
	}
}
```

bin/plugin/commands/test-plugins.js
```diff
@@ -458,23 +460,65 @@ function doRunStandalonePluginTests( settings ) {
...
+	if ( ! plugins ) {
+		log(
+			formats.error(
+				'The given plugin configuration is invalid, the plugins are missing, or they are misspelled.'
+			)
+		);
+
+		// Return with exit code 1 to trigger a failure in the test pipeline.
+		process.exit( 1 );
+	}
```
westonruter:
```suggestion
```

bin/plugin/commands/test-plugins.js
```diff
@@ -445,8 +447,8 @@ function doRunStandalonePluginTests( settings ) {
 		process.exit( 1 );
 	}
 
-	const plugins = pluginsConfig.modules;
-	if ( ! plugins ) {
+	const standalonePlugins = pluginsConfig.modules;
+	if ( ! standalonePlugins ) {
```
westonruter: This `if` statement can be removed as well.

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,76 @@
...
+
+	// Resolve the absolute path to the plugins.json file.
+	const pluginsFile = path.join(
+		__dirname,
+		'../../../' + settings.pluginsJsonFile
+	);
+
+	try {
+		// Read the plugins.json file synchronously.
+		const { modules, plugins } = require( pluginsFile );
```
westonruter, line 51: Question: In `test-plugns.js` it is doing:

```js
const pluginsConfig = JSON.parse( pluginsJsonFileContent );
```

Shouldn't it rather use `require` like is used here? At least the same loading approach should be used between the two files.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -69,15 +86,20 @@ jobs:
...
           SLUG: ${{ matrix.slug }}
           VERSION: ${{ matrix.version }}
-          BUILD_DIR: ./build/${{ matrix.slug }}
-          ASSETS_DIR: ./build/${{ matrix.slug }}/.wordpress-org
+          BUILD_DIR: ./${{ matrix.directory }}/${{ matrix.slug }}
+          ASSETS_DIR: ./${{ matrix.directory }}/${{ matrix.slug }}/.wordpress-org
```
westonruter, line 115: I just noticed that `BUILD_DIR` points to the parent directory of `.wordpress-org`. Doesn't that mean `.wordpress-org` will get distributed with the plugin accidentally? Or will this not be the case due to `.gitattributes`?

joemcgill: The `ASSETS_DIR` variable is what is used by [the 10up deploy action](https://github.com/10up/action-wordpress-plugin-deploy?tab=readme-ov-file#optional-environment-variables) to tell the action where to find the asset directory for whatever is being deployed. Since we're declaring a different build directory, we also need to declare a different assets directory, otherwise the assets from Performance Lab would get deployed to each plugin.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -40,7 +57,7 @@ jobs:
             result=$(echo "${{ steps.get-version.outputs.version }}" | awk '/^(\*|[0-9]+(\.[0-9]+){0,2}(-[a-zA-Z0-9.]+)?)$/ {print "Matched"}')
             if [[ -n "$result" ]]; then
               # Set the manual input values in JSON format for use in the matrix.
-              echo "matrix={\"include\":[{\"slug\":\"${{ inputs.slug }}\",\"version\":\"${{ steps.get-version.outputs.version }}\",\"dry-run\":\"${{ inputs.dry-run }}\"}]}" >> $GITHUB_OUTPUT
+              echo "matrix={\"include\":[{\"slug\":\"${{ inputs.slug }}\",\"version\":\"${{ steps.get-version.outputs.version }}\",\"directory\":\"${{ steps.get-plugin-directory.outputs.directory }}\",\"dry-run\":\"true\"}]}" >> $GITHUB_OUTPUT
```
joemcgill, line 60: Don't forget to change the hard-coded `dry-run` setting here back to `inputs.dry-run`. Same below.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -69,15 +86,20 @@ jobs:
       - name: Install npm dependencies
         run: npm ci
       - name: Building standalone plugins
+        if: ${{ matrix.directory != 'plugins' }}
         run: npm run build-plugins
```
joemcgill, line 92: The language here is now confusing because the name is "Building standalone plugins" but in only runs when we're _not_ in the plugins directory. Can we remove this entirely in this branch, since we're going to be moving all of the standalone plugins to the plugins directory anyway?

joemcgill (commented):
Left a bit of feedback, but want to give this another look with fresh eyes and do some local testing.

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: Yes, for security reasons, and as we did it for all our workflow changes, once it is tested, we will remove it.

bin/plugin/commands/get-plugin-dir.js
mukeshpanchal27: Updated in https://github.com/WordPress/performance/pull/1000/commits/15b1cc0404f04e5e491ec36f0f5c3e7b764f2a86

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: Thanks @joemcgill for answering.

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: Yes. Will do that once testing done.

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: Yes, but for now, we need plugins from the build, so I have excluded these steps for the plugin, as we don't need a build process for it.

- Build - https://github.com/WordPress/performance/actions/runs/8015338254/job/21895352814?pr=1000
- Plugin - https://github.com/WordPress/performance/actions/runs/8015338254/job/21895352942?pr=1000

.github/workflows/deploy-standalone-plugins.yml
joemcgill: @mukeshpanchal27 looking at the test runs on this PR, it does look like the `ASSETS_DIR` isn't working as expected though and instead of the files in this folder getting added, the folder itself is getting added. See: https://github.com/WordPress/performance/actions/runs/8015338254/job/21895352814?pr=1000#step:6:71

Something to look into.

plugins.json
joemcgill: I spent some time looking into how we could modify the release job to avoid the duplication of version numbers here for this, and I'm not sure it's worth the effort for now. Once we finish migrating all of the standalone modules to the plugins folder we can clean this up more easily.

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: @joemcgill The workflow works similarly to how it did for the `2.9.0` release https://github.com/WordPress/performance/actions/runs/7978814725/job/21784762898. Could you please point out what's not working?

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,70 @@
...
+	doRunGetPluginDir( {
+		pluginsJsonFile: 'plugins.json', // Path to plugins.json file.
+		slug: opt.slug, // Plugin/module slug.
+	} );
+};
+
+/**
+ * Prints directory root for plugin or module based on the slug.
+ *
+ * @param {Object} settings Plugin settings.
```
westonruter:
```suggestion

@param {Object} settings                 Plugin settings.
@param {string} settings.pluginsJsonFile Path to plugins JSON file.
@param {string} settings.slug            Slug for the plugin or module.
```

Same could be done in `get-plugin-version.js`

bin/plugin/commands/get-plugin-dir.js
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Command to get directory for plugin/module based on the slug.
+ *
+ * @param {Object} opt Command options.
+ */
+exports.handler = async ( opt ) => {
+	doRunGetPluginDir( {
+		pluginsJsonFile: 'plugins.json', // Path to plugins.json file.
+		slug: opt.slug, // Plugin/module slug.
```
westonruter:
```suggestion
 * @param {Object} opt      Command options.
 * @param {string} opt.slug Plugin/module slug.
 */
exports.handler = async ( opt ) => {
	doRunGetPluginDir( {
		pluginsJsonFile: 'plugins.json', // Path to plugins.json file.
		slug: opt.slug,
```

Same could be done in `get-plugin-version.js`

westonruter (approved):
Just a couple trivial suggestions not blocking merge, but otherwise LGTM.

.github/workflows/deploy-standalone-plugins.yml
felixarntz: @westonruter Regarding your question, yes due to the `.gitattributes` file the `.wordpress-org` directories should not be part of the ZIP builds. At least this is already the case for modules built into plugins (via `/modules/**/.wordpress-org export-ignore`).

@mukeshpanchal27 We need to make sure the `/plugins/**/.wordpress-org` directories are ignored for the plugin ZIP / deployment as well.

plugins.json
felixarntz: @mukeshpanchal27 @joemcgill I just voiced a similar concern as @westonruter in https://github.com/WordPress/performance/issues/935#issuecomment-1965535471, and his feedback is precisely why the requirements only mentioned plugin slugs (no need to duplicate the version number).

I'm okay for this to be addressed in a separate issue, but since this PR is actually making the change and the `feature/modules-to-plugins` branch currently uses the _correct_ format, it's a little strange IMO to go ahead with this change just to change it back afterwards.

If this PR is blocking other work and therefore would benefit from a merge ASAP, I'm okay to fix this again in a follow up issue, but it should be addressed shortly after this PR, in time for 3.0, as it requires changes to the `plugins.json` file format/spec, which will affect documentation.

felixarntz (changes requested):
@mukeshpanchal27 This looks solid to me, except for two things:
* The change of including the plugin versions in `plugins.json` is unnecessary, based on the other feedback. The way the file is structured in `feature/modules-to-plugins` is correct.
* Per https://github.com/WordPress/performance/pull/1000/files#r1503461064, let's double check that the `.wordpress-org` directories of the plugins are excluded from the ZIP builds. We probably just have to add an according directive to the `.gitattributes` file for that to work.

plugins.json
mukeshpanchal27: Thanks for the feedback.

In https://github.com/WordPress/performance/pull/1000/commits/a4cf3bf058093370ccdf067027edd53049493922, I have reverted the version-related changes for plugins and instead fetched the version from the `readme.txt` file. I updated the script accordingly. I opted not to use the plugin's main file because different plugins have different main files, such as `plugin/load.php` or `plugin/plugin.php`.

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: Understood, I've added it to the exclude list.

mukeshpanchal27:
Thanks @felixarntz @westonruter @swissspidy @joemcgill for the review. I've made updates based on the suggestions and discussions:

- Removed versions from the plugins in the `plugins.json` file.
- Extended the version script to read versions from `readme.txt`.
- Excluded `.wordpress-org` for plugins.
- Updated docblocks according to the suggestions.

swissspidy (approved):

felixarntz (approved):
Looks great, thanks @mukeshpanchal27!

PR #1002, opened 2024-02-21
Separate `phpcs.xml.dist` Files for Each Plugin to Isolate Text Domains

### LINKED ISSUE #997: Separate `phpcs.xml.dist` Files for Each Plugin to Isolate Text Domains
Follow-up: https://github.com/WordPress/performance/pull/972#discussion_r1480801832

Currently, our project utilizes a single `phpcs.xml.dist` file containing text domains for multiple plugins. However, this setup poses a risk of accidental misuse of text domains, especially during copy/paste operations between plugins. To mitigate this risk and enhance code quality, we propose maintaining separate `phpcs.xml.dist` files for each plugin.

cc. @joemcgill

mukeshpanchal27:
## Summary

Fixes #997

## Checklist


composer.json
```diff
@@ -24,7 +24,7 @@
     "post-update-cmd": "if php -r 'exit( version_compare( phpversion(), \"8.1\", \">=\" ) ? 0 : 1 );'; then composer --working-dir=build-cs update --no-interaction; else echo 'Skipping composer update for build-cs since not on PHP 8.1+. You are running: '; php -v; fi",
     "phpstan": "build-cs/vendor/bin/phpstan analyse --memory-limit=2048M -c phpstan.neon.dist",
     "format": "build-cs/vendor/bin/phpcbf --standard=phpcs.xml.dist --report-summary --report-source",
-    "lint": "build-cs/vendor/bin/phpcs --standard=phpcs.xml.dist",
+    "lint": "build-cs/vendor/bin/phpcs",
```
thelovekesh, line 27: The same is updated in https://github.com/WordPress/performance/pull/995. I think a rebase with trunk should update it.

phpcs.ruleset.xml
```diff
@@ -0,0 +1,100 @@
...
+	<rule ref="PHPCompatibility"/>
+	<config name="testVersion" value="7.0-"/>
+
+	<rule ref="WordPress-Docs"/>
+	<rule ref="WordPress-Extra" />
+	<rule ref="WordPress.WP.I18n"/>
+
+	<arg value="ps"/>
+	<arg name="extensions" value="php"/>
+
```
thelovekesh, line 14:
```suggestion
<arg name="parallel" value="20"/>

```
Would be great to include this to speed up linting. See https://github.com/WordPress/performance/pull/953/files#r1469088154

plugins/auto-sizes/.gitattributes
thelovekesh, line 1: I am not fully sure but as per the previous discussion, there will be no build steps in standalone plugins. Is this something we really need?

plugins/auto-sizes/.gitignore
thelovekesh, line 1: I think the `.gitignore` in the root should do this already. If some path is not working we can make it [relative](https://git-scm.com/docs/gitignore#_pattern_format)?

plugins/auto-sizes/.gitattributes
swissspidy: Shouldn't the `.gitattributes` file in the root cover all this? I don't see the need to duplicate this file.

thelovekesh:
Tests the PHPCS config with the following commands by changing the text domains, and it's working as expected:

<details><summary>Commands output</summary>

```sh
➜  performance git:(fix/997-separate-phpcs) ✗ build-cs/vendor/bin/phpcs                                             
......E..................................................... 60 / 69 (87%)
.........                                                    69 / 69 (100%)



FILE: /home/thelovekesh/Desktop/loki/wordpressdev/public/content/plugins/performance/modules/database/audit-autoloaded-options/can-load.php
--------------------------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------------------------------------------------------------------
 15 | ERROR | Mismatched text domain. Expected 'performance-lab' or 'default' but got 'performance'. (WordPress.WP.I18n.TextDomainMismatch)
--------------------------------------------------------------------------------------------------------------------------------------------

Time: 2.3 secs; Memory: 36MB

➜  performance git:(fix/997-separate-phpcs) ✗ build-cs/vendor/bin/phpcs plugins/auto-sizes --standard=plugins/auto-sizes/phpcs.xml.dist
.E 2 / 2 (100%)



FILE: /home/thelovekesh/Desktop/loki/wordpressdev/public/content/plugins/performance/plugins/auto-sizes/hooks.php
---------------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
---------------------------------------------------------------------------------------------------------------------------------
 29 | ERROR | Mismatched text domain. Expected 'auto-sizes' or 'default' but got 'auto-'. (WordPress.WP.I18n.TextDomainMismatch)
---------------------------------------------------------------------------------------------------------------------------------

Time: 85ms; Memory: 16MB
```

</details>

swissspidy:
Btw I am looking into those PHP unit test failures, probably has to do with yesterday's MariaDB update.

phpcs.xml.dist
```diff
@@ -1,108 +1,13 @@
...
-	<rule ref="WordPress-Docs"/>
-	<rule ref="WordPress-Extra" />
-	<rule ref="WordPress.WP.I18n"/>
-	<config name="text_domain" value="performance-lab,default,auto-sizes"/>
-
-	<arg value="ps"/>
-	<arg name="extensions" value="php"/>
+	<rule ref="phpcs.ruleset.xml"/>
+	
+	<config name="text_domain" value="performance-lab,default"/>
```
swissspidy, line 6: Why are we even allowing the default text domain?

plugins/auto-sizes/phpcs.xml.dist
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0"?>
+<ruleset name="WordPress Coding Standards for Auto-sizes for Lazy-loaded Images Plugin">
+
+	<rule ref="phpcs.ruleset.xml"/>
+
+	<config name="text_domain" value="auto-sizes,default"/>
```
westonruter:
```suggestion
	<config name="text_domain" value="auto-sizes"/>
```

phpcs.xml.dist
westonruter: Agreed. We're not using it, as far as I can see.

```suggestion
	<config name="text_domain" value="performance-lab"/>
```

Should only be allowed if we're intentionally re-using core strings.

plugins/auto-sizes/phpcs.xml.dist
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0"?>
+<ruleset name="WordPress Coding Standards for Auto-sizes for Lazy-loaded Images Plugin">
+
+	<rule ref="phpcs.ruleset.xml"/>
```
westonruter: This is a bit confusing, as I would expect it to be `../../phpcs.ruleset.xml`. But it needs to be this way because phpcs gets invoked from the repo root?

phpcs.xml.dist
swissspidy: Yep, and if we want to intentionally re-use core strings, we can add ignore comments with explanations.

plugins/auto-sizes/.gitattributes
mukeshpanchal27: File removed in https://github.com/WordPress/performance/pull/1002/commits/94550b0e12181133ad76c7df3b534406f0a9b9ba

phpcs.xml.dist
mukeshpanchal27: Thanks for raising this. We use the `default` text domain in two files with multiple instances. Do we need to add an inline comment to ignore these? I'm happy to do it either way.
- https://github.com/WordPress/performance/blob/trunk/admin/load.php uses it in 5 places.
- https://github.com/WordPress/performance/blob/trunk/admin/plugins.php uses it in 30 places.

thelovekesh: I see it's being used in multiple places within two files, so adding a comment for each line is not a good idea. Rather two points came to my mind:
- Let it be in PHPCS config, as we may add this text domain in future code.
- Try to add this on the file level instead?

composer.json
mukeshpanchal27: Open PR #1007 that merge trunk branch changes.

phpcs.xml.dist
westonruter: OK, I missed those two instances. In that case, could this line:

```xml
<config name="text_domain" value="performance-lab,default"/>
```

Only be applied to `admin/load.php` and `admin/plugins.php`? I don't think PHPCS has such a capability. 

In that case, I'd say we leave `default` as-is in the PL plugin's `phpcs.xml.dist` but make sure we omit it from the standalone plugins (unless we really need it). We can revisit later, as it leaving `default` would not be any different than the behavior today.

swissspidy: Agreed, let's restrict usage to just these two files, either via PHPCS config specific to the files or file-level comments.

westonruter: But I don't think `config` can be selectively applied to files and I don't think comments can be used to override config values. So either we just keep `default` in the config for the PL plugin, or we add comments to disable PHPCS warnings about using `default` in those two files (which I think is what your second suggestion).

phpcs.ruleset.xml
```diff
@@ -0,0 +1,93 @@
+<?xml version="1.0"?>
+<ruleset name="WordPress Coding Standards for Performance related Plugins">
```
joemcgill: We could take this opportunity to simplify the ruleset name, since the description could be more verbose. Ex:

```suggestion
<ruleset name="WPP-Base">
```

phpcs.xml.dist
```diff
@@ -1,108 +1,23 @@
...
-				<element value="createTextNode"/>
-				<element value="textContent"/>
-			</property>
-		</properties>
-	</rule>
+	<file>./admin</file>
+	<file>./load.php</file>
+	<file>./modules</file>
+	<file>./server-timing</file>
+	<file>./tests</file>
```
joemcgill: I think it's better to exclude than include things here, otherwise we can potentially miss linting new files and folders that we add to the plugin. For example, there is an uncaught error in our `plugin-tests` folder at the moment:

```suggestion
	<file>.</file>
	<exclude-pattern>./build/*</exclude-pattern>
	<exclude-pattern>./node_modules/*</exclude-pattern>
	<exclude-pattern>./plugins/*</exclude-pattern>
	<exclude-pattern>./vendor/*</exclude-pattern>
```

Note that by excluding the plugins here, we have to make sure we're checking phpcs separately for each plugin.

plugins/auto-sizes/phpcs.xml.dist
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0"?>
+<ruleset name="WordPress Coding Standards for Auto-sizes for Lazy-loaded Images Plugin">
```
joemcgill: We could then apply a similar naming scheme to our plugin-specific rulesets:

```suggestion
<ruleset name="WPP-AutoSizes">
	<description>WordPress Coding Standards for Auto-sizes for Lazy-loaded Images Plugin</description>
```

composer.json
joemcgill, line 27: This will only result in running the `phpcs.xml.dist` file in the project root. This is ok, but we'll need to add another script to lint each plugin based on it's own ruleset.
```suggestion
    "lint": "build-cs/vendor/bin/phpcs",
    "lint:all": [
        "@lint",
        "@lint:auto-sizes"
    ],
    "lint:auto-sizes": "@lint -- ./plugins/auto-sizes --standard=./plugins/auto-sizes/phpcs.xml.dist"
```

phpcs.xml.dist
```diff
@@ -1,108 +1,23 @@
 <?xml version="1.0"?>
 <ruleset name="WordPress Coding Standards for Performance Plugin">
```
joemcgill:
```suggestion
<ruleset name="WPP-PerfLab">
	<description>WordPress Coding Standards for the Performance Lab Plugin</description>
```

joemcgill (changes requested):
I like where this is headed, though I think we will need to update the various ways that we run PHPCS via local scripts to make sure we're running the correct ruleset for each plugin. I left a suggestion, but I'm not totally sold that this is the most maintainable approach.

composer.json
westonruter: Related to https://github.com/WordPress/performance/pull/995 which also removes the explicit `phpcs.xml.dist` reference in `trunk`.

swissspidy: I assume PHPCS can‘t automatically detect various config files and run them all separately?

westonruter: Right. It only looks at parent directories to find a config file: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#using-a-default-configuration-file

It doesn't look into subdirectories for potential config overrides, as far as I know.

thelovekesh: Correct, and AFAIK most tools do the same(while they can go up to the root directory to look for possible base config). They take reference to the current directory where the process is being run.

phpcs.xml.dist
mukeshpanchal27: Added exclude rules for two files in https://github.com/WordPress/performance/pull/1002/commits/5edaedf973cbcb84c953c685bac443539535dffa

composer.json
mukeshpanchal27: Thanks for all the feedback. The command has been added in https://github.com/WordPress/performance/pull/1002/commits/6b03ad06b053a637fea15d46c3570def70c8c2a9 and it works as expected. You can see the results here: https://github.com/WordPress/performance/actions/runs/8043866559/job/21966609437?pr=1002#step:9:9

swissspidy (approved):

phpcs.xml.dist
```diff
@@ -1,108 +1,30 @@
...
-				<element value="parentNode"/>
-				<element value="createTextNode"/>
-				<element value="textContent"/>
-			</property>
-		</properties>
+	<!-- Exclude specific files from the TextDomainMismatch sniff -->
+	<rule ref="WordPress.WP.I18n.TextDomainMismatch">
+		<exclude-pattern>admin/load.php</exclude-pattern>
+		<exclude-pattern>admin/plugins.php</exclude-pattern>
 	</rule>
```
westonruter: IMO, it would be better to either add `phpcs:disable WordPress.WP.I18n`/`phpcs:enable WordPress.WP.I18n`/`phpcs:ignore WordPress.WP.I18n` comments in the files.

The `admin/load.php` file has 35 instances of the `performance-lab` text domain, and 5 instances of the `default` text domain. Here it would seem `phpcs:ignore` for each 5 instances would make sense.

For `admin/plugins.php`, there are 30 instances of `default` and only 3 for `performance-lab`, so if anything this would be the file to exclude entirely. The instances of `default` seem to all be inside of the `perflab_render_plugin_card()` function, so it could be put into a separate file like `admin/plugin-card-rendering.php` to isolate the `default` domain.

But this is not a blocker for merging this.

westonruter (approved):

thelovekesh (approved):

phpcs.xml.dist
```diff
@@ -1,108 +1,30 @@
...
-	<file>./load.php</file>
-	<file>./modules</file>
-	<file>./plugins</file>
-	<file>./server-timing</file>
-	<file>./tests</file>
+<ruleset name="WPP-PerfLab">
+	<description>WordPress Coding Standards for the Performance Lab Plugin</description>
+	<rule ref="phpcs.ruleset.xml"/>
+	
+	<config name="text_domain" value="performance-lab"/>
```
felixarntz: Why not keep `default` in here as it has been so far? That's not related to a specific plugin, but intended for strings that are found exactly like that in WordPress core. Using that approach is totally reasonable within a plugin, so I think we should keep that text domain included here rather than working around it with exceptions.
```suggestion
	<config name="text_domain" value="performance-lab,default"/>
```

phpcs.xml.dist
felixarntz: I think it would be simpler and cleaner to keep allowing `default`, as we have so far. That's not prohibited by any means. See my other review feedback.

.gitattributes
```diff
@@ -23,6 +23,7 @@
 /package.json export-ignore
 /package-lock.json export-ignore
 /phpcs.xml.dist export-ignore
+/phpcs.ruleset.xml export-ignore
 /phpstan.neon.dist export-ignore
 /phpunit.xml.dist export-ignore
```
felixarntz, line 28: I think we'll also have to cater here for the individual `/plugins/**/phpcs.xml.dist` files now that they may exist per individual plugin.

If I'm not mistaken, without this the standalone plugin builds would include these files, but they should be omitted.

felixarntz (changes requested):
@mukeshpanchal27 This looks great to me overall. A few small points of feedback.

.gitattributes
westonruter: Right, although I think actually we should rather do this:

```diff
- /phpcs.xml.dist export-ignore
+ phpcs.xml.dist export-ignore
```

We should just ignore these PHPCS files anywhere in the repo from being exported.

phpcs.xml.dist
mukeshpanchal27: Added "default" text domain back in https://github.com/WordPress/performance/pull/1002/commits/4927e3c518cdc03ec91295d11632826361e22545

mukeshpanchal27:
@felixarntz @joemcgill I have addressed the feedback in the PR. It's ready for merge now. Could you please review it when you have a moment?

plugins/auto-sizes/phpcs.xml.dist
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<ruleset name="WPP-AutoSizes">
+	<description>WordPress Coding Standards for Auto-sizes for Lazy-loaded Images Plugin</description>
+
+	<rule ref="../../phpcs.ruleset.xml"/>
+
+	<config name="text_domain" value="auto-sizes"/>
```
felixarntz, line 7: Related to my previous feedback, we could add `default` as an allowed text domain for all plugins, but definitely not required as long as we don't use any core strings in them.

felixarntz (approved):

plugins/auto-sizes/phpcs.xml.dist
westonruter: I think we should keep this omitted unless we require core strings.

felixarntz:
@joemcgill Can you please take another look at this as you previously requested changes?

joemcgill (approved):
Thanks for the updates. I think this looks good now.

PR #1005, opened 2024-02-21
Added global documentation

### LINKED ISSUE #1003: Need to Add Global Documentation in plugins.php file
Need to add Global Documentation according to [PHP Documentation Standards](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/)  in below file.

**1. admin/plugins.php file**

keyurptl:
Added Global documentation according to [PHP Documentation Standards.](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/)

Fixes: https://github.com/WordPress/performance/issues/1003

### DIFF
--- a/admin/plugins.php
+++ b/admin/plugins.php
@@ -123,9 +123,7 @@ function perflab_render_plugins_ui() {
  * @link https://github.com/WordPress/wordpress-develop/blob/0b8ca16ea3bd9722bd1a38f8ab68901506b1a0e7/src/wp-admin/includes/class-wp-plugin-install-list-table.php#L467-L830
  *
  * @global string $page
- * 
- * 
- * 
+ *
  * @param array $plugin_data Plugin data from the WordPress.org API.
  */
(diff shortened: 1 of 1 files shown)


westonruter (approved):

mukeshpanchal27 (changes requested):
Thanks, @keyurptl, for the PR. We don't need this global variable as it will not be used on the page. Instead of adding the document, we should remove it from the code.

Please correct me if I missed anything. Thank you!

keyurptl:
> Thanks, @keyurptl, for the PR. We don't need this global variable as it will not be used on the page. Instead of adding the document, we should remove it from the code.
> 
> Please correct me if I missed anything. Thank you!

@mukeshpanchal27 Thanks for the feedback.

I have updated changes as per mentioned.

felixarntz (approved):
Thank you @keyurptl!

mukeshpanchal27 (approved):
Thanks @keyurptl for the updates. LGTM!

PR #1008, opened 2024-02-22
Update plugin building 

### LINKED ISSUE #893: Add JS build step to utilize local copy of web-vitals.js

thelovekesh:
## Summary

Fixes #893

## Checklist


webpack.config.js
```diff
@@ -70,4 +91,85 @@ const webVitals = () => {
...
+	if ( 'plugin' !== env.build ) {
+		return {
+			entry: {},
+			output: {},
+		};
+	}
+
+	const from = path.resolve( __dirname );
+	const to = path.resolve( __dirname, 'build' );
+	const pluginDistIncludes = [
```
westonruter: Isn't this list essentially the inverse of the current `.gitattributes`? Since this is an allowlist it will have to be manually updated. But then there is also a `pluginDistIgnores` which is essentially the same as the `.gitattributes`? Wouldn't this be simpler to just rather go with `.distignore` instead of the `.gitattributes`?

thelovekesh: Humm, I agree with the negation thing since it's replicating the behavior of `.gitattributes`. TBH I wasn't much aware of `.distignore` but after exploring it I understand it's just "another" ignore file but creating distribution builds.

thelovekesh: Since `CopyWebpackPlugin` uses globby which provides support for such ignore files, I have switched it to use `.distignore` which is pretty much neat.

westonruter (changes requested):
I think the scope of this PR needs to be pulled back because this branch is still working with modules and without the standalone plugins. In the `feature/modules-to-plugins` branch which will hopefully soon get merged, it has a build script for plugins already:

https://github.com/WordPress/performance/blob/36fe0dce276a81db16a7bd0dc13c453d0d207314/package.json#L26

See the [`build-plugins` command](https://github.com/WordPress/performance/blob/36fe0dce276a81db16a7bd0dc13c453d0d207314/bin/plugin/commands/build-plugins.js).

But then there's another PR https://github.com/WordPress/performance/pull/1000 which is branched off of that which goes further for supporting standalone plugins.

So I think all that should be done in the immediate term is to switch from `.gitattributes` to `.distignore`. But we should perhaps wait altogether until we we finally get the `feature/modules-to-plugins` branch merged into `trunk`, as otherwise it's likely there will be re-work needed or merge conflicts.

thelovekesh:
>  it has a build script for plugins already:

IMO instead of developing CLI scripts, which frequently need to be modified for new use cases, we should utilize a tool like Webpack, which is ideally equipped and made to handle such use cases. Also, it gives us the flexibility to consume WP Scripts configurations which includes better presets for such cases.

thelovekesh:
> https://github.com/WordPress/performance/pull/1024#issuecomment-1979221706

Pulling this conversation for bundling plugins here.

cc @felixarntz

package.json
```diff
@@ -29,6 +29,7 @@
     "readme": "./bin/plugin/cli.js readme",
     "translations": "./bin/plugin/cli.js translations",
     "build": "wp-scripts build",
+    "build-plugin": "wp-scripts build --env build=plugin",
     "build-plugins": "./bin/plugin/cli.js build-plugins",
```
felixarntz, line 33: This is confusing to me. What is the purpose of `build-plugin` and `build`? We already have `build-plugins` which is for the modules to be built into plugins (which can soon be removed). Do we really need `build-plugin` or can it be combined somehow? If not, let's use a different name that clarifies what this is for.

felixarntz (commented):
@thelovekesh I am unsure what is the purpose of this PR, it looks like it is mixing concerns a bit.

I know it started as a PR for #897, which makes sense given it's targeting the `feature/image-loading-optimization` branch.

I don't think we should use this PR as a solution for the problem of ignoring development files and assets for the standalone plugins. That should be in its own PR against `trunk`. If you already have something in mind that can fix that problem, it would be great if you could open a new PR against `trunk` so we can consider it separately. This work shouldn't be coupled to a specific module.

thelovekesh:
Yes @felixarntz. This PR is confusing from the global point of view since this was originally authored as a POC for the ILO plugin. Hence, I will raise a PR to solve this on the trunk and later ILO can rebase with trunk and thus it will be solved for all the WIP and upcoming plugins.

cc @westonruter

felixarntz:
@thelovekesh Sounds great, thanks!

PR #1010, opened 2024-02-23
Introduce ILO_URL_Metrics_Group and ILO_URL_Metrics_Group_Collection classes

westonruter:
## Summary

This was a sub-PR of #988. It has now been merged.

This implements these aspects of https://github.com/WordPress/performance/issues/933:

> * Introduce class to represent a grouped URL metrics collection by breakpoint (e.g. `ILO_Grouped_URL_Metric_Collection`).
>     * The class could receive a list (indexed array) of `ILO_URL_Metric` instances and then take care of grouping them accordingly. The grouping internally could just be a map of `breakpoint => ILO_URL_Metric[]`.
>     * The class should probably implement `Traversable`, so that it can be iterated just like an array (over the breakpoints).
>     * The class could have a method `unshift( ILO_URL_Metric $metric )` that replaces the `ilo_unshift_url_metrics()` function and integrates the given metric properly, while discarding excess metrics beyond the configured sample size at the same time.
>     * Some of the grouping "configuration" (e.g. breakpoints, sample size) may also be useful to provide via constructor.
> * The `ilo_group_url_metrics_by_breakpoint()` function could then use these two classes.
>     * It seems the module currently _always_ uses the breakpoints from `ilo_get_breakpoint_max_widths()` and the sample size from `ilo_get_url_metrics_breakpoint_sample_size()`. If so, we could probably avoid those function calls elsewhere and _only_ call them in `ilo_group_url_metrics_by_breakpoint()`. In other words, only the group class constructor would take them as parameters (for flexibility and testing), but the wrapper function to instantiate the group class wouldn't, it would just use the default ones, to avoid having to pass those parameters around in several places.
> * Any functions that currently expect a `$grouped_url_metrics` array should be updated to expect an instance of the new group class. If there are particular usages where another method on that class would be helpful, it could be added based on need.
> * Functions that currently expect a (non-grouped) `$url_metrics` array but then immediately group them without doing anything else to the original array should be modified to expect an instance of the new group class as well, avoiding another call to the grouping function. This also helps separating responsibilities.

## Relevant technical choices

* A new `ILO_Grouped_URL_Metrics` class is introduced which now has methods which were previously functions. URL metrics which are added to instances of this class are initially-grouped and kept grouped as new URL metrics are added. The method names are improved, for example `ilo_unshift_url_metrics` is now just `add`. The constructor takes a list of the URL metrics, the breakpoints, the sample size, and the freshness TTL.
* A new method `are_all_groups_populated` is added which internalizes the logic which was previously in `ilo_optimize_template_output_buffer()`. This eliminates the need to understand the internal data structure.
* Another new method `flatten` is introduced which returns a list of `ILO_URL_Metric` instances. This is used when obtaining the URL metrics to serialize into the post type, as well as internally when calling the `add` method.

## Checklist


modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
```diff
@@ -0,0 +1,295 @@
...
+	/**
+	 * Gets needed minimum viewport widths.
+	 *
+	 * @since n.e.x.t
+	 * @access private
+	 *
+	 * @param float $current_time Current time, defaults to `microtime(true)`.
+	 * @return array<int, array{int, bool}> Array of tuples mapping minimum viewport width to whether URL metric(s) are needed.
+	 */
+	public function get_needed_minimum_viewport_widths( float $current_time = null ): array {
```
felixarntz: It would be great to find a simpler name for this.

Some ideas:
* `get_incomplete_minimum_viewport_widths()`
* `get_incomplete_viewport_widths()`
* `get_incomplete_group_identifiers()` (which should be clear enough given that the groups are always by minimum viewport width)

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
```diff
@@ -0,0 +1,295 @@
...
+			}
+		);
+	}
+
+	/**
+	 * Checks whether all groups have URL metrics.
+	 *
+	 * @return bool Whether all groups have URL metrics.
+	 */
+	public function are_all_groups_populated(): bool {
```
felixarntz: Maybe the name could also be simplified. Maybe we could call it `is_populated()`?

Separately, I would recommend we add another method here called `is_complete()`: This would return true if all groups are not just populated but have enough metrics to satisfy the sample size. This could be used in at least one place below in favor of the more complicated current implementation of `get_needed_minimum_viewport_widths()`.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
```diff
@@ -0,0 +1,295 @@
...
+ * @since n.e.x.t
+ */
+
+/**
+ * URL metrics grouped by minimum viewport width for the provided breakpoints.
+ *
+ * @since n.e.x.t
+ * @access private
+ */
+final class ILO_Grouped_URL_Metrics {
```
felixarntz: Not a big issue, but I'd prefer to call this `ILO_Grouped_URL_Metrics_Collection` or `ILO_Grouped_URL_Metric_Collection`. Or the other way around, `ILO_URL_Metric_Group_Collection`.

In any case, the "collection" in the name avoids the idea of odd method names like `are_...()` and we could just use `is...()` instead, and it's a common naming pattern for this kind of class.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
```diff
@@ -0,0 +1,295 @@
...
+	 * Gets minimum viewport widths for the groups of URL metrics divided by the breakpoints.
+	 *
+	 * @return int[]
+	 */
+	public function get_minimum_viewport_widths(): array {
+		return array_keys( $this->groups );
+	}
+
+	/**
+	 * Unshifts a new URL metric, potentially pushing out older URL metrics when exceeding the sample size.
```
felixarntz: I forgot, is it actually relevant that this is "unshifted"? If so, we may still want to clarify the documentation here. Potentially say something like that the metric will be prepended to the correct group, and the last entry from that group will be discarded if the sample size is exceeded.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	/**
+	 * Gets needed minimum viewport widths.
+	 *
+	 * @since n.e.x.t
+	 * @access private
+	 *
+	 * @param float $current_time Current time, defaults to `microtime(true)`.
+	 * @return array<int, array{int, bool}> Array of tuples mapping minimum viewport width to whether URL metric(s) are needed.
```
felixarntz: I think a simpler API, which would also be better in line with the method name, would be to return a simple indexed array of the minimum viewport widths.

I'm not sure there is a need for this to return all minimum viewport widths with a boolean indicator.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
```diff
@@ -0,0 +1,295 @@
...
+	 * a supported LCP element.) If there is a supported LCP element at the breakpoint, then the array value is an array
+	 * representing that element, including its breadcrumbs. If two adjoining breakpoints have the same value, then the
+	 * latter is dropped.
+	 *
+	 * @since n.e.x.t
+	 * @access private
+	 *
+	 * @return array LCP elements keyed by its minimum viewport width. If there is no supported LCP element at a breakpoint, then `false` is used.
+	 */
+	public function get_lcp_elements_by_minimum_viewport_widths(): array {
```
felixarntz: I would prefer to leave this method out of the class and keep as a standalone function for now. It is more specific than all other methods in the class in that it is tied to LCP related logic, whereas the rest of this class is more generic and reusable.

modules/images/image-loading-optimization/optimization.php
```diff
@@ -149,16 +149,15 @@ function ilo_optimize_template_output_buffer( string $buffer ): string {
...
-		$url_metrics,
-		microtime( true ),
+	$grouped_url_metrics = new ILO_Grouped_URL_Metrics(
+		$post ? ilo_parse_stored_url_metrics( $post ) : array(),
 		ilo_get_breakpoint_max_widths(),
 		ilo_get_url_metrics_breakpoint_sample_size(),
 		ilo_get_url_metric_freshness_ttl()
 	);
 
+	$needed_minimum_viewport_widths = $grouped_url_metrics->get_needed_minimum_viewport_widths( microtime( true ) );
```
felixarntz: See my above suggestion of an `is_complete()` method. That could be used here for the `$needs_detection` variable below, replacing the slightly complex logic there and simplifying things. As in:
`$needs_detection = ! $grouped_url_metrics->is_complete()`

modules/images/image-loading-optimization/storage/rest-api.php
```diff
@@ -113,14 +113,15 @@ function ilo_register_endpoint() {
...
-	$needed_minimum_viewport_widths = ilo_get_needed_minimum_viewport_widths(
+	$grouped_url_metrics = new ILO_Grouped_URL_Metrics(
 		$post ? ilo_parse_stored_url_metrics( $post ) : array(),
-		microtime( true ),
 		ilo_get_breakpoint_max_widths(),
 		ilo_get_url_metrics_breakpoint_sample_size(),
 		ilo_get_url_metric_freshness_ttl()
 	);
 
+	$needed_minimum_viewport_widths = $grouped_url_metrics->get_needed_minimum_viewport_widths( microtime( true ) );
```
felixarntz: Even here I'm not sure this method is the right solution. Maybe the `foreach` loop below could be avoided if instead we just had a method like `is_group_complete( $minimum_viewport_width )`. Then that could be used here instead, as far as I can tell.

felixarntz (commented):
@westonruter Overall this looks good. I think there are a few things we can tidy up and simplify in terms of the methods the class exposes.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: I forgot to update the phpdoc when I renamed the method from `unshift` to `add`. The fact that it does an unshift is a non-important implementation detail which should be hidden.

```suggestion
	 * Adds a new URL metric to a group.
	 *
	 * Once a group reaches the sample size, the oldest URL metric is pushed out.
```

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: It took me longer than I wished to prove this out but I believe I've finally done so. In db8ae4a6a0802b2fc964fc4f21bfa3be6200e6fd I added a test which:

1. Establishes a single breakpoint for which there are two viewport groups.
2. Fully populates the group for the wider viewports and leaves the group empty for narrower viewports.
3. Attempts to over-populate the wider viewport group with yet another URL metric.

When switching to only return the simple indexed array of the minimum viewport widths needing URL metrics (as I did in 794061d10183286ec7b9041f5a2e54cc8a4d37af), the REST API endpoint [fails](https://github.com/WordPress/performance/actions/runs/8103921850/job/22149500014) to properly block attempts to overpopulate the wider viewport group. The reason is that the narrower viewport group is encountered first when iterating which then sets `$last_was_needed` to `true`:

https://github.com/WordPress/performance/blob/794061d10183286ec7b9041f5a2e54cc8a4d37af/modules/images/image-loading-optimization/storage/rest-api.php#L128-L137

It never even considers the wider viewport group since it isn't returned by `get_needed_minimum_viewport_widths()`. Contrast this with the previous working code:

https://github.com/WordPress/performance/blob/ceeab6ec00024a1bf87e57da164e6284b67fd1ca/modules/images/image-loading-optimization/storage/rest-api.php#L128-L137

Note that in this case it iterates over _all_ viewport widths ensuring that the `$last_was_needed` can be switched back to `false` to block a submission.

felixarntz: Did you see my comment https://github.com/WordPress/performance/pull/1010#discussion_r1506883862? Wouldn't something like that address the problem in a more elegant way? We don't need to get a list that includes all viewports to find out whether the viewport we already know (as specified in the request) is full.

Looking at my previous suggestion again, I think I missed one thing though, the minimum viewport width is of course not necessarily the exact viewport width provided. So we need a method that, for a given viewport width, returns the appropriate breakpoint / minimum viewport width that identifies the relevant group. Still, I think we could simplify this via a method in the class that checks for a given viewport width whether its information is needed.

For even more separation of concerns, this could work as follows:
```php
/*
 * Get the correct breakpoint / minimum viewport width that the given viewport width belongs to.
 * For example, for breakpoints 0, 600, and 1200:
 * - 599 would return 0
 * - 1199 would return 600
 * - 1200 would return 1200
 * - 8000 would return 1200
 */
$minimum_viewport_width = $grouped_url_metrics->get_minimum_viewport_width( $viewport_width );

/*
 * If the group for that breakpoint / minimum viewport width is already complete/full
 * (i.e. as large as the sample size), fail with an error.
 */
if ( $grouped_url_metrics->is_group_complete( $minimum_viewport_width ) ) {
	return new WP_Error( '...' );
}
```

This would abstract away the logic so we wouldn't need to have a complicated `foreach` in here.

felixarntz: By the way, I wonder whether in terms of naming we should use "breakpoint" instead of "minimum viewport width". They are not exactly the same technically, but "breakpoint" is just a lot more concise for all those method and variable names. Plus no confusion with a viewport (i.e. a concrete viewport that isn't just the "minimum viewport" for a group).

westonruter: Yes, the logic can be encapsulated into a method to avoid having to iterate over the values in the REST API handler. I wanted to first sanity check why `get_needed_minimum_viewport_widths()` can't just return an array of integers for the viewport widths that actually need URL metrics.

Regarding the naming to use breakpoint, it can get a bit tricky because the breakpoints are the division between the groups, whereas the minimum viewport widths correspond to the number of groups themselves. The breakpoint count is always one less than the number of minimum viewport widths. 

Let me think about how this method might be named a bit more.

westonruter: I've introduced an `is_group_filled` method in b7d0819f.

felixarntz: Regarding "breakpoint" vs "minimum viewport width", the difference you've explained makes sense. I personally still feel using the term "breakpoint" is fine and is just more approachable from a code perspective. While `0` isn't really a "breakpoint", we're talking about "grouping by breakpoint" in a few places, so every group represents a specific breakpoint. So from that perspective I'd also be inclined to go with the much shorter "breakpoint" terminology.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: OK. Moved back in bdeaaccd.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: I'd rather not as it would become inconsistent terminology. Breakpoints are the end-user API which are exposed via the `ilo_breakpoint_max_widths()` filter. Breakpoints get used internally as "minimum viewport widths" (which is not end-user friendly), but these are private APIs not intended for end users.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: I've renamed it to `is_every_group_populated()` in 3523e98b. I think this is more clear than something shorter like `is_populated()` because that could mean is _any_ group populated.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: I think `needed` is fine because it's relevant not only if the group doesn't have the sample size but also if any of the URL metrics in the group are stale.

modules/images/image-loading-optimization/optimization.php
westonruter: Nevertheless, the `$needed_minimum_viewport_widths` needs to be passed to below to JavaScript so the client knows whether or not it should try to send a URL metric:

https://github.com/WordPress/performance/blob/4d8baaa822aadc623d0cf222361e2f887e7c3ac3/modules/images/image-loading-optimization/optimization.php#L315-L319

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: I feel inclined to keep it as `ILO_Grouped_URL_Metrics`. I feel adding `Collection` on the end makes it redundant. I solved the issue with the `are_` method name.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
felixarntz: Not sure I understand. Are you saying so far breakpoints are the "maximum widths" rather than the "minimum widths of the next larger group"?

Alternatively, how about "breakpoint groups" as terminology? Or "group IDs"? I just want to avoid the excessively long "minimum viewport widths".

Separately, WDYT about https://github.com/WordPress/performance/pull/1010#discussion_r1508340377? Looking at https://github.com/WordPress/performance/pull/1010/commits/b7d0819f3d2fe9d5fe9c0056caa8f0c1766e13ca#diff-4369a8ab8272ee7760fdc9ce4442c633c520114a2d593c859b7807e7ddb806dcR163, I still don't understand why we have to call that method here. IMO it would be more intuitive to have one method to get the correct "breakpoint group ID" or "minimum viewport width" for the group that the given viewport width falls into, and another method to check whether that group is filled.

westonruter: Breakpoints are the maximum widths before entering the next group. Given 3 breakpoints 480, 600, and 782 then there would be 4 groups for the following viewport ranges:

* 0-480
* 481-600
* 601-782
* 782-∞

So there is always one more group than there are breakpoints. I organized around "minimum viewport widths" since it makes sense to anchor off of the zero value, rather than to work from the other end at infinity (or `PHP_INT_MAX`). I don't have a problem with "minimum viewport widths" as a name, as it is much more explicit at what it is talking about. Another name like "group ID" or "breakpoint group" is more general and you'd have to remind yourself what it is referring to.

https://github.com/WordPress/performance/blob/3523e98b62ce2627502f1b75d28c5de1cee48023/modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php#L132-L140

> Separately, WDYT about [#1010 (comment)](https://github.com/WordPress/performance/pull/1010#discussion_r1508340377)? Looking at [b7d0819#diff-4369a8ab8272ee7760fdc9ce4442c633c520114a2d593c859b7807e7ddb806dcR163](https://github.com/WordPress/performance/commit/b7d0819f3d2fe9d5fe9c0056caa8f0c1766e13ca#diff-4369a8ab8272ee7760fdc9ce4442c633c520114a2d593c859b7807e7ddb806dcR163), I still don't understand why we have to call that method here. IMO it would be more intuitive to have one method to get the correct "breakpoint group ID" or "minimum viewport width" for the group that the given viewport width falls into, and another method to check whether that group is filled.

It could be refactored to access the data in a different way specifically for the purposes of how the data is needed by the REST API here, to specifically look up one single viewport width to see if its group is filled or not. However, the data is needed as well in `ilo_optimize_template_output_buffer()` in two places:

https://github.com/WordPress/performance/blob/3523e98b62ce2627502f1b75d28c5de1cee48023/modules/images/image-loading-optimization/optimization.php#L159-L167

https://github.com/WordPress/performance/blob/3523e98b62ce2627502f1b75d28c5de1cee48023/modules/images/image-loading-optimization/optimization.php#L317-L319

In the first, it needs to look at all the groups together to see if _any_ of them need URL metrics. This is because at this point we don't know the user's viewport width. So it doesn't make sense here to look up a specific group since there is no specific viewport yet known.

In the second instance, we export the data to `detect.js` so that the client logic (which _does_ know the viewport width) can decide whether or not to send the URL metrics to the server.

So in both of these cases, we need the full data returned by `get_needed_minimum_viewport_widths` and this is re-used in the REST API endpoint via this new `is_group_filled` method.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: I spent some time thinking about this some more. I just pushed up some more changes. Now instead of `get_needed_minimum_viewport_widths()` the method is called `get_viewport_group_lacking_statuses()`. Not any less verbose. But maybe more clear?

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: > So in both of these cases, we need the full data returned by `get_needed_minimum_viewport_widths` and this is re-used in the REST API endpoint via this new `is_group_filled` method.

With the latest changes, these methods are now called `get_viewport_group_lacking_statuses` and `is_viewport_group_lacking`, respectively.

westonruter: As I've been commenting above, I've been working on refactoring this further. I've introduced a `ILO_URL_Metrics_Group_Status` which contains both the `minimum_viewport_width` and the `is_lacking` data. This object is then passed around which I think makes it more clear and readable.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
```diff
@@ -0,0 +1,237 @@
...
+	 * Adds a new URL metric to a group.
+	 *
+	 * Once a group reaches the sample size, the oldest URL metric is pushed out.
+	 *
+	 * @since n.e.x.t
+	 * @access private
+	 *
+	 * @param ILO_URL_Metric $new_url_metric New URL metric.
+	 */
+	public function add( ILO_URL_Metric $new_url_metric ) {
```
felixarntz: Maybe better to name it `add_metric()`, to clarify? Otherwise it would be unclear whether this is for adding a metric or adding a group or adding a viewport, for instance.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
```diff
@@ -0,0 +1,237 @@
...
+	 * @since n.e.x.t
+	 * @access private
+	 *
+	 * @param ILO_URL_Metric[] $url_metrics URL metrics.
+	 * @return array<int, ILO_URL_Metric[]> URL metrics grouped by breakpoint. The array keys are the minimum widths for a viewport to lie within
+	 *                                      the breakpoint. The returned array is always one larger than the provided array of breakpoints, since
+	 *                                      the breakpoints reflect the max inclusive boundaries whereas the return value is the groups of page
+	 *                                      metrics with viewports on either side of the breakpoint boundaries.
+	 */
+	private function group_url_metrics_by_breakpoint( array $url_metrics ): array {
```
felixarntz: In relation to the other naming discussions, maybe better as `group_url_metrics_by_viewport()`? As you previously mentioned, `0` for instance isn't really a breakpoint. But every group represents a viewport.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
```diff
@@ -0,0 +1,237 @@
...
+	public function is_group_lacking( int $viewport_width ): bool {
+		$last_was_lacking = false;
+		foreach ( $this->get_group_statuses() as $status ) {
+			if ( $viewport_width >= $status->get_minimum_viewport_width() ) {
+				$last_was_lacking = $status->is_lacking();
+			} else {
+				break;
+			}
+		}
+		return $last_was_lacking;
```
felixarntz: I still don't understand the need for this logic. Why do we need to cycle through multiple groups?

The breakpoints define the minimum viewport of a group, and the maximum viewport of the previous group is that breakpoint minus 1. So for instance if `$viewport_width` is `500` and there are groups `0-399` and `400-699`, that viewport falls in the second group, and in the second group only. I don't think we should use larger viewport widths to populate data for smaller viewport groups.

Based on that, I would reiterate my previous proposal of implementing:
* one dedicated method to identify the correct group for an arbitrary viewport width (could e.g. return the minimum viewport width of that group, or maybe we want to introduce a dedicated class representing a group?)
* this method, which would only call the before method and then check if that group is not full yet

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
```diff
@@ -0,0 +1,237 @@
...
+	/**
+	 * Determines whether the URL metrics group for a given viewport is lacking URL metrics.
+	 *
+	 * Either the group does not have enough URL metrics for the desired sample size,
+	 * or some of the URL metrics are stale.
+	 *
+	 * @param int $viewport_width Viewport width.
+	 * @return bool Whether group is lacking.
+	 */
+	public function is_group_lacking( int $viewport_width ): bool {
```
felixarntz: Why not name this the opposite to make it a positive term? It's cognitively a bit easier to work with such logic rather than negative naming (think negating negatives is more complex).

We already have the term "populated" below as describing that a group is not empty.
I would suggest we use the term "full" to describe a group has enough entries to satisfy the sample size.

In other words, I would suggest to name this `is_group_full()`, and obviously return the reverse boolean.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
```diff
@@ -0,0 +1,237 @@
...
+						$is_lacking = true;
+						break;
+					}
+				}
+			}
+			$statuses[] = new ILO_URL_Metrics_Group_Status( $minimum_viewport_width, $is_lacking );
+		}
+
+		return $statuses;
+	}
```
felixarntz: See my above comments:
* Why does a group status not also include its upper (maximum) viewport width?
* I think rather than a more specific "URL metrics group status" class, we should have a "URL metrics group" class.
    * We already have "URL metric" and "grouped URL metrics", so IMO "URL metrics group" is the missing piece between the two. A class for their status would be more specific, and IMO is unnecessary as it could be implemented as part of the group class.
    * We could then create those instances as part of the `group_url_metrics_by_breakpoint()` method, and the `get_groups()` method would be the logical place to return those objects from - or maybe another one, if we wanted to keep a method in place to return _just_ the minimum viewport widths.

Based on that, I think this method could be replaced with something simpler. That said, we should probably first figure out the problem of not considering maximum viewport widths here - maybe I'm missing something but I don't see why those aren't being considered.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
```diff
@@ -0,0 +1,237 @@
...
+
+	/**
+	 * Checks whether every group is populated with at least one URL metric each.
+	 *
+	 * They aren't necessarily filled to the sample size, however.
+	 *
+	 * @return bool Whether all groups have URL metrics.
+	 */
+	public function is_every_group_populated(): bool {
```
felixarntz: Per my earlier comment, it would be useful to also have a similar method to see whether every group is "full". This would signify that for sure no tracking is needed for the URL. I think a `is_every_group_full()` next to `is_every_group_populated()` and the more specific `is_group_full( $group )` (see above) would make a lot of sense.

modules/images/image-loading-optimization/class-ilo-url-metrics-group-status.php
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @phpstan-type Data array{
+ *                        minimumViewportWidth: int,
+ *                        isLacking: bool,
+ *                    }
+ *
+ * @since n.e.x.t
+ * @access private
+ */
+final class ILO_URL_Metrics_Group_Status implements JsonSerializable {
```
felixarntz: See above, I think this class is too specific. I don't see the need for it, but if this information needs to be in a dedicated class per group, I think it would make more sense in a general "URL metric group" class.

modules/images/image-loading-optimization/optimization.php
```diff
@@ -149,28 +149,26 @@ function ilo_optimize_template_output_buffer( string $buffer ): string {
...
-		array_column( $needed_minimum_viewport_widths, 1 ),
-		true
-	);
+	$needs_detection = false;
+	foreach ( $group_statuses as $group_status ) {
+		if ( $group_status->is_lacking() ) {
+			$needs_detection = true;
+			break;
+		}
+	}
```
felixarntz: See above: Instead of this logic, why not use a class method like `is_every_group_full()` to simplify this?
```suggestion
	$needs_detection = ! $grouped_url_metrics()->is_every_group_full();
```

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
felixarntz: @westonruter 
> So in both of these cases, we need the full data returned by `get_needed_minimum_viewport_widths` and this is re-used in the REST API endpoint via this new `is_group_filled` method.

The full list is really only needed in the `ilo_get_detection_script()` call. For the `$needs_detection` check a more intuitive solution would be to just call a method on the class, see https://github.com/WordPress/performance/pull/1010#discussion_r1511727457.

> With the latest changes, these methods are now called `get_viewport_group_lacking_statuses` and `is_viewport_group_lacking`, respectively.

Why? See https://github.com/WordPress/performance/pull/1010#discussion_r1511718561, a positive term would be more intuitive. Whether it's "full" or "filled" doesn't really matter from my perspective, but either is IMO easier to understand and shorter. Negative method names are more complicated, particularly as they may be used with negation.

felixarntz: > As I've been commenting above, I've been working on refactoring this further. I've introduced a `ILO_URL_Metrics_Group_Status` which contains both the `minimum_viewport_width` and the `is_lacking` data. This object is then passed around which I think makes it more clear and readable.

See https://github.com/WordPress/performance/pull/1010#discussion_r1511725755 for basically a response to this.

modules/images/image-loading-optimization/optimization.php
felixarntz: Resolving this in favor of the similar conversation as part of https://github.com/WordPress/performance/pull/1010#discussion_r1511735762

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: Updated in 2b0ede3f0086a2e6d0c41f37ce981b34fa3efdca

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: There are three possible statuses for a URL metrics group:

1. It doesn't have enough URL metrics for the sample size, and/or
2. Some of the URL metrics are stale and need to be refreshed. Or
3. There is a complete number of URL metrics for the desired sample size and they are all fresh (non-stale).

I was thinking of the `lacking` term since it can encapsulate both the first and second statuses. Terms like "full" or "complete" could be misleading since they might describe either status 2 or 3 when the full sample size is present.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: +1. Coincidentally, after I put that `ILO_URL_Metrics_Group_Status` class together on Friday, over the weekend I was ruminating on it a bit  more and also came to the conclusion that it would make more sense to have an `ILO_URL_Metrics_Group` class which exposes the status as an aspect. I didn't get a chance to follow up until now.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
felixarntz: Thanks for outlining the three states, that makes sense.

I'm not sure about the conclusion though. To me `lacking` is not more clear than `full` or `complete`. I agree `full` may be more limited in meaning, but IMO `complete` would work perfectly well. When you read `lacking`, you may just as much think that the group is lacking items because it doesn't have a sufficient amount, so not really different than the opposite of `complete`.

I think it doesn't really matter though _why_ a group is not `complete`. That's an internal detail that the class itself solves and the caller doesn't need to bother in the first place. That's why from a semantic standpoint, I think `lacking` and `complete` are equally good or bad. However, I think `complete` works better as it is a positive term.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: I went ahead and renamed it to `ILO_URL_Metrics_Group_Collection` in eb864da9.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: Inverted `lacking` to `complete` in da7d2ad0526fddbcb775d1ebca27d7ec67a0fb03.

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: Added the `is_every_group_complete()` method in 28ee2b8c35f17ed6b837f99bf0647fca533304cb.

modules/images/image-loading-optimization/optimization.php
westonruter: See 28ee2b8

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: There is now a `ILO_URL_Metrics_Group_Collection::get_group_for_viewport_width()` method which returns a `ILO_URL_Metrics_Group` instance. The `is_complete()` method on this group is then now called in `rest-api.php`:

https://github.com/WordPress/performance/blob/fadd66edf0aa6dba5e0a88e76a1e600e45676c87/modules/images/image-loading-optimization/storage/rest-api.php#L124-L131

modules/images/image-loading-optimization/class-ilo-grouped-url-metrics.php
westonruter: This has been refactored as follows:

https://github.com/WordPress/performance/blob/55606ace9b614d5e7c9ca128dc26a3e31c3df482/modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php#L115-L119

Based on the breakpoints alone, empty groups are constructed initially:

https://github.com/WordPress/performance/blob/55606ace9b614d5e7c9ca128dc26a3e31c3df482/modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php#L127-L136

And then they are populated with the URL metrics, if there are any.

modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
```diff
@@ -0,0 +1,241 @@
...
+	 */
+	private $groups;
+
+	/**
+	 * Breakpoints in max widths.
+	 *
+	 * @var int[]
+	 * @phpstan-var int<1, max>[]
+	 */
+	private $breakpoints;
```
felixarntz: So this is for instance 480, 800, and ... (what's used for the maximum?), which would mean the groups are:
* 0-480
* 481-800
* 800-infinite

Do I understand that correctly?

modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
```diff
@@ -0,0 +1,241 @@
...
+	 * @param int[]            $breakpoints   Breakpoints in max widths.
+	 * @param int              $sample_size   Sample size for the maximum number of viewports in a group between breakpoints.
+	 * @param int              $freshness_ttl Freshness age (TTL) for a given URL metric.
+	 */
+	public function __construct( array $url_metrics, array $breakpoints, int $sample_size, int $freshness_ttl ) {
+		// Set breakpoints.
+		sort( $breakpoints );
+		$breakpoints = array_values( array_unique( $breakpoints, SORT_NUMERIC ) );
+		foreach ( $breakpoints as $breakpoint ) {
+			if ( $breakpoint <= 0 ) {
```
felixarntz: See above, how is the maximum / infinite value handled? The last breakpoint group would not really have a maximum. Does that value need to be provided or is it handled automatically? I just wonder whether additional validation is needed.

modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
```diff
@@ -0,0 +1,241 @@
...
+	 * @return ILO_URL_Metrics_Group[] Groups.
+	 */
+	private function create_groups(): array {
+		$groups    = array();
+		$min_width = 0;
+		foreach ( $this->breakpoints as $max_width ) {
+			$groups[]  = new ILO_URL_Metrics_Group( array(), $min_width, $max_width, $this->sample_size, $this->freshness_ttl );
+			$min_width = $max_width + 1;
+		}
+		$groups[] = new ILO_URL_Metrics_Group( array(), $min_width, PHP_INT_MAX, $this->sample_size, $this->freshness_ttl );
```
felixarntz, line 138: I see. Just thinking out loud, would it be more straightforward to just add `PHP_INT_MAX` to `$breakpoints` in the constructor instead?

modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
```diff
@@ -0,0 +1,241 @@
...
+	 * Gets grouped keyed by the minimum viewport width.
+	 *
+	 * @return ILO_URL_Metrics_Group[] Groups.
+	 */
+	public function get_groups(): array {
+		return $this->groups;
+	}
+
+	/**
+	 * Get group for viewport width.
```
felixarntz:
```suggestion
	 * Gets group for viewport width.
```

modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
```diff
@@ -0,0 +1,241 @@
...
+		}
+		return true;
+	}
+
+	/**
+	 * Gets URL metrics from all groups merged into one list.
+	 *
+	 * @return ILO_URL_Metric[] All URL metrics.
+	 */
+	public function get_merged_url_metrics(): array {
```
felixarntz: "merged" to me is a bit of a technicality. Maybe better "flattened", or simply "all"?

modules/images/image-loading-optimization/class-ilo-url-metrics-group.php
```diff
@@ -0,0 +1,172 @@
...
+	/**
+	 * Adds a URL metric to the group if it is in the bounds of the max and min viewport width.
+	 *
+	 * @param ILO_URL_Metric $url_metric URL metric.
+	 * @return bool Whether the URL metric was added.
+	 */
+	public function add_url_metric( ILO_URL_Metric $url_metric ): bool {
+		if ( ! $this->is_viewport_width_in_range( $url_metric->get_viewport()['width'] ) ) {
+			return false;
+		}
```
felixarntz: Rather than checking this here and returning a boolean, a more explicit API would be to check whether `is_viewport_width_in_range` in the `URL_Metrics_Group_Collection` class _before_ calling `add_url_metric`, with the method here throwing an exception if the given metric doesn't fit in the group.

This prevents mistakes in a way the developer will for sure find out about it. Returning a success/failure value from a method that does something other than check for a specific thing isn't very intuitive and often may go ignored. Think about all the `register_post_type()` calls in plugins and how uncommon it is to check for the return value - even though there is one.

modules/images/image-loading-optimization/class-ilo-url-metrics-group.php
```diff
@@ -0,0 +1,172 @@
...
+	public function get_url_metrics(): array {
+		return $this->url_metrics;
+	}
+
+	/**
+	 * Counts the URL metrics in the group.
+	 *
+	 * @return int Count of URL metrics in the group.
+	 */
+	public function count(): int {
```
felixarntz, line 215: I recall there being some recommendation not to use certain method names for "reserved" PHP terms like `count`. I can't pinpoint it, but in any case it may be safer to go with another name. How about `get_count()` or `get_size()` or just `size()`?

Alternatively, you could implement the `Countable` interface to be able to call `count( $url_metrics_group )` on the class instance itself. Not a must have, but worth considering, alongside other array interfaces like `Traversable`, which would allow doing `foreach ( $url_metrics_group as $url_metric )` on the class instance.

modules/images/image-loading-optimization/detection.php
```diff
@@ -39,16 +39,24 @@ function ilo_get_detection_script( string $slug, array $needed_minimum_viewport_
...
+		'urlMetricsNonce'         => ilo_get_url_metrics_storage_nonce( $slug ),
+		'urlMetricsGroupStatuses' => array_map(
+			static function ( ILO_URL_Metrics_Group $group ): array {
+				return array(
+					'minimumViewportWidth' => $group->get_minimum_viewport_width(),
+					'complete'             => $group->is_complete(),
+				);
+			},
+			$group_collection->get_groups()
+		),
```
felixarntz, line 57: +1, this puts this more specific logic only where needed without requiring the class based API to return that shape.

felixarntz (approved):
@westonruter Thanks for the iterations, this looks great!

A few minor comments for consideration, please take a look, mostly nit-picks or suggestions.

modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
westonruter: If two breakpoints were provided: 480 and 800, then there would be 3 groups with viewport ranges:

1. 0-600
2. 601-800
3. 801-∞

A breakpoint must be greater than zero, because it is not possible to have a viewport that is zero pixels wide. A breakpoint wouldn't in practice be the maximum because that would mean there could be a viewport at maximum + 1. In this way, the inclusion of `max` in the typing is not accurate. It should rather be `int<1, max-1>[]` but I don't think that's supported.

westonruter: Further improved typing in 34e3d062 to clarify the range of values allowed for breakpoints.

modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
westonruter: The last group has a maximum width value set `PHP_INT_MAX`. As such, an actual breakpoint wouldn't be able to have that value, since then it would try to create an even bigger breakpoint group which has a minimum viewport width of `PHP_INT_MAX + 1` but still a maximum viewport width of `PHP_INT_MAX` which would be smaller than the minimum viewport width.

modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
westonruter: I don't think so per https://github.com/WordPress/performance/pull/1010#discussion_r1515634477.

modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
westonruter: Went back to flattened in 49ec2c5b.

modules/images/image-loading-optimization/class-ilo-url-metrics-group.php
westonruter: I implemented `Countable` and `IteratorAggregate` in 0cce1326370817c2be9841c9a0a02213050dd3f7 and ebb74327169b0b0407b49fb729ee02c8342f2119.

modules/images/image-loading-optimization/class-ilo-url-metrics-group.php
westonruter: OK, I've implemented this in c362a6f42f7f9cb92bc969d93ff952dbf17c0c67. Nevertheless, a URL metric should always fit into a group contained in a collection because when the collection creates the groups it will ensure the groups span all non-negative integers.

modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
felixarntz: I guess I was thinking for the last group you wouldn't add the +1. That would mean every breakpoint represents a group, while right now there's always one less breakpoint than there are groups.

I guess both makes sense from different perspectives, so the current approach WFM.

modules/images/image-loading-optimization/class-ilo-url-metrics-group.php
felixarntz: Right. This is just to make a clean API, should someone try to do something custom with it. Plus for any potential future iterations within the plugin code itself.

PR #945, opened 2024-01-17
Auto-sizes: Disallow direct file access

joemcgill:
## Summary

Disallows direct file access.

## Relevant technical choices

This implements the common pattern of bailing early if `ABSPATH` is not defined, following the suggestion when this was submitted to the WP Plugin repo.

## Checklist

- [x] PR has either `[Focus]` or `Infrastructure` label.
- [x] PR has a `[Type]` label.
- [x] PR has a milestone or the `no milestone` label.

PR #1011, opened 2024-02-23
Move published modules to standalone plugins

### LINKED ISSUE #654: Remove the modules that were published as standalone plugins from the Performance Lab plugin

thelovekesh:
## Summary

- Move published modules to the `plugins` directory. Moved modules include:
    - `dominant-color-images`
    - `webp-uploads`
- Move modules test cases to `tests/plugins` directory.
- Add testsuite entry in the PHPUnit config.
- Update plugin files headers and doc blocks.
- Update `@since` and `@package` tags.

Fixes #654 

## Relevant technical choices

- Update `@since` tags based on the module changelog.
- Update `@package` tags based on the module name.
- Rename the main plugin file based on the plugin name.
- Remove checks included to return if "module" was already activated.
- Update plugin version as per `plugin.json` metadata.

## Checklist


.github/CODEOWNERS
```diff
@@ -49,11 +44,15 @@
...
-/tests/modules/images/dominant-color-images                                 @pbearne @spacedmonkey
-/tests/testdata/modules/images/dominant-color-images                        @pbearne @spacedmonkey
-
 # Plugin: Auto-sizes for Lazy-Loaded Images
 /plugins/auto-sizes                         @joemcgill
 /tests/plugins/auto-sizes                   @joemcgill
+
+# Plugin: WebP Uploads
+/plugins/webp-uploads                       @adamsilverstein @felixarntz
+/tests/plugins/webp-uploads                 @adamsilverstein @felixarntz
```
westonruter, line 57: What about `/tests/testdata/plugins/webp-uploads`? This was present above but it is missing here.

plugins.json
```diff
@@ -9,5 +9,9 @@
       "version": "1.0.5"
     }
   },
-  "plugins": [ "auto-sizes" ]
+  "plugins": [
+    "auto-sizes",
+    "dominant-color-images",
+    "webp-uploads"
```
westonruter: Versions will need to be added once https://github.com/WordPress/performance/pull/1000 is merged.

plugins/dominant-color-images/.gitattributes
westonruter, line 1: Shouldn't this `.gitattributes` file rather be deleted in favor of having this information in the root `.gitattributes`?

plugins/dominant-color-images/class-dominant-color-image-editor-imagick.php
westonruter, line 1: The text domains in this file (and maybe others) need to be updated from `performance-lab` to `dominant-color-images`. See https://plugins.trac.wordpress.org/browser/dominant-color-images/tags/1.0.1/class-dominant-color-image-editor-imagick.php#L64

plugins/webp-uploads/.gitattributes
westonruter, line 1: Ditto from above regarding this file.

plugins/webp-uploads/helper.php
```diff
@@ -2,14 +2,11 @@
...
- * @package performance-lab
+ * @package webp-uploads
+ *
  * @since 1.0.0
  */
 
-if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Exit if accessed directly.
-}
-
```
westonruter: Why is this deleted?

westonruter (changes requested):

.github/CODEOWNERS
thelovekesh: Moved it to `tests/plugins/webp-uploads/helpers` for being autoloaded by PHPUnit in https://github.com/WordPress/performance/pull/1011/commits/370a592e4fe2c54f65c05e0fed12d5ebb0fe9868

plugins/dominant-color-images/.gitattributes
thelovekesh: Yes, makes sense to me. Also, I don't see its benefit in the current setup since it's being added in the dist plugin.

plugins/dominant-color-images/class-dominant-color-image-editor-imagick.php
thelovekesh: Aah yes. I missed it. Thanks for pointing out.

plugins/webp-uploads/helper.php
thelovekesh: I am not sure if I missed any use cases. Since this file is just declaring functions and not invoking them, do we still need this safeguard?

phpcs.xml.dist
```diff
@@ -8,7 +8,7 @@
 	<rule ref="WordPress-Docs"/>
 	<rule ref="WordPress-Extra" />
 	<rule ref="WordPress.WP.I18n"/>
-	<config name="text_domain" value="performance-lab,default,auto-sizes"/>
+	<config name="text_domain" value="performance-lab,default,auto-sizes,dominant-color-images,webp-uploads"/>
```
thelovekesh: Note: Needs to be updated once https://github.com/WordPress/performance/pull/1002 is merged.

thelovekesh:
@mukeshpanchal27 JFYI recently added unit tests setup for standalone plugins doesn't seem to be running multisite test cases - https://github.com/WordPress/performance/actions/runs/8027561041/job/21931681466?pr=1011. I am aiming to amend this setup in https://github.com/WordPress/performance/pull/1013.

plugins/webp-uploads/helper.php
mukeshpanchal27: Yes.

mukeshpanchal27:
> @mukeshpanchal27 JFYI recently added unit tests setup for standalone plugins doesn't seem to be running multisite test cases - https://github.com/WordPress/performance/actions/runs/8027561041/job/21931681466?pr=1011. I am aiming to amend this setup in #1013.

Great catch. It was indeed missed. I've added the missing configuration in PR #1014.

default-enabled-modules.php
```diff
@@ -1,8 +1,6 @@
 <?php
 /* THIS IS A GENERATED FILE. DO NOT EDIT DIRECTLY. */
 return array(
-	'images/dominant-color-images',
```
mukeshpanchal27: Note: It will auto updated in release process. Check https://make.wordpress.org/performance/handbook/performance-lab/releasing-the-plugin/#update-list-of-default-enabled-modules

plugins/dominant-color-images/dominant-color-images.php
```diff
@@ -0,0 +1,27 @@
+<?php
```
mukeshpanchal27: The plugin released with [load.php](https://plugins.trac.wordpress.org/browser/dominant-color-images/trunk/load.php) we should rename this file back to `load.php`

plugins/webp-uploads/webp-uploads.php
```diff
@@ -0,0 +1,30 @@
+<?php
```
mukeshpanchal27: Check similar comment `dominant-color-images` plugin.

plugins/dominant-color-images/dominant-color-images.php
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Plugin Name: Dominant Color Images
+ * Plugin URI: https://github.com/WordPress/performance/tree/trunk/modules/images/dominant-color-images
```
mukeshpanchal27: This should links to `/plugins/` folder now.

plugins/webp-uploads/webp-uploads.php
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Plugin Name: WebP Uploads
+ * Plugin URI: https://github.com/WordPress/performance/tree/trunk/modules/images/webp-uploads
```
mukeshpanchal27: This should links to `/plugins/` folder now.

.github/CODEOWNERS
```diff
@@ -49,11 +44,15 @@
...
+
+# Plugin: Dominant Color Images
+/plugins/dominant-color-images                                              @pbearne @spacedmonkey
+/tests/plugins/dominant-color-images                                        @pbearne @spacedmonkey
+/tests/testdata/modules/images/dominant-color-images                        @pbearne @spacedmonkey
```
mukeshpanchal27: Per https://github.com/WordPress/performance/pull/1011/files#r1501248677 do we need to move this into helpers?

mukeshpanchal27 (changes requested):
Thanks @thelovekesh for the PR. This PR is blocked by #1000 and #1002. Left some initial feedbacks.

.github/CODEOWNERS
thelovekesh: Hmm, yes, because it would be better to keep things related to a standalone plugin in a single place.

plugins/webp-uploads/helper.php
thelovekesh: @mukeshpanchal27 Not sure I understand. Can you please share some info here?

.github/CODEOWNERS
thelovekesh: done in https://github.com/WordPress/performance/pull/1011/commits/69b6275b08602d0b7fff6e5f7886a145c17a700a

plugins/webp-uploads/helper.php
westonruter: As I recall, the plugin review team requested this short-circuit logic. But I don't think it is in the handbook, so it may not be a hard requirement. I think @joemcgill got that feedback.

westonruter (approved):
LGTM

felixarntz (changes requested):
@thelovekesh This looks good so far.

I'm putting a "Request changes" for now, just so that we don't merge it yet until the foundation has been completed. I'm worried that if we merge this already while the foundation is still being worked on in other PRs (e.g. #1002 that you highlighted) we'll run into merge conflicts and things may become too confusing. Let's merge the removal/migration of the modules as the last step.

thelovekesh:
@felixarntz @mukeshpanchal27 Since #1000 and #1002 are now merged, we can pick it up next?

mukeshpanchal27:
@thelovekesh, could you please rebase the PR as it seems to have conflicts?

thelovekesh:
@mukeshpanchal27 Resolved the conflicts, Can you please take a look now?

plugins/dominant-color-images/load.php
```diff
@@ -0,0 +1,27 @@
+<?php
+/**
+ * Plugin Name: Dominant Color Images
+ * Plugin URI: https://github.com/WordPress/performance/tree/trunk/plugins/images/dominant-color-images
```
mukeshpanchal27:
```suggestion
 * Plugin URI: https://github.com/WordPress/performance/tree/trunk/plugins/dominant-color-images
```

plugins/webp-uploads/load.php
```diff
@@ -0,0 +1,30 @@
+<?php
+/**
+ * Plugin Name: WebP Uploads
+ * Plugin URI: https://github.com/WordPress/performance/tree/trunk/plugins/images/webp-uploads
```
mukeshpanchal27:
```suggestion
 * Plugin URI: https://github.com/WordPress/performance/tree/trunk/plugins/webp-uploads
```

plugins/dominant-color-images/load.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+// Define required constants.
+define( 'DOMINANT_COLOR_IMAGES_VERSION', '1.0.1' );
+
```
mukeshpanchal27, line 29: Why removed below function from file? Without this code it generate `Plugin could not be activated because it triggered a fatal error.` when some one use old PL plugin and new module.

```
if ( function_exists( 'dominant_color_metadata' ) ) {
	return;
}
```

plugins/webp-uploads/load.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+// Define required constants.
+define( 'WEBP_UPLOADS_VERSION', '1.0.5' );
+
```
mukeshpanchal27, line 29: Why removed below function from file? Without this code it generate `Plugin could not be activated because it triggered a fatal error.` when some one use old PL plugin and new module.

```
if ( function_exists( 'webp_uploads_create_sources_property' ) ) {
	return;
}
```

plugins/webp-uploads/helper.php
mukeshpanchal27: Check #945 for more details.

mukeshpanchal27 (dismissed):

plugins.json
```diff
@@ -9,5 +9,9 @@
       "version": "1.0.5"
     }
   },
```
mukeshpanchal27: Remove the modules.

```
"modules": {}
```

plugins/dominant-color-images/load.php
thelovekesh: Shouldn't the plugin take precedence over the module if it's published as a plugin?

westonruter: Yeah, that makes sense to me. Old PL versions with that `if` intact will short circuit when the standalone plugins is active.

felixarntz:
Please take a look at https://github.com/WordPress/performance/issues/656#issuecomment-1969542264 where we can coordinate the next steps on how to proceed with this PR and #1019.

thelovekesh:
@felixarntz I think we can change the base to trunk now?

felixarntz:
@thelovekesh Thanks, I've updated it. There are a few small merge conflicts, but they should be straightforward to resolve.

.github/CODEOWNERS
```diff
@@ -49,15 +44,18 @@
...
 /plugins/speculation-rules                  @felixarntz
 /tests/plugins/speculation-rules            @felixarntz
+
+# Plugin: WebP Uploads
+/plugins/webp-uploads                       @adamsilverstein @felixarntz
+/tests/plugins/webp-uploads                 @adamsilverstein @felixarntz
+
+# Plugin: Dominant Color Images
+/plugins/dominant-color-images                                              @pbearne @spacedmonkey
+/tests/plugins/dominant-color-images                                        @pbearne @spacedmonkey
```
felixarntz: Nit-pick, let's align the spacing with the other entries.

phpunit.xml.dist
```diff
@@ -18,6 +18,12 @@
 		<testsuite name="speculation-rules">
 			<directory suffix=".php">./tests/plugins/speculation-rules</directory>
 		</testsuite>
+		<testsuite name="dominant-color-images">
+			<directory suffix=".php">./tests/plugins/dominant-color-images</directory>
+		</testsuite>
+		<testsuite name="webp-uploads">
+			<directory suffix=".php">./tests/plugins/webp-uploads</directory>
+		</testsuite>
```
felixarntz: Nit-pick: In e.g. `composer.json` and `plugins.json`, you sort the plugins alphabetically. Maybe we can do that here too?
```suggestion
		<testsuite name="dominant-color-images">
			<directory suffix=".php">./tests/plugins/dominant-color-images</directory>
		</testsuite>
		<testsuite name="speculation-rules">
			<directory suffix=".php">./tests/plugins/speculation-rules</directory>
		</testsuite>
		<testsuite name="webp-uploads">
			<directory suffix=".php">./tests/plugins/webp-uploads</directory>
		</testsuite>
```

plugins/dominant-color-images/load.php
felixarntz: It's safer to keep the `defined( ... )` constant check above the constant to prevent double loading. Yes, this probably shouldn't happen, but IMO better safe than sorry, and the other standalone plugins have that condition too.

If we decide to remove them, let's remove them everywhere rather than making partial decisions.

plugins/dominant-color-images/load.php
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @package dominant-color-images
+ */
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+// Define required constants.
```
felixarntz, line 23: See other comment thread:
```suggestion
if ( defined( 'DOMINANT_COLOR_IMAGES_VERSION' ) ) {
	return;
}

```

plugins/webp-uploads/load.php
felixarntz: The function check shouldn't be needed anymore, it was introduced before the versions constants existed. But for safety and consistency, we should keep the constant check in place for now (see my other comment).

plugins/webp-uploads/load.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @package webp-uploads
+ */
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+// Define required constants.
```
felixarntz, line 23: See other comment:
```suggestion
if ( defined( 'WEBP_UPLOADS_VERSION' ) ) {
	return;
}
```

tests/plugins/webp-uploads/helpers/class-wp-image-doesnt-support-webp.php
```diff
@@ -2,7 +2,7 @@
 /**
  * A WP_Image_Editor mock that doesn't support WebP.
```
felixarntz: Let's name this folder `data` instead of `helpers`, similar to how you did it for Dominant Color Images, and in accordance with both the original files being in the `testdata` folder.

tests/plugins/webp-uploads/helpers/class-wp-image-edit.php
```diff
@@ -2,7 +2,7 @@
 /**
  * A WP_Image_Edit class to avoid abstraction to handle image edits.
  *
- * @package performance-lab
+ * @package webp-uploads
```
felixarntz: See above, the folder with this file should have a consistent name with how you did it for Dominant Color Images.

felixarntz (commented):
@thelovekesh This is almost good to go, only some minor feedback.

One overall concern: It looks like you removed all the individual `.gitattributes` files. Don't we still need them to ensure the extra files aren't part of the plugin ZIP builds? We will at least need to exclude `.wordpress-org` and `phpcs.xml.dist` for every plugin.

mukeshpanchal27 (commented):
For consistency, could we use the `data` folder within the `webp-uploads` plugin to store images for testing, similar to how it's done for the `dominant-color-images` plugin? Currently, `webp-uploads` uses images from the `tests/testdata/modules/images` folder.

phpunit.xml.dist
thelovekesh: Seems like I missed it while resolving merge conflicts.

.github/CODEOWNERS
thelovekesh: Done https://github.com/WordPress/performance/pull/1011/commits/486014e1b6bd074ebbf110ef6a234b2248178c97

plugins/dominant-color-images/load.php
thelovekesh: Done in https://github.com/WordPress/performance/pull/1011/commits/a6c1ddb7f219e09faaec4a396f7c9e78de90bdc5

plugins/webp-uploads/load.php
thelovekesh: Done in https://github.com/WordPress/performance/pull/1011/commits/a6c1ddb7f219e09faaec4a396f7c9e78de90bdc5

plugins/dominant-color-images/load.php
thelovekesh: done in https://github.com/WordPress/performance/pull/1011/commits/a6c1ddb7f219e09faaec4a396f7c9e78de90bdc5

plugins/webp-uploads/load.php
thelovekesh: done in https://github.com/WordPress/performance/pull/1011/commits/a6c1ddb7f219e09faaec4a396f7c9e78de90bdc5

tests/plugins/webp-uploads/helpers/class-wp-image-doesnt-support-webp.php
thelovekesh: done in https://github.com/WordPress/performance/pull/1011/commits/de7732047571df69397865caf202c80a39331475

tests/plugins/webp-uploads/helpers/class-wp-image-edit.php
thelovekesh: Renamed in https://github.com/WordPress/performance/pull/1011/commits/de7732047571df69397865caf202c80a39331475

thelovekesh:
> For consistency, could we use the `data` folder within the `webp-uploads` plugin to store images for testing, similar to how it's done for the `dominant-color-images` plugin? Currently, `webp-uploads` uses images from the `tests/testdata/modules/images` folder.

@mukeshpanchal27 I think those images are used by `dominant-color-images` plugin as well. IMO we can keep shared resources in testdata and standalone in plugin-specific data directory?

thelovekesh:
> One overall concern: It looks like you removed all the individual .gitattributes files. Don't we still need them to ensure the extra files aren't part of the plugin ZIP builds? We will at least need to exclude .wordpress-org and phpcs.xml.dist for every plugin.

@felixarntz I removed it since it looks like it's not being used. Since `.gitattributes` is git specific, I can't find any instances in the build/deploy setup where this is being used for standalone plugins or modules. As a result, I can see ignored files/directories are there in the published plugin as well - https://plugins.svn.wordpress.org/dominant-color-images/trunk/.

Also, we might need to switch from `.gitattributes` to `.distignore`(see https://github.com/WordPress/performance/issues/893#issuecomment-1947398713) since some WIP plugins have assets that are not in the git tree but need to be included in the plugin release.

felixarntz:
@thelovekesh `.gitattributes` is supported by the 10up build action, no need to mention it in any of our build/deploy workflows. That's why it's used in the root of the project for the main PL plugin as well. At some point in the past we used `.distignore`, but it was less practical as it's not considered by GitHub itself, as far as I can remember.

Let me take a look at the `dominant-color-images` plugin, I believe the assets are in that one because of the same bug happening in the past, and they would need to be manually deleted to no longer be in there.

Let's use `.gitattributes` for now, as we have been, and we can rethink this in the future.

thelovekesh:
@felixarntz I think the 10up deploy action only takes `.gitattributes` into account if [`BUILD_DIR` is false](https://github.com/10up/action-wordpress-plugin-deploy/blob/abb939a0d0bfd01063e8d1933833209201557381/deploy.sh#L92). For the main plugin we don't pass any `BUILD_DIR` but for the standalone plugins we do. Can you please validate it once?

felixarntz:
@thelovekesh Oh, that's a great catch! So it looks like there's no way to use either of the two files for the 10up deploy workflow to automatically handle it?

Maybe we have to implement something in our own workflow that removes those files before triggering the 10up deploy workflow? 🤔

thelovekesh:
@felixarntz Yes, and I think instead of handling it with manual CLI script we should switch to a build tool to better handle bundling use cases.

As far as we are concerned about `.distignore`, it shouldn't hurt our workflow in any way. `.gitattributes` play nice when we only want to build the stuff that is in the git tree.

---

Coming to the release of standalone plugins, we need to move WPOrg assets to some place(maybe tmp?) and then pass it path in `ASSETS_DIR`. It should do the trick.

felixarntz:
@thelovekesh In the interest of unblocking other PRs, let's merge this PR now without the correct handling of the assets.

Similarly to what I outlined in #1008, it would be great if you could open another PR against `trunk` to fix specifically that bug, to ensure that development files and brand assets for the standalone plugins are not part of the builds and deployments.

felixarntz (approved):

PR #956, opened 2024-01-29
Add test suite for Plugins folder

mukeshpanchal27:
## Summary

Fixes #934

**NOTE: Checking test suite for plugins folder.**

## Checklist

- [x] PR has either `[Focus]` or `Infrastructure` label.
- [x] PR has a `[Type]` label.
- [x] PR has a milestone or the `no milestone` label.

PR #1013, opened 2024-02-24
Add standalone PHPUnit configs in `plugins/`

thelovekesh:
## Summary

Fixes task 2 of https://github.com/WordPress/performance/issues/1012

## Relevant technical choices
- Add `phpunit.xml.dist` to standalone plugins.
- Move test helpers to a new class.
- Add tests bootstrap files to plugins.

## Checklist

- [x] PR has either `[Focus]` or `Infrastructure` label.
- [x] PR has a `[Type]` label.
- [x] PR has a milestone or the `no milestone` label.

PR #1014, opened 2024-02-26
Fix unit test for Multisite

mukeshpanchal27:
## Summary

In #956, we introduced a test suite for plugins, updating only the `phpunit.xml.dist` file. However, we missed updating the multisite config file `tests/multisite.xml`, causing it to show "No tests executed!" without an error. This PR adds the missing configuration for the multisite test suite. Thanks to @thelovekesh for spotting this oversight.

## Checklist

### DIFF
--- a/tests/multisite.xml
+++ b/tests/multisite.xml
@@ -10,9 +10,13 @@
 		<const name="WP_TESTS_MULTISITE" value="1" />
 	</php>
 	<testsuites>
-		<testsuite name="default">
+		<testsuite name="performance-lab">
 			<directory suffix=".php">./</directory>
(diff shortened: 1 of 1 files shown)


thelovekesh:
Note: As discovered in https://github.com/WordPress/performance/pull/1013, wp-env based setup has some issues(https://github.com/WordPress/performance/pull/1013#issuecomment-1962731568). Since it's a docker-based setup we don't have much control over it. So in the same PR, I have updated the `wp-env` based setup with a more controllable and super-fast setup which simulates real scenario WordPress setup.

thelovekesh (approved):

PR #1018, opened 2024-02-28
Refactor storage lock functions into ILO_Storage_Lock class

westonruter:
## Summary

_This is a sub-PR of #1010. Review/merge it first._

Spiritually motivated by #933.

## Relevant technical choices

Top-level functions are moved into a new `ILO_Storage_Lock` class as static methods. A namespace would achieve the same thing. This makes them much easier to read. 

## Checklist


felixarntz (approved):
@westonruter This looks great to me.

You mention this is a sub PR of #1010, but I believe we could even merge this already if you decouple it from the #1010 codebase, as it doesn't appear to have overlap, at least I don't see any. Not a big deal either way, I'm saying this only in case you want to already unblock this.

PR #1022, opened 2024-03-01
Add caching to the wordpress.org query to get plugin information.

### LINKED ISSUE #1015: PL admin performance can be improved
## Summary

PL plugin is using [`plugins_api()`](https://developer.wordpress.org/reference/functions/plugins_api/) to retrieve the data of the standalone plugin from the WPOrg API. 

https://github.com/WordPress/performance/blob/0f110ffe1a50b1d49b4e6309ec311dc6327d304b/admin/plugins.php#L21-L30

Currently, when a user loads the admin page it gets fired  to retrieve the data of each available plugin:
https://github.com/WordPress/performance/blob/0f110ffe1a50b1d49b4e6309ec311dc6327d304b/load.php#L336-L342

Even when the data is retrieved for all plugins, it gets fired again when the user saves the settings or loads the page, and hence page loading time increases.

To avoid hitting the WPOrg API multiple times and reduce the page load time second time onwards, we should consider caching the plugin information for a day or so in the:
- Object cache if present.
- Else, save the response data as transients.

thelovekesh:
## Summary

Fixes https://github.com/WordPress/performance/issues/1015

## Checklist

### DIFF
--- a/admin/plugins.php
+++ b/admin/plugins.php
@@ -18,6 +18,12 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @return array Array of plugin data, or empty if none/error.
  */
 function perflab_query_plugin_info( string $plugin_slug ) {
+	$plugin = get_transient( 'perflab_plugin_info_' . $plugin_slug );
+
+	if ( $plugin ) {
+		return $plugin;
+	}
+
 	$plugin = plugins_api(
 		'plugin_information',
 		array(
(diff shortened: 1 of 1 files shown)


admin/plugins.php
```diff
@@ -37,6 +43,8 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 		$plugin = (array) $plugin;
 	}
 
+	set_transient( 'perflab_plugin_info_' . $plugin_slug, $plugin, HOUR_IN_SECONDS );
```
westonruter, line 46: The TTL could probably be higher but I don't feel strongly about it.

westonruter (approved):
LGTM

felixarntz (approved):
Good catch @thelovekesh! LGTM.

PR #1024, opened 2024-03-04
Fix missing `ABSPATH` checks in Speculation Rules plugin

felixarntz:
## Summary

The `speculation-rules` plugin was missing the `defined( 'ABSPATH' )` checks that are sort of mandatory for plugins.

## Checklist

### DIFF
--- a/admin/plugins.php
+++ b/admin/plugins.php
@@ -18,6 +18,12 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @return array Array of plugin data, or empty if none/error.
  */
 function perflab_query_plugin_info( string $plugin_slug ) {
+	$plugin = get_transient( 'perflab_plugin_info_' . $plugin_slug );
+
+	if ( $plugin ) {
+		return $plugin;
+	}
+
 	$plugin = plugins_api(
 		'plugin_information',
 		array(
@@ -36,6 +42,8 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 	if ( is_object( $plugin ) ) {
 		$plugin = (array) $plugin;
 	}
+
+	set_transient( 'perflab_plugin_info_' . $plugin_slug, $plugin, HOUR_IN_SECONDS );
 
 	return $plugin;
 }
--- a/plugins/auto-sizes/.gitattributes
+++ b/plugins/auto-sizes/.gitattributes
@@ -1,3 +1,0 @@
-/.wordpress-org export-ignore
-/.gitattributes export-ignore
-/phpcs.xml.dist export-ignore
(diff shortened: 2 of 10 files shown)


westonruter:
I thought the root `.gitattributes` was updated to account for these standalone plugin development files. The initial `/` was removed from the relevant paths so that the files are excluded anywhere in the directory tree. (As I recall. Currently AFK and can't easily confirm.)

westonruter:
Maybe I'm thinking of what was done in #1011.

felixarntz:
@westonruter 
> I thought the root `.gitattributes` was updated to account for these standalone plugin development files. The initial `/` was removed from the relevant paths so that the files are excluded anywhere in the directory tree. (As I recall. Currently AFK and can't easily confirm.)

I initially thought this would do it, but missed to consider the important part: To publish the plugins (just like for the modules) the 10up deploy script is instructed to look at the specific plugin folder. So it'll basically be as if the root `.gitattributes` doesn't exist, so we need those individual ones. It's still not a bad idea to make the root exclusions more generic, but I don't think it's sufficient for this.

This is similar to my #1011 review I just left. cc @thelovekesh

thelovekesh:
@felixarntz please check my comment here - https://github.com/WordPress/performance/pull/1011#issuecomment-1978101869

Since we are passing `BUILD_DIR` for standalone plugins, I don't think the 10up deploy script will take plugin-specific `.gitattributes` into account as per - https://github.com/10up/action-wordpress-plugin-deploy/blob/abb939a0d0bfd01063e8d1933833209201557381/deploy.sh#L149

thelovekesh:
To produce plugin builds, please also refer to https://github.com/WordPress/performance/pull/1008, which introduces Webpack (extending WP Script configs) and is capable of handling almost every use case. 

This allows us to keep utilizing a single global `.distignore` file rather than several ignore files. It may also generate any plugin build with a single command, such as this one (which is presently limited to the main plugin):

```
npm run build-plugin -- --env plugin=perflab # Build the main plugin
npm run build-plugin -- --env plugin=speculation-rules # Build the speculation-rules plugin
```

felixarntz:
@thelovekesh Great point. Let's continue the discussion on how we want to solve this problem in #1008, given neither `.gitattributes` nor `.distignore` work.

I've removed that part from this PR, so it should be good to go, only focusing on the `ABSPATH` checks now.

thelovekesh:
@felixarntz SGTM. Just out of curiosity, I wanted to know where `.distignore` is causing the issues. AFAIK it's just an ignore file that contains some patterns and is mostly used by multiple tools out there to ignore such patterns from distribution, because `.gitignore` and `.gitattributes` are bound to git specific things.

thelovekesh (approved):
LGTM 🚀

mukeshpanchal27 (approved):

PR #1027, opened 2024-03-05
Update tested WordPress version to 6.5

felixarntz:
## Summary

This PR updates Performance Lab and its standalone plugins' tested up to WordPress version to the upcoming 6.5. It also bumps the standalone plugin versions accordingly.

## Checklist

### DIFF
--- a/plugins/auto-sizes/readme.txt
+++ b/plugins/auto-sizes/readme.txt
@@ -2,9 +2,9 @@
 
 Contributors:      wordpressdotorg
 Requires at least: 6.4
-Tested up to:      6.4
+Tested up to:      6.5
 Requires PHP:      7.0
-Stable tag:        1.0.0
+Stable tag:        1.0.1
 License:           GPLv2 or later
 License URI:       https://www.gnu.org/licenses/gpl-2.0.html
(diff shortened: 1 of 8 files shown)


thelovekesh (approved):

plugins/auto-sizes/readme.txt
```diff
@@ -48,6 +48,10 @@ Contributions are always welcome! Learn more about how to get involved in the [C
 
 == Changelog ==
 
+= 1.0.1 =
+
+* Update tested WordPress version to 6.5. ([1027](https://github.com/WordPress/performance/pull/1027))
```
westonruter, line 53: It's too bad a version bump is needed to do just this. Previously I've just manually bumped the version in the `readme.txt` of the latest SVN tag as a way to avoid having to do a release.

westonruter (approved):

plugins/auto-sizes/readme.txt
felixarntz: Agreed. I think there's a way to automate this too via GitHub workflow somehow. But until we have this, probably fine to do this.

Issue #911, opened 2023-12-18
Making standalone plugins usable before they're released to the plugin repo


joemcgill:
## Feature Description

Once #656 is complete, the intention was that new modules would be initially released as standalone plugins, and not bundled with the main Performance Lab plugin. This creates a situation where new plugin modules can not easily be tested or used prior to them being available on the [plugin repo](https://wordpress.org/plugins/). Given the current long wait times for new plugins to be approved on the repo, this would create a significant problem for efficiently developing and testing out new features and no clear way of testing the feature prior to an initial release.

Currently, a new module can be activated from the Performance Lab settings screen before it has been removed as a standalone plugin, so a new module can be tested by checking out a feature branch of this repo that includes the new module and activating it in the settings screen.

### Initial ideas to consider

As originally discussed in https://github.com/WordPress/performance/pull/904#issuecomment-1858540201, there are a few ways to address this issue:

1. Allow new modules to initially be included as experimental features that can be activated in the Performance Lab plugin prior to being released as a standalone plugin. This would require maintaining the migration logic developed in #656 in the plugin indefinitely, as it will be needed as new plugins are released.
2. Update the release process for this repo so that we can publish zip files of standalone modules here, before they are available on the plugin repo. This is probably worth looking into regardless, so we can test pre-release zips prior to publishing a new release of a standalone plugin. However, this is still less convenient than being able to test updates from a single repo.

### Currently effected modules in development

* #556
* #805
* #886
* #897
* #904

felixarntz:
I think I would prefer option 1, as it allows automatic updates easily (simply via Performance Lab). It's also not a lot of additional work required since most of the necessary infrastructure is implemented already.

Any code could start as a module, and once it's approved as standalone plugin, it can be deployed that way. Users would then see the migration notice as needed, and 1-2 releases later the module would be removed from PL.

The main difference that doing this long-term would potentially require is to adjust the standalone plugin build process slightly:
* Right now all plugins get built and deployed from their module code within the `modules` directory.
* Going forward, all plugins were originally going to be built and deployed from their code that's _already_ in plugin shape (correct headers etc.), coming from another directory (e.g. `plugins`).
* If we keep the approach of migrating modules to plugins regularly, it's probably best to support both, i.e. the build process should be able to build and deploy standalone plugins from either of the two locations.
    * It should be relatively straightforward, since building modules into plugins is the most complex piece and that one is already implemented.
[...]

PR #1028, opened 2024-03-05
Add standalone plugins to the wp-env config

joemcgill:
## Summary
This updates the included `.wp-env.json` config file to ensure all standalone plugins are installed on the environment during development and testing. This behavior can be overridden locally using a `.wp-env.override.json` file.

See #911

## Checklist


felixarntz (approved):

thelovekesh (approved):

.wp-env.json
```diff
@@ -1,11 +1,17 @@
...
+  ],
   "env": {
     "tests": {
       "config": {
         "FS_METHOD": "direct"
       }
     }
   }
-}
+}
```
westonruter: EOF EOL should be retained, no?

westonruter (approved):

.wp-env.json
felixarntz: Per `.editorconfig`, JSON files should _not_ have a final newline.

westonruter: Why was that added? It's not in https://github.com/WordPress/wordpress-develop/blob/trunk/.editorconfig nor https://github.com/WordPress/gutenberg/blob/trunk/.editorconfig

felixarntz: No good reason I think, other than it has always been like that. The reverse question though: Why _should_ JSON files have a newline at the end?

westonruter: AFAIK because it's standard for text files to end in newlines. For example https://gist.github.com/OleksiyRudenko/d51388345ea55767b7672307fe35adf3 which links to https://stackoverflow.com/a/729795/93579.

See also https://unix.stackexchange.com/a/18789/218144. 

In short, it seems text utilities expect newlines to work properly.

mukeshpanchal27 (approved):

.wp-env.json
joemcgill: Looking at the other .json files in the project, they all mostly have extra lines and it looks like my editor is auto-removing the lines when I save any of these files because of the .editorconfig. I'll open a new PR for us to make the switch, but for consistency will update the file here.

```suggestion
}

```

Issue #1012, opened 2024-02-24
DX improvements in new monorepo


thelovekesh:
As per https://github.com/WordPress/performance/issues/654, we are moving published modules as standalone plugins, making the current repo a mono repo for multiple plugins.

This brings a new challenge to streamline the current tooling and build setup used to maintain coding standards, testing, and publishing the plugin. With this, we are exploring integrating existing workflows in such a way that it doesn't hurt the DX and keeps the maintenance cost low.

### Tasklist
- [x] [Done] PHPCS setup for mono repo https://github.com/WordPress/performance/issues/997 (https://github.com/WordPress/performance/pull/1002)
- [ ]  Setup PHPUnit on plugin level instead of testsuite in main PHPUnit config(See: https://github.com/WordPress/performance/issues/934#issuecomment-1960892281 and https://github.com/WordPress/performance/issues/934#issuecomment-1962093966). Testsuite-based setup was introduced in https://github.com/WordPress/performance/pull/956. 
- [x] [Done] Reconsider build setup for both main and standalone plugins. Currently, we use a custom CLI tool which often requires tuning on a use case basis which can be replaced with a tool like Webpack with build steps for each use case. (https://github.com/WordPress/performance/pull/1033)

thelovekesh:
> But assuming that the tests all remain in /tests and don't also move to the /plugins directory, there would still need to be a plugin-specific testsuite in the plugin's dedicated phpunit.xml.dist: from (https://github.com/WordPress/performance/issues/934#issuecomment-1962093966)

@westonruter I think it would make more sense to include the plugin test in the plugin directory itself. It will help us in multiple ways:
- With this new approach, we won't need to switch between directories or hunt down related files scattered throughout multiple directories.
- When a plugin becomes part of the core, removing it is much easier. Simply delete the dedicated directory, and there's no need for a tedious search and delete across the entire plugin.
- No need to modify configurations based on changes in the `plugins/` directory. 
- This new approach ensures a smooth testing experience with just one command for any plugin.

    ```sh
    ./vendor/bin/phpunit -c plugins/foo/phpunit.xml.dist # Run tests for plugin: Foo 
    ./vendor/bin/phpunit -c plugins/bar/phpunit.xml.dist # Run tests for plugin: Bar
    ```

Please take a look at the POC - https://github.com/WordPress/performance/pull/1013

westonruter:
[...]

Issue #1032, opened 2024-03-06
Enhance onboarding experience for Performance Lab


felixarntz:
#1031 focuses on revising the UI of the Performance Lab settings screen to be more intuitive about the features it allows enabling.

Separately, we should think about how we can improve the onboarding experience in particular, specifically about enabling all non-experimental features (see #1045) with one click, or at least have some kind of onboarding where those features are recommended.

There are 3 distinct use-cases to consider here, all of which we need to find a solution for:
1. A new installation of the Performance Lab plugin: The administrator should be proactively informed of the available features and which ones in particular are recommended to activate (all non-experimental ones).
2. An update of the Performance Lab plugin to a version that comes with one or more new features (i.e. new standalone plugins exposed as features): The administrator should be proactively informed of the availability of new features, especially if those features are not experimental.
3. An update of the Performance Lab plugin to a version where one of the features has "graduated" from "experimental" to "non-experimental": Unless the site already has the feature active, the administrator should be proactively informed about the feature now being recommended to activate.

Let's think about what this experience could look like. We may not need a full multi step onboarding wizard for this, as it doesn't really involve multiple steps. But at the same time, we shouldn't settle for just an admin notice or pointer, but rather shoot for a more prominent and more intuitive user experience, with minimal clicks required to get to the recommended features being active.

westonruter:
As I just mentioned in https://github.com/WordPress/performance/issues/1031#issuecomment-1980042835, we also need to be wary of readonly filesystems. If someone, for example, installs Performance Lab on the Pantheon dev environment and then pushes it to production without going through the onboarding experience on dev, they will then fail to be able to successfully do the onboarding on the live environment because it is a readonly file system.

felixarntz:
@joemcgill @mukeshpanchal27 I have updated the description to consider the third use-case of features graduating from experimental to non-experimental (also see #1045).

Let's discuss what would be a good starting point in implementing such an onboarding experience. A simple admin notice? An admin pointer plus dedicated UI on top of the PL settings screen? A custom screen purely for the onboarding (akin to a more common onboarding wizard experience where the user's entire focus is on the one step to go through)? Just some ideas to consider.

felixarntz:
Sharing a few ideas here:
* Taking into consideration that neither of the 3 use-cases outlined in the issue description requires a stepped flow, I think that a dedicated onboarding flow screen is overkill for this.
[...]

PR #1033, opened 2024-03-06
Update plugin(s) build setup for distribution

thelovekesh:
## Summary

Fixes task 3 of https://github.com/WordPress/performance/issues/1012

As mentioned by @westonruter in https://github.com/WordPress/performance/issues/893#issuecomment-1947398713, we are now moving towards use cases where we can have to build files that are not in the git tree and hence will be ignored by `git archive`.

Also as outlined in https://github.com/WordPress/performance/pull/1024#issuecomment-1979221706, in the new mono repo setup we will require a build setup because the repo level `.gitattributes` or `.distignore` won't work just to ignore files from standalone distributed plugins.

This PR proposes a simple Webpack-based plugin bundling solution that solves both problems.
## Relevant technical choices

- Switch to Webpack for handling plugin(s) building tasks with `copy-webpack-plugin` Webpack plugin.
- Remove dependency from custom-written CLI scripts to:
    - Get the plugin version from the `readme.txt` file. This task is now handled by Webpack which creates a `manifest.json` with plugin versions.
    - Get the plugin path. This is not required as we are completely removing the modules.
    - Build the standalone plugins. This task is now handled by Webpack.
- Add plugin(s) deployment status on the repo.
<details><summary>Screenshot</summary>

image
</details> 

- Add a workflow step to add the plugin(s) built to release assets with their checksum to ensure assets' integrity.

<details><summary>Screenshot</summary>

image
</details> 

## Checklist


.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -12,88 +13,81 @@ on:
...
         with:
           node-version-file: '.nvmrc'
           cache: npm
+
       - name: Install npm dependencies
-        run: npm ci
-      - name: Building standalone plugins
-        if: ${{ matrix.directory != 'plugins' }}
-        run: npm run build-plugins
-      - name: Set version
```
thelovekesh, line 81: @felixarntz @mukeshpanchal27 Can you please confirm if need to set this version? I think that was required for modules?

.github/workflows/deploy-dotorg.yml
```diff
@@ -7,17 +7,33 @@ on:
...
+        run: npm run build:plugin:performance-lab
+
       - name: WordPress plugin deploy
         uses: 10up/action-wordpress-plugin-deploy@stable
         env:
           SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
           SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          BUILD_DIR: 'build/performance-lab'
+          ASSETS_DIR: '.wordpress-org'
           SLUG: performance-lab
```
felixarntz: What is the purpose of all these changes? I don't see a good reason for them. The main Performance Lab process has not needed a build process so far, and I don't see why it would need one now.

I understand we need to fix the process for standalone plugins, but for Performance Lab it has worked perfectly well the way it has been so far. The additions here, if anything, will make the process slower as far as I can tell, due to installing NPM dependencies etc.

.distignore
```diff
@@ -0,0 +1,87 @@
+############
```
felixarntz: Similar to my questions elsewhere: Why `.distignore` instead of `.gitattributes`?

`.gitattributes` has the benefit that it is also processed by GitHub. When you click "Download ZIP", you only get a ZIP of the actual Performance Lab plugin without the dev ops files. With `.distignore` that is no longer the case.

If we replace `.gitattributes` with `.distignore`, we need a compelling reason why. Can you explain your rationale?

package.json
```diff
@@ -12,18 +12,26 @@
...
+    "webpackbar": "^6.0.1"
   },
   "scripts": {
     "changelog": "./bin/plugin/cli.js changelog",
     "since": "./bin/plugin/cli.js since",
     "readme": "./bin/plugin/cli.js readme",
     "translations": "./bin/plugin/cli.js translations",
-    "build-plugins": "./bin/plugin/cli.js build-plugins",
+    "build-plugins": "npm-run-all build:plugin:*",
+    "build:plugin:performance-lab": "webpack --mode production --env plugin=performance-lab",
```
felixarntz: See above: I don't see why we need a build process for Performance Lab. Its process works perfectly fine with only relying on `.gitattributes` for the files and directories to ignore.

.github/workflows/deploy-standalone-plugins.yml
felixarntz: This is needed to set the correct `VERSION` environment variable for the 10up deployment workflow. You removed this below, but it has to be maintained to make sure we publish the correct version.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -12,88 +13,81 @@ on:
...
+      - name: Deploy Standalone Plugin - ${{ matrix.plugin }}
         uses: 10up/action-wordpress-plugin-deploy@stable
         with:
-          dry-run: ${{ matrix.dry-run }}
+          dry-run: ${{ github.event_name == 'workflow_dispatch' && inputs.dry-run || false }}
         env:
           SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
           SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
-          SLUG: ${{ matrix.slug }}
-          VERSION: ${{ steps.set_version.outputs.version }}
```
felixarntz, line 97: See above, this must not be removed.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -12,88 +13,81 @@ on:
...
+
 jobs:
-  release:
-    name: Prepare Deployment
+  pre-run:
+    name: Pre-run
     runs-on: ubuntu-latest
     outputs:
-      matrix: ${{ steps.set-matrix.outputs.matrix }}
+      matrix: ${{ steps.set-matrix.outputs.json }}
```
felixarntz: What does this change do?

felixarntz (changes requested):
@thelovekesh For building the standalone plugins, this looks solid so far.

However, I don't think it makes sense to introduce a build process for Performance Lab, this is a solution in search of a problem IMO. And I don't see a good reason to use `.distignore` over `.gitattributes`, please clarify why you're proposing this change.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -12,88 +13,65 @@ on:
...
+
       - name: Set matrix
         id: set-matrix
         run: |
+          PLUGINS=$(jq -r '.plugins' plugins.json)
           if ${{ github.event_name == 'workflow_dispatch' }}; then
-            result=$(echo "${{ steps.get-version.outputs.version }}" | awk '/^(\*|[0-9]+(\.[0-9]+){0,2}(-[a-zA-Z0-9.]+)?)$/ {print "Matched"}')
-            if [[ -n "$result" ]]; then
-              # Set the manual input values in JSON format for use in the matrix.
-              echo "matrix={\"include\":[{\"slug\":\"${{ inputs.slug }}\",\"version\":\"${{ steps.get-version.outputs.version }}\",\"directory\":\"${{ steps.get-plugin-directory.outputs.directory }}\",\"dry-run\":\"${{ inputs.dry-run }}\"}]}" >> $GITHUB_OUTPUT
```
mukeshpanchal27, line 48: Why the manual workflow is removed? https://github.com/WordPress/performance/actions/workflows/deploy-standalone-plugins.yml?query=branch%3Aupdate%2Fplugin-bundling

mukeshpanchal27 (changes requested):

.github/workflows/deploy-standalone-plugins.yml
thelovekesh: @mukeshpanchal27 Still working on the workflow files. Will update you once done.

.github/workflows/deploy-standalone-plugins.yml
thelovekesh: Thanks for confirming. I have re-added this. Now the Webpack generates the build manifest of plugin(s) which can provide version information.

<details><summary>manifest.json</summary>

```json
{
  "performance-lab": "2.9.0",
  "auto-sizes": "1.0.1",
  "dominant-color-images": "1.0.2",
  "speculation-rules": "1.1.0",
  "webp-uploads": "1.0.6"
}
```

</details>

.github/workflows/deploy-dotorg.yml
thelovekesh: @felixarntz Thanks for pointing this out. I have updated the PR description with the required details on why we need to switch to another build setup.

Now as you mentioned:

>.gitattributes has the benefit that it is also processed by GitHub. When you click "Download ZIP", you only get a ZIP of the actual Performance Lab plugin without the dev ops files. With .distignore that is no longer the case.

In cases where the PL plugin has such files that are not in the git tree and are generated by a build setup, this thing will not be the ideal solution. IMO prod-ready zip should be tagged to plugin release assets which also provide the flexibility to download releases with GitHub APIs.

Now a build step for the main plugin brings a few things to the table:
- A prod-ready zip is tagged on the release assets.
- In case someone needs the PL plugin built locally, it provides the way to do it.
- Since we already created a build system for standalone plugins, the same will also work with the main plugin. So I think it would be better to take the control over distribution creation process of the PL plugin as well from deploy action like we are doing for other plugins.

.distignore
thelovekesh: Please check my above comment and https://github.com/WordPress/performance/issues/893#issuecomment-1947398713

.github/workflows/deploy-standalone-plugins.yml
thelovekesh: It simplifies the matrix for the plugins list. Now we just provide the array of plugins in the matrix and all other data is fetched from build manifest.

.github/workflows/deploy-dotorg.yml
felixarntz: @thelovekesh 
> In cases where the PL plugin has such files that are not in the git tree and are generated by a build setup, this thing will not be the ideal solution.

Can you provide a concrete example where that is the case? I don't see any files being generated by the build setup for Performance Lab.

> IMO prod-ready zip should be tagged to plugin release assets which also provide the flexibility to download releases with GitHub APIs.

That is a good point, but it doesn't require us to have a build step. This can be accomplished with the 10up deploy action alone, see https://github.com/10up/action-wordpress-plugin-deploy?tab=readme-ov-file#inputs.

I think the reasons you're pointing out are valid in theory, but not in practice. As far as I can tell, the Performance Lab plugin doesn't require to be built, it works just from the source - unless I'm missing something. Unless there is a concrete usage that requires the build process, I think we should avoid it, as it introduces a process for something that doesn't need it, which makes things slower and is one more thing to maintain.

.distignore
felixarntz: Replied to your reply there :)

.github/workflows/deploy-dotorg.yml
thelovekesh: @felixarntz I think there are two possible scenarios here:
- PL plugin which doesn't have any build setup as of now, so works well with the current `.gitattributes` based setup.
- Standalone plugins, where:
    - ILO and Web Worker Offloading have a build step that includes vendor library.
    - We need to generate builds for these which should have the vendor files(which are git ignored) and should not have config/WP Assets files.

Now we have a webpack-based build step(by keeping standalone plugins in mind) that will build plugins and ignore files that are not required but include files that are git ignored(which doesn't seem possible with `.gitattributes`), hence we switched to `.distignore`.

> Unless there is a concrete usage that requires the build process, I think we should avoid it, as it introduces a process for something that doesn't need it, which makes things slower and is one more thing to maintain.

Now the same build setup can be used to generate for build PL plugin as well at no extra cost, I don't think it will require any maintenance.

felixarntz: @thelovekesh Based on your reply, there's no reason to use a build process for Performance Lab. I completely agree the other standalone plugins need it, but that doesn't mean we need to use it for PL as well. Even if the maintenance cost is little, there is no benefit to using it for Performance Lab. So let's use it only for the standalone plugins.

> Now we have a webpack-based build step(by keeping standalone plugins in mind) that will build plugins and ignore files that are not required but include files that are git ignored(which doesn't seem possible with `.gitattributes`), hence we switched to `.distignore`.

I don't understand this. `.gitattributes` is used by GitHub e.g. to ignore files when users download a ZIP file, and it's also used by the 10up deploy script. I don't think it interferes with a custom Webpack build process. Of course 3rd party libraries like `web-vitals` need to be in `.gitignore`, which means they don't even need to be listed in `.gitattributes` because they won't be part of the repo. For the build process of the standalone plugins, that means they have to run `npm install` before building the standalone plugins, and those aren't subject to `.gitattributes` anyway - only the main PL plugin is.

In other words: I think as long as we continue to separate the processes, one for deploying PL, another one for building and deploying standalone plugins, there should not be an issue. If I'm missing something, can you please provide more detail on which problem you're encountering because of `.gitattributes`?

thelovekesh: > In other words: I think as long as we continue to separate the processes, one for deploying PL, and another one for building and deploying standalone plugins, there should not be an issue. If I'm missing something, can you please provide more detail on which problem you're encountering because of .gitattributes?

Okay, I see. In essence, my goal was to make this procedure more efficient for every plugin. I believe that if we leave `.gitattributes` alone, I will have to explicitly specify which patterns to ignore for the standalone plugin build rather than obtaining them from a global `.distignore`.

I don't mind if the release workflow for PL and standalone plugins is kept separate, but I still think that the build setup should retain `.distignore` since the 10up deploy action [supports](https://github.com/10up/action-wordpress-plugin-deploy?tab=readme-ov-file#excluding-files-from-deployment) it (if you want to handle the build from the 10up action) or with plugin build setup if you agree to keep `.distignore`. 

Attaching the build zip to the release section is always an option if we are only keeping the `.gitattributes` because it offers processed ZIP. In any case, once we have build files for the PL plugin, we will need to delete it. Who knows, maybe we'll use WordPress components for https://github.com/WordPress/performance/issues/1032. However, as you pointed out, that's only a theory 😎

felixarntz: > I believe that if we leave `.gitattributes` alone, I will have to explicitly specify which patterns to ignore for the standalone plugin build rather than obtaining them from a global `.distignore`.

This is the part I do not understand. How does this work when using `.distignore` but not when using `.gitattributes`? As far as my understanding goes, neither of these two files are used by the 10up deploy action when we specify a custom `BUILD_DIR`, as you pointed out before. So what's the difference between these two files for the build process you're adding here? Where in the build process is `.distignore` used? Maybe I'm missing something, so can you please clarify?

thelovekesh: Aah okay. I am using `.distignore` to pass patterns to be ignored in the `copy-webpack-plugin`.

https://github.com/WordPress/performance/blob/9f38e400ecdac6ddf3abc505adf0d44e11db273b/webpack.config.js#L82

felixarntz: I see! Thanks for clarifying. From that perspective, I agree using `.distignore` makes sense.

However, we would still lose out on the current default GitHub behavior from using `.gitattributes` here.

How about the following:
* We keep `.gitattributes` as is for now.
* We use `.distignore` too, but _only_ for the standalone plugins. This means that list can be a lot shorter (probably only needs to cover 3-5 file/folder names and it doesn't have to be a "copy" of `.gitattributes`, which avoids a maintenance nightmare. I think you'd only need to put `.wordpress-org`, `phpcs.xml.dist`, and `.distignore` in there.

I'm not sure this is necessarily a great end solution, but I think it would be good enough for now and satisfy all our use-cases. As long as Performance Lab doesn't need a build process, it doesn't need to use `.distignore` either, so we could limit `.distignore` to only the few files/folders used for the standalone plugins. We should add a comment on top of `.distignore` to clarify this is only used for the standalone plugin build process.

WDYT?

thelovekesh: Yes @felixarntz, it makes sense to me. I believe we should keep the `.gitattributes` file as it is also processed by https://github.com/github-linguist/linguist (and the AMP plugin uses it like https://github.com/ampproject/amp-wp/blob/develop/.gitattributes). This is after researching the usage of `.gitattributes` in a GitHub repository in greater detail.

The only drawback is that the 10up deploy action will prioritize `.distignore` if we retain both `.gitattributes` and `.distignore`([ref](https://github.com/10up/action-wordpress-plugin-deploy/blob/abb939a0d0bfd01063e8d1933833209201557381/deploy.sh#L94)). 

Since Webpack is now using `.distignore`, I would suggest moving it to `bin/webpack`. As a result, the standalone plugin will continue to use `.distignore`, and the PL plugin will continue to use `.gitattributes`. 

We can easily transfer it to the plugin root if we ever decide to use `.distignore` for PL plugins as well.

thelovekesh: Also, I don't see any issues if we use `.distignore` with the 10up deploy action as well. Then it will be like:
- Use `.distignore` for creating distribution builds.
- Use `.gitattributes` for GitHub related chores and ZIP processing.

felixarntz: @thelovekesh 
> Since Webpack is now using `.distignore`, I would suggest moving it to `bin/webpack`. As a result, the standalone plugin will continue to use `.distignore`, and the PL plugin will continue to use `.gitattributes`. 

This sounds great to me.

> Also, I don't see any issues if we use `.distignore` with the 10up deploy action as well. Then it will be like:
> - Use `.distignore` for creating distribution builds.
> - Use `.gitattributes` for GitHub related chores and ZIP processing.

While we could do that, this would mean we would still need to basically maintain the large file list in two places. So I would prefer to go with your prior suggestion and use `.gitattributes` in the root (for both GitHub and Performance Lab build/deployment) and `.distignore` in `bin/webpack` (for the standalone plugins build/deployment). That way we can keep `.distignore` light and only focused on the few files that need to be ignored for the standalone plugin folders.

thelovekesh: @felixarntz It looks like the ignore file is not working as expected when moving it to `bin/webpack` directory since it's adding cwd path to patterns. Hence I have directally added the ingore paths in Webpack config for now since they are very few. Please let me know if it looks good to you.

https://github.com/WordPress/performance/blob/41b45ca5826671fe5cb7599704a688310336d0d3/webpack.config.js#L68-L71

felixarntz: @thelovekesh Thanks, that works for me! Given it's so few files, it should be okay.

package.json
```diff
@@ -12,18 +12,26 @@
...
+    "webpackbar": "^6.0.1"
   },
   "scripts": {
     "changelog": "./bin/plugin/cli.js changelog",
     "since": "./bin/plugin/cli.js since",
     "readme": "./bin/plugin/cli.js readme",
     "translations": "./bin/plugin/cli.js translations",
-    "build-plugins": "./bin/plugin/cli.js build-plugins",
+    "build-plugins": "npm-run-all 'build:plugin:!(performance-lab)'",
+    "build:plugin:performance-lab": "rm -rf build/performance-lab && mkdir -p build/performance-lab && git archive HEAD | tar -x -C build/performance-lab",
```
thelovekesh, line 30: @felixarntz Added this command, in case we ever need to confirm what's being added in the build by deploy action.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -1,99 +1,154 @@
...
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v3
-      - name: Setup Node.js (.nvmrc)
-        uses: actions/setup-node@v3
-        with:
-          node-version-file: '.nvmrc'
-          cache: npm
-      - name: Install npm dependencies
-        run: npm ci
```
felixarntz, line 30: You're moving these steps to `deploy` further below, which runs for every plugin to deploy. Maybe that's the only way, but would it be possible to instead leave these actions here?

The reason is that checking out the repository and installing dependencies will be the same for every plugin. So it would be more efficient to do it once in the beginning, regardless of how many or which plugins are being deployed afterwards.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -1,99 +1,154 @@
...
+      - name: Set plugin version
+        working-directory: ./build
         run: |
-          if [ -z "${{ matrix.version }}" ]; then
-            echo "version=$(node ./bin/plugin/cli.js get-plugin-version --slug=${{ matrix.slug }})" >> $GITHUB_OUTPUT
-          else
-            echo "version=${{ matrix.version }})" >> $GITHUB_OUTPUT
-          fi
-      - name: Deploy Standalone Plugin - ${{ matrix.slug }}
+          echo "PLUGIN_VERSION=$(jq -r '."${{ matrix.plugin }}"' manifest.json)" >> $GITHUB_ENV
```
felixarntz, line 78: Curious: Why use `$GITHUB_ENV` instead of `$GITHUB_OUTPUT` like before?

felixarntz (commented):
@thelovekesh This looks quite good to me now in terms of the overall approach. A few small comments/questions below.

I'm not very familiar with GitHub workflows, so it would be great to get @swissspidy's feedback on those changes in particular.

.github/workflows/deploy-standalone-plugins.yml
thelovekesh: @mukeshpanchal27 I have updated the workflow files. Please feel free to re-review them.

thelovekesh: And JFI, we are now using build manifest instead of using a CLI script to determine the plugin versions.

.github/workflows/deploy-standalone-plugins.yml
thelovekesh: I think it's best to use `$GITHUB_OUTPUT` when you want to share the info across the jobs. For the same job, an env variable is suffice.

.github/workflows/deploy-standalone-plugins.yml
thelovekesh: @felixarntz These steps are here just to install NPM deps so that we can run a few CLI commands to get the plugin version and path to the matrix.

In the previous setup, we were running `npm ci` in two places:
- release job
- deployment job(which is also a matrix)

In the current setup, we were running `npm ci`  only once:
- deploy job (which is a matrix)

package.json
thelovekesh: @felixarntz I believe that we should use the above command to verify distribution files(maybe just before a release?). I can see that the `build-cs` directory is being committed in the svn at https://plugins.trac.wordpress.org/browser/performance-lab/trunk for the most current release 😞. 

Also, it would be good if we log an issue to remove such redundant files/dirs from all our plugins in the next release.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -1,99 +1,154 @@
...
+          cd ./build/${{ matrix.plugin }}
+          zip -r ../dist/${{ matrix.plugin }}.zip .
+
+      - name: Generate checksum
+        working-directory: ./build/dist
+        run: |
+          mkdir -p $RUNNER_TEMP/plugin-checksums
+          find . -type f -print0 | sort -z | xargs -r0 shasum -a 256 -b | sed 's# \*\./# *#' > $RUNNER_TEMP/plugin-checksums/checksums.txt
+          shasum -a 256 -U -c $RUNNER_TEMP/plugin-checksums/checksums.txt
+          cat $RUNNER_TEMP/plugin-checksums/checksums.txt | while read sum file; do echo "$sum $file" > ${file#\*}.sha256; done
```
swissspidy: Do we really need to create checksums? A bit overkill IMO. I doubt anyone would ever look at these.

If someone wants checksum verification, WordPress.org already provides that information anyway through a nice API.

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -1,99 +1,154 @@
...
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.plugin }}
+          path: ./build/dist
+
+      - name: Start deployment
+        uses: bobheadxi/deployments@v1
```
swissspidy, line 126: Interesting action! 👍 👀

swissspidy (approved):
Overall looks good 👍  Looking forward to seeing this in action.

mukeshpanchal27 (commented):
@thelovekesh, thank you for the updates.

It seems that `npm run test-plugins` is broken. Have you checked the automated and manual plugin release workflow with `dry-run` set to true? This way, we can thoroughly test all steps of the workflow. Could you please ensure the following:

- Added a pull_request for checking the automated workflow.
- Set both auto and manual workflow DRY Run to true for security.
- Updated SVN credentials for double security.

thelovekesh:
Thanks @mukeshpanchal27. It seems like breaking changes are unrelated to this PR. I can see them on https://github.com/WordPress/performance/pull/1028 as well - https://github.com/WordPress/performance/actions/runs/8182298569/job/22373438106

>  Have you checked the automated and manual plugin release workflow with dry-run set to true?

Yes, I have tested them on my fork. 
- Standalone plugins release logs - https://github.com/thelovekesh/performance/actions/workflows/deploy-standalone-plugins.yml
- PL plugin release logs - https://github.com/thelovekesh/performance/actions/workflows/deploy-dotorg.yml
- Release assets linking - https://github.com/thelovekesh/performance/releases/tag/test-release
- Deployment status - https://github.com/thelovekesh/performance/deployments/

.github/workflows/deploy-standalone-plugins.yml
thelovekesh: It's not a "must use," but providing a means to verify the integrity of your distributed assets is never bad.

.gitattributes
```diff
@@ -10,6 +10,7 @@
 /tests export-ignore
 /plugin-tests export-ignore
 /plugins export-ignore
+/build-cs export-ignore
```
felixarntz, line 13: Great catch!

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -1,99 +1,154 @@
...
+        with:
+          name: ${{ matrix.plugin }}
+          path: ./build/dist
+
+      - name: Upload release assets
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            ./build/dist/${{ matrix.plugin }}.zip
+            ./build/dist/${{ matrix.plugin }}.zip.sha256
```
felixarntz: As far as I can tell, this will add the release assets to _any_ GitHub release, correct? Does that mean when we publish a new Performance Lab release and all relevant standalone plugins are updated as part of that, the Performance Lab release on GitHub will have all those standalone plugins' ZIP files attached?

I'm not saying this is necessarily a bad idea, I just want to make sure we have thought this through. In the future, we may consider having individual GitHub releases for the standalone plugins, so in such a scenario we probably wouldn't want to do this, but rather attach the ZIP file only to the corresponding release of _that_ plugin.

felixarntz (commented):
@thelovekesh @swissspidy How does this PR determine which plugins to deploy an update for? Here's what I mean:
* When a new Performance Lab release is triggered for instance, not all standalone plugins will need to be updated.
* A plugin should only be updated if its new version is greater than the latest version currently available on wordpress.org.

I'm not certain the workflow currently accounts for that. I know the 10up deploy action handles this internally already, but I don't see anything in the other steps of our workflow that take that into account. Can you double check? 🤔 

Also, another somewhat related comment on which ZIP files get attached to which GitHub release. FWIW, if that complicates getting this merged, we could always leave that out and add separately, also to solve it holistically for Performance Lab as well. We need to rethink our release strategy anyway.

.github/workflows/deploy-standalone-plugins.yml
thelovekesh: > As far as I can tell, this will add the release assets to any GitHub release, correct? Does that mean when we publish a new Performance Lab release and all relevant standalone plugins are updated as part of that, the Performance Lab release on GitHub will have all those standalone plugins' ZIP files attached?

That's correct. 

> I'm not saying this is necessarily a bad idea, I just want to make sure we have thought this through. In the future, we may consider having individual GitHub releases for the standalone plugins, so in such a scenario we probably wouldn't want to do this, but rather attach the ZIP file only to the corresponding release of that plugin.

@felixarntz As long as we are releasing all plugins on release, IMO we can keep it. As mentioned in https://wordpress.slack.com/archives/C02KGN5K076/p1710262256960409, once we rethink on how we streamline the release this can be taken care of.

thelovekesh: Also, it would be better if we open a separate ticket for this and track all the discussions there.

thelovekesh:
> I'm not certain the workflow currently accounts for that. I know the 10up deploy action handles this internally already, but I don't see anything in the other steps of our workflow that take that into account. Can you double check? 🤔

@felixarntz At this point, no it doesn't take that into account and it's that way only. We need to re-consider on how we release our plugins and IMO that should be a part of different issue/PR.

felixarntz:
@thelovekesh 
> > I'm not certain the workflow currently accounts for that. I know the 10up deploy action handles this internally already, but I don't see anything in the other steps of our workflow that take that into account. Can you double check? 🤔
> 
> @felixarntz At this point, no it doesn't take that into account and it's that way only. We need to re-consider on how we release our plugins and IMO that should be a part of different issue/PR.

I agree with you regarding the attachment of ZIP files to the release that it can be solved separately. However, the part about detecting whether a plugin even needs to be updated has to be handled here, otherwise it breaks existing behavior. For example:
* Let's say "WebP Uploads" is currently live on wordpress.org in version 1.0.5.
* In this repository, it currently has version 1.0.5 in its plugin main file and `readme.txt`.
* Publishing a new release on GitHub must not attach a ZIP file for WebP Uploads to that release, because 1.0.5 was already published before.

In other words, this would be incorrect since the 1.0.5 ZIP generated wouldn't _actually_ be the 1.0.5 release. That's why this needs to be fixed here to prevent this bug.

thelovekesh:
sure @felixarntz. I have updated the workflow only to deploy a plugin if it's not already deployed on the WPOrg.

.github/workflows/deploy-standalone-plugins.yml
thelovekesh: Addressed in https://github.com/WordPress/performance/pull/1033/commits/0d5451d8bcd9c4619a1ccce54bf7e6e981b5987c

felixarntz (approved):
@thelovekesh Looks great to me!

Before merge, @swissspidy could you take another brief look at the new additions since you had previously approved?

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -1,99 +1,201 @@
...
+          name: ${{ matrix.plugin }}
+          path: ./build/dist
+
+      - name: Upload release assets
+        if: steps.artifact-existence.outputs.exists == 'true'
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            ./build/dist/${{ matrix.plugin }}.zip
+            ./build/dist/${{ matrix.plugin }}.zip.sha256
```
westonruter, line 201: So this will attach all standalone plugin ZIPs to the Performance Lab plugin release? If so, it seems like a missing piece in general for the standalone plugin story is how to trigger a release of a standalone plugin without having to do a release of Performance Lab as well. I guess that can still be done by [manually invoking the workflow](https://github.com/WordPress/performance/actions/workflows/deploy-standalone-plugins.yml), but then there would be no record of the standalone plugin release among the GitHub releases.

thelovekesh: Yes, that's right. For now, there are two scenarios:
- Release trigger.
- On workflow dispatch.

In the second case, there will be no release record as mentioned by @westonruter. In the last Performance chat, we discussed opening a new issue which I have just opened https://github.com/WordPress/performance/issues/1061 to discuss it.

felixarntz: @westonruter Correct, at the moment the manual workflow is the way to do it, if we need to release something apart from a PL release.

A discussion on how to better address that is long overdue and has come up in a few places before. Let's tackle that in a separate issue. Potentially we can just create GitHub releases for the individual plugins and have the workflows adjust based on e.g. the release name or something like that what gets deployed.

thelovekesh: > So this will attach all standalone plugin ZIPs to the Performance Lab plugin release?

Only for such plugins that are not already released on WPOrg. See https://github.com/WordPress/performance/pull/1033/commits/0d5451d8bcd9c4619a1ccce54bf7e6e981b5987c

bin/webpack/utils.js
```diff
@@ -0,0 +1,84 @@
...
+ * Determine the plugin version from the readme.txt file.
+ *
+ * @param {string} pluginPath The path to the plugin.
+ *
+ * @return {string|false} The plugin version or false if not found.
+ */
+const getPluginVersion = ( pluginPath ) => {
+	const readmePath = path.resolve( pluginPath, 'readme.txt' );
+
+	const fileContent = fs.readFileSync( readmePath, 'utf-8' );
```
mukeshpanchal27, line 36: Not blocker: Better to use try catch for error handling.

```
let fileContent = '';
try {
	fileContent = fs.readFileSync( readmePath, 'utf-8' );
} catch ( error ) {
	log(
		formats.error(
			`Error reading the file "${ readmePath }": "${ error }"`
		)
	);
}
```

thelovekesh: There's no reason to handle the error here, in my opinion. We must end the operation here if it is unable to read or resolve the `readme.txt`. We don't need to add our error statements as NodeJS errors are already verbose.

swissspidy (approved):
Still looks good to me now. I'd say let's just get this in and give it a try, we can always refine later if needed.

mukeshpanchal27 (approved):

PR #1034, opened 2024-03-07
Update the .editorconfig to add a new line for json files

joemcgill:
## Summary

See: https://github.com/WordPress/performance/pull/1028#discussion_r1513755830

## Checklist

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -21,7 +21,6 @@ indent_size = 2
 trim_trailing_whitespace = false
 
 [*.json]
-insert_final_newline = true
 indent_style = space
 indent_size = 2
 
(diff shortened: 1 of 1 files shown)


thelovekesh (approved):

.editorconfig
```diff
@@ -21,7 +21,7 @@ indent_size = 2
 trim_trailing_whitespace = false
 
 [*.json]
-insert_final_newline = false
+insert_final_newline = true
```
mukeshpanchal27:
```suggestion
```

@joemcgill Instead of setting it to `true`, could we remove it so it will use the global configuration, which is already set to `true`?

mukeshpanchal27 (approved):

PR #1037, opened 2024-03-07
Remove the extension "-jpg" from the generated file name of the full sized webp version

### LINKED ISSUE #545: Allow for modification or removal of the -jpg suffix added to webp images, generated from jpg files
## Feature Description

As per a user request in the forums allow for modification or removal of the naming structure of _webp_ images that are generated from _jpg_ files. At present when _webp_ files are created from uploaded _jpg_ they include a -jpg suffix. 

Example: If I upload `myimage.jpg` the respective webp version will be titled `myimage-jpg.webp`

- [Support topic](https://wordpress.org/support/topic/remove-the-jpg-added-to-the-webp-filename/#post-16038138)
- This is the first such report in the WordPress support forums

Update: The below support topics also reference issues or queries regarding the file names that were added:
- https://wordpress.org/support/topic/modified-file-names/
- https://wordpress.org/support/topic/images-rename/

rilwis:
## Summary

In #545, a problem when uploading a `image.jpg` file is that the plugin will create the full sized version of the image with the name `image-jpg.webp`. This PR changes this behavior by not automatically append `-jpg` to the file name. The file name is only appended when there's an existing image with the same name.

Fixes #545

## Relevant technical choices

Previously, the plugin relies on `$editor->generate_filename`. But this function always consider a suffix and add `-{$suffix}` to the file name. There's no way to eliminate this. So I have to create a helper function `webp_uploads_generate_filename` to handle this.

There is 2 things that this function consider:

1. Whether the image size is `full`
2. Whether the filter above is `true`

## Checklist


plugins/webp-uploads/helper.php
```diff
@@ -175,6 +172,45 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
 	);
 }
 
+/**
+ * Generate a file name for an image with a different extension.
```
```suggestion
 * Generates a file name for an image with a different extension.
```

plugins/webp-uploads/helper.php
```diff
@@ -175,6 +172,45 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
...
+ *
+ * @param WP_Image_Editor $editor    The image editor that handle the image manipulation.
+ * @param string          $file      Full path to the image file.
+ * @param string          $size      The image size.
+ * @param string          $extension The image extension.
+ *
+ * @since n.e.x.t
+ *
+ * @return string
```
```suggestion
 * @return string File name with a different extension.
```

plugins/webp-uploads/helper.php
```diff
@@ -175,6 +172,45 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
...
+ * Generates a file name for an image with a different extension.
+ *
+ * @param WP_Image_Editor $editor    The image editor that handle the image manipulation.
+ * @param string          $file      Full path to the image file.
+ * @param string          $size      The image size.
+ * @param string          $extension The image extension.
+ *
+ * @since n.e.x.t
+ *
+ * @return string File name with a different extension.
```
mukeshpanchal27:
```suggestion
 * @since n.e.x.t
 *
 * @param WP_Image_Editor $editor    The image editor that handle the image manipulation.
 * @param string          $file      Full path to the image file.
 * @param string          $size      The image size.
 * @param string          $extension The image extension.
 * @return string File name with a different extension.
```

felixarntz:
@rilwis Thank you for the PR.

However, I'm not sure we should apply this change. The reason the extension is added is to prevent conflicts. This was a conscious decision which actually came in response to bugs reported. For example, if a user uploads `image.jpg` and `image.jpeg`, they would both become `image.webp` and thus cause a conflict.

I realize this is a very specific scenario, but it is a bug that actually happened and was reported. So I'm not sure we should allow removing the original file extension suffix just for better looking file names - unless there is another reason that the suffix is a problem.

rilwis:
@felixarntz Thanks for your comment. This PR doesn't change the default behavior that you described. When people making the change via the filter, they have to do that with a clear intention. So it is the user's problem when they upload images with the same file name like `image.jpg` and `image.jpeg`.

More details are discussed in the issue #545 and I think @mukeshpanchal27 can provide more info, because he suggested the filter.

felixarntz:
@rilwis Having a filter for that seems okay, but I don't find it a great solution. Filters can only be set by developers, but then users need to live with the consequences. If you're the developer _and_ administrator of a site, that makes sense. But many sites are not like that, the people that use the site and the people that develop the site are different.

I think a better solution would be to handle it automatically, in a smart way that only applies the suffix when needed:
* By default, don't add a suffix for the original file extension.
    * For example, if you upload `image.jpg`, you get `image.webp`.
* However, if a file of the same name already exists, add the original extension to prevent a conflict.
    * For example, if you upload `image.jpeg` after uploading `image.jpg`, you get `image-jpeg.webp` in addition to `image.webp`.

This way, it just works, no developer intervention required, and satisfies both issues:
* There is no chance for conflicts.
* The file names look nicer (unless there's a conflict).

rilwis:
@felixarntz After checking the source code of WordPress, I found that WordPress already has a mechanism to prevent duplicated file names:

https://github.com/WordPress/WordPress/blob/7ae05b77532b5902c0f67a5791abaec45220b7ca/wp-includes/functions.php#L2577

This function checks only the file name **without** extension and append numbers if neccessary.

So when a user uploads `image.jpeg` after uploading `image.jpg`, you'll get `image-1.jpeg` automatically. And with our code, it will be converted to `image-1.webp`. I've tested and saw it works that way.

I've updated the code below. Please take a look.

mukeshpanchal27:
Thanks, @rilwis, for working on the PR. I appreciate the way you implemented the changes. However, as @felixarntz mentioned, we should go with the prefix to avoid conflicts. Since the PR generates image names like WP did, I'm happy to proceed with it, but we need to get Felix's feedback on it. Once we agree, we'll need to address the unit tests as some of them are failing now. Thanks! cc @adamsilverstein

rilwis:
@mukeshpanchal27 Tests are updated. There's also a test for uploading `image.jpeg` and `image.jpg`.

felixarntz:
@rilwis 
> https://github.com/WordPress/WordPress/blob/7ae05b77532b5902c0f67a5791abaec45220b7ca/wp-includes/functions.php#L2577
> 
> This function checks only the file name **without** extension and append numbers if neccessary.
> 
> So when a user uploads `image.jpeg` after uploading `image.jpg`, you'll get `image-1.jpeg` automatically. And with our code, it will be converted to `image-1.webp`. I've tested and saw it works that way.

I'm aware of the code that does this, but as far as I remember, that function doesn't consider `image.jpeg` and `image.jpg` the same file name - because they aren't. So it only becomes a problem when you change both of their file extensions to `.webp`, and that happens in a completely different place in the code which currently doesn't check for duplicate file names.

But maybe I'm missing something. If so, can you point me to where that function detects those files to be the same file name even though their file extensions are different?

When you tested this, did you test it in both configurations? See _Settings > Media_, it needs to work both when only generating WebP as well as when generating JPEG _and_ WebP.

rilwis (commented):
@felixarntz: You're right. I looked at the wrong place. The code I refered above was about checking the image with suffix "-scaled" and rename if it already exists.

I tested again with the "Generate JPEG files in addition to WebP" settings enabled and saw if `image.jpeg` is uploaded after `image.jpg`, then they generate the same `image.webp`.

If the settings "Generate JPEG files in addition to WebP" is disabled (unchecked), then WordPress auto appends `-1` to the webp versions.

I've updated the PR with a check for `file_exists` to rename the file if needed. Please take a look.

westonruter:
@rilwis Hi! Could you update the PR description to reflect the latest state of this PR? That will help us get new reviewers so we can get this merged for the next release.

rilwis:
@westonruter Done.

plugins/webp-uploads/helper.php
```diff
@@ -183,6 +180,36 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
...
+ * @param WP_Image_Editor $editor    The image editor that handle the image manipulation.
+ * @param string          $file      Full path to the image file.
+ * @param string          $size      The image size.
+ * @param string          $extension The image extension.
+ * @return string File name with a different extension.
+ */
+function webp_uploads_generate_filename( WP_Image_Editor $editor, string $file, string $size, string $extension ): string {
+	$ext = pathinfo( $file, PATHINFO_EXTENSION );
+	if ( $extension === $ext ) {
+		return $file;
```
westonruter, line 198: Could test coverage be added for this?

plugins/webp-uploads/helper.php
```diff
@@ -183,6 +180,36 @@ function webp_uploads_generate_additional_image_source( $attachment_id, $image_s
...
+	}
+
+	$dir    = trailingslashit( pathinfo( $file, PATHINFO_DIRNAME ) );
+	$name   = wp_basename( $file, ".$ext" );
+	$suffix = 'full' === $size ? '' : '-' . $editor->get_suffix();
+
+	// Add "-{$ext}" to the filename if an image with the same name already exists.
+	$file = "{$dir}{$name}{$suffix}.{$extension}";
+	if ( file_exists( $file ) ) {
+		$file = "{$dir}{$name}{$suffix}-{$ext}.{$extension}";
```
westonruter, line 208: Could test coverage be added for this as well?

westonruter (changes requested):
Running unit tests for this locally, I'm getting a single failure:

```
There was 1 failure:

1) WebP_Uploads_Helper_Tests::it_should_add_original_image_extension_to_the_webp_file_name_to_ensure_it_is_unique with data set #0 ('/var/www/html/wp-content/plug...e.jpeg', '/var/www/html/wp-content/plug...ge.jpg')
Failed asserting that 'image-35-300x300.webp' ends with "-1-300x300.webp".

/var/www/html/wp-content/plugins/performance/tests/plugins/webp-uploads/helper-tests.php:576
```

This is confusing because there is no error in the GHA-run tests.

@adamsilverstein @thelovekesh This seems somehow related to WebP and/or env issues that I seem to recall you mentioning?

thelovekesh:
> This is confusing because there is no error in the GHA-run tests.

@westonruter what environment are you using for this?

westonruter:
@thelovekesh I'm testing locally with wp-env which is supposed to be the same as on GHA.

adamsilverstein (commented):
awaiting changes or refinement

plugins/webp-uploads/helper.php
```diff
@@ -184,6 +181,36 @@ function webp_uploads_generate_additional_image_source( int $attachment_id, stri
...
+ *
+ * @since n.e.x.t
+ *
+ * @param WP_Image_Editor $editor    The image editor that handle the image manipulation.
+ * @param string          $file      Full path to the image file.
+ * @param string          $size      The image size.
+ * @param string          $extension The image extension.
+ * @return string File name with a different extension.
+ */
+function webp_uploads_generate_filename( WP_Image_Editor $editor, string $file, string $size, string $extension ): string {
```
adamsilverstein, line 195: can we use core's [helper](https://developer.wordpress.org/reference/classes/wp_image_editor/generate_filename/) instead?

rilwis: Please see the details in the PR description.

adamsilverstein: Ah sorry, I missed why you were using a custom approach.

phanduynam:
Has this issue been resolved?

reteinformatica:
> Has this issue been resolved?

NO.

phanduynam:
I think we should focus on solving this problem because it has existed for a long time.

phanduynam:
please update avif image format also

phanduynam:
@adamsilverstein please update more avif format images

KautsarJuhari:
Yeah, because of this system my WP gallery is a mess now. the problem is because the added -prefix.webp make an image mention in block editor not point out correctly

phanduynam:

adamsilverstein:
>  I think we should focus on solving this problem because it has existed for a long time.

The PR has a couple of outstanding comments that need to be resolved before this can be merged. This is a very low priority issue that is _not user facing_ and only affects file names as far as I can tell.

>  Yeah, because of this system my WP gallery is a mess now. the problem is because the added -prefix.webp make an image mention in block editor not point out correctly

@KautsarJuhari can you clarify this comment? file naming should in no way impact functionality of the gallery - the image meta data links to the images no matter the name.

PR #1039, opened 2024-03-08
Implement ILO_URL_Metrics_Group_Collection as iterator

westonruter:
## Summary

In follow-up to #1010 which implemented `ILO_URL_Metrics_Group` as an iterator, this PR does the same for `ILO_URL_Metrics_Group_Collection` for consistency.

It also fixes a typo discovered by @felixarntz in https://github.com/WordPress/performance/commit/34e3d062a2812458f4b1270a609d3238251dab84#r139510810.

Lastly, a `ILO_URL_Metric::get_viewport_width()` convenience getter method is added since we only are needing to get the width and not the height (so far).

## Checklist

### DIFF
--- a/modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
+++ b/modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
@@ -230,6 +230,9 @@ final class ILO_URL_Metrics_Group_Collection implements Countable, IteratorAggre
 	 * @return ILO_URL_Metric[] All URL metrics.
 	 */
 	public function get_flattened_url_metrics(): array {
+		// The duplication of iterator_to_array is not a mistake. This collection is an
+		// iterator and the collection contains iterator instances. So to flatten the
+		// two levels of iterators we need to nest calls to iterator_to_array().
 		return array_merge(
 			...array_map(
 				'iterator_to_array',
--- a/modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
+++ b/modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
@@ -152,7 +152,7 @@ final class ILO_URL_Metrics_Group_Collection implements Countable, IteratorAggre
 	 */
 	public function add_url_metric( ILO_URL_Metric $new_url_metric ) {
 		foreach ( $this->groups as $group ) {
-			if ( $group->is_viewport_width_in_range( $new_url_metric->get_viewport()['width'] ) ) {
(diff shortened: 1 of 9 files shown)


modules/images/image-loading-optimization/class-ilo-url-metrics-group.php
```diff
@@ -79,7 +79,7 @@ public function __construct( array $url_metrics, int $minimum_viewport_width, in
 		}
 		if ( $minimum_viewport_width >= $maximum_viewport_width ) {
 			throw new InvalidArgumentException(
-				esc_html__( 'The minimum viewport width must be larger than the maximum viewport width.', 'performance-lab' )
+				esc_html__( 'The minimum viewport width must smaller larger than the maximum viewport width.', 'performance-lab' )
```
felixarntz: 😁 
```suggestion
				esc_html__( 'The minimum viewport width must be smaller than the maximum viewport width.', 'performance-lab' )
```

modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
```diff
@@ -239,11 +232,27 @@ public function is_every_group_complete(): bool {
 	public function get_flattened_url_metrics(): array {
 		return array_merge(
 			...array_map(
-				static function ( ILO_URL_Metrics_Group $group ): array {
-					return iterator_to_array( $group );
-				},
-				$this->groups
+				'iterator_to_array',
+				iterator_to_array( $this )
```
felixarntz, line 239: Maybe worth adding a comment here to explain because someone may easily think there's some duplication mistake here - which I know is not the case but may not be obvious.

felixarntz (approved):
@westonruter Looks great, except for the typo fix. Once you've addressed that, good to go IMO.

modules/images/image-loading-optimization/class-ilo-url-metrics-group.php
westonruter: Thanks. I need more sleep 😧

modules/images/image-loading-optimization/class-ilo-url-metrics-group-collection.php
westonruter: Good idea. It will look even more confusing once we update to PHP 8.1 with first-class callables:

```php
return array_merge(
	...array_map(
		iterator_to_array(...),
		iterator_to_array( $this )
	)
);
```

Added comment in 91c521eb2e65b3e20f3dcf8ca9bff523a6049f.

Issue #1030, opened 2024-03-06
Remove modules infrastructure


felixarntz:
Following #1029, there are effectively no longer any modules as part of Performance Lab. This should remain that way. As such, all modules infrastructure should be entirely removed.

Let's focus on the minimum effort needed to clean up in this issue. Further enhancements to user experience should be considered in separate issues.

Here is a non comprehensive list of what needs to be removed:
* All UI related to toggling modules (i.e. the checkboxes).
* All UI related to the module migration.
* All PHP functions related to module settings.
* All PHP functions related to loading modules, module activation/deactivation etc.
* All PHP functions related to migrating from modules to their standalone plugins.
* The `perflab_is_standalone_plugin_loaded()` PHP function (as it relates to modules).
* All unit test coverage related to module infrastructure.
* All CLI scripts in `bin` that deal with modules exclusively.
* All GitHub workflows that deal with modules exclusively (such as building and testing modules, publishing modules, ...).
* Utility files, like `module-i18n.php`, `default-enabled-modules.php`, and their usage and surrounding infrastructure.
* Any reference to modules, really :)

Additional changes to be made:
* The `plugins.json` file should no longer support a `"modules"` entry. Tooling (e.g. scripts and GitHub workflows) should be updated to no longer consider that `"modules"` entry. For example, the `deploy-standalone-plugins.yml` workflow should be updated to only work with the actual plugins in the `/plugins` folder and no longer reference modules, which means for example scripts like `get-plugin-dir` can be removed as it will now always be the `/plugins` directory.
* The `$source` parameter of the `perflab_get_standalone_plugin_version_constants()` function should be deprecated (or removed) and effectively always be for plugins.
* The generator tag should be adjusted to not include any information about modules. In other words, it output should look like `Performance Lab %1$s; plugins: %2$s` going forward.
* The copy of the default admin pointer should be adjusted to no longer say "performance features included in the plugin", but just "performance features".
* The main `readme.txt` should be updated to no longer list the available modules.
* The `readme` JS script should be adjusted to no longer try to populate the list of available modules into `readme.txt` (the changelog functionality should be kept though).

These lists may not be complete, and further things to remove/update may come up as a PR is being worked on.

felixarntz:
@mukeshpanchal27 @joemcgill @thelovekesh Let's work on this once #1029 is completed.

This should be a big PR, but almost only red lines, which is always a good thing :)

mukeshpanchal27:
[...]

PR #1041, opened 2024-03-09
Rename "Image Loading Optimization" to "Optimization Detective"

westonruter:
## Checklist


.github/CODEOWNERS
```diff
@@ -60,6 +60,6 @@
 /plugins/dominant-color-images              @pbearne @spacedmonkey
 /tests/plugins/dominant-color-images        @pbearne @spacedmonkey
 
-# Plugin: Image Loading Optimization
-/plugins/image-loading-optimization         @westonruter
-/tests/plugins/image-loading-optimization   @westonruter
+# Plugin: Optimization Detective
+/plugins/optimization-detective         @westonruter
+/tests/plugins/optimization-detective   @westonruter
```
mukeshpanchal27: Nit-pick, let's align the spacing with the other entries.

plugins/optimization-detective/can-load.php
```diff
@@ -3,7 +3,7 @@
  * Can load function to determine if ILO can load.
```
mukeshpanchal27: Do we still need `can-load` for plugin?

plugins/optimization-detective/class-od-data-validation-exception.php
```diff
@@ -1,8 +1,8 @@
 <?php
 /**
- * Image Loading Optimization: ILO_Data_Validation_Exception class
+ * Optimization Detective: OD_Data_Validation_Exception class
  *
- * @package image-loading-optimization
+ * @package optimization-detective
  * @since n.e.x.t
```
mukeshpanchal27: `n.e.x.t` should be updated to a specific version number.

.github/CODEOWNERS
westonruter: Updated in 44a032f9738d803867a580617be92de387a765ba

plugins/optimization-detective/can-load.php
westonruter: Good point! This is only for modules. Removed in 2106b9a4e60a8e6b9e5e469c5c7908c352c3a0fd.

plugins/optimization-detective/class-od-data-validation-exception.php
westonruter: Updated in 30e6b9601c4e37088589c7627bc5f40e99f62a25.

But when do we use `n.e.x.t` then?

felixarntz (commented):
I haven't reviewed the PR, but we just had a conversation about our plugin names this morning, and this name here isn't necessarily a good idea IMO.

I just opened https://github.com/WordPress/performance/issues/1046 to discuss more holistically.

plugins/optimization-detective/class-od-data-validation-exception.php
mukeshpanchal27: Apologies for the delayed response. The current `npm run since -- -r {version}` command runs on the entire file system and updates the same version. Therefore, if we use this command for Performance Lab file changes, it will also update the same version in other plugins, such as this one. We need to enhance that command to update the version for the plugin based on its configuration. In the meantime, we should use the actual version.

.github/CODEOWNERS
```diff
@@ -60,6 +60,6 @@
 /plugins/dominant-color-images              @pbearne @spacedmonkey
 /tests/plugins/dominant-color-images        @pbearne @spacedmonkey
 
-# Plugin: Image Loading Optimization
-/plugins/image-loading-optimization         @westonruter
-/tests/plugins/image-loading-optimization   @westonruter
+# Plugin: Optimization Detective
+/plugins/optimization-detective             @westonruter
+/tests/plugins/optimization-detective       @westonruter
```
felixarntz: Feel free to add me as second code owner here. While I've mostly been just supporting/reviewing, I'd like to remain involved on this one.

load.php
```diff
@@ -330,12 +330,12 @@ function perflab_get_standalone_plugin_version_constants( $source = 'plugins' )
...
 	return array(
-		'webp-uploads'               => 'WEBP_UPLOADS_VERSION',
-		'dominant-color-images'      => 'DOMINANT_COLOR_IMAGES_VERSION',
-		'image-loading-optimization' => 'IMAGE_LOADING_OPTIMIZATION_VERSION',
-		'performant-translations'    => 'PERFORMANT_TRANSLATIONS_VERSION',
-		'auto-sizes'                 => 'IMAGE_AUTO_SIZES_VERSION',
-		'speculation-rules'          => 'SPECULATION_RULES_VERSION',
+		'webp-uploads'            => 'WEBP_UPLOADS_VERSION',
+		'dominant-color-images'   => 'DOMINANT_COLOR_IMAGES_VERSION',
+		'optimization-detective'  => 'OPTIMIZATION_DETECTIVE_VERSION',
```
felixarntz: Per the #1046 discussion, let's remove this from here for now, as having it here will lead it to show up in the PL recommended features. The only other purpose for this function was to check for the constant to avoid conflict between a plugin and its equivalent module, which is no longer relevant as of #1030, and in case of Optimization Detective wouldn't be relevant anyway since it's always only going to have been a plugin.

Maybe instead put a TODO comment here to add it once deemed stable enough / split into image loading optimization feature plugin.

plugins/optimization-detective/load.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Plugin Name: Image Loading Optimization
+ * Plugin Name: Optimization Detective
  * Plugin URI: https://github.com/WordPress/performance/issues/869
  * Description: Improves accuracy of optimizing the loading of the LCP image by leveraging client-side detection with real user metrics. Also enables output buffering of template rendering which can be filtered.
```
felixarntz, line 5: Can we change this to be more focused on the infrastructure / API part of this, as discussed in #1046? The image loading optimization part is more of a proof of concept use-case for the infrastructure.

This applies to `readme.txt` as well.

felixarntz (commented):
@westonruter Almost LGTM. Just a few small things.

plugins/optimization-detective/load.php
westonruter: I was thinking we'd do that after the plugin review to avoid confusion for the plugin review team. So the first version we actually commit to SVN would have that.

felixarntz: WFM. :)

.github/CODEOWNERS
westonruter:
```suggestion
/plugins/optimization-detective             @westonruter @felixarntz
/tests/plugins/optimization-detective       @westonruter @felixarntz
```

load.php
westonruter: Like so?

```suggestion
		// TODO: Add image loading optimization plugin, dependent of Optimization Detective, once ready for end users.
```

felixarntz: SGTM!

felixarntz (approved):

PR #1042, opened 2024-03-09
Migrate Site Health checks from being modules to becoming part of the plugin core

### LINKED ISSUE #1029: Make remaining modules (all Site Health) part of Performance Lab core
As recently discussed (also see https://github.com/WordPress/performance/issues/911#issuecomment-1974233702), the concept of modules will be entirely removed as part of the 3.0.0 release.

At this point, all features are already moved to standalone plugins exclusively, except for the Site Health modules where this was an explicit design decision, to keep features that don't directly enhance performance part of Performance Lab. This philosophy remains, with a slight adjustment of no longer using modules:

**Any features that directly enhance performance should be standalone plugins.**
**Any features that don't directly enhance performance (such as measurement features or recommendations like Site Health) should be part of Performance Lab core.**

Site Health checks should continue to remain part of Performance Lab. However, going forward, this should no longer be in the form of modules, but simply as part of Performance Lab core. This means:
* They'll be permanently active, as long as Performance Lab is active.
* There won't be a UI to control them. Their only UI will be the respective check itself within Site Health.
* They should still be implemented in distinct folders, in a standalone manner.
* They can still be ported over to WordPress core, once deemed eligible. At that point, they would be removed from Performance Lab.

thelovekesh:
## Summary

Fixes #1029

## Checklist


includes/site-health/load.php
```diff
@@ -0,0 +1,27 @@
...
+$site_health_tests = array(
+	'audit-autoloaded-options',
+	'audit-enqueued-assets',
+	'webp-support'
+);
+
+foreach ( $site_health_tests as $test ) {
+	require_once __DIR__ . '/' . $test . '/helper.php';
+	require_once __DIR__ . '/' . $test . '/hooks.php';
+}
```
felixarntz: I think this is a bit "too smart".

We don't want to mandate which files a Site Health check should have. For that reason, I think it's better to make this a simple list of the 6 `require_once`s necessary, so that _any_ new file can be simply added to that list.

Also there's a PHPCS lint error in the array, but that shouldn't be relevant anymore if you apply my suggested changes.

includes/site-health/load.php
```diff
@@ -0,0 +1,27 @@
+<?php
+/**
+ * Site Health tests loader.
```
felixarntz: Nit-pick: Those are commonly referred to as "checks", not "tests". The latter could be confused with unit test coverage.
```suggestion
 * Site Health checks loader.
```

load.php
```diff
@@ -30,10 +30,14 @@
...
-require_once PERFLAB_PLUGIN_DIR_PATH . 'server-timing/class-perflab-server-timing.php';
-require_once PERFLAB_PLUGIN_DIR_PATH . 'server-timing/load.php';
-require_once PERFLAB_PLUGIN_DIR_PATH . 'server-timing/defaults.php';
+// Load server-timing package.
+require_once PERFLAB_PLUGIN_DIR_PATH . 'includes/server-timing/class-perflab-server-timing-metric.php';
+require_once PERFLAB_PLUGIN_DIR_PATH . 'includes/server-timing/class-perflab-server-timing.php';
+require_once PERFLAB_PLUGIN_DIR_PATH . 'includes/server-timing/load.php';
+require_once PERFLAB_PLUGIN_DIR_PATH . 'includes/server-timing/defaults.php';
+
+// Load site health tests.
```
felixarntz: See above.
```suggestion
// Load Site Health checks.
```

load.php
```diff
@@ -30,10 +30,14 @@
 	define( 'PERFLAB_OBJECT_CACHE_DROPIN_VERSION', false );
 }
 
-require_once PERFLAB_PLUGIN_DIR_PATH . 'server-timing/class-perflab-server-timing-metric.php';
-require_once PERFLAB_PLUGIN_DIR_PATH . 'server-timing/class-perflab-server-timing.php';
-require_once PERFLAB_PLUGIN_DIR_PATH . 'server-timing/load.php';
-require_once PERFLAB_PLUGIN_DIR_PATH . 'server-timing/defaults.php';
+// Load server-timing package.
```
```suggestion
// Load Server-Timing API.
```

tests/load-tests.php
```diff
@@ -331,6 +301,6 @@ static function () {
 		WP_Filesystem();
 
 		// Simulate that the original object-cache.copy.php file exists.
-		$wp_filesystem->put_contents( PERFLAB_PLUGIN_DIR_PATH . 'server-timing/object-cache.copy.php', file_get_contents( PERFLAB_PLUGIN_DIR_PATH . 'server-timing/object-cache.copy.php' ) );
+		$wp_filesystem->put_contents( PERFLAB_PLUGIN_DIR_PATH . 'includes/server-timing/object-cache.copy.php', file_get_contents( PERFLAB_PLUGIN_DIR_PATH . 'includes/server-timing/object-cache.copy.php' ) );
```
felixarntz, line 252: This looks good, but there's still another production reference to this file that you need to update: Please check the main PL `load.php` file. It still references the old location, causing a unit test to fail.

felixarntz (commented):
@thelovekesh This mostly looks good. Please check the below comments.

Looking at the unit test failures, only the 5th failure is actually valid. The other tests just fail because there aren't any modules anymore, which is fine. Please remove those tests.

includes/site-health/load.php
thelovekesh: I'm also thinking the same. Thanks for pointing in the right direction.

tests/load-tests.php
thelovekesh: I have fixed the path for the server-timing API. 

@felixarntz I am not sure of all the testing steps for this API. Would you please check if it's working on your local? Else please share the steps and I'll test it as per requirements.

felixarntz: @thelovekesh The logic places this file to have parts of the Server-Timing API eligible to load sooner. The extra checks for placing the file though are needed since some sites may already have an `object-cache.php` drop-in present, or the server environment may prevent writing that file. So all of this has to be handled gracefully.

includes/server-timing/object-cache.copy.php
```diff
@@ -58,17 +58,17 @@ function perflab_load_server_timing_api_from_dropin() {
...
 		}
 
-		require_once $plugin_dir . 'server-timing/class-perflab-server-timing-metric.php';
-		require_once $plugin_dir . 'server-timing/class-perflab-server-timing.php';
-		require_once $plugin_dir . 'server-timing/load.php';
-		require_once $plugin_dir . 'server-timing/defaults.php';
+		require_once $plugin_dir . 'includes/server-timing/class-perflab-server-timing-metric.php';
+		require_once $plugin_dir . 'includes/server-timing/class-perflab-server-timing.php';
+		require_once $plugin_dir . 'includes/server-timing/load.php';
+		require_once $plugin_dir . 'includes/server-timing/defaults.php';
```
felixarntz, line 71: @thelovekesh Now that I see this change, moving the files for the Server-Timing API is tricky from a backward compatibility perspective. Here's why:
* The `wp-content/object-cache.php` drop-in may already be placed on some sites, from an older Performance Lab version. As such, it would try to load the old paths, which wouldn't work. It wouldn't cause a fatal error because of the checks above, but it just wouldn't load the API.
* In order to fix that, we'll have to enhance the logic that places that drop-in file (i.e. copies from `object-cache.copy.php` to the actual drop-in) so that it detects whether the current file is outdated.
* I'm happy to open another PR for that. However, at a minimum, let's update the drop-in version number in this PR from `2` to `3` (in both the file header and constant).

felixarntz: @thelovekesh I've opened #1047 for the purpose of updating an older version of the drop-in, please take a look. Depending which of the two PRs gets merged first, the other one will require a minor update (to sync the version numbers and file paths).

includes/site-health/load.php
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @package performance-lab
+ * @since n.e.x.t
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+if ( ! has_filter( 'user_has_cap', 'wp_maybe_grant_site_health_caps' ) ) {
```
mukeshpanchal27:
```suggestion
// Bail early if Site Health feature is disabled.
if ( ! has_filter( 'user_has_cap', 'wp_maybe_grant_site_health_caps' ) ) {
```

mukeshpanchal27 (commented):
Thanks for the PR! can we remove `https://github.com/WordPress/performance/tree/enhancement/move-modules-to-pl-core/tests/testdata/demo-modules` directory.

felixarntz (approved):
@thelovekesh Thank you for the updates. Other than the remaining feedback from @mukeshpanchal27, LGTM!

includes/site-health/load.php
swissspidy: Why are we even using this check and not actually `current_user_can( 'view_site_health_checks' )`?

felixarntz: This is too early to check `current_user_can()`. It's not impossible, but not a good practice since the user is typically only initialized before `init`.

That said, I personally think we can remove this check entirely. It was present before to prevent users from activating useless modules, but that is no longer relevant. The filters that are added in the required files below aren't harmful even if Site Health is disabled. So I'd suggest to remove the check entirely.

thelovekesh:
> Thanks for the PR! can we remove https://github.com/WordPress/performance/tree/enhancement/move-modules-to-pl-core/tests/testdata/demo-modules directory.

@mukeshpanchal27 As discussed in today's performance meeting, we will remove modules in a separate PR.

includes/site-health/load.php
thelovekesh: Removed in https://github.com/WordPress/performance/pull/1042/commits/88cd219b4b13d27f1045567bcb8d5d13091db1c9

thelovekesh:
@mukeshpanchal27 Can you please re-review this PR?

swissspidy (approved):

mukeshpanchal27 (approved):

PR #1043, opened 2024-03-10
Improve URL handling in Optimization Detective

westonruter:
* Compute the current URL server-side instead of letting client specify the URL in JavaScript. Previously there was no validation that the supplied URL is actually valid or that it corresponds to the server-computed slug from the normalized query vars. True the URL is not used except as metadata, but it was not ideal that proper validation was not being performed. Now the validity of the supplied URL is guaranteed as it is used alongside the slug in the computation of the nonce.
* Each stored URL metric now includes the original URL for which the data was gathered, being stored alongside the viewport, timestamp, and elements. The URL is also stored as the `post_title` of the `od-url-metrics` post so there is apparently duplication with the URLs stored in the `post_content`. However, this may not be the case since the URL metrics being stored the post are grouped because they share the same normalized query vars. Each of the URL metrics have have a slightly different actual URL (although they should share the same canonical URL). Storing the original URL with each URL metric can assist with debugging issues with query var normalization.
* Removes JSON pretty-printing of `post_content` for URL Metrics posts.
* Removed TODO for adding a version identifier for the stored URL metrics since strict schema validation now achieves this by dropping entries which no longer match the schema.

## Checklist


plugins/optimization-detective/storage/data.php
```diff
@@ -85,9 +85,49 @@ function od_get_normalized_query_vars(): array {
...
+ *
+ * This is essentially the REQUEST_URI prefixed by the scheme and host for the home URL.
+ * This is needed in particular due to subdirectory installs.
+ *
+ * @since n.e.x.t
+ * @access private
+ *
+ * @return string Current URL.
+ */
+function od_get_current_url(): string {
```
westonruter, line 99: Inspired by the same function in the AMP plugin: https://github.com/ampproject/amp-wp/blob/5708e19c60cf0f88ce522349fba955266b6892f6/includes/amp-helper-functions.php#L617-L662

plugins/optimization-detective/storage/data.php
```diff
@@ -85,9 +85,49 @@ function od_get_normalized_query_vars(): array {
...
+	$current_url = $parsed_url['scheme'] . '://' . $parsed_url['host'];
+	if ( isset( $parsed_url['port'] ) ) {
+		$current_url .= ':' . $parsed_url['port'];
+	}
+	$current_url .= '/';
+
+	if ( isset( $_SERVER['REQUEST_URI'] ) ) {
+		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+		$current_url .= ltrim( wp_unslash( $_SERVER['REQUEST_URI'] ), '/' );
+	}
```
felixarntz, line 122: Thinking about my favorite edge-case of sites within a subdirectory, this works as expected, right? I guess since `$_SERVER['REQUEST_URI']` would include the _full_ path, including the part that's potentially part of `home_url()`, this should be fine. Just want to make sure since the function doesn't otherwise consider a path that may be part of `home_url()` already.

plugins/optimization-detective/storage/data.php
```diff
@@ -85,9 +85,49 @@ function od_get_normalized_query_vars(): array {
...
+	return esc_url_raw( $current_url );
```
felixarntz, line 123: Why does this use `esc_url_raw()`?

plugins/optimization-detective/storage/post-type.php
```diff
@@ -190,7 +190,7 @@ static function ( OD_URL_Metric $url_metric ): array {
 			},
 			$group_collection->get_flattened_url_metrics()
 		),
-		JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES // TODO: No need for pretty-printing.
+		JSON_UNESCAPED_SLASHES // No need for escaped slashes since not printed to frontend.
```
felixarntz, line 193: Completely unrelated question: Why is this relevant? Do frontend slashes have to be escaped? I'm asking since the Speculation Rules plugin JSON output contains backslashes which Domenic suggested to remove in my Speculation Rules post draft code example. I'm confused about whether those _should_ or _shouldn't_ be there. 🤔

felixarntz (approved):
@westonruter LGTM, just a few small questions.

plugins/optimization-detective/storage/post-type.php
westonruter: Good question. The slashes should always be present when the JSON is inline in HTML, as this prevents malicious strings from breaking out of the containing `script`. For example, if someone somehow got `</script>` into the data being serialized to JSON, this would prematurely terminate the `script` element, whereas if slashes are escaped, then `<\/script>` would not.

In the context of `post_content` here where the JSON is never being printed to HTML, there is no need for this escaping.

westonruter: Er, so that should probably be restored to Speculation Rules 😊

plugins/optimization-detective/storage/data.php
westonruter: Yes, `REQUEST_URI` should include the full path including the subdirectory root path. Otherwise, this function here only uses `home_url()` to obtain the default `scheme`, `host`, and `port`. The `path` is not used.

plugins/optimization-detective/storage/data.php
westonruter: Well, humm, since potentially malicious data is present in `REQUEST_URI`. Perhaps this is overkill, but it has worked well in the AMP plugin.

felixarntz: I would think this is already handled by `$wpdb->prepare()` when storing the value. But in any case, doesn't hurt I guess.

plugins/optimization-detective/storage/post-type.php
felixarntz: I see, thanks for explaining. I believe Speculation Rules already does this correctly then, feel free to check. I was referring to the Make Core post draft I've been working on.

PR #1044, opened 2024-03-11
Remove branch restrictions so workflows can run on PRs for any branch

### LINKED ISSUE #998: Expand workflow and linting checks to run on all branches
Follow-up: https://github.com/WordPress/performance/pull/972#discussion_r1486605682

Currently, the test plugin and linting check workflow are configured to run only on specific branches, potentially leading to sub-PRs not receiving the necessary checks. It has been observed that certain sub-PRs #984 and #983 are not triggering checks, causing delays and potential oversights in the review process. To address this issue and ensure consistent code quality across all branches, we propose expanding the workflow and linting checks to run on all branches.

cc. @westonruter

westonruter:
## Summary

This will ensure that checks run on sub-PRs for feature branches.

Fixes #998

## Checklist

### DIFF
--- a/.github/workflows/js-lint.yml
+++ b/.github/workflows/js-lint.yml
@@ -16,10 +16,6 @@ on:
       - '**/package.json'
       - 'package-lock.json'
   pull_request:
-    branches:
-      - trunk
-      - 'release/**'
-      - 'feature/**'
     # Only run if JS/JSON/Lint/NVM files changed.
     paths:
       - '.github/workflows/js-lint.yml'
--- a/.github/workflows/php-lint.yml
+++ b/.github/workflows/php-lint.yml
@@ -14,10 +14,6 @@ on:
       - 'composer.json'
       - 'composer.lock'
   pull_request:
-    branches:
-      - trunk
-      - 'release/**'
-      - 'feature/**'
     # Only run if PHP-related files changed.
     paths:
       - '.github/workflows/php-lint.yml'
--- a/.github/workflows/php-test-plugins.yml
+++ b/.github/workflows/php-test-plugins.yml
@@ -17,10 +17,6 @@ on:
       - 'composer.json'
       - 'composer.lock'
   pull_request:
-    branches:
-      - trunk
-      - 'release/**'
-      - 'feature/**'
     # Only run if PHP-related files changed.
     paths:
       - '.github/workflows/php-test-plugins.yml'
--- a/.github/workflows/php-test-standalone-plugins.yml
+++ b/.github/workflows/php-test-standalone-plugins.yml
@@ -17,6 +17,10 @@ on:
       - 'composer.json'
       - 'composer.lock'
   pull_request:
+    branches:
+      - trunk
+      - 'release/**'
+      - 'feature/**'
     # Only run if PHP-related files changed.
     paths:
       - '.github/workflows/php-test-standalone-plugins.yml'
--- a/.github/workflows/php-test-standalone-plugins.yml
+++ b/.github/workflows/php-test-standalone-plugins.yml
@@ -17,10 +17,6 @@ on:
       - 'composer.json'
       - 'composer.lock'
   pull_request:
-    branches:
-      - trunk
-      - 'release/**'
-      - 'feature/**'
     # Only run if PHP-related files changed.
     paths:
       - '.github/workflows/php-test-standalone-plugins.yml'
--- a/.github/workflows/php-test.yml
+++ b/.github/workflows/php-test.yml
@@ -17,10 +17,6 @@ on:
       - 'composer.json'
       - 'composer.lock'
   pull_request:
-    branches:
-      - trunk
-      - 'release/**'
-      - 'feature/**'
     # Only run if PHP-related files changed.
     paths:
       - '.github/workflows/php-test.yml'
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -2,10 +2,6 @@ name: Pull Request Validation
 
 on:
   pull_request:
-    branches:
-      - trunk
-      - 'release/**'
-      - 'feature/**'
     types:
       - labeled
       - unlabeled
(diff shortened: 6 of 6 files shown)


thelovekesh (approved):

felixarntz (commented):
@westonruter Overall makes sense. However, do we need all of those to run on "random" PRs? Some of those checks are most relevant on PRs against `trunk` and feature branches - I'm not even sure how common of a use-case it is to have PRs that _aren't_ against those kinds of branches.

Anyways, I'd say linting and unit testing probably makes sense to have no restriction on. But the plugin testing and PR validation, not sure. Sure, we could run all, but there's also a time and resource cost to this, so I'd say let's limit it to what's truly useful.

westonruter:
@felixarntz The use case is very common for me. To prevent getting blocked and to avoid overly-large pull requests, I open sub-PRs. In this case, it is very useful to have all checks run as if the PR was opened against `trunk` or a feature branch.

What other random PRs do you have in mind?

felixarntz:
@westonruter 
> @felixarntz The use case is very common for me. To prevent getting blocked and to avoid overly-large pull requests, I open sub-PRs. In this case, it is very useful to have all checks run as if the PR was opened against `trunk` or a feature branch.

Makes sense. In an ideal world, such sub PRs wouldn't be necessary, but I agree they're sometimes better than getting blocked. When you say "sub-PR", this still means the PR will eventually be reviewed against the "real" base branch though correct? As far as I understand, you only create those PRs to be able to continue working while the PRs that have to be merged first are still waiting for review/approvals. Is that correct? Something I'd like to avoid for instance is that those "sub-PRs" would be the foundation for actual review processes even before the previous PR gets merged. That could get messy very easily. For review, it's always better to have the previous work merged so that the actual changes are obvious to review.

> What other random PRs do you have in mind?

Nothing really, with "random PRs" I just meant any PRs that are not against `trunk` or feature branches :)

westonruter:
> When you say "sub-PR", this still means the PR will eventually be reviewed against the "real" base branch though correct? As far as I understand, you only create those PRs to be able to continue working while the PRs that have to be merged first are still waiting for review/approvals. Is that correct?

That is correct.

felixarntz:
In terms of the workflows, I think some workflows are better suited for once the PR is actually ready for review (e.g. the label/milestone check, I don't see its value on a PR that's effectively not ready for review anyway). As in:
* Let's say you have PR branch `a` against `feature/something`. Good to review.
* While waiting for feedback on `a`, you create branch `b` and open a PR from `b` against `a`. At this point, having a few basic things like linting and unit testing is probably still helpful. Broader things like standalone plugin testing or correct label usage though seems overkill.
* At some point the branch from `a` against `feature/something` gets merged.
* This also means the PR for `b` can be updated to be against `feature/something`. At this point it's ready for full review, and the complete suite of workflows makes sense to run.

LMKWYT.

westonruter:
@felixarntz OK, I've reverted the changes to the `pr-validation` and `php-test-standalone-plugins` workflow.

TBH, I hadn't before looked deeply at `npm run test-php` versus `npm run test-php-plugins`, except that I ran the latter once when I intended to run the former. I'm not totally clear yet on what standalone plugin testing is doing. Is it truly just doing this one single assertion for each plugin?

https://github.com/WordPress/performance/blob/06d6ffeb05fb5c7d3bb67253b21b924b62fe2a9e/plugin-tests/tests/test-standalone.php#L13-L15

It seems there would be a much simpler way to go about testing for a fatal, and that would be to not use PHPUnit at all but rather to:

1. Install all of the plugins
2. Activate each plugin one-by-one using WP-CLI
3. When each plugin is active, do a `curl` on the homepage to make sure it returns a 200 status code and `</html>`.

In the future, it would be better to use Playwright to do some actual E2E testing rather than a simple `curl`, like @swissspidy has [blogged about](https://pascalbirchler.com/wordpress-performance-testing/).

As it stands right now, the existing standalone tests appear to be broken. As seen in [this job](https://github.com/WordPress/performance/actions/runs/8184134244/job/22378090944#step:9:15):

```
Error copying plugin "undefined": ENOENT: no such file or directory, lstat './plugins/undefined/'
Error copying plugin "undefined": ENOENT: no such file or directory, lstat './plugins/undefined/'
Error copying plugin "undefined": ENOENT: no such file or directory, lstat './plugins/undefined/'
Error copying plugin "undefined": ENOENT: no such file or directory, lstat './plugins/undefined/'
```

Nevertheless, the jobs are passing. So the `php-test-standalone-plugins` workflow is currently wasting resources.

mukeshpanchal27:
Thanks, @westonruter, for the PR. I had it on my list to open a PR, but you beat me to it.

The standalone tests break after the changes made in #1011. We should change the condition here https://github.com/WordPress/performance/blob/trunk/bin/plugin/commands/build-plugins.js#L29 so that it with skip the module part if we don't have module lists.

I will open a PR soon to address this issue. The tests should not work for modules but should work for plugins only. Both scripts should be updated accordingly.

mukeshpanchal27 (approved):

swissspidy:
> It seems there would be a much simpler way to go about testing for a fatal, and that would be to not use PHPUnit at all but rather to:

@westonruter I think that test is just a dummy so that we have at least 1 test there. It's not really trying to detect fatal errors or anything.

westonruter:
> I think that test is just a dummy so that we have at least 1 test there. It's not really trying to detect fatal errors or anything.

@swissspidy Right, I didn't mean that somehow `assertTrue( true )` could possibly fail the assertion. Indeed, PHPStan flags it as being an error:

> phpstan: Call to method PHPUnit\Framework\Assert::assertTrue() with true will always evaluate to true.

It seems PHPUnit isn't the right tool for the job here if we're just wanting to check for a plugin fatal error.

westonruter:
> It seems PHPUnit isn't the right tool for the job here if we're just wanting to check for a plugin fatal error.

Filed: https://github.com/WordPress/performance/issues/1051

PR #1047, opened 2024-03-11
Enhance `object-cache.php` drop-in placement logic to support updating to a newer version of the file

felixarntz:
## Summary

See https://github.com/WordPress/performance/pull/1042#discussion_r1520452415: Given that PR contains changes to the `object-cache.php` drop-in which require updating it in order for the underlying Server-Timing logic to still be loaded correctly, the logic to place that file needs to support _updating_ an older version of the file to its latest version that comes with Performance Lab.

So far, the logic was naively checking for purely the existence of the drop-in via its version constant, without checking whether that version is the latest version though.

## Relevant technical choices

* The logic bails if the PL drop-in is already present _and_ in the latest version (or newer, though that's unlikely).
* Otherwise, it enters the logic as before. The new change there is that, if there is an existing drop-in, it may be an older version of the Performance Lab one.
    * If the version constant is not set, for sure it's _not_ the PL drop-in.
    * Otherwise, it likely is the PL drop-in, though just to make sure this is verified by checking the file contents.
    * If it is indeed an older version of the PL drop-in, that file is deleted, so that afterwards the previously existing logic can copy over the new drop-in.
* A new constant `PERFLAB_OBJECT_CACHE_DROPIN_LATEST_VERSION` contains always the latest version of the drop-in. For that, we need a separate constant as we can't access the `PERFLAB_OBJECT_CACHE_DROPIN_VERSION` constant defined in the plugin's own `object-cache.copy.php` file, as that file is never actually loaded.
* A filter `perflab_object_cache_dropin_version` is introduced, purely to facilitate unit testing related to the `PERFLAB_OBJECT_CACHE_DROPIN_VERSION` constant. This filter is needed because constants can't be "un-defined", so code relying on constants is otherwise untestable.

## Checklist


load.php
```diff
@@ -410,7 +423,10 @@ function perflab_maybe_set_object_cache_dropin() {
 		$dropin_path = WP_CONTENT_DIR . '/object-cache.php';
 
 		/**
```
swissspidy: While we're here, multi-line inline comments should not use two asterisks

```suggestion
		/*
```

load.php
```diff
@@ -29,6 +29,7 @@
 if ( ! defined( 'PERFLAB_OBJECT_CACHE_DROPIN_VERSION' ) ) {
 	define( 'PERFLAB_OBJECT_CACHE_DROPIN_VERSION', false );
 }
+define( 'PERFLAB_OBJECT_CACHE_DROPIN_LATEST_VERSION', 2 );
```
swissspidy: Let's add a unit test that verifies `PERFLAB_OBJECT_CACHE_DROPIN_LATEST_VERSION` is identical to the value in `server-timing/object-cache.copy.php`

load.php
```diff
@@ -398,6 +394,23 @@ function perflab_maybe_set_object_cache_dropin() {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param int|bool $current_dropin_version The drop-in version as defined by the
+	 *                                         `PERFLAB_OBJECT_CACHE_DROPIN_VERSION` constant.
+	 */
+	$current_dropin_version = apply_filters( 'perflab_object_cache_dropin_version', PERFLAB_OBJECT_CACHE_DROPIN_VERSION );
+
+	// Bail if already placed in the latest version or newer.
+	if ( $current_dropin_version && $current_dropin_version >= PERFLAB_OBJECT_CACHE_DROPIN_LATEST_VERSION ) {
```
swissspidy, line 400: Will these always be integers? Otherwise `version_compare` would be more robust or future proof

load.php
```diff
@@ -398,6 +394,23 @@ function perflab_maybe_set_object_cache_dropin() {
...
+	 * Filters the value of the `object-cache.php` drop-in constant.
+	 *
+	 * This filter should not be used outside of tests.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param int|bool $current_dropin_version The drop-in version as defined by the
+	 *                                         `PERFLAB_OBJECT_CACHE_DROPIN_VERSION` constant.
+	 */
+	$current_dropin_version = apply_filters( 'perflab_object_cache_dropin_version', PERFLAB_OBJECT_CACHE_DROPIN_VERSION );
```
swissspidy: Do we really need this filter? Can't we just modify the file contents in tests?

load.php
felixarntz: Was thinking about that. I went with the simpler numeric comparison given we're just using integers. If you feel strongly about `version_compare()`, we can use that. I guess I've never used `version_compare()` with integers, does that work?

load.php
felixarntz: I don't think so. The file is loaded by PHP, so how would we change its file contents _and_ have it reflected by the already running PHP process? Since constants aren't testable, I think this is the only way.

load.php
felixarntz: Done in https://github.com/WordPress/performance/pull/1047/commits/56ec805a7925c6223394a547fba8c2e28b8b91e7

load.php
felixarntz: Added in https://github.com/WordPress/performance/pull/1047/commits/56ec805a7925c6223394a547fba8c2e28b8b91e7

load.php
swissspidy: It does work, yeah, even though the docs say it expects a string in the form of a typical version number.

I don't feel strongly about that, so feel free to leave as-is

load.php
swissspidy: Ah right, because the plugin always defines the constant, so it will be always defined when running tests:

```php
if ( ! defined( 'PERFLAB_OBJECT_CACHE_DROPIN_VERSION' ) ) {
	define( 'PERFLAB_OBJECT_CACHE_DROPIN_VERSION', false );
}
```

Do we really need that though? We can just use `defined()`.

felixarntz: That wouldn't really change anything in terms of testability. We need to be able to test how the logic behaves with e.g.:
* the constant being `false`
* the constant having the latest version number in it
* the constant having an older version number in it

This can't be accomplished without a filter around the constant check because we can't dynamically define (and undefine) constants in tests.

load.php
westonruter: Integer comparison seems fine to me. It would be analogous to `$wp_db_version`.

load.php
westonruter: Would it be possible to use [runkit7](https://www.php.net/manual/en/intro.runkit7.php) to override the constant in tests instead?

swissspidy: Right, I was thinking modifying the file on the filesystem to modify the constant.

felixarntz: @westonruter @swissspidy I'm not sure how either of those would work. FWIW, WP core also has a few filters around certain constants purely for testing purposes, so I think that's an _okay_ practice.

Unless there's a straightforward way around that, I think it would be better to consider such approaches in a separate issue/PR.

westonruter: In that case, maybe good to add an [`@internal`](https://docs.phpdoc.org/2.9/references/phpdoc/tags/internal.html) tag to the docblock. We did that for such filters in the AMP plugin that we wanted to make explicitly not intended for external use.

felixarntz: @westonruter Added in https://github.com/WordPress/performance/pull/1047/commits/6e4c0be0c7ed3f86ec16f21809a4f46ccbfc519c

tests/load-tests.php
```diff
@@ -313,6 +369,20 @@ public function test_perflab_object_cache_dropin_may_be_disabled_via_filter() {
 		$this->assertFalse( $wp_filesystem->exists( WP_CONTENT_DIR . '/object-cache.php' ) );
 	}
 
+	public function test_perflab_object_cache_dropin_version_matches_latest() {
+		$file_content = file_get_contents( PERFLAB_PLUGIN_DIR_PATH . 'server-timing/object-cache.copy.php' );
+
+		// Get the version from the file header and the constant.
+		preg_match( '/^ \* Version: (\d+)$/m', $file_content, $matches );
```
westonruter: Just as a sanity check, as this will make sure that the file actually contains the expected contents. If not, then possibly `(int) $matches[1]` below could evaluate to `0` since `$matches[1]` would be `null` (with a warning).

```suggestion
		$this->assertTrue( (bool) preg_match( '/^ \* Version: (\d+)$/m', $file_content, $matches ) );
```

tests/load-tests.php
```diff
@@ -313,6 +369,20 @@ public function test_perflab_object_cache_dropin_may_be_disabled_via_filter() {
...
+		$file_header_version = (int) $matches[1];
+		preg_match( '/define\( \'PERFLAB_OBJECT_CACHE_DROPIN_VERSION\', (\d+) \)\;/', $file_content, $matches );
```
westonruter:
```suggestion
		$this->assertTrue( (bool) preg_match( '/define\( \'PERFLAB_OBJECT_CACHE_DROPIN_VERSION\', (\d+) \)\;/', $file_content, $matches ) );
```

tests/load-tests.php
```diff
@@ -296,6 +296,62 @@ public function test_perflab_maybe_set_object_cache_dropin_with_conflict() {
...
+	public function test_perflab_maybe_set_object_cache_dropin_with_latest_version() {
+		global $wp_filesystem;
+
+		$this->set_up_mock_filesystem();
+
+		$latest_file_content = file_get_contents( PERFLAB_PLUGIN_DIR_PATH . 'server-timing/object-cache.copy.php' );
+		$wp_filesystem->put_contents( WP_CONTENT_DIR . '/object-cache.php', $latest_file_content );
+
+		// Simulate PL constant is set to the value from the current file.
+		preg_match( '/define\( \'PERFLAB_OBJECT_CACHE_DROPIN_VERSION\', (\d+) \)\;/', $latest_file_content, $matches );
```
westonruter:
```suggestion
		$this->assertTrue( (bool) preg_match( '/define\( \'PERFLAB_OBJECT_CACHE_DROPIN_VERSION\', (\d+) \)\;/', $latest_file_content, $matches ) );
```

westonruter (approved):
LGTM, but with some test assertion hardening.

mukeshpanchal27 (approved):

felixarntz:
PR is updated to address the remaining feedback as well as to cater for the `object-cache.copy.php` updates merged via #1042.

swissspidy (approved):

PR #1048, opened 2024-03-12
Support changing autoload value for largest autoloaded options in Site Health check

### LINKED ISSUE #889: Support changing autoload value for largest autoloaded options
As discussed a while ago (I believe initially at WCUS 2022 contributor day), the autoloaded options Site Health check needs to be actionable. While showing the largest autoloaded options is great, without a way to fix the issue it's not all that helpful.

The Site Health check should make use of the new `wp_set_option_autoload()` function introduced in WordPress 6.4 (see [this dev note post](https://make.wordpress.org/core/2023/10/17/new-option-functions-in-6-4/)) to allow disabling autoload for a given option. Any changed option should be stored persistently in a list so that WordPress keeps track of it. The Site Health check should also allow reverting any changed option (in case changing it causes problems).

## Requirements
All of the below requirements should only take effect _if_ the `wp_set_option_autoload()` function is available.

* The list of the largest autoloaded options should have a new column with a button to "Disable autoload".
* When the user clicks this button, the option's autoload value should be set to `no`, and the option's name should be added to the array of options for which the user changed the value (e.g. a new option `perflab_aao_modified_options`).
* If the `perflab_aao_modified_options` list is not empty, the list of these options should be displayed at the bottom of the Site Health check, with each option being accompanied with a "Revert to autoload" CTA link.
    * Important: This list must be displayed even when the check passes, as the user must at any time be able to revert a decision they made before.

mukeshpanchal27:
## Summary

Fixes #889

## Checklist


modules/database/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,45 @@ function perflab_aao_get_autoloaded_options_table() {
 	$autoload_summary = perflab_aao_query_autoloaded_options();
 
 	$html_table = sprintf(
-		'<table class="widefat striped"><thead><tr><th scope="col">%s</th><th scope="col">%s</th></tr></thead><tbody>',
+		'<table class="widefat striped"><thead><tr><th scope="col">%s</th><th scope="col">%s</th><th scope="col">%s</th></tr></thead><tbody>',
 		__( 'Option Name', 'performance-lab' ),
-		__( 'Size', 'performance-lab' )
+		__( 'Size', 'performance-lab' ),
+		__( 'Action', 'performance-lab' )
```
westonruter: These should actually all use `esc_html__()`.

modules/database/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,45 @@ function perflab_aao_get_autoloaded_options_table() {
...
 	);
 
 	foreach ( $autoload_summary as $value ) {
-		$html_table .= sprintf( '<tr><td>%s</td><td>%s</td></tr>', $value->option_name, size_format( $value->option_value_length, 2 ) );
+		$disable_button = sprintf( '<a class="button update-autoload" data-option-name="%s" data-autoload="no" data-value="%s" href="#">%s</a>', esc_attr( $value->option_name ), esc_attr( $value->option_value_length ), __( 'Disable Autoload', 'performance-lab' ) );
```
westonruter:
```suggestion
		$disable_button = sprintf( '<a class="button update-autoload" data-option-name="%s" data-autoload="no" data-value="%s" href="#">%s</a>', esc_attr( $value->option_name ), esc_attr( $value->option_value_length ), esc_html__( 'Disable Autoload', 'performance-lab' ) );
```

modules/database/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,45 @@ function perflab_aao_get_autoloaded_options_table() {
...
+
+	if ( empty( $perflab_aao_disabled_options ) ) {
+		return;
+	}
+
+	$html_table = sprintf(
+		'<br><br><table class="widefat striped"><thead><tr><th scope="col">%s</th><th scope="col">%s</th><th scope="col">%s</th></tr></thead><tbody>',
+		__( 'Option Name', 'performance-lab' ),
+		__( 'Size', 'performance-lab' ),
+		__( 'Action', 'performance-lab' )
```
westonruter:
```suggestion
		esc_html__( 'Option Name', 'performance-lab' ),
		esc_html__( 'Size', 'performance-lab' ),
		esc_html__( 'Action', 'performance-lab' )
```

modules/database/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,45 @@ function perflab_aao_get_autoloaded_options_table() {
...
+	);
+
+	foreach ( $perflab_aao_disabled_options as $option_name => $value ) {
+		$disable_button = sprintf( '<a class="button update-autoload" data-option-name="%s" data-autoload="yes" href="#">%s</a>', esc_attr( $option_name ), __( 'Revert to Autoload', 'performance-lab' ) );
```
westonruter:
```suggestion
		$disable_button = sprintf( '<a class="button update-autoload" data-option-name="%s" data-autoload="yes" href="#">%s</a>', esc_attr( $option_name ), esc_html__( 'Revert to Autoload', 'performance-lab' ) );
```

modules/database/audit-autoloaded-options/update-autoload.js
```diff
@@ -0,0 +1,36 @@
...
+			option_name: optionName,
+			autoload,
+			value,
+		};
+
+		$.ajax( {
+			url:
+				perflabAutoloadSettings.root +
+				'perflab-aao/v1/update-autoload/' +
+				optionName,
```
westonruter: Per above, I don't think this needs to be part of the route. It's already part of the `data` being sent anyway.

modules/database/audit-autoloaded-options/update-autoload.js
```diff
@@ -0,0 +1,36 @@
...
+			data,
+			success( response ) {
+				if ( response.success ) {
+					button.attr( 'disabled', true );
+				} else {
+					alert( response.message );
+				}
+			},
+			error( error ) {
+				alert( error.errorMessage );
```
westonruter: Can we use something other than `alert()` for the success/error messages. It doesn't have a great UX.

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,101 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+
+/**
+ * Register custom REST API route for updating autoload.
+ *
+ * @since n.e.x.t
+ */
+function perflab_aao_register_rest_route() {
+	register_rest_route(
+		'perflab-aao/v1',
+		'/update-autoload/(?P<option_name>[a-zA-Z0-9_-]+)',
```
westonruter: Perhaps the option name shouldn't be part of the route, as this could artificially limit which options are able to be manipulated. It's not necessary either, as the option name is being passed among the request body.

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,101 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+	if ( empty( $option_name ) || empty( $autoload ) ) {
+		return new WP_REST_Response(
+			array(
+				'success' => false,
+				'message' => __( 'Invalid option name or autoload value.', 'performance-lab' ),
+			),
+			400
+		);
+	}
+
```
westonruter: I think this could be missing some validation here. Shouldn't it first be checking to see if the option exists? If not, then it should return an error. Maybe this is overkill as the validation is handled by `wp_set_option_autoload()`.

westonruter (changes requested):

modules/database/audit-autoloaded-options/update-autoload.js
mukeshpanchal27: Nice catch.

modules/database/audit-autoloaded-options/update-autoload.js
mukeshpanchal27: Instead of displaying an `alert`, I append that error message after the button. Could you please take a look?

mukeshpanchal27:
@westonruter, I have addressed the review feedback. The PR is ready for another round of review. Thank you!

modules/database/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,45 @@ function perflab_aao_get_autoloaded_options_table() {
...
+
+	$html_table = sprintf(
+		'<br><br><table class="widefat striped"><thead><tr><th scope="col">%s</th><th scope="col">%s</th><th scope="col">%s</th></tr></thead><tbody>',
+		esc_html__( 'Option Name', 'performance-lab' ),
+		esc_html__( 'Size', 'performance-lab' ),
+		esc_html__( 'Action', 'performance-lab' )
+	);
+
+	foreach ( $perflab_aao_disabled_options as $option_name => $value ) {
+		$disable_button = sprintf( '<a class="button update-autoload" data-option-name="%s" data-autoload="yes" href="#">%s</a>', esc_attr( $option_name ), esc_html__( 'Revert to Autoload', 'performance-lab' ) );
```
westonruter: Question: Why is this a link and not just a `button`?

modules/database/audit-autoloaded-options/update-autoload.js
```diff
@@ -0,0 +1,45 @@
+/* eslint camelcase: "off" */
+/* global jQuery:false, perflabAutoloadSettings:false */
+
+( function ( $, document ) {
+	$( document ).on( 'click', '.update-autoload', function ( e ) {
+		e.preventDefault();
```
westonruter: If a `button[type=button]` were used, then this wouldn't be needed.

```suggestion
```

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -20,9 +20,119 @@
...
 	return $tests;
 }
 add_filter( 'site_status_tests', 'perflab_aao_add_autoloaded_options_test' );
+
+/**
+ * Register custom REST API route for updating autoload.
+ *
+ * @since n.e.x.t
+ */
+function perflab_aao_register_rest_route() {
```
westonruter: There aren't any tests for this yet, right?

modules/database/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,45 @@ function perflab_aao_get_autoloaded_options_table() {
...
-		__( 'Size', 'performance-lab' )
+		'<table class="widefat striped"><thead><tr><th scope="col">%s</th><th scope="col">%s</th><th scope="col">%s</th></tr></thead><tbody>',
+		esc_html__( 'Option Name', 'performance-lab' ),
+		esc_html__( 'Size', 'performance-lab' ),
+		esc_html__( 'Action', 'performance-lab' )
 	);
 
 	foreach ( $autoload_summary as $value ) {
-		$html_table .= sprintf( '<tr><td>%s</td><td>%s</td></tr>', $value->option_name, size_format( $value->option_value_length, 2 ) );
+		$disable_button = sprintf( '<a class="button update-autoload" data-option-name="%s" data-autoload="no" data-value="%s" href="#">%s</a>', esc_attr( $value->option_name ), esc_attr( $value->option_value_length ), esc_html__( 'Disable Autoload', 'performance-lab' ) );
```
westonruter: This could be a `button` instead of a link, right? (Same as below)

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -20,9 +20,119 @@
...
+function perflab_aao_update_autoload_rest( $request ) {
+	$option_name = $request->get_param( 'option_name' );
+	$autoload    = $request->get_param( 'autoload' );
+	$value       = $request->get_param( 'value' );
+
+	if ( empty( $option_name ) || empty( $autoload ) ) {
+		return new WP_REST_Response(
+			array(
+				'success' => false,
+				'message' => esc_html__( 'Invalid option name or autoload value.', 'performance-lab' ),
```
westonruter: This shouldn't be escaped in this context since it isn't where it is output. It should be escaped right when printed.

```suggestion
				'message' => __( 'Invalid option name or autoload value.', 'performance-lab' ),
```

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -20,9 +20,119 @@
...
+		);
+	}
+
+	// Check if the option exists.
+	if ( false === get_option( $option_name ) ) {
+		// Option doesn't exist, return an error or handle the situation accordingly.
+		return new WP_REST_Response(
+			array(
+				'success' => false,
+				'message' => esc_html__( 'The option does not exist.', 'performance-lab' ),
```
westonruter:
```suggestion
				'message' => __( 'The option does not exist.', 'performance-lab' ),
```

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -20,9 +20,119 @@
...
+			$modified_options[ $option_name ] = $value;
+		}
+		update_option( 'perflab_aao_modified_options', $modified_options );
+
+		return new WP_REST_Response(
+			array(
+				'success' => true,
+				'message' => sprintf(
+					/* translators: 1: Autoload value, 2: Option name */
+					esc_html__( 'Autoload %1$s for option %2$s.', 'performance-lab' ),
```
westonruter:
```suggestion
					__( 'Autoload %1$s for option %2$s.', 'performance-lab' ),
```

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -20,9 +20,119 @@
...
+					$option_name
+				),
+			),
+			200
+		);
+	} else {
+		return new WP_REST_Response(
+			array(
+				'success' => false,
+				'message' => esc_html__( 'Failed to update autoload status.', 'performance-lab' ),
```
westonruter:
```suggestion
				'message' => __( 'Failed to update autoload status.', 'performance-lab' ),
```

modules/database/audit-autoloaded-options/helper.php
westonruter: Also, I can see that the JS is unexpectedly adding an `disabled` attribute to an `a` element instead of a `button`.

modules/database/audit-autoloaded-options/update-autoload.js
```diff
@@ -0,0 +1,45 @@
...
+				} else {
+					button.append(
+						'<span class=""> ' + response.message + '</span>'
+					);
+					$(
+						'<div style="margin-top:5px;border:1px solid #d63638;padding-left:5px;">' +
+							response.message +
+							'</div>'
+					).insertAfter( button );
+				}
```
westonruter: It doesn't seem like the `else` condition ever occurs, since a non-200 status code is always returned for errors along with `success => false`. In fact, there's no need for sending `success` at all as the HTTP status can be relied on entirely.

Also, `attr()` isn't quite right here. It should be `prop()` here, although this only works with an actual `button` and not an `a` element.

```suggestion
				button.prop( 'disabled', true );
```

See https://api.jquery.com/prop/#:~:text=Attributes%20vs.%20Properties

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -20,9 +20,119 @@
...
+			400
+		);
+	}
+
+	// Check if the option exists.
+	if ( false === get_option( $option_name ) ) {
+		// Option doesn't exist, return an error or handle the situation accordingly.
+		return new WP_REST_Response(
+			array(
+				'success' => false,
```
westonruter:
```suggestion
```

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -20,9 +20,119 @@
...
+ */
+function perflab_aao_update_autoload_rest( $request ) {
+	$option_name = $request->get_param( 'option_name' );
+	$autoload    = $request->get_param( 'autoload' );
+	$value       = $request->get_param( 'value' );
+
+	if ( empty( $option_name ) || empty( $autoload ) ) {
+		return new WP_REST_Response(
+			array(
+				'success' => false,
```
westonruter:
```suggestion
```

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -20,9 +20,119 @@
...
+		if ( 'yes' === $autoload && isset( $modified_options[ $option_name ] ) ) {
+			unset( $modified_options[ $option_name ] );
+		} else {
+			$modified_options[ $option_name ] = $value;
+		}
+		update_option( 'perflab_aao_modified_options', $modified_options );
+
+		return new WP_REST_Response(
+			array(
+				'success' => true,
```
westonruter:
```suggestion
```

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -20,9 +20,119 @@
...
+					$autoload,
+					$option_name
+				),
+			),
+			200
+		);
+	} else {
+		return new WP_REST_Response(
+			array(
+				'success' => false,
```
westonruter:
```suggestion
```

modules/database/audit-autoloaded-options/update-autoload.js
```diff
@@ -0,0 +1,45 @@
...
+			},
+			error( error ) {
+				$(
+					'<div style="margin-top:5px;border:1px solid #d63638;padding-left:5px;">' +
```
westonruter: This should re-use the `error` class instead of custom styles.

modules/database/audit-autoloaded-options/update-autoload.js
```diff
@@ -0,0 +1,45 @@
...
+						error.responseJSON.message +
+						'</div>'
+				).insertAfter( button );
```
westonruter: The message needs to be escaped by setting it via `text()`.

```suggestion
				const errorMessage = $( '<div class="error"></div>' );
				errorMessage.text( error.responseJSON.message );
				button.after( errorMessage );
```

Also, what happens when it keeps failing? It adds more errors without removing previous ones:

image

Perhaps the error container element should be in the DOM output by PHP and then populated and shown when an error occurs. When the error resolves, the element can be hidden again.

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -20,9 +20,119 @@
...
+	if ( 'site-health.php' !== $hook_suffix ) {
+		return;
+	}
+	wp_enqueue_script( 'perflab-aao-update-autoload', plugin_dir_url( __FILE__ ) . 'update-autoload.js', array( 'jquery' ), 'n.e.x.t', true );
+
+	wp_localize_script(
+		'perflab-aao-update-autoload',
+		'perflabAutoloadSettings',
+		array(
+			'root' => sanitize_url( get_rest_url() ),
```
westonruter: I don't think this is needed because it's coming straight from WP:

```suggestion
			'root' => get_rest_url(),
```

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -20,9 +20,119 @@
  */
 function perflab_aao_add_autoloaded_options_test( $tests ) {
 	$tests['direct']['autoloaded_options'] = array(
-		'label' => __( 'Autoloaded options', 'performance-lab' ),
+		'label' => esc_html__( 'Autoloaded options', 'performance-lab' ),
```
westonruter: I don't see where this label is ever shown. It may not need `esc_html()`.

modules/database/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,45 @@ function perflab_aao_get_autoloaded_options_table() {
...
+	$html_table = sprintf(
+		'<br><br><table class="widefat striped"><thead><tr><th scope="col">%s</th><th scope="col">%s</th><th scope="col">%s</th></tr></thead><tbody>',
+		esc_html__( 'Option Name', 'performance-lab' ),
+		esc_html__( 'Size', 'performance-lab' ),
+		esc_html__( 'Action', 'performance-lab' )
+	);
+
+	foreach ( $perflab_aao_disabled_options as $option_name => $value ) {
+		$disable_button = sprintf( '<a class="button update-autoload" data-option-name="%s" data-autoload="yes" href="#">%s</a>', esc_attr( $option_name ), esc_html__( 'Revert to Autoload', 'performance-lab' ) );
+		$html_table    .= sprintf( '<tr><td>%s</td><td>%s</td><td>%s</td></tr>', $option_name, size_format( $value, 2 ), $disable_button );
```
westonruter: Very important escaping here:

```suggestion
		$html_table    .= sprintf( '<tr><td>%s</td><td>%s</td><td>%s</td></tr>', esc_html( $option_name ), size_format( $value, 2 ), $disable_button );
```

modules/database/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,45 @@ function perflab_aao_get_autoloaded_options_table() {
...
+		'<table class="widefat striped"><thead><tr><th scope="col">%s</th><th scope="col">%s</th><th scope="col">%s</th></tr></thead><tbody>',
+		esc_html__( 'Option Name', 'performance-lab' ),
+		esc_html__( 'Size', 'performance-lab' ),
+		esc_html__( 'Action', 'performance-lab' )
 	);
 
 	foreach ( $autoload_summary as $value ) {
-		$html_table .= sprintf( '<tr><td>%s</td><td>%s</td></tr>', $value->option_name, size_format( $value->option_value_length, 2 ) );
+		$disable_button = sprintf( '<a class="button update-autoload" data-option-name="%s" data-autoload="no" data-value="%s" href="#">%s</a>', esc_attr( $value->option_name ), esc_attr( $value->option_value_length ), esc_html__( 'Disable Autoload', 'performance-lab' ) );
+		$html_table    .= sprintf( '<tr><td>%s</td><td>%s</td><td>%s</td></tr>', $value->option_name, size_format( $value->option_value_length, 2 ), $disable_button );
```
westonruter: Critical escaping here:

```suggestion
		$html_table    .= sprintf( '<tr><td>%s</td><td>%s</td><td>%s</td></tr>', esc_html( $value->option_name ), size_format( $value->option_value_length, 2 ), $disable_button );
```

westonruter (changes requested):

modules/database/audit-autoloaded-options/hooks.php
```diff
@@ -20,9 +20,119 @@
...
 }
 add_filter( 'site_status_tests', 'perflab_aao_add_autoloaded_options_test' );
+
+/**
+ * Register custom REST API route for updating autoload.
+ *
+ * @since n.e.x.t
+ */
+function perflab_aao_register_rest_route() {
+	register_rest_route(
```
swissspidy: This is lacking a `permission_callback`

modules/database/audit-autoloaded-options/helper.php
felixarntz: If we continue using AJAX for this logic, then using a `button` would indeed be more appropriate.

Alternatively, if we want to keep the implementation simpler, we could use an "action URL" request to handle the change. If so, continuing to use a link (`a`) would make sense.

modules/database/audit-autoloaded-options/hooks.php
felixarntz: Yes, this doesn't need to be escaped, it's handled by core. Also somewhat unrelated to this PR anyway.

modules/database/audit-autoloaded-options/update-autoload.js
felixarntz: As far as I can tell, the success callback doesn't handle moving the option to the other table, which creates a weird user experience which would require the user to reload the page to see what it should _actually_ look like after performing the action.

I see two options here:
* Either we fix this by expanding the JS logic. Doing it with AJAX can create a better user experience - but only if done right. The end result should be the same as if you refreshed the page afterwards.
* Or we ditch the AJAX logic and REST endpoint and go with a much faster and easier to implement "action URL" that handles the logic (e.g. via `admin_action_{$action}` hook). This would be _a lot_ easier to implement in a way that doesn't feel weird.

Again, in theory AJAX can make for a smoother experience than a page reload - but not if the AJAX experience is actually sub par in what it does. Worth noting that handling this via AJAX and JS requires some JS code that duplicates functionality which PHP already handles, so there's always a maintenance overhead. Not sure whether it's worth it.

felixarntz (changes requested):
@mukeshpanchal27 I don't have much to add to the existing feedback. Most importantly from my perspective though, we either need to do the AJAX + JS approach the right way, or alternatively go with the much simpler alternative of using a page reload and a regular WP Admin callback via "action URL".

mukeshpanchal27:
Thanks, @felixarntz, @westonruter, and @swissspidy for the feedback. Per the discussion in https://github.com/WordPress/performance/pull/1048#discussion_r1523685027, I also believe that the `action URL` approach is the right choice. Additionally, I've added an admin notice when the option is updated.

modules/database/audit-autoloaded-options/update-autoload.js
mukeshpanchal27: In https://github.com/WordPress/performance/pull/1048/commits/781871b617d14ff061c572ceef3a9d48bec3ebab i have update approach.

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+		wp_die( esc_html__( 'Invalid option name or autoload value.', 'performance-lab' ) );
+	}
+
+	// Check if the option exists.
+	if ( false === get_option( $option_name ) ) {
+		// Option doesn't exist, return an error or handle the situation accordingly.
+		wp_die( esc_html__( 'The option does not exist.', 'performance-lab' ) );
+	}
+
+	$result = wp_set_option_autoload( $option_name, $autoload );
```
swissspidy, line 69: Looks like Performance Lab requires WP 6.3+ but this function has only been added in 6.4.

We either should bump our requirement or add some safeguards.

includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,70 @@ function perflab_aao_get_autoloaded_options_table() {
...
+	);
+
+	$nonce = wp_create_nonce( 'perflab_aao_update_autoload' );
+	foreach ( $perflab_aao_disabled_options as $option_name => $value ) {
+		$url            = esc_url_raw(
+			add_query_arg(
+				array(
+					'action'      => 'perflab_aao_update_autoload',
+					'_wpnonce'    => $nonce,
+					'option_name' => esc_attr( $option_name ),
```
westonruter: No escaping should be done here. It will be handled when output below.

```suggestion
					'option_name' => $option_name,
```

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+		$modified_options = get_option( 'perflab_aao_modified_options', array() );
+		if ( 'yes' === $autoload && isset( $modified_options[ $option_name ] ) ) {
+			unset( $modified_options[ $option_name ] );
+		} else {
+			$modified_options[ $option_name ] = $value;
+		}
+		update_option( 'perflab_aao_modified_options', $modified_options );
+
+		wp_safe_redirect( admin_url( 'site-health.php?autoload_updated=true' ) );
+		exit;
```
westonruter: To facilitate unit testing:

```suggestion
		if ( wp_safe_redirect( admin_url( 'site-health.php?autoload_updated=true' ) ) ) {
			exit;
		}
```

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+ * @global string $pagenow The filename of the current screen.
+ */
+function perflab_aao_admin_notices() {
+	if ( 'site-health.php' !== $GLOBALS['pagenow'] ) {
+		return;
+	}
+
+	if ( isset( $_GET['autoload_updated'] ) && 'true' === $_GET['autoload_updated'] ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		wp_admin_notice(
+			esc_html__( 'The option has been successfully updated!', 'performance-lab' ),
```
westonruter:
```suggestion
			esc_html__( 'The option has been successfully updated.', 'performance-lab' ),
```

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+ *
+ * @since n.e.x.t
+ */
+function perflab_aao_handle_update_autoload() {
+	// Verify nonce.
+	$nonce = isset( $_GET['_wpnonce'] ) ? sanitize_text_field( $_GET['_wpnonce'] ) : '';
+
+	if ( ! wp_verify_nonce( $nonce, 'perflab_aao_update_autoload' ) ) {
+		wp_die( esc_html__( 'Nonce verification failed.', 'performance-lab' ) );
+	}
```
westonruter: I believe this can be simplified as follows:

```suggestion
	check_admin_referer( 'perflab_aao_update_autoload' );
```

includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,70 @@ function perflab_aao_get_autoloaded_options_table() {
...
 	foreach ( $autoload_summary as $value ) {
-		$html_table .= sprintf( '<tr><td>%s</td><td>%s</td></tr>', $value->option_name, size_format( $value->option_value_length, 2 ) );
+		$url            = esc_url_raw(
+			add_query_arg(
+				array(
+					'action'      => 'perflab_aao_update_autoload',
+					'_wpnonce'    => $nonce,
+					'option_name' => esc_attr( $value->option_name ),
+					'autoload'    => 'no',
+					'value'       => esc_attr( $value->option_value_length ),
```
westonruter: Could be `value_length` or `value_size` be used instead? It's not clear otherwise what the `value` is.

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+
+	$option_name = isset( $_GET['option_name'] ) ? sanitize_text_field( $_GET['option_name'] ) : '';
+	$autoload    = isset( $_GET['autoload'] ) ? sanitize_text_field( $_GET['autoload'] ) : '';
+	$value       = isset( $_GET['value'] ) ? sanitize_text_field( $_GET['value'] ) : '';
```
westonruter:
```suggestion
	$value       = isset( $_GET['value'] ) ? (int) $_GET['value'] : 0;
```

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+function perflab_aao_handle_update_autoload() {
+	// Verify nonce.
+	$nonce = isset( $_GET['_wpnonce'] ) ? sanitize_text_field( $_GET['_wpnonce'] ) : '';
+
+	if ( ! wp_verify_nonce( $nonce, 'perflab_aao_update_autoload' ) ) {
+		wp_die( esc_html__( 'Nonce verification failed.', 'performance-lab' ) );
+	}
+
+	$option_name = isset( $_GET['option_name'] ) ? sanitize_text_field( $_GET['option_name'] ) : '';
+	$autoload    = isset( $_GET['autoload'] ) ? sanitize_text_field( $_GET['autoload'] ) : '';
```
westonruter:
```suggestion
	$autoload    = isset( $_GET['autoload'] ) ? rest_sanitize_boolean( $_GET['autoload'] ) : false;
```

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+		// Option doesn't exist, return an error or handle the situation accordingly.
+		wp_die( esc_html__( 'The option does not exist.', 'performance-lab' ) );
+	}
+
+	$result = wp_set_option_autoload( $option_name, $autoload );
+
+	if ( $result ) {
+		// Update modified options list.
+		$modified_options = get_option( 'perflab_aao_modified_options', array() );
+		if ( 'yes' === $autoload && isset( $modified_options[ $option_name ] ) ) {
```
westonruter:
```suggestion
		if ( $autoload && isset( $modified_options[ $option_name ] ) ) {
```

includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,70 @@ function perflab_aao_get_autoloaded_options_table() {
...
+
+	$nonce = wp_create_nonce( 'perflab_aao_update_autoload' );
+	foreach ( $perflab_aao_disabled_options as $option_name => $value ) {
+		$url            = esc_url_raw(
+			add_query_arg(
+				array(
+					'action'      => 'perflab_aao_update_autoload',
+					'_wpnonce'    => $nonce,
+					'option_name' => esc_attr( $option_name ),
+					'autoload'    => 'yes',
```
westonruter:
```suggestion
					'autoload'    => 'true',
```

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+ * @since n.e.x.t
+ */
+function perflab_aao_handle_update_autoload() {
+	// Verify nonce.
+	$nonce = isset( $_GET['_wpnonce'] ) ? sanitize_text_field( $_GET['_wpnonce'] ) : '';
+
+	if ( ! wp_verify_nonce( $nonce, 'perflab_aao_update_autoload' ) ) {
+		wp_die( esc_html__( 'Nonce verification failed.', 'performance-lab' ) );
+	}
+
```
westonruter, line 51: Shouldn't this bail if any of the required parameters are missing?

```suggestion

if ( ! isset( $_GET['option_name'], $_GET['autoload'], $_GET['value'] ) ) {
	wp_die( esc_html__( 'Missing required parameter.', 'performance-lab' ) );
}
```

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+	$option_name = isset( $_GET['option_name'] ) ? sanitize_text_field( $_GET['option_name'] ) : '';
```
westonruter: If the above is done, then this can be, but don't forget to unslash all input vars:

```suggestion
	$option_name = sanitize_text_field( wp_unslash( $_GET['option_name'] ) );
```

includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,70 @@ function perflab_aao_get_autoloaded_options_table() {
...
+	$nonce = wp_create_nonce( 'perflab_aao_update_autoload' );
 	foreach ( $autoload_summary as $value ) {
-		$html_table .= sprintf( '<tr><td>%s</td><td>%s</td></tr>', $value->option_name, size_format( $value->option_value_length, 2 ) );
+		$url            = esc_url_raw(
+			add_query_arg(
+				array(
+					'action'      => 'perflab_aao_update_autoload',
+					'_wpnonce'    => $nonce,
+					'option_name' => esc_attr( $value->option_name ),
+					'autoload'    => 'no',
```
westonruter:
```suggestion
					'autoload'    => 'false',
```

includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,70 @@ function perflab_aao_get_autoloaded_options_table() {
...
 
+	$nonce = wp_create_nonce( 'perflab_aao_update_autoload' );
 	foreach ( $autoload_summary as $value ) {
-		$html_table .= sprintf( '<tr><td>%s</td><td>%s</td></tr>', $value->option_name, size_format( $value->option_value_length, 2 ) );
+		$url            = esc_url_raw(
+			add_query_arg(
+				array(
+					'action'      => 'perflab_aao_update_autoload',
+					'_wpnonce'    => $nonce,
+					'option_name' => esc_attr( $value->option_name ),
```
westonruter:
```suggestion
					'option_name' => $value->option_name,
```

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+
+	$result = wp_set_option_autoload( $option_name, $autoload );
+
+	if ( $result ) {
+		// Update modified options list.
+		$modified_options = get_option( 'perflab_aao_modified_options', array() );
+		if ( 'yes' === $autoload && isset( $modified_options[ $option_name ] ) ) {
+			unset( $modified_options[ $option_name ] );
+		} else {
+			$modified_options[ $option_name ] = $value;
```
westonruter: The value of `$value` is not getting validated. See above for casting to an `int`.

westonruter (changes requested):

includes/site-health/audit-autoloaded-options/hooks.php
mukeshpanchal27: Per the policy outlined in https://make.wordpress.org/performance/handbook/performance-lab/version-support-policy/#wordpress-core-versions, Performance Lab is committed to supporting the latest stable version of WordPress core. Additionally, considering that this change is expected to be released in version 3.0, which is slated for mid-April, and WP 6.5 is scheduled for release on March 26, we propose bumping the minimum version for Performance Lab to 6.4.

What do you think? cc. @felixarntz

mukeshpanchal27: Open followup issue #1056

mukeshpanchal27:
Thanks, @westonruter, for the suggestions and feedback.

I have updated it slightly so that it also works for reverting autoload values.

includes/site-health/audit-autoloaded-options/hooks.php
westonruter: Do it. 🚀

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+	$result = wp_set_option_autoload( $option_name, $autoload );
+
+	if ( $result ) {
+		// Update modified options list.
+		$modified_options = get_option( 'perflab_aao_modified_options', array() );
+		if ( $autoload && isset( $modified_options[ $option_name ] ) ) {
+			unset( $modified_options[ $option_name ] );
+		} else {
+			$modified_options[ $option_name ] = $value_length;
+		}
```
westonruter: Is this always right? If `isset( $modified_options[ $option_name ] )` is true then this will cause `$modified_options[ $option_name ]` to be set, but that would seem wrong because it is supposed to only be populated with the options for which autoload is disabled? But it will always get set then regardless of whether `$autoload` is true or false.

mukeshpanchal27: Great catch! I didn't notice that oversight. In the commit https://github.com/WordPress/performance/pull/1048/commits/93c6a42ae3e38e1ae9f5d12957b608d0c19c53ee, I have updated the logic to address this issue. Now, the `$modified_options[$option_name]` is only set when `$autoload` is false, ensuring that it tracks modified options correctly, aligning with the intended behavior of the function. Thank you for bringing this to my attention!

includes/site-health/audit-autoloaded-options/hooks.php
felixarntz: Replied on the issue, SGTM!

includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,70 @@ function perflab_aao_get_autoloaded_options_table() {
...
 	foreach ( $autoload_summary as $value ) {
-		$html_table .= sprintf( '<tr><td>%s</td><td>%s</td></tr>', $value->option_name, size_format( $value->option_value_length, 2 ) );
+		$url            = esc_url_raw(
+			add_query_arg(
+				array(
+					'action'       => 'perflab_aao_update_autoload',
+					'_wpnonce'     => $nonce,
+					'option_name'  => $value->option_name,
+					'autoload'     => 'false',
+					'value_length' => esc_attr( $value->option_value_length ),
```
felixarntz: Maybe call this `option_size` instead? Similar to how the table is labelled.

It's a clever solution to store the size so that we can still show it. However, this value may get out of date, so I wonder whether we should instead implement this more dynamically. Last but not least it feels a bit strange that the option size has to be passed to change the autoload value.

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+function perflab_aao_handle_update_autoload() {
+	check_admin_referer( 'perflab_aao_update_autoload' );
+
+	if ( ! isset( $_GET['option_name'], $_GET['autoload'] ) ) {
+		wp_die( esc_html__( 'Missing required parameter.', 'performance-lab' ) );
+	}
+
+	$option_name  = sanitize_text_field( wp_unslash( $_GET['option_name'] ) );
+	$autoload     = isset( $_GET['autoload'] ) ? rest_sanitize_boolean( $_GET['autoload'] ) : false;
+	$value_length = isset( $_GET['value_length'] ) ? (int) $_GET['value_length'] : 0;
```
felixarntz: Per my above feedback, it's a bit strange we need to pass this here. But more importantly it can get out of date, e.g. when an option is updated. Also, since it's not mandatory, it may be `0` which would be confusing.

I think it would be better to not pass this value to the action URL and instead dynamically retrieve it from the database when displaying the disabled options table.

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+
+	if ( $result ) {
+		// Update modified options list.
+		$modified_options = get_option( 'perflab_aao_modified_options', array() );
+		if ( ! $autoload ) {
+			$modified_options[ $option_name ] = $value_length;
+		} elseif ( isset( $modified_options[ $option_name ] ) ) {
+			unset( $modified_options[ $option_name ] );
+		}
+		update_option( 'perflab_aao_modified_options', $modified_options );
```
felixarntz: Maybe better to call this `disabled_options` rather than `modified_options`? This better indicates that autoload for those options was disabled, rather than just "modified", which could be in any way.

includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,70 @@ function perflab_aao_get_autoloaded_options_table() {
...
+ *
+ * @return string HTML formatted table.
+ */
+function perflab_aao_get_disabled_autoloaded_options_table() {
+	$perflab_aao_disabled_options = get_option( 'perflab_aao_modified_options', array() );
+
+	if ( empty( $perflab_aao_disabled_options ) ) {
+		return '';
+	}
+
```
felixarntz, line 191: Instead of storing the option size from the time of disabling autoload for the option, here you could run a database query that gets those sizes.

It could look rather similar to the other query, however you would need to use something like `WHERE option_name IN (...)` with the concrete list of the options for which autoload was disabled. This way the table will always display up to date values.

It also brings a real user benefit: Maybe an option is very large, but later the plugin developer modifies what is stored there, so the option becomes smaller after an update. This table will then show it and the admin may consider reverting it back to autoload.

felixarntz (commented):
@mukeshpanchal27 This looks very close now. I only have one broader suggestion on how to handle the option size to display to the user more dynamically.

includes/site-health/audit-autoloaded-options/helper.php
mukeshpanchal27: Thank you for bringing this to my attention. Initially, I considered using a query, but to avoid it, I stored the option value. However, I overlooked updating the option even if it is disabled. In https://github.com/WordPress/performance/pull/1048/commits/4e857a448d883663e79f3b48e53376d97fb1269f, I have made the necessary updates as per your direction.

mukeshpanchal27:
Thank you, @felixarntz and @westonruter, for your feedback. The PR is now ready for the final review and merge.

includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -26,3 +26,86 @@ function perflab_aao_add_autoloaded_options_test( $tests ) {
...
+
+	if ( $result ) {
+		// Update disabled options list.
+		$disabled_options = get_option( 'perflab_aao_disabled_options', array() );
+		if ( ! $autoload ) {
+			$disabled_options[] = $option_name;
+		} elseif ( in_array( $option_name, $disabled_options, true ) ) {
+			$key = array_search( $option_name, $disabled_options, true );
+			unset( $disabled_options[ $key ] );
+		}
```
felixarntz: Two small improvements can be made here:
* It's worth ensuring that the same option can't be added to the `$disabled_options` list twice.
* You don't need to use both `in_array()` and `array_search()`. If you need the key, you can use `array_search()` alone.

For example:
```suggestion

		$key = array_search( $option_name, $disabled_options, true );
		if ( ! $autoload && false === $key ) {
			$disabled_options[] = $option_name;
		} elseif ( $autoload && false !== $key ) {
			unset( $disabled_options[ $key ] );
		}

```

includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -144,13 +144,83 @@ function perflab_aao_get_autoloaded_options_table() {
...
+			sprintf(
+				"SELECT option_name, LENGTH(option_value) AS option_value_length FROM $wpdb->options WHERE option_name IN (%s) ORDER BY option_value_length DESC",
+				implode( ',', array_fill( 0, count( $disabled_options ), '%s' ) )
+			),
+			$disabled_options
+		)
+	);
+
+	$html_table = sprintf(
+		'<br><br><table class="widefat striped"><thead><tr><th scope="col">%s</th><th scope="col">%s</th><th scope="col">%s</th></tr></thead><tbody>',
```
felixarntz: Rather than just showing the table below the other table without any other context, we should provide a minimal explanation or heading for it. On another note, I'd suggest to avoid using `<br>` to create spacing, as this element should be used for line breaks within text.

I think we could replace it as follows:
```suggestion
		'<p>%s</p><table class="widefat striped"><thead><tr><th scope="col">%s</th><th scope="col">%s</th><th scope="col">%s</th></tr></thead><tbody>',
		__( 'The following table shows the options for which you have previously disabled Autoload.', 'performance-lab' ),
```

felixarntz (approved):
@mukeshpanchal27 Two small last things, otherwise this is good to go IMO. Great stuff!

westonruter:
Will there be tests added for this?

swissspidy:
> Will there be tests added for this?

This would be a good candidate for an e2e test. Makes me wonder why we don't have any 🤔

mukeshpanchal27:
> > Will there be tests added for this?
> 
> This would be a good candidate for an e2e test. Makes me wonder why we don't have any 🤔

We don't currently have e2e tests set up in the PL repository. I'm exploring how we can add tests for these changes. 😅

swissspidy:
I saw that, hence the wondering why :)

here, some php unit tests should be possible

mukeshpanchal27:
@swissspidy @westonruter I've added unit tests. It appears that the unit tests are failing in WP 6.3 due to the introduction of `wp_set_option_autoload` in 6.4. We'll need to update the workflow accordingly.

swissspidy:
I thought #1062 updated the workflows? Why are we still testing 6.3? Let's just fix that here.

.github/workflows/php-test.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         wp: [ 'latest' ]
         include:
           - php: '7.4'
-            wp: '6.3'
+            wp: '6.4'
```
felixarntz, line 49: @mukeshpanchal27 @swissspidy Strange that this is not failing in `trunk` 🤔 

But anyway, great to fix here.

felixarntz (approved):

felixarntz:
@westonruter Which kinds of tests are you envisioning? In terms of PHPUnit, maybe a test of the "action function" to update the autoload values would be useful. I'm not sure it's worth adding tests for the Site Health UI bits as that's relatively trivial and IMO more difficult to test than it provides value.

westonruter (approved):

PR #1049, opened 2024-03-12
Fix `test-plugins` command

mukeshpanchal27:
## Summary

This PR resolves the issue with the `test-plugins` command. After transitioning the infrastructure from modules to plugins, it is now functioning correctly.

## Checklist

### DIFF
--- a/.github/workflows/php-test-standalone-plugins.yml
+++ b/.github/workflows/php-test-standalone-plugins.yml
@@ -8,6 +8,7 @@ on:
     # Only run if PHP-related files changed.
     paths:
       - '.github/workflows/php-test-standalone-plugins.yml'
+      - 'bin/plugin/commands/test-plugins.js'
       - 'plugin-tests/**'
       - '**.php'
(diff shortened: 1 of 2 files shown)


mukeshpanchal27:
Do we need to add the command JS file in the paths here: https://github.com/WordPress/performance/blob/trunk/.github/workflows/php-test-standalone-plugins.yml#L26-L34 so that when it gets updated, the workflow is triggered?

westonruter:
@mukeshpanchal27 yes, good idea to add `bin/plugin/commands/test-plugins.js` to the list of paths.

felixarntz (approved):
@mukeshpanchal27 Good catch! Once you have added https://github.com/WordPress/performance/pull/1049#issuecomment-1991365827, this is good to go from my perspective.

mukeshpanchal27:
Thanks @westonruter @felixarntz. PR is ready for second sign-off.

westonruter (approved):

PR #1050, opened 2024-03-12
Remove checklist from pull request template

westonruter:
## Summary

At WordCamp Asia Contributor Day, a contributor experienced friction when attempting to contribute because they were unable to action on the checklist in the pull request template. External contributors cannot set any labels, so they may feel blocked from contributing. I think the checklist should be removed for this reason, and also because the checklist is redundant with the [pr-validation](https://github.com/WordPress/performance/blob/trunk/.github/workflows/pr-validation.yml) workflow which will serve the same purpose as the checklist, but in terms of failing checks. Contributors who are part of the project know the labels they need to add (although we've also discussed adding these labels may not always be of much value anymore anyway). Also, the checklist adds additional overhead for contributors who regularly open PRs: they have to not only add the labels but also check the checkboxes. Removing the checklist will improve efficiency and reduce friction for new contributors.

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,3 +6,13 @@ Fixes #
 ## Relevant technical choices
 
 <!-- Please describe your changes. -->
+
+
+
+<!--
+For maintainers only, please make sure:
+
+- PR has either `[Focus]` or `Infrastructure` label.
+- PR has a `[Type]` label.
+- PR has a milestone or the `no milestone` label.
+-->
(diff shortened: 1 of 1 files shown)


felixarntz (commented):
@westonruter How about providing this purely as information rather than a list of checkboxes?

We could put something like: _For maintainers: Make sure the PR has ..._

Regarding the purpose of those, let's discuss separately. The reason those requirements are present is automatic changelog generation. However, there are definitely parts of the requirements which are now outdated, such as the `[Focus]` labels (since PL no longer contains different performance features that fall under those focuses, using this as "categorization" for the changelog is now pointless as everything effectively falls under "Infrastructure"). Instead, we may need a requirement to provide the label for which plugin the PR is for.

westonruter:
> How about providing this purely as information rather than a list of checkboxes?
> 
> We could put something like: _For maintainers: Make sure the PR has ..._

OK, done in 5e42dd3

felixarntz (approved):

swissspidy (approved):

Issue #1046, opened 2024-03-11
Review current plugin names and their descriptions for their accuracy and helpfulness


felixarntz:
See https://github.com/WordPress/performance/issues/1031#issuecomment-1989043559: We should revisit our current standalone plugins' names and descriptions for how accurate and useful they are. In particular, we should consider the following:
* Plugin names and descriptions are used to describe features within Performance Lab.
* Our plugins are not _products_. They are features for WordPress core. As such, we should avoid plugin names that sound too much like a product / brand name, and instead simply find a good name that clarifies the feature. Example: "Optimization Detective" doesn't clarify what it does.
* Additionally, we should avoid technicalities in the name, to make them intuitive. Example: "WebP Uploads" would probably be better as something like "WebP Images" (the term "Uploads" is only used because of Media Library jargon). "Speculation Rules" should probably be something like "Speculative preloading" because the Speculation Rules API is a technical detail.
* Plugin descriptions should not mention the word "plugin". They can just start with a third-person verb (e.g. "Adds xyz" or similar).

For reference, here are our current plugin names and descriptions to review:
* Auto-sizes for Lazy-loaded Images
    * This plugin implements the HTML spec for adding `sizes="auto"` to lazy-loaded images.
* Dominant Color Images
    * Adds support to store the dominant color of newly uploaded images and create a placeholder background of that color.
* Optimization Detective
    * Improves accuracy of optimizing the loading of the LCP image by leveraging client-side detection with real user metrics. Also enables output buffering of template rendering which can be filtered.
* Performant Translations
    * Making internationalization/localization in WordPress faster than ever before.
* Speculation Rules
    * Uses the Speculation Rules API to prerender linked URLs upon hover by default.
* WebP Uploads
    * Creates WebP versions for new JPEG image uploads if supported by the server.

Let's brainstorm.

cc @adamsilverstein @joemcgill @mukeshpanchal27 @swissspidy @westonruter

westonruter:
> Example: "WebP Uploads" would probably be better as something like "WebP Images" (the term "Uploads" is only used because of Media Library jargon). 

Further, we discussed ideally not including WebP here either, so it would be rather "Modern Images".

> * Our plugins are not _products_. They are features for WordPress core. As such, we should avoid plugin names that sound too much like a product / brand name, and instead simply find a good name that clarifies the feature. Example: "Optimization Detective" doesn't clarify what it does.

[...]

PR #1052, opened 2024-03-13
Prefix standalone plugins with 'PL' in list table and add plugin row meta link to Performance Lab settings screen

westonruter:
## Summary

See #1046. When the PL plugin is active, this prefixes all installed standalone plugins with PL so that they appear grouped together. A "Performance Lab" plugin row meta link is also added which points to the settings screen.

Performance Lab Deactivated | Performance Lab Activated
--|--
localhost_8888_wp-admin_plugins php_plugin_status=all paged=1 s | localhost_8888_wp-admin_plugins php

## Relevant technical choices

The `all_plugins` filter is employed to loop overall installed plugins, and if any one matches one of the standalone plugin slugs, it prefixes the Name with "PL". After this, the plugins are re-sorted so that the PL plugins appear together.

Also, the plugin row meta is filtered for the standalone plugins so that a "Performance Lab" link is added for the plugin's settings screen.

### DIFF
--- a/admin/load.php
+++ b/admin/load.php
@@ -912,3 +912,38 @@ function perflab_print_modules_page_style() {
 </style>
 	<?php
 }
+
+/**
+ * Filters all plugins to prefix Performance Lab standalone plugins with "PL".
+ *
+ * @since n.e.x.t
+ *
+ * @param array<string, array{ Name: string }> $plugins Plugins.
+ * @return array<string, array{ Name: string }> Filtered plugins.
+ */
+function perflab_filter_plugins_list( array $plugins ): array {
+	$standalone_plugins = perflab_get_standalone_plugins();
+
+	$did_prefix  = false;
+	$name_prefix = __( 'PL', 'performance-lab' ) . ' ';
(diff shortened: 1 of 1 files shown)


felixarntz (commented):
I'm not a fan of the "PL" prefix. This is as if "Jetpack Boost" was called "JP Boost". When you see one of those plugins, you probably wouldn't know what it means.

Anyway, let's wait for the decision in #1046 before considering a change like this. Potentially we'll want to change the plugin names themselves instead, in which case this wouldn't be relevant.

westonruter:
"Jetpack" is much shorter than "Performance Lab" so it makes more sense to not abbreviate.

Anyway, this PR is a POC to assist with the discussion of what the possibilities are.

ThierryA:
> "Jetpack" is much shorter than "Performance Lab" so it makes more sense to not abbreviate.

Why would the length be an issue? I would understand how it is less pleasant in the code base, for a plugin name it doesn't really matter. Folks reading the plugin name would not understand what "PL" stands for, kind of against the "don't make me think approach".

We should rename the project to Perf Lab 😂

PR #1055, opened 2024-03-14
Implement garbage collection of URL metrics including at uninstallation

### LINKED ISSUE #895: Implement garbage collection for URL metrics posts
## Feature Description

In the Image Loading Optimization module (see #869), detection logic may run when a visitor lands on a page to collect metrics for that URL, namely which element is LCP and which images were in the initial viewport. This information is then sent to a REST API endpoint and is then stored in an `ilo_url_metrics` post. Over time, URL metrics will be updated in this post as more visitors land on the page. However, a post gets deleted and the URL for which an `ilo_url_metrics` post was created, then at present the `ilo_url_metrics` will persist indefinitely. There should be a cron job that does garbage collection of these posts `ilo_url_metrics` when they are excessively stale. For example, if the `post_modified_gmt` for a given post is over a month old, then the post could be deleted.

🆕 This should also involve implementing an `uninstall.php` which deletes all instances of the custom post type at uninstallation.

westonruter:
## Summary

This PR implements two things:

* Garbage collection of URL Metrics posts which haven't been modified in a month.
* Introduction of `uninstall.php` which deletes all of the URL Metrics and unschedules the garbage collection event.

Fixes #895

## Relevant technical choices

Note that the first commit in this PR (https://github.com/WordPress/performance/pull/1055/commits/21a5cec031fd5862a85dac9974aeae4f045d0abc) refactors the post type logic from a set of functions to a single class `OD_URL_Metrics_Post_Type`. To facilitate review, it may be helpful to look at that commit separately from the ones that follow: 21a5cec031fd5862a85dac9974aeae4f045d0abc...1a9532c.

This implements garbage collection of URL Metrics posts which haven't been modified in a month. By default, a URL Metric has a freshness TTL of 1 day. This means that after 1 day, the detection logic will attempt to gather a new URL Metric to refresh the one that is stale.  There is also a sample size of 3 for each URL Metrics group (divided by the breakpoints), so if there are 3 breakpoints there would be 4 groups each of which seeks to have a total of 3 fresh URL Metrics. So there could be a maximum of 12 updates to a given URL Metrics post on a given day. If there are no visitors to a given URL for an entire month, and thus the URL Metrics post for that URL hasn't been modified in a month, then it is eligible to be deleted. The garbage collection process is configured to run on a daily basis.

An `uninstall.php` is also added which deletes all URL Metrics posts from the current site and other sites on a multisite network (up to 100). The post deletion here is done in batch using `$wpdb->delete()` for the sake of speed. We used [the same approach in the AMP plugin](https://github.com/ampproject/amp-wp/blob/5708e19c60cf0f88ce522349fba955266b6892f6/includes/uninstall-functions.php#L51-L86).


plugins/optimization-detective/class-od-url-metrics-post-type.php
```diff
@@ -0,0 +1,329 @@
...
+		$scheduled_event = wp_get_scheduled_event( self::GC_CRON_EVENT_NAME );
+		if ( $scheduled_event && self::GC_CRON_RECURRENCE !== $scheduled_event->schedule ) {
+			wp_unschedule_event( $scheduled_event->timestamp, self::GC_CRON_EVENT_NAME );
+			$scheduled_event = null;
+		}
+
+		if ( ! $scheduled_event ) {
+			wp_schedule_event( time(), self::GC_CRON_RECURRENCE, self::GC_CRON_EVENT_NAME );
+		}
+	}
```
westonruter: h/t https://github.com/ampproject/amp-wp/blob/5708e19c60cf0f88ce522349fba955266b6892f6/src/BackgroundTask/RecurringBackgroundTask.php#L34-L53

plugins/optimization-detective/class-od-url-metrics-post-type.php
```diff
@@ -0,0 +1,329 @@
...
+				'query_var'        => false,
+				'delete_with_user' => false,
+				'can_export'       => false,
+				'supports'         => array( 'title' ),
+				// The original URL is stored in the post_title, and the post_name is a hash of the query vars.
+			)
+		);
+
+		add_action( 'admin_init', array( __CLASS__, 'schedule_garbage_collection' ) );
+		add_action( self::GC_CRON_EVENT_NAME, array( __CLASS__, 'delete_stale_posts' ) );
```
westonruter: h/t https://github.com/ampproject/amp-wp/blob/5708e19c60cf0f88ce522349fba955266b6892f6/src/BackgroundTask/RecurringBackgroundTask.php#L25-L26

plugins/optimization-detective/class-od-url-metrics-post-type.php
```diff
@@ -0,0 +1,329 @@
...
+		// Delete all URL Metrics posts.
+		$wpdb->delete(
+			$wpdb->posts,
+			array(
+				'post_type' => self::SLUG,
+			)
+		);
+
+		// phpcs:enable WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
+	}
```
westonruter: h/t https://github.com/ampproject/amp-wp/blob/5708e19c60cf0f88ce522349fba955266b6892f6/includes/uninstall-functions.php#L57-L86

plugins/optimization-detective/class-od-url-metrics-post-type.php
```diff
@@ -0,0 +1,329 @@
+<?php
```
felixarntz: IMO it would make more sense to have this file in the `storage` folder, like `post-type.php` before.

plugins/optimization-detective/class-od-url-metrics-post-type.php
```diff
@@ -0,0 +1,329 @@
...
+			)
+		);
+
+		// Delete all URL Metrics posts.
+		$wpdb->delete(
+			$wpdb->posts,
+			array(
+				'post_type' => self::SLUG,
+			)
+		);
```
felixarntz: Don't we have to clear some caches after running this to avoid serving stale data? I think for the specific post IDs, deleting the caches is irrelevant as they shouldn't be ever accessed once they're gone, but we need to pay attention to query caching. I think at a minimum we have to update WordPress's "last changed" value?

plugins/optimization-detective/class-od-url-metrics-post-type.php
```diff
@@ -0,0 +1,329 @@
...
+					'before' => $one_month_ago,
+				),
+			)
+		);
+
+		foreach ( $query->posts as $post ) {
+			if ( self::SLUG === $post->post_type ) { // Sanity check.
+				wp_delete_post( $post->ID, true );
+			}
+		}
```
felixarntz: I wonder how costly this is, given it can be 100 posts. It may be more efficient to run custom DB queries like you do in the `delete_all_posts()` method, using `ID IN (... )`, rather than lots of individual queries plus all those hooks.

plugins/optimization-detective/class-od-url-metrics-post-type.php
```diff
@@ -0,0 +1,329 @@
...
+
+	/**
+	 * Parses post content in URL metrics post.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param WP_Post $post URL metrics post.
+	 * @return OD_URL_Metric[] URL metrics.
+	 */
+	public static function parse_post_content( WP_Post $post ): array {
```
felixarntz: I think this method name could be made more clear in what it does (_how_ or _why_ does it parse post content). Arguably mentioning "post content" isn't necessary since what needs to be provided is the post, so the fact that it uses the content is an implementation detail. For example the method could be called `get_url_metrics_from_post()`.

plugins/optimization-detective/class-od-url-metrics-post-type.php
felixarntz: It's a bit odd adding general hooks in the same method where the post type is registered. Two alternative approaches I think would be cleaner:
* Either break out those two hook additions into a separate method, e.g. `add_hooks()`.
* Or use this method as the single "add hook" entry point (could be called `add_hooks()`, or you stick with `register()`), and move the `register_post_type()` call into a distinct `register_post_type()` method, which this entry point method then adds as an `init` action callback.

I think I prefer the second approach. This way you have one dedicated method to register the post type, but the only method that needs to be called to bootstrap what the class does is the one entry point method which adds the three hooks.

plugins/optimization-detective/class-od-url-metrics-post-type.php
```diff
@@ -0,0 +1,329 @@
...
+	const GC_CRON_RECURRENCE = 'daily';
+
+	/**
+	 * Registers post type for URL metrics storage.
+	 *
+	 * This the configuration for this post type is similar to the oembed_cache in core.
+	 *
+	 * @since 0.1.0
+	 */
+	public static function register() {
```
felixarntz: Using static methods should usually be reserved for stateless functionality, however most of the methods in this class modify state (either by what's registered in WordPress or the database). Static methods are difficult and sometimes impossible to unit test if they modify state.

To be fair, in the specific cases here I don't see a strong reason to make them regular instance methods. But we can't change it later once this is part of an API, so I think it would still be a more future-proof choice to avoid static methods.

We can add a static `instance()` method or similar to make it trivial to get the "main instance" of the class. That doesn't make the code more complicated, but enhances testability for the future.

plugins/optimization-detective/class-od-url-metrics-post-type.php
westonruter: Since the only plugin code that would be querying these posts is being uninstalled, I didn't think it was worthwhile clearing the cache which could be a significant performance hit.

What last changed value are you referring to?

plugins/optimization-detective/class-od-url-metrics-post-type.php
westonruter: In this case I was thinking it is worthwhile to run the hooks since the plugin code is not uninstalled and there should be potential side effects.

plugins/optimization-detective/class-od-url-metrics-post-type.php
westonruter: Humm, but the class is stateless. There aren't any instances variables. A namespace would essentially work the same here as a class.

westonruter: Also, this class is marked as being `@access private` (although `@internal` maybe would be better). So the API is advertising as not being public for others to use.

plugins/optimization-detective/class-od-url-metrics-post-type.php
felixarntz: I didn't mean that we should clear (flush) the entire cache, but rather invalidate relevant caches affected by the deletion of these posts.

We need to update the "last changed" value that applies to all WordPress posts, so that `WP_Query` won't serve any stale data. This value is changed whenever any post, of any post type, is changed.

This is trivial to do by calling `wp_cache_set_posts_last_changed()`

plugins/optimization-detective/class-od-url-metrics-post-type.php
felixarntz: The class isn't really stateless even though it looks like it. It calls global functions that modify state in WordPress. I think the only method in the class that is actually stateless is the `parse_post_content()` method which returns URL metrics from a given post.

The other methods either change state of WordPress or at least rely on persistence layers outside of the class.

westonruter: I meant stateless in the context of the class. So there's nothing here about this being `static` that impacts testability.

felixarntz: Yeah that's fair. Let's keep as is for now, given the class isn't part of any external API (via the `@access private`).

plugins/optimization-detective/class-od-url-metrics-post-type.php
westonruter: Moved in 368112d1

plugins/optimization-detective/class-od-url-metrics-post-type.php
westonruter: Added in a620bd0a

plugins/optimization-detective/class-od-url-metrics-post-type.php
westonruter: Renamed in a301c938

felixarntz:
@westonruter Pending https://github.com/WordPress/performance/pull/1055#discussion_r1530845169, this LGTM. :)

plugins/optimization-detective/class-od-url-metrics-post-type.php
westonruter: Done in f6e06834

plugins/optimization-detective/load.php
```diff
@@ -38,10 +38,9 @@
 require_once __DIR__ . '/storage/class-od-url-metrics-post-type.php';
 require_once __DIR__ . '/storage/data.php';
 require_once __DIR__ . '/storage/rest-api.php';
+OD_URL_Metrics_Post_Type::add_hooks();
```
felixarntz: Curious: Why not move it below all the `require_once`s like you had it before? Feels a bit odd to have "logic" mixed up with loading files, but that may be a personal preference.

felixarntz (approved):
@westonruter Looks great, just one small detail.

plugins/optimization-detective/load.php
felixarntz: Or actually: Why not put this into `hooks.php`? I'd argue that file should probably be loaded last and add all the hooks, similar to how `default-filters.php` mostly does in WP core.

westonruter: Alternatively, this call could be in `storage.php` which gets required here, but it would be the only code in the file. Down further, `optimization.php` adds a hook as well:

https://github.com/WordPress/performance/blob/18d36fea4d1db730c9443ff776e1f5c0b68d5013/plugins/optimization-detective/optimization.php#L29

Should these two hooks be moved to `load.php` so that none of the requires have any side effects?

westonruter: I had put it here because it goes with the other storage logic.

westonruter: Oh, `hooks.php`, right 😊

westonruter: Done in b74f7ebe180ec7bbe1062cfd57894970d9d21351

plugins/optimization-detective/load.php
```diff
@@ -27,19 +27,22 @@
...
 define( 'OPTIMIZATION_DETECTIVE_VERSION', '0.1.0' );
 
-require_once __DIR__ . '/hooks.php';
-
 // Storage logic.
 require_once __DIR__ . '/class-od-data-validation-exception.php';
 require_once __DIR__ . '/class-od-url-metric.php';
 require_once __DIR__ . '/class-od-url-metrics-group.php';
 require_once __DIR__ . '/class-od-url-metrics-group-collection.php';
 require_once __DIR__ . '/class-od-storage-lock.php';
```
westonruter: I guess all of these should go into `storage` then too.

westonruter: Moved in ~b5623fd0~ 814f57db749bd2fd91c71658721df2f428c8f66a

felixarntz: @westonruter Maybe. `OD_Storage_Lock` probably yes, for the URL metric classes not necessarily - though I don't feel strongly about it. Those are the central infrastructure classes used for the entire functionality, IMO they aren't _just_ storage related as e.g. the post type class is.

I'm okay with either, though please align the test files structure with whatever you go with here as well (e.g. right now the post type class tests are not in the `storage` folder even though that class is in the `storage` folder). Nit-picking... :)

westonruter: Done in 33b02728b262edd2415e5b9d0f06060ac56295fc

PR #1057, opened 2024-03-15
Add generator meta tag for Embed Optimizer

westonruter:
This matches the `meta[name=generator]` added by WebP Uploads and Dominant Color Images.

### DIFF
--- a/plugins/embed-optimizer/hooks.php
+++ b/plugins/embed-optimizer/hooks.php
@@ -155,7 +155,7 @@ function embed_optimizer_trigger_error( string $function_name, string $message, 
 }
 
 /**
- * Displays the HTML generator tag for the WebP Uploads plugin.
+ * Displays the HTML generator tag for the Embed Optimizer plugin.
  *
  * See {@see 'wp_head'}.
  *
--- a/plugins/embed-optimizer/hooks.php
+++ b/plugins/embed-optimizer/hooks.php
@@ -153,3 +153,19 @@ function embed_optimizer_trigger_error( string $function_name, string $message, 
 	}
 	wp_trigger_error( $function_name, $message, $error_level );
 }
+
+/**
+ * Displays the HTML generator tag for the WebP Uploads plugin.
+ *
+ * See {@see 'wp_head'}.
+ *
+ * @since 0.1.0
+ */
+function embed_optimizer_render_generator() {
+	if (
+		defined( 'EMBED_OPTIMIZER_VERSION' )
+	) {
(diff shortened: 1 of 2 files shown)


plugins/embed-optimizer/hooks.php
```diff
@@ -153,3 +153,19 @@ function embed_optimizer_trigger_error( string $function_name, string $message,
 	}
 	wp_trigger_error( $function_name, $message, $error_level );
 }
+
+/**
+ * Displays the HTML generator tag for the WebP Uploads plugin.
```
adamsilverstein:
```suggestion
 * Displays the HTML generator tag for the Embed Optimizer plugin.
```

plugins/embed-optimizer/hooks.php
```diff
@@ -153,3 +153,19 @@ function embed_optimizer_trigger_error( string $function_name, string $message,
...
+ * Displays the HTML generator tag for the Embed Optimizer plugin.
+ *
+ * See {@see 'wp_head'}.
+ *
+ * @since 0.1.0
+ */
+function embed_optimizer_render_generator() {
+	if (
+		defined( 'EMBED_OPTIMIZER_VERSION' )
+	) {
```
westonruter, line 167: This condition seems pointless if we always define the constant. But it is present in the other plugins.

tests/plugins/embed-optimizer/embed-optimizer-test.php
```diff
@@ -156,4 +156,17 @@ function ( array $attrs, string $javascript ) {
 		$this->assertStringStartsWith( '<script type="module" nonce="abc123">', $script );
 		$this->assertStringEndsWith( '</script>', $script );
 	}
+
+	/**
+	 * Test printing the meta generator tag.
+	 *
+	 * @covers ::embed_optimizer_render_generator
+	 */
+	public function test_embed_optimizer_render_generator() {
```
adamsilverstein, line 165: Nice, very thorough!

plugins/embed-optimizer/hooks.php
adamsilverstein: 🤷🏼 very defensive coding

adamsilverstein (approved):
Helpful!

plugins/embed-optimizer/hooks.php
westonruter: Maybe we should also check if `esc_attr()` exists and wrap it in a `try`/`catch` block then 😄

Issue #1051, opened 2024-03-12
Improve fatal detection in php-test-standalone-plugins with E2E test rather than using PHPUnit


westonruter:
Currently, php-test-standalone-plugins workflow (via the [test-plugins](https://github.com/WordPress/performance/blob/trunk/bin/plugin/commands/test-plugins.js) command) is just checking to see if each standalone plugin causes a fatal when it is active in isolation from the other plugins:

https://github.com/WordPress/performance/blob/06d6ffeb05fb5c7d3bb67253b21b924b62fe2a9e/plugin-tests/tests/test-standalone.php#L13-L15

This seems like using the wrong tool, PHPUnit, for the job. Indeed, PHPStan flags this assertion as an error:

> phpstan: Call to method PHPUnit\Framework\Assert::assertTrue() with true will always evaluate to true.

It seems better to not use PHPUnit here, but rather to use `curl` to fetch the homepage and make sure it returns a 200 status code and includes `</html>` in the response body. Even better would be to start using Playwright to do actual E2E tests.

I also think the test-plugins command should also be renamed to reduce confusion with test-php-plugins and test-php, for example, test-e2e.

Originally discussed in https://github.com/WordPress/performance/pull/1044#issuecomment-1989679007 and https://github.com/WordPress/performance/pull/1044#issuecomment-1991866728.

swissspidy:
My previous comment:

> I think that test is just a dummy so that we have at least 1 test there. It's not really trying to detect fatal errors or anything.

What I mean with that:

This test is not supposed to detect fatal errors. It's just a dummy so that PHPUnit doesn't complain that there is no simple test.

We don't need any e2e tests here IMHO, we simply need to write some actual PHPUnit tests testing plugin functionality. Then we can remove this particular test here.

westonruter:
> Even better would be to start using Playwright to do actual E2E tests.

See https://github.com/WordPress/performance/issues/1096

Issue #1031, opened 2024-03-06
Improve UX/UI of enabling performance features via standalone plugins


felixarntz:
After careful consideration, the new UI on the Performance Lab settings screen to install and activate the standalone plugins was not the right call from a UX perspective. It is confusing to find the UI to install and activate plugins replicated in that area, and it fails to convey the point of why it's there.

**We should iterate on the UI, to focus it on features rather than plugins.**

The standalone plugins are, in a way, just a technicality. Their goal is to deliver performance features to WordPress sites in advance of a potential future WordPress core merge. Therefore, we should have cards of _features_, rather than cards of _plugins_. Under the hood, clicking the "Enable feature" button would still install and activate the underlying plugin. But that shouldn't be the focus of the UI, and it doesn't make sense to have it broken down into two distinct tasks either (install and activate can happen at once).

We can take inspiration from how Jetpack solves this problem. Consider the below screenshot:
<img width="1005" alt="Screenshot 2024-03-05 at 7 19 07 PM" src="https://github.com/WordPress/performance/assets/3531426/c9d063ca-05f1-41db-adb8-24659e98fa1a">

Most of the cards refer to other plugins, and they are conveniently installed and activated under the hood. In our case, we wouldn't need a "Learn more" since all that Jetpack exposes there is information on the different plans and pricing, which doesn't apply here. We can surface sufficient information right in the card.

Let's focus this issue on implementing an MVP solution that revises the plugin cards to become feature cards, with plugin installation and activation happening in the background. We can reuse the existing design where possible (i.e. no need for different branding or anything).

**Note:** The modules part of the Performance Lab screen UI is being removed as part of #1029. So this can be ignored here as it will be gone soon anyway.

felixarntz:
For an MVP, I propose the following:
* Reduce the plugin cards so that they only show the following:
    * Feature name
    * Brief feature description (value proposition)
    * A button to activate the feature or, if the feature is already active, an indicator that it is active
* The feature name and description shouldn't be simply the content pulled from .org, but rather custom content that is more tailored to the user experience of what the feature unlocks. Some of the plugin names and descriptions may also work well for the feature names and descriptions, but we should reconsider each one to think whether we can make improvements. For example "WebP Uploads" may work better as something like "WebP format for images", or shorter "WebP images". Or "Speculation Rules" may work better as something like "Speculative preloading" or "Navigational preloading".
[...]

PR #1060, opened 2024-03-18
Remove modules infrastructure and UI from the plugin

### LINKED ISSUE #1030: Remove modules infrastructure

mukeshpanchal27:
## Summary

Fixes #1030

Things needs to updates:
- Document: https://make.wordpress.org/performance/handbook/performance-lab/releasing-the-plugin/

**Note:** In `.github/workflows/deploy-standalone-plugins.yml`, I've included some 'TODO' items that need to be addressed and removed before the merge.


mukeshpanchal27:
For the CODEOWNERS files:
- Should we remove `Focus` areas?
- Do we need to add a new section called `Site Health`, and include `WebP Support Health Check`, `Autoloaded Options Health Check`, and `Enqueued Assets Health Check` with their respective owners? What do you think?

.github/workflows/deploy-standalone-plugins.yml
thelovekesh, line 1: This workflow is updated at https://github.com/WordPress/performance/pull/1033. @felixarntz @mukeshpanchal27 Can we please aim to merge https://github.com/WordPress/performance/pull/1033 before we merge this PR?

plugins.json
```diff
@@ -1,5 +1,4 @@
 {
```
thelovekesh, line 1: Should this be updated to
```json
[
  "auto-sizes",
  "dominant-color-images",
  "speculation-rules",
  "webp-uploads"
]
```
Since `plugins.json` only store plugin names anyway.

readme.txt
```diff
@@ -40,13 +32,13 @@ Currently the plugin includes the following performance modules:
...
+The primary purpose of the Performance Lab plugin is to allow testing of various performance plugins for which the goal is to eventually land in WordPress core. It is essentially a collection of "feature plugins", which makes it different from other performance plugins that offer performance features which are not targeted at WordPress core and potentially rely on functionality that would not be feasible to use in WordPress core. The list of available plugins will regularly change: Existing plugins may be removed after they have been released in WordPress core, while new modules may be added in any release.
```
thelovekesh: > Existing plugins may be removed after they have been released in WordPress core, while new modules may be added in any release.

I think modules is confusing here given we are removing them completely. Should we term it as standalone or feature plugin here?

admin/load.php
```diff
@@ -15,13 +15,8 @@
  * @since 1.0.0
  */
 function perflab_add_modules_page() {
```
thelovekesh: Do we need to change the functions name as well?

.github/workflows/deploy-standalone-plugins.yml
felixarntz: @thelovekesh That works for me. @mukeshpanchal27 What do you think? If we merge #1033 first, there will be merge conflicts here in the workflow files. But since #1033 makes more complex changes to those workflows, it's probably easier to merge that and then fix the little extra things in here.

admin/load.php
felixarntz: +1

Maybe just call it `perflab_add_settings_page()`, so it's not tied to modules/plugins internals. Or it could be called `perflab_add_features_page()`. WDYT?

You can add `@since` annotations like `@since n.e.x.t Renamed to ...`.

admin/load.php
```diff
@@ -15,13 +15,8 @@
...
-	// Don't add a page if active modules are controlled programmatically.
-	if ( has_filter( 'perflab_active_modules' ) ) {
-		return false;
-	}
-
 	$hook_suffix = add_options_page(
-		__( 'Performance Modules', 'performance-lab' ),
+		__( 'Performance Plugins', 'performance-lab' ),
```
felixarntz: How about we already name this "Performance Features" instead, in preparation for #1031?

admin/load.php
```diff
@@ -30,7 +25,12 @@ function perflab_add_modules_page() {
 
 	// Add the following hooks only if the screen was successfully added.
 	if ( false !== $hook_suffix ) {
-		add_action( "load-{$hook_suffix}", 'perflab_load_modules_page', 10, 0 );
+		// Handle script enqueuing for settings page.
+		add_action( 'admin_enqueue_scripts', 'perflab_enqueue_modules_page_scripts' );
+
+		// Handle admin notices for settings page.
+		add_action( 'admin_notices', 'perflab_plugin_admin_notices' );
+
```
felixarntz: Why this change? This adds those action hooks unconditionally, which means they may load all over WP Admin.

I think we should maintain `perflab_load_modules_page()` (just renamed to `perflab_load_settings_page()`, similar to @thelovekesh's previous suggestion) and keep the action hooks in there.

admin/load.php
```diff
@@ -39,350 +39,18 @@ function perflab_add_modules_page() {
...
-	}
-}
-
-/**
- * Renders the modules page.
+ * Renders the plugin page.
  *
  * @since 1.0.0
  */
 function perflab_render_modules_page() {
```
felixarntz: See above, can you rename this accordingly as well?

plugins.json
felixarntz: Maybe. I'd say let's keep as is for now as the extra property doesn't really hurt, and keeping it avoids having to make more changes in this PR.

readme.txt
```diff
@@ -9,19 +9,11 @@ License:           GPLv2 or later
...
-
-Currently the plugin includes the following performance modules:
-
-* **Dominant Color Images:** Adds support to store the dominant color of newly uploaded images and create a placeholder background of that color.
-* **WebP Support Health Check:** Adds a WebP support check in Site Health status.
-* **WebP Uploads:** Creates WebP versions for new JPEG image uploads if supported by the server.
-* **Enqueued Assets Health Check:** Adds a CSS and JS resource check in Site Health status.
-* **Autoloaded Options Health Check:** Adds a check for autoloaded options in Site Health status.
+The Performance Lab plugin is a collection of plugins focused on enhancing performance of your site, most of which should eventually be merged into WordPress core. The plugin allows to individually enable and test the plugins to get their benefits before they become available in WordPress core, and to provide feedback to further improve the solutions.
```
felixarntz: See #1031: We should focus on the _features_ rather than that they are plugins.
```suggestion
The Performance Lab plugin is a collection of features focused on enhancing performance of your site, most of which should eventually be merged into WordPress core. The plugin allows to individually enable and test the features to get their benefits before they become available in WordPress core, and to provide feedback to further improve the solutions.
```

readme.txt
```diff
@@ -40,13 +32,13 @@ Currently the plugin includes the following performance modules:
 = After activation =
 
 1. Visit the new **Settings > Performance** menu.
-2. Enable the individual modules you would like to use.
+2. Enable the individual plugins you would like to use.
```
felixarntz:
```suggestion
2. Enable the individual features you would like to use.
```

readme.txt
felixarntz:
```suggestion
The primary purpose of the Performance Lab plugin is to allow testing of various performance features for which the goal is to eventually land in WordPress core. It is essentially a collection of "feature plugins", which makes it different from other performance plugins that offer performance features which are not targeted at WordPress core and potentially rely on functionality that would not be feasible to use in WordPress core. The list of available features will regularly change: Existing features may be removed after they have been released in WordPress core, while new features may be added in any release.
```

load.php
```diff
@@ -40,178 +40,26 @@
...
-}
-
 /**
  * Gets the content attribute for the generator tag for the Performance Lab plugin.
  *
  * This attribute is then used in {@see perflab_render_generator()}.
  *
  * @since 1.1.0
  * @since 2.9.0 The generator tag now includes the active standalone plugin slugs.
+ * @since n.e.x.t Remove modules from the generator tag.
```
```suggestion
 * @since n.e.x.t The generator tag no longer includes module slugs.
```

load.php
```diff
@@ -231,90 +79,14 @@ function perflab_render_generator() {
...
-	_deprecated_function( __FUNCTION__, 'Performance Lab 2.9.0', "perflab_get_standalone_plugin_version_constants( 'modules' )" );
-	return perflab_get_standalone_plugin_version_constants( 'modules' );
-}
-
-/**
- * Gets the standalone plugin constants used for each available standalone plugin, or module with a standalone plugin.
+ * Gets the standalone plugin constants used for each available standalone plugin.
  *
  * @since 2.9.0
+ * @since n.e.x.t Remove $source parameter.
```
```suggestion
 * @since n.e.x.t The $source parameter was removed.
```

felixarntz (commented):
@mukeshpanchal27 This looks great so far. I left a few comments.

.github/workflows/deploy-standalone-plugins.yml
mukeshpanchal27: Yes, I will take care of that. Thank you for the heads up.

admin/load.php
mukeshpanchal27: Addressed in https://github.com/WordPress/performance/pull/1060/commits/66f0492278e40c1c5a31790b9d20515977e6d3b3

thelovekesh:
@mukeshpanchal27 #1033 caused merge conflicts 😐

.github/workflows/deploy-standalone-plugins.yml
```diff
@@ -91,9 +93,13 @@ jobs:
...
           dry-run: ${{ matrix.dry-run }}
         env:
-          SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
-          SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          # TODO Once the workflow is tested, we will remove the comment and use the secret SVN access.
+          #SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
+          #SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          # TODO Once the workflow is tested, we will remove this test credential.
+          SVN_PASSWORD: SVN_PASSWORD
+          SVN_USERNAME: SVN_USERNAME
```
felixarntz: Please remember to revert this once testing is complete.

Also see the merge conflicts with this file based on #1033.

admin/load.php
```diff
@@ -13,376 +13,56 @@
...
- *                                will rely on {@see perflab_get_modules()}.
- * @param array|null $focus_areas Associative array of focus area data, keyed by focus area slug. By default, this will
- *                                rely on {@see perflab_get_focus_areas()}.
+ * @since n.e.x.t Renamed to perflab_load_settings_page(), and the
+ *                $module and $hook_suffix parameters were removed.
  */
-function perflab_load_modules_page( $modules = null, $focus_areas = null ) {
-	global $wp_settings_sections;
-
+function perflab_load_settings_page() {
```
felixarntz: Since you went with `features_page` in the other function above, please align that here as well.

readme.txt
```diff
@@ -9,19 +9,11 @@ License:           GPLv2 or later
 License URI:       https://www.gnu.org/licenses/gpl-2.0.html
 Tags:              performance, images, javascript, site health, measurement, object caching
 
-Performance plugin from the WordPress Performance Team, which is a collection of standalone performance modules.
+Performance plugin from the WordPress Performance Team, which is a collection of standalone performance plugins.
```
felixarntz: Please update the plugin description like this everywhere, including in `load.php`, `composer.json` etc.
```suggestion
Performance plugin from the WordPress Performance Team, which is a collection of standalone performance features.
```

mukeshpanchal27:
@felixarntz, should we update the setting page slug from the existing `/wp-admin/options-general.php?page=perflab-modules` to `/wp-admin/options-general.php?page=perflab-features`?

thelovekesh:
Overall changes look good to me. I just want to verify whether an updater script is required to remove any options or configurations added to the database by the previous PL plugin that will not be utilized in the next version.

felixarntz:
@mukeshpanchal27 
> @felixarntz, should we update the setting page slug from the existing `/wp-admin/options-general.php?page=perflab-modules` to `/wp-admin/options-general.php?page=perflab-features`?

Great catch! I think this is a good idea, though I would suggest we go with the more generally applicable `/wp-admin/options-general.php?page=performance-lab`. Let's also rename the constant, e.g. to `PERFLAB_SCREEN` or `PERFLAB_FEATURES_SCREEN`.

It would be great to add another hook though which catches requests to `/wp-admin/options-general.php?page=perflab-modules` and redirects those to the new location, e.g. on the `admin_page_access_denied` action hook. For reference, consider [this example](https://github.com/google/site-kit-wp/blob/12bb2a84fa62f9b329ffd750be6876001bc6dd11/includes/Core/Admin/Screens.php#L317) in the Google Site Kit plugin.

@thelovekesh
> I just want to verify whether an updater script is required to remove any options or configurations added to the database by the previous PL plugin that will not be utilized in the next version.

Great point as well! It would be great to implement an update routine which deletes the `'perflab_modules_settings'` option if it is present. Since the option would be autoloaded anyway, I think we could maybe do that simply on `admin_init` if the current user has `manage_options` capabilities. This check should be cheap enough to run in that action hook, and it has the benefit that it works reliably at scale, even on large multisites.

.github/workflows/php-test-standalone-plugins.yml
```diff
@@ -76,8 +76,6 @@ jobs:
           node --version
           composer --version
           php -v
-      - name: Building standalone plugins
-        run: npm run build-plugins
```
westonruter, line 80: Why are the plugins no longer needing to be built? In `bin/plugin/commands/test-plugins.js` it is not accounting for assets which may have been generated by Webpack as it is just copying the source files from `/plugins` to `/build`. It seems rather that `test-plugins.js` should be updated to reuse the output from `npm run build-plugins` which is already outputting the plugins into `/build`. cc @thelovekesh

bin/plugin/commands/test-plugins.js
```diff
@@ -414,55 +414,32 @@ function doRunStandalonePluginTests( settings ) {
...
 						formats.error(
-							`Built plugin path "${ settings.builtPluginsDir }${ modules[ item ].slug }" not found, skipping and removing from plugin list`
+							`Error copying plugin "${ plugin }": ${ e.message }`
 						)
 					);
 					return false;
 				}
-				return true;
 			} )
-			.map( ( item ) => modules[ item ].slug );
```
westonruter, line 436: Per above, is this entire try/catch block even needed if we already have `npm run build-plugins`?

uninstall.php
```diff
@@ -37,6 +37,5 @@
  * @since 1.4.0
  */
 function perflab_delete_plugin_option() {
-	delete_option( 'perflab_modules_settings' );
```
westonruter, line 40: Should this remain? What if someone was on Performance Lab v2 and then they upgrade to v3 but then uninstall. Won't this option be then remaining? Perhaps leave this line here but add a comment to say it was from an older version. 

That brings to mind that it seems at present there is an _autoloaded_ `perflab_modules_settings` option which is getting loaded needlessly. It seems there should be an upgrade routine added for 2.9.0 to 3.0.0 which deletes this option or at least makes it not be autoloaded.

westonruter (changes requested):

uninstall.php
westonruter: Oh, this may be what @thelovekesh mentioned in https://github.com/WordPress/performance/pull/1060#issuecomment-2009992383.

westonruter: Heh, and @felixarntz in https://github.com/WordPress/performance/pull/1060#issuecomment-2010175828.

felixarntz: @westonruter Yeah, I don't think we need to keep it if we instead have a "migration" logic that deletes this option after an update. That way the old option is only referenced in one place, and it makes more sense to delete it after the update rather than only once (or rather if) the plugin is uninstalled.

.github/workflows/php-test-standalone-plugins.yml
mukeshpanchal27: Thank you for pointing that out. It's on my list to address as a follow-up PR since it scratches the current one.

mukeshpanchal27:
@westonruter, I'll update the `test-plugins` command in a follow-up PR.

@felixarntz @thelovekesh, the feedback has been addressed. The PR is now ready for another round of review.

load.php
```diff
@@ -261,6 +261,41 @@ function perflab_maybe_remove_object_cache_dropin() {
...
+}
+add_action( 'admin_page_access_denied', 'perflab_no_access_redirect_module_to_peformance_feature_page' );
+
+/**
+ * Cleanup function to delete 'perflab_modules_settings' option if present.
+ *
+ * @since n.e.x.t
+ */
+function perflab_cleanup_option() {
+	if ( current_user_can( 'manage_options' ) && get_option( 'perflab_modules_settings' ) ) {
```
thelovekesh: Do we need the `get_option()` check? I think `delete_option()` handles it on its own.

thelovekesh (approved):
LGTM! Just a minor feedback ;)

bin/plugin/commands/test-plugins.js
thelovekesh, line 1: Do we still need this given we aren't converting modules to plugins anymore? I am not sure if we really need this.

load.php
mukeshpanchal27: Remove it in https://github.com/WordPress/performance/pull/1060/commits/959377cec9dddc5d8997b7bb5e9bd03e49868f90

bin/plugin/commands/test-plugins.js
mukeshpanchal27: That command simply checks if the plugin activates and doesn't return any errors, so I don't believe we need to remove it. I'd appreciate getting other opinions on this matter.

.github/workflows/php-test-standalone-plugins.yml
felixarntz: Based on what @thelovekesh wrote in https://github.com/WordPress/performance/pull/1060#discussion_r1534051355, do we even need this anymore? The purpose of the `test-plugins` script and this GitHub workflow was purely to test whether standalone plugins don't cause some weird fatal error when built from their modules.

Since we don't build modules into plugins anymore, I think this isn't useful any more. For that reason, I would suggest to remove the whole `test-plugins` script and everything around it. Since this is only removing more things, it would be great if you could do it as part of this PR.

I think some of the code changes in this PR are a bit confusing because they're changing those scripts when they don't really have a good purpose anymore as a whole.

felixarntz (commented):
@mukeshpanchal27 Unless I'm missing something, I think we should remove the `test-plugins` script and everything using it, including the workflow, as this was intended to test plugins that were built from modules, which is no longer relevant. See https://github.com/WordPress/performance/pull/1060/files#r1534169130.

cc @westonruter @thelovekesh

bin/plugin/commands/test-plugins.js
felixarntz: Just replied in https://github.com/WordPress/performance/pull/1060#discussion_r1534169130 :)

westonruter:
> Unless I'm missing something, I think we should remove the `test-plugins` script and everything using it, including the workflow, as this was intended to test plugins that were built from modules, which is no longer relevant. See https://github.com/WordPress/performance/pull/1060/files#r1534169130.

I think there is a better way to check for fatals and that would be through E2E tests in https://github.com/WordPress/performance/issues/1051. I think it makes sense to remove the existing `test-plugins` script as it currently stands.

This can be done in a separate PR, AFAIK.

load.php
```diff
@@ -507,171 +262,44 @@ function perflab_maybe_remove_object_cache_dropin() {
...
-	}
-	$module = require $module_activation_file;
-	if ( ! is_callable( $module ) ) {
-		return;
+	if ( current_user_can( 'manage_options' ) ) {
+		wp_safe_redirect(
+			add_query_arg( 'page', PERFLAB_SCREEN )
+		);
+		exit;
 	}
```
westonruter: This would be better (as in more testable) if the `exit` only happens conditionally based on whether the redirect happens:

```suggestion
	if ( 
		current_user_can( 'manage_options' ) && 
		wp_safe_redirect( add_query_arg( 'page', PERFLAB_SCREEN ) ) 
	) {
		exit;
	}
```

But apparently this isn't being tested right now, so not critical to change here.

westonruter (approved):
Just a nit, but this looks like it can be merged and then further iterated on in subsequent PRs.

felixarntz (approved):
I agree with @westonruter's assessment, let's merge this to unblock other PRs. We can remove the `test-plugins` command in a subsequent PR.

PR #1062, opened 2024-03-19
Bump minimum required WordPress version to 6.4

### LINKED ISSUE #1056: Bump minimum required WP version to 6.4
For context on why I'm opening this issue, please see https://make.wordpress.org/performance/handbook/performance-lab/version-support-policy/. Specifically, the following:

"The Performance Lab plugin commits to supporting the latest stable version of WordPress core."

Since we use some WP 6.4 functions see https://github.com/WordPress/performance/pull/1048#discussion_r1524550877, I propose changing the minimum WP version set to 6.4. This way, we won't need to handle safeguard or old version compatibility code.

mukeshpanchal27:
## Summary

Fixes #1056

### DIFF
--- a/plugins/speculation-rules/hooks.php
+++ b/plugins/speculation-rules/hooks.php
@@ -27,7 +27,6 @@ function plsr_print_speculation_rules() {
 	// This workaround is needed for WP 6.4. See <https://core.trac.wordpress.org/ticket/60320>.
 	$needs_html5_workaround = (
 		! current_theme_supports( 'html5', 'script' ) &&
-		version_compare( strtok( get_bloginfo( 'version' ), '-' ), '6.4', '>=' ) &&
 		version_compare( strtok( get_bloginfo( 'version' ), '-' ), '6.5', '<' )
 	);
 	if ( $needs_html5_workaround ) {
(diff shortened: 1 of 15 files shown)


plugins/speculation-rules/hooks.php
```diff
@@ -27,7 +27,6 @@ function plsr_print_speculation_rules() {
 	// This workaround is needed for WP 6.4. See <https://core.trac.wordpress.org/ticket/60320>.
 	$needs_html5_workaround = (
 		! current_theme_supports( 'html5', 'script' ) &&
-		version_compare( strtok( get_bloginfo( 'version' ), '-' ), '6.4', '>=' ) &&
```
westonruter: I think this makes sense to still keep since it was limited to 6.4. Otherwise this logic would seem to say that the issue was fixed in 6.5 but it existed for all previous versions. When we bump the minimum version to 6.5 then we can remove the entire condition.

westonruter (approved):
LGTM, but with one suggestion

felixarntz (approved):

thelovekesh (approved):

plugins/speculation-rules/hooks.php
mukeshpanchal27: Got it. Condition reverted in https://github.com/WordPress/performance/pull/1062/commits/81a78fd5024ece2e0a6ce142a934a2669051713b

PR #1063, opened 2024-03-19
Merge trunk into feature/image-loading-optimization

westonruter:
Note that this fixes merge conflicts. Pay special attention to changes in these files:

```
        both modified:   .github/CODEOWNERS
        both modified:   .github/workflows/deploy-dotorg.yml
        both modified:   .github/workflows/deploy-standalone-plugins.yml
        deleted by them: .github/workflows/php-test-standalone-plugins.yml
        both modified:   .wp-env.json
        both modified:   composer.json
        both modified:   load.php
        both modified:   package-lock.json
        both modified:   package.json
        both modified:   phpunit.xml.dist
        both modified:   plugins.json
        both modified:   tests/plugins/speculation-rules/speculation-rules-test.php
        both added:      webpack.config.js
```

Diff with `trunk`: https://github.com/WordPress/performance/compare/trunk...merge/trunk-into-image-loading-optimization


felixarntz:
@westonruter Is this urgent? Otherwise I'd suggest to wait until you can include #1060 in this as well.

westonruter:
Not urgent. I'll wait.

westonruter:
> @westonruter Is this urgent? Otherwise I'd suggest to wait until you can include #1060 in this as well.

@felixarntz Now up to date with `trunk` as of this moment with the many PRs that were just merged.

felixarntz (commented):
@westonruter With all the merge conflicts, this is difficult to review.

Looking at https://github.com/WordPress/performance/compare/trunk...merge/trunk-into-image-loading-optimization, I am especially curious about whether the build process is handled efficiently here and whether it needs to run in all the instances where it appears to be run in the feature branch (as far as I know, only `optimization-detective` requires a build process at this point because of `web-vitals`).

I'm not opposed to blindly approving this to unblock the work, but I would probably raise the same question once an actual PR from the feature branch to `trunk` would exist, so doesn't hurt figuring this out already :)

westonruter:
@felixarntz yes, it is difficult to review.

> Looking at [trunk...merge/trunk-into-image-loading-optimization](https://github.com/WordPress/performance/compare/trunk...merge/trunk-into-image-loading-optimization), I am especially curious about whether the build process is handled efficiently here and whether it needs to run in all the instances where it appears to be run in the feature branch (as far as I know, only `optimization-detective` requires a build process at this point because of `web-vitals`).

Of the 4 instances, and @thelovekesh would be best to speak to this, it could possibly be omitted from:

https://github.com/WordPress/performance/blob/efb7c1e50802d916a56114d25ea7f664393eea58/.github/workflows/php-test.yml#L65-L66

and

https://github.com/WordPress/performance/blob/efb7c1e50802d916a56114d25ea7f664393eea58/.github/workflows/deploy-dotorg.yml#L23-L24

Since the Performance Lab plugin doesn't _currently_ require assets to be built.

But it would definitely need to be present in the two other locations, since it is specifically for the standalone plugins, including Optimization Detective and (soon Web Worker Offloading):

https://github.com/WordPress/performance/blob/efb7c1e50802d916a56114d25ea7f664393eea58/.github/workflows/php-test-plugins.yml#L64-L65

and

https://github.com/WordPress/performance/blob/efb7c1e50802d916a56114d25ea7f664393eea58/.github/workflows/deploy-standalone-plugins.yml#L37-L38

Maybe we merge this now and then get confirmation when merging into `trunk`?

felixarntz (approved):
Approving to unblock this. Let's make sure to iron out the workflow changes before `trunk` merge though. This kind of change that affects all other plugins too needs to be carefully evaluated.

PR #1065, opened 2024-03-20
Update PHPUnit tests infrastructure

thelovekesh:
## Summary

Previously https://github.com/WordPress/performance/pull/1013
Fixes task 2 of https://github.com/WordPress/performance/issues/1012

## Relevant technical choices
- Move tests to their plugin directories and move any helper files to the `tests/data` directory.
- Add separate `phpunit.xml.dist` to plugin directories. The current setup also allows to override of PHPUnit config per plugin e.g. put a `phpunit.xml` file in the plugin directory and PHPUnit will override `phpunit.xml.dist`.
- Move test commands from composer to npm scripts. NPM allows to run scripts from any directory and its config feature helps provide required args from CLI to npm commands out of the box. See: https://docs.npmjs.com/cli/v10/using-npm/config#command-line-flags


thelovekesh:
Failing tests are unrelated. It wasn't discovered previously as something was going wrong in the `wp-env` based setup. 
See: https://github.com/WordPress/performance/pull/1013#issuecomment-1962589426 and https://github.com/WordPress/performance/pull/1013#issuecomment-1962731568

thelovekesh:
@felixarntz @westonruter Should I skip the failing tests and fix them in a new follow-up PR? Seems like to fix this we need to make some changes in the `dominant-color-images` plugin to specify which image editor needs to be loaded if both `gd` and `imagick` extensions are present and we want to use a specific extension.

felixarntz (changes requested):
@thelovekesh There is a lot of changes to review here. I'd like for us to take a step back.

This PR makes several changes, and I'm not sure all of them are justified. It would be great if you could provide more context on what your goal is with this PR and why/how those changes help us.

For instance, I see the tests are being moved to the individual plugin folders. That generally makes sense to me. However, I also see the `wp-env` setup is replaced with `install-wp-tests.sh`. That to me doesn't make sense because having that `install-wp-tests.sh` script here is far worse from a maintenance perspective than using a setup like `wp-env` which is supposed to "just work".

Let's discuss that more. Generally, I would prefer that we make one change at a time, to avoid such very difficult-to-review PRs.

thelovekesh:
@felixarntz Most of the changes are related to moving test files from `tests/plugins` to their respective directory and it was previously worked at https://github.com/WordPress/performance/pull/1013 and discussed at https://github.com/WordPress/performance/issues/1012

> However, I also see the wp-env setup is replaced with install-wp-tests.sh. That to me doesn't make sense because having that install-wp-tests.sh script here is far worse from a maintenance perspective than using a setup like wp-env which is supposed to "just work".

There are many reasons to not use `wp-env` in the CI. With this setup, the tests are being executed in half of the time `wp-env` takes to start the environment. Also, we identified a few issues in the setup itself which are mentioned in https://github.com/WordPress/performance/issues/973#issuecomment-1962732898. Also from the maintenance perspective, I have used and seen plugins like AMP, and Web Stories using it for years and it doesn't require any maintenance so far.

> Generally, I would prefer that we make one change at a time, to avoid such very difficult-to-review PRs.

If that's the case, I would love to break this PR into chunks. Just LMK your opinion.

westonruter:
> There are many reasons to not use `wp-env` in the CI. With this setup, the tests are being executed in half of the time `wp-env` takes to start the environment. Also, we identified a few issues in the setup itself which are mentioned in [#973 (comment)](https://github.com/WordPress/performance/issues/973#issuecomment-1962732898). Also from the maintenance perspective, I have used and seen plugins like AMP, and Web Stories using it for years and it doesn't require any maintenance so far.

If contributors have a blessed way to run tests without wp-env this may prove advantageous given many developers now are unable to use Docker given the licensing change to Docker Desktop.

felixarntz:
@thelovekesh @westonruter While there are certainly different ways to set up CI workflows and tests, I feel quite strongly about using `wp-env`, for various reasons:
* It's an established project in the WordPress ecosystem, which allows us not to have to worry about all of this ourselves.
* It's used by both WordPress core and Gutenberg, the other two key projects that contributors to our repository are likely familiar with.
* Generally, it "just works", whether in CI or locally. You need Docker, run `npm install`, and the `wp-env` command, and that's it. You don't need to set up anything else. Plugins like [Web Stories point to requirements of a local development environment](https://github.com/GoogleForCreators/web-stories-wp/blob/main/docs/getting-started.md#local-environment), which may look different for everyone and requires more setup that is not documented as part of the repo and thus more complicated and more error-prone. Other than repositories using `wp-env`, I have not seen any plugin repositories on GitHub that were trivial to set up the development environment. We shouldn't move towards a model that makes getting started more complicated than it has to be.

So while based on the above, I'm opposed to using something custom in favor of `wp-env`, of course we can discuss this. But that should happen in an issue of its own. If you break that part out of this PR, we can focus on the other changes, which make more sense to me.

thelovekesh:
> Generally, it "just works", whether in CI or locally.

TBH I don't think so after what's discovered in https://github.com/WordPress/performance/pull/1013#issuecomment-1962731568 and that's why `dominant-color-images` tests are failing.

> It's used by both WordPress core and Gutenberg, the other two key projects that contributors to our repository are likely familiar with.

And they can use it locally even for running their tests. But in CI, I don't think it will affect any user workflow. Rather it will act as a source of truth that tests are being run as expected in an expected environment. In the current setup, I don't think we are even able to validate the environment configuration apart from PHP version.

felixarntz:
> > Generally, it "just works", whether in CI or locally.
> 
> TBH I don't think so after what's discovered in [#1013 (comment)](https://github.com/WordPress/performance/pull/1013#issuecomment-1962731568) and that's why `dominant-color-images` tests are failing.

I was referring to the development environment generally "just working", as in not requiring any external setup that is not part of the repository's documentation. I completely get your point though, but I would argue this kind of thing may happen and we can try to fix it upstream in https://github.com/WordPress/gutenberg/tree/trunk/packages/env.

> Rather it will act as a source of truth that tests are being run as expected in an expected environment. In the current setup, I don't think we are even able to validate the environment configuration apart from PHP version.

I'm not sure I understand what you mean. As far as I know, we can customize how `wp-env` is configured to a certain degree. And if anything, using it should imply a consistent environment better than if we were requiring _any_ WordPress development environment to be set up. What concretely is `wp-env` lacking for CI that a custom implementation could achieve? And would it make sense to potentially add those things to the upstream package?

thelovekesh:
> I'm not sure I understand what you mean. 

I referred to the logging of environment details like PHP, WP, and PHPUnit versions.

> I would argue this kind of thing may happen and we can try to fix it upstream

And how do we identify such problems? If we hadn't run our test suite with WP Tests setup we wouldn't have identified this issue in the first place. WP Tests setup is also from the WP ecosystem and used widely(just came to know site-kit also uses install-wp-tests 🤪). 

 Maybe we can bring this discussion to https://github.com/WordPress/performance/issues/973 and discuss it widely?

felixarntz:
@thelovekesh We can consider enhancing `wp-env` to give us the logging that we're looking for. `wp-env` is an established solution that works out of the box, and is part of the WordPress project, so I don't think it's a blocker. We can contribute to it upstream to cater for our problems. Unless we find that our proposed changes don't make it into the product, I think it's inefficient to replicate similar functionality with our own solution, worse from a maintenance perspective.

Regarding `install-wp-tests.sh`, I know this is established too, but I really don't think it's a great practice for this file to just be copied around. It seems every repository is including it, there's no one place to include it from, which is not great from a maintenance perspective. You could almost argue it's not even an established approach, as everyone has their own versions of this file which may differ.

Overall, I completely agree this should be discussed elsewhere. Though I think replacing `wp-env` with something else should be discussed in its own issue, as it has wide-reaching implications. #973 doesn't look right for this as it's merely focused on expanding capabilities of our workflows. Let's not discuss too many different things in the same issue, as that ends up very difficult to navigate. For `wp-env`, we will need to carefully consider _why_ it may not work for us and _how_ an alternative solution would work better.

thelovekesh:
> We can consider enhancing wp-env to give us the logging that we're looking for. wp-env is an established solution that works out of the box, and is part of the WordPress project, so I don't think it's a blocker. We can contribute to it upstream to cater for our problems.

We can always contribute upstream to make it a better tool for local development while I think fitting it into CI is a bit overkill, and I would say `install-wp-tests.sh` is also a part of the WordPress project designed specifically for testing purposes.

>  I know this is established too, but I really don't think it's a great practice for this file to just be copied around. It seems every repository is including it, there's no one place to include it from, which is not great from a maintenance perspective.

I think this is also the case with `wp-env` where we have to add a workaround to make it work in CI by adding a `wp-env` override config so a little maintenance burden will be on both sides.

As mentioned above, to discuss it more widely I have opened https://github.com/WordPress/performance/issues/1085 and I will limit the scope of this PR to organizing test files and improving DX in testing.

felixarntz:
@thelovekesh Sounds good. Let me know once you've updated the PR.

thelovekesh:
Thanks @felixarntz. I have made the required updates and PR is ready for review. Please note, that the PR looks huge but most changes are of moving files from one directory to another.

tests/multisite.xml
thelovekesh, line 1: This config does two things:
- Set `WP_TESTS_MULTISITE` constant
- Exclude the `ms-excluded` group.

Passing this file as config takes away the power of overriding config and adds duplicated config. We can also achieve the same thing using `env WP_MULTISITE=1 composer phpunit -- --exclude-group ms-excluded` and re-use the existing PHPUnit config. Please let me know if it makes sense to remove it or not.

phpunit.xml.dist
```diff
@@ -5,35 +5,12 @@
...
-		</testsuite>
-		<testsuite name="webp-uploads">
-			<directory suffix=".php">./tests/plugins/webp-uploads</directory>
 		</testsuite>
 	</testsuites>
-	<groups>
-		<exclude>
-			<group>ms-required</group>
-		</exclude>
-	</groups>
```
thelovekesh, line 38: Neither the PL plugin nor standalone plugin uses this group at the moment. Once this is required, it can be added in required plugin config.

.github/CODEOWNERS
```diff
@@ -32,7 +32,6 @@
 # Module: WebP Support Health Check
 /modules/images/webp-support                                                @adamsilverstein
```
mukeshpanchal27: We removed the module infrastructure, so we should update this accordingly. I have opened PR #1094 to address it.

.github/workflows/php-test-plugins.yml
```diff
@@ -71,6 +71,6 @@ jobs:
       - name: Composer update
         run: npm run wp-env run tests-cli -- --env-cwd="wp-content/plugins/$(basename $(pwd))" composer update --no-interaction
       - name: Running single site unit tests for plugins
-        run: npm run test-php-plugins
```
mukeshpanchal27, line 74: As you removed `multisite.xml`, Could you remove `tests/multisite.xml`?

thelovekesh: Did that in https://github.com/WordPress/performance/pull/1065/commits/bd66bb964b629a07c0bb83036f4bd7162187ac3e

mukeshpanchal27: I mean we should remove that file reference from workflow file.

thelovekesh: Ahh okay. Done in https://github.com/WordPress/performance/pull/1065/commits/61eb09297b58a19c373ca2381538a5a9291452c1.

thelovekesh:
@felixarntz @westonruter Can you please give this a round of review? So that I can work on it over the weekend if there is any feedback.

package.json
```diff
@@ -42,10 +42,25 @@
     "format-php": "composer format",
     "phpstan": "composer phpstan",
     "lint-php": "composer lint:all",
-    "test-php": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/$(basename $(pwd)) composer test",
-    "test-php-multisite": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/$(basename $(pwd)) composer test-multisite",
-    "test-php-plugins": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/$(basename $(pwd)) composer test:plugins",
-    "test-php-multisite-plugins": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/$(basename $(pwd)) composer test-multisite:plugins",
+    "test:php": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer phpunit",
+    "test-php-plugins": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance/plugins/$npm_config_plugin env WP_MULTISITE=$npm_config_multisite php ../../vendor/bin/phpunit --verbose",
```
mukeshpanchal27: Does this cmd work for you?

package.json
```diff
@@ -42,10 +42,25 @@
     "format-php": "composer format",
     "phpstan": "composer phpstan",
     "lint-php": "composer lint:all",
-    "test-php": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/$(basename $(pwd)) composer test",
-    "test-php-multisite": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/$(basename $(pwd)) composer test-multisite",
-    "test-php-plugins": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/$(basename $(pwd)) composer test:plugins",
-    "test-php-multisite-plugins": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/$(basename $(pwd)) composer test-multisite:plugins",
+    "test:php": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer phpunit",
```
mukeshpanchal27: We should update cmd related changes to handbook. Here: https://make.wordpress.org/performance/handbook/performance-lab/#useful-commands

phpunit.xml.dist
```diff
@@ -5,35 +5,12 @@
 	convertErrorsToExceptions="true"
 	convertNoticesToExceptions="true"
 	convertWarningsToExceptions="true"
+	defaultTestSuite="default"
 	>
 	<testsuites>
-		<testsuite name="performance-lab">
+		<testsuite name="default">
```
mukeshpanchal27, line 11: Could we utilize a root file as a configuration for all plugins, considering that all files have the same configuration? 🤔

package.json
thelovekesh: This non-user-facing command is extended by commands like `test:php:auto-sizes` etc. In case someone wants to use it directly, a `--plugin` flag should be passed e.g. `npm run test-php-plugins --plugin=auto-sizes`.

thelovekesh: My thought process here is to use `-` in non-user-facing commands and use `:` separator in user-facing commands. Please LMK if it makes sense.

package.json
thelovekesh: cc @felixarntz. 

Is there a way to update docs on `make.wordpress.org` from this repo?

phpunit.xml.dist
thelovekesh: I did try it but can't find a way to achieve it with a single config 😞

composer.json
```diff
@@ -41,56 +42,13 @@
     "lint:embed-optimizer": "@lint -- ./plugins/embed-optimizer --standard=./plugins/embed-optimizer/phpcs.xml.dist",
     "lint:optimization-detective": "@lint -- ./plugins/optimization-detective --standard=./plugins/optimization-detective/phpcs.xml.dist",
     "lint:speculation-rules": "@lint -- ./plugins/speculation-rules --standard=./plugins/speculation-rules/phpcs.xml.dist",
-    "lint:webp-uploads": "@lint -- ./plugins/webp-uploads --standard=./plugins/webp-uploads/phpcs.xml.dist",
-    "test": "phpunit --verbose --testsuite performance-lab",
```
joemcgill, line 45: Why remove all of the test scripts from composer? I think it would be better if scripts that rely on PHP functionality were maintained in the composer.json and provide a node script that referenced composer than moving all of these over to the package.json file.

package.json
joemcgill: This set of commands is getting pretty verbose. I think it could be easier easier to have a single bash file or node CLI script to run these tests, which can accept different arguments for a single plugin or configuration (e.g., multisite) rather than needing to maintain separate scripts for each plugin here.

joemcgill (changes requested):
I'm still unclear why it is more beneficial to move each plugins tests, config, and test data to their respective plugin's directory, rather than maintaining a single test folder that covers all of the plugins in the mono-repo. There are a few things that I think could be downsides to this approach, or that are at least missing from the current implementation, including:

* All tests are being run independently, meaning we wouldn't catch any compatibility failures between plugins. Since we generally intend for many of these to be run together, I think that's something we should be looking out for.
* Inability to share test data between plugins.
* Inability to share fixtures or test helpers between plugins.

Ideally, as I mentioned in https://github.com/WordPress/performance/issues/1012#issuecomment-1967337536, I think we would be better off having a single combined test suite that covered all of the plugins, but could be filtered to individual plugins for convenience, similar to how you can filter a large test suite by passing a `--group` or `--filter` arg to PHPUnit.

felixarntz:
+1 to @joemcgill's feedback. FWIW the discussion on whether to use individual PHPUnit test suites per plugin in #1012 was never resolved, so IMO it's a bit premature that this PR now implements it.

While it's possible that we won't find a solution that all of us fully agree with, we should still make an effort to facilitate / formalize a decision in the issue before diving into the code to do it. I think we should take a step back here and instead continue the conversation in the issue. What concrete problems are we trying to solve here, and how would individual test suites per plugin address them?

thelovekesh:
> I think it could be easier to have a single bash file or node CLI script to run these tests, which can accept different arguments for a single plugin or configuration (e.g., multisite) rather than needing to maintain separate scripts for each plugin here.

@joemcgill NPM provides this out of the box so I don't think we need to handle it via some script. As you can see above, a command defined which uses npm config args:

```
"test-php-plugins": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance/plugins/$npm_config_plugin env WP_MULTISITE=$npm_config_multisite php ../../vendor/bin/phpunit --verbose",
```

Now we can simply run this script by passing `npm run test-php-plugins --plugin=plugin-name` and/or `test-php-plugins --plugin=plugin-name --multisite=1`.

Now coming to the point of using Composer, I think it's become hard to run composer scripts from different directories. Since this is a mono repo now, it's not guaranteed that someone will always develop from the root directory, on the other hand, NPM has inbuilt features for such use cases.

composer.json
thelovekesh: @joemcgill @felixarntz In conjunction with both this and https://github.com/WordPress/performance/pull/1065#discussion_r1546422938, I have added a node script to run PHPUnit tests in https://github.com/WordPress/performance/pull/1065/commits/0ba78d9b6678f672c23d59d442712ae5e3df3f80 and added the scripts in `package.json` due to some NPM specific features.

Now the commands works like:
```
npm run test:php # Run tests for PL plugin
npm run test:php:plugins # Run tests for all plugins
npm run test:php:plugins --plugin=foo # Run tests for foo plugin
npm run test:php:plugins --plugin=foo -- --phpunit-args # Run tests for foo plugin and pass args to PHPUnit.
// ... and same as above for mutlsite just need to append :multisite in commands.
```

Please LMKWYT.

westonruter: This is nice. Previously in order to run just the tests for Optimization Detective, I would do this:

```bash
npm run wp-env run tests-cli bash
```

And then in the container:

```bash
cd wp-content/plugins/performance/
composer run-script test:optimization-detective # (run as many times as needed)
```

Great to have a command to run just one plugin's tests:

```bash
npm run test:php:plugins --plugin=optimization-detective
```

thelovekesh: One more advantage is, that you can now pass args to PHPUnit by escaping once only like:

```
npm run test:php:plugins --plugin=optimization-detective -- --filter=Foo_Tests
```
is makes it more better IMO.

.github/workflows/php-test-plugins.yml
```diff
@@ -52,10 +50,10 @@ jobs:
       WP_ENV_PHP_VERSION: ${{ matrix.php }}
       WP_ENV_CORE: ${{ matrix.wp == 'trunk' && 'WordPress/WordPress' || format( 'https://wordpress.org/wordpress-{0}.zip', matrix.wp ) }}
     steps:
-      - uses: styfle/cancel-workflow-action@0.12.1
-      - uses: actions/checkout@v4
+      - uses: styfle/cancel-workflow-action@0.11.0
+      - uses: actions/checkout@v3
       - name: Setup Node.js (.nvmrc)
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v3
```
mukeshpanchal27, line 56: Should be reverted as trunk now have latest dependancy.

bin/phpunit/test.js
```diff
@@ -0,0 +1,89 @@
...
+	if (
+		'performance-lab' !== WPP_PLUGIN &&
+		! plugins.includes( WPP_PLUGIN )
+	) {
+		// eslint-disable-next-line no-console
+		console.error(
+			formats.error(
+				`The plugin ${ WPP_PLUGIN } is not a valid plugin managed as part of this project.`
+			)
+		);
```
mukeshpanchal27, line 34: Can we utilize `log` from `require( '../plugin/lib/logger' )`?

```suggestion
		log(
			formats.error(
				`The plugin ${ WPP_PLUGIN } is not a valid plugin managed as part of this project.`
			)
		);
```

tests/bootstrap-functions.php
```diff
@@ -0,0 +1,122 @@
...
+			return normalize_path( $tests_dir ) . '/';
+		}
+
+		unset( $tests_dir );
+	}
+
+	/*
+	* Last resort: see if this is a typical WP-CLI scaffold command set-up where a subset of
+	* the WP test files have been put in the system temp directory.
+	*/
```
mukeshpanchal27, line 98:
```suggestion
	/*
	 * Last resort: see if this is a typical WP-CLI scaffold command set-up where a subset of
	 * the WP test files have been put in the system temp directory.
	 */
```

tests/bootstrap-functions.php
```diff
@@ -0,0 +1,122 @@
...
+		}
+
+		unset( $dev_dir );
+	}
+
+	/*
+	* If neither of the constants was set, check whether the plugin is installed
+	* in `src/wp-content/plugins`. In that case, this file would be in
+	* `src/wp-content/plugins/performance/tests`.
+	*/
```
mukeshpanchal27, line 84:
```suggestion
	/*
	 * If neither of the constants was set, check whether the plugin is installed
	 * in `src/wp-content/plugins`. In that case, this file would be in
	 * `src/wp-content/plugins/performance/tests`.
	 */
```

bin/phpunit/test.js
westonruter: I think better to use `console.error()` as it goes to `STDERR` instead of `STDOUT`.

Otherwise, this could be added to logger as an alias in the same way that was done for `console.log`.

bin/phpunit/test.js
```diff
@@ -0,0 +1,89 @@
...
+const pluginBasePath = path.resolve( __dirname, '../../' );
+const pluginBaseName = path.basename( pluginBasePath );
+const pluginsDir = path.resolve( pluginBasePath, 'plugins' );
+const phpunitBin = path.resolve( pluginBasePath, 'vendor/bin/phpunit' );
+const wpEnvBin = path.resolve( pluginBasePath, 'node_modules/.bin/wp-env' );
+const wpPluginsDir = `/var/www/html/wp-content/plugins/${ pluginBaseName }`;
+
+const _plugins = []; // Store absolute paths to plugins.
+
+if ( WPP_PLUGIN ) {
```
felixarntz, line 24: @thelovekesh Unrelated to the ongoing discussion in #1012: If we implement new JS tooling scripts like this, I think they should be incorporated into the existing foundation in `bin/plugin`. It uses `commander` for a full CLI suite, which has some benefits like built-in help / documentation functionality and error handling infrastructure. This could be reused rather than implementing standalone commands like this.

plugins/dominant-color-images/tests/dominant-color-image-editor-imagick-test.php
```diff
@@ -7,10 +7,27 @@
...
+	 * Setup before class.
+	 */
+	public static function wpSetUpBeforeClass() {
+		// Setup site options if it's a multisite network.
+		if ( is_multisite() ) {
+			$site_exts = explode( ' ', get_site_option( 'upload_filetypes', 'jpg jpeg png gif' ) );
+
+			// Add `tiff` and `bmp` to the list of allowed file types.
+			// These are removed by default in multisite.
+			$site_exts[] = 'tiff';
```
adamsilverstein, line 24: do we need tiff and bmp for tests? what about webp/avif?

adamsilverstein:
> TBH I don't think so after what's discovered in [#1013 (comment)](https://github.com/WordPress/performance/pull/1013#issuecomment-1962731568) and that's why `dominant-color-images` tests are failing.
> 
> >

@thelovekesh if the issue for tests passing is with the Imagick implementation, we may be able to fix this by adding a filter to the test to have it use GD instead?

adamsilverstein (commented):
Overall this looks great @thelovekesh - a few moire points from other commenters to address then this should be good to go. Thanks for limiting the scope here to reorganization of files.

thelovekesh:
@westonruter @adamsilverstein Thanks for all the feedback on this. With the recent updates to the trunk, this PR is now out of sync again, and rebasing it has become a significant task.

If you believe these changes are valuable, I will create a new PR once we agree on https://github.com/WordPress/performance/issues/1012.

westonruter:
@thelovekesh How about we create 3 PRs for the items included in this PR:

> 1. Move tests to their plugin directories and move any helper files to the `tests/data` directory.
> 1. Add separate `phpunit.xml.dist` to plugin directories. The current setup also allows to override of PHPUnit config per plugin e.g. put a `phpunit.xml` file in the plugin directory and PHPUnit will override `phpunit.xml.dist`.
> 1. Move test commands from composer to npm scripts. NPM allows to run scripts from any directory and its config feature helps provide required args from CLI to npm commands out of the box. See: https://docs.npmjs.com/cli/v10/using-npm/config#command-line-flags

I think the first point is the least controversial and can be merged promptly, and then we can move on to build on that momentum with the additional improvements?

thelovekesh:
Thanks @westonruter. The above suggestion makes sense to me. I have raised a PR for the first step - https://github.com/WordPress/performance/pull/1262.

PR #1066, opened 2024-03-20
Add Embed Optimizer assets and update readme

westonruter:
Updating assets and readme for initial commit on WordPress.org.


plugins/embed-optimizer/load.php
```diff
@@ -2,7 +2,7 @@
 /**
  * Plugin Name: Embed Optimizer
  * Plugin URI: https://github.com/WordPress/performance/tree/trunk/plugins/embed-optimizer
- * Description: Optimize the performance of embeds.
+ * Description: Optimize the performance of embeds by lazy-loading them.
```
felixarntz: https://github.com/WordPress/performance/issues/1046#issuecomment-2005134234 proposes `Optimizes the performance of embeds by lazy-loading iframes and scripts.`.

Doesn't mean we have to go with that, but let's align this. At the very minimum, this needs to be changed to use a third-person verb.

plugins/embed-optimizer/readme.txt
```diff
@@ -9,11 +9,11 @@ License:           GPLv2 or later
 License URI:       https://www.gnu.org/licenses/gpl-2.0.html
 Tags:              performance, embeds
 
-Optimize the performance of embeds.
+Optimize the performance of embeds by lazy-loading them.
```
felixarntz: See above.

plugins/embed-optimizer/load.php
westonruter: Updated in 175e2ea6

felixarntz (approved):

PR #1067, opened 2024-03-20
Optimization Detective: Add directory assets, update readme, and add generator meta tag

westonruter:
See #869.

* Add WordPress.org plugin directory assets.
* Update readme.txt.
* Add generator meta tag.


plugins/optimization-detective/readme.txt
```diff
@@ -9,11 +9,11 @@ License:           GPLv2 or later
 License URI:       https://www.gnu.org/licenses/gpl-2.0.html
 Tags:              performance, images
 
-Improves accuracy of optimizing the loading of the LCP image by leveraging client-side detection with real user metrics.
+Uses real user metrics to improve heuristics WordPress applies on the frontend to improve image loading priority.
```
westonruter, line 12: @felixarntz Note:  Per https://github.com/WordPress/performance/issues/1046#issuecomment-2005134234, the "Provides an API to" part will be added once the Image Prioritization logic is split out into a dependent plugin. Also, when this does get submitted to SVN, the plugin name will get "(Developer Preview)" added to it.

felixarntz (approved):

PR #1071, opened 2024-03-21
Remove workflow to test standalone plugins built from modules as no longer relevant

felixarntz:
## Summary

Follow up to #1060.
See #1030.

This removes the `test-plugins` script and surrounding workflow, which was used only to check for fatal errors in plugins built from standalone plugins. This is no longer relevant given modules no longer exist.

Note: This change has nothing to do with PHPUnit tests. The `php-test-plugins.yml` workflow is the workflow that covers the standalone plugins' PHPUnit tests, and is still relevant.


thelovekesh (approved):
Thanks @felixarntz.

westonruter (approved):

PR #1072, opened 2024-03-21
Move `admin` to `includes` directory

thelovekesh:
## Summary

Previously https://github.com/WordPress/performance/pull/1042

As discussed in https://github.com/WordPress/performance/issues/1029#issuecomment-1980938480, we discussed moving the top-level PL plugin code to the `includes` directory. In https://github.com/WordPress/performance/pull/1042 we already moved the `site-health` and `server-timing` to the `includes` directory but it seems like the `admin` directory was not moved to the right place. This PR aims to do that.


westonruter (approved):
LGTM

felixarntz (approved):
Great catch, thanks @thelovekesh!

Issue #775, opened 2023-07-10
Increase PHPStan to rule level to 1 (and beyond)


westonruter:
## Feature Description

https://github.com/WordPress/performance/pull/730 introduced PhpStan and started with level 0 conformance. Now we can increase the [rule level](https://phpstan.org/user-guide/rule-levels) to further improve the codebase.

Once this is resolved, another issue should be created to bump the rule level again.

- [x] https://github.com/WordPress/performance/pull/1188
- [x] https://github.com/WordPress/performance/pull/1195
- [x] https://github.com/WordPress/performance/pull/1198
- [x] https://github.com/WordPress/performance/pull/1199
- [x] https://github.com/WordPress/performance/pull/1200
- [x] https://github.com/WordPress/performance/pull/1201
- [x] https://github.com/WordPress/performance/pull/1202
- [x] https://github.com/WordPress/performance/pull/1203
- [x] #1205 
- [x] https://github.com/WordPress/performance/pull/1210
- [x] https://github.com/WordPress/performance/pull/1241
- [ ] https://github.com/WordPress/performance/pull/1211 (maybe not)
- [x] https://github.com/WordPress/performance/issues/1209

ValeriiVasyliev:
On it

westonruter:
Over halfway to level 6 now!

westonruter:
I'm declaring this done! Level 8 seems to be sufficient. There is https://github.com/WordPress/performance/pull/1211 for updating to level 9, but as I mention in that PR, it seems of diminishing value.

PR #1073, opened 2024-03-21
Fix static analysis issues and code quality in PerfLab load.php files

westonruter:
## Summary

I noticed a few issues in the `load.php` and `admin/load.php` files, mainly identified with a stricter level of static analysis in PHPStan (#775) and/or PhpStorm's static analysis. There are other issues in the other PHP files, but I just wanted to limit the scope to these two files (and a JS file). The following issues are fixed:

* The use of `esc_js()` is brittle and should be replaced with `wp_json_encode()`.
* Using `var` in JS is obsolete in favor of `const`/`let`.
* Lacking PHP7 typing on function parameters and return values.
* LanguageTool is complaining about grammar in a couple places.
* Most substantively, the `deactivate_plugins()` function was being used as if it returned a value. It returns `void`.


admin/load.php
```diff
@@ -131,8 +131,8 @@ function perflab_render_pointer( $pointer_id = 'perflab-admin-pointer', $args =
 	<script id="<?php echo esc_attr( $pointer_id ); ?>" type="text/javascript">
 		jQuery( function() {
 			// Pointer Options.
-			var options = {
-				content: '<h3><?php echo esc_js( $args['heading'] ); ?></h3><p><?php echo wp_kses( $args['content'], $wp_kses_options ); ?></p>',
+			const options = {
+				content: <?php echo wp_json_encode( '<h3>' . esc_html( $args['heading'] ) . '</h3><p>' . wp_kses( $args['content'], $wp_kses_options ) . '</p>' ); ?>,
```
felixarntz: Why is `wp_json_encode()` better here? It isn't really JSON, so this feels a bit weird to me semantically.

admin/load.php
```diff
@@ -73,7 +73,7 @@ function perflab_render_settings_page() {
  *
  * @param string $hook_suffix The current admin page.
  */
-function perflab_admin_pointer( $hook_suffix ) {
+function perflab_admin_pointer( string $hook_suffix ) {
```
felixarntz: Using PHP7 type hinting is a bit of a personal preference. I am personally in favor, but we haven't been doing it for the most part, so I feel this shouldn't be dictated by a tool.

I am not strictly opposed to the change, but it feels out of scope here and IMO should rather be the result of a conscious decision that the team is aware of.

admin/load.php
westonruter: Well, a few reasons:

1. It avoids the author having to keep track of the open and closing quotation marks, and remember whether they are single or double.
2. It consolidates the echoes into one rather than splitting it into two.

This also fixes a couple issues:

1. There is no escaping of the heading being done, so it's adding an `esc_html()`. If using `esc_js()` then it should probably should have been `esc_js( esc_html( $args['heading'] ) )`
2. Similarly, the `wp_kses( $args['content'], $wp_kses_options )` should have also been wrapped in `esc_js()` here in case the `content` contains some single quote mark or backslashes or some other character needing escaping.

By using `wp_json_encode()`, the code is made more robust against bugs being introduced.

admin/load.php
westonruter: ok, I reverted in 2269c6017335c995800d28c0a8948e5c921e8559

felixarntz (approved):

mukeshpanchal27 (approved):

PR #1064, opened 2024-03-19
Simplify PR requirements while allowing changelog script to apply to any standalone plugin

felixarntz:
## Summary

This PR changes and simplifies the requirements for labeling of PRs, while at the same time enhancing the changelog generation script to be flexible to support any standalone plugin.

The proposed changes imply the following:
* The `[Focus] ...` or `Infrastructure` labels are no longer required (and are no longer used by any infrastructure).
* Milestones should exist for every individual standalone plugin rather than only for the Performance Lab plugin. They always need to include a version number.
    * Examples: "Performance Lab 3.0.0", "WebP Uploads 1.0.6".
    * Or we could use the slugs, such as "performance-lab 3.0.0", "webp-uploads 1.0.6", which could be beneficial to further integrate with other plugin specific automation workflows. Either way, this PR does not mandate any particular format, so this could be decided separately.
* The `no milestone` label now actually _means_ "no milestone" rather than "omit this PR from the changelog". The only PRs this should be used for is infrastructure PRs that don't apply to a specific plugin (e.g. build / tooling that affects the entire monorepo).
* A new `skip changelog` label should be used to indicate that a PR should be omitted from the changelog. With this label, it is now possible to assign a target milestone to a PR, without that mandating a changelog entry for it.

The changelog script can be used the same way as before. But now it is flexible to handle _any_ milestone for any standalone plugin rather than only allowing `PL Plugin ...` milestones.

For example:
`npm run changelog -- -m "PL Plugin 3.0.0"` or `npm run changelog -- -m "WebP Uploads 1.0.6"` (given that milestone exists)

In a follow up PR, we could consider enhancing the readme script to also support any standalone plugin, by applying the changes to the correct `readme.txt` file.

PR #1075, opened 2024-03-21
Enhance readme CLI command to support standalone plugins

felixarntz:
## Summary

Follow up to #1064: This PR enhances the `npm run readme` command to support standalone plugins. Given the changelog command was already updated, the only necessary change was to dynamically identify _which_ `readme.txt` file to update, as before it would always update the root `readme.txt` file.

## Relevant technical choices

* A `path` option was added, however that's purely for flexibility and is optional.
* By default, the `path` will automatically be detected via the `milestone` option.
    * Going forward, milestones should be named `plugin-slug v.e.r.s.i.o.n` so that this script can detect the relevant plugin automatically.
    * This will most likely be beneficial for other future workflow automation enhancements too.

You can test this by e.g. running:
* `npm run readme -- -m "performance-lab 3.0.0"`
* `npm run readme -- -m "speculation-rules 1.1.0"`

Don't commit the changes though, it's just for testing, particularly as both changelogs at this point contain a few manually added entries that are worth keeping.


bin/plugin/commands/readme.js
```diff
@@ -48,20 +55,53 @@ exports.handler = async ( opt ) => {
...
+	if ( ! slug ) {
+		throw new Error(
+			`The ${ milestone } milestone does not start with a valid plugin slug.`
+		);
+	}
+
+	if ( 'performance-lab' === slug[ 1 ] ) {
+		return 'readme.txt';
+	}
+
```
westonruter, line 81: It could be useful to also check if the path `plugins/${ slug[ 1 ] }/readme.txt` exists as an additional check.

```suggestion

	if ( ! fs.existsSync( `plugins/${ slug[ 1 ] }` ) ) {
		throw new Error( `Unknown plugin with slug '${ slug[ 1 ] }'` );
	}

```

Otherwise, If I add a typo like:

```bash
npm run readme -- -m "speculation-rulez 1.1.0"
```

Then I get an error:

> Error: Cannot find milestone by title: speculation-rulez 1.1.0

But this is the same error you'd get if the version had a typo.

Just an additional safeguard.

westonruter (changes requested):

bin/plugin/commands/readme.js
felixarntz: Great idea, fixed in https://github.com/WordPress/performance/pull/1075/commits/fcd1af2288118b9c4211b2b7602d6ee48251348d.

westonruter (approved):

mukeshpanchal27 (approved):
Thanks for the PR @felixarntz. LGTM!

PR #1068, opened 2024-03-20
Bump minimum WP version for Optimization Detective

westonruter:
See #1056.

PR #1078, opened 2024-03-22
Update readme and phpdoc for initial Optimization Detective release

westonruter:
This updates the `readme.txt` for Optimization Detective along with some phpdoc fixes for an initial release on the WordPress Plugin Directory. It also fixes a bad condition in a `_doing_it_wrong()` check.

The bulk of the readme changes are in a Google Doc which will be added here once reviewed.

### DIFF
--- a/plugins/optimization-detective/detection.php
+++ b/plugins/optimization-detective/detection.php
@@ -23,9 +23,9 @@ function od_get_detection_script( string $slug, OD_URL_Metrics_Group_Collection 
 	/**
 	 * Filters the time window between serve time and run time in which loading detection is allowed to run.
 	 *
-	 * Allow this amount of milliseconds between when the page was first generated (and perhaps cached) and when the
+	 * This the allowance of milliseconds between when the page was first generated (and perhaps cached) and when the
 	 * detect function on the page is allowed to perform its detection logic and submit the request to store the results.
-	 * This avoids situations in which there is missing detection metrics in which case a site with page caching which
+	 * This avoids situations in which there is missing URL Metrics in which case a site with page caching which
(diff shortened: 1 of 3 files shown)


plugins/optimization-detective/readme.txt
```diff
@@ -1,4 +1,4 @@
-=== Optimization Detective ===
+=== Optimization Detective (Developer Preview) ===
 
 Contributors:      wordpressdotorg
 Requires at least: 6.3
```
mukeshpanchal27:
```suggestion
Requires at least: 6.4
```

Should be 6.4

westonruter: Oh, I missed that in https://github.com/WordPress/performance/pull/1068.

Thanks!

plugins/optimization-detective/detection.php
```diff
@@ -23,9 +23,9 @@ function od_get_detection_script( string $slug, OD_URL_Metrics_Group_Collection
 	/**
 	 * Filters the time window between serve time and run time in which loading detection is allowed to run.
 	 *
-	 * Allow this amount of milliseconds between when the page was first generated (and perhaps cached) and when the
+	 * This the allowance of milliseconds between when the page was first generated (and perhaps cached) and when the
```
felixarntz: Typo / missing word here.

felixarntz (approved):
LGTM, just one thing.

plugins/optimization-detective/detection.php
westonruter: God catch

```suggestion
	 * This is the allowance of milliseconds between when the page was first generated (and perhaps cached) and when the
```

😉

PR #1079, opened 2024-03-22
Merge Optimization Detective into trunk

westonruter:
Since Optimization Detective has been approved by the Plugin Review team, it is ready to be committed to WordPress.org. Therefore, it should be merged into `trunk` and the feature branch should be retired.

Blocked by:

- [x] https://github.com/WordPress/performance/pull/1080
- [x] https://github.com/WordPress/performance/pull/1078


westonruter:
Plugin build: [optimization-detective.zip](https://github.com/WordPress/performance/files/14715459/optimization-detective.zip)

.gitignore
```diff
@@ -32,6 +33,7 @@ build-cs/composer.lock
 
 node_modules/
 vendor/
+plugins/optimization-detective/detection/web-vitals*
```
felixarntz, line 36: I think it would be a good idea to use `build` directories within the individual standalone plugins for things like this. If we want to exclude built resources, it shouldn't be necessary to add exclusions for every individual one.

includes/admin/server-timing.php
```diff
@@ -43,16 +43,18 @@ function perflab_add_server_timing_page() {
...
+		 * This settings section technically includes a field, however it is directly rendered as part of the section
+		 * callback due to requiring custom markup.
+		 */
+		add_settings_section(
+			'output-buffering',
+			__( 'Output Buffering', 'performance-lab' ),
+			'perflab_render_server_timing_page_output_buffering_section',
+			PERFLAB_SERVER_TIMING_SCREEN
+		);
+	}
```
felixarntz, line 57: Is this change really needed? Is it harmful to have both output buffers active?

I'm asking since I don't like the idea of excluding PL specific logic because of another standalone plugin. There are lots of WordPress plugins that have output buffers, so it's a bit random we would cater for the one from Optimization Detective here just because it's also our plugin. What would be the problem of having the Server-Timing output buffering active in combination with the one from Optimization Detective?

includes/admin/server-timing.php
```diff
@@ -93,7 +95,7 @@ static function () {
 				);
 				?>
 			</p>
-			<?php if ( ! perflab_server_timing_use_output_buffer() ) : ?>
+			<?php if ( ! perflab_server_timing_use_output_buffer() && ! has_filter( 'template_include', 'od_buffer_output' ) ) : ?>
```
felixarntz, line 98: See above.

includes/server-timing/class-perflab-server-timing.php
```diff
@@ -225,6 +225,21 @@ public function use_output_buffer() {
...
+				'od_template_output_buffer',
+				function ( $buffer ) {
+					$this->send_header();
+					return $buffer;
+				},
+				PHP_INT_MAX
+			);
+			return $passthrough;
+		}
+
```
felixarntz, line 242: See above.

phpcs.xml.dist
```diff
@@ -9,9 +9,13 @@
 		<exclude-pattern>includes/server-timing/object-cache.copy.php</exclude-pattern>
 	</rule>
 
+	<!-- Check up to 20 files simultaneously. -->
+	<arg name="parallel" value="20"/>
```
felixarntz, line 13: Why is this added? Not opposed but would be good to have context since this applies to the whole repository.

.gitignore
```diff
@@ -23,6 +23,7 @@ nbproject/
 
 build
 .wp-env.override.json
+*.asset.php
```
felixarntz, line 26: See my other comment below, if we used a `build` directory, we wouldn't need to explicitly exclude this from here and a few other places.

phpcs.xml.dist
```diff
@@ -9,9 +9,13 @@
...
+
 	<file>.</file>
 	<exclude-pattern>./build/*</exclude-pattern>
 	<exclude-pattern>./node_modules/*</exclude-pattern>
 	<exclude-pattern>./plugins/*</exclude-pattern>
 	<exclude-pattern>./vendor/*</exclude-pattern>
+	<exclude-pattern>*.asset.php</exclude-pattern>
```
felixarntz, line 20: See above.

felixarntz (commented):
@westonruter Almost looks good, a few points of feedback. Please take a look, though we could also consider addressing some of them afterwards.

phpcs.xml.dist
westonruter: It was added in https://github.com/WordPress/performance/commit/c00f6bc7d3e97b5588b5a413cb06d4d8735fd21e as part of #953.

We use it in the AMP plugin to speed up PHPCS: https://github.com/ampproject/amp-wp/blob/d4200c4b26446541282aef3c3cc2acd3b93674d7/.phpcs.xml.dist#L187

includes/admin/server-timing.php
westonruter: The thinking was that if Optimization Detective introduces an official way to do output buffering (for Core-43258), that this can be reused.

.gitignore
westonruter: Sounds good. Let's do that in another PR to `trunk`.

includes/admin/server-timing.php
westonruter: And that if a PL standalone plugin is already doing output buffering, it's kinda unnecessary to do yet more output buffering on top.

phpcs.xml.dist
westonruter: FWIW, Gutenberg does this as well: https://github.com/WordPress/gutenberg/blob/4056660341852e2b627127d2dacbbedd38cb6e3a/phpcs.xml.dist#L80

adamsilverstein (approved):
Looks good, maybe we can address additional feedback in separate PR?

includes/admin/server-timing.php
westonruter: In other words, it's the Performance Lab plugin taking advantage of the capabilities of one of its own feature plugins.

felixarntz: If there was a standard way to do this, that would be great indeed. However, I'm not sure there's good value in connecting the plugins in this way. It comes with its own set of questions, such as:
* Which of the two plugins should be the "source of truth" for output buffering when both are active?
* If PL uses OD's output buffering capabilities, how do we deal with the somewhat confusing implications of that for the Server-Timing screen? It's strange to just hide the control without any explanation.

I'm not sure it's worth discussing those questions given the two plugins could just work independently. And again, it's somewhat arbitrary anyway. If any other plugin can have its own output buffering, I don't think it's important that these two plugins don't both add their own.

For those reasons I'd prefer to either remove this integration to simplify things, or alternatively to polish it. Could be done in a follow up PR though.

westonruter: Since it doesn't impact the Optimization Detective plugin's codebase, let's do in another PR.

felixarntz (approved):
Good to merge, though let's address the two points from my previous review separately.

westonruter:
> let's address the two points from my previous review separately.

See https://github.com/WordPress/performance/pull/1082

phpcs.xml.dist
westonruter: Actually, I tested this and it doesn't seem to help here in Performance Lab. It may be due to the AMP plugin having many more PHP files being tested at once where it then makes an impact. I'm reverting in https://github.com/WordPress/performance/pull/1082.

PR #1084, opened 2024-03-23
Prevent detection script injection from breaking import maps in classic themes

### LINKED ISSUE #1083: Detection script injected by Optimization Detective breaks import maps for classic themes
## Bug Description

When URL Metrics are needed, Optimization Detective injects the detection script at the end of the `head` tag. This turns out to break import maps in classic themes which output the `script[type=importmap]` in the footer. The import map script must be on the page before any module script. An error is in the console:

> An import map is added after module script load was triggered.

When the page has an interactive block present (which uses the Interactivity API and thus tries to import `@wordpress/interactivity` module), another error then occurs because the import map failed to load:

> Uncaught TypeError: Failed to resolve module specifier "@wordpress/interactivity". Relative references must start with either "/", "./", or "../".

This issue does not occur in block themes because the import map script is printed there since it is aware of all module scripts which are later going to be loaded.

Since this involves import maps and module scripts, it is a WordPress 6.5+ issue.

To fix this issue, the detection script should be injected at the end of the `body` tag instead. To facilitate this injection, it may be worthwhile to print a unique placeholder comment at `print_footer_scripts` after we know the import map script has been printed, and then do a replacement of that. Otherwise, searching for `</body>` to replace may not actually target the closing `body` tag (as it may be inside of a comment, for example).

## Steps to reproduce

1. Activate a classic theme, such as Twenty Twenty-One.
1. Add an Image block to a post and enable click-to-expand.
1. Activate Optimization Detective.
1. View the post while being logged-out. 

## Screenshots

image

## Additional Context

 - PHP Version: 8.1
 - Browser: [e.g. chrome, safari]
 - Plugin Version: [e.g. 22]

westonruter:
## Summary

Move detection script module from being injected at the end of `head` to instead happen at the end of `body`. This prevents the script module from breaking import maps, which are printed in the footer in classic themes rather than in the head (as done in block themes). The import map script must be printed before all script modules.

Fixes #1083

## Relevant technical choices

<details><summary>Outdated technical choices</summary>

Previously the detection script module was injected just before the `</head>` closing tag. This broke import maps in classic themes, as an import map script must be added before any script modules, and classic themes output import map scripts in the footer. So instead of injecting the detection script at the end of the `head`, this PR changes it so that they get injected at the point where `wp_print_footer_scripts` happens. This injection happens by means of printing a placeholder string during template rendering and then this placeholder is replaced with the script module during output buffer processing. The placeholder is randomized to prevent accidental replacement of content. A placeholder is used instead of injecting before `</body>` since it may be that this exists inside of a comment, like `<!--</body>-->` in which case the replacement would be invalid.

Note that the injection of the preload links continues to happen just before `</head>`. The reason that a similar placeholder string is not used is that when a string appears in the `head`, the DOM specifies that this should cause the `head` to immediately close and implicitly open the `body`. In case there is any `DOMDocument` processing being done on the page in addition to our output buffer, the presence of a string placeholder would break the DOM. 

Additionally, a mustache-ish tag is used for the placeholder instead of an HTML comment because other plugins may try to optimize the page by removing all HTML comments to reduce page weight.

Ultimately, the injection of the preload links in the `head` and the detection script in the `body` should leverage the HTML API (via `WP_HTML_Processor`) once it is updated to support tag node insertions, which [it does not support yet](https://github.com/WordPress/wordpress-develop/blob/ea7888f114a9fe3c318716684d21703fe146ecbd/src/wp-includes/html-api/class-wp-html-processor.php#L28).

</details> 

Previously the detection script module was injected just before the `</head>` closing tag (using search/replace). This broke import maps in classic themes, as an import map script must be added before any script modules, and classic themes output import map scripts in the footer. Since it was using search/replace, it was also vulnerable to injecting the script in the wrong place, for example if the HTML contained `<!--</head>-->`. Now the HTML Tag Processor is extended (thanks to @dmsnell) to inject the script tag right before the `</body>` closing tag (and not confused by `<!--</body>-->`). The HTML Tag Processor is also used to inject any preload links at the end of the `</head>`. Note also that what was known as `OD_HTML_Tag_Processor` has been renamed to `OD_HTML_Tag_Walker`, and a new class `OD_HTML_Tag_Processor` is introduced which directly extends `WP_HTML_Tag_Processor`. The `OD_HTML_Tag_Walker` class (and the old `OD_HTML_Tag_Processor` class) did not extend `WP_HTML_Tag_Processor` but used it by composition. The new `OD_HTML_Tag_Processor` includes a method for injecting HTML in the `head` and `body`.

Finally, this PR removes the use of `ext-dom` to do equality assertions in PHPUnit. This was overkill and it made it difficult to debug when the assertions started failing.

----

This also fixes (46ca5f8) a trivial bug where an extra slash was present in the path to the web-vitals.js library (e.g. `optimization-detective//build/web-vitals.js`).

### DIFF
--- a/plugins/optimization-detective/class-od-html-tag-processor.php
+++ b/plugins/optimization-detective/class-od-html-tag-processor.php
@@ -194,12 +194,17 @@ final class OD_HTML_Tag_Processor {
 					return false;
 				}
 
+				// WordPress 6.5 changed the $end arg in the WP_HTML_Text_Replacement constructor to $length.
+				static $is_old_text_replacement_signature = null;
+				if ( null === $is_old_text_replacement_signature ) {
+					$is_old_text_replacement_signature = version_compare( get_bloginfo( 'version' ), '6.5', '<' );
+				}
+
 				$start = $this->bookmarks[ $bookmark ]->start;
 
 				$this->lexical_updates[] = new WP_HTML_Text_Replacement(
 					$start,
-					// In WordPress 6.5, the signature was changed from $end to $length.
-					version_compare( get_bloginfo( 'version' ), '6.5', '<' ) ? $start : 0,
+					$is_old_text_replacement_signature ? $start : 0,
 					$html
 				);
 				return true;
--- a/plugins/optimization-detective/class-od-html-tag-processor.php
+++ b/plugins/optimization-detective/class-od-html-tag-processor.php
@@ -164,7 +164,12 @@ final class OD_HTML_Tag_Processor {
 	/**
 	 * Processor.
 	 *
-	 * @var WP_HTML_Tag_Processor
+	 * This contains an instance of an anonymous subclass of WP_HTML_Tag_Processor which adds an append_html method.
+	 * The use of the mixed type is due to the lack of typing being available for such an anonymous class. PHPCS
+	 * requires that there be a type provided, and yet if it is just WP_HTML_Tag_Processor then PHPStan complains about
+	 * the append_html method not being defined.
+	 *
+	 * @var WP_HTML_Tag_Processor|mixed
 	 */
 	private $processor;
 
@@ -422,7 +427,6 @@ final class OD_HTML_Tag_Processor {
 	 * @return bool Whether successful.
 	 */
 	public function append_head_html( string $html ): bool {
-		// @phpstan-ignore-next-line
 		$success = $this->processor->append_html( self::END_OF_HEAD_BOOKMARK, $html );
 		if ( ! $success ) {
 			$this->warn( __( 'Unable to append markup to the HEAD.', 'optimization-detective' ) );
@@ -437,7 +441,6 @@ final class OD_HTML_Tag_Processor {
 	 * @return bool Whether successful.
 	 */
 	public function append_body_html( string $html ): bool {
-		// @phpstan-ignore-next-line
 		$success = $this->processor->append_html( self::END_OF_BODY_BOOKMARK, $html );
 		if ( ! $success ) {
 			$this->warn( __( 'Unable to append markup to the BODY.', 'optimization-detective' ) );
--- a/plugins/optimization-detective/class-od-html-tag-processor.php
+++ b/plugins/optimization-detective/class-od-html-tag-processor.php
@@ -194,9 +194,12 @@ final class OD_HTML_Tag_Processor {
 					return false;
 				}
 
+				$start = $this->bookmarks[ $bookmark ]->start;
+
 				$this->lexical_updates[] = new WP_HTML_Text_Replacement(
-					$this->bookmarks[ $bookmark ]->start,
-					0,
+					$start,
+					// In WordPress 6.5, the signature was changed from $end to $length.
+					version_compare( get_bloginfo( 'version' ), '6.5', '<' ) ? $start : 0,
 					$html
 				);
 				return true;
--- a/plugins/optimization-detective/class-od-html-tag-processor.php
+++ b/plugins/optimization-detective/class-od-html-tag-processor.php
@@ -1,0 +1,62 @@
+<?php
+/**
+ * Optimization Detective: OD_HTML_Tag_Processor class
+ *
+ * @package optimization-detective
+ * @since 0.1.1
+ */
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Extension to WP_HTML_Tag_Processor that supports injecting HTML.
+ *
+ * @since 0.1.1
+ * @access private
+ */
+final class OD_HTML_Tag_Processor extends WP_HTML_Tag_Processor {
+
+	/**
+	 * Whether the old (pre-WP 6.5) signature for WP_HTML_Text_Replacement is needed.
+	 *
+	 * WordPress 6.5 changed the $end arg in the WP_HTML_Text_Replacement constructor to $length.
+	 *
+	 * @var bool
+	 */
+	private $old_text_replacement_signature_needed;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param string $html HTML to process.
+	 */
+	public function __construct( $html ) {
+		$this->old_text_replacement_signature_needed = version_compare( get_bloginfo( 'version' ), '6.5', '<' );
+		parent::__construct( $html );
+	}
+
+	/**
+	 * Appends HTML to the provided bookmark.
+	 *
+	 * @param string $bookmark Bookmark.
+	 * @param string $html     HTML to inject.
+	 * @return bool Whether the HTML was appended.
+	 */
+	public function append_html( string $bookmark, string $html ): bool {
+		if ( ! $this->has_bookmark( $bookmark ) ) {
+			return false;
+		}
+
+		$start = $this->bookmarks[ $bookmark ]->start;
+
+		$this->lexical_updates[] = new WP_HTML_Text_Replacement(
+			$start,
+			$this->old_text_replacement_signature_needed ? $start : 0,
+			$html
+		);
+		return true;
+	}
+}
--- a/plugins/optimization-detective/class-od-html-tag-processor.php
+++ b/plugins/optimization-detective/class-od-html-tag-processor.php
@@ -134,6 +134,20 @@ final class OD_HTML_Tag_Processor {
 	const XPATH_PATTERN = '^(/\*\[\d+\]\[self::.+?\])+$';
 
 	/**
+	 * Bookmark for the end of the HEAD.
+	 *
+	 * @var string
+	 */
+	const END_OF_HEAD_BOOKMARK = 'end_of_head';
+
+	/**
+	 * Bookmark for the end of the BODY.
+	 *
+	 * @var string
+	 */
+	const END_OF_BODY_BOOKMARK = 'end_of_body';
+
+	/**
 	 * Open stack tags.
 	 *
 	 * @var string[]
@@ -160,7 +174,31 @@ final class OD_HTML_Tag_Processor {
 	 * @param string $html HTML to process.
 	 */
 	public function __construct( string $html ) {
-		$this->processor = new WP_HTML_Tag_Processor( $html );
+
+		$processor = new class( $html ) extends WP_HTML_Tag_Processor {
+
+			/**
+			 * Appends HTML to the provided bookmark.
+			 *
+			 * @param string $bookmark Bookmark.
+			 * @param string $html     HTML to inject.
+			 * @return bool Whether the HTML was appended.
+			 */
+			public function append_html( string $bookmark, string $html ): bool {
+				if ( ! $this->has_bookmark( $bookmark ) ) {
+					return false;
+				}
+
+				$this->lexical_updates[] = new WP_HTML_Text_Replacement(
+					$this->bookmarks[ $bookmark ]->start,
+					0,
+					$html
+				);
+				return true;
+			}
+		};
+
+		$this->processor = new $processor( $html );
 	}
 
 	/**
@@ -253,6 +291,13 @@ final class OD_HTML_Tag_Processor {
 							$tag_name
 						)
 					);
+				}
+
+				// Set bookmarks for insertion of preload links and the detection script module.
+				if ( 'HEAD' === $popped_tag_name ) {
+					$p->set_bookmark( self::END_OF_HEAD_BOOKMARK );
+				} elseif ( 'BODY' === $popped_tag_name ) {
+					$p->set_bookmark( self::END_OF_BODY_BOOKMARK );
 				}
 
 				array_splice( $this->open_stack_indices, count( $this->open_stack_tags ) + 1 );
@@ -368,6 +413,36 @@ final class OD_HTML_Tag_Processor {
 	 */
 	public function remove_attribute( string $name ): bool {
 		return $this->processor->remove_attribute( $name );
+	}
+
+	/**
+	 * Append HTML to the HEAD.
+	 *
+	 * @param string $html HTML to inject.
+	 * @return bool Whether successful.
+	 */
+	public function append_head_html( string $html ): bool {
+		// @phpstan-ignore-next-line
+		$success = $this->processor->append_html( self::END_OF_HEAD_BOOKMARK, $html );
+		if ( ! $success ) {
+			$this->warn( __( 'Unable to append markup to the HEAD.', 'optimization-detective' ) );
+		}
+		return $success;
+	}
+
+	/**
+	 * Append HTML to the BODY.
+	 *
+	 * @param string $html HTML to inject.
+	 * @return bool Whether successful.
+	 */
+	public function append_body_html( string $html ): bool {
+		// @phpstan-ignore-next-line
+		$success = $this->processor->append_html( self::END_OF_BODY_BOOKMARK, $html );
+		if ( ! $success ) {
+			$this->warn( __( 'Unable to append markup to the BODY.', 'optimization-detective' ) );
+		}
+		return $success;
 	}
 
 	/**
--- a/plugins/optimization-detective/class-od-html-tag-processor.php
+++ b/plugins/optimization-detective/class-od-html-tag-processor.php
@@ -180,7 +180,7 @@ final class OD_HTML_Tag_Processor {
 	 */
 	public function __construct( string $html ) {
 
-		$processor = new class( $html ) extends WP_HTML_Tag_Processor {
+		$this->processor = new class( $html ) extends WP_HTML_Tag_Processor {
 
 			/**
 			 * Appends HTML to the provided bookmark.
@@ -202,8 +202,6 @@ final class OD_HTML_Tag_Processor {
 				return true;
 			}
 		};
-
-		$this->processor = new $processor( $html );
 	}
 
 	/**
--- a/plugins/optimization-detective/class-od-html-tag-walker.php
+++ b/plugins/optimization-detective/class-od-html-tag-walker.php
@@ -1,9 +1,9 @@
 <?php
 /**
- * Optimization Detective: OD_HTML_Tag_Processor class
+ * Optimization Detective: OD_HTML_Tag_Walker class
  *
  * @package optimization-detective
- * @since 0.1.0
+ * @since 0.1.1
  */
 
 // Exit if accessed directly.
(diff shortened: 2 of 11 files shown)


plugins/optimization-detective/optimization.php
```diff
@@ -10,6 +10,8 @@
 	exit; // Exit if accessed directly.
 }
 
+const OD_DETECTION_SCRIPT_PLACEHOLDER_COMMENT = '<!--OPTIMIZATION_DETECTIVE_DETECTION_SCRIPT-->';
```
westonruter: This is probably safer, since another optimization plugin may strip out comments.

```suggestion
const OD_DETECTION_SCRIPT_PLACEHOLDER_COMMENT = '{{{OPTIMIZATION_DETECTIVE_DETECTION_SCRIPT}}}';
```

felixarntz (commented):
@westonruter One thing I don't understand (maybe I'm not understanding your explanation fully) is why the script is now being injected via replacing a placeholder.

Instead of injecting a placeholder into `wp_print_footer_scripts` and replacing it as part of output buffering, why can't we simply inject the actual script into `wp_print_footer_scripts`?

westonruter:
@felixarntz Right, because we should only inject the detection script if we need it. If detection isn't needed, then the script should be omitted. Whether detection is needed is determined at the beginning of output buffer processing in `od_optimize_template_output_buffer()`:

https://github.com/WordPress/performance/blob/1b4485fc43630ece10fef6dc130780ae1b865ac4/plugins/optimization-detective/optimization.php#L186-L196

If the script injection were to be moved outside of the output buffer callback, it would look something like this:

```php
add_action(
	'wp_print_footer_scripts',
	static function () {
		$slug = od_get_url_metrics_slug( od_get_normalized_query_vars() );
		$post = OD_URL_Metrics_Post_Type::get_post( $slug );

		$group_collection = new OD_URL_Metrics_Group_Collection(
			$post ? OD_URL_Metrics_Post_Type::get_url_metrics_from_post( $post ) : array(),
			od_get_breakpoint_max_widths(),
			od_get_url_metrics_breakpoint_sample_size(),
			od_get_url_metric_freshness_ttl()
		);
		
		if ( ! $group_collection->is_every_group_complete() ) {
			echo od_get_detection_script( $slug, $group_collection );
		}
	}
);
```

Nevertheless, as is, this would entail getting the same data twice. But assuming that is resolved with a class refactor to remember that state... there is another downside to doing this, and that is mentioned in this todo:

https://github.com/WordPress/performance/blob/1b4485fc43630ece10fef6dc130780ae1b865ac4/plugins/optimization-detective/optimization.php#L344-L348

During optimization, it may be that we find that XPaths in the URL Metrics don't match against something in the document. When this happens, it really should be forcing `$needs_detection` to be true so that we can start collecting data. More would be involved there, as we'd have to have a way to make sure the endpoint allows receiving a new URL Metric even though the groups are complete, but that was in part why I injected the detection script as part of the output buffer processing. Also, the preload links can only be injected during output buffer processing as well.

westonruter:
Hold on. In talking with Dennis apparently there _is_ a way to inject tags with the tag processor.

westonruter:
@dmsnell It seems like the tag insertion in WP 6.4 isn't working. It's resulting in a document like this:

```html
<html lang="en">
        <head>
                <meta charset="utf-8">
                <title>...</title>
        </head>
        <body>
                <img data-od-xpath="/*[0][self::HTML]/*[1][self::BODY]/*[0][self::IMG]" src="https://example.com/foo.jpg" alt="Foo" width="1200" height="800" loading="lazy">
        <script type="module">/* import detect ... */</script>

<html lang="en">
        <head>
                <meta charset="utf-8">
                <title>...</title>
        </head>
        <body>
                <img data-od-xpath="/*[0][self::HTML]/*[1][self::BODY]/*[0][self::IMG]" src="https://example.com/foo.jpg" alt="Foo" width="1200" height="800" loading="lazy">
        </body>
</html>
```

When it should be resulting in the following, which it does in WP 6.5:

```html
<html lang="en">
    <head>
        <meta charset="utf-8">
        <title>...</title>
    </head>
    <body>
        <img data-od-xpath="/*[0][self::HTML]/*[1][self::BODY]/*[0][self::IMG]" src="https://example.com/foo.jpg" alt="Foo" width="1200" height="800" loading="lazy">
        <script type="module">/* import detect ... */</script>
    </body>
</html>
```

plugins/optimization-detective/class-od-html-tag-processor.php
```diff
@@ -160,7 +179,31 @@ final class OD_HTML_Tag_Processor {
...
+				$this->lexical_updates[] = new WP_HTML_Text_Replacement(
+					$this->bookmarks[ $bookmark ]->start,
+					0,
+					$html
+				);
+				return true;
+			}
+		};
+
+		$this->processor = new $processor( $html );
```
dmsnell: when creating the anonymous class, isn't it already instantiated? I don't think we want this second `new`

plugins/optimization-detective/class-od-html-tag-processor.php
```diff
@@ -160,7 +179,31 @@ final class OD_HTML_Tag_Processor {
...
+			 */
+			public function append_html( string $bookmark, string $html ): bool {
+				if ( ! $this->has_bookmark( $bookmark ) ) {
+					return false;
+				}
+
+				$this->lexical_updates[] = new WP_HTML_Text_Replacement(
+					$this->bookmarks[ $bookmark ]->start,
+					0,
+					$html
```
dmsnell: oh shoot, this may be a result of the change of signature of the span/range functions from `(start, end)` to `(start, length)`.

on WordPress 6.4 the span would be from `$this->bookmarks[ $bookmark ]->start` to `$this->bookmarks[ $bookmark ]->end` I think. I wouldn't understand why a `0` value here though would duplicate the string.

also it's suspiciously close to what would happen if we weren't to shadow the `$html` from line 183, but that seems impossible.

dmsnell (commented):
in WordPress 6.5 we did change the span representation and maybe that's in play here, since we're creating the private internal classes. if this code needs to work with both versions then it would likely have to detect that.

alternatively you could also try another approach that I've done, which is to trap externally the edits you want to make. that would ensure you don't have to worry about versioning.

so for example, once you get the head-end and body-end bookmarks, hold on to them until the end and then use them to perform your text operations.

now at the end you can perform your script and link injection in one fell swoop and not worry about the internal `WP_HTML_Text_Replacement` class changes.

```php
new class ( $html ) extends WP_HTML_Tag_Processor {
	public function split_at_start_of_bookmarks( ...$names ) {
		$splits = array();
		$at     = 0;

		foreach ( $names as $name ) {
			$mark     = $this->bookmarks[ $name ];
			$splits[] = substr( $this->html, $at, $mark->start );
			$at       = $mark->start;
		}

		if ( $at < strlen( $this->html ) ) {
			$splits[] = substr( $this->html, $at );
		}

		return $splits;
	}
}

$links    = array();
$scripts  = array();

while ( $processor->next_tag( … ) ) {
	if ( $need_a_link ) {
		$links[] = "<link rel=anything>";
	}

	…

	if ( ! $found_head && 'HEAD' === $tag_name && $processor->is_tag_closer() ) {
		$processor->set_bookmark( HEAD_BOOKMARK );
	}

	…
}

// Processing is done, inject content.
list( $head, $body, $footer ) = $processor->split_at_start_of_bookmarks( HEAD_MARK, BODY_MARK );
return $head . implode( "\n", $links ) . $body . implode( "\n", $scripts ) . $footer;
```

again, this is only one way of doing it to avoid trying to detect the running version of WordPress. it's probably easiest to perform that version check and adjust whether you add `0` or the start of the bookmark again (length vs. ending location).

plugins/optimization-detective/class-od-html-tag-processor.php
westonruter: Oh, right. Fixed in 7a6debcb9ee097dcf4e9f536b111fcf4644d326a

plugins/optimization-detective/class-od-html-tag-processor.php
westonruter: Great! Easily fixed then in 03ad8f9f30311d55907addd7b8e6558cf33c48b3.

plugins/optimization-detective/class-od-html-tag-processor.php
```diff
@@ -160,7 +179,32 @@ final class OD_HTML_Tag_Processor {
 	 * @param string $html HTML to process.
 	 */
 	public function __construct( string $html ) {
-		$this->processor = new WP_HTML_Tag_Processor( $html );
+
+		$this->processor = new class( $html ) extends WP_HTML_Tag_Processor {
```
felixarntz: This seems a bit fragile to me. Can we make it an actual named class?

felixarntz (commented):
@westonruter I don't feel like I'm the right person to review this, would be great to get @dmsnell's approval as he's far more familiar with the internals of the HTML tag processor.

I only have one high level feedback, but overall it looks good to me.

plugins/optimization-detective/class-od-html-tag-processor.php
westonruter: Fragile how? It's what @dmsnell suggested so I follow his lead for the use of the anonymous class.

westonruter: I've invited @dmsnell to the Performance team on GitHub for proper approval granting 😄

tests/plugins/optimization-detective/optimization-tests.php
```diff
@@ -469,11 +469,11 @@ public function data_provider_test_od_optimize_template_output_buffer(): array {
...
 							<meta charset="utf-8">
 							<title>...</title>
-							<link as="image" data-od-added-tag="" fetchpriority="high" href="https://example.com/foo.jpg" rel="preload" media="screen">
+							<link data-od-added-tag rel="preload" fetchpriority="high" as="image" href="https://example.com/foo.jpg" media="screen">
 						</head>
 						<body>
-							<img src="https://example.com/foo.jpg" alt="Foo" width="1200" height="800" fetchpriority="high" data-od-added-fetchpriority data-od-removed-loading="lazy">
-							<img src="https://example.com/bar.jpg" alt="Bar" width="10" height="10" loading="lazy" data-od-removed-fetchpriority="high">
+							<img data-od-added-fetchpriority data-od-removed-loading="lazy" fetchpriority="high" src="https://example.com/foo.jpg" alt="Foo" width="1200" height="800" >
+							<img data-od-removed-fetchpriority="high" src="https://example.com/bar.jpg" alt="Bar" width="10" height="10" loading="lazy" >
```
westonruter, line 476: Note: these changes are purely syntactic and not semantic. The equivalent DOM tree is produced. The difference is made here because no longer is DOMDocument being used for comparison in PHPUnit. Using DOMDocument normalized differences in attribute order. But with the removal of DOMDocument, the attribute order and value syntax has to match.

The same goes for the other changes below, other than the move of the `script` from the `head` to the footer, which is the only semantic change.

plugins/optimization-detective/class-od-html-tag-processor.php
dmsnell: I'm not sure about what kind of fragility we're concerned about here. Extending `WP_HTML_Tag_Processor` provides guarantees about the behavior of the underlying methods.

The anonymous class was more or less an example I tossed out as a convenient way to create or expose risky behaviors from the HTML API without exporting that to the outside world. It creates these methods but they shouldn't be instantiable or callable from outside, and won't show up in documentation. Also they are convenient if your styling rules forbid having more than one class in the same file.

plugins/optimization-detective/class-od-html-tag-processor.php
```diff
@@ -160,7 +179,32 @@ final class OD_HTML_Tag_Processor {
...
+				if ( ! $this->has_bookmark( $bookmark ) ) {
+					return false;
+				}
+
+				$start = $this->bookmarks[ $bookmark ]->start;
+
+				$this->lexical_updates[] = new WP_HTML_Text_Replacement(
+					$start,
+					// In WordPress 6.5, the signature was changed from $end to $length.
+					version_compare( get_bloginfo( 'version' ), '6.5', '<' ) ? $start : 0,
```
dmsnell: this probably doesn't matter much, but there's no need to recompute this on every call. the WordPress version won't change during the runtime (at least outside of the Playground where it could), so this could be set in the constructor or as a method-level static var.

dmsnell (approved):
I don't see any obvious issues with the HTML API use. Thanks for continuing to lean into it and explore its boundaries.

plugins/optimization-detective/class-od-html-tag-processor.php
felixarntz: I'm just questioning why we use an anonymous class here. We haven't done that anywhere, even with other classes we consider private. So it's a new pattern and I'm not sure we want to go that route, as it would lead to more such cases in the future.

As @dmsnell says, the anonymous class was just "a convenient way". Changing it to a regular class will probably take just a few minutes, but follow our patterns established so far.

I just don't see any good reason for an anonymous class, other than it was very quick to throw in here and we don't have to think about naming. We can just mark it as `@access private` like we do with other classes, then it doesn't matter if someone else tries to use it - that's on them, and there's no support or guarantees for it.

plugins/optimization-detective/class-od-html-tag-processor.php
westonruter: Good point. Although this will only be called twice at most, once for the head insertion and once for body. I'll add a static var.

plugins/optimization-detective/class-od-html-tag-processor.php
westonruter: It makes sense to me to use an anonymous class here since this is the only place where this class is being used. Making it a regular class would mean yet another class in the global namespace that isn't being used anywhere except this one place.

plugins/optimization-detective/class-od-html-tag-processor.php
westonruter: Done in b3fbf5c2

plugins/optimization-detective/class-od-html-tag-processor.php
dmsnell: fun fact I learned recently: anonymous classes _aren't_ anonymous 😄

_thankfully_, they are compiled up-front, given a unique name, and behave like regular classes. if you return one from a function then all the returned values will be of the same "anonymous" class.

```php
function make() {
	return new class () {
		public function me() {
			echo "I am no sweet unique butterfly 😭";
		}
	}
}

$a = make();
$b = make();
true === ( get_class( $a ) === get_class( $b ) );
true === ( $a instanceof $b );
true === ( $b instanceof $a );
```

this is unrelated to the current discussion, but I found that interesting when I learned.

felixarntz: @westonruter 
> It makes sense to me to use an anonymous class here since this is the only place where this class is being used. Making it a regular class would mean yet another class in the global namespace that isn't being used anywhere except this one place.

IMO that's not a good argument against using a regular class. For example, the `OD_HTML_Tag_Processor` class is also only instantiated and used in one place in the plugin code, so I don't see the difference. Also, as mentioned before, it's a new pattern and I'm not sure it's one we want to encourage.

westonruter: @felixarntz The other thing is that the use of `WP_HTML_Tag_Processor` should eventually be replaced with `WP_HTML_Processor` once it implements support for the remaining tags. This class will eventually support node insertion itself, so what is currently an anonymous class here should eventually be removed. IMO just because we haven't used a pattern before (here because it's a new language feature in PHP 7) isn't a good argument against using it. But if you feel strongly I'll do the work to move it to a regular class.

felixarntz: I'd prefer to do it in a class. Once we no longer need it, we can deprecate it, or even remove if marked private.

westonruter: I'd prefer not 😄 

I guess you feel strongly then. I'll do it.

westonruter: OK, this is done.

felixarntz: Thanks @westonruter, I think this is in line with how we've been writing classes so far. FWIW there are no anonymous classes in WP core either, and since this is a feature plugin I think it makes sense to follow core's patterns too.

plugins/optimization-detective/class-od-html-tag-walker.php
```diff
@@ -0,0 +1,443 @@
...
+/**
+ * Walker leveraging WP_HTML_Tag_Processor which gathers breadcrumbs for computing XPaths while iterating the open_tags() generator.
+ *
+ * Eventually this class should be made largely obsolete once `WP_HTML_Processor` is fully implemented to support all HTML tags.
+ *
+ * @since 0.1.0
+ * @since 0.1.1 Renamed from OD_HTML_Tag_Processor to OD_HTML_Tag_Walker
+ * @access private
+ */
+final class OD_HTML_Tag_Walker {
```
felixarntz, line 23: Just double checking as this is hard to review: I _think_ you renamed the class that was originally wrapping `WP_HTML_Tag_Processor` from `OD_HTML_Tag_Processor` to this, and the new `OD_HTML_Tag_Processor` is the new extending class for `WP_HTML_Tag_Processor`?

That would make sense to me. I assume there are no further notable changes in this class here that haven't been reviewed?

felixarntz (approved):
This LGTM, just one question to make sure we didn't miss it.

plugins/optimization-detective/class-od-html-tag-walker.php
westonruter: Yes, I noted that in the `@since` tag and I also updated the PR description to explain this change.

plugins/optimization-detective/class-od-html-tag-processor.php
westonruter: Right, but this was intended to anticipate `WP_HTML_Processor` having such node insertion functionality. By the time this is proposed for core I was expecting the anonymous class would no longer be needed. Anyway, let's move on 😄

dmsnell: might be good to consider anonymous classes at times when we have something dangerous like this that we don't want to expose. it's unfortunate we have no way of hiding things otherwise.

PR #1089, opened 2024-03-25
Fix lint-staged running phpcs checks for standalone plugins

westonruter:
I found that lint-staged was completely skipping phpcs checks for PHP files in standalone plugins. This is because it was attempting to run `composer run lint-staged` on files in the `plugins` directory, which is excluded by the `phpcs.xml.dist` config. This fixes the problem by adding plugin-specific patterns for lint-staged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,8 +51,34 @@
   },
   "lint-staged": {
     "*.php": [
-      "composer run-script lint",
       "composer run-script phpstan"
+    ],
+    "./*.php": [
+      "composer run-script lint"
+    ],
+    "./tests/*.php": [
+      "composer run-script lint"
+    ],
+    "./{includes,tests/includes}/**/*.php": [
+      "composer run-script lint"
+    ],
+    "./{plugins/auto-sizes,tests/plugins/auto-sizes}/**/*.php": [
+      "composer run-script lint:auto-sizes"
+    ],
+    "./{plugins/dominant-color-images,tests/plugins/dominant-color-images}/**/*.php": [
+      "composer run-script lint:dominant-color-images"
+    ],
+    "./{plugins/embed-optimizer,tests/plugins/embed-optimizer}/**/*.php": [
+      "composer run-script lint:embed-optimizer"
+    ],
+    "./{plugins/optimization-detective,tests/plugins/optimization-detective}/**/*.php": [
+      "composer run-script lint:optimization-detective"
+    ],
+    "./{plugins/speculation-rules,tests/plugins/speculation-rules}/**/*.php": [
+      "composer run-script lint:speculation-rules"
+    ],
+    "./{plugins/webp-uploads,tests/plugins/webp-uploads}/**/*.php": [
+      "composer run-script lint:webp-uploads"
     ],
     "*.js": [
       "npm run lint-js"
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "./tests/*.php": [
       "composer run-script lint"
     ],
-    "./{includes,tests/includes}/**/*.php": [
+    "./{includes,tests/includes,tests/testdata,tests/utils}/**/*.php": [
       "composer run-script lint"
     ],
     "./{plugins/auto-sizes,tests/plugins/auto-sizes}/**/*.php": [


package.json
```diff
@@ -51,9 +51,35 @@
   },
   "lint-staged": {
```
thelovekesh, line 52: @westonruter Since there are multiple entries for `lint-staged`, should we switch to `.lintstagedrc.json`? 

Also, if we add `.lintstagedrc.json` in each plugin, then it will auto-resolve the staged files in that particular directory. For the experiment I put `plugins/webp-uploads/.lintstagedrc.json` and when I made changes to the files in `plugins/webp-uploads` it automatically ran on staged files within `plugins/webp-uploads`.

// path: `plugins/webp-uploads/.lintstagedrc.json`
```json
{
  "*.php": [
    "composer run-script lint --working-dir=../../",
    "composer run-script phpstan --working-dir=../../"
  ],
  "*.js": [
    "npm run lint-js"
  ]
}
```

output:
```sh
➜  performance git:(trunk) ✗ git commit -m "Temp commit"   
✔ Preparing lint-staged...
✔ Preparing lint-staged...
❯ Running tasks for staged files...
  ❯ plugins/webp-uploads/.lintstagedrc.json — 1 file
    ❯ *.php — 1 file
      ✔ composer run-script lint --working-dir=../../
      ✖ composer run-script phpstan --working-dir=../../
    ↓ *.js — no files [SKIPPED]
◼ Applying modifications from tasks...
◼ Cleaning up temporary files...
``

Comment (westonruter) [reply package.json:52]: Hummm, this doesn't seem to work. I tried putting that `.lintstagedrc.json` at `plugins/webp-uploads`. I then added an intentional PHPCS failure, like putting the wrong text domain in a translation function. It didn't block the commit. I think this is because `plugins` is still excluded in the root `phpcs.xml.dist`. No PHPCS errors were reported. The PHPStan checks are running successfully, however.

Comment (westonruter) [reply package.json:52]: To get it to work, I have to explicitly run `lint:webp-uploads` and not just `lint`:

```json
{
  "*.php": [
    "composer run-script lint:webp-uploads --working-dir=../../",
    "composer run-script phpstan --working-dir=../../"
  ],
  "*.js": [
    "npm run lint-js"
  ]
}
```

However, I'm not sure this is better than having a single root config because only the lint command is plugin-specific currently. Having plugin-specific `.lintstagedrc.json` files would introduce more duplication.

package.json
```diff
@@ -51,9 +51,35 @@
...
+    "./{plugins/embed-optimizer,tests/plugins/embed-optimizer}/**/*.php": [
+      "composer run-script lint:embed-optimizer"
+    ],
+    "./{plugins/optimization-detective,tests/plugins/optimization-detective}/**/*.php": [
+      "composer run-script lint:optimization-detective"
+    ],
+    "./{plugins/speculation-rules,tests/plugins/speculation-rules}/**/*.php": [
+      "composer run-script lint:speculation-rules"
+    ],
+    "./{plugins/webp-uploads,tests/plugins/webp-uploads}/**/*.php": [
```
westonruter, line 80: These can be simplified once #1065 is merged.

```suggestion
    "./plugins/auto-sizes/**/*.php": [
      "composer run-script lint:auto-sizes"
    ],
    "./plugins/dominant-color-images/**/*.php": [
      "composer run-script lint:dominant-color-images"
    ],
    "./plugins/embed-optimizer/**/*.php": [
      "composer run-script lint:embed-optimizer"
    ],
    "./plugins/optimization-detective/**/*.php": [
      "composer run-script lint:optimization-detective"
    ],
    "./plugins/speculation-rules/**/*.php": [
      "composer run-script lint:speculation-rules"
    ],
    "./plugins/webp-uploads/**/*.php": [
```

package.json
thelovekesh: > However, I'm not sure this is better than having a single root config because only the lint command is plugin-specific currently. Having plugin-specific .lintstagedrc.json files would introduce more duplication.

Humm and it's a long-known trade-off in mono repos. It will bring duplication but lead to simpler management on plugin level but we can have a root level `.lintstagedrc.json` for now.

thelovekesh: Did some experiments on https://github.com/WordPress/performance/pull/1090 with dynamic lint-staged config. 
@westonruter Can you please test it?

westonruter:
Closing in favor of #1090

PR #1090, opened 2024-03-26
Add PHP `lint` scripts in npm; Fix `lint-staged` config

thelovekesh:
## Summary

As mentioned by @westonruter in https://github.com/WordPress/performance/issues/1012#issuecomment-2018519984, for standalone plugins, PHPCS always takes `phpcs.xml.dist` into account since it is explicitly set to phpcs.xml.dist:

https://github.com/WordPress/performance/blob/5e570fae64d45079c6ab42c6a48d7b9554138014/composer.json#L39

This PR aims to move `lint` and `format` scripts from `composer.json` to `package.json` scripts to fix the above behavior and running commands from any directory will be easy with `npm run`.

Also it adds the `lint-staged` config to run the correct linting commands on staged files based on the correct PHPCS config.

Closes #1089

## Relevant technical choices

- Use the https://docs.npmjs.com/cli/v10/using-npm/config#command-line-flags feature to `cd` into a specific plugin directory and let PHPCS decide which specific config file to use.


lint-staged.config.js
```diff
@@ -0,0 +1,44 @@
...
+const micromatch = require( 'micromatch' );
+
+/**
+ * Internal dependencies
+ */
+const { plugins } = require( './plugins.json' );
+
+module.exports = {
+	'**/*.js': ( files ) =>
+		`npm run lint-js ${ files.length > 10 ? '' : files.join( ' ' ) }`,
```
westonruter: Why the `files.length > 10` checks?

lint-staged.config.js
```diff
@@ -0,0 +1,44 @@
...
+		plugins.forEach( ( plugin ) => {
+			const pluginFiles = micromatch(
+				files,
+				`**/plugins/${ plugin }/**`
+			);
+
+			if ( pluginFiles.length ) {
+				commands.push(
+					`npm run lint:php:${ plugin } ${
+						pluginFiles.length > 10 ? '' : pluginFiles.join( ' ' )
```
westonruter: Rather than just joining with a space, there should probably should be some shell argument escaping just in case (however unlikely) a file has a space in it or for more nefarious reasons.

lint-staged.config.js
```diff
@@ -0,0 +1,44 @@
...
+module.exports = {
+	'**/*.js': ( files ) =>
+		`npm run lint-js ${ files.length > 10 ? '' : files.join( ' ' ) }`,
+	'**/*.php': ( files ) => {
+		const commands = [ 'composer phpstan' ];
+
+		plugins.forEach( ( plugin ) => {
+			const pluginFiles = micromatch(
+				files,
+				`**/plugins/${ plugin }/**`
```
westonruter: Naturally this would need to wait for #1065 to be merged first so that plugin-specific PHPCS rules are checked for test files.

lint-staged.config.js
thelovekesh: It's just to run linting on whole codebase in case there are multiple files which are staged.

lint-staged.config.js
thelovekesh: I think path of staged files are resolved and passed as string. Do wr still need to think about this?

lint-staged.config.js
thelovekesh: Currently all the tests files adheres to root PHPCS config. Once they are moved to respective plugin directories, they will use plugin specific PHPCS config. In any case this file will remain same so #1065 isn't a blocker for this.

lint-staged.config.js
westonruter: Ultimately, what should be getting constructed here is:

```bash
npm run lint:php:auto-sizes 'foo.js' 'bar.js' 'baz quux.js'
```

So the filenames should be getting quoted. Alternatively, can an array be passed as a command? So this could instead be:

```js
[ 'npm', 'run', `lint:php:${ plugin }`, ...pluginFiles ]
```

That may handle the escaping automatically.

Doing a quick search for a built-in Node equivalent to `escapeshellarg()` is coming up blank.

lint-staged.config.js
westonruter: If modifying more than 10 files though, perhaps you don't want to check _all_ the files? Why not just pass the files always?

lint-staged.config.js
westonruter: I don't think that's the case. For example, webp-uploads has a plugin-specific translation string:

https://github.com/WordPress/performance/blob/2bc2f96d4061cf0201433eb64e1c22a741bc01e9/tests/plugins/webp-uploads/helper-tests.php#L345

thelovekesh: Ahh, interesting. At this moment neither of the PHPCS configs are linting the `tests/plugins` directory as the `plugins` directory is excluded in the root PHPCS config. So they will be linted once #1065 is merged but no changes will be required in lint staged config. So we can merge it right?

lint-staged.config.js
thelovekesh: LInt staged receives an array of commands and not its parts IMO. But I have updated the logic of joining files:

```js
const joinFiles = ( files ) => {
	return files.map( ( file ) => `'${ file }'` ).join( ' ' );
};
```
and it's working as expected.

lint-staged.config.js
thelovekesh: My initial thought was to run linting on the whole codebase if changes are big. In big changes, sometimes a few things are missed. Let's say you removed some dependencies from `package.json` but they are still in some unstaged files, ESlint has a rule to check that but such things can also be caught on bundling, CI, or runtime.

Though I am not fully sure if we need to keep this but IMP if changes are big it makes sense to run linting on the whole codebase.

westonruter: I'd say to remove the `files.length` check. The 10 seems a bit arbitrary. In any case, the GHA workflow will check against the entire codebase anyway.

lint-staged.config.js
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @return {string} Joined files.
+ */
+const joinFiles = ( files ) => {
+	return files.map( ( file ) => `'${ file }'` ).join( ' ' );
+};
+
+module.exports = {
+	'**/*.js': ( files ) =>
+		`npm run lint-js ${ files.length > 10 ? '' : joinFiles( files ) }`,
```
westonruter:
```suggestion
		`npm run lint-js ${ joinFiles( files ) }`,
```

lint-staged.config.js
```diff
@@ -0,0 +1,55 @@
...
+		plugins.forEach( ( plugin ) => {
+			const pluginFiles = micromatch(
+				files,
+				`**/plugins/${ plugin }/**`
+			);
+
+			if ( pluginFiles.length ) {
+				commands.push(
+					`npm run lint:php:${ plugin } ${
+						pluginFiles.length > 10 ? '' : joinFiles( pluginFiles )
```
westonruter:
```suggestion
						joinFiles( pluginFiles )
```

lint-staged.config.js
```diff
@@ -0,0 +1,55 @@
...
+				);
+			}
+		} );
+
+		const otherFiles = micromatch( files, `!plugins/**` );
+
+		if ( otherFiles.length ) {
+			commands.push(
+				`npm run lint:php ${
+					otherFiles.length > 10 ? '' : joinFiles( otherFiles )
```
westonruter:
```suggestion
					joinFiles( otherFiles )
```

lint-staged.config.js
thelovekesh: Humm makes sense. Reverting it.

.github/workflows/php-lint.yml
```diff
@@ -52,7 +52,9 @@ jobs:
         run: composer validate --strict
       - name: Composer Install
         run: composer install --no-interaction --no-progress
+      - name: Install `npm-run-all` utility
+        run: npm install npm-run-all -g
```
westonruter: Why would this need to be installed globally if it's already installed locally?

lint-staged.config.js
```diff
@@ -0,0 +1,48 @@
...
+
+module.exports = {
+	'**/*.js': ( files ) => `npm run lint-js ${ joinFiles( files ) }`,
+	'**/*.php': ( files ) => {
+		const commands = [ 'composer phpstan' ];
+
+		plugins.forEach( ( plugin ) => {
+			const pluginFiles = micromatch(
+				files,
+				`**/plugins/${ plugin }/**`
```
westonruter: PhpStorm is complaining here: "Invalid number of arguments, expected 3"

```suggestion
				`**/plugins/${ plugin }/**`,
				{}
```

It seems micromatch isn't properly doing its jsdoc to make the `options` arg optional.

lint-staged.config.js
```diff
@@ -0,0 +1,48 @@
...
+			);
+
+			if ( pluginFiles.length ) {
+				commands.push(
+					`npm run lint:php:${ plugin } ${ joinFiles( pluginFiles ) }`
+				);
+			}
+		} );
+
+		const otherFiles = micromatch( files, `!plugins/**` );
```
westonruter:
```suggestion
		const otherFiles = micromatch( files, `!plugins/**`, {} );
```

lint-staged.config.js
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Internal dependencies
+ */
+const { plugins } = require( './plugins.json' );
+
+/**
+ * Join and escape filenames for shell.
+ *
+ * @param {Array} files Files to join.
```
westonruter:
```suggestion
 * @param {string[]} files Files to join.
```

lint-staged.config.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @return {string} Joined files.
+ */
+const joinFiles = ( files ) => {
+	return files.map( ( file ) => `'${ file }'` ).join( ' ' );
+};
+
+module.exports = {
+	'**/*.js': ( files ) => `npm run lint-js ${ joinFiles( files ) }`,
```
westonruter, line 28: Circling back to our previous chat, I was hoping that lint-staged would supported what node does with passing args to commands. For [example](https://nodejs.org/api/child_process.html#child_process_asynchronous_process_creation):

```js
const { spawn } = require('node:child_process');
const ls = spawn('ls', ['-lh', '/usr']);
```

Each argument is passed in an array. But alas it seems lint-staged doesn't support this. I see it has been opened as a feature request: https://github.com/lint-staged/lint-staged/issues/989.

lint-staged.config.js
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Join and escape filenames for shell.
+ *
+ * @param {Array} files Files to join.
+ *
+ * @return {string} Joined files.
+ */
+const joinFiles = ( files ) => {
+	return files.map( ( file ) => `'${ file }'` ).join( ' ' );
```
westonruter, line 20: Oh, I see in https://github.com/lint-staged/lint-staged/pull/875 that an `escape()` function was added. So this should rather be:

```suggestion
	return files.map( escape ).join( ' ' );
```

Along with the `require` above.

lint-staged.config.js
```diff
@@ -0,0 +1,48 @@
+/**
+ * External dependencies
+ */
+const micromatch = require( 'micromatch' );
```
westonruter, line 5:
```suggestion
const micromatch = require( 'micromatch' );
const { escape } = require( 'lint-staged/utils' )
```

westonruter (changes requested):

westonruter:
Oh, and `lint-staged.config.js` needs to be added to `.gitattributes`.

.github/workflows/php-lint.yml
thelovekesh: We aren't installing NPM deps in the PHP Lint workflow, and installing this utility globally took ~2 seconds while `npm install` takes ~20 seconds.

lint-staged.config.js
thelovekesh: The above PR was not merged. And the paths that come in the files array are already normalized. I am not sure if we need to rethink this.

lint-staged.config.js
thelovekesh: Done as a part of https://github.com/WordPress/performance/pull/1090/commits/52d07aef7d9d94a9378d91e10db4368d1e62509f.

lint-staged.config.js
westonruter: Oops, I missed that it wasn't merged.

I tried staging 4 files with irregular filenames:

* foo\bar.php
* food bard.php
* food"bard.php
* food'bard.php

When I then tried to commit, I see that they are passed to a command as follows:

```bash
npm run lint:php \
'/home/westonruter/repos/performance/foo/bar.php' \
'/home/westonruter/repos/performance/food bard.php' \
'/home/westonruter/repos/performance/food"bard.php' \
'/home/westonruter/repos/performance/food'bard.php'
```

So interestingly, it normalizes the backslash to a forward slash which makes the file not found. In the second case, space is accounted for by the quotes. In the last example, the single quote is an error because it is not escaped.

Granted, it is highly unlikely for any of these scenarios to be encountered. But to guard against them, additional escaping would be needed since it doesn't seem the normalization in lint-staged is doing any escaping.

thelovekesh: Not sure how to implement it correctly to support it on both Posix and Win. Even if I add a workaround to handle the above cases, there might be other cases that can break things on different OS.

IMO we should keep it as it is and wait to get it fixed upstream.  Also, if such a file exists and `lint-staged` fails, the user can learn about it and either try to fix the file name or pass a `--no-verify`.

_Found an issue here where it was discussed - https://github.com/nodejs/node/issues/34840._

westonruter (approved):

lint-staged.config.js
thelovekesh: > So interestingly, it normalizes the backslash to a forward slash which makes the file not found.

`lint-staged` does the normalizing here. Should this be reported upstream? But I think the author can rule it out by mentioning it's an irregular file name.

westonruter: Yeah, it's highly irregular. Doubtful anyone will have an issue there.

composer.json
```diff
@@ -25,23 +25,6 @@
...
-      "@lint:optimization-detective",
-      "@lint:speculation-rules",
-      "@lint:webp-uploads"
-    ],
-    "lint:auto-sizes": "@lint -- ./plugins/auto-sizes --standard=./plugins/auto-sizes/phpcs.xml.dist",
-    "lint:dominant-color-images": "@lint -- ./plugins/dominant-color-images --standard=./plugins/dominant-color-images/phpcs.xml.dist",
-    "lint:embed-optimizer": "@lint -- ./plugins/embed-optimizer --standard=./plugins/embed-optimizer/phpcs.xml.dist",
-    "lint:optimization-detective": "@lint -- ./plugins/optimization-detective --standard=./plugins/optimization-detective/phpcs.xml.dist",
-    "lint:speculation-rules": "@lint -- ./plugins/speculation-rules --standard=./plugins/speculation-rules/phpcs.xml.dist",
-    "lint:webp-uploads": "@lint -- ./plugins/webp-uploads --standard=./plugins/webp-uploads/phpcs.xml.dist",
```
felixarntz: Why are all of these being removed from here? Similar to @joemcgill's feedback in https://github.com/WordPress/performance/pull/1065#discussion_r1546419716, I think PHP based tooling makes sense to be accessible as a Composer script.

Can't this be solved in a PHP way? What's the benefit of moving all of these to NPM?

At a minimum, I think they should remain accessible via `composer.json` like before, while NPM could effectively be a "wrapper".

felixarntz (changes requested):
@thelovekesh While this PR appears to solve the problem of plugin-specific `phpcs.xml` files not being respected, I'm not sure about the approach. Why are we moving PHP specific logic to require NPM? That seems a bit backwards to me, especially since it still requires writing a custom script to make it work.

Why not solve it with a PHP solution, which IMO would be more adequate for the problem? If we need to dynamically support plugin-specific `phpcs.xml` that a developer may have placed, that should be possible in a much less disruptive way by writing a custom PHP script.

By doing that, we'd keep the PHP linting tied to PHP (with NPM merely being a wrapper if anything), and this PR would contain fewer changes.

composer.json
thelovekesh: @felixarntz I have added https://github.com/WordPress/performance/pull/1065#issuecomment-2034000560 in response to https://github.com/WordPress/performance/pull/1065#discussion_r1546419716.

> At a minimum, I think they should remain accessible via composer.json like before, while NPM could effectively be a "wrapper".

Wouldn't that cause duplication and maintenance burden? Also, I would like to know why we want it on composer specifically while NPM has the same feature, a better build to handle mono-repo use cases.

thelovekesh:
> Why not solve it with a PHP solution, which IMO would be more adequate for the problem? If we need to dynamically support plugin-specific phpcs.xml that a developer may have placed, that should be possible in a much less disruptive way by writing a custom PHP script.

@felixarntz Why do we need a custom PHP script while NPM has it out of the box? And even if we keep it in composer, the number of commands would be the same. I can also argue about using PHP since all the CLI-based tasks are being handled by NodeJS CLI in this repo.

composer.json
felixarntz: @thelovekesh There would only be duplication if we have plugin specific scripts in both `composer.json` and `package.json`. What I'm proposing though is something like this:
* We abstract the PHPCS specific call in a custom PHP script that takes the plugin slug as a parameter and takes the potential `phpcs.xml` from the plugin folder into account.
* The `composer.json` can then use that instead of manually calling `phpcs` with the correct parameters.
* No changes to `package.json` would be needed.

PHPCS is PHP specific tooling, and `composer.json` is the appropriate location to manage such PHP specific tooling.

felixarntz:
@thelovekesh 
> Why do we need a custom PHP script while NPM has it out of the box?

My point is that NPM _doesn't_ have it out of the box. If it did, maybe that would be a good reason. But you implemented a custom NPM specific script https://github.com/WordPress/performance/pull/1090/files#diff-29c15f0b5d8d1144bd22153c302cde16c2de092d65b122a618cbdb3023336346 to make it work. At that point, it would be better to solve the problem in PHP and `composer.json`, as the affected tooling is purely for PHP, and NPM requires a custom script as well.

westonruter:
> My point is that NPM _doesn't_ have it out of the box. If it did, maybe that would be a good reason. But you implemented a custom NPM specific script https://github.com/WordPress/performance/pull/1090/files#diff-29c15f0b5d8d1144bd22153c302cde16c2de092d65b122a618cbdb3023336346 to make it work.

@felixarntz Actually, no. This is unrelated. This is exclusively to fix the lint-staged issue I mentioned in https://github.com/WordPress/performance/issues/1012#issuecomment-2018691843. It could be moved into a separate PR.

felixarntz:
@westonruter 
> > My point is that NPM _doesn't_ have it out of the box. If it did, maybe that would be a good reason. But you implemented a custom NPM specific script https://github.com/WordPress/performance/pull/1090/files#diff-29c15f0b5d8d1144bd22153c302cde16c2de092d65b122a618cbdb3023336346 to make it work.
> 
> @felixarntz Actually, no. This is unrelated. This is exclusively to fix the lint-staged issue I mentioned in [#1012 (comment)](https://github.com/WordPress/performance/issues/1012#issuecomment-2018691843). It could be moved into a separate PR.

Ah I see, thank you for clarifying. Looking at the new `package.json` scripts, what's preventing us from leaving that in `composer.json`? Can't we use `cd` there and run PHPCS that same way?

thelovekesh:
@felixarntz there are a few scenarios here:
- The current directory is the root,
- The current directory is `plugins/{some-plugin}`
- Command can handle custom args so we don't need to write our scripts.

By default, the composer works well in the first case. If you run any script from another directory it will ask:
```
No composer.json in current directory, do you want to use the one at /home/thelovekesh/Desktop/work/wpdev/public/content/plugins/performance? [Y,n]? y
Always want to use the parent dir? Use "composer config --global use-parent-dir true" to change the default.
```
but we can change this behavior by setting `use-parent-dir` to true in the global config which I don't think is a good idea.

Coming to the third point, if you pass some custom argument to NPM scripts like `npm run foo --bar=baz`, NPM CLI will convert it to `$npm_config_bar` which can be used to determine the value of `--bar` argument anytime. To achieve this in composer we have to write a custom script. I hope it explains the reason behind choosing NPM over composer.

felixarntz:
Thanks for outlining this @thelovekesh.

What I'm curious about is why we need to support running these commands from a specific `/plugins/{plugin}` directory. Which benefits does developing from a subdirectory like this actually provide? I know @westonruter had mentioned this before somewhere, but I'm unsure why it's useful.

As this is a monorepo, IMO it makes sense to develop from the root folder, just like in any other repo, especially because some tooling is shared, and should be shared to avoid duplication. You probably wouldn't develop or run CI from a specific subdirectory of an individual plugin repository either.

thelovekesh:
Thanks @felixarntz. We organize the tool within directories so it knows which configuration file to use. For example, let's say you're working with 'auto-sizes' and it has its own PHPCS config. If we run it from the main directory, we have to tell it where the config is. But if we run PHPCS from inside the 'auto-sizes' directory, it figures out the config on its own. And if you want to replace the config with a higher priority one like 'phpcs.xml', PHPCS handles that automatically.

In the context of CI, we can run it however we like it since it's automation, and passing the config path is not a big task.

thelovekesh:
> As this is a monorepo, IMO it makes sense to develop from the root folder, just like in any other repo, especially because some tooling is shared, and should be shared to avoid duplication.

I'm uncertain if that's accurate, considering tools like NPM, which offer various features for monorepos out of the box. They might not provide the capability to execute npm scripts from child directories while maintaining the root directory as the context.  

The flexibility and nature of the tool determine whether such configurations are possible, as discussed earlier. While some tools offer this flexibility, others may not.

westonruter:
> As this is a monorepo, IMO it makes sense to develop from the root folder, just like in any other repo, especially because some tooling is shared, and should be shared to avoid duplication. You probably wouldn't develop or run CI from a specific subdirectory of an individual plugin repository either.

@felixarntz If I'm working on a specific plugin, I'm more likely to change into the directory containing that plugin code so I can focus specifically on files and directories for that plugin. This is also why I think it makes sense to move the plugin tests to the plugin's own directory rather than in a shared location. This is what Gutenberg does, for example: https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library

thelovekesh:
Just adding to this, Jetpack also does the same.

.github/workflows/php-lint.yml
thelovekesh: Removed it as part of new enhancements.

composer.json
thelovekesh: @felixarntz Can you please take a look at the new changes? As suggested by @joemcgill, I have added a PHP script to handle linting in plugins and also added an NPM wrapper in composer commands.

package.json
```diff
@@ -39,23 +41,16 @@
     "build:plugin:webp-uploads": "webpack --mode production --env plugin=webp-uploads",
     "format-js": "wp-scripts format",
     "lint-js": "wp-scripts lint-js",
-    "format-php": "composer format",
     "phpstan": "composer phpstan",
-    "lint-php": "composer lint:all",
+    "lint:php": "composer lint",
+    "lint:php:fix": "composer format",
+    "lint:php:plugins": "cross-env WPP_PLUGIN=$npm_config_plugin php -f bin/phpcs/lint.php",
+    "lint:php:plugins:fix": "cross-env WPP_FIX=1 WPP_PLUGIN=$npm_config_plugin php -f bin/phpcs/lint.php",
```
felixarntz, line 50: Why do we need these here? What prevents us from implementing this as a PHP-based command in `composer.json`?

A more specific question, so that I understand correctly: Where does the `$npm_config_plugin` variable come from here?

felixarntz, line 50: Nit-pick: The other `package.json` commands use hyphens (`-`), not colons (`:`), so better to stick with that for consistency. Also I would suggest we stick with the "format" name instead of "lint...fix", which is more intuitive to remember and aligned with the similar JS tooling.

felixarntz (commented):
Thanks @thelovekesh, I think it's going in a better direction now.

Adding some follow up feedback below.

package.json
westonruter: The `$npm_config_plugin` variable comes from the `--plugin` command line arg to `npm`. https://dev.to/felipperegazio/handling-command-line-arguments-in-npm-scripts-2ean

bin/phpcs/lint.php
```diff
@@ -0,0 +1,105 @@
...
+ * @return void
+ */
+function logger( $str, $type = 'i' ) {
+	$str = match ( $type ) {
+		'e' => "\033[31m$str \033[0m\n",
+		's' => "\033[32m$str \033[0m\n",
+		'w' => "\033[33m$str \033[0m\n",
+		'i' => "\033[36m$str \033[0m\n",
+		default => $str,
+	};
```
westonruter: Ooo. Fancy. But `match` requires PHP8. Should this use `switch` instead for now?

package.json
thelovekesh: > Why do we need these here? What prevents us from implementing this as a PHP-based command in composer.json?

@felixarntz We can do it, but then if we run `lint:plugins` from npm and it requires to pass PHPCS args, we will need to escape the args twice e.g. `npm run lint:php:plugins -- -- --help`.

thelovekesh: I think the build commands just change that and even in composer we are using `:` notation and this is pretty much common even in the Gutenberg repo, and the same with the `:fix` notation.

bin/phpcs/lint.php
thelovekesh: Humm makes sense. I will make it a switch case, and it will not be fancy anymore 😢

westonruter: We can make it fancy again in the future 😄

package.json
thelovekesh: @felixarntz @westonruter Can you please share your opinion here?

westonruter: If we can avoid the extra `--` in the commands, I'm all for it!

felixarntz: > > Why do we need these here? What prevents us from implementing this as a PHP-based command in composer.json?
> 
> @felixarntz We can do it, but then if we run `lint:plugins` from npm and it requires to pass PHPCS args, we will need to escape the args twice e.g. `npm run lint:php:plugins -- -- --help`.

IMO we shouldn't have to use NPM to run those PHP scripts in the first place. If we implement this with Composer, we can probably find a way to do it like `composer lint:plugins -- --plugin=webp-uploads`.

Of course we could also consider making it work well for both. Wouldn't we be able to use `npm run lint:php:plugins --plugin=webp-uploads` by having that command call `composer lint:plugins -- --plugin=$npm_config_plugin`? With something along those lines, we could get the best of both worlds.

westonruter: > IMO we shouldn't have to use NPM to run those PHP scripts in the first place. If we implement this with Composer, we can probably find a way to do it like `composer lint:plugins -- --plugin=webp-uploads`.

Isn't it better for DX to have all the commands available in either `npm` _or_ `composer` but not both? It is not helpful to have to try to remember whether a task is accomplished with `npm` or `composer` each time you need to run it. If we do have the scripts in `composer.json`, then I think they should just be there for redundancy. But with redundancy, then there is more maintenance overhead. 

Also, if we can avoid having to add that extra `--` it will also be helpful. I don't know about you, but I often forget and then it doesn't work as expected, so then I have to go back and remember to add it, and put it in the right spot.

thelovekesh: I believe one of the primary advantages of sticking with NPM is its ability to execute commands from any directory, providing flexibility. However, if we were to integrate this script into Composer and interpret the `--plugin` flag within it, we would face the challenge of distinguishing between arguments meant for the script and those intended for PHPCS.

NPM comes to our aid by treating it as a configuration variable if a specific flag is present, converting it into an environment variable. Subsequently, within the script, we can access the environment variables prefixed with `WPP_` to discern which arguments are intended for scripts, while any other arguments passed to the script are considered parameters for the PHPCS execution within it.

thelovekesh: @felixarntz As you mentioned in https://github.com/WordPress/performance/pull/1090#discussion_r1548554987, what if we keep NPM as a wrapper inside the composer? It will work well in that case IMO. I understand your apprehension about including PHP-specific elements in Composer. However, in terms of practical execution, the process remains the same whether we call it within NPM or Composer. It's essentially an implementation detail that shouldn't disrupt any workflows, as the `lint:php` command will perform its intended function regardless of the script runner in which it's executed.

 Inside composer scripts, add `"lint:plugins": "npm run lint:php:plugins"` and we can call it like:
```
composer lint:plugins # Lint all plugins.
composer lint:plugins -- --plugin=foo # Lint specific plugin
composer lint:plugins -- --plugin=auto-sizes -- --help # Pass args to PHPCS.
```

felixarntz: > Isn't it better for DX to have all the commands available in either `npm` _or_ `composer` but not both? It is not helpful to have to try to remember whether a task is accomplished with `npm` or `composer` each time you need to run it.

I'm not saying we have to have them in both, that was just to try to satisfy both your and my concerns. To me, there's a simple way to separate without having to remember one by one: Commands for PHP go into `composer.json`, while commands for JS or other (not PHP) environment related things go into `package.json`.

felixarntz: Again, the bigger question and what we need to further discuss in #1012 is whether the approach of this PR makes sense. Trying to solve both developing from the root directory and developing from the individual plugin directory already shows as of this PR how many workarounds it requires and headaches it gives due to those workarounds being controversial to some contributors.

While my preference for various reasons is to encourage development from the root directory, either way that we decide for the solution may differ:
* If everyone contributes to this from the root directory, then I'm not sure whether this change is needed.
* If everyone contributes from the individual plugin's directory, then we should simply have `composer.json` files per plugin directory (similar to how other monorepos of that approach do it) which would make it work without such workarounds.

Both of this would be much simpler than what this PR currently does.

westonruter: > Commands for PHP go into `composer.json`, while commands for JS or other (not PHP) environment related things go into `package.json`.

Looking at Gutenberg, I see they do have composer scripts: https://github.com/WordPress/gutenberg/blob/6490514a4e5fd14635de1faaa6d395201e248620/composer.json#L49-L52

But those seem to be utilities which are invoked by npm scripts:

https://github.com/WordPress/gutenberg/blob/6490514a4e5fd14635de1faaa6d395201e248620/package.json#L290-L291
https://github.com/WordPress/gutenberg/blob/6490514a4e5fd14635de1faaa6d395201e248620/package.json#L301
https://github.com/WordPress/gutenberg/blob/6490514a4e5fd14635de1faaa6d395201e248620/package.json#L337-L338

So npm scripts are always a superset of the composer scripts and it seems the npm scripts are those which developers are expected to be invoked directly. It makes sense to me for all scripts for developers to run to be in one spot.

lint-staged.config.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+// Get plugin base name to match regex more accurately.
+// Else it can cause issues when this plugin is places in `wp-content/plugins` directory.
+const PLUGIN_BASE_NAME = path.basename( __dirname );
+
+module.exports = {
+	'**/*.js': ( files ) => `npm run lint-js ${ joinFiles( files ) }`,
+	'**/*.php': ( files ) => {
+		const commands = [ 'composer phpstan' ];
```
westonruter, line 30: As noted by @thelovekesh in https://github.com/WordPress/performance/pull/1228#issuecomment-2125658548, this intentionally does _not_ include the list of changed files as arguments as PHPStan should be run across the entire codebase.

package.json
```diff
@@ -41,21 +43,16 @@
...
     "lint-js": "wp-scripts lint-js",
-    "format-php": "composer format",
     "phpstan": "composer phpstan",
-    "lint-php": "composer lint:all",
-    "test-php": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test:plugins",
-    "test-php-multisite": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test-multisite:plugins",
+    "lint:php": "composer lint",
+    "lint:php:fix": "composer format",
+    "lint:php:plugins": "cross-env WPP_PLUGIN=$npm_config_plugin php -f bin/phpcs/lint.php",
+    "lint:php:plugins:fix": "cross-env WPP_FIX=1 WPP_PLUGIN=$npm_config_plugin php -f bin/phpcs/lint.php",
```
westonruter, line 50: I guess these need to be updated?

```suggestion
    "lint:php:plugins": "cross-env WPP_PLUGIN=$npm_config_plugin php -f tools/phpcs/lint.php",
    "lint:php:plugins:fix": "cross-env WPP_FIX=1 WPP_PLUGIN=$npm_config_plugin php -f tools/phpcs/lint.php",
```

tools/phpcs/lint.php
```diff
@@ -0,0 +1,112 @@
...
+ * Would run phpcs on plugin1 and plugin2.
+ *
+ * $ WPP_PLUGIN=plugin1,plugin2 WPP_FIX=1 php bin/phpcs/lint.php
+ * Would run phpcbf on plugin1 and plugin2.
+ *
+ * $ php bin/phpcs/lint.php
+ * Would run phpcs on all plugins.
+ *
+ * $ WPP_FIX=1 php bin/phpcs/lint.php
+ * Would run phpcbf on all plugins.
```
westonruter, line 17:
```suggestion
 * $ WPP_PLUGIN=plugin1,plugin2 php tools/phpcs/lint.php
 * Would run phpcs on plugin1 and plugin2.
 *
 * $ WPP_PLUGIN=plugin1,plugin2 WPP_FIX=1 php tools/phpcs/lint.php
 * Would run phpcbf on plugin1 and plugin2.
 *
 * $ php tools/phpcs/lint.php
 * Would run phpcs on all plugins.
 *
 * $ WPP_FIX=1 php tools/phpcs/lint.php
 * Would run phpcbf on all plugins.
```

.github/workflows/php-lint.yml
```diff
@@ -22,6 +23,7 @@ on:
       - 'phpstan.neon.dist'
       - 'composer.json'
       - 'composer.lock'
+      - 'bin/phpcs'
```
westonruter, line 26: Is this needed now that the script ends in `.php` (`tools/phpcs/lint.php`)?

```suggestion
```

.github/workflows/php-lint.yml
```diff
@@ -13,6 +13,7 @@ on:
...
```
westonruter, line 16: Is this needed now that the script ends in `.php` (`tools/phpcs/lint.php`)?

```suggestion
```

thelovekesh:
@westonruter I think we need to split this PR into two parts as well, similar to what we did in https://github.com/WordPress/performance/pull/1065#issuecomment-2140072571

1. Fix lint-staged per plugin 
2. Running PHPCS within plugins.

For this PR, the main blocker was the second part. I think we can discuss that a little more and then raise a PR.

thelovekesh:
Closing in favor of the above comment. cc @westonruter

PR #1091, opened 2024-03-26
Improve UX/UI of enabling performance features via standalone plugins

### LINKED ISSUE #1031: Improve UX/UI of enabling performance features via standalone plugins

mukeshpanchal27:
## Summary

Fixes #1031

Before:
Screenshot 2024-03-26 at 5 19 40 PM

After:
Screenshot 2024-03-26 at 5 17 54 PM


swissspidy:
Could you perhaps add a before/after screenshot/video to the PR description? Makes reviews a bit easier :)

mukeshpanchal27:
@swissspidy Screenshot added in PR description.

includes/admin/load.php
```diff
@@ -224,31 +228,85 @@ function perflab_enqueue_modules_page_scripts() {
 	wp_enqueue_script( 'plugin-install' );
 
 	wp_enqueue_script(
-		'perflab-plugin-management',
-		plugin_dir_url( __FILE__ ) . 'js/perflab-plugin-management.js',
+		'perflab-install-activate-plugins',
+		plugin_dir_url( __FILE__ ) . 'js/perflab-install-activate-plugins.js',
 		array( 'jquery' ),
```
thelovekesh: Do you think this can be omitted now? I don't see jQuery being used with the script.

includes/admin/js/perflab-install-activate-plugins.js
```diff
@@ -0,0 +1,53 @@
...
+				data.append( 'action', 'perflab_install_activate_plugins' );
+				data.append( 'nonce', perflab_install_activate_plugins.nonce );
+				data.append( 'slug', target.dataset.slug );
+
+				fetch( perflab_install_activate_plugins.ajaxurl, {
+					method: 'POST',
+					credentials: 'same-origin',
+					body: data,
+				} )
+					.then( function ( response ) {
```
thelovekesh: I think we can use `async/await` instead of this `.then()` callbacks.

includes/admin/load.php
```diff
@@ -224,31 +228,85 @@ function perflab_enqueue_modules_page_scripts() {
...
 		)
 	);
+
+	wp_localize_script(
+		'perflab-install-activate-plugins',
+		'perflab_install_activate_plugins',
+		array(
+			'ajaxurl' => admin_url( 'admin-ajax.php' ),
+			'nonce'   => wp_create_nonce( 'perflab-install-activate-plugins' ),
+		)
```
thelovekesh: I think it should go through `wp_json_encode()`.

includes/admin/load.php
```diff
@@ -224,31 +228,85 @@ function perflab_enqueue_modules_page_scripts() {
...
  *
- * @since 2.8.0
+ * @since n.e.x.t
  */
-function perflab_activate_plugin() {
-	// Do not proceed if plugin query arg is not present.
-	if ( empty( $_GET['plugin'] ) ) {
-		return;
+function perflab_install_activate_plugins_callback() {
+	if ( ! wp_verify_nonce( $_REQUEST['nonce'], 'perflab-install-activate-plugins' ) ) {
```
thelovekesh: One more check can be added to check if `$_REQUEST['nonce']` exists.

includes/admin/load.php
```diff
@@ -224,31 +228,85 @@ function perflab_enqueue_modules_page_scripts() {
...
+		$status['errorMessage'] = __( 'Invalid nonce: Please refresh and try again.', 'performance-lab' );
```
thelovekesh: I see that `$status` is not initialized. If `$status['errorMessage']` is the only status we are passing, can't we pass it directly in `wp_send_json_error();`?

thelovekesh:
When the user loads the `performance-lab` options page for the first time, the options page takes time to load since it queries the plugin's API. As the number of plugins grows this time will increase. Can we do something to make this step async? 

Also, can we use `@wordpress/components` here instead of SSR'd content? That could handle the above case but it would require a build step in the PL plugin.

thelovekesh:
@mukeshpanchal27 On the WP trunk the rendered screen seems to be breaking. Can you please confirm? Or am I missing something 🤔 

image

swissspidy:
This doesn't yet address all the suggestions for an initial improvement aka MVP outlined at https://github.com/WordPress/performance/issues/1031#issuecomment-1980033829, specifically:

> * The feature name and description shouldn't be simply the content pulled from .org, but rather custom content that is more tailored to the user experience of what the feature unlocks.
> * We can hard-code the list of features, i.e. the name, description, and underlying plugin slug.
> * The logic to get information from wordpress.org can be removed.

The latter two points should also alleviate the load time concern mentioned above

> Also, can we use `@wordpress/components` here instead of SSR'd content? That could handle the above case but it would require a build step in the PL plugin.

That's a bit out of scope for an MVP like this, but something we should consider for future iterations.

felixarntz:
@swissspidy 
> This doesn't yet address all the suggestions for an initial improvement aka MVP outlined at [#1031 (comment)](https://github.com/WordPress/performance/issues/1031#issuecomment-1980033829), specifically:
> 
> > * The feature name and description shouldn't be simply the content pulled from .org, but rather custom content that is more tailored to the user experience of what the feature unlocks.
> > * We can hard-code the list of features, i.e. the name, description, and underlying plugin slug.
> > * The logic to get information from wordpress.org can be removed.

This was initially defined as a requirement, but later discarded. See https://github.com/WordPress/performance/issues/1031#issuecomment-1989043559 for the latest requirements.

> > Also, can we use `@wordpress/components` here instead of SSR'd content? That could handle the above case but it would require a build step in the PL plugin.
> 
> That's a bit out of scope for an MVP like this, but something we should consider for future iterations.

Agreed, this is out of scope here. Let's discuss such a change separately.

cc @mukeshpanchal27

includes/admin/plugins.php
```diff
@@ -129,108 +128,37 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+		$action_links[] = sprintf(
+			'<button type="button" class="button button-disabled" disabled="disabled">%s</button>',
+			esc_html__( 'Active', 'default' )
+		);
+	} elseif (
+		$compatible_php &&
+		$compatible_wp &&
+		current_user_can( 'install_plugins', $status['file'] ) &&
+		current_user_can( 'activate_plugin', $status['file'] )
+	) {
```
felixarntz: It would be better here to only check for `install_plugins` if the plugin actually needs to be installed. Since the `$status` variable already provides this information, that should be straightforward to do.

Also, please note that `install_plugins` does not use the specific plugin basename, so it needs to be removed.
```suggestion
	} elseif (
		$compatible_php &&
		$compatible_wp &&
		current_user_can( 'activate_plugin', $status['file'] ) &&
		(
			'install' !== $status ||
			current_user_can( 'install_plugins' )
		)
	) {
```

includes/admin/plugins.php
```diff
@@ -129,108 +128,37 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+		$action_links[] = sprintf(
+			'<button type="button" class="button perflab-install-active-plugin" data-slug="%s">%s</button>',
+			esc_attr( $plugin_data['slug'] ),
+			esc_html__( 'Activate', 'default' )
+		);
+	} else {
+		$action_links[] = sprintf(
+			'<button type="button" class="button button-disabled" disabled="disabled">%s</button>',
+			esc_html( _x( 'Cannot Install', 'plugin', 'default' ) )
+		);
```
felixarntz: Let's correctly identify whether the plugin cannot be _activated_ or not be _installed_. Or alternatively, we always use "Cannot Activate" as that is more in line with the language on the active buttons.

Since this is a special case scenario, I'd say more granular explanations are still more helpful in conveying what's wrong.
```suggestion
		$explanation    = 'install' !== $status || current_user_can( 'install_plugins' ) ? _x( 'Cannot Activate', 'plugin', 'default' ) : _x( 'Cannot Install', 'plugin', 'default' );
		$action_links[] = sprintf(
			'<button type="button" class="button button-disabled" disabled="disabled">%s</button>',
			esc_html( $explanation )
		);
```

includes/admin/plugins.php
```diff
@@ -129,108 +128,37 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
-				$action_links[] = sprintf(
-					'<button type="button" class="button button-disabled" disabled="disabled">%s</button>',
-					esc_html( _x( 'Installed', 'plugin', 'default' ) )
-				);
-			}
-			break;
+	if ( is_plugin_active( $status['file'] ) ) {
+		$action_links[] = sprintf(
+			'<button type="button" class="button button-disabled" disabled="disabled">%s</button>',
+			esc_html__( 'Active', 'default' )
```
felixarntz: Better to use the actual core string including the context:
```suggestion
			esc_html( _x( 'Active', 'plugin', 'default' ) )
```

includes/admin/load.php
```diff
@@ -262,79 +320,45 @@ function perflab_activate_plugin() {
...
+		$status['errorMessage'] = $result->get_error_message();
+		wp_send_json_error( $status );
 	}
 
-	if ( wp_safe_redirect( $referer ) ) {
-		exit;
-	}
+	wp_send_json_success( $status );
 }
-add_action( 'admin_action_perflab_activate_plugin', 'perflab_activate_plugin' );
```
felixarntz, line 283: Similar to my feedback in #1048, why use AJAX instead of the much simpler admin "action URL" implementation? I'd argue here this is even more relevant, since the previous code was using that approach, and I don't see a benefit for changing this here.

Using AJAX makes the PR unnecessarily complex. I would advise to go back to a regular admin action, and modify the existing action to both install and activate, rather than implementing a whole new JavaScript asset and JSON response behavior. Last but not least, using AJAX also comes with a higher risk that something else in the UI of the page is missed and not updated, so that's another reason to go with the simpler "action URL" approach.

includes/admin/load.php
```diff
@@ -262,79 +320,45 @@ function perflab_activate_plugin() {
...
-		<div class="notice notice-success is-dismissible">
-			<p><?php esc_html_e( 'Plugin activated.', 'default' ); ?></p>
-		</div>
-		<?php
-	} elseif ( isset( $_GET['deactivate'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-		?>
-		<div class="notice notice-success is-dismissible">
-			<p><?php esc_html_e( 'Plugin deactivated.', 'default' ); ?></p>
-		</div>
-		<?php
```
felixarntz: See my other comment, the "activate" specific notice could be maintained if we continued using a simple admin action.

I would recommend to change the wording to use "Feature" intstead of "Plugin" here.

felixarntz (changes requested):
Thank you @mukeshpanchal27, leaving a bit of feedback.

includes/admin/load.php
westonruter: Not if using `wp_localize_script()`. Using `wp_json_encode()` is only relevant if using `wp_add_inline_script()` where you have to manually encode. For example:

```php
wp_add_inline_script(
    'perflab-install-activate-plugins',
    sprintf( 'var perflab_install_activate_plugins = %s;', wp_json_encode(
		array(
			'ajaxurl' => admin_url( 'admin-ajax.php' ),
			'nonce'   => wp_create_nonce( 'perflab-install-activate-plugins' ),
		)
    ) ),
    'before'
);
```

includes/admin/js/perflab-install-activate-plugins.js
```diff
@@ -0,0 +1,53 @@
...
+	document.addEventListener( 'DOMContentLoaded', function () {
+		document.addEventListener( 'click', function ( event ) {
+			if (
+				event.target.classList.contains(
+					'perflab-install-active-plugin'
+				)
+			) {
+				const target = event.target;
+				target.classList.add( 'updating-message' );
+				target.innerHTML = wp.i18n.__( 'Activating…' );
```
westonruter: What about the text domain?

includes/admin/load.php
westonruter: Actually, I think `wp_verify_nonce()` can rather be replaced with `check_ajax_referer()`:

```suggestion
	if ( ! check_ajax_referer( 'perflab-install-activate-plugins', 'nonce', false ) ) {
```

includes/admin/load.php
```diff
@@ -224,31 +228,85 @@ function perflab_enqueue_modules_page_scripts() {
...
+		$status['errorMessage'] = __( 'Sorry, you are not allowed to manage plugins for this site. Please contact the administrator.', 'performance-lab' );
+		wp_send_json_error( $status );
+	}
 
-	check_admin_referer( "perflab_activate_plugin_{$plugin}" );
+	require_once ABSPATH . 'wp-admin/includes/plugin-install.php';
+	require_once ABSPATH . 'wp-admin/includes/class-wp-upgrader.php';
+	require_once ABSPATH . 'wp-admin/includes/class-wp-ajax-upgrader-skin.php';
+
+	$plugin = $_REQUEST['slug'];
```
```suggestion
	$plugin = isset( $_REQUEST['slug'] ) ? wp_unslash( $_REQUEST['slug'] ) : '';
```

Could be further sanitized if necessary.

includes/admin/js/perflab-install-activate-plugins.js
swissspidy: Good catch, for the JS translations we can't easily use the core default text domain, so we should use `performance-lab` here instead of nothing

includes/admin/load.php
mukeshpanchal27: Thank you for bringing this up. Following other feedback, I've opted for the admin action approach, so I've removed this JavaScript.

includes/admin/load.php
mukeshpanchal27: Initially, I thought the AJAX approach would work for these changes. However, based on your direction, I updated the approach.

includes/admin/js/perflab-install-activate-plugins.js
mukeshpanchal27: Updated in https://github.com/WordPress/performance/pull/1091/commits/2630a2fea49e9c856bb59851ab3cf03e2529c179

mukeshpanchal27:
> WP trunk the rendered screen seems to be breaking. Can you please confirm? Or am I missing something 🤔

@thelovekesh Trunk version works fine for me. Could you please check if you can replicate the issue?

@felixarntz @westonruter PR ready for review.

includes/admin/load.php
```diff
@@ -204,51 +210,75 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
-	);
+/**
+ * Register admin actions for handling performance feature plugin installation/activation.
+ *
+ * @since n.e.x.t
+ */
+function perflab_register_admin_actions() {
+	add_action( 'admin_action_perflab_install_activate_plugins', 'perflab_install_activate_plugins_callback' );
 }
+add_action( 'admin_init', 'perflab_register_admin_actions' );
```
swissspidy: Why the extra function? Can't we just do the following?

```suggestion
add_action( 'admin_action_perflab_install_activate_plugins', 'perflab_install_activate_plugins_callback' );
```

includes/admin/load.php
```diff
@@ -204,51 +210,75 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
+
+	if ( ! $plugin ) {
+		wp_die( esc_html__( 'Invalid plugin.', 'performance-lab' ) );
 	}
 
-	$plugin = sanitize_text_field( wp_unslash( $_GET['plugin'] ) );
+	$api = perflab_query_plugin_info( $plugin );
 
-	check_admin_referer( "perflab_activate_plugin_{$plugin}" );
+	// Return early if plugin API return an error.
```
swissspidy:
```suggestion
	// Return early if plugin API returns an error.
```

includes/admin/load.php
```diff
@@ -262,79 +292,92 @@ function perflab_activate_plugin() {
...
+	document.addEventListener( 'DOMContentLoaded', function () {
+		document.addEventListener( 'click', function ( event ) {
+			if (
+				event.target.classList.contains(
+					'perflab-install-active-plugin'
+				)
+			) {
+				const target = event.target;
+				target.classList.add( 'updating-message' );
+				target.innerHTML = wp.i18n.__( 'Activating…', 'performance-lab' );
```
swissspidy: WordPress doesn't (yet) support JS translations for inline scripts.

So right now there is no way to translate this string.

A few options:

1. Put this back into an external file and call `wp_set_script_translations()`
2. Register a script handle and call `wp_set_script_translations()`, but continue printing as inline script
3. Print this in PHP instead and use as variable here (e.g. `$activating = esc_js( ... );` at the top of the function and then use `$activating` here.

felixarntz: Taking a step back here, why do we need this script even? It makes sense to me in an AJAX based workflow where you're waiting for a response, but now that this button simply loads another URL, I think changing the button's appearance is a bit odd. 🤔

includes/admin/load.php
```diff
@@ -204,51 +210,75 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
-			'settings' => array(
-				'totals' => wp_get_update_data(),
-			),
-		)
-	);
-
+function perflab_enqueue_features_page_scripts() {
 	wp_enqueue_script( 'thickbox' );
 	wp_enqueue_style( 'thickbox' );
+}
```
felixarntz: Does this have any use, given there are no other scripts now? If not, we should probably no longer enqueue it.

includes/admin/plugins.php
```diff
@@ -129,108 +128,52 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+			'install' !== $status ||
+			current_user_can( 'install_plugins' )
+		)
+	) {
+		$url = esc_url_raw(
+			add_query_arg(
+				array(
+					'action'   => 'perflab_install_activate_plugins',
+					'_wpnonce' => wp_create_nonce( 'perflab_install_activate_plugins' ),
+					'slug'     => $plugin_data['slug'],
```
felixarntz: Passing the plugin basename (`$status['file']`) here would allow us to simplify things a lot. We wouldn't need to to do all the guess work: If the plugin is already installed, we can use the value to activate it. If the value is empty, we know it's not installed.
```suggestion
					'action'   => 'perflab_install_activate_plugins',
					'_wpnonce' => wp_create_nonce( 'perflab_install_activate_plugins' ),
					'slug'     => $plugin_data['slug'],
					'file'     => $status['file']
```

includes/admin/load.php
```diff
@@ -204,51 +210,75 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
+	require_once ABSPATH . 'wp-admin/includes/class-wp-upgrader.php';
+	require_once ABSPATH . 'wp-admin/includes/class-wp-ajax-upgrader-skin.php';
+
+	if ( ! isset( $_GET['slug'] ) ) {
+		wp_die( esc_html__( 'Missing required parameter.', 'performance-lab' ) );
+	}
+
+	if ( ! current_user_can( 'install_plugins' ) || ! current_user_can( 'activate_plugins' ) ) {
+		wp_die( esc_html__( 'Sorry, you are not allowed to manage plugins for this site. Please contact the administrator.', 'performance-lab' ) );
+	}
```
felixarntz: Related to my previous feedback: We should check permissions more granularly. A few things here:
* Please use `activate_plugin` (singular) with the plugin basename instead of the generic `activate_plugins`.
* We should only check for the `install_plugins` capability if the plugin isn't already installed yet.

Based on my other recommendation to pass the `file` (which contains the basename _if_ the plugin is already installed), we can simplify this function quite a bit:
1. If `file` is empty, we know the plugin needs to be installed and then activated. Otherwise, we know the plugin only needs to be activated.
2. If the plugin needs to be installed, we need to check the `install_plugins` capability. Only then we run all the installation specific logic.
3. Then, we determine the plugin basename (only relevant if `file` is empty, i.e. after installation of the plugin).
4. Then we check for the `activate_plugin` capability passing the basename. Then we can run the activation logic.

includes/admin/load.php
```diff
@@ -204,51 +210,75 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
+			wp_die( esc_html( $result->get_error_message() ) );
+		} elseif ( is_wp_error( $skin->result ) ) {
+			wp_die( esc_html( $skin->result->get_error_message() ) );
+		} elseif ( $skin->get_errors()->has_errors() ) {
+			wp_die( esc_html( $skin->get_error_messages() ) );
+		}
+	}
 
 	// If `$plugin` is a plugin slug rather than a plugin basename, determine the full plugin basename.
 	if ( ! str_contains( $plugin, '/' ) ) {
```
felixarntz: This check isn't useful now because `$plugin` is _always_ the slug based on your logic changes. Instead, related to my other feedback, we should check whether the passed `file` is empty, and if so use the `slug` to determine the file / basename below.

includes/admin/load.php
```diff
@@ -204,51 +210,75 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
+
+	if ( ! isset( $_GET['slug'] ) ) {
+		wp_die( esc_html__( 'Missing required parameter.', 'performance-lab' ) );
+	}
+
+	if ( ! current_user_can( 'install_plugins' ) || ! current_user_can( 'activate_plugins' ) ) {
+		wp_die( esc_html__( 'Sorry, you are not allowed to manage plugins for this site. Please contact the administrator.', 'performance-lab' ) );
+	}
+
+	$plugin = sanitize_text_field( wp_unslash( $_GET['slug'] ) );
```
felixarntz: Let's rename this variable since `$plugin` is not clear in what this is. The variable contains the slug, so it should be called `$slug` or `$plugin_slug`.

includes/admin/load.php
```diff
@@ -204,51 +210,75 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
+	$api = perflab_query_plugin_info( $plugin );
 
-	check_admin_referer( "perflab_activate_plugin_{$plugin}" );
+	// Return early if plugin API return an error.
+	if ( ! $api ) {
+		wp_die( esc_html__( 'An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration.', 'performance-lab' ) );
+	}
+
+	$plugins             = get_plugins();
+	$is_plugin_installed = ! ( isset( $plugins[ $plugin . '/load.php' ] ) || isset( $plugins[ $plugin . '/' . $plugin . '.php' ] ) );
```
felixarntz: This is not reliable, we can't make any assumptions on the plugin file name. Instead, let's rely on the passed `file` value, which is determined by WordPress core's `install_plugin_install_status()` function.

Something like this:
```php
$is_plugin_installed = ! isset( $_GET['file'] ) || ! $_GET['file'];
```

includes/admin/load.php
```diff
@@ -204,51 +210,75 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
+	}
+
+	$plugin = sanitize_text_field( wp_unslash( $_GET['slug'] ) );
+
+	if ( ! $plugin ) {
+		wp_die( esc_html__( 'Invalid plugin.', 'performance-lab' ) );
 	}
 
-	$plugin = sanitize_text_field( wp_unslash( $_GET['plugin'] ) );
+	$api = perflab_query_plugin_info( $plugin );
```
felixarntz: This is only needed if the plugin isn't installed yet. Let's put it into the corresponding condition, to not run it unnecessarily if the plugin only needs to be activated.

includes/admin/load.php
```diff
@@ -204,51 +210,75 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
  *
- * @since 2.8.0
+ * @since n.e.x.t
  */
-function perflab_activate_plugin() {
-	// Do not proceed if plugin query arg is not present.
-	if ( empty( $_GET['plugin'] ) ) {
-		return;
+function perflab_install_activate_plugins_callback() {
+	check_admin_referer( 'perflab_install_activate_plugins' );
```
felixarntz: Let's rename the function and the nonce action to use singular "plugin", not "plugins", since it only handles a single plugin.

includes/admin/plugins.php
```diff
@@ -129,108 +128,52 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
-					);
+	if ( is_plugin_active( $status['file'] ) ) {
+		$action_links[] = sprintf(
+			'<button type="button" class="button button-disabled" disabled="disabled">%s</button>',
+			esc_html( _x( 'Active', 'plugin', 'default' ) )
+		);
+	} elseif (
+		$compatible_php &&
+		$compatible_wp &&
+		current_user_can( 'activate_plugin', $status['file'] ) &&
```
felixarntz: Thinking about this further, we need to be careful here. We can only check for the specific `activate_plugin` capability if the plugin is already installed. If it's not installed, `$status['file']` will be empty. In that case we need to fall back to the generic `activate_plugins` capability.
```suggestion
		(
			$status['file'] && current_user_can( 'activate_plugin', $status['file'] ) ||
			current_user_can( 'activate_plugins' )
		) &&
```

includes/admin/load.php
mukeshpanchal27: Updated!

includes/admin/load.php
mukeshpanchal27: Make sense. Removed the JS.

includes/admin/load.php
mukeshpanchal27: @felixarntz This JS is required for the `Learn more` button popup. Additionally, `wp_enqueue_script( 'plugin-install' )` has been added for autosizing the popup as we already fix previously in #896.

swissspidy:
Should we remove the "Learn more" link? It's redundant right now as you can just click on the title to open the modal.

mukeshpanchal27:
> Should we remove the "Learn more" link? It's redundant right now as you can just click on the title to open the modal.

Certainly, if we agree to remove it, I'll go ahead and do so. However, it's worth noting that it was included in the requirement outlined here: https://github.com/WordPress/performance/issues/1031#issuecomment-1989043559

includes/admin/load.php
```diff
@@ -204,137 +207,142 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
-				'totals' => wp_get_update_data(),
-			),
-		)
-	);
-
+function perflab_enqueue_features_page_scripts() {
 	wp_enqueue_script( 'thickbox' );
 	wp_enqueue_style( 'thickbox' );
 
 	wp_enqueue_script( 'plugin-install' );
```
felixarntz: Maybe add a comment to explain why these are being enqueued? For example:
```suggestion
	// These assets are needed for the "Learn more" popover.
	wp_enqueue_script( 'thickbox' );
	wp_enqueue_style( 'thickbox' );
	wp_enqueue_script( 'plugin-install' );
```

includes/admin/load.php
```diff
@@ -204,137 +207,142 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
 
-	check_admin_referer( "perflab_activate_plugin_{$plugin}" );
+	if ( ! $plugin_slug ) {
+		wp_die( esc_html__( 'Invalid plugin.', 'performance-lab' ) );
+	}
+
+	$is_plugin_installed = ! isset( $_GET['file'] ) || ! $_GET['file'];
+
+	// Install the plugin if it is not installed yet.
+	if ( $is_plugin_installed ) {
```
felixarntz: I think this logic is the wrong way around, given the variable name.
```suggestion
	$is_plugin_installed = isset( $_GET['file'] ) && $_GET['file'];

	// Install the plugin if it is not installed yet.
	if ( ! $is_plugin_installed ) {
```

includes/admin/load.php
```diff
@@ -204,137 +207,142 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
+		// Check if the user have plugin installation capability.
+		if ( ! current_user_can( 'install_plugins' ) ) {
+			wp_die( esc_html__( 'Sorry, you are not allowed to manage plugins for this site. Please contact the administrator.', 'performance-lab' ) );
```
felixarntz: We could make use of a core string here instead to simplify this and avoid "extra work" for translators :)
```suggestion
			wp_die( esc_html__( 'Sorry, you are not allowed to install plugins on this site.', 'default' ) );
```

includes/admin/load.php
```diff
@@ -204,137 +207,142 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
+		}
+
+		$api = perflab_query_plugin_info( $plugin_slug );
+
+		// Return early if plugin API returns an error.
+		if ( ! $api ) {
+			wp_die( esc_html__( 'An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration.', 'performance-lab' ) );
```
felixarntz: Same here, we can make use of a core string instead.
```suggestion
			wp_die(
				wp_kses(
					sprintf(
						/* translators: %s: Support forums URL. */
						__( 'An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration. If you continue to have problems, please try the <a href="%s">support forums</a>.', 'default' ),
						__( 'https://wordpress.org/support/forums/', 'default' )
					),
					array( 'a' => array( 'href' => true ) )
				)
			);
```

includes/admin/load.php
```diff
@@ -204,137 +207,142 @@ function perflab_dismiss_wp_pointer_wrapper() {
...
-		$plugin            = $plugin . '/' . $plugin_file_names[0];
+		$plugin_basename   = $plugin_slug . '/' . $plugin_file_names[0];
+	} else {
+		$plugin_basename = $_GET['file'];
 	}
 
-	if ( ! current_user_can( 'activate_plugin', $plugin ) ) {
-		wp_die( esc_html__( 'Sorry, you are not allowed to activate this plugin.', 'default' ) );
+	if ( ! current_user_can( 'activate_plugin', $plugin_basename ) ) {
+		wp_die( esc_html__( 'Sorry, you are not allowed to manage plugins for this site. Please contact the administrator.', 'performance-lab' ) );
```
felixarntz: Same here:
```suggestion
		wp_die( esc_html__( 'Sorry, you are not allowed to activate this plugin.', 'default' ) );
```

felixarntz (commented):
@mukeshpanchal27 This looks great, almost good to go. Just a few last points.

felixarntz:
@swissspidy @mukeshpanchal27 I think we should keep the "Learn more" link. Even though you can access the modal from two places, the Learn more link is more obvious in allowing that than the title being clickable. FWIW, that's also how the regular plugin cards at _Plugins > Add New_ handle it.

includes/admin/load.php
westonruter: Perhaps simpler as:

```php
$is_plugin_installed = ! empty( $_GET['file'] );
```

westonruter: Still, seems like there should be some validation of this input here. And store the sanitized value in `$file`, as was done to store `$_GET['slug']` in `$plugin_slug`.

felixarntz: Regarding the `empty`, I'd prefer not to use it, based on e.g. https://dev.to/klnjmm/never-use-empty-function-in-php-4pb0 or https://localheinz.com/articles/2023/05/10/avoiding-empty-in-php/. It's not a big deal either way, but the way it currently is written is more explicit.

+1 to sanitizing the input and storing it in `$file`!

mukeshpanchal27: Sanitization added in https://github.com/WordPress/performance/pull/1091/commits/46357b03a53a67c1c404c8af99f62ec87a87f61b

felixarntz (approved):
@mukeshpanchal27 Awesome work, LGTM!

thelovekesh:
> @thelovekesh Trunk version works fine for me. Could you please check if you can replicate the issue?

@mukeshpanchal27 In this particular core [commit](https://github.com/WordPress/wordpress-develop/commit/2b08a77f5c953bd7f5236d4b29b0a42010cf913b#diff-fd27ee048c40fddec930c410d341428d36aa6546c1a7014d2ff0feffeb98002f), few new classes were added to `/wp-admin/css/list-tables.css` which are causing some styles issues I shared above in https://github.com/WordPress/performance/pull/1091#issuecomment-2020583850. Can you please check it once again? I think following steps should help with testing on `wp-env`.
- Destroy existing `wp-env` containers - `npm run wp-env destroy`.
- Add override config `.wp-env.override.json`:
   ```json
    {
      "core": "WordPress/WordPress#master"
    }
   ```
- Run `wp-env` again - `npm run wp-env start`.

felixarntz:
Thanks @thelovekesh, this is a great catch. In that case, it looks like we may need to output CSS conditionally, depending on WordPress version.

The current CSS works correctly for WP <6.5. But based on the changes from that commit, it looks like for WP >=6.5 we will have to include different CSS.

@mukeshpanchal27 Could you test this and conditionally include the CSS for WP >=6.5?

mukeshpanchal27:
@thelovekesh Great catch. I have added the CSS fix that was introduced in the latest (6.5) version. Could you please take a look when you have a moment?

felixarntz (approved):
Thanks @mukeshpanchal27 for the update. I've tested this in 6.4 and 6.5-RC3 and it looks good in both now. ✅

thelovekesh (approved):
LGTM

PR #1098, opened 2024-03-28
[POC] Performance Dashboard (Web Vitals)

swissspidy:
## Summary

See #1324

Leverage Optimization Detective to collect Core Web Vitals field data inside WordPress. See what real users are experiencing on your site, without CrUX.

Optimization Detective is uniquely positioned to also collect web vitals RUM data. This data can then be presented in the WordPress dashboard for site owners to keep an eye on their site's performance.

### Demo

The proof of concept looks as follows:

Dedicated dashboard page:

Screenshot 2024-03-28 at 17 28 25

Admin dashboard widget:

Screenshot 2024-04-10 at 10 16 17

### Advantages

* Works for smaller sites without CrUX field data
* You don't have to wait for CrUX data to be available
* Get data from other browsers
* No external data source required
* Tightly integrated into WordPress

Main difference to Optimization Detective _today_ is that ideally CWV

### Potential

In the future we could:

* Provide WP-specific call-to-actions, e.g. to fix the offending LCP image in the block editor
* Identify plugins causing LCP/CLS/etc. issues and provide call-to-actions
* Identify plugins causing CLS issues
* Expose (aggregated) CWV data publicly for analysis purposes
	* Opt-in or opt-out
	* Alternatively send to some telemetry server
* Use [Long Animation Frames API](https://developer.chrome.com/docs/web-platform/long-animation-frames) to identify offenders
* Collect `Server-Timing` measurements per page
	* Could even identify impact by plugin via [Origination](https://github.com/GoogleChromeLabs/wp-origination)
* Add insights to admin bar
* Add insights to block editor

### Related

I found an existing [Core Web Vitals Monitor](https://wordpress.org/plugins/core-web-vitals-monitor/) WordPress plugin which also does this but is somehow not maintained nor popular. We might be able to learn something from it.

See also:

* https://github.com/WordPress/gutenberg/issues/33578

## Relevant technical choices

(TODO)


plugins/optimization-detective/storage/rest-api.php
```diff
@@ -85,6 +88,26 @@ function od_register_endpoint() {
...
 	);
+
+	register_rest_route(
+		OD_REST_API_NAMESPACE,
+		OD_CHART_ROUTE . '/(?P<id>[\d]+)',
+		array(
+			'methods'             => WP_REST_Server::READABLE,
+			'args' => array(
+				'id' => array(
+					'description' => __( 'Unique identifier for the object.', 'web-stories' ),
```
westonruter:
```suggestion
					'description' => __( 'Unique identifier for the object.', 'optimization-detective' ),
```

plugins/optimization-detective/dashboard.php
```diff
@@ -0,0 +1,86 @@
...
+add_action( 'admin_menu', 'od_add_submenu_page' );
+
+function od_render_dashboard_page() {
+	$asset_file = plugin_dir_path( __FILE__ ) . 'build/performance-dashboard.asset.php';
+
+	if ( ! file_exists( $asset_file ) ) {
+		return;
+	}
+
+	$asset = include $asset_file;
```
westonruter: Probably better to error if the file is absent, as it means a build needs to happen. In fact, `load.php` will raise an error if a build hasn't happened.

```suggestion
	$asset = require $asset_file;
```

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -14,21 +14,27 @@
...
+ *                                timestamp: int,
+ *                                viewport: RectData,
+ *                                elements: ElementData[]
+ *                            }
+ * @phpstan-type WebVitalData array{
+ *                                LCP: int,
+ *                                CLS: int,
+ *                                INP: int,
+ *                                TTFB: int,
+ *                            }
```
westonruter: I was wrong to use `int` for `timestamp`. The WebVitals metrics also should be floats in looking at some sample data I'm storing:

```json
{"webVitals":{"LCP":240.89999997615814,"CLS":0,"INP":null,"TTFB":83.19999998807907},"timestamp":1711666589.886095}
```

```suggestion
 * @phpstan-type Data         array{
 *                                url: string,
 *                                timestamp: float,
 *                                viewport: RectData,
 *                                elements: ElementData[]
 *                            }
 * @phpstan-type WebVitalData array{
 *                                LCP: float,
 *                                CLS: float,
 *                                INP: float|null,
 *                                TTFB: float,
 *                            }
```

plugins/optimization-detective/detect.js
```diff
@@ -361,39 +381,80 @@ export default async function detect( {
...
+
+		webVitalsMetrics.LCP = [];
+		webVitalsMetrics.CLS = [];
+		webVitalsMetrics.INP = [];
+		webVitalsMetrics.TTFB = [];
 	}
 
-	// Clean up.
-	breadcrumbedElementsMap.clear();
+	log( 'Adding event listeners for backgrounding and unloading.' );
```
westonruter, line 462:
```suggestion
	if ( isDebug ) {
		log( 'Adding event listeners for backgrounding and unloading.' );
	}
```

westonruter: Meh, I have a lot of this `if( isDebug ) { log(...) }` code in this file. Probably should introduce a `debugLog`, `debugWarn`, and `debugError()` functions that remove this repetition.

plugins/optimization-detective/detect.js
```diff
@@ -86,13 +89,22 @@ function error( ...message ) {
  * @property {DOMRectReadOnly} boundingClientRect - Bounding client rectangle.
  */
 
+/**
+ * @typedef {Object} WebVitalsMetrics
+ * @property {number} lcp  - Largest contentful paint.
+ * @property {number} cls  - Cumulative layout shift.
+ * @property {number} inp  - Interaction to next paint.
+ * @property {number} ttfb - Time to first byte.
```
westonruter: Since the metrics can be null and they are uppercase below.

```suggestion
 * @property {?number} LCP  - Largest contentful paint.
 * @property {?number} CLS  - Cumulative layout shift.
 * @property {?number} INP  - Interaction to next paint.
 * @property {?number} TTFB - Time to first byte.
```

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -148,6 +155,34 @@ public static function get_json_schema(): array {
...
+								'required' => false,
+							),
+							'TTFB'              => array(
+								'type'     => 'number',
+								'required' => false,
+							),
+							'INP'              => array(
+								'type'     => 'number',
+								'required' => false,
+							),
```
westonruter:
```suggestion
							'LCP'              => array(
								'type'     => 'number',
								'required' => false,
								'minimum'  => 0,
							),
							'CLS'              => array(
								'type'     => 'number',
								'required' => false,
								'minimum'  => 0,
							),
							'TTFB'              => array(
								'type'     => 'number',
								'required' => false,
								'minimum'  => 0,
							),
							'INP'              => array(
								'type'     => 'number',
								'required' => false,
								'minimum'  => 0,
							),
```

plugins/optimization-detective/detect.js
```diff
@@ -324,9 +338,15 @@ export default async function detect( {
...
+		webVitals: {
+			LCP: null,
+			CLS: null,
+			INP: null,
+			TTFB: null,
+		},
 	};
 
-	const lcpMetric = lcpMetricCandidates.at( -1 );
+	const lcpMetric = webVitalsMetrics.LCP.at( -1 );
```
westonruter: I'm finding this is now returning `undefined`, and so none of the elements are getting `isLCP: true`. This appears to be because the promise was removed so it is not waiting for LCP to be reported. Perhaps it is better to capture the elements and then every time that the `onLCP` callback fires, it goes over all of the elements and then sets the `isLCP` flag.

plugins/optimization-detective/storage/rest-api.php
```diff
@@ -73,6 +75,7 @@ function od_register_endpoint() {
 				return od_handle_rest_request( $request );
 			},
 			'permission_callback' => static function () {
+				return true;
```
westonruter: Tip: What I do during development is I have a mini plugin with the following filters:

```php
add_filter( 'od_url_metric_storage_lock_ttl', '__return_zero' );
add_filter( 'od_url_metric_freshness_ttl', '__return_zero' );
add_filter( 'od_url_metrics_breakpoint_sample_size', function () { return 1; } );
```

Then you won't need this.

```suggestion
```

plugins/optimization-detective/storage/rest-api.php
```diff
@@ -29,6 +29,8 @@
  */
 const OD_URL_METRICS_ROUTE = '/url-metrics:store';
 
+const OD_CHART_ROUTE = '/url-metrics:chart';
```
westonruter, line 32: Aside: I was meaning to encapsulate these functions into an `OD_REST_API_Endpoints` class, but I didn't get around to it yet. I won't start that while this PR is open to avoid merge conflicts.

plugins/optimization-detective/performance-dashboard.js
```diff
@@ -0,0 +1,186 @@
+import { __ } from '@wordpress/i18n';
+import domReady from '@wordpress/dom-ready';
+import { createRoot, useEffect, useState } from '@wordpress/element';
+import {
+	// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
+	__experimentalGrid as Grid,
+	ComboboxControl,
+} from '@wordpress/components';
+import apiFetch from '@wordpress/api-fetch';
+import { Chart } from 'react-google-charts';
```
westonruter: I wonder if a chart is going to be the best visualization here? The "freshness TTL" for a URL metric by default is 1 day. So really all of the stored URL metrics, for a site that gets regular traffic, should not be older than a day since they would get continually refreshed. So I believe the graph would typically only show data for one day:

image

Capturing historical data hasn't been a goal for Optimization Detective.

Would it be better to have a visualization similar to what PSI has?

image

plugins/optimization-detective/storage/rest-api.php
swissspidy: Good idea! 😄

plugins/optimization-detective/performance-dashboard.js
swissspidy: I thought it would be useful to store data for 30 days or so, so you can see trends/changes. Otherwise yeah this would need some other visualization.

plugins/optimization-detective/storage/rest-api.php
swissspidy: I don‘t know how long this PR will be open as it‘s just a POC right now and things can still move around quite a bit.
A dedicated REST controller class makes sense though. Feel free to start a PR, I‘m happy to resolve any conflicts.

plugins/optimization-detective/performance-dashboard.js
westonruter: Also, something that just came to mind is that the admin bar could be populated with the current CWV field metrics for any page that the user is viewing on the front end. This could make discovery easier, although probably not in exclusion of having a dashboard too. Clicking the metrics in the admin bar could take them to the dashboard or the graphs could be shown in some dropdown from from the admin bar, similar to what Site Kit does.

swissspidy: Oh yeah, I was thinking of adding this in the admin bar, the post editor, maybe even in a dashboard widget. Lots of possibilities.

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -148,6 +155,34 @@ public static function get_json_schema(): array {
 						'additionalProperties' => false,
 					),
 				),
+				'webVitals'  => array(
```
felixarntz: @westonruter and I chatted about this yesterday. Assuming that, at least for now, we won't bake this functionality directly into Optimization Detective, it would be valuable to allow expanding the schema with additional properties, which would effectively allow to send arbitrary key value pairs based on the schema definition.

That way, a dashboard and underlying functionality like this could be implemented in a standalone plugin and still function by registering the additional supported data with the Optimization Detective API.

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -192,16 +228,28 @@ public function get_timestamp(): float {
...
 
+	/**
+	 * Gets web vitals.
+	 *
+	 * @return array Web Vitals.
+	 * @phpstan-return WebVitalData
+	 */
+	public function get_web_vitals(): array {
+		return $this->data['webVitals'];
+	}
```
felixarntz, line 265: Related: Instead of this, assuming the functionality would live in a standalone plugin, we could provide a simple `get( $key )` method that would allow retrieving the data. Or be "even more clever" and have a `__call()` method that automatically resolves it based on the schema, so that you could call `get_web_vitals()` which would resolve to returning `$this->data['webVitals']`. Maybe the latter is overkill, but one of the two could work.

plugins/optimization-detective/class-od-url-metric.php
westonruter: That being said, I think it does make sense to bake this in. If anything, there could be a way to opt-out of it. Since we're already using `onLCP` but just not capturing the LCP metric, it makes sense that we do. And if we do that, might as well capture INP and CLS as well. 

This doesn't preclude there also being a way to extend the schema with arbitrary key/value pairs for other plugins to send custom metrics.

felixarntz: I'd argue this shouldn't be built-in just like no other feature should be built-in, as the Optimization Detective plugin is just for the API.

It doesn't make sense to store those values if no feature is using them, so I think a separate plugin with a Web Vitals dashboard that relies on Optimization Detective to collect the data makes more sense.

swissspidy: Either approach has pros and cons, let‘s not focus on tha too much while this is still a POC. The main thing here is that the POC highlights missing pieces and extension points in the API, which is already a great outcome.

> It doesn't make sense to store those values if no feature is using them, so I think a separate plugin with a Web Vitals dashboard that relies on Optimization Detective to collect the data makes more sense.

With what I said above, that means OD would need a clear API for other plugins to collect additional metrics in JS and process that data on the server. Something to think about as this POC progresses.

plugins/optimization-detective/performance-dashboard.js
swissspidy: Dashboard widget POC:

Screenshot 2024-04-10 at 10 16 17

(Colors would change depending on score)

adamsilverstein:
@swissspidy I like the concept, but couldn't quite get this building and running locally to test.

Can you add instructions for building and testing?

To get the build even working, I had to rename the js files for .jsx to avoid a  `Unexpected token: operator (<)` from Terser when running npm run build:plugins, maybe I am missing a step? I tried collecting some data but never saw the dashboard load.

swissspidy:
That sounds odd 🤔  But since it's in very early stages, things can be a bit bumpy :)

I'll add more detailed instructions once it's more robust and less in flux.

During development, what I did is:

- Run `npm run start:plugin:optimization-detective` to build the JS files

- Add a small plugin to disable some safeguards (hat tip Weston):

```php
<?php
/**
 * Plugin Name: Optimization Detective Debugging
 * Version: 0.1.0
 */

add_filter( 'od_url_metric_storage_lock_ttl', '__return_zero' );
add_filter( 'od_url_metric_freshness_ttl', '__return_zero' );
add_filter( 'od_url_metrics_breakpoint_sample_size', function () { return 1; } );

add_filter(
    'register_od_url_metrics_post_type_args',
    static function( $args ) {
    
        $args['supports'][] = 'content';
        $args['public'] = true;

        return $args;
    }
);
```

- In a new session, visit a few pages, move around etc. so that `visibilitychange` is triggered and data is collected

westonruter:
I was thinking, in order to facilitate the historical data use case, perhaps it would make sense to fire an action whenever a URL metric is stored, like so:

```diff
--- a/plugins/optimization-detective/storage/rest-api.php
+++ b/plugins/optimization-detective/storage/rest-api.php
@@ -160,6 +160,16 @@ function od_handle_rest_request( WP_REST_Request $request ) {
 		return $result;
 	}
 
+	/**
+	 * Fires when a new URL metric has been stored.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param OD_URL_Metric   $url_metric Stored URL metric.
+	 * @param WP_REST_Request $request    REST request to store the URL metric.
+	 */
+	do_action( 'od_url_metric_stored', $url_metric, $request );
+
 	return new WP_REST_Response(
 		array(
 			'success' => true,
```

This could extract data from the URL metric or the storage request to then store in another location, like a custom table which is better suited for aggregating CWV metrics. This presumes that the CWV dashboard is eventually built in a dependent plugin and not in Optimization Detective itself.

swissspidy:
Interesting idea. Depending on what storage method we settle on, this could be a useful way to do that. 

Aside: For custom tables it's worth noting that some environments don't like them. They would definitely be a valid option for stats like this though. Just wanted to flag it :)

> This presumes that the CWV dashboard is eventually built in a dependent plugin and not in Optimization Detective itself.

Related discussion: https://github.com/WordPress/performance/pull/1098#discussion_r1550488934

We'd need to define the boundaries for data collection and storage, extension points for other plugins, etc.

joemcgill:
I just became aware of this demo and am really intrigued by the potential usefulness here. My initial thought was to agree with @felixarntz, that this probably shouldn't be a feature of the Optimization Detective plugin, but perhaps a feature in another plugin that makes use of the Optimization Detective API. 

Personally, I think this is the perfect kind of feature for Performance Lab, now that it's been repositioned from a collection of modules, to a central place to inform site owners of performance optimizations that they can make and surface performance features they can install via our standalone plugins.

swissspidy:
I started this before we did all the refactoring with plugin dependencies etc. So yeah this can now easily be its own plugin. I plan to pick it up again at some point 👍

joemcgill:
> So yeah this can now easily be its own plugin

I was suggesting that it become a feature of the Performance Lab plugin, rather than a separate plugin. One of the unfortunate side-effects of needing to unbundle the Performance Lab modules into standalone plugins (#656) is that it limits the usefulness of what is included in the main PL plugin. In initial discussions about this change, one of the ideas was to have PL be responsible for helping site owners understand the performance characteristics of their own site through the use of tools like the Server Timing headers and Site Health Checks, as well as be a central place to surface suggestions about performance features that they could install to improve their site. I think a performance dashboard like this fits directly into that vision and improve the usefulness of the main PL plugin.

swissspidy:
I see. It requires Optimization Detective though as a dependency, that might make things confusing.

felixarntz:
+1 to @joemcgill's suggestion of making this part of Performance Lab. Our self-imposed guidance when we made the split was that only features that actually perform optimizations should be their own plugins, while features around metrics, monitoring, and recommendations should be part of the Performance Lab plugin.

The Optimization Detective dependency concern by @swissspidy makes this maybe tricky, but I think we can find a solution for it (without actually making Optimization Detective a dependency for Performance Lab, which we definitely shouldn't do).

How about this:
* The dashboard only surfaces Server-Timing based data out of the box.
* If Optimization Detective is active, the dashboard also surfaces all the additional metrics that having Optimization Detective unlocks.
* If Optimization Detective is not active, we display a prompt that recommends the user to install and activate that. We can definitely include some reasons why, since of course the metrics it would unlock are much more powerful than Server-Timing metrics.

swissspidy:
Interesting suggestion. Moving it to the proposal issue for visibility.

plugins/optimization-detective/detect.js
```diff
@@ -364,43 +377,107 @@ export default async function detect( {
...
+			sendData();
+		}
+	} );
+
+	// NOTE: Safari does not reliably fire the `visibilitychange` event when the
+	// page is being unloaded. If Safari support is needed, you should also flush
+	// the queue in the `pagehide` event.
+	addEventListener( 'pagehide', () => {
+		sendData();
+	} );
```
westonruter, line 482: I've largely adopted this in https://github.com/WordPress/performance/pull/1373 now so that instead of sending the gathered URL metric as soon as the page has loaded and is idle, it instead sends it via `navigator.sendBeacon()` when the page is left. It just sends it once per page. The intention is that client-side extensions can hook in to finalize the URL metric object before it are sent, as can be seen in that PR currently for using a `ResizeObserver` to update the ultimate height of embeds rather than just the initial height.

plugins/optimization-detective/class-od-url-metric.php
westonruter: I'm adding URL metric schema extensibility in https://github.com/WordPress/performance/pull/1492

plugins/optimization-detective/class-od-url-metric.php
westonruter: Instead of `__call()` I added `__get()` in #1492. It seems a magic getter is better here. However, for static analysis neither `__call()` nor `__get()` would be helpful, so I added a `get($key)` method as well.

swissspidy:
I don't think I have the capacity to further work on this, so I am closing this one for now.

PR #1100, opened 2024-03-28
Clean up outdated module references

felixarntz:
## Summary

This PR removes or replaces all remaining "module" references that are no longer accurate. The only exceptions are changelog entries and documentation related to migrating away from modules.

* Replaces the word "module" with "plugin" or, in a few places, another more accurate replacement.
* Removes WebP Uploads specific FAQ items from the main Performance Lab `readme.txt`.
* Renames the `tests/testdata/modules` directory to `tests/testdata/plugins`.
* Removes a few module specific references that are no longer relevant.

This PR does not change _any_ functionality.

PR #1101, opened 2024-03-28
Update plugin names and descriptions to better describe their features

### LINKED ISSUE #1046: Review current plugin names and their descriptions for their accuracy and helpfulness

felixarntz:
## Summary

Fixes #1046 (note that the prefixing discussion and implementation should be part of a separate issue)

This is a sub PR of #1100. Please merge that one first.

## Relevant technical choices

* Applies plugin names and descriptions from https://github.com/WordPress/performance/issues/1046#issuecomment-2005134234.
* Bumps minor version as needed since the name change is more significant than a patch version increase IMO.
* Adds changelog entries about the name change.


plugins/dominant-color-images/readme.txt
```diff
@@ -1,15 +1,15 @@
...
-Stable tag:        1.0.2
+Stable tag:        1.1.0
 License:           GPLv2 or later
 License URI:       https://www.gnu.org/licenses/gpl-2.0.html
 Tags:              performance, images, dominant color
 
-Adds support to store the dominant color of newly uploaded images and create a placeholder background of that color.
+Displays placeholders based on an image's dominant color while the image is loading.
 
 == Description ==
```
westonruter, line 14: Perhaps the first line of the description should say:

> _This plugin was formerly known as Dominant Color Images._

plugins/speculation-rules/readme.txt
```diff
@@ -9,15 +9,15 @@ License:           GPLv2 or later
 License URI:       https://www.gnu.org/licenses/gpl-2.0.html
 Tags:              performance, javascript, speculation rules, prerender, prefetch
 
-Uses the Speculation Rules API to prerender linked URLs upon hover by default.
+Enables browsers to speculatively prerender or prefetch pages when hovering over links.
 
 == Description ==
```
westonruter, line 14: Ditto above about mentioning the old name.

plugins/webp-uploads/readme.txt
```diff
@@ -1,15 +1,15 @@
...
-Stable tag:        1.0.6
+Stable tag:        1.1.0
 License:           GPLv2 or later
 License URI:       https://www.gnu.org/licenses/gpl-2.0.html
 Tags:              performance, images, webp
 
-Creates WebP versions for new JPEG image uploads if supported by the server.
+Converts images to more modern formats such as WebP or AVIF during upload.
 
 == Description ==
```
westonruter, line 14: Ditto above about mentioning the old name.

plugins/dominant-color-images/readme.txt
felixarntz: @westonruter How about putting those explanations at the end of the respective `== Description ==` section?

westonruter: Sounds good

felixarntz: Done in https://github.com/WordPress/performance/pull/1101/commits/0c04a92f8cdf1eb4067a0ca11f2c0b92ee4a7b24

plugins/speculation-rules/readme.txt
felixarntz: Done in https://github.com/WordPress/performance/pull/1101/commits/0c04a92f8cdf1eb4067a0ca11f2c0b92ee4a7b24

plugins/webp-uploads/readme.txt
felixarntz: Done in https://github.com/WordPress/performance/pull/1101/commits/0c04a92f8cdf1eb4067a0ca11f2c0b92ee4a7b24

westonruter (approved):

thelovekesh (approved):

PR #1102, opened 2024-03-28
Add Speculative Loading generator tag

felixarntz:
## Summary

Adds a generator tag to the Speculative Loading plugin, similar to how the other standalone plugins include one.

## Relevant technical choices

* Uses the new plugin name per #1046 / #1101

### DIFF
--- a/plugins/dominant-color-images/hooks.php
+++ b/plugins/dominant-color-images/hooks.php
@@ -178,11 +178,6 @@ add_filter( 'wp_enqueue_scripts', 'dominant_color_add_inline_style' );
  * @since 1.0.0
  */
 function dominant_color_render_generator() {
-	if (
-		defined( 'DOMINANT_COLOR_IMAGES_VERSION' ) &&
-		! str_starts_with( DOMINANT_COLOR_IMAGES_VERSION, 'Performance Lab ' )
-	) {
-		echo '<meta name="generator" content="Dominant Color Images ' . esc_attr( DOMINANT_COLOR_IMAGES_VERSION ) . '">' . "\n";
-	}
+	echo '<meta name="generator" content="Dominant Color Images ' . esc_attr( DOMINANT_COLOR_IMAGES_VERSION ) . '">' . "\n";
 }
 add_action( 'wp_head', 'dominant_color_render_generator' );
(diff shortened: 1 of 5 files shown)


westonruter (approved):

westonruter:
Maybe take this as an opportunity to align the other functions with this one? Namely, the constant checks seem to be a vestige of the module past.

https://github.com/WordPress/performance/blob/71055b340907330f5f07c0e501a0959f4bdd9f1f/plugins/dominant-color-images/hooks.php#L180-L187

https://github.com/WordPress/performance/blob/71055b340907330f5f07c0e501a0959f4bdd9f1f/plugins/webp-uploads/hooks.php#L772-L778

https://github.com/WordPress/performance/blob/71055b340907330f5f07c0e501a0959f4bdd9f1f/plugins/embed-optimizer/hooks.php#L164-L170

Aside: It seems Auto Sizes is missing a generator meta tag also.

felixarntz:
@westonruter 
> Maybe take this as an opportunity to align the other functions with this one? Namely, the constant checks seem to be a vestige of the module past.

Done in https://github.com/WordPress/performance/pull/1102/commits/c6252cf99935fe34cadbe3233738b6bb6c3d771e

> Aside: It seems Auto Sizes is missing a generator meta tag also.

Let's do that in a separate PR.

Another idea: Should we use the plugin slug instead of the plugin name for all those individual generator tags? That would be more in line with the main Performance Lab generator tag, and furthermore would avoid having to make a trade-off between using outdated plugin names vs the maintenance burden of having to adjust tooling every time we may change the plugin name.

Since those generator tags are still new and I assume aren't used much, it would probably be fine to change them once at this point to be the slug, and after that they'll be "stable" forever.

westonruter:
Good idea! Let's use slugs. 🐌

westonruter:
> Let's do that in a separate PR.

Why? You're already in here modifying other plugins' generator meta tags. Will reduce noise to just do it all here.

felixarntz:
> Good idea! Let's use slugs. 🐌

I'll do that in a separate follow up PR, as it's a distinct change affecting several plugins in a way that I'd put it in the changelog.

> Why? You're already in here modifying other plugins' generator meta tags. Will reduce noise to just do it all here.

Similar to the above, I prefer to do this separately as it would touch another plugin and should be mentioned in the changelog. It's quick anyway.

thelovekesh (approved):

mukeshpanchal27 (approved):

mukeshpanchal27:
I would also like to do the follow-up work on a separate PR.

PR #1103, opened 2024-03-29
Use plugin slug for generator tag

felixarntz:
## Summary

Based on the discussion in https://github.com/WordPress/performance/pull/1102#issuecomment-2026401373: Since plugin names can change over time, the original approach of using the plugin names for generator tags is not great. Therefore this PR proposes using the plugin slug, which is immutable. In the long run, this will be more maintainable as it won't result in tooling or queries requiring an update.

## Relevant technical choices

* Use the plugin slug instead of plugin name for all existing generator tags, including Performance Lab.
* Update existing tests, and add a missing test for the Speculation Rules plugin's generator tag.
* Add readme entries for the standalone plugins (for PL this will be auto-generated, while for the other standalone plugins we'll start with that after the next "set of releases").
* Bump Optimization Detective to 0.1.1 to include this change.

### DIFF
--- a/plugins/optimization-detective/readme.txt
+++ b/plugins/optimization-detective/readme.txt
@@ -4,7 +4,7 @@ Contributors:      wordpressdotorg
 Requires at least: 6.4
 Tested up to:      6.5
 Requires PHP:      7.0
-Stable tag:        0.1.0
+Stable tag:        0.1.1
 License:           GPLv2 or later
 License URI:       https://www.gnu.org/licenses/gpl-2.0.html
 Tags:              performance, images
@@ -136,6 +136,10 @@ Contributions are always welcome! Learn more about how to get involved in the [C
 The [plugin source code](https://github.com/WordPress/performance/tree/trunk/plugins/optimization-detective) is located in the [WordPress/performance](https://github.com/WordPress/performance) repo on GitHub.
 
 == Changelog ==
+
+= 0.1.1 =
+
+* Use plugin slug for generator tag. ([1103](https://github.com/WordPress/performance/pull/1103))
 
 = 0.1.0 =
 
--- a/plugins/speculation-rules/hooks.php
+++ b/plugins/speculation-rules/hooks.php
@@ -54,6 +54,7 @@ add_action( 'wp_footer', 'plsr_print_speculation_rules' );
  * @since 1.1.0
  */
 function plsr_render_generator_meta_tag() {
-	echo '<meta name="generator" content="Speculative Loading ' . esc_attr( SPECULATION_RULES_VERSION ) . '">' . "\n";
+	// Use the plugin slug as it is immutable.
+	echo '<meta name="generator" content="speculation-rules ' . esc_attr( SPECULATION_RULES_VERSION ) . '">' . "\n";
 }
 add_action( 'wp_head', 'plsr_render_generator_meta_tag' );
--- a/tests/plugins/speculation-rules/speculation-rules-test.php
+++ b/tests/plugins/speculation-rules/speculation-rules-test.php
@@ -19,6 +19,23 @@ class Speculation_Rules_Tests extends WP_UnitTestCase {
 		parent::tear_down();
 	}
 
+	public function test_hooks() {
+		$this->assertSame( 10, has_action( 'wp_footer', 'plsr_print_speculation_rules' ) );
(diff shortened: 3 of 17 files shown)


plugins/optimization-detective/readme.txt
```diff
@@ -137,6 +137,10 @@ The [plugin source code](https://github.com/WordPress/performance/tree/trunk/plu
 
 == Changelog ==
 
+= 0.1.1 =
+
+* Use plugin slug for generator tag. ([1103](https://github.com/WordPress/performance/pull/1103))
```
mukeshpanchal27, line 142:
```suggestion
* Use plugin slug for generator tag. ([1103](https://github.com/WordPress/performance/pull/1103))
* Bump minimum required WP version to 6.4. ([1078](https://github.com/WordPress/performance/pull/1078))
```

mukeshpanchal27 (approved):

adamsilverstein:
+1 to making this change, makes sense for the long run. One issue worth noting that will arise from this change will be that any queries against a feature will need to take both old and new slugs into account to work properly.

plugins/speculation-rules/hooks.php
```diff
@@ -54,6 +54,7 @@ function plsr_print_speculation_rules() {
  * @since 1.1.0
  */
 function plsr_render_generator_meta_tag() {
-	echo '<meta name="generator" content="Speculative Loading ' . esc_attr( SPECULATION_RULES_VERSION ) . '">' . "\n";
+	// Use the plugin slug as it is immutable.
+	echo '<meta name="generator" content="speculation-rules ' . esc_attr( SPECULATION_RULES_VERSION ) . '">' . "\n";
```
adamsilverstein, line 58: probably want to bump the plugin version as well, otherwise users won't get this update.

joemcgill (commented):
Should we take this opportunity to add generator tags to the auto-sizes plugin as well? We should also make sure that all plugins have tests in place to cover this behavior.

plugins/speculation-rules/hooks.php
felixarntz: @adamsilverstein That version is already bumped (and not published), so this PR doesn't need to bump it again. :)

felixarntz:
@adamsilverstein 
> One issue worth noting that will arise from this change will be that any queries against a feature will need to take both old and new slugs into account to work properly.

Slugs never change, so that's the main reason for why this change is proposed. Only now this one time we would have to update tooling, to switch from the plugin names to plugin slugs.

@joemcgill 
> Should we take this opportunity to add generator tags to the auto-sizes plugin as well?

I was going to do that in a separate PR.

> We should also make sure that all plugins have tests in place to cover this behavior.

+1, let me check that every plugin is covered by such a test.

plugins/optimization-detective/readme.txt
felixarntz: @mukeshpanchal27 I don't think this is relevant here as the plugin was already released with that bump included, see https://wordpress.org/plugins/optimization-detective/.

felixarntz:
@joemcgill Tests for all generator tags and that they include the slugs are now present.

tests/plugins/speculation-rules/speculation-rules-test.php
```diff
@@ -19,6 +19,11 @@ public function tear_down() {
 		parent::tear_down();
 	}
 
+	public function test_hooks() {
```
joemcgill, line 22: Good idea. I don't think we are validating that hooks are always present for each of the plugins. Should this type of test be added everywhere, since I don't believe we can run tests against the output of `wp_head` to confirm that these are actually being printed without introducing side effects?

joemcgill (approved):
Thanks @felixarntz. One follow-up question, but this is looking good.

tests/plugins/speculation-rules/speculation-rules-test.php
felixarntz: @joemcgill I generally like to have this kind of test as it helps prevent accidents with hooks added or removed, but I wouldn't want to mandate it. IMO it's a good practice, but I still think this is something we can leave up to the typical freedom of choice of the respective plugin developers / PR authors and PR reviewers.

I only added this here since the hook check previously was baked into the generator test method, which IMO it shouldn't be since it's a different kind of test.

PR #1104, opened 2024-04-01
Avoid needless array allocation in rgb to hex conversion

ju1ius:
The current implementation of the `dominant_color_rgb_to_hex` function validates the RGB color by:
1. allocating an array of 256 64-bit integers
2. performing three linear scans of the array (one for each color channel)

Which is not a very performant way to check if a variable is an integer in the range `[0..255]`.

This PR replaces the current implementation by simple type checks and integer comparisons.


plugins/dominant-color-images/helper.php
```diff
@@ -172,8 +172,12 @@ function dominant_color_has_transparency( $attachment_id ) {
...
  */
 function dominant_color_rgb_to_hex( $red, $green, $blue ) {
-	$range = range( 0, 255 );
-	if ( ! in_array( $red, $range, true ) || ! in_array( $green, $range, true ) || ! in_array( $blue, $range, true ) ) {
+	if ( ! (
+		\is_int( $red ) && \is_int( $green ) && \is_int( $blue )
+		&& $red >= 0 && $red <= 255
+		&& $green >= 0 && $green <= 255
+		&& $blue >= 0 && $blue <= 255
+	) ) {
```
felixarntz, line 191: This could be further simplified by not having the entire clause negated.
```suggestion
	if (
		! is_int( $red ) || $red < 0 || $red > 255
		|| ! is_int( $green ) || $green < 0 || $green > 255
		|| ! is_int( $blue ) || $blue < 0 || $blue > 255
	) {
```

felixarntz (commented):
@ju1ius Thank you for the PR, great catch!

I have one suggestion on how your logic could be further simplified, mostly easier to understand when reading the code.

plugins/dominant-color-images/helper.php
ju1ius: This feels weird to me.

```php
// One would normally write A.
if (!isValid($r, $b, $b)) return null;
else return toHex($r, $g, $b);
// or B.
if (isValid($r, $b, $b)) return toHex($r, $g, $b);
else return null;
// but not C.
if (isInvalid($r, $b, $b)) return null;
else return toHex($r, $g, $b);
// and certainly not D.
if (!isInvalid($r, $b, $b)) return toHex($r, $g, $b);
else return null;
```
If you really want to remove the negation, the most readable IMO would be option B:
```php
if (
    \is_int($r) && $r >= 0 && $r <= 255
    && \is_int($g) && $g >= 0 && $g <= 255
    && \is_int($b) && $b >= 0 && $b <= 255
) {
    return sprintf('%02x%02x%02x', $r, $g, $b);
}

return null;
```

felixarntz: I would agree with you if we had a function like `isValid` to call here, but since we don't have one, I think we shouldn't reason about readability / understandability based on a hypothetical function. Your option B would be okay from a readability perspective, but then it would no longer use an early return, which is a good convention to avoid a lot of nesting of if clauses.

So IMO it makes sense to either go with my suggestion, or introduce a separate function like your `isValid` that we can call here to simplify readability even more.

westonruter: @ju1ius Will you be adding a function like `dominant_color_is_valid_rgb()`?

ju1ius: Hi, @westonruter.

The `dominant_color_rgb_to_hex` function currently has two call sites and in both of them the input is guaranteed to be valid RGB, so the validation steps are useless.

In the [Imagick editor case](https://github.com/WordPress/performance/blob/603610d03a93a06451d6fc593b6314bb8424901e/plugins/dominant-color-images/class-dominant-color-image-editor-imagick.php#L37), the [ImagickPixel::getColor](https://www.php.net/manual/en/imagickpixel.getcolor.php) method guarantees that the `r`, `g` and `b` keys of the returned array are integers in the [0, 255] range.

In the [GD editor case](https://github.com/WordPress/performance/blob/603610d03a93a06451d6fc593b6314bb8424901e/plugins/dominant-color-images/class-dominant-color-image-editor-gd.php#L37), the [imagecolorat](https://www.php.net/manual/function.imagecolorat.php) function returns `int|false`. The PHP type coercion rules and the following bit operations therefore guarantee that the `$r`, `$g` and `$b` variables are integers in the [0, 255] range.

Therefore my suggestion would be to just inline the `sprintf` calls, remove the `dominant_color_rgb_to_hex` function to reduce API surface, and optionally fix the error checking in the GD editor so that a `WP_Error` is returned instead of `"000000"`.

westonruter: Thanks for that. In 31fd033d129b49cd558d34f9ae8cc3071667d503 I've added an explicit check to ensure that we only use the `imagecolorat()` return value if it is an `int`.

Also, note that in #1203 the use of PHP type hints is introduced so that this code can be simplified to:

```php
	if ( ! (
		$red >= 0 && $red <= 255 && 
		$green >= 0 && $green <= 255 &&
		$blue >= 0 && $blue <= 255
	) ) {
```

westonruter (approved):

plugins/dominant-color-images/helper.php
westonruter: I've applied that simplification with fixing the merge conflict for that PR: 7e85330e7a04b19e3256d3ad1fb51798e3ac717d

PR #1106, opened 2024-04-01
Add crate-ci/typos config file with false positives.

felixarntz:
## Summary

It looks like our Spell Check underlying repository was updated and now complains about the word "dur", which however is intentional, as it abbreviates "duration" and is in fact a syntax requirement for the Server-Timing API. See failure here for reference: https://github.com/WordPress/performance/actions/runs/8513515250/job/23317345691?pr=1105

This fix is based on the documentation in https://github.com/crate-ci/typos?tab=readme-ov-file#false-positives

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,0 +1,3 @@
+[default.extend-words]
+# Don't correct the abbreviated "duration"
+dur = "dur"


joemcgill (approved):

mukeshpanchal27 (approved):

_typos.toml
thelovekesh, line 1: @felixarntz I think this file should have been placed under `.github` since it's being used in CI at the moment.

felixarntz: @thelovekesh Does the CI workflow automatically pick it up too if it's in the `.github` directory instead of the root directory?

thelovekesh: @felixarntz I think we can use `config` argument:

```diff
diff --git a/.github/workflows/spell-check.yml b/.github/workflows/spell-check.yml
index 7fe8b067..cd468152 100644
--- a/.github/workflows/spell-check.yml
+++ b/.github/workflows/spell-check.yml
@@ -10,3 +10,5 @@ jobs:
       - uses: actions/checkout@v3
       - name: Search for misspellings
         uses: crate-ci/typos@master
+        with:
+          config: ./.github/typos.toml
```

westonruter: Good. As at the moment this file is incorrectly getting included with the build since it isn't being ignored for export/dist, right?

thelovekesh: Oh yes. But moving it to `.github` will fix that.

felixarntz: Feel free to open a follow up PR. Thanks!

PR #1111, opened 2024-04-04
Implement support for annotating certain plugins as experimental

### LINKED ISSUE #1045: Allow marking certain standalone plugins as experimental
Follow up to #1031: Similar to how modules can be marked as experimental, we should maintain the ability to have certain standalone plugins marked as experimental in the Performance Lab UI. This indicates that a certain plugin may be in an earlier phase and is just starting out, or it contains "riskier" features that we are for instance less confident about being used in production. While Performance Lab as a whole is sort of a Beta plugin, some of our standalone plugins are "more beta than others".

As with modules, of course any plugin can and should eventually graduate to non-experimental.

The primary reason to continue indicating and supporting the distinction between experimental and non-experimental modules, other than end user awareness, is the onboarding flow (see #1032). One of its requirements is to have a one-click button to activate all "recommended" features - which effectively means activating all "non-experimental" features.

For this issue, let's implement the following (after #1031 is done):
* Additionally to the standalone plugins' slug and version constants hard-coded in the Performance Lab codebase, each standalone plugin should receive a boolean indicator for whether it's experimental or not.
* Of the current 5 standalone plugins, only the `auto-sizes` plugin should be marked as experimental. This is mostly because there's currently no browser where its functionality is actually supported (given Chrome recently rolled back the feature).
* The feature cards for the plugins on the PL settings screen should show a note "experimental" for those plugins which are experimental. Experimental plugins should be displayed after the non-experimental ones.

felixarntz:
## Summary

Fixes #1045

## Relevant technical choices

* Introduces a new `perflab_get_standalone_plugin_data()` function so that all standalone plugin data can be defined in one place.
    * I'm sure we wouldn't want to have to maintain the plugin list in two places, e.g. one for whether each plugin is experimental or not, and another one for each plugin's constant.
* Marks `auto-sizes` and `embed-optimizer` as experimental for now.
    * `auto-sizes` because its browser behavior isn't really available yet, `embed-optimizer` because it's very new and a `0.x` version.
* Displays an "(experimental)" annotation on the feature card behind every plugin that is marked experimental.
* Displays the experimental plugins after all the non-experimental plugins.

See screenshot of what this would look like based on the PR implementation:
Screenshot 2024-04-03 at 5 57 44 PM


felixarntz:
**Note for reviewers:** I have opened this against `release/3.0.0` since it's a technically relatively simple change which IMO would be great to include if we can get it across the finish line (we're still 1.5 weeks away from that release).

If there's substantial feedback or further discussion required that means this can't land in 3.0.0, we can adjust the base branch and change it to target 3.1.0.

felixarntz (commented):
Another note: The CI failures in https://github.com/WordPress/performance/actions/runs/8547642712/job/23420147133?pr=1111 are unrelated, so we should fix them separately. See #1012.

mukeshpanchal27 (approved):
Thanks @felixarntz for the PR. It look good to me.

mukeshpanchal27:
> Another note: The CI failures in https://github.com/WordPress/performance/actions/runs/8547642712/job/23420147133?pr=1111 are unrelated, so we should fix them separately. See #1012.

The PR is #1112 not 1012

swissspidy:
I was hoping we can just mark them as experimental based on the version number (i.e. < 1.0.0 is experimental), but turns out Auto Sizes is already at 1.0.0 🙃

felixarntz:
@swissspidy 
> I was hoping we can just mark them as experimental based on the version number (i.e. < 1.0.0 is experimental), but turns out Auto Sizes is already at 1.0.0 🙃

Yeah, we probably need an extra flag. It's also worth considering that in certain edge-cases we may even want to designate plugins as experimental again when they weren't before. Potentially this could apply to Dominant Color Images / Image Placeholders as right now nobody is really exploring a path forward or maintaining it, and it received a lot of concerns/pushback when initially proposed in a Make Core post. Just thinking out loud, I'm not saying we should do that. But probably a flag is helpful for several reasons.

Could you (or someone else) please give another review to this PR? Since #1112 was now merged into this, the CI failures are resolved, so this is good for a review.

load.php
```diff
@@ -79,29 +79,56 @@ function perflab_render_generator() {
...
+ * Gets the standalone plugins recommended by Performance Lab and their data.
  *
- * @since 2.9.0
- * @since n.e.x.t The $source parameter was removed.
+ * @since n.e.x.t
  *
- * @return array<string, string> Map of plugin slug and the version constant used.
+ * @return array Associative array of $plugin_slug => $plugin_data pairs.
  */
-function perflab_get_standalone_plugin_version_constants() {
```
felixarntz, line 89: Note for reviewers: This function is simply moved down (see bottom of the file), it's not being replaced as potentially the diff here may suggest.

load.php
```diff
@@ -79,29 +79,56 @@ function perflab_render_generator() {
...
 /**
- * Gets the standalone plugin constants used for each available standalone plugin.
+ * Gets the standalone plugins and their data.
  *
- * @since 2.9.0
- * @since n.e.x.t The $source parameter was removed.
+ * @since n.e.x.t
  *
- * @return array<string, string> Map of plugin slug and the version constant used.
+ * @return array Associative array of $plugin_slug => $plugin_data pairs.
```
westonruter:
```suggestion
 * @return array<string, array{'constant': string, 'experimental'?: bool}> Associative array of $plugin_slug => $plugin_data pairs.
```

westonruter (approved):
LGTM but with a suggestion to formalize the description of a return value with an array shape for static analysis.

PR #1112, opened 2024-04-04
Add support for autoloading enhancements in WordPress 6.6 trunk to autoloaded options Site Health check

felixarntz:
## Summary

This PR fixes the test failures seen in https://github.com/WordPress/performance/actions/runs/8547642712/job/23420147133?pr=1111, which are unrelated to the PR they're reported on. The next time we'll make a PHP change against `trunk`, we should see the same failures too.

## Relevant technical choices

The failures happen due to https://core.trac.wordpress.org/changeset/57920 which makes an intentional change regarding large autoloaded options, to no longer autoload them unless `$autoload = true` is explicitly provided. Additionally, there are now more database values than `yes` that should trigger autoloading. All of this is expected. Plugins that perform low-level optimizations around autoloaded options are likely to run into similar failures - all of which is straightforward to fix though. This is why the underlying Trac ticket is annotated with `needs-dev-note`.

So we need to do two things:
* In the production code, update our low-level database queries to take into account the new values that should result in options being autoloaded, via the new core function `wp_autoload_values_to_autoload()` that exists for this purpose.
* In the tests, explicitly force the excessively large options in the tests to be autoloaded, otherwise they obviously won't get picked up as autoloaded options.


felixarntz:
Once this fix is merged into `release/3.0.0`, we should also merge it into `trunk` to avoid failures in that branch.

includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -101,7 +101,22 @@ function perflab_aao_autoloaded_options_test() {
...
+	if ( function_exists( 'wp_autoload_values_to_autoload' ) ) {
+		$autoload_values = wp_autoload_values_to_autoload();
+	} else {
+		$autoload_values = array( 'yes' );
+	}
+
+	return (int) $wpdb->get_var(
+		$wpdb->prepare(
+			sprintf(
+				'SELECT SUM(LENGTH(option_value)) FROM ' . $wpdb->prefix . 'options WHERE autoload IN (%s)',
```
mukeshpanchal27:
```suggestion
				"SELECT SUM(LENGTH(option_value)) FROM $wpdb->options WHERE autoload IN (%s)",
```

tests/includes/site-health/audit-autoloaded-options/audit-autoloaded-options-test.php
```diff
@@ -59,7 +59,22 @@ public function test_perflab_aao_autoloaded_options_test_warning() {
...
+		if ( function_exists( 'wp_autoload_values_to_autoload' ) ) {
+			$autoload_values = wp_autoload_values_to_autoload();
+		} else {
+			$autoload_values = array( 'yes' );
+		}
+
+		$autoloaded_options_size = $wpdb->get_var(
+			$wpdb->prepare(
+				sprintf(
+					'SELECT SUM(LENGTH(option_value)) FROM ' . $wpdb->prefix . 'options WHERE autoload IN (%s)',
```
mukeshpanchal27:
```suggestion
					"SELECT SUM(LENGTH(option_value)) FROM $wpdb->options WHERE autoload IN (%s)",
```

mukeshpanchal27 (approved):
Overall look good to me. Left some nit-pick

includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -101,7 +101,22 @@ function perflab_aao_autoloaded_options_test() {
  */
 function perflab_aao_autoloaded_options_size() {
 	global $wpdb;
-	return (int) $wpdb->get_var( 'SELECT SUM(LENGTH(option_value)) FROM ' . $wpdb->prefix . 'options WHERE autoload = \'yes\'' );
+
+	if ( function_exists( 'wp_autoload_values_to_autoload' ) ) {
+		$autoload_values = wp_autoload_values_to_autoload();
+	} else {
+		$autoload_values = array( 'yes' );
+	}
```
joemcgill: Rather than repeating this several times, how about consolidating it to our own `perflab_aao_get_autoload_values()` helper?

joemcgill (approved):
Just one suggestion but not a blocker.

includes/site-health/audit-autoloaded-options/helper.php
felixarntz: @joemcgill Great idea, done in https://github.com/WordPress/performance/pull/1112/commits/031658609d93296d6eb556f086a78eba92d879e2

PR #1113, opened 2024-04-04
Merge `release/3.0.0` additions into `trunk` to unblock

felixarntz:
This is simply a merge of #1111 and #1112 from the 3.0.0 release branch into `trunk`.

Normally we would only do this after the 3.0.0 release, but because #1112 avoids CI failures in `trunk` and #1111 unblocks working on #1032, it makes sense to already merge these two changes back at this point.


westonruter (approved):

joemcgill (approved):
This seems fine to me, though we'll need to make sure that the version bumps that get made to update the `@since` values also get merged back to trunk after the release. This should happen automatically if the [release steps in the handbook](https://make.wordpress.org/performance/handbook/performance-lab/releasing-the-plugin/) are followed, but want to make sure we're keeping an eye on it.

PR #1115, opened 2024-04-06
Reuse `WP_HTML_Processor::is_void()` now that WP 6.5 is minimum supported version

westonruter:
This removes a TODO to reuse `WP_HTML_Processor::is_void()` instead of replicating that functionality which was not available until WP 6.4. Now that 6.4 is the minimum-supported version, that method can be reused.

This could have been done previously in #1068 when the minimum version was bumped.

### DIFF
--- a/plugins/optimization-detective/class-od-html-tag-processor.php
+++ b/plugins/optimization-detective/class-od-html-tag-processor.php
@@ -22,36 +22,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 final class OD_HTML_Tag_Processor {
 
 	/**
-	 * HTML void tags (i.e. those which are self-closing).
-	 *
-	 * @link https://html.spec.whatwg.org/multipage/syntax.html#void-elements
-	 * @see WP_HTML_Processor::is_void()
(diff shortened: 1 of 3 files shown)


felixarntz (approved):
Good catch @westonruter!

felixarntz (changes requested):
@westonruter Actually, it looks like https://github.com/WordPress/performance/actions/runs/8583685696/job/23573872147?pr=1115 has failures related to this change. Can you check?

westonruter:
@felixarntz Ah, this may have been an oversight in core that certain void tags were missed. I've opened Core-60961 to discuss that. In the meantime, probably better to let this PR sit.

felixarntz:
@westonruter SG. We should probably set the base to `trunk` then.

westonruter:
Oh, and actually `WP_HTML_Processor::is_void()` _was_ updated in 6.5 to include these deprecated void tags. So it's just that the version in 6.4 doesn't include them. So unless the change is backported to 6.4, this PR will have to wait until 6.5 is the minimum-supported version.

westonruter:
This was done in #1333.

PR #1116, opened 2024-04-08
Delete option when uninstalling the Modern Image Formats plugin

mukeshpanchal27:
## Summary
We removed the module infrastructure in #1060 from the Performance Lab plugin. In #345, we introduced an uninstaller routine that removes options when the plugin is uninstalled from the site.

We should move the `uninstall.php` file to the WebP Uploads folder, as that plugin adds that option.

### DIFF
--- a/plugins/webp-uploads/uninstall.php
+++ b/plugins/webp-uploads/uninstall.php
@@ -2,8 +2,8 @@
 /**
  * Plugin uninstaller logic.
  *
- * @package performance-lab
- * @since 1.2.0
+ * @package webp-uploads
+ * @since n.e.x.t
  */
 
 // If uninstall.php is not called by WordPress, bail.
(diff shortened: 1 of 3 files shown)


mukeshpanchal27:
For now, I've moved this PR to the `3.0` milestone. If anyone has concerns, they can update the base branch to `trunk` for future releases.

swissspidy (approved):

thelovekesh (approved):

plugins/webp-uploads/uninstall.php
```diff
@@ -2,8 +2,8 @@
 /**
  * Plugin uninstaller logic.
  *
- * @package performance-lab
- * @since 1.2.0
+ * @package webp-uploads
+ * @since n.e.x.t
```
felixarntz: The `n.e.x.t` replacements script at the moment still only works for the main PL plugin, so for the other plugins so far we have to handle it ourselves until we have improved the script.
```suggestion
 * @since 1.1.0
```

plugins/webp-uploads/uninstall.php
```diff
@@ -24,18 +24,18 @@
...
 }
 
-perflab_delete_plugin_option();
+webp_uploads_delete_plugin_option();
 
 /**
  * Delete the current site's option.
  *
- * @since 1.4.0
+ * @since n.e.x.t
```
felixarntz:
```suggestion
 * @since 1.1.0
```

felixarntz (commented):
@mukeshpanchal27 Great catch!

Note that the WebP Uploads 1.1.0 release will be published next Monday, so let's add it to the corresponding milestone, especially since you also implemented this against the release branch.

felixarntz (commented):
@mukeshpanchal27 Can you please also add an entry for this to the WebP Upload's `readme.txt`? This cycle is the last time that we're adding those manually, afterwards we can rely on the new script to handle it for us.

westonruter (approved):

PR #1119, opened 2024-04-09
Update `since` command for standalone plugin usage

### LINKED ISSUE #1118: Enhance `npm run since` command to support standalone plugins
As a follow up to #1064 and #1075 which made the readme changelog NPM script reusable for the standalone plugins as well, this issue is about doing the same for the "since" NPM script, which replaces `n.e.x.t` placeholders with actual version numbers.

So far the script only supports Performance Lab plugin and requires a `--release (-r)` argument. We should add a `--plugin (-p)` argument to it so that a specific plugin slug can be provided. This would then instruct the command to make the replacements only for the specific plugin. For a smooth transition / BC we should use `performance-lab` as the default for that new argument.

It's worth noting that this unlocks a slightly better workflow experience for publishing plugins. One of the current problems is that when you publish a Performance Lab release, you also publish any pending standalone plugin releases, and it's difficult to figure out whether all plugins that have bumped versions are actually ready because all of this is managed completely manually and intertwined with specific enhancement/bug PRs. While #1061 is the main issue about shaping the future goal, this is unrelated and merely improving the current status quo without actually changing the way we do releases. So it makes sense to be a standalone enhancement.

Here's what having standalone plugin support in our "readme" and "since" commands unlocks:
- Instead of hard-coding specific since annotations when working on enhancement/bug PRs for standalone plugins, we should simply use "n.e.x.t", similar to Performance Lab.
- Similarly, we should no longer bump any standalone plugin versions as part of another enhancement/bug PR.
- Instead, whenever maintainers decide that a specific standalone plugin is ready for release, they would need to do the following:
    - Change the relevant GitHub milestone from using "n.e.x.t" to using the actual version number decided.
    - Open a "release preparation" PR that simply bumps the versions in the plugin main file (header and constant) and readme (header), replaces the plugin's `n.e.x.t` code placeholders with the correct version (via the "since" NPM script) and populates the readme changelog for that version/milestone (via the "readme" NPM script). This is very similar to what https://make.wordpress.org/performance/handbook/performance-lab/releasing-the-plugin/#preparing-the-release describes for the Performance Lab plugin.

This workflow would notably simplify the current overhead as it would avoid:
- having to manually include those things in specific enhancement/bug PRs
- having to search through the codebase to keep track of which plugins are ready for release (to not accidentally release unfinished standalone plugin updates when releasing a new Performance Lab version)

Going forward, it should be straightforward to know whether a standalone plugin is ready for release by checking whether its latest milestone already has a version number, since the version bumps in the code happen at a similar time. And even if the latter PR is not merged yet, it wouldn't be a problem because the code simply wouldn't be deployed, and it could be done later by dispatching the workflow.

Once the Performance Lab 3.0.0 release has gone out next Monday, I propose we go with the above workflow improvement. It will make standalone plugin releases a more intentional action without changing any technical infrastructure (beyond the "since" script enhancement in this issue). We should then update the relevant documentation [here](https://make.wordpress.org/performance/handbook/performance-lab/releasing-a-standalone-plugin/) and [here](https://make.wordpress.org/performance/handbook/performance-lab/releasing-the-plugin/) to include those preparation steps.

cc @westonruter @mukeshpanchal27 @thelovekesh @joemcgill

thelovekesh:
## Summary

Fixes #1118

## Relevant technical choices

- Add `--plugin` flag to pass the specific plugin name.
- If no `--plugin` flag is passed, it will run on PL plugin.
- Return with error if `--plugin` is passed with wrong plugin name.

### DIFF
--- a/bin/plugin/cli.js
+++ b/bin/plugin/cli.js
@@ -5,6 +5,11 @@
  */
 const program = require( 'commander' );
 
+/**
+ * Internal dependencies
+ */
+const { formats } = require( './lib/logger' );
+
 const withOptions = ( command, options ) => {
 	options.forEach( ( { description, argname, defaults } ) => {
 		command = command.option( argname, description, defaults );
@@ -17,7 +22,7 @@ const catchException = ( handler ) => {
 		try {
 			await handler( ...args );
 		} catch ( error ) {
-			console.error( error ); // eslint-disable-line no-console
+			console.error( formats.error( error.message ) ); // eslint-disable-line no-console
 			process.exitCode = 1;
 		}
 	};
--- a/bin/plugin/commands/since.js
+++ b/bin/plugin/commands/since.js
@@ -20,7 +20,7 @@ const { plugins } = require( '../../../plugins.json' );
 exports.options = [
 	{
 		argname: '-p, --plugin <plugin>',
-		description: 'Plugin name',
+		description: 'Plugin slug',
 		defaults: 'performance-lab',
 	},
 	{
--- a/bin/plugin/commands/since.js
+++ b/bin/plugin/commands/since.js
@@ -18,6 +18,10 @@ const { log, formats } = require( '../lib/logger' );
 
 exports.options = [
 	{
+		argname: '-p, --plugin <plugin>',
+		description: 'Plugin name',
+	},
+	{
 		argname: '-r, --release <release>',
 		description: 'Release version number',
 	},
@@ -38,13 +42,23 @@ exports.handler = async ( opt ) => {
 		return;
 	}
 
-	const patterns = [
-		path.resolve( __dirname, '../../../**/*.php' ),
-		path.resolve( __dirname, '../../../**/*.js' ),
-	];
+	const patterns = [];
+	const pluginRoot = path.resolve( __dirname, '../../../' );
+	const ignore = [ '**/node_modules', '**/vendor', '**/bin', '**/build' ];
+
+	if ( opt.plugin ) {
+		const pluginPath = path.resolve( pluginRoot, 'plugins', opt.plugin );
+
+		patterns.push( `${ pluginPath }/**/*.php` );
+		patterns.push( `${ pluginPath }/**/*.js` );
+	} else {
+		ignore.push( '**/plugins' );
+		patterns.push( `${ pluginRoot }/**/*.php` );
+		patterns.push( `${ pluginRoot }/**/*.js` );
+	}
 
 	const files = await glob( patterns, {
-		ignore: [ __filename, '**/node_modules', '**/vendor' ],
+		ignore,
 	} );
 
 	const regexp = new RegExp( '@since(\\s+)n.e.x.t', 'g' );
--- a/bin/plugin/commands/since.js
+++ b/bin/plugin/commands/since.js
@@ -21,6 +21,7 @@ exports.options = [
 	{
 		argname: '-p, --plugin <plugin>',
 		description: 'Plugin name',
+		defaults: 'performance-lab',
 	},
 	{
 		argname: '-r, --release <release>',
@@ -43,7 +44,7 @@ exports.handler = async ( opt ) => {
 		return;
 	}
 
-	if ( opt.plugin && ! plugins.includes( opt.plugin ) ) {
+	if ( opt.plugin !== 'performance-lab' && ! plugins.includes( opt.plugin ) ) {
 		log(
 			formats.error(
 				`The plugin "${ opt.plugin }" is not found in the plugins.json file.`
@@ -56,7 +57,7 @@ exports.handler = async ( opt ) => {
 	const pluginRoot = path.resolve( __dirname, '../../../' );
 	const ignore = [ '**/node_modules', '**/vendor', '**/bin', '**/build' ];
 
-	if ( opt.plugin ) {
+	if ( opt.plugin !== 'performance-lab' ) {
 		const pluginPath = path.resolve( pluginRoot, 'plugins', opt.plugin );
 
 		patterns.push( `${ pluginPath }/**/*.php` );
--- a/bin/plugin/commands/since.js
+++ b/bin/plugin/commands/since.js
@@ -14,7 +14,8 @@ const { plugins } = require( '../../../plugins.json' );
 /**
  * @typedef WPSinceCommandOptions
  *
- * @property {string} version Version number.
+ * @property {string} plugin  Plugin slug.
+ * @property {string} release Release version number.
  */
 
 exports.options = [
--- a/bin/plugin/commands/since.js
+++ b/bin/plugin/commands/since.js
@@ -47,7 +47,7 @@ exports.handler = async ( opt ) => {
 	if ( opt.plugin !== 'performance-lab' && ! plugins.includes( opt.plugin ) ) {
 		log(
 			formats.error(
-				`The plugin "${ opt.plugin }" is not found in the plugins.json file.`
+				`The plugin "${ opt.plugin }" is not a valid plugin managed as part of this project.`
 			)
 		);
 		return;
--- a/bin/plugin/commands/since.js
+++ b/bin/plugin/commands/since.js
@@ -9,6 +9,7 @@ const fs = require( 'fs' );
  * Internal dependencies
  */
 const { log, formats } = require( '../lib/logger' );
+const { plugins } = require( '../../../plugins.json' );
 
 /**
  * @typedef WPSinceCommandOptions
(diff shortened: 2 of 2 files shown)


bin/plugin/commands/since.js
```diff
@@ -17,6 +18,10 @@ const { log, formats } = require( '../lib/logger' );
  */
 
 exports.options = [
+	{
+		argname: '-p, --plugin <plugin>',
+		description: 'Plugin name',
```
felixarntz:
```suggestion
		description: 'Plugin name',
		defaults: 'performance-lab',
```

bin/plugin/commands/since.js
```diff
@@ -17,6 +18,10 @@ const { log, formats } = require( '../lib/logger' );
  */
```
felixarntz, line 18: Worth adding the new `plugin` entry here. Also it looks like this was already wrong before, the value isn't called `version`, but `release`.

bin/plugin/commands/since.js
```diff
@@ -38,13 +43,32 @@ exports.handler = async ( opt ) => {
 		return;
 	}
 
-	const patterns = [
-		path.resolve( __dirname, '../../../**/*.php' ),
-		path.resolve( __dirname, '../../../**/*.js' ),
-	];
+	if ( opt.plugin && ! plugins.includes( opt.plugin ) ) {
```
felixarntz: We also need to support someone passing `performance-lab` (which should result in the default behavior).

By making it the default, we could use:
```suggestion
	if ( opt.plugin !== 'performance-lab' && ! plugins.includes( opt.plugin ) ) {
```

bin/plugin/commands/since.js
```diff
@@ -38,13 +43,32 @@ exports.handler = async ( opt ) => {
...
+		log(
+			formats.error(
+				`The plugin "${ opt.plugin }" is not found in the plugins.json file.`
```
felixarntz:
```suggestion
				`The plugin "${ opt.plugin }" is not a valid plugin managed as part of this project.`
```

bin/plugin/commands/since.js
```diff
@@ -38,13 +43,32 @@ exports.handler = async ( opt ) => {
...
+			)
+		);
+		return;
+	}
+
+	const patterns = [];
+	const pluginRoot = path.resolve( __dirname, '../../../' );
+	const ignore = [ '**/node_modules', '**/vendor', '**/bin', '**/build' ];
+
+	if ( opt.plugin ) {
```
felixarntz: See above:
```suggestion
	if ( opt.plugin !== 'performance-lab' ) {
```

felixarntz: Worth adding a comment to explain:
```suggestion
	/*
	 * For a standalone plugin, use the specific plugin directory.
	 * For Performance Lab, use the root directory and ignore the standalone plugin directories.
	 */
	if ( opt.plugin ) {
```

felixarntz (commented):
Thank you for the PR @thelovekesh, this looks great!

Just a few minor points of feedback.

bin/plugin/commands/since.js
felixarntz: This is not the plugin name, but plugin slug.
```suggestion
		description: 'Plugin slug',
```

thelovekesh: @felixarntz Any specific reason to add this, since by default it will run on the PL plugin anyway?

felixarntz: @thelovekesh From an API perspective it's better to have an actual value defined. We shouldn't use a default that is impossible to manually pass IMO.

This also creates a consistent DX. No matter which plugin you run the command for, you can pass the plugin slug via `--plugin`. The only reason that you don't _have_ to do this for `performance-lab` is backward compatibility.

Last but not least, having a distinct value for `performance-lab` will help later if we decide to move that plugin's code to `./plugins/performance-lab` and therefore treat it similar to the other plugins.

thelovekesh: Humm, I see this change in the readme command as well.

bin/plugin/commands/since.js
thelovekesh: done in https://github.com/WordPress/performance/pull/1119/commits/bf5547f2d58cd0671a8232c708fcb234ea0c8016

bin/plugin/commands/since.js
felixarntz: How do you mean? The readme command checks the provided plugin slug against the milestones, which means `performance-lab ...` would also be covered. But since it's not in `plugins.json`, here we have to manually check for it.

thelovekesh: By this change, I meant using performance-lab check for plugin root.

felixarntz (approved):
@thelovekesh Thank you for the updates, LGTM!

bin/plugin/commands/since.js
```diff
@@ -38,13 +45,39 @@ exports.handler = async ( opt ) => {
...
+	if (
+		opt.plugin !== 'performance-lab' &&
+		! plugins.includes( opt.plugin )
+	) {
+		log(
+			formats.error(
+				`The plugin "${ opt.plugin }" is not a valid plugin managed as part of this project.`
+			)
+		);
+		return;
```
westonruter: How about exiting with an error code?

```suggestion
		process.exit(1);
```

felixarntz: @thelovekesh @westonruter This actually shouldn't exit or output the error at all. This is handled centrally by the CLI tool, and changing it here unnecessarily couples error handling to the output. Let's change this to simply `throw new Error( ... )`, which is consistent with the other commands.

thelovekesh: @felixarntz I think it simply exits when an error is thrown. 

<details><summary>See Output</summary>

```
Error: The plugin "foo" is not a valid plugin managed as part of this project.
    at exports.handler (/home/thelovekesh/Desktop/work/google/wpdev/public/content/plugins/performance/bin/plugin/commands/since.js:61:9)
    at Command.<anonymous> (/home/thelovekesh/Desktop/work/google/wpdev/public/content/plugins/performance/bin/plugin/cli.js:18:10)
    at Command.listener [as _actionHandler] (/home/thelovekesh/Desktop/work/google/wpdev/public/content/plugins/performance/node_modules/commander/lib/command.js:482:17)
    at /home/thelovekesh/Desktop/work/google/wpdev/public/content/plugins/performance/node_modules/commander/lib/command.js:1283:65
    at Command._chainOrCall (/home/thelovekesh/Desktop/work/google/wpdev/public/content/plugins/performance/node_modules/commander/lib/command.js:1177:12)
    at Command._parseCommand (/home/thelovekesh/Desktop/work/google/wpdev/public/content/plugins/performance/node_modules/commander/lib/command.js:1283:27)
    at /home/thelovekesh/Desktop/work/google/wpdev/public/content/plugins/performance/node_modules/commander/lib/command.js:1081:27
    at Command._chainOrCall (/home/thelovekesh/Desktop/work/google/wpdev/public/content/plugins/performance/node_modules/commander/lib/command.js:1177:12)
    at Command._dispatchSubcommand (/home/thelovekesh/Desktop/work/google/wpdev/public/content/plugins/performance/node_modules/commander/lib/command.js:1077:23)
    at Command._parseCommand (/home/thelovekesh/Desktop/work/google/wpdev/public/content/plugins/performance/node_modules/commander/lib/command.js:1248:19)
```
</details>

Which I don't think is handled by the CLI tool as per its API docs - https://www.npmjs.com/package/commander#display-error

Also, the message it's logging is pretty verbose + CLI doesn't need to handle anything after it, so exiting here shouldn't be an issue or we handle it the CLI way i.e. using `program.error()`.

westonruter: Humm. So when an an exception is thrown it should actually set an error `exitCode` already:

https://github.com/WordPress/performance/blob/9d0a86975a685bdde48bf640ca2f85495d994273/bin/plugin/cli.js#L15-L24

But that's not working?

thelovekesh: Ohh seems like I missed this. The exit code is 1 yes. Are we ok with a stack trace along with an error message? UX-wise, it can confuse someone with a CLI error while a simple error message in red color shows user has passed the wrong argument.

felixarntz: Can we please open a separate issue or PR to discuss this?

This PR is for adding the `since` enhancement, so we should simply throw an error here as it works elsewhere. If the way the CLI tool handles errors has some issues, I'd be supportive of improving that. But let's handle that as its own problem.

westonruter: I see the problem. The issue is that it was doing `console.error( error )` instead of `console.error( error.message )`. I've fixed that and added the red color in b65b8a8d.

bin/plugin/cli.js
```diff
@@ -17,7 +22,7 @@ const catchException = ( handler ) => {
 		try {
 			await handler( ...args );
 		} catch ( error ) {
-			console.error( error ); // eslint-disable-line no-console
+			console.error( formats.error( error.message ) ); // eslint-disable-line no-console
```
thelovekesh, line 25: cool 😎

felixarntz (approved):
Thanks @thelovekesh @westonruter, current code looks great!

bin/plugin/commands/since.js
westonruter: image

westonruter (approved):

PR #1123, opened 2024-04-11
WPCS Improvement

Chintesh:
## Summary

Fixes #

## Relevant technical choices

I've refactored the code to address PHPCS (PHP CodeSniffer) warnings and improve code readability and standards compliance.

- Used `isset()` to check if a variable is set before accessing it to avoid illegal string offset warnings.
- Replaced `strpos()` with `str_contains()` for more readable string containment checks.
- ~Utilized `filter_input()` with appropriate sanitization for better handling of input data from superglobals.~
- ~Ensured that all global variables like `$_SERVER` are properly validated before use to prevent potential security vulnerabilities.~

These changes aim to enhance the overall quality and maintainability of the codebase by adhering to best practices recommended by PHPCS.

### DIFF
--- a/plugins/auto-sizes/hooks.php
+++ b/plugins/auto-sizes/hooks.php
@@ -30,7 +30,7 @@ function auto_sizes_update_image_attributes( $attr ) {
 	}
 
 	// Don't add 'auto' to the sizes attribute if it already exists.
-	if ( false !== strpos( $attr['sizes'], 'auto,' ) ) {
+	if ( str_contains( $attr['sizes'], 'auto,' ) ) {
 		return $attr;
 	}
 
--- a/plugins/optimization-detective/optimization.php
+++ b/plugins/optimization-detective/optimization.php
@@ -73,6 +73,10 @@ function od_maybe_add_template_output_buffer_filter() {
  * @return bool Whether response can be optimized.
  */
 function od_can_optimize_response(): bool {
+
+	// Retrieve the sanitized value of the 'REQUEST_METHOD' server variable.
+	$method = filter_input( INPUT_SERVER, 'REQUEST_METHOD', FILTER_SANITIZE_STRING );
+	
 	$able = ! (
 		// Since there is no predictability in whether posts in the loop will have featured images assigned or not. If a
 		// theme template for search results doesn't even show featured images, then this wouldn't be an issue.
@@ -80,7 +84,7 @@ function od_can_optimize_response(): bool {
 		// Since injection of inline-editing controls interfere with breadcrumbs, while also just not necessary in this context.
 		is_customize_preview() ||
 		// Since the images detected in the response body of a POST request cannot, by definition, be cached.
-		'GET' !== $_SERVER['REQUEST_METHOD'] ||
+		( isset( $method ) && 'GET' !== $method ) ||
 		// The aim is to optimize pages for the majority of site visitors, not those who administer the site. For admin
 		// users, additional elements will be present like the script from wp_customize_support_script() which will
 		// interfere with the XPath indices. Note that od_get_normalized_query_vars() is varied by is_user_logged_in()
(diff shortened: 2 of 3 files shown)


plugins/optimization-detective/optimization.php
```diff
@@ -73,14 +73,18 @@ function od_maybe_add_template_output_buffer_filter() {
  * @return bool Whether response can be optimized.
  */
 function od_can_optimize_response(): bool {
+
+	// Retrieve the sanitized value of the 'REQUEST_METHOD' server variable.
+	$method = filter_input( INPUT_SERVER, 'REQUEST_METHOD', FILTER_SANITIZE_STRING );
```
westonruter: I don't think we should use `filter_input()` for two reasons:

1. It's not used in core.
2. It's not able to be unit tested (e.g. by setting `$_SERVER['REQUEST_METHOD']`).

You can see there is a unit test failing for the second reason: https://github.com/WordPress/performance/actions/runs/8651242612/job/23721516648?pr=1123#step:9:84

You can see the test setup is overriding `$_SERVER['POST_METHOD']`: 

https://github.com/WordPress/performance/blob/2fd48b05409136580dd6bbdd4812dde95ccdd389/tests/plugins/optimization-detective/optimization-tests.php#L124-L130

plugins/auto-sizes/hooks.php
```diff
@@ -30,7 +30,7 @@ function auto_sizes_update_image_attributes( $attr ) {
 	}
 
 	// Don't add 'auto' to the sizes attribute if it already exists.
-	if ( false !== strpos( $attr['sizes'], 'auto,' ) ) {
+	if ( str_contains( $attr['sizes'], 'auto,' ) ) {
```
westonruter, line 33: `str_contains()` is only available as of PHP8. However, since WP 5.9 and later implement a [polyfill](https://github.com/WordPress/wordpress-develop/blob/5afdac7e057bf4faf097079c0704f6e834fdce97/src/wp-includes/compat.php#L455-L475) for this, then this is fine.

plugins/optimization-detective/optimization.php
```diff
@@ -73,14 +73,18 @@ function od_maybe_add_template_output_buffer_filter() {
...
+	
 	$able = ! (
 		// Since there is no predictability in whether posts in the loop will have featured images assigned or not. If a
 		// theme template for search results doesn't even show featured images, then this wouldn't be an issue.
 		is_search() ||
 		// Since injection of inline-editing controls interfere with breadcrumbs, while also just not necessary in this context.
 		is_customize_preview() ||
 		// Since the images detected in the response body of a POST request cannot, by definition, be cached.
-		'GET' !== $_SERVER['REQUEST_METHOD'] ||
+		( isset( $method ) && 'GET' !== $method ) ||
```
westonruter: There isn't a need for `isset()` if the `$method` variable was set above, is there? Here it's just checking if it is `null`. There's no string offset going on.

plugins/speculation-rules/helper.php
```diff
@@ -93,7 +93,7 @@ function plsr_get_speculation_rules() {
 		),
 	);
 
-	// Allow adding a class on any links to prevent prerendering.
+	// Allow adding a class on any links to prevent pre-rendering.
```
westonruter: I don't think this change should be made since "prerendering" is the term used. See https://developer.mozilla.org/en-US/docs/Glossary/Prerender

westonruter (changes requested):
Thanks for the PR! 

I'm curious about how you identified the WPCS issues because we're already running PHPCS checks with WPCS on this repo?

plugins/optimization-detective/optimization.php
Chintesh: Got it, I have Updated this

Thanks @westonruter

plugins/speculation-rules/helper.php
Chintesh: I have Reverted it

Thanks @westonruter

plugins/optimization-detective/optimization.php
Chintesh: I have Updated this

Thanks @westonruter

westonruter (approved):

felixarntz (approved):
Thank you for the PR @Chintesh!

PR #1128, opened 2024-04-12
Add missing uninstall.php to remove plugin's database option

westonruter:
In preparing for the PL 3.0.0 release, I was looking at the `build` directory after running `npm run build-plugins`. I noticed that while `webp-uploads` has an `uninstall.php` now (#1116), and an `uninstall.php` is also present for `optimization-detective`, there is none for Speculative Loading. This adds the missing `uninstall.php`, adapted from the one from `webp-uploads`.

This also bumps the plugin version to 1.2.0 and updates the changelog.

### DIFF
--- a/plugins/speculation-rules/readme.txt
+++ b/plugins/speculation-rules/readme.txt
@@ -104,8 +104,6 @@ Contributions are always welcome! Learn more about how to get involved in the [C
 
 = 1.2.0 =
 
-**Enhancements**
-
 * Add missing uninstall.php to remove plugin's database option. ([1128](https://github.com/WordPress/performance/pull/1128))
 
 = 1.1.0 =
--- a/plugins/speculation-rules/readme.txt
+++ b/plugins/speculation-rules/readme.txt
@@ -4,7 +4,7 @@ Contributors:      wordpressdotorg
 Requires at least: 6.4
 Tested up to:      6.5
 Requires PHP:      7.0
-Stable tag:        1.1.0
+Stable tag:        1.2.0
 License:           GPLv2 or later
 License URI:       https://www.gnu.org/licenses/gpl-2.0.html
 Tags:              performance, javascript, speculation rules, prerender, prefetch
--- a/plugins/speculation-rules/readme.txt
+++ b/plugins/speculation-rules/readme.txt
@@ -102,6 +102,12 @@ Contributions are always welcome! Learn more about how to get involved in the [C
 
 == Changelog ==
 
+= 1.2.0 =
+
+**Enhancements**
+
+* Add missing uninstall.php to remove plugin's database option. ([1128](https://github.com/WordPress/performance/pull/1128))
+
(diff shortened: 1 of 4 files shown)


plugins/speculation-rules/uninstall.php
```diff
@@ -0,0 +1,41 @@
+<?php
+/**
+ * Plugin uninstaller logic.
+ *
+ * @package speculation-rules
+ * @since 1.1.0
```
felixarntz: Per the version you used elsewhere:
```suggestion
 * @since 1.2.0
```

plugins/speculation-rules/uninstall.php
```diff
@@ -0,0 +1,41 @@
...
+		restore_current_blog();
+	}
+}
+
+speculation_rules_delete_plugin_option();
+
+/**
+ * Delete the current site's option.
+ *
+ * @since 1.1.0
```
felixarntz:
```suggestion
 * @since 1.2.0
```

plugins/speculation-rules/uninstall.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+function speculation_rules_delete_plugin_option() {
```
felixarntz: Maybe use `plsr_` instead of `speculation_rules_` for consistency with the other functions in the plugin.

felixarntz (approved):
Thanks @westonruter. Approving, though minor nit-pick findings below.

plugins/speculation-rules/readme.txt
```diff
@@ -102,6 +102,12 @@ Contributions are always welcome! Learn more about how to get involved in the [C
 
 == Changelog ==
 
+= 1.2.0 =
+
+**Enhancements**
+
+* Add missing uninstall.php to remove plugin's database option. ([1128](https://github.com/WordPress/performance/pull/1128))
```
mukeshpanchal27:
```suggestion
= 1.2.0 =

* Add missing uninstall.php to remove plugin's database option. ([1128](https://github.com/WordPress/performance/pull/1128))
```

Not blocker: For the plugins changelog, we don't use titles; we only use them for the PL plugin.

mukeshpanchal27 (approved):

plugins/speculation-rules/readme.txt
westonruter: Humm, this was generated by the `npm run changelog` (which I hacked to get the PR even though it wasn't closed yet).

felixarntz: @mukeshpanchal27 @westonruter In the manual approach we've taken so far we didn't add those headings, but that's mostly an oversight from doing it manually.

I don't feel strongly about this. Technically since we have been doing it manually for this round of releases, it could be done either way. But going forward (after the PL 3.0.0 release) we should be entirely relying on the changelog generator script, i.e. including the titles.

PR #1130, opened 2024-04-12
Bump minimum PHP requirement to 7.2

### LINKED ISSUE #1129: Bump the minimum PHP version to 7.2
As WordPress bumps the [minimum required PHP version to 7.2](https://core.trac.wordpress.org/ticket/58719) for 6.6 version, we should do the same for Performance lab and all standalone plugins.

thelovekesh:
## Summary
https://core.trac.wordpress.org/ticket/58719 is recently merged in Core which bumps the minimum PHP to 7.2. Now that we can bump to 7.2, all composer packages required by this repo are PHP 7.2+ compatible and we can safely merge root `composer.json` and `build-cs/composer.json`.

Fixes https://github.com/WordPress/performance/issues/1129

## Relevant technical choices

- Merge root `build-cs/composer.json` in root `composer.json`.
- Set composer platform.php to 7.2 and bump PHP constraint to 7.2.


thelovekesh:
Once I set platform.php to 7.2, PHPStan start reporting few errors:

```log
Error: Function str_starts_with not found.
Error: Function str_starts_with not found.
Error: Function str_contains not found.
Error: Function str_contains not found.
Error: Function str_contains not found.
Error: Function str_contains not found.
Error: Function str_contains not found.
Error: Function str_starts_with not found.
Error: Function str_starts_with not found.
Error: Function str_starts_with not found.
```

It means we are using functions that are not supported in PHP ^7.2? 🤔

thelovekesh:
@swissspidy Can you share your thoughts here? Also wanted to know if we are good to keep this or need to wait till 6.6 comes out.

swissspidy:
Let's wait until 6.6 is released. After that we can bump our minimum requirement too.

Performance Lab & co. still support WordPress 6.4, FWIW. But I think it's OK to bump to PHP 7.2 regardless.

Also, if you bump the PHP requirement in Composer you must also update the `Requires PHP` header in the main plugin files (and the readme.txt files, although we should actually remove it from there, it's useless)

westonruter:
> Performance Lab & co. still support WordPress 6.4, FWIW. But I think it's OK to bump to PHP 7.2 regardless.

@swissspidy Meaning, bump to PHP 7.2 when WP 6.6 is released. At this time WP 6.5 will be the minimum supported version which still supports PHP 7.0, but that is fine?

IMO, I'd be fine with bumping to 7.2 now even though we still support WP 6.4. Since core trunk now requires PHP 7.2, if we follow suit now we can more closely follow core in development of feature plugins. For example, there may be code in trunk we copy into our plugins and vice-versa. If we're on the same PHP version now, that will be easier. And if we force users to prematurely upgrade to PHP 7.2, then I don't think this is a bad thing at all.

swissspidy:
That‘s what I meant, yes. I don‘t have a strong opinion though. Core won‘t change to 7.2+ specific code right away, so code copying is probably not an issue. But if you think it‘s easier this way, we can also bump now

westonruter:
I think it's better for us to err on the side of being more current than on the side of holding back, especially since it reduces our maintenance burden and improves tool compatibility. And since core is adopting 7.2 anyway and our codebase is feature plugins intended for core merge, being on whatever the next PHP version is for core makes sense to me. The only downside is not getting potential testers from PHP 7.0 and PHP 7.1. But their usage is only 1.5% and 0.9% respectively, so that seems like a non-issue.

mukeshpanchal27:
It seems like this PR is removing `build-cs` rather than bumping PHP to 7.2. Could you separate it so it's easier for review?

swissspidy:
As per my above comment, the PR should do both

bin/phpcs/phpcs.ruleset.xml
```diff
@@ -3,7 +3,7 @@
 	<description>Sniffs for WordPress plugins</description>
 
 	<rule ref="PHPCompatibility"/>
-	<config name="testVersion" value="7.0-"/>
+	<config name="testVersion" value="7.4-"/>
```
swissspidy:
```suggestion
	<config name="testVersion" value="7.2-"/>
```

phpstan.neon.dist
```diff
@@ -1,5 +1,5 @@
 includes:
-	- build-cs/vendor/phpstan/phpstan/conf/bleedingEdge.neon
+	- phar://phpstan.phar/conf/bleedingEdge.neon
```
swissspidy, line 2: No idea why PHPStan didn't complain before, but we need to update the config file here to ignore any errors about `str_contains` and similar, as WP polyfills those.

Example:

https://github.com/WordPress/plugin-check/blob/6721eb4d281bf39469289590233277012e66807a/phpstan.neon.dist#L17-L27

swissspidy:
> Once I set platform.php to 7.2, PHPStan start reporting few errors:
> 
> ```
> Error: Function str_starts_with not found.
> Error: Function str_starts_with not found.
> Error: Function str_contains not found.
> Error: Function str_contains not found.
> Error: Function str_contains not found.
> Error: Function str_contains not found.
> Error: Function str_contains not found.
> Error: Function str_starts_with not found.
> Error: Function str_starts_with not found.
> Error: Function str_starts_with not found.
> ```
> 
> It means we are using functions that are not supported in PHP ^7.2? 🤔

I missed this comment.

Yes and no, WordPress provides polyfilles for those.

phpstan.neon.dist
thelovekesh: PHPStan takes platform PHP into account if it's present when it initialize the PHP parser - https://github.com/phpstan/phpstan-src/blob/aa801e413e81c8552878a5d7ef418f00532c6f5b/src/Php/PhpVersionFactoryFactory.php#L36

swissspidy: Ah that makes sense, it was missing before.

swissspidy (approved):

westonruter:
Let's wait to merge this until after the 3.0.0 release (in an hour or so) because there will be a merge conflict in `composer.json` since `ext-dom` was removed.

westonruter:
OK, merge conflicts ready to be resolved 😬

thelovekesh:
> OK, merge conflicts ready to be resolved 😬

Done.

westonruter (approved):

composer.json
```diff
@@ -1,31 +1,50 @@
...
-    "php": ">=7|^8"
+  "autoload-dev": {
+    "psr-4": {
+      "PerformanceLab\\Tests\\": "tests/utils"
+    }
+  },
+  "config": {
+    "allow-plugins": {
+      "dealerdirect/phpcodesniffer-composer-installer": true,
+      "phpstan/extension-installer": true
```
westonruter, line 39: Is this missing `composer/installers`? I'm getting the following when I do `composer install`:

image

westonruter:
```
composer/installers contains a Composer plugin which is currently not in your allow-plugins config. See https://getcomposer.org/allow-plugins
Do you trust "composer/installers" to execute code and wish to enable it now? (writes "allow-plugins" to composer.json) [y,n,d,?] ?
y - add package to allow-plugins in composer.json and let it run immediately
n - add package (as disallowed) to allow-plugins in composer.json to suppress further prompts
d - discard this, do not change composer.json and do not allow the plugin to run
? - print help
Do you trust "composer/installers" to execute code and wish to enable it now? (writes "allow-plugins" to composer.json) [y,n,d,?] d
```

I went with `d`. I'm not seeing it happen now when I do `composer install`. 😕

thelovekesh: This might be due to stale vendor content. I don't see if we are using this plugin so I removed it.

PR #1142, opened 2024-04-15
Exclude rel=nofollow links from prefetch/prerender

### LINKED ISSUE #1140: Products added to cart in WooCommerce without clicking on Add to Cart button when Speculative Loading active
## Bug Description

As reported on the [support forum](https://wordpress.org/support/topic/eager-mode-causes-woocommerce-products-to-be-added-to-the-cart/), users browsing a site with WooCommerce and the Speculative Loading plugin can end up getting products added to their shopping cart by just hovering over the Add to Cart button when in moderate eagerness (the default setting). What's worse is when the eager eagerness setting is enabled, all the products on the page can be added to the cart just by visiting the page.

## Steps to reproduce

1. Activate and configure WooCommerce (and install the Storefront theme, perhaps)
2. Activate the Speculative Loading plugin
3. Access the shop page and hover over an Add to Cart button

## Screenshots

[Screen recording 2024-04-15 11.49.08.webm](https://github.com/WordPress/performance/assets/134745/d3ff2249-7235-4edb-a3db-b1cda2e3cee8)

## Additional Context

When not using a block theme, this is the code to create the add-to-cart button:
https://github.com/woocommerce/woocommerce/blob/a8e84d8c8b20558f66ed33dcbee6c1dec300752f/plugins/woocommerce/includes/wc-template-functions.php#L1363-L1403
https://github.com/woocommerce/woocommerce/blob/a8e84d8c8b20558f66ed33dcbee6c1dec300752f/plugins/woocommerce/templates/loop/add-to-cart.php#L27-L32

Note an `a` tag is always used. Contrast with the block code:
https://github.com/woocommerce/woocommerce/blob/a8e84d8c8b20558f66ed33dcbee6c1dec300752f/plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php#L74-L231

Here a `button` is used when AJAX is enabled.

In both cases, however, `rel=nofollow` is present (although it can be filtered out by the `woocommerce_loop_add_to_cart_args` filter). It's unfortunate that since these add-to-cart links aren't idempotent (i.e. they have an action) that they don't also have `_wpnonce` query parameters.

westonruter:
## Summary

WooCommerce has `rel=nofollow` on its Add to Cart buttons (which are often links instead of `button` elements). These links are not idempotent and they must be excluded from prefech/prerender. 

Fixes #1140

Before | After
--|--
image | image

## Relevant technical choices

Alternatively this could have excluded URLs that include the `add-to-cart` query parameter. However, this would seem to be overly-specific to WooCommerce. The issue https://github.com/WICG/nav-speculation/issues/309 was opened by @tunetheweb to consider whether `rel=nofollow` links should be excluded by default as well, so that is what this PR is going with now as well.

Eventually there may need to be a filter in addition to `plsr_speculation_rules_href_exclude_paths` which applies on the entire speculation rules array so that a plugin can add exclusions which are not URL-specific.

### Before

```json
{
  "prerender": [
    {
      "source": "document",
      "where": {
        "and": [
          {
            "href_matches": "/*"
          },
          {
            "not": {
              "href_matches": [
                "/wp-login.php",
                "/wp-admin/*"
              ]
            }
          },
          {
            "not": {
              "selector_matches": ".no-prerender"
            }
          }
        ]
      },
      "eagerness": "moderate"
    }
  ]
}
```

### After

```json
{
  "prerender": [
    {
      "source": "document",
      "where": {
        "and": [
          {
            "href_matches": "/*"
          },
          {
            "not": {
              "href_matches": [
                "/wp-login.php",
                "/wp-admin/*"
              ]
            }
          },
          {
            "not": {
              "selector_matches": "a[rel=nofollow]"
            }
          },
          {
            "not": {
              "selector_matches": ".no-prerender"
            }
          }
        ]
      },
      "eagerness": "moderate"
    }
  ]
}
```

### DIFF
--- a/plugins/speculation-rules/helper.php
+++ b/plugins/speculation-rules/helper.php
@@ -87,7 +87,7 @@ function plsr_get_speculation_rules() {
 							'href_matches' => $href_exclude_paths,
 						),
 					),
-					// Also exclude rel=nofollow links, as plugins like WooCommerce attribute their add-to-cart links (which should buttons).
+					// Also exclude rel=nofollow links, as plugins like WooCommerce attribute their add-to-cart links.
 					array(
 						'not' => array(
 							'selector_matches' => 'a[rel=nofollow]',
--- a/plugins/speculation-rules/helper.php
+++ b/plugins/speculation-rules/helper.php
@@ -77,7 +77,7 @@ function plsr_get_speculation_rules() {
 			'source'    => 'document',
 			'where'     => array(
 				'and' => array(
-					// Prerender any URLs within the same site.
+					// Include any URLs within the same site.
 					array(
 						'href_matches' => $prefixer->prefix_path_pattern( '/*' ),
 					),
@@ -85,6 +85,12 @@ function plsr_get_speculation_rules() {
 					array(
 						'not' => array(
 							'href_matches' => $href_exclude_paths,
+						),
+					),
+					// Also exclude rel=nofollow links, as plugins like WooCommerce attribute their add-to-cart links (which should buttons).
+					array(
+						'not' => array(
+							'selector_matches' => 'a[rel=nofollow]',
 						),
 					),
 				),
--- a/plugins/speculation-rules/helper.php
+++ b/plugins/speculation-rules/helper.php
@@ -87,7 +87,7 @@ function plsr_get_speculation_rules() {
 							'href_matches' => $href_exclude_paths,
 						),
 					),
-					// Also exclude rel=nofollow links, as plugins like WooCommerce attribute their add-to-cart links.
+					// Also exclude rel=nofollow links, as plugins like WooCommerce use that on their add-to-cart links.
 					array(
 						'not' => array(
 							'selector_matches' => 'a[rel=nofollow]',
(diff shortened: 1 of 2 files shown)


plugins/speculation-rules/helper.php
```diff
@@ -87,6 +87,12 @@ function plsr_get_speculation_rules() {
 							'href_matches' => $href_exclude_paths,
 						),
 					),
+					// Also exclude rel=nofollow links, as plugins like WooCommerce attribute their add-to-cart links (which should buttons).
```
adamsilverstein: typo here? anyway, maybe remove this part and suggest Woo fix upstream.
```suggestion
					// Also exclude rel=nofollow links, as plugins like WooCommerce attribute their add-to-cart links.
```

adamsilverstein (commented):
Looks good, one small doc typo/suggestion. Also, the tests are complaining about your changes.

plugins/speculation-rules/helper.php
```diff
@@ -87,6 +87,12 @@ function plsr_get_speculation_rules() {
 							'href_matches' => $href_exclude_paths,
 						),
 					),
+					// Also exclude rel=nofollow links, as plugins like WooCommerce attribute their add-to-cart links.
```
tunetheweb: This reads weird to me.

```suggestion
					// Also exclude rel=nofollow links, as plugins like WooCommerce use that on their add-to-cart links.
```

tunetheweb (approved):
LGTM with one suggestion.

plugins/speculation-rules/helper.php
westonruter: I was trying to be clever and use "attribute" as a verb. But clearly my pun was a dud.

swissspidy (approved):

PR #1147, opened 2024-04-16
Do not show admin pointer on the Performance screen and dismiss the pointer when visited

### LINKED ISSUE #1135: Setting screen tooltip should go away after visiting the settings page
## Bug Description

The tool tip guiding users to the settings page shows even after they visit it. Instead, it should go away as if "Dismiss" was clicked.

This tooltip pops up when the user installs the plugin so they can more easily find the settings screen:

image

Once the user clicks the "Settings > Performance" link in the tooltip, or otherwise visits the settings page, we should no longer show the tooltip.

## Steps to reproduce

1. Install the plugin
2. Click "Settings > Performance" in the tooltip to visit the settings page
3. Go back and visit the Dashboard

Expected results: the tooltip should no longer show
Actual results: the tooltip is still showing

Ellusu:
## Summary

Fixes: #1135

## Relevant technical choices

I added the possibility to set the value of dismissed_wp_pointers at the beginning of the function if we are already in the plugin settings page.

I had to set $current_user at the beginning of the function

### DIFF
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -82,7 +82,7 @@ function perflab_admin_pointer( $hook_suffix ) {
 		return;
 	}
 	$current_user = get_current_user_id();
-	$dismissed = array_filter( explode( ',', (string) get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) ) );
+	$dismissed    = array_filter( explode( ',', (string) get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) ) );
 
 	if ( in_array( 'perflab-admin-pointer', $dismissed, true ) ) {
 		return;
@@ -91,7 +91,7 @@ function perflab_admin_pointer( $hook_suffix ) {
 	if ( ! in_array( $hook_suffix, array( 'index.php', 'plugins.php' ), true ) ) {
 
 		// Do not show on the settings page and dismiss the pointer.
-		if ( ! empty( $_GET['page'] ) && PERFLAB_SCREEN === $_GET['page'] && ( ! in_array( 'perflab-admin-pointer', $dismissed, true ))) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		if ( ! empty( $_GET['page'] ) && PERFLAB_SCREEN === $_GET['page'] && ( ! in_array( 'perflab-admin-pointer', $dismissed, true ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 			$dismissed[] = 'perflab-admin-pointer';
 			update_user_meta( $current_user, 'dismissed_wp_pointers', implode( ',', $dismissed ) );
 		}
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -77,26 +77,25 @@ function perflab_render_settings_page() {
  * @param string $hook_suffix The current admin page.
  */
 function perflab_admin_pointer( $hook_suffix ) {
-	$current_user = get_current_user_id();
-	$dismissed    = array_filter( explode( ',', (string) get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) ) );
-
-	if ( ! in_array( $hook_suffix, array( 'index.php', 'plugins.php' ), true ) ) {
-
-		// Do not show on the settings page and dismiss the pointer.
-		if ( isset( $_GET['page'] ) && PERFLAB_SCREEN === $_GET['page'] && ( ! in_array( 'perflab-admin-pointer', $dismissed, true ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-			$dismissed[] = 'perflab-admin-pointer';
-			update_user_meta( $current_user, 'dismissed_wp_pointers', implode( ',', $dismissed ) );
-		}
-
-		return;
-	}
-
 	// Do not show admin pointer in multisite Network admin or User admin UI.
 	if ( is_network_admin() || is_user_admin() ) {
 		return;
 	}
+	$current_user = get_current_user_id();
+	$dismissed = array_filter( explode( ',', (string) get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) ) );
 
 	if ( in_array( 'perflab-admin-pointer', $dismissed, true ) ) {
+		return;
+	}
+
+	if ( ! in_array( $hook_suffix, array( 'index.php', 'plugins.php' ), true ) ) {
+
+		// Do not show on the settings page and dismiss the pointer.
+		if ( ! empty( $_GET['page'] ) && PERFLAB_SCREEN === $_GET['page'] && ( ! in_array( 'perflab-admin-pointer', $dismissed, true ))) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			$dismissed[] = 'perflab-admin-pointer';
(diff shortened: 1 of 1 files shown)


includes/admin/load.php
```diff
@@ -77,7 +77,16 @@ function perflab_render_settings_page() {
  * @param string $hook_suffix The current admin page.
  */
 function perflab_admin_pointer( $hook_suffix ) {
+	$current_user = get_current_user_id();
+
 	if ( ! in_array( $hook_suffix, array( 'index.php', 'plugins.php' ), true ) ) {
+
+		// Do not show on the settings page and dismiss the pointer
```
mukeshpanchal27:
```suggestion
		// Do not show on the settings page and dismiss the pointer.
```

includes/admin/load.php
```diff
@@ -77,7 +77,16 @@ function perflab_render_settings_page() {
...
+		if ( !empty( $_GET['page'] ) && $_GET['page'] === PERFLAB_SCREEN ) {
```
mukeshpanchal27:
```suggestion
		if ( ! empty( $_GET['page'] ) && PERFLAB_SCREEN === $_GET['page'] ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```

includes/admin/load.php
```diff
@@ -86,7 +95,6 @@ function perflab_admin_pointer( $hook_suffix ) {
 		return;
 	}
 
-	$current_user = get_current_user_id();
 	$dismissed    = explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) );
```
mukeshpanchal27:
```suggestion
	$dismissed = explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) );
```

mukeshpanchal27 (commented):
Thanks @Ellusu, Left some initial feedback for PHPCS fix.

includes/admin/load.php
```diff
@@ -77,7 +77,16 @@ function perflab_render_settings_page() {
...
  */
 function perflab_admin_pointer( $hook_suffix ) {
+	$current_user = get_current_user_id();
+
 	if ( ! in_array( $hook_suffix, array( 'index.php', 'plugins.php' ), true ) ) {
+
+		// Do not show on the settings page and dismiss the pointer.
+		if ( ! empty( $_GET['page'] ) && PERFLAB_SCREEN === $_GET['page'] ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			$dismissed[] = 'perflab-admin-pointer';
+			update_user_meta( $current_user, 'dismissed_wp_pointers', implode( ',', $dismissed ) );
```
adamsilverstein, line 96: wouldn't this overwrite existing dismissed pointers? This probably needs to get the existing list first and append this pointer if it doesn't exist. The core function `wp_ajax_dismiss_wp_pointer` has a good boilerplate for doing this.

includes/admin/load.php
```diff
@@ -77,7 +77,16 @@ function perflab_render_settings_page() {
  * @param string $hook_suffix The current admin page.
  */
 function perflab_admin_pointer( $hook_suffix ) {
+	$current_user = get_current_user_id();
+
 	if ( ! in_array( $hook_suffix, array( 'index.php', 'plugins.php' ), true ) ) {
+
+		// Do not show on the settings page and dismiss the pointer.
```
adamsilverstein, line 93:
```suggestion
		// Dismiss the admin pointer once the user reaches the settings screen.
```
Docblock update suggestion

adamsilverstein:
Thanks @Ellusu - left a couple of suggestions.

includes/admin/load.php
westonruter: I think this would be the case if the `$dismissed` assignment line from below is also moved up after setting `$current_user`.

Ellusu: thanks @adamsilverstein, @mukeshpanchal27 e @westonruter, made a new commit

includes/admin/load.php
```diff
@@ -77,7 +77,17 @@ function perflab_render_settings_page() {
  * @param string $hook_suffix The current admin page.
  */
 function perflab_admin_pointer( $hook_suffix ) {
+	$current_user = get_current_user_id();
+	$dismissed    = array_filter( explode( ',', (string) get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) ) );
+
 	if ( ! in_array( $hook_suffix, array( 'index.php', 'plugins.php' ), true ) ) {
+
+		// Do not show on the settings page and dismiss the pointer.
+		if ( ! empty( $_GET['page'] ) && PERFLAB_SCREEN === $_GET['page'] && ( ! in_array( 'perflab-admin-pointer', $dismissed, true ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
westonruter: Minor suggestion based on [prior feedback](https://github.com/WordPress/performance/pull/1091#discussion_r1543908524) to use `isset()` rather than `! empty()`:

```suggestion
		if ( isset( $_GET['page'] ) && PERFLAB_SCREEN === $_GET['page'] && ( ! in_array( 'perflab-admin-pointer', $dismissed, true ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```

westonruter (approved):
Works for me!

Before:

[Screen recording 2024-04-17 14.32.10.webm](https://github.com/WordPress/performance/assets/134745/4672a1ba-322c-4b35-93f0-e15c782ad823)

After:

[Screen recording 2024-04-17 14.33.11.webm](https://github.com/WordPress/performance/assets/134745/38f7b87c-8b92-472e-afa9-5176817f0fc1)

includes/admin/load.php
```diff
@@ -77,7 +77,17 @@ function perflab_render_settings_page() {
  * @param string $hook_suffix The current admin page.
  */
 function perflab_admin_pointer( $hook_suffix ) {
```
mukeshpanchal27, line 79:
```
// Do not show admin pointer in multisite Network admin or User admin UI.
if ( is_network_admin() || is_user_admin() ) {
	return;
}
```

Let's check this condition at the top of the function, as the changes do not need to run on multisite.

includes/admin/load.php
```diff
@@ -86,9 +96,6 @@ function perflab_admin_pointer( $hook_suffix ) {
 		return;
 	}
 
-	$current_user = get_current_user_id();
-	$dismissed    = explode( ',', (string) get_user_meta( $current_user, 'dismissed_wp_pointers', true ) );
-
 	if ( in_array( 'perflab-admin-pointer', $dismissed, true ) ) {
 		return;
 	}
```
mukeshpanchal27: Move this check after line no 81 because the update-related condition is checked on all pages, so it is not needed if the pointer is already disabled.

mukeshpanchal27 (changes requested):
Thanks, @Ellusu, for the update. I left some feedback for adjustments to the conditions. Other than that, it looks solid to me.

includes/admin/load.php
Ellusu: Edit thanks :)

includes/admin/load.php
Ellusu: Edit thanks :)

mukeshpanchal27 (approved):
Thanks @Ellusu for the updates. The changes look good to me.

PR #1148, opened 2024-04-16
Prevent showing warning notice about read-only file system when all plugins are installed and fix plugin links

### LINKED ISSUE #1138: Warning notice about read-only file system shown even when all plugins are installed
## Bug Description

On my site hosted by Pantheon, I have all PL plugins installed, although not all are active. Nevertheless, I am still getting the warning notice which is shown when there is a read-only filesystem. This warning should not be shown when all the plugins are active. 

## Steps to reproduce

1. Install all PL plugins.
2. Make the filesystem read-only.
3. Access the Performance screen.

## Screenshots

Screenshot 2024-04-15 10 49 05

mukeshpanchal27:
## Summary

Fixes #1138

### DIFF
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -341,8 +341,14 @@ function perflab_print_features_page_style() {
 function perflab_plugin_admin_notices() {
 	if ( ! current_user_can( 'install_plugins' ) ) {
 		$are_all_plugins_active = true;
+		$installed_plugin_slugs = array_map( 
+			static function ( $name ) { 
+				return strtok( $name, '/' ); 
+			}, 
+			array_keys( get_plugins() )
+		);
 		foreach ( perflab_get_standalone_plugin_version_constants() as $plugin_slug => $constant_name ) {
-			if ( ! defined( $constant_name ) ) {
+			if ( ! in_array( $plugin_slug, $installed_plugin_slugs, true ) ) {
 				$are_all_plugins_active = false;
 				break;
 			}
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -340,13 +340,23 @@ function perflab_print_features_page_style() {
  */
 function perflab_plugin_admin_notices() {
 	if ( ! current_user_can( 'install_plugins' ) ) {
-		wp_admin_notice(
-			esc_html__( 'Due to your site\'s configuration, you may not be able to activate the performance features, unless the underlying plugin is already installed. Please install the relevant plugins manually.', 'performance-lab' ),
-			array(
-				'type' => 'warning',
-			)
-		);
-		return;
+		$is_all_plugins_actived = true;
+		foreach ( perflab_get_standalone_plugin_version_constants() as $plugin_slug => $constant_name ) {
+			if ( ! defined( $constant_name ) ) {
+				$is_all_plugins_actived = false;
+				break;
+			}
+		}
+
+		if ( ! $is_all_plugins_actived ) {
+			wp_admin_notice(
+				esc_html__( 'Due to your site\'s configuration, you may not be able to activate the performance features, unless the underlying plugin is already installed. Please install the relevant plugins manually.', 'performance-lab' ),
+				array(
+					'type' => 'warning',
+				)
+			);
+			return;
+		}
 	}
 
 	if ( isset( $_GET['activate'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -340,15 +340,15 @@ function perflab_print_features_page_style() {
  */
 function perflab_plugin_admin_notices() {
 	if ( ! current_user_can( 'install_plugins' ) ) {
-		$is_all_plugins_activated = true;
+		$are_all_plugins_active = true;
 		foreach ( perflab_get_standalone_plugin_version_constants() as $plugin_slug => $constant_name ) {
 			if ( ! defined( $constant_name ) ) {
-				$is_all_plugins_activated = false;
+				$are_all_plugins_active = false;
 				break;
 			}
 		}
 
-		if ( ! $is_all_plugins_activated ) {
+		if ( ! $are_all_plugins_active ) {
 			wp_admin_notice(
 				esc_html__( 'Due to your site\'s configuration, you may not be able to activate the performance features, unless the underlying plugin is already installed. Please install the relevant plugins manually.', 'performance-lab' ),
 				array(
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -340,8 +340,8 @@ function perflab_print_features_page_style() {
  */
 function perflab_plugin_admin_notices() {
 	if ( ! current_user_can( 'install_plugins' ) ) {
-		$are_all_plugins_active = true;
-		$installed_plugin_slugs = array_map(
+		$are_all_plugins_installed = true;
+		$installed_plugin_slugs    = array_map(
 			static function ( $name ) {
 				return strtok( $name, '/' );
 			},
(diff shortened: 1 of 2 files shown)


includes/admin/load.php
```diff
@@ -340,13 +340,23 @@ function perflab_print_features_page_style() {
...
 function perflab_plugin_admin_notices() {
 	if ( ! current_user_can( 'install_plugins' ) ) {
-		wp_admin_notice(
-			esc_html__( 'Due to your site\'s configuration, you may not be able to activate the performance features, unless the underlying plugin is already installed. Please install the relevant plugins manually.', 'performance-lab' ),
-			array(
-				'type' => 'warning',
-			)
-		);
-		return;
+		$is_all_plugins_activated = true;
```
swissspidy:
```suggestion
		$are_all_plugins_active = true;
```

Same everywhere

includes/admin/load.php
```diff
@@ -340,13 +340,23 @@ function perflab_print_features_page_style() {
...
-		wp_admin_notice(
-			esc_html__( 'Due to your site\'s configuration, you may not be able to activate the performance features, unless the underlying plugin is already installed. Please install the relevant plugins manually.', 'performance-lab' ),
-			array(
-				'type' => 'warning',
-			)
-		);
-		return;
+		$are_all_plugins_active = true;
+		foreach ( perflab_get_standalone_plugin_version_constants() as $plugin_slug => $constant_name ) {
+			if ( ! defined( $constant_name ) ) {
```
swissspidy: Wait, this checks whether all plugins are _active_, but what we need to check is whether they are all already _installed_. Or am I missing something?

westonruter: Right, this needs to rather check if `$plugin_slug` is among the installed plugins:

```suggestion
			if ( ! in_array( $plugin_slug, $installed_plugin_slugs, true ) ) {
```

includes/admin/load.php
```diff
@@ -340,13 +340,23 @@ function perflab_print_features_page_style() {
...
 function perflab_plugin_admin_notices() {
 	if ( ! current_user_can( 'install_plugins' ) ) {
-		wp_admin_notice(
-			esc_html__( 'Due to your site\'s configuration, you may not be able to activate the performance features, unless the underlying plugin is already installed. Please install the relevant plugins manually.', 'performance-lab' ),
-			array(
-				'type' => 'warning',
-			)
-		);
-		return;
+		$are_all_plugins_active = true;
```
```suggestion
		$are_all_plugins_active = true;
		$installed_plugin_slugs = array_map( 
			static function ( $name ) { 
				return strtok( $name, '/' ); 
			}, 
			array_keys( get_plugins() )
		);
```

westonruter: And `$are_all_plugins_active` should rather be `$are_all_plugins_installed`

westonruter:
This seems to work now as expected when file mods are disabled.

When not all plugins are installed:

Screenshot 2024-04-17 10 09 20

When all plugins are installed but not all are active:

Screenshot 2024-04-17 10 10 17

westonruter:
I just discovered what appears to be a closely-related issue here. When file mods are disabled, clicking the plugin name (or the Learn more link) opens a lightbox and shows a permissions error.

Expected | Actual
--|--
Screenshot 2024-04-17 10 15 16 | Screenshot 2024-04-17 10 15 27

It makes sense that the lightbox wouldn't work due to the Install/Activate button at the bottom. So when file mods are disabled, these links should rather open the plugin's directory page in a new tab instead. Either this can be done in this PR or another, but might as well do it here.

mukeshpanchal27:
Nice catch @westonruter I have updated the links as per [WP core](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/includes/class-wp-plugins-list-table.php#L1182-L1206).

Could you please take a look. Thanks!

includes/admin/plugins.php
```diff
@@ -193,27 +193,45 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+		/* translators: %s: Plugin name. */
+		$aria_label = sprintf( __( 'Visit plugin site for %s', 'default' ), $name );
+
+		$details_link = $plugin_data['homepage'];
+
+		$action_links[] = sprintf(
+			'<a href="%s" aria-label="%s" target="_blank">%s</a>',
+			esc_url( $details_link ),
+			esc_attr( $aria_label ),
+			__( 'Visit plugin site', 'default' )
```
westonruter:
```suggestion
			esc_html__( 'Visit plugin site', 'default' )
```

includes/admin/plugins.php
```diff
@@ -279,7 +297,7 @@ function perflab_render_plugin_card( array $plugin_data ) {
 		<div class="plugin-card-top">
 			<div class="name column-name">
 				<h3>
-					<a href="<?php echo esc_url( $details_link ); ?>" class="thickbox open-plugin-details-modal">
+					<a href="<?php echo esc_url( $details_link ); ?>"<?php echo wp_kses_post( $title_attr ); ?>>
```
westonruter: Since Kses isn't designed for this context, probably better to just add a `phpcs` comment to disable the rule here, since the variable is essentially a constant.

includes/admin/plugins.php
```diff
@@ -193,27 +193,45 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
-				'tab'       => 'plugin-information',
-				'plugin'    => $plugin_data['slug'],
-				'TB_iframe' => 'true',
-				'width'     => 600,
-				'height'    => 550,
-			),
-			admin_url( 'plugin-install.php' )
-		)
-	);
+	$title_attr = '';
```
westonruter: Why is this called `$title_attr`? Makes me think it is for the `title` attribute. I guess this is coming from core?

mukeshpanchal27: No. It's local variable that take `class` and `target` attributes values for the title href.

includes/admin/plugins.php
```diff
@@ -193,27 +193,45 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
+			esc_attr( $name ),
+			esc_html__( 'Learn more', 'performance-lab' )
+		);
+	} elseif ( ! empty( $plugin_data['homepage'] ) ) {
+		$title_link_attr = ' target="_blank"';
+
+		/* translators: %s: Plugin name. */
+		$aria_label = sprintf( __( 'Visit plugin site for %s', 'default' ), $name );
+
+		$details_link = $plugin_data['homepage'];
```
westonruter: I just noticed there is no `else` condition for if `$plugin_data['homepage']` is blank. In that case, it seems `$details_link` will never be defined and the reference below will be an error. PhpStorm's static analysis is warning me about this:

image

Is `homepage` ever blank? If so, I think we can fall back to using the WordPress.org plugin directory page in this case (which is always the same as the `homepage` in our plugins anyway):

```suggestion
	} else {
		$title_link_attr = ' target="_blank"';

		/* translators: %s: Plugin name. */
		$aria_label = sprintf( __( 'Visit plugin site for %s', 'default' ), $name );

		$details_link = __( 'https://wordpress.org/plugins/', 'default' ) . $plugin_data['slug'];
```

See related logic in core: https://github.com/WordPress/wordpress-develop/blob/80096ddf29d3ffa4d5654f5f788df7f598b48756/src/wp-admin/includes/plugin-install.php#L714-L717

westonruter: I'm confused by what `homepage` refers to. I guess it is always the WordPress.org plugin directory page? I guess the `Plugin URI` from the bootstrap file is not even returned in the API response?

westonruter: Oh no, the `homepage` _is_ the `Plugin URI`. In our case, this is not good because it points to the plugin directory on the GitHub repo. Therefore, this doesn't seem right. I've updated my suggestion above to not use `$plugin_data['homepage']` at all.

includes/admin/plugins.php
```diff
@@ -193,27 +193,45 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
-				'tab'       => 'plugin-information',
-				'plugin'    => $plugin_data['slug'],
-				'TB_iframe' => 'true',
-				'width'     => 600,
-				'height'    => 550,
-			),
-			admin_url( 'plugin-install.php' )
-		)
-	);
+	$title_link_attr = '';
```
westonruter: With my suggestion below accepted, this variable assignment can then be removed because it will never be used.

```suggestion
```

westonruter (approved):

joemcgill (approved):

includes/admin/plugins.php
westonruter: I just realized my suggestion lacked the trailing slash from https://github.com/WordPress/wordpress-develop/blob/364ea85faca123a56a3631594ec7667164b1482a/src/wp-admin/includes/plugin-install.php#L715

Added in cc43ccf

PR #1155, opened 2024-04-16
Update readme with browser support and FAQ section about analytics and personalization

westonruter:
Added FAQ lifted from @tunetheweb in a support topic reply: https://wordpress.org/support/topic/how-does-this-affect-pageview-tracking/#post-17686997

### DIFF
--- a/plugins/speculation-rules/readme.txt
+++ b/plugins/speculation-rules/readme.txt
@@ -88,6 +88,16 @@ add_filter(
 
 As mentioned above, adding the `no-prerender` CSS class to a link will prevent it from being prerendered (but not prefetched). Additionally, links with `rel=nofollow` will neither be prefetched nor prerendered because some plugins add this to non-idempotent links (e.g. add to cart); such links ideally should rather be buttons which trigger a POST request or at least they should use `wp_nonce_url()`.
 
+= How will this impact analytics and personalization? =
+
+Prerendering can affect analytics and personalization.
+
+For client-side JavaScript, is recommended to delay these until the page clicks and some solutions (like Google Analytics) already do this automatically for prerender. See [Impact on Analytics](https://developer.chrome.com/docs/web-platform/prerender-pages#impact-on-analytics). Additionally, cross-origin iframes are not loaded until activation which can further avoid issues here.
+
+Speculating on hover (moderate) increases the chance the page will be loaded, over preloading without this signal, and thus reduces the risk here. Alternatively, the plugin offers to only speculate on mouse/pointer down (conservative) which further reduces the risk here and is an option for sites which are concerned about this, at the cost of having less of a lead time and so less of a performance gain.
+
+A prerendered page is linked to the page that prerenders it, so personalisation may already be known by this point and changes (e.g. browsing other products, or logging in/out) may require a new page load, and hence a new prerender anyway, which will take these into account. But it definitely is something to be aware of and test!
+
 = Where can I submit my plugin feedback? =
 
 Feedback is encouraged and much appreciated, especially since this plugin may contain future WordPress core features. If you have suggestions or requests for new features, you can [submit them as an issue in the WordPress Performance Team's GitHub repository](https://github.com/WordPress/performance/issues/new/choose). If you need help with troubleshooting or have a question about the plugin, please [create a new topic on our support forum](https://wordpress.org/support/plugin/speculation-rules/#new-topic-0).
--- a/plugins/speculation-rules/readme.txt
+++ b/plugins/speculation-rules/readme.txt
@@ -23,7 +23,7 @@ A filter can be used to exclude certain URL paths from being eligible for prefet
 
 = Browser support =
 
-The Speculation Rules API is a new web API, and the specific syntax used by the plugin currently requires using Chrome or Edge (specifically Chromium 121+); other Chromium-based browsers like Opera should also work, although Arc and Brave may not work if uBlock Origin is enabled.
+The Speculation Rules API is a new web API, and the functionality used by the plugin is supported in Chromium-based browsers such as Chrome, Edge, or Opera using version 121 or above. Other browsers such as Safari and Firefox will ignore the functionality with no ill effects but will not benefit from the speculative loading. Note that extensions may disable preloading by default (for example, uBlock Origin does this) .
 
 Other browsers will not see any adverse effects, however the feature will not work for those clients.
 
--- a/plugins/speculation-rules/readme.txt
+++ b/plugins/speculation-rules/readme.txt
@@ -23,7 +23,7 @@ A filter can be used to exclude certain URL paths from being eligible for prefet
 
 = Browser support =
 
-The Speculation Rules API is a new web API, and the functionality used by the plugin is supported in Chromium-based browsers such as Chrome, Edge, or Opera using version 121 or above. Other browsers such as Safari and Firefox will ignore the functionality with no ill effects but will not benefit from the speculative loading. Note that extensions may disable preloading by default (for example, uBlock Origin does this) .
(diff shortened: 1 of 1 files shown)


tunetheweb (approved):

plugins/speculation-rules/readme.txt
```diff
@@ -23,7 +23,7 @@ A filter can be used to exclude certain URL paths from being eligible for prefet
 
 = Browser support =
 
-The Speculation Rules API is a new web API, and the specific syntax used by the plugin currently requires using Chrome 121+.
+The Speculation Rules API is a new web API, and the specific syntax used by the plugin currently requires using Chrome or Edge (specifically Chromium 121+); other Chromium-based browsers like Opera should also work, although Arc and Brave may not work if uBlock Origin is enabled.
```
westonruter: @tunetheweb I've updated the browser support section here as we've chatted about today.

adamsilverstein: How about something simpler like "...currently requires a recent Chromium based browser like Edge or Chrome"

tunetheweb: Confirmed Opera does work by default.
Confirmed Arc and Brave disable this by default but not sure if we need to specifically mention ALL browsers.

WDYT about this?:

```suggestion
The Speculation Rules API is a new web API, and the functionality used by the plugin is supported in Chromium-based browsers such as Chrome, Edge, or Opera using version 121 or above. Other browsers such as Safari and Firefox will ignore the functionality with no ill effects but will not benefit from the speculative loading. Note that extensions may disable preloading by default (for example, uBlock Origin does this) .
```

tunetheweb (approved):
LGTM

mukeshpanchal27 (approved):

plugins/speculation-rules/readme.txt
```diff
@@ -88,6 +88,16 @@ add_filter(
...
 
+= How will this impact analytics and personalization? =
+
+Prerendering can affect analytics and personalization.
+
+For client-side JavaScript, is recommended to delay these until the page clicks and some solutions (like Google Analytics) already do this automatically for prerender. See [Impact on Analytics](https://developer.chrome.com/docs/web-platform/prerender-pages#impact-on-analytics). Additionally, cross-origin iframes are not loaded until activation which can further avoid issues here.
+
+Speculating on hover (moderate) increases the chance the page will be loaded, over preloading without this signal, and thus reduces the risk here. Alternatively, the plugin offers to only speculate on mouse/pointer down (conservative) which further reduces the risk here and is an option for sites which are concerned about this, at the cost of having less of a lead time and so less of a performance gain.
```
felixarntz, line 97: I find "over preloading without this signal" complicated to understand. What's a "signal"? I definitely think the language here can be simplified.

Is this supposed to mean "speculating on hover increases the chance ... over speculating on eager"?

Also since "speculating on hover (moderate)" is the default in this plugin anyway, I think it's worth formulating this response with that context, just to reiterate here that this is the default that the plugin uses.

plugins/speculation-rules/readme.txt
```diff
@@ -88,6 +88,16 @@ add_filter(
 
 As mentioned above, adding the `no-prerender` CSS class to a link will prevent it from being prerendered (but not prefetched). Additionally, links with `rel=nofollow` will neither be prefetched nor prerendered because some plugins add this to non-idempotent links (e.g. add to cart); such links ideally should rather be buttons which trigger a POST request or at least they should use `wp_nonce_url()`.
 
+= How will this impact analytics and personalization? =
+
+Prerendering can affect analytics and personalization.
+
+For client-side JavaScript, is recommended to delay these until the page clicks and some solutions (like Google Analytics) already do this automatically for prerender. See [Impact on Analytics](https://developer.chrome.com/docs/web-platform/prerender-pages#impact-on-analytics). Additionally, cross-origin iframes are not loaded until activation which can further avoid issues here.
```
felixarntz, line 95: What is "until the page clicks"? Is that the most suitable term here? I would assume there's an easier-to-understand way to indicate that the page has actually become active / that the user is actually visiting the page.

plugins/speculation-rules/readme.txt
```diff
@@ -88,6 +88,16 @@ add_filter(
...
+
+Speculating on hover (moderate) increases the chance the page will be loaded, over preloading without this signal, and thus reduces the risk here. Alternatively, the plugin offers to only speculate on mouse/pointer down (conservative) which further reduces the risk here and is an option for sites which are concerned about this, at the cost of having less of a lead time and so less of a performance gain.
+
+A prerendered page is linked to the page that prerenders it, so personalisation may already be known by this point and changes (e.g. browsing other products, or logging in/out) may require a new page load, and hence a new prerender anyway, which will take these into account. But it definitely is something to be aware of and test!
```
felixarntz, line 99: What can be done if the test leads to discovering problems? I think something should be included about that, e.g. something about modifying the personalization logic or, if that's not possible/feasible, excluding the relevant pages from speculative loading (via the filter the plugin provides).

felixarntz (commented):
@westonruter I realize this was merged, but had it open for a while and wanted to take a look now. Leaving a bit of feedback. If there's consensus, we should be able to quickly address in a follow up PR.

cc @tunetheweb

westonruter:
@felixarntz I'll defer to @tunetheweb as he's the wordsmith here.

westonruter:
Maybe we should copy the readme into a Google Doc to better collaborate on the copy (and leverage the new [Copy/Paste as Markdown feature](https://support.google.com/docs/answer/12014036?hl=en)!)

plugins/speculation-rules/readme.txt
tunetheweb: Fair point. The term we usually use is "when the prerender is activated"/"activation" since this matches the [`activationStart ` time in the Performance Navigation Timing API](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/activationStart)

```suggestion
For client-side JavaScript, it is recommended to delay these until the prerender is activated (for example but clicking on the link) and some solutions (like Google Analytics) already do this automatically for prerender. See [Impact on Analytics](https://developer.chrome.com/docs/web-platform/prerender-pages#impact-on-analytics). Additionally, cross-origin iframes are not loaded until activation which can further avoid issues here.
```

plugins/speculation-rules/readme.txt
tunetheweb: It's supposed to me "speculating on hover increases the chance of the page actually being visited, and so reduces the chance of any negative impact on analytics due to pages being prerendered (and being included in analytics) but not actually being seen by a user".

I agree the wording could be clearer.

And btw, it may should a bit dismissive to say "hey, if there's a bigger chance of a page being visited, then it's OK if your analytics are off sometimes", but the reality is this already happens in many cases. For example, people opening in a new tab and not actually visiting it. Or open a link, not reading it, but intending to come back to it (but never doing that).

plugins/speculation-rules/readme.txt
tunetheweb: The options (from best to worse IMHO) are:
- Cancel speculations on personalization changes (note this happens automatically on navigations anyway, which is probably when most personalization signal changes and other state changes happen anyway).
- Update on page activation with client-side JS (plus: also could handle stale background tabs, negative: can lead to a flash of old content(
- Don't speculation on hover and only do on mousedown / "conservative" eagerness. (negative, less gain).
- Exclude personalized pages (but if this is a loging widget for example that might be ALL pages)
- Don't use speculation rules on those types of sites.

Probably not a bad idea to spell these out.

westonruter:
I created a doc and shared it with ya'll.

westonruter:
I've opened #1459 to track this work.

PR #1162, opened 2024-04-18
Use PHP 7.2 features in Optimization Detective

westonruter:
In #1130 the PHP version was bumped to 7.2. This allows us to make use of new PHP language features which further harden static analysis (related to #775). Changes in this PR include:

* Use the `void` return type
* Use the nullable type
* Add value types for arrays
* Add improved typing to prevent having to ignore the following errors in PHPStan level 6:
  * `#^(Function|Method) .+? return type has no value type specified in iterable type array#`
  * `#^(Function|Method) .+? has parameter .+? with no value type specified in iterable type array#`
  * `#^(Function|Method) .+? has no return type specified#`

### DIFF
--- a/plugins/optimization-detective/class-od-url-metric.php
+++ b/plugins/optimization-detective/class-od-url-metric.php
@@ -61,7 +61,7 @@ final class OD_URL_Metric implements JsonSerializable {
 	/**
 	 * Gets JSON schema for URL Metric.
 	 *
-	 * @return array Schema.
+	 * @return array<string, mixed> Schema.
 	 */
 	public static function get_json_schema(): array {
 		$dom_rect_schema = array(
--- a/plugins/optimization-detective/class-od-url-metric.php
+++ b/plugins/optimization-detective/class-od-url-metric.php
@@ -46,6 +46,7 @@ final class OD_URL_Metric implements JsonSerializable {
 	 * Constructor.
 	 *
 	 * @param array $data URL metric data.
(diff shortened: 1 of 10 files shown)


thelovekesh (approved):

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -46,6 +46,7 @@ final class OD_URL_Metric implements JsonSerializable {
 	 * Constructor.
 	 *
 	 * @param array $data URL metric data.
+	 * @phpstan-param Data $data URL metric data.
 	 *
```
joemcgill: Lets separate custom params from other params. See [this example](https://github.com/WordPress/plugin-check/blob/trunk/includes/Checker/Check_Collection.php#L40-L51).

```suggestion
	 *
	 * @phpstan-param Data $data URL metric data.
	 *
	 * @param array $data URL metric data.
	 *
```

plugins/optimization-detective/optimization.php
```diff
@@ -27,7 +27,7 @@
  * @access private
  * @link https://core.trac.wordpress.org/ticket/43258
  *
- * @param string $passthrough Optional. Filter value. Default null.
+ * @param string $passthrough Value for the template_include filter which is passed through.
```
joemcgill, line 30: Good catch ✨

joemcgill (approved):
Just one suggestion, but looks good. Pre-approving.

PR #1143, opened 2024-04-15
Exclude _wpnonce URLs in speculation rules

westonruter:
_This is a sub-PR of #1142._

Fixes #1141

Before | After
--|--
Screenshot 2024-04-15 14 49 27 | Screenshot 2024-04-15 14 49 33

The URLPattern used to exclude URLs with `_wpnonce` is `/*\?*(^|&)_wpnonce=*`. Take special note that the second `*` wildcard only matches chars in the query segment. Additionally, the `^` anchors the match to the beginning of the `search` segment, and not to the beginning of the entire string. (Remember: `URLPattern` is not regex.) This means that only the `_wpnonce` query parameter will matched, whether it is the first query parameter or not. It also won't incorrectly match `other_wpnonce`.

The following URLs were tested to be properly included and excluded:

URL | Status | URLPattern Tester
--|--|--
`http://example.com/foo/?_wpnonce=1234` | Excluded | [Test](https://urlpattern.com/#eyJ1cmxQYXR0ZXJuIjoiaHR0cDovL2V4YW1wbGUuY29tLypcXD8qKF58Jilfd3Bub25jZT0qIiwidXJsIjoiaHR0cDovL2V4YW1wbGUuY29tL2Zvby8/X3dwbm9uY2U9MTIzNCJ9)
`http://example.com/?_wpnonce=1234` | Excluded | [Test](https://urlpattern.com/#eyJ1cmxQYXR0ZXJuIjoiaHR0cDovL2V4YW1wbGUuY29tLypcXD8qKF58Jilfd3Bub25jZT0qIiwidXJsIjoiaHR0cDovL2V4YW1wbGUuY29tLz9fd3Bub25jZT0xMjM0In0=)
`http://example.com/foo/?amp=1&_wpnonce=1234` | Excluded | [Test](https://urlpattern.com/#eyJ1cmxQYXR0ZXJuIjoiaHR0cDovL2V4YW1wbGUuY29tLypcXD8qKF58Jilfd3Bub25jZT0qIiwidXJsIjoiaHR0cDovL2V4YW1wbGUuY29tL2Zvby8/YW1wPTEmX3dwbm9uY2U9MTIzNCJ9)
[...]

PR #1164, opened 2024-04-19
Fix composition of href exclude paths to account for JSON encoding and differing site/home URLs

### LINKED ISSUE #1163: Speculative Loading invalid key for a URL pattern error - exclude_paths filter
## Bug Description

Trying to simply exclude a URL pattern and running into what appears to be a bug with the `plsr_speculation_rules_href_exclude_paths` filter.

Getting this error:

> Invalid key "0" for a URL pattern object found.

image

## Steps to reproduce

1. Fresh WordPress Install with Speculative Loading plugin running. 
2. Add a filter with the following exclusion and URL pattern. Using example from WordPress.org docs ([source](https://wordpress.org/plugins/speculation-rules/#how%20can%20i%20prevent%20certain%20urls%20from%20being%20prefetched%20and%20prerendered%3F)).

```php
add_filter(
    'plsr_speculation_rules_href_exclude_paths',
    function ( array $exclude_paths ): array {
        $exclude_paths[] = '/go/*';
        return $exclude_paths;
    }
);
```

3. Log out of the site and test in Chrome DevTools. Error should flag. 

image

## Additional Context

 - PHP Version: 8.2
 - OS: macOS Sonoma 14.4.1
 - Browser: Chrome Version 123.0.6312.124
 - Plugin Version: 1.2.1

If I'm doing something wrong, feel free to clarify and close the issue. Thanks!

westonruter:
## Summary

When filtering `plsr_speculation_rules_href_exclude_paths` to add additional paths, make sure the resulting de-duplicated array has sequential keys starting from zero. If this is not the case, then `json_encode()` interprets the array as an object which causes a speculation rules failure in the browser.

Additionally, we found that the base exclude hrefs were getting prefixed with the site URL path and then prefixed again with the home URL path with the merge with the additional filtered exclude href exclude paths. On sites with different site URL and home URL (e.g. `/wp/` and `/blog/`) this can result in the base URLs getting double-prefixed in an erroneous prefix (e.g. `/blog/wp/`). This is fixed by avoiding the redundant prefixing, so that only the href exclude paths provided by the filter are prefixed with the home URL path. This PR also ensures that the nonce/`_wpnonce` action URLs (from #1143) are prefixed with the home URL path and not the site URL path.

Fixes #1163

## Relevant technical choices

The result of calling `array_unique()` is passed into `array_values()` to ensure that the keys are numerically sequential from zero. Tests are updated to ensure this is the case.

Additionally, the `$base_href_exclude_paths` are no longer provided as the initial value to pass into the `plsr_speculation_rules_href_exclude_paths` filter. The values in `$base_href_exclude_paths` get merged again with the result of the filter, meaning they will always be present regardless. It is misleading to pass them into the filter, as a developer may think they can remove a path by eliminating it from the array passed to the filter when they cannot. Eliminating this redundancy alone fixes the symptom of non-sequential array keys, but the changes in this PR go further to ensure there is no possibility of non-sequential array keys when the the base exclude paths are merged with the return value from the filter.

### DIFF
--- a/plugins/speculation-rules/helper.php
+++ b/plugins/speculation-rules/helper.php
@@ -58,7 +58,7 @@ function plsr_get_speculation_rules() {
 	 *
 	 * @param string[] $href_exclude_paths Additional paths to disable speculative prerendering for. The base exclude paths,
 	 *                                     such as for wp-admin, cannot be removed.
-	 * @param string $mode                 Mode used to apply speculative prerendering. Either 'prefetch' or 'prerender'.
+	 * @param string   $mode               Mode used to apply speculative prerendering. Either 'prefetch' or 'prerender'.
 	 */
 	$href_exclude_paths = (array) apply_filters( 'plsr_speculation_rules_href_exclude_paths', array(), $mode );
 
--- a/plugins/speculation-rules/helper.php
+++ b/plugins/speculation-rules/helper.php
@@ -42,7 +42,7 @@ function plsr_get_speculation_rules() {
 	$base_href_exclude_paths = array(
 		$prefixer->prefix_path_pattern( '/wp-login.php', 'site' ),
 		$prefixer->prefix_path_pattern( '/wp-admin/*', 'site' ),
-		$prefixer->prefix_path_pattern( '/*\\?*(^|&)_wpnonce=*', 'site' ),
+		$prefixer->prefix_path_pattern( '/*\\?*(^|&)_wpnonce=*', 'home' ),
 	);
 
 	/**
--- a/plugins/speculation-rules/helper.php
+++ b/plugins/speculation-rules/helper.php
@@ -62,13 +62,18 @@ function plsr_get_speculation_rules() {
 	 */
 	$href_exclude_paths = (array) apply_filters( 'plsr_speculation_rules_href_exclude_paths', $href_exclude_paths, $mode );
 
-	// Ensure that there are no duplicates and that the base paths cannot be removed.
-	$href_exclude_paths = array_unique(
-		array_map(
-			array( $prefixer, 'prefix_path_pattern' ),
-			array_merge(
-				$base_href_exclude_paths,
-				$href_exclude_paths
+	// Ensure that:
+	// 1. There are no duplicates.
+	// 2. The base paths cannot be removed.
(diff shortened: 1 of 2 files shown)


westonruter:
Build for testing: [speculation-rules.zip](https://github.com/WordPress/performance/files/15033767/speculation-rules.zip)

tunetheweb (approved):
LGTM form a spec rules perspective but will leave others to comment on the PHP side of things.

adamsilverstein (approved):
Changes look good, thanks for the tests!

plugins/speculation-rules/helper.php
```diff
@@ -57,18 +56,24 @@ function plsr_get_speculation_rules() {
...
+	// Ensure that:
+	// 1. There are no duplicates.
+	// 2. The base paths cannot be removed.
+	// 3. The array has sequential keys (i.e. array_is_list()).
+	$href_exclude_paths = array_values(
+		array_unique(
+			array_map(
+				array( $prefixer, 'prefix_path_pattern' ),
+				array_merge(
+					$base_href_exclude_paths,
```
joemcgill: Now that `$base_href_exclude_paths` are being run through `$prefixer::prefix_path_pattern` here, should we avoid doing so above?

joemcgill (approved):
One question, but looks good.

plugins/speculation-rules/helper.php
westonruter: Oh! I think you identified another bug, specifically when the `site_url` and `home_url` are different. I've added a test case to show this: 74e52a21 & d77b46c1. The problem is that the base exclude paths for wp-admin and wp-login are initially prefixed with the site URL, but then later after merging with the filter-supplied exclude paths, it is getting prefixed _again_ and this time with the home URL. The result is the prefix from both the home URL _and_ the site URL is getting applied to the base URLs. I've proposed a fix for this in 01d9153aa7701d02198ac90c1102479345e7c449. Please check.

westonruter:
Updated build for testing: [speculation-rules.zip](https://github.com/WordPress/performance/files/15043596/speculation-rules.zip)

joemcgill (approved):
This update looks good to me. Thanks for testing!

Issue #1157, opened 2024-04-17
Speculative loading considerations


AntonyXSI:
## Bug Description
I have used prefetching on hundreds of WordPress sites for 5+ years with great success to CWV metrics and improving the user experience in general. Good to see this being explored. It's great a feature, however I don't believe that speculative loading would be suitable as an on by default core feature. Due to it having the ability to decrease scalability by increasing server resource usage, particularly for sites on shared hosting servers. It really needs to be configured depending on the site/server resources available therefore would be best suited as a plugin.

Happy to share what default settings has worked for the sites I use it on:

1. Excluding logged in users. Prefetching really shines with page caching as the server is able to deliver the content almost instantly with near zero CPU usage without waiting for the page to be generated. Logged in visits are predominately dynamic non cacheable requests. 

Using prefetching/prerendering on dynamic pages specifically on shared hosting where it can often take 1s+ for a page to be generated, would result in higher CPU usage and resources more easily bottlenecking. This is from pages being requested that may not be visited, as well as the ease/rate for links to be requested by hovering over them rather than manually when navigating through the site, and fundamentally the dynamic requests taking a lot longer. This affects the ability of the server to handle the same amount of traffic, and traffic spikes. 

2. Excluding URLs with query strings. This would solve needing to manually exclude the nonce, add to cart strings etc. Another example would be plugins using href links for filters, wishlists etc. These plugins don't always abide by the practice of adding the nofollow tag to links. Query parameters URLs are often excluded by page caching plugins as well. Including these would increase CPU usage and potentially cause undesirable actions (adding items to the cart on). A blanket exclude seems to be the best option to avoid compatibility issues with different plugins and custom code.

3. Excluding links that contain "." or file extensions i.e to prevent a user from downloading large PDFs or images potentially bottlenecking bandwidth. An example would be images being wrapped in a link to the original uncropped image. Hovering over images in a gallery for example could result in lots of large unoptimized images being downloaded unnecessarily in the background. As images will begin to render almost instantly prerendering images, pdfs etc may not be as beneficial as prerendering pages. I'm also not sure what happens if .exe,.pdf or other large files are prefetched that would normally be downloaded once clicked? Does Chrome drop the request or continue the download from the same prefetch request? (Note: Excluding the uploads directory is implemented in #1167.)

[...]

Issue #1156, opened 2024-04-16
Entirety of speculation rules should be filterable, not just the paths to exclude


westonruter:
## Feature Description

Currently the Speculative Loading plugin provides a `plsr_speculation_rules_href_exclude_paths` filter to exclude URLs (and URL patterns) from being speculatively loaded. However, two situations came up recently where this was not sufficient. 

1. In #1140, WooCommerce was found to create non-idempotent add-to-cart links which add products to cart when prefetched/prerendered. WooCommerce seems to try to get around the non-idempotency by adding `rel=nofollow` to the links. However, there is no way to exclude links via such an attribute without manually modifying the default rules, which was done in https://github.com/WordPress/performance/pull/1142. (This was to avoid having to add a WooCommerce-specific URLPattern to exclude URLs with an `add-to-cart` query param, since excluding links with `rel=nofollow` may be generally advisable: https://github.com/WICG/nav-speculation/issues/309).
2. In a [comment](https://github.com/WordPress/performance/issues/1140#issuecomment-2060034357) on the aforementioned issue, a user wanted to customize the speculation rules so that specific URLs were opted-in for eager prerendering. That is, instead of specifying a list of URLs/URLPatterns to _exclude_, they wanted to provide a list of URLs/URLPatterns to _include_. This is not possible at present either.

To account for these two use cases, I suggest that the entire set of speculation rules (speculation ruleset?) be filterable, doing something like this:

```diff
--- a/plugins/speculation-rules/helper.php
+++ b/plugins/speculation-rules/helper.php
@@ -109,5 +109,13 @@ function plsr_get_speculation_rules() {
 		);
 	}
 
-	return array( $mode => $rules );
+	$ruleset = array( $mode => $rules );
+
+	/**
+	 * Filters the entire speculation ruleset.
+	 *
+	 * @param array  $ruleset Ruleset.
+	 * @param string $mode    Mode used to apply speculative prerendering. Either 'prefetch' or 'prerender'.
+	 */
+	return (array) apply_filters( 'plsr_speculation_ruleset', $ruleset, $mode );
 }
```

Also, @swissspidy suggested in https://github.com/WordPress/performance/issues/1144#issuecomment-2058311409 that a JSON Schema could be written which could validate whatever is being returned by the filter. If not valid, it could trigger a `_doing_it_wrong()`.

westonruter:
[...]

PR #1167, opened 2024-04-19
Prevent speculatively loading links to the uploads, content, plugins, template, or stylesheet directories

westonruter:
This was brought up by @AntonyXSI in #1157:

> 3\. Excluding links that contain "." or file extensions i.e to prevent a user from downloading large PDFs or images potentially bottlenecking bandwidth. An example would be images being wrapped in a link to the original uncropped image. Hovering over images in a gallery for example could result in lots of large unoptimized images being downloaded unnecessarily in the background. As images will begin to render almost instantly prerendering images, pdfs etc may not be as beneficial as prerendering pages. I'm also not sure what happens if .exe,.pdf or other large files are prefetched that would normally be downloaded once clicked? Does Chrome drop the request or continue the download from the same prefetch request?

See Chrome's behavior for speculative loading of an image and a PDF: https://github.com/WordPress/performance/issues/1157#issuecomment-2065459362.

In this PR, the uploads directory is excluded from speculative loading. The uploads directory path is added to the base href excludes.

Before | After
--|--
image | image

What if the uploads directory is pointing to a CDN instead? Well, it will be ignored anyway because the speculation rules are configured to only do same-origin speculative loads:

https://github.com/WordPress/performance/blob/cb5a37e8871e06564ac23e793b2121d6181dc0b8/plugins/speculation-rules/helper.php#L83-L91

So they will be excluded anyway. (The same goes for a site with a site customizing the `site_url` to point to another domain, cf. #1164). But what if someone _wants_ to do cross-origin speculative loads? That could be made possible with #1156 where the `href_matches` is changed from `/*` to instead be `*://*/*`. Nevertheless, it seems Chrome does not currently support this:

image

Nevertheless, the above was specifically for prerendering. If I try instead with the mode set to prefetch, I get different results:

image

So it _can_ work with prefetch in some situations.

We should keep that in mind for the future.

### DIFF
--- a/plugins/speculation-rules/class-plsr-url-pattern-prefixer.php
+++ b/plugins/speculation-rules/class-plsr-url-pattern-prefixer.php
@@ -107,8 +107,9 @@ class PLSR_URL_Pattern_Prefixer {
 	 */
 	public static function get_default_contexts(): array {
 		return array(
-			'home' => self::escape_pattern_string( trailingslashit( wp_parse_url( home_url( '/' ), PHP_URL_PATH ) ) ),
-			'site' => self::escape_pattern_string( trailingslashit( wp_parse_url( site_url( '/' ), PHP_URL_PATH ) ) ),
+			'home'    => self::escape_pattern_string( trailingslashit( wp_parse_url( home_url( '/' ), PHP_URL_PATH ) ) ),
+			'site'    => self::escape_pattern_string( trailingslashit( wp_parse_url( site_url( '/' ), PHP_URL_PATH ) ) ),
+			'uploads' => self::escape_pattern_string( trailingslashit( wp_parse_url( wp_upload_dir( null, false )['baseurl'], PHP_URL_PATH ) ) ),
 		);
 	}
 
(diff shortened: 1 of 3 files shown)


tunetheweb (approved):

felixarntz:
@westonruter Would it make sense to instead prevent speculative loading of any links to something within `wp-content`? Uploads is probably most important, but I'm not sure we'd want to speculatively load anything bundled in a plugin or theme either, as it wouldn't be a regular WordPress URL.

westonruter:
@felixarntz yeah, good point. That should be added as another entry specific to `wp-content` since the `uploads` directory can be configured to point somewhere else.

westonruter:
For completeness, the following URLs should be excluded: `WP_CONTENT_URL`, `WP_PLUGIN_URL`, `get_stylesheet_directory_uri()`, and `get_template_directory_uri()`. All of these can be overridden from their defaults.

adamsilverstein (approved):
Nice work!

tunetheweb (approved):
LGTM

westonruter:
Build for testing: [speculation-rules.zip](https://github.com/WordPress/performance/files/15240067/speculation-rules.zip)

PR #1168, opened 2024-04-20
WPCS improvement

mrmakadia94:
## Summary

Fixes #

## Relevant technical choices

- use  unslashed and sanitization for `$_GET['file']`
- use `$plugin_data['requires'] ?? null` to reduce cognitive load.

### DIFF
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -248,7 +248,7 @@ function perflab_install_activate_plugin_callback() {
 		wp_die( esc_html__( 'Invalid plugin.', 'performance-lab' ) );
 	}
 
-	$is_plugin_installed = isset( $_GET['file'] ) && sanitize_text_field( wp_unslash( $_GET['file'] ) );
+	$is_plugin_installed = isset( $_GET['file'] ) && $_GET['file'];
 
 	// Install the plugin if it is not installed yet.
 	if ( ! $is_plugin_installed ) {
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -248,7 +248,7 @@ function perflab_install_activate_plugin_callback() {
 		wp_die( esc_html__( 'Invalid plugin.', 'performance-lab' ) );
 	}
 
-	$is_plugin_installed = isset( $_GET['file'] ) && $_GET['file'];
+	$is_plugin_installed = isset( $_GET['file'] ) && sanitize_text_field( wp_unslash( $_GET['file'] ) );
 
 	// Install the plugin if it is not installed yet.
 	if ( ! $is_plugin_installed ) {
--- a/includes/admin/plugins.php
+++ b/includes/admin/plugins.php
@@ -142,8 +142,8 @@ function perflab_render_plugin_card( array $plugin_data ) {
 	$version     = $plugin_data['version'];
 	$name        = wp_strip_all_tags( $title . ' ' . $version );
 
-	$requires_php = isset( $plugin_data['requires_php'] ) ? $plugin_data['requires_php'] : null;
-	$requires_wp  = isset( $plugin_data['requires'] ) ? $plugin_data['requires'] : null;
+	$requires_php = $plugin_data['requires_php'] ?? null;
+	$requires_wp  = $plugin_data['requires'] ?? null;
 
 	$compatible_php = is_php_version_compatible( $requires_php );
 	$compatible_wp  = is_wp_version_compatible( $requires_wp );
--- a/includes/admin/plugins.php
+++ b/includes/admin/plugins.php
@@ -142,8 +142,8 @@ function perflab_render_plugin_card( array $plugin_data ) {
 	$version     = $plugin_data['version'];
 	$name        = wp_strip_all_tags( $title . ' ' . $version );
 
-	$compatible_php = is_php_version_compatible( $plugin_data['requires_php'] );
-	$compatible_wp  = is_wp_version_compatible( $plugin_data['requires'] );
+	$compatible_php = ! $plugin_data['requires_php'] || is_php_version_compatible( $plugin_data['requires_php'] );
+	$compatible_wp  = ! $plugin_data['requires'] || is_wp_version_compatible( $plugin_data['requires'] );
 	$action_links   = array();
 
 	$status = install_plugin_install_status( $plugin_data );
--- a/includes/admin/plugins.php
+++ b/includes/admin/plugins.php
@@ -142,11 +142,8 @@ function perflab_render_plugin_card( array $plugin_data ) {
 	$version     = $plugin_data['version'];
 	$name        = wp_strip_all_tags( $title . ' ' . $version );
 
-	$requires_php = $plugin_data['requires_php'] ?? null;
-	$requires_wp  = $plugin_data['requires'] ?? null;
-
-	$compatible_php = is_php_version_compatible( $requires_php );
-	$compatible_wp  = is_wp_version_compatible( $requires_wp );
+	$compatible_php = is_php_version_compatible( $plugin_data['requires_php'] );
+	$compatible_wp  = is_wp_version_compatible( $plugin_data['requires'] );
 	$action_links   = array();
 
 	$status = install_plugin_install_status( $plugin_data );


includes/admin/load.php
```diff
@@ -248,7 +248,7 @@ function perflab_install_activate_plugin_callback() {
 		wp_die( esc_html__( 'Invalid plugin.', 'performance-lab' ) );
 	}
 
-	$is_plugin_installed = isset( $_GET['file'] ) && $_GET['file'];
+	$is_plugin_installed = isset( $_GET['file'] ) && sanitize_text_field( wp_unslash( $_GET['file'] ) );
```
mukeshpanchal27: Since we're only checking if the file is not null and we're already sanitizing the value in line 298, adding sanitization here would be redundant.

includes/admin/plugins.php
```diff
@@ -142,8 +142,8 @@ function perflab_render_plugin_card( array $plugin_data ) {
 	$version     = $plugin_data['version'];
 	$name        = wp_strip_all_tags( $title . ' ' . $version );
 
-	$requires_php = isset( $plugin_data['requires_php'] ) ? $plugin_data['requires_php'] : null;
-	$requires_wp  = isset( $plugin_data['requires'] ) ? $plugin_data['requires'] : null;
+	$requires_php = $plugin_data['requires_php'] ?? null;
+	$requires_wp  = $plugin_data['requires'] ?? null;
```
mukeshpanchal27: Not blocker: WordPress code did not support it at the moment so we can do this changes in future when WP officially support it.

mukeshpanchal27 (changes requested):
Thanks for the PR, @mrmakadia94. The changes you proposed seem redundant. Please review my feedback.

includes/admin/plugins.php
westonruter: Actually, `??` is being used in WordPress core: https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop+%2F+%5C%3F%5C%3F+%2F&type=code

We can use this "null coalescing operator" since PHP7 is the minimum supported version.

includes/admin/load.php
westonruter: It seems strange that the presence of the `file` parameter is the signal that the plugin is installed or not. This is coming from core logic?

westonruter: No, I see it is in `perflab_render_plugin_card()`. 

@mukeshpanchal27 I think we can harden/simplify things by only using the slug. This will eliminate the need for the `file` query parameter, and we add additional validation to restrict the slug to only those which are found in `plugins.json`. For example:

```diff
diff --git a/includes/admin/load.php b/includes/admin/load.php
index d4a12608..c179945c 100644
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -247,8 +247,18 @@ function perflab_install_activate_plugin_callback() {
 	if ( ! $plugin_slug ) {
 		wp_die( esc_html__( 'Invalid plugin.', 'performance-lab' ) );
 	}
-
-	$is_plugin_installed = isset( $_GET['file'] ) && $_GET['file'];
+	// TODO: Make sure that $plugin_slug is inside of plugins.json. If not, then call wp_die().
+
+	// Check if installed and determine the plugin basename.
+	$is_plugin_installed = false;
+	$plugin_basename     = null;
+	foreach ( array_keys( get_plugins() ) as $plugin_file ) {
+		if ( strtok( $plugin_file, '/' ) === $plugin_slug ) {
+			$is_plugin_installed = true;
+			$plugin_basename     = $plugin_file; // TODO: The variable name "$plugin_basename" seems misleading. To follow core convention, it should be "$plugin_file", right?
+			break;
+		}
+	}
 
 	// Install the plugin if it is not installed yet.
 	if ( ! $is_plugin_installed ) {
@@ -294,8 +304,6 @@ function perflab_install_activate_plugin_callback() {
 
 		$plugin_file_names = array_keys( $plugins );
 		$plugin_basename   = $plugin_slug . '/' . $plugin_file_names[0];
-	} else {
-		$plugin_basename = sanitize_text_field( wp_unslash( $_GET['file'] ) );
 	}
 
 	if ( ! current_user_can( 'activate_plugin', $plugin_basename ) ) {
diff --git a/includes/admin/plugins.php b/includes/admin/plugins.php
index 1721cf41..5436cbf5 100644
--- a/includes/admin/plugins.php
+++ b/includes/admin/plugins.php
@@ -174,7 +174,6 @@ function perflab_render_plugin_card( array $plugin_data ) {
 					'action'   => 'perflab_install_activate_plugin',
 					'_wpnonce' => wp_create_nonce( 'perflab_install_activate_plugin' ),
 					'slug'     => $plugin_data['slug'],
-					'file'     => $status['file'],
 				),
 				admin_url( 'options-general.php' )
 			)

```

mrmakadia94: Thanks @mukeshpanchal27 for the guide regarding redundant changes, I remove those changes on PR.

includes/admin/plugins.php
mukeshpanchal27: Yes, some code utilizes it, but let's check some discussions regarding it.

- https://core.trac.wordpress.org/ticket/58874
- https://wordpress.slack.com/archives/C02RQBWTW/p1695705092378319?thread_ts=1695704573.754599&cid=C02RQBWTW
- https://github.com/WordPress/wordpress-develop/pull/5302

westonruter: It seems the main concern is back porting to old WP versions. Given that is not relevant here for new Performance features, I say we keep it.

mukeshpanchal27: Ok. Let's keep it then.

includes/admin/load.php
westonruter: I've submitted my above patch as https://github.com/WordPress/performance/pull/1170

westonruter (approved):

includes/admin/plugins.php
```diff
@@ -142,8 +142,8 @@ function perflab_render_plugin_card( array $plugin_data ) {
...
 
 	$compatible_php = is_php_version_compatible( $requires_php );
 	$compatible_wp  = is_wp_version_compatible( $requires_wp );
```
westonruter: Actually, I found out that the `requires_php` and `requires` are _always_ set. If they aren't present in the plugin, then they are returned in the API response as `false`. Take, for example, this ancient plugin which has no `Requires` nor `Requires PHP`: https://wordpress.org/plugins/automatic-glossary/

See the API response: https://api.wordpress.org/plugins/info/1.0/automatic-glossary.json

It includes:

```
  "requires": false,
  "requires_php": false,
  "requires_plugins": [],
```

Therefore, we can simplify this further. Instead of:

```php
	$requires_php = $plugin_data['requires_php'] ?? null;
	$requires_wp  = $plugin_data['requires'] ?? null;

	$compatible_php = is_php_version_compatible( $requires_php );
	$compatible_wp  = is_wp_version_compatible( $requires_wp );
```

This can rather be:

```php
	$compatible_php = ! $plugin_data['requires_php'] || is_php_version_compatible( $plugin_data['requires_php'] );
	$compatible_wp  = ! $plugin_data['requires'] || is_wp_version_compatible( $plugin_data['requires'] );
```

Technically it could be even simpler as:

```php
	$compatible_php = is_php_version_compatible( $plugin_data['requires_php'] );
	$compatible_wp  = is_wp_version_compatible( $plugin_data['requires'] );
```

However the phpdoc for these functions says it takes a `string` as input, while it actually works with `false` as well due to `empty()` checks.

westonruter (changes requested):

includes/admin/plugins.php
westonruter: I've added an array shape to specify what the data looks like in another PR: #1175

mrmakadia94: @westonruter updated the changes as per request.

includes/admin/plugins.php
```diff
@@ -142,11 +142,8 @@ function perflab_render_plugin_card( array $plugin_data ) {
 	$version     = $plugin_data['version'];
 	$name        = wp_strip_all_tags( $title . ' ' . $version );
 
-	$requires_php = isset( $plugin_data['requires_php'] ) ? $plugin_data['requires_php'] : null;
-	$requires_wp  = isset( $plugin_data['requires'] ) ? $plugin_data['requires'] : null;
-
-	$compatible_php = is_php_version_compatible( $requires_php );
-	$compatible_wp  = is_wp_version_compatible( $requires_wp );
+	$compatible_php = is_php_version_compatible( $plugin_data['requires_php'] );
+	$compatible_wp  = is_wp_version_compatible( $plugin_data['requires'] );
```
westonruter: @mrmakadia94 Actually, that last [suggestion](https://github.com/WordPress/performance/pull/1168#discussion_r1576963271), while it works, it technically is not correct because `is_php_version_compatible()` and `is_wp_version_compatible()` are only supposed to take strings. So I believe this is right:

```suggestion
	$compatible_php = ! $plugin_data['requires_php'] || is_php_version_compatible( $plugin_data['requires_php'] );
	$compatible_wp  = ! $plugin_data['requires'] || is_wp_version_compatible( $plugin_data['requires'] );
```

mrmakadia94: @westonruter please check now, i updated the PR.

westonruter (approved):
Thanks!

mukeshpanchal27 (approved):

PR #1172, opened 2024-04-23
Refactor optimization logic in Optimization Detective

westonruter:
This PR refactors the logic to facilitate extracting optimization logic into separate dependent plugins, for example Image Prioritizer (#1088).

To review the changes here, I suggest reviewing commit-by-commit, as I've carefully rebased changes to keep each commit as atomic as possible.

The biggest change is the introduction of the `OD_Preload_Link_Collection` class. Instead of building up an array data structure while walking over the document and then manually constructing HTML `link` tags once iteration is over, the use of this new class makes the logic much clearer as the pending preload links are added during iteration in a well-defined object interface. 

Additionally, instead of passing around an array of LCP element data arrays keyed by the minimum viewport widths (via `od_get_lcp_elements_by_minimum_viewport_widths()`), this logic is now moved over to a more logical place as a `get_lcp_element` method on the `OD_URL_Metrics_Group` class. The minimum viewport and maximum viewport can be obtained via the methods on the `OD_URL_Metrics_Group` instance. This simplifies construction of the preload links, as each group knows its minimum and maximum viewport width, without having to iterate over other keys to discover which one is the next-largest.

Also, there was a somewhat subtle distinction when obtaining the LCP element data for a given viewport group: it could return an `array`, `false`, or `null`. The array case meant that there was an LCP element found. The `false` case meant that there were URL Metrics gathered, but no supported LCP element was present. Lastly, the `null` meant that there was no data available at all. In c6681cfb0673c266eb21d2f864cfdfbe926f1e7f and 07e19c1253d549cf7e26db2744cda805d35ddcac I've merged these `false` and `null` states into just `null`. The logic was going ahead and preloading an LCP element from a smaller viewport group when the next-largest viewport group was missing data. This is now undone so that it will only preload images which are _actually_ detected as being LCP elements.

Work in this PR leads into #1235 which will address #1088.

PR #1175, opened 2024-04-23
Avoid passing incomplete data to perflab_render_plugin_card() and show error when plugin directory API query fails

westonruter:
## Summary

When working on Image Prioritizer (#1172), I added the plugin to the `perflab_get_standalone_plugin_data()` but then I started getting PHPUnit test failures. I also got a bunch of warnings on the Performance screen:

> image

The issue here is that Image Prioritizer is not on WordPress.org yet, so the call to `perflab_query_plugin_info()` is returning an empty array. Nevertheless, this logic in `perflab_render_plugin_card()` is not working as expected:

https://github.com/WordPress/performance/blob/cb5a37e8871e06564ac23e793b2121d6181dc0b8/includes/admin/plugins.php#L130-L134

This is because we are merging the response from WordPress.org with our own data from `perflab_get_standalone_plugin_data()` meaning `$plugin_data` will _never_ be empty:

https://github.com/WordPress/performance/blob/cb5a37e8871e06564ac23e793b2121d6181dc0b8/includes/admin/plugins.php#L76-L88

This scenario could happen not only if the plugin is not on WordPress.org but also if the query fails due to WordPress.org being down. So with this PR an error is now shown when an error occurs:

image

This PR also removes the standalone plugin version from being referenced on the Performance screen. It was only referenced in `aria-label` and `data-title` attributes.

Lastly, when a WordPress.org Plugins API query fails during installation, an error message is now displayed rather than just a generic error.

Related: #1170

### DIFF
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -260,14 +260,14 @@ function perflab_install_activate_plugin_callback() {
 		$api = perflab_query_plugin_info( $plugin_slug );
 
 		// Return early if plugin API returns an error.
-		if ( ! $api ) {
+		if ( $api instanceof WP_Error ) {
 			wp_die(
 				wp_kses(
 					sprintf(
 						/* translators: %s: Support forums URL. */
 						__( 'An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration. If you continue to have problems, please try the <a href="%s">support forums</a>.', 'default' ),
 						__( 'https://wordpress.org/support/forums/', 'default' )
-					),
+					) . ' ' . $api->get_error_message(),
 					array( 'a' => array( 'href' => true ) )
 				)
 			);
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -304,7 +304,7 @@ function perflab_install_activate_plugin_callback() {
 
 	$result = activate_plugin( $plugin_basename );
 	if ( is_wp_error( $result ) ) {
-		wp_die( esc_html( $result->get_error_message() ) );
+		wp_die( wp_kses_post( $result->get_error_message() ) );
 	}
 
 	$url = add_query_arg(
--- a/includes/admin/plugins.php
+++ b/includes/admin/plugins.php
@@ -15,7 +15,7 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @since 2.8.0
  *
  * @param string $plugin_slug The string identifier for the plugin in questions slug.
- * @return array|WP_Error Array of plugin data or WP_Error if failed.
+ * @return array{name: string, slug: string, short_description: string, requires: string|false, requires_php: string|false, requires_plugins: string[], download_link: string}|WP_Error Array of plugin data or WP_Error if failed.
  */
 function perflab_query_plugin_info( string $plugin_slug ) {
 	$plugin = get_transient( 'perflab_plugin_info_' . $plugin_slug );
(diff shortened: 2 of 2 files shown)


includes/admin/plugins.php
```diff
@@ -125,22 +151,15 @@ function perflab_render_plugins_ui() {
...
-	}
 
-	// Remove any HTML from the description.
+	$name        = wp_strip_all_tags( $plugin_data['name'] );
 	$description = wp_strip_all_tags( $plugin_data['short_description'] );
-	$title       = $plugin_data['name'];
 
 	/** This filter is documented in wp-admin/includes/class-wp-plugin-install-list-table.php */
 	$description = apply_filters( 'plugin_install_description', $description, $plugin_data );
-	$version     = $plugin_data['version'];
```
westonruter, line 142: I didn't see a need for the version to be mentioned among the features, so I removed it.

includes/admin/load.php
```diff
@@ -260,14 +260,14 @@ function perflab_install_activate_plugin_callback() {
...
-		if ( ! $api ) {
+		if ( $api instanceof WP_Error ) {
 			wp_die(
 				wp_kses(
 					sprintf(
 						/* translators: %s: Support forums URL. */
 						__( 'An unexpected error occurred. Something may be wrong with WordPress.org or this server&#8217;s configuration. If you continue to have problems, please try the <a href="%s">support forums</a>.', 'default' ),
 						__( 'https://wordpress.org/support/forums/', 'default' )
-					),
+					) . ' ' . $api->get_error_message(),
```
westonruter, line 278: Now the specific error is shown rather than just the generic message.

includes/admin/load.php
```diff
@@ -304,7 +304,7 @@ function perflab_install_activate_plugin_callback() {
 
 	$result = activate_plugin( $plugin_basename );
 	if ( is_wp_error( $result ) ) {
-		wp_die( esc_html( $result->get_error_message() ) );
+		wp_die( wp_kses_post( $result->get_error_message() ) );
```
westonruter, line 313: This fixes HTML mistakenly being escaped:

image

Now:

image

includes/admin/plugins.php
```diff
@@ -74,20 +88,41 @@ function perflab_render_plugins_ui() {
...
+
+		// Skip if the plugin is not on WordPress.org or there was a network error.
+		if ( $api_data instanceof WP_Error ) {
+			wp_admin_notice(
+				esc_html(
+					sprintf(
+						/* translators: %s is the plugin slug */
+						__( 'Failed to query WordPress.org Plugin Directory for plugin "%s".', 'performance-lab' ) . ' %s',
+						$plugin_slug,
+						$api_data->get_error_message()
```
swissspidy:
```suggestion
						/* translators: 1: plugin slug. 2: error message. */
						__( 'Failed to query WordPress.org Plugin Directory for plugin "%1$s". %2$s', 'performance-lab' ),
						$plugin_slug,
						$api_data->get_error_message()
```

westonruter: Actually I did it this way to reuse the core translation string.

westonruter: Oh no, that was somewhere else I guess.

swissspidy (approved):

includes/admin/plugins.php
westonruter: Oh yeah, it should do the same thing as I commented at https://github.com/WordPress/performance/pull/1175/files#r1576999210

westonruter: Meh. That's too generic. Nevermind.

includes/admin/plugins.php
```diff
@@ -24,25 +24,39 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
+
 	$plugin = plugins_api(
 		'plugin_information',
 		array(
 			'slug'   => $plugin_slug,
-			'fields' => array(
-				'short_description' => true,
-				'icons'             => true,
-			),
+			'fields' => array_fill_keys( $fields, true ),
```
mukeshpanchal27, line 42: Nice ❤️

mukeshpanchal27 (approved):

Issue #996, opened 2024-02-16
Fallback script obsolete - remove it


westonruter:
See discussion below where we decided we should drop the fallback script, especially since the plugin now supports the `<picture>` element which provides smoother fallback experience, and because WebP is so widely supported.

<details><summary>Old description for "Script handling in WebP Uploads can be improved"</summary>

Currently an inline script is printed to the page which conditionally loads a `fallback.js` script:

https://github.com/WordPress/performance/blob/11be3253d7c40baa2a08b32eb93520b93c22665f/modules/images/webp-uploads/hooks.php#L699-L705

Note how it adds a `data-rest-api` data attribute here. This is then read in `fallback.js` via:

https://github.com/WordPress/performance/blob/11be3253d7c40baa2a08b32eb93520b93c22665f/modules/images/webp-uploads/fallback.js#L91-L93

This value is then used to construct the `script` URL to load:

https://github.com/WordPress/performance/blob/11be3253d7c40baa2a08b32eb93520b93c22665f/modules/images/webp-uploads/fallback.js#L129-L137

A couple things here:

1. Wouldn't it be better to rely on `fetch()` than to use JSONP?
2. Instead of exporting the URL in an HTML attribute, why not expose it via a global variable in the same inline script that is being printed to begin with?
3. The query args for `restPath` would be better constructed with `URLSearchParams`, or if IE11 compat is really desired, fall back to using `document.createElement('a').href`.

Also, in that inline script it would be slightly safer if this line used `wp_json_encode()`:

https://github.com/WordPress/performance/blob/11be3253d7c40baa2a08b32eb93520b93c22665f/modules/images/webp-uploads/hooks.php#L683

Like:

```js
s.src = <?php echo wp_json_encode( esc_url_raw( plugins_url( '/fallback.js', __FILE__ ) ) ); ?>; 
```

Lastly, instead of relying on data attributes on the script tag, the necessary data may make more sense to export into JS via `wp_add_inline_script_tag()` or else a JSON `script`. The reason is that the `script` tag for JS may end up getting concatenated with other scripts, resulting in those attributes getting lost.

</details>

adamsilverstein:
I wonder if we should consider adding `picture` element support to replace this approach? Support is [solid](https://caniuse.com/picture) and using `picture` would provide a superior user experience across the board. Since we have the `sources` data, outputting a picture element should be straightforward. See https://github.com/WordPress/performance/issues/21 for previous discussion.

westonruter:
[...]

Issue #469, opened 2022-08-09
Mapping output from WebP to PNG loads replacement code does not work


erikyo:
## Bug Description

I know the main intent was to provide a replacement jpeg->webp but for an ecommerce I had to change the normal operation of the plugin (otherwise it would have been of little use since the images were almost all png with transparency)

## Steps to reproduce

1. Add the filter 
```php
add_filter('webp_uploads_upload_image_mime_transforms', function() {
    return array(
        'image/png' => array( 'image/png', 'image/webp' ),
        'image/webp' => array( 'image/webp', 'image/png' ),
    );
});
```
2. Create a page and upload a bunch of png images
3. When you view that page and the fallback script is activated it tries to replace images [jpg images](https://github.com/WordPress/performance/blob/trunk/modules/images/webp-uploads/load.php#L529) only

## Additional Context

- was previously discussed here https://github.com/WordPress/performance/pull/360#issuecomment-1155988625

mxbclang:
Tagging @adamsilverstein based on our discussion of this in today's chat.

adamsilverstein:
It looks like our JavaScript has [hard coded](https://github.com/WordPress/performance/blob/308f61a3af95dc83e5a8d34b8e7c58d5d3354160/plugins/webp-uploads/fallback.js#L32) mime types it looks for; this is something we will need to address when adding AVIF support. Leveraging `picture` element support would likely eliminate the need for this script entirely.

adamsilverstein:
the new version of the fallback script in https://github.com/WordPress/performance/pull/1176 addresses AVIF and WebP.

@erikyo We are now working on [adding picture element](https://github.com/WordPress/performance/pull/73) support - this should work better than the fallback script in your use case, could you give that a test to see if it resolves this issue?

erikyo:
yes i can confirm that now I get it working with this 👇filter

```php
add_filter('webp_uploads_upload_image_mime_transforms', function() {
	return array(
		'image/png' => array( 'image/webp' )
	);
});
```

thanks!

Issue #371, opened 2022-06-13
Convert PNG to AVIF/WebP on upload


mxbclang:
As noted in [this support topic](https://wordpress.org/support/topic/need-support-to-convert-png-images-to-webp/), add the ability to automatically convert PNG images to WebP.

erikyo:
I think it's definitely a good idea 💯
if jpgs are not always the ideal source for webp encoding, because the result vary depending on the source quality, pngs would give amazing results -- especially considering that a lossy webp weighs up to 20 times less than a png (png -> webp lossy). This would be a nice performance boost! 
Another advantage is that in this case there we have generation loss so, at the same SSIM, the images will weight about the 30% compared to a standard jpg.

there would be two options: 
- convert the png to lossy
- convert the png to lossless

My vote goes for **png->lossy** because the images of websites aren't intended to be 1:1 with the original so we can take the advantage of the compression techniques like chroma subsampling that decimates the image filesize. 

Another suggestion is to use the png as the source image and not to create the subsizes in png. 
I currently use exactly this way: I upload png images that I keep it only as source in order to create the copy and subsizes in webp. In this way I get much higher quality than a jpg, it takes up less space, and I only have one extra image (the source png).

EDIT: the bug of the webp missing transparency is related to versions before [gd 2.1.1](https://github.com/libgd/libgd/releases) (6 years ago, php 5.6)... I forgot to mention about, but i guess actually isn't a real issue since it's [deprecated](https://libgd.github.io/pages/docs.html). here more [info](https://github.com/rosell-dk/webp-convert/issues/238#issuecomment-545928597)

maisondasilva:
@felixarntz Any news about this, I loved the plugin, but most of my images are .png and that would be very good, I did an online conversion test and saw that the images would be half the size!

Thanks

nikolasholm:
Any update on this yet?

ficod:
Any plans on this feature? First post in this thread is dated Jun 13, 2022... 
Is there any progress made or it's just "planned for the future"?

predaytor:
@ficod, yeah, I decided to convert them manually with Imagick:
```php
<?php

namespace App\Media;

/**
 * Set image quality based on mime type.
 *
 * @param int    $quality    The image quality level.
 * @param string $mime_type  The mime type of the image.
 *
 * @return int The modified image quality level.
 */
function set_image_quality($quality, $mime_type)
{
    // Check if the mime type is 'image/webp'
[...]

PR #1176, opened 2024-04-23
Modern Image Formats: add AVIF output support

### LINKED ISSUE #1151: Add AVIF support to the Modern Image Formats plugin
## Feature Description

Now that WordPress 6.5+ supports the AVIF format, we should add support in the **Modern Image Formats** plugin (previously "WebP uploads").

## Details

* This feature should allow the user to choose between outputting WebP or AVIF images for uploaded images. We could add a dropdown to select the output format in the "uploading files" section of Settings->Media eg  "Choose the image output format". 
* The AVIF option should only be available if the server actually supports AVIF. otherwise the UI should remain unchanged
* When the user selects the AVIF format, all images are output in AVIF where they are currently output in WebP (nothing else changes)
* The current checkbox for "Generate JPEG files in addition to WebP" needs to be renamed. We can remove the WebP mention, eg. "In addition, generate JPEG images" or "Generate JPEG fallback images"

## Questions
* Can we make AVIF the default for new installs?

### Related: 
https://github.com/WordPress/performance/issues/11
https://github.com/WordPress/performance/issues/1133

adamsilverstein:
## Summary

Add AVIF support for Modern Image Formats plugin.

Fixes https://github.com/WordPress/performance/issues/1151

## Relevant technical choices

* When the server supports AVIF, add a dropdown under settings->media where users can choose between AVIF and WebP as the output format
* Slightly rename the field that enables generating JPEG images alongside the modern output format (so it isn't just about WebP plug JPEG).
  * note: I left the option name unchanged for backwards compatibility. 
* I removed the CSS class that were  hiding part of the settings UI, that felt hacky and made adding the new field more challenging.

### Media settings

The new settings fields on media, the new dropdown is only shown if AVIF is supported on the server:

image

note: the default selection is AVIF when available.

### Uploads
* Note: activate twenty-twentyone theme

#### Output set to AVIF 
image

#### Output set to WebP
image

#### AVIF
image

#### WebP
image

Note: there appears to be a bug when the additional JPEG option is selected. I will double check, I believe this problem is already present in trunk. Looking carefully at the final two images where this setting is enabled, the AVIF/WebP images are only generated in the default (small/medium/medium/large/large and -scaled) variations it looks like custom sizes and maybe the high dpi images are missing in the modern format.  I can create a separate issue to investigate this further.


plugins/webp-uploads/helper.php
```diff
@@ -16,20 +16,24 @@
...
  * @return array<string, array<string>> An array of valid mime types, where the key is the mime type and the value is the extension type.
  */
 function webp_uploads_get_upload_image_mime_transforms() {
+
+	// Check the selected output format.
+	$output_format = get_option( 'perflab_generate_avif_and_webp', wp_image_editor_supports( array( 'mime_type' => 'image/avif' ) ) ?  'avif' : 'webp' );
+
 	$default_transforms = array(
-		'image/jpeg' => array( 'image/webp' ),
-		'image/webp' => array( 'image/webp' ),
```
adamsilverstein, line 25: self format is the default so i removed this line

plugins/webp-uploads/settings.php
```diff
@@ -48,45 +77,41 @@ function webp_uploads_add_media_settings_field() {
...
+
+	$selected = get_option( 'perflab_generate_avif_and_webp', 'avif' );
 	?>
-		<label for="perflab_generate_webp_and_jpeg">
-			<input name="perflab_generate_webp_and_jpeg" type="checkbox" id="perflab_generate_webp_and_jpeg" aria-describedby="perflab_generate_webp_and_jpeg_description" value="1"<?php checked( '1', get_option( 'perflab_generate_webp_and_jpeg' ) ); ?> />
-			<?php esc_html_e( 'Generate JPEG files in addition to WebP', 'webp-uploads' ); ?>
-		</label>
-		<p class="description" id="perflab_generate_webp_and_jpeg_description"><?php esc_html_e( 'Enabling JPEG in addition to WebP can improve compatibility, but will effectively double the filesystem storage use of your images.', 'webp-uploads' ); ?></p>
+			<label for="perflab_generate_avif_and_webp">
+				<select name="perflab_generate_avif_and_webp" id="perflab_generate_avif_and_webp" aria-describedby="perflab_generate_avif_and_webp_description">
```
adamsilverstein: my assumption is that it only ever makes sense to output to a single format (plus optionally the JPEG fallback) - hence using a dropdown for format selection here

plugins/webp-uploads/helper.php
```diff
@@ -16,20 +16,24 @@
...
  *
  * @since 1.0.0
+ * @since n.e.x.t Added support for AVIF.
  *
  * @return array<string, array<string>> An array of valid mime types, where the key is the mime type and the value is the extension type.
  */
 function webp_uploads_get_upload_image_mime_transforms() {
+
+	// Check the selected output format.
+	$output_format = get_option( 'perflab_generate_avif_and_webp', wp_image_editor_supports( array( 'mime_type' => 'image/avif' ) ) ?  'avif' : 'webp' );
```
westonruter: Micro-nit:

```suggestion
	$output_format = get_option( 'perflab_generate_avif_and_webp', wp_image_editor_supports( array( 'mime_type' => 'image/avif' ) ) ? 'avif' : 'webp' );
```

plugins/webp-uploads/helper.php
```diff
@@ -16,20 +16,24 @@
...
+		'image/jpeg' => array( 'image/' . $output_format ),
 	);
 
-	// Check setting for whether to generate both JPEG and WebP.
+	// Check setting for whether to generate both JPEG and the modern output format.
 	if ( true === (bool) get_option( 'perflab_generate_webp_and_jpeg' ) ) {
 		$default_transforms = array(
-			'image/jpeg' => array( 'image/jpeg', 'image/webp' ),
-			'image/webp' => array( 'image/webp', 'image/jpeg' ),
+			'image/jpeg' => array( 'image/jpeg', 'image/' . $output_format ),
```
westonruter: For PHPCS:

```suggestion
			'image/jpeg'              => array( 'image/jpeg', 'image/' . $output_format ),
```

plugins/webp-uploads/settings.php
```diff
@@ -12,11 +12,26 @@
...
+ * @since n.e.x.t The `perflab_generate_avif_and_webp` was added to enable selecting between AVIF and WebP output.
  */
 function webp_uploads_register_media_settings_field() {
+	if ( wp_image_editor_supports( array( 'mime_type' => 'image/avif' ) ) ) {
+		register_setting(
+			'media',
+			'perflab_generate_avif_and_webp',
+			array(
+				'type'         => 'string',
+				'default'      => 'avif',
```
westonruter: Should this not default to `webp`?

plugins/webp-uploads/settings.php
```diff
@@ -48,45 +77,41 @@ function webp_uploads_add_media_settings_field() {
...
-function webp_uploads_generate_webp_jpeg_setting_callback() {
-	if ( ! is_multisite() ) {
-		?>
-			</td>
-			<td class="td-full">
-		<?php
-	}
+function webp_uploads_generate_avif_webp_setting_callback() {
+
+	$selected = get_option( 'perflab_generate_avif_and_webp', 'avif' );
```
westonruter: AFAIK no default value is needed here since the default is defined where the setting is registered.

```suggestion
	$selected = get_option( 'perflab_generate_avif_and_webp' );
```

westonruter (changes requested):

plugins/webp-uploads/settings.php
adamsilverstein: Hmm, AVIF is the better/newer format, why not default to it when available?

adamsilverstein:
It would be good to add some tests here. I was expecting existing tests to fail, since I changed the default output to AVIF - maybe the test runners don't have AVIF support.

plugins/webp-uploads/settings.php
westonruter: I guess because WebP is more widely-supported? And it is the format that this plugin was focused on previously. So if users update this plugin, then they will all of a sudden start getting AVIF images generated instead of WebP ones, which may not be what they expect.

adamsilverstein: > I guess because WebP is more widely-supported? 

It only defaults to AVIF if you have it available, otherwise the dropdown is not shown and webp  is used as the default output type.

> So if users update this plugin, then they will all of a sudden start getting AVIF images generated instead of WebP ones, which may not be what they expect

Ah, good point, how about I change it so that only new installs get AVIF by default (if available)?

westonruter: In regards to support, I mean not server support but rather client support (and support in popular web apps).

adamsilverstein: Browser support for AVIF is actually pretty solid: https://caniuse.com/avif. I don't think we should be too concerned with the "support in popular web apps" issue for this plugin.

plugins/webp-uploads/settings.php
```diff
@@ -12,11 +12,26 @@
 }
 
 /**
- * Registers setting for generating both JPEG and WebP versions for image uploads.
+ * Registers setting for generating JPEG in addition to the selected modern format for image uploads.
  *
  * @since 1.0.0
+ * @since n.e.x.t The setting was made more general to coveroutputting JPEG as a secondary type. The "webp" option naming
+ *        was left unchanged for backward compatibility.
```
westonruter: Should this indent to match "The" on the line above?

plugins/webp-uploads/helper.php
```diff
@@ -16,20 +16,24 @@
...
  *
  * @since 1.0.0
+ * @since n.e.x.t Added support for AVIF.
  *
  * @return array<string, array<string>> An array of valid mime types, where the key is the mime type and the value is the extension type.
  */
 function webp_uploads_get_upload_image_mime_transforms() {
+
+	// Check the selected output format.
+	$output_format = get_option( 'perflab_generate_avif_and_webp', wp_image_editor_supports( array( 'mime_type' => 'image/avif' ) ) ? 'avif' : 'webp' );
```
westonruter: Since the `wp_image_editor_supports( array( 'mime_type' => 'image/avif' )` code is appearing in multiple places in this PR, should there be a helper function like `webp_uploads_supports_avif()`?

plugins/webp-uploads/settings.php
```diff
@@ -12,11 +12,26 @@
...
+ * @since n.e.x.t The `perflab_generate_avif_and_webp` was added to enable selecting between AVIF and WebP output.
  */
 function webp_uploads_register_media_settings_field() {
+	if ( wp_image_editor_supports( array( 'mime_type' => 'image/avif' ) ) ) {
+		register_setting(
+			'media',
+			'perflab_generate_avif_and_webp',
```
westonruter: Would this setting name more sense as `perflab_modern_image_format`? Then it could go beyond AVIF in the future?

plugins/webp-uploads/helper.php
westonruter: Also, the `perflab_generate_avif_and_webp` option here only exists if AVIF support is present. Otherwise, the option will never exist which will result in an additional failing DB query for each request unless a persistent object cache is present, right? Should the same condition then be used? Like:

```suggestion
	$output_format = wp_image_editor_supports( array( 'mime_type' => 'image/avif' ) ) ? get_option( 'perflab_generate_avif_and_webp' ) : 'webp';
```

plugins/webp-uploads/settings.php
```diff
@@ -35,10 +50,24 @@ function webp_uploads_register_media_settings_field() {
  * @since 1.0.0
  */
 function webp_uploads_add_media_settings_field() {
+
+	// If AVIF is supported, add a dropdown to select the output format between AVIF and WebP.
+	if ( wp_image_editor_supports( array( 'mime_type' => 'image/avif' ) ) ) {
```
westonruter: Should some field still be printed even when AVIF is not supported by the server to inform the user that this is why AVIF isn't available? I guess there is a Site Health test for that, but only if Performance Lab is installed, right?

plugins/webp-uploads/settings.php
adamsilverstein: right, fixed.

plugins/webp-uploads/settings.php
adamsilverstein: > Should some field still be printed even when AVIF is not supported by the server

I guess we could show a dropdown with WebP and AVIF, but AVIF is disabled, wdyt? (Thats actually what I do in my https://github.com/adamsilverstein/modern-images-wp plugin)

> I guess there is a Site Health test for that, but only if Performance Lab is installed, right?

Right, I see that landed in https://github.com/WordPress/performance/pull/1177 🎉

plugins/webp-uploads/settings.php
adamsilverstein: Good point, I already dislike having "webp" everywhere so no reason to repeat that. Will update.

adamsilverstein: updated

adamsilverstein:
I'm going to work on expanding the tests to cover both WebP and AVIF output.

plugins/webp-uploads/settings.php
westonruter: Having AVIF alone disabled may not be enough. It might need a separate inline WP notice, since when there is just one disabled option and one enabled/selected option, it's not entirely clear that the disabled option is disabled from the gray text alone:

[Screen recording 2024-05-01 14.31.09.webm](https://github.com/WordPress/performance/assets/134745/dc5b5c4b-a351-46e7-bbdd-4347534d41b6)

adamsilverstein: That makes sense. Maybe we can do both the disabled option and the inline notice.

adamsilverstein:
Still working on tests for AVIF here, I also started working on making the fallback script more flexible so it will work for either AVIF or WebP.

Related: https://github.com/WordPress/performance/issues/469 & https://github.com/WordPress/performance/issues/996

adamsilverstein:
Tests were failing with PHP 8.0 because some environments report AVIF support when in fact they do not support AVIF, see https://github.com/php/php-src/issues/12197. To get tests passing locally, I made the check for AVIF support more robust.

plugins/webp-uploads/helper.php
westonruter: You removed `'image/webp' => array( 'image/webp' ),`? I see you removed a non-self format `'image/jpeg' => array( 'image/webp' ),`

plugins/webp-uploads/helper.php
```diff
@@ -340,3 +346,28 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
 	}
 	return false;
 }
+
+/**
+ * Helper function that checks if a mime type is supported by the server.
```
```suggestion
 * Checks if a mime type is supported by the server.
```

plugins/webp-uploads/helper.php
```diff
@@ -340,3 +346,28 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
...
+ *
+ * Includes special handling for false positives on AVIF support.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $mime_type The mime type to check.
+ *
+ * @return bool Whether the server supports a given mime type.
+ */
+function webp_uploads_mime_type_supported( $mime_type ) {
```
```suggestion
function webp_uploads_mime_type_supported( string $mime_type ): bool {
```

plugins/webp-uploads/helper.php
```diff
@@ -16,20 +16,26 @@
  * for example an image/jpeg can be converted into an image/webp.
  *
  * @since 1.0.0
+ * @since n.e.x.t Added support for AVIF.
  *
  * @return array<string, array<string>> An array of valid mime types, where the key is the mime type and the value is the extension type.
  */
 function webp_uploads_get_upload_image_mime_transforms() {
```
westonruter: Might as well since these lines are being touched:

```suggestion
function webp_uploads_get_upload_image_mime_transforms(): array {
```

plugins/webp-uploads/hooks.php
```diff
@@ -683,16 +695,16 @@ function webp_uploads_wepb_fallback() {
 		s.src = '<?php echo esc_url_raw( plugins_url( '/fallback.js', __FILE__ ) ); ?>';
```
westonruter:
```suggestion
		s.src = <?php echo wp_json_encode( esc_url_raw( plugins_url( '/fallback.js', __FILE__ ) ) ); ?>;
```

plugins/webp-uploads/hooks.php
```diff
@@ -683,16 +695,16 @@ function webp_uploads_wepb_fallback() {
...
 			i.onload = undefined;
-			i.src = p + 'h4AAABXRUJQVlA4TBEAAAAvAQAAAAfQ//73v/+BiOh/AAA=';
+			i.src = 'data:image/webp;base64,UklGRh4AAABXRUJQVlA4TBEAAAAvAQAAAAfQ//73v/+BiOh/AAA=';
 		};
 
 		i.onerror = function() {
 			d.body.appendChild( s );
 		};
-	} )( document, 'img', 'script', 'data:image/webp;base64,UklGR' );
+	} )( document, 'img', 'script', '<?php echo esc_html( $detection_string ); ?>' );
```
westonruter:
```suggestion
	} )( document, 'img', 'script', <?php echo wp_json_encode( $detection_string ) ); ?> );
```

plugins/webp-uploads/hooks.php
```diff
@@ -666,12 +666,24 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
  */
 function webp_uploads_wepb_fallback() {
 	// Get mime type transforms for the site.
-	$transforms = webp_uploads_get_upload_image_mime_transforms();
+	$transforms   = webp_uploads_get_upload_image_mime_transforms();
+	$image_format = get_option( 'perflab_modern_image_format' );
```
westonruter: This `perflab_modern_image_format` setting isn't always registered, right? So the `$image_format` could be `false` here. Does that matter?

plugins/webp-uploads/helper.php
```diff
@@ -340,3 +346,28 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
...
+ *
+ * @return bool Whether the server supports a given mime type.
+ */
+function webp_uploads_mime_type_supported( $mime_type ) {
+	if ( ! wp_image_editor_supports( array( 'mime_type' => $mime_type ) ) ) {
+		return false;
+	}
+
+	// In certain server environments GD can report a false positive for AVIF support, see https://github.com/php/php-src/issues/12197.
+	// To ensure the server actually supports AVIF, test for the function used by core..
```
```suggestion
	// To ensure the server actually supports AVIF, test for the function used by core.
```

plugins/webp-uploads/hooks.php
```diff
@@ -683,16 +695,16 @@ function webp_uploads_wepb_fallback() {
...
+			i.src = 'data:image/webp;base64,UklGRh4AAABXRUJQVlA4TBEAAAAvAQAAAAfQ//73v/+BiOh/AAA=';
 		};
 
 		i.onerror = function() {
 			d.body.appendChild( s );
 		};
-	} )( document, 'img', 'script', 'data:image/webp;base64,UklGR' );
+	} )( document, 'img', 'script', '<?php echo esc_html( $detection_string ); ?>' );
 	<?php
 	$javascript = ob_get_clean();
```
westonruter: Here's an alternative which avoids the use of `ob_start()` and ensures that the JS gets IDE syntax checking without needing to mix JS and PHP code: 

```php
	$js_function = <<<JS
		function webpUploadsDetectFallback( d, i, s, p, fallbackSrc ) {
			s = d.createElement( s );
			s.src = fallbackSrc;

			i = d.createElement( i );
			i.src = p;
			i.onload = function() {
				i.onload = undefined;
				i.src = 'data:image/webp;base64,UklGRh4AAABXRUJQVlA4TBEAAAAvAQAAAAfQ//73v/+BiOh/AAA=';
			};

			i.onerror = function() {
				d.body.appendChild( s );
			};
		}
JS;

	wp_print_inline_script_tag(
		sprintf(
			'( %s )( "document", "img", "script", %s, %s )',
			preg_replace( '/\s+/', '', $js_function ),
			wp_json_encode( $detection_string ),
			wp_json_encode( plugins_url( '/fallback.js', __FILE__ ) )
		),
		array(
			'id'            => 'webpUploadsFallbackWebpImages',
			'data-rest-api' => esc_url_raw( trailingslashit( get_rest_url() ) ),
		)
	);
```

In PhpStorm:

Before | After
--|--
image | image

plugins/webp-uploads/helper.php
adamsilverstein: well, I changed it to `'image/jpeg' => array( 'image/' . $output_format )` which is now either webp or avif. 

Also, I have since restored these lines because having the self mapping actually changed the behavior - the plugin doesn't create the root source mapping if only one transform is included.

plugins/webp-uploads/hooks.php
adamsilverstein: sure, i can change this...

plugins/webp-uploads/hooks.php
adamsilverstein: Good point, addressed in 8899b66bdbd98ae4480f87bf0d7fb698f720f9a3

adamsilverstein:
Notice warning users their server lacks AVIF support. Language borrowed from Health Check for AVIF
image


AVIF disabled in the dropdown due to lack of server support.
image

plugins/webp-uploads/settings.php
adamsilverstein: resolving for now

plugins/webp-uploads/settings.php
adamsilverstein: I added both an inline notice and disabled "AVIF" in the dropdown in [`88dedcb` (#1176)](https://github.com/WordPress/performance/pull/1176/commits/88dedcb2b40ef9d7a128fa5f7abf5ba9a16f79fe) and [`3896293` (#1176)](https://github.com/WordPress/performance/pull/1176/commits/3896293ca512c7680939555801b9ed8b95f58033)

plugins/webp-uploads/hooks.php
adamsilverstein: Updated this as suggested

adamsilverstein:
While testing the fallback script locally I noticed that if I check the box to output JPEG as well as WebP or AVIF, the output HTML lacks the `srcset` I expect to see. Unchecking the box, the srcset returns. 

Looking at the generated meta data, it seems fine so I am going to dig further. I don't **think** this is the intended behavior. Maybe @mukeshpanchal27 knows, did this previously work with the full `srcset` here?

adamsilverstein:
> While testing the fallback script locally I noticed that if I check the box to output JPEG as well as WebP or AVIF, the output HTML lacks the `srcset` I expect to see. Unchecking the box, the srcset returns.

I [tested](https://convincing-porcupine-aa1903.instawp.xyz/test-with-jpeg/) this with the current shipping plugin and observed the same issue, so this is unrelated to the changes in this PR. I searched and didn't find an existing issue for this, so I will go ahead and open one.

adamsilverstein:
> > While testing the fallback script locally I noticed that if I check the box to output JPEG as well as WebP or AVIF, the output HTML lacks the `srcset` I expect to see. Unchecking the box, the srcset returns.
> 
> I [tested](https://convincing-porcupine-aa1903.instawp.xyz/test-with-jpeg/) this with the current shipping plugin and observed the same issue, so this is unrelated to the changes in this PR. I searched and didn't find an existing issue for this, so I will go ahead and open one.

I created this issue to investigate separately: https://github.com/WordPress/performance/issues/1207

plugins/webp-uploads/hooks.php
```diff
@@ -681,38 +681,54 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
-	<?php
-	$javascript = ob_get_clean();
+	}
+JS;
 
 	wp_print_inline_script_tag(
-		preg_replace( '/\s+/', '', $javascript ),
+		sprintf(
+			'( %s )( "document", "img", "script", %s, %s )',
+			preg_replace( '/\s+/', '', $js_function ),
```
westonruter: I'm surprised the JS isn't erroring with removal of all whitespace. I think this would be safer. although it doesn't cause an error currently:

```suggestion
			preg_replace( '/\s+/', ' ', $js_function ),
```

plugins/webp-uploads/hooks.php
```diff
@@ -681,38 +681,54 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+			wp_json_encode( $detection_string ),
+			wp_json_encode( plugins_url( '/fallback.js', __FILE__ ) )
+		),
 		array(
 			'id'            => 'webpUploadsFallbackWebpImages',
```
westonruter: Can we make this a module? Maybe that would be bad since it would wait for DOMContentLoaded:

```suggestion
			'id'            => 'webpUploadsFallbackWebpImages',
			'type'          => 'module',
```

tests/plugins/webp-uploads/data/class-wp-image-doesnt-support-webp.php
```diff
@@ -12,18 +12,21 @@
...
 	 *
 	 * @param string $mime_type The mime type to check.
 	 * @return bool
 	 */
 	public static function supports_mime_type( $mime_type ) {
-		return 'image/webp' !== $mime_type;
+		return (
+			'image/webp' !== $mime_type &&
+			'image/avif' !== $mime_type
+			);;
```
westonruter:
```suggestion
		);
```

plugins/webp-uploads/helper.php
```diff
@@ -348,3 +354,33 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
...
+
+/**
+ * Checks if a mime type is supported by the server.
+ *
+ * Includes special handling for false positives on AVIF support.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $mime_type The mime type to check.
+ *
```
```suggestion
```

plugins/webp-uploads/helper.php
```diff
@@ -348,3 +354,33 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
...
+	}
+
+	// In certain server environments Image editors can report a false positive for AVIF support.
+	if ( 'image/avif' === $mime_type ) {
+		$editor = _wp_image_editor_choose( 'image/avif' );
+		if ( 'WP_Image_Editor_GD' === $editor ) {
+			return function_exists( 'imageavif' );
+		}
+		if ( 'WP_Image_Editor_Imagick' === $editor ) {
+			return ! empty( Imagick::queryformats('AVIF') );
```
westonruter: Would this be better?

```suggestion
			return (bool) Imagick::queryformats('AVIF');
```

plugins/webp-uploads/hooks.php
```diff
@@ -681,38 +681,54 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
 	// Get mime type transforms for the site.
-	$transforms = webp_uploads_get_upload_image_mime_transforms();
-
-	// We need to add fallback only if jpeg alternatives for the webp images are enabled for the server.
-	$preserve_jpegs_for_jpeg_transforms = isset( $transforms['image/jpeg'] ) && in_array( 'image/jpeg', $transforms['image/jpeg'], true ) && in_array( 'image/webp', $transforms['image/jpeg'], true );
-	$preserve_jpegs_for_webp_transforms = isset( $transforms['image/webp'] ) && in_array( 'image/jpeg', $transforms['image/webp'], true ) && in_array( 'image/webp', $transforms['image/webp'], true );
-	if ( ! $preserve_jpegs_for_jpeg_transforms && ! $preserve_jpegs_for_webp_transforms ) {
+	$transforms   = webp_uploads_get_upload_image_mime_transforms();
+	$image_format = get_option( 'perflab_modern_image_format' );
+	$image_format = empty( $image_format ) ? 'webp' : $image_format;
```
westonruter: This would seem a bit safer since someone could populate the option with anything:

```suggestion
	$image_format = in_array( $image_format, array( 'webp', 'avif' ), true ) ? $image_format : 'webp';
```

plugins/webp-uploads/settings.php
```diff
@@ -12,11 +12,25 @@
...
+ * @since n.e.x.t The setting was made more general to cover outputting JPEG as a secondary type. The "webp" option naming
+ *                was left unchanged for backward compatibility.
+ * @since n.e.x.t The `perflab_modern_image_format` was added to enable selecting an output format.
+ *                Currently includes AVIF and WebP.
  */
 function webp_uploads_register_media_settings_field() {
+	register_setting(
+		'media',
+		'perflab_modern_image_format',
+		array(
```
westonruter, line 27: What about the `sanitize_callback`?

```suggestion
		array(
			'sanitize_callback' => static function ( $image_format ): string { return in_array( $image_format, array( 'webp', 'avif' ), true ) ? $image_format : 'webp'; },
```

plugins/webp-uploads/hooks.php
adamsilverstein: why, isn't whitespace ignored in JavaScript?

plugins/webp-uploads/hooks.php
adamsilverstein: would this be supported in IE? I can test with IE and AVIF using browserstack to see if that would work.

This approach is only about supporting older browsers and is already a crutch and non-performance so I'd rather make it as compatibility-helping as possible.

plugins/webp-uploads/helper.php
adamsilverstein: I'm sure it must be because you are suggesting it :) Curious why though?

Do you think its clearer/easier to read? The function literally returns an empty array when AVIF is not available, so an empty check seems very appropriate and I tend to prefer empty anyway because it works well across types.

plugins/webp-uploads/settings.php
adamsilverstein: oh, nice!

plugins/webp-uploads/hooks.php
westonruter: I was thinking of JS like `var foo = bar`, if the whitespace is removed then this becomes `varfoo=bar` which would be an error. This isn't an issue with the JS as written right now though.

plugins/webp-uploads/hooks.php
westonruter: OK, yeah, IE doesn't support modules.

plugins/webp-uploads/helper.php
westonruter: I've just been reminded over and over that using `empty()` can be bad:

* https://localheinz.com/articles/2023/05/10/avoiding-empty-in-php/

westonruter: If we know that `Imagick::queryformats('AVIF')` returns an array, and an empty array casted to `bool` is `false`, then this will work as well. Alternatively, this would explicitly be what is being checked:

```suggestion
			return count( Imagick::queryformats('AVIF') ) !== 0;
```

adamsilverstein: > I've just been reminded over and over that using empty() can be bad:

Ok, somehow I got the opposite advice! Thanks for sharing the links, that is helpful to get the perspective. Sure, I'll go ahead and change it.

adamsilverstein: used count approach in [`be67798` (#1176)](https://github.com/WordPress/performance/pull/1176/commits/be67798d2fa3404f66fc3596949112356b68abbc)

plugins/webp-uploads/hooks.php
adamsilverstein: oh good point, thanks for clarifying. accepting the suggestion!

plugins/webp-uploads/hooks.php
```diff
@@ -681,38 +681,54 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
-	ob_start();
+	$detection_string = '';
+	// The fallback script can only handle a single image format at a time.
+	if ( 'webp' === $image_format ) {
+		$detection_string = 'data:image/webp;base64,UklGRh4AAABXRUJQVlA4TBEAAAAvAQAAAAfQ//73v/+BiOh/AAA=';
+	} elseif ( 'avif' === $image_format ) {
+		$detection_string = 'data:image/avif;base64,AAAAIGZ0eXBhdmlmAAAAAGF2aWZtaWYxbWlhZk1BMUIAAADybWV0YQAAAAAAAAAoaGRscgAAAAAAAAAAcGljdAAAAAAAAAAAAAAAAGxpYmF2aWYAAAAADnBpdG0AAAAAAAEAAAAeaWxvYwAAAABEAAABAAEAAAABAAABGgAAAB0AAAAoaWluZgAAAAAAAQAAABppbmZlAgAAAAABAABhdjAxQ29sb3IAAAAAamlwcnAAAABLaXBjbwAAABRpc3BlAAAAAAAAAAIAAAACAAAAEHBpeGkAAAAAAwgICAAAAAxhdjFDgQ0MAAAAABNjb2xybmNseAACAAIAAYAAAAAXaXBtYQAAAAAAAAABAAEEAQKDBAAAACVtZGF0EgAKCBgANogQEAwgMg8f8D///8WfhwB8+ErK42A=';
+	}
+
+	if ( empty( $detection_string ) ) {
```
westonruter:
```suggestion
	if ( ! $detection_string ) {
```

plugins/webp-uploads/hooks.php
```diff
@@ -681,38 +681,54 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+	// The fallback script can only handle a single image format at a time.
+	if ( 'webp' === $image_format ) {
+		$detection_string = 'data:image/webp;base64,UklGRh4AAABXRUJQVlA4TBEAAAAvAQAAAAfQ//73v/+BiOh/AAA=';
+	} elseif ( 'avif' === $image_format ) {
+		$detection_string = 'data:image/avif;base64,AAAAIGZ0eXBhdmlmAAAAAGF2aWZtaWYxbWlhZk1BMUIAAADybWV0YQAAAAAAAAAoaGRscgAAAAAAAAAAcGljdAAAAAAAAAAAAAAAAGxpYmF2aWYAAAAADnBpdG0AAAAAAAEAAAAeaWxvYwAAAABEAAABAAEAAAABAAABGgAAAB0AAAAoaWluZgAAAAAAAQAAABppbmZlAgAAAAABAABhdjAxQ29sb3IAAAAAamlwcnAAAABLaXBjbwAAABRpc3BlAAAAAAAAAAIAAAACAAAAEHBpeGkAAAAAAwgICAAAAAxhdjFDgQ0MAAAAABNjb2xybmNseAACAAIAAYAAAAAXaXBtYQAAAAAAAAABAAEEAQKDBAAAACVtZGF0EgAKCBgANogQEAwgMg8f8D///8WfhwB8+ErK42A=';
+	}
+
+	if ( ! $detection_string ) {
+		return;
+	}
```
westonruter: Oh, hah. Now that the `$image_format` is validated, it can only ever be `webp` or `avif`, so PHPStan is correctly detecting that it will never be falsy here.

```suggestion
```

plugins/webp-uploads/hooks.php
```diff
@@ -681,38 +681,54 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+
+	// We need to add fallback only if jpeg alternatives for the image_format images are enabled for the server.
+	$preserve_jpegs_for_jpeg_transforms       = isset( $transforms['image/jpeg'] ) && in_array( 'image/jpeg', $transforms['image/jpeg'], true ) && in_array( 'image/' . $image_format, $transforms['image/jpeg'], true );
+	$preserve_jpegs_for_image_type_transforms = isset( $transforms[ 'image/' . $image_format ] ) && in_array( 'image/jpeg', $transforms[ 'image/' . $image_format ], true ) && in_array( 'image/' . $image_format, $transforms[ 'image/' . $image_format ], true );
+	if ( ! $preserve_jpegs_for_jpeg_transforms && ! $preserve_jpegs_for_image_type_transforms ) {
 		return;
 	}
 
-	ob_start();
+	$detection_string = '';
```
westonruter, line 724:
```suggestion
```

plugins/webp-uploads/hooks.php
```diff
@@ -681,38 +681,50 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
 	}
 
-	ob_start();
+	// The fallback script can only handle a single image format at a time.
+	if ( 'webp' === $image_format ) {
+		$detection_string = 'data:image/webp;base64,UklGRh4AAABXRUJQVlA4TBEAAAAvAQAAAAfQ//73v/+BiOh/AAA=';
+	} elseif ( 'avif' === $image_format ) {
+		$detection_string = 'data:image/avif;base64,AAAAIGZ0eXBhdmlmAAAAAGF2aWZtaWYxbWlhZk1BMUIAAADybWV0YQAAAAAAAAAoaGRscgAAAAAAAAAAcGljdAAAAAAAAAAAAAAAAGxpYmF2aWYAAAAADnBpdG0AAAAAAAEAAAAeaWxvYwAAAABEAAABAAEAAAABAAABGgAAAB0AAAAoaWluZgAAAAAAAQAAABppbmZlAgAAAAABAABhdjAxQ29sb3IAAAAAamlwcnAAAABLaXBjbwAAABRpc3BlAAAAAAAAAAIAAAACAAAAEHBpeGkAAAAAAwgICAAAAAxhdjFDgQ0MAAAAABNjb2xybmNseAACAAIAAYAAAAAXaXBtYQAAAAAAAAABAAEEAQKDBAAAACVtZGF0EgAKCBgANogQEAwgMg8f8D///8WfhwB8+ErK42A=';
+	}
+
```
westonruter: Last lint fix :)

```suggestion
```

adamsilverstein: on it!

plugins/webp-uploads/helper.php
```diff
@@ -348,3 +354,32 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
...
+ * @return bool Whether the server supports a given mime type.
+ */
+function webp_uploads_mime_type_supported( string $mime_type ): bool {
+	if ( ! wp_image_editor_supports( array( 'mime_type' => $mime_type ) ) ) {
+		return false;
+	}
+
+	// In certain server environments Image editors can report a false positive for AVIF support.
+	if ( 'image/avif' === $mime_type ) {
+		$editor = _wp_image_editor_choose( 'image/avif' );
```
westonruter: Apparently this is supposed to be:

```suggestion
		$editor = _wp_image_editor_choose( array( 'mime_type' => 'image/avif' ) );
```

plugins/webp-uploads/helper.php
```diff
@@ -348,3 +354,32 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
...
+		if ( 'WP_Image_Editor_GD' === $editor ) {
```
westonruter: And it seems `_wp_image_editor_choose` returns an instance of the object, not the class. So should this rather be:

```suggestion
		if ( $editor instanceof WP_Image_Editor_GD ) {
```

Or why `WP_Image_Editor_GD` specifically and not `WP_Image_Editor`?

westonruter: Ah, because of the `function_exists()` check.

plugins/webp-uploads/helper.php
```diff
@@ -348,3 +354,32 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
...
+			return function_exists( 'imageavif' );
+		}
+		if ( 'WP_Image_Editor_Imagick' === $editor ) {
```
```suggestion
		if ( $editor instanceof WP_Image_Editor_Imagick && class_exists( 'Imagick' ) ) {
```

plugins/webp-uploads/helper.php
```diff
@@ -348,3 +354,32 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
...
+			return 0 !== count( Imagick::queryformats( 'AVIF' ) );
```
westonruter: Apparently `capital_F_dangit`? https://www.php.net/manual/en/imagick.queryformats.php

```suggestion
			return 0 !== count( Imagick::queryFormats( 'AVIF' ) );
```

westonruter:
These `assertIsArray` assertions were removed in #1203 (pending merge) since they are now redundant, so they can be removed here too.

plugins/webp-uploads/helper.php
adamsilverstein: note `_wp_image_editor_choose` returns a string

plugins/webp-uploads/helper.php
adamsilverstein: `_wp_image_editor_choose` returns a string, not the editor class

plugins/webp-uploads/helper.php
westonruter: Oops. I must have looked at the wrong spot. In that case, how about:

```suggestion
		if ( is_a( $editor, WP_Image_Editor_GD::class, true ) ) {
```

This will support subclasses of `WP_Image_Editor_GD` being able to be used here.

plugins/webp-uploads/helper.php
westonruter: My bad. In that case, per above:

```suggestion
		if ( is_a( $editor, WP_Image_Editor_Imagick::class, true ) && class_exists( 'Imagick' ) ) {
```

plugins/webp-uploads/helper.php
westonruter: For example, with Dominant Color Images active, this could be `Dominant_Color_Image_Editor_GD` or `Dominant_Color_Image_Editor_Imagick`.

plugins/webp-uploads/hooks.php
```diff
@@ -681,38 +681,52 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
 		i = d.createElement( i );
-		i.src = p + 'jIAAABXRUJQVlA4ICYAAACyAgCdASoCAAEALmk0mk0iIiIiIgBoSygABc6zbAAA/v56QAAAAA==';
+		i.src = p;
 		i.onload = function() {
 			i.onload = undefined;
-			i.src = p + 'h4AAABXRUJQVlA4TBEAAAAvAQAAAAfQ//73v/+BiOh/AAA=';
 		};
 
 		i.onerror = function() {
 			d.body.appendChild( s );
```
westonruter, line 750: It looks like `webpUploadsOutputFormat` isn't yet set as a global variable at this point. This would fix that:

```suggestion
			window.webpUploadsOutputFormat = webpUploadsOutputFormat;
			d.body.appendChild( s );
```

But perhaps it should be set as a data attribute on the `script` tag the same way that was done for `data-rest-api`?

plugins/webp-uploads/helper.php
adamsilverstein: > For example, with Dominant Color Images active, this could be Dominant_Color_Image_Editor_GD or Dominant_Color_Image_Editor_Imagick.

Ah right, good point.

> 		`if ( is_a( $editor, WP_Image_Editor_GD::class, true ) ) {`

I like this approach. Do you mean to have $editor as the second argument and the class as the first argument though? https://www.php.net/manual/en/function.is-a.php

adamsilverstein: Oh, NM I was reading that wrong - this isgood. would be nice to have some tests for the case you described though, `Dominant_Color_Image_Editor_Imagick`

plugins/webp-uploads/hooks.php
adamsilverstein: Good suggestion, I haven't quite gotten to testing this part yet.

adamsilverstein: Actually, this isn't a global, it is passed as the 6th argument to the `webpUploadsDetectFallback` function. I guess we can document that function.

westonruter: Right, but it is passed into that function but then it isn't passed on anywhere from that. How does it get passed to `wpPerfLab.webpUploadsFallbackWebpImages()` in `fallback.js`?

adamsilverstein: trying that..

adamsilverstein: addressed in [`d3dd80a` (#1176)](https://github.com/WordPress/performance/pull/1176/commits/d3dd80aa10549672b77d99d059ceed321fb4c203)

adamsilverstein:
This is close, needs some additional testing.

plugins/webp-uploads/fallback.js
```diff
@@ -1,8 +1,11 @@
-/* eslint no-var: "off", camelcase: "off" */
+/* eslint no-var: "off", camelcase: "off", no-undef: "off" */
 
 window.wpPerfLab = window.wpPerfLab || {};
 
 ( function ( document ) {
+	const webpUploadsOutputFormat = document
+		.getElementById( 'webpUploadsFallbackWebpImages' )
+		.getAttribute( 'output-format' );
```
westonruter:
```suggestion
		.getAttribute( 'data-output-format' );
```

plugins/webp-uploads/hooks.php
```diff
@@ -681,41 +681,58 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+		sprintf(
+			'( %s )( document, "img", "script", %s, %s )',
+			preg_replace( '/\s+/', ' ', $js_function ),
+			wp_json_encode( $detection_string ),
+			wp_json_encode( plugins_url( '/fallback.js', __FILE__ ) )
+		),
 		array(
 			'id'            => 'webpUploadsFallbackWebpImages',
 			'data-rest-api' => esc_url_raw( trailingslashit( get_rest_url() ) ),
+			'output-format' => webp_uploads_get_image_output_format(),
```
westonruter:
```suggestion
			'data-output-format' => webp_uploads_get_image_output_format(),
```

plugins/webp-uploads/fallback.js
```diff
@@ -1,8 +1,11 @@
-/* eslint no-var: "off", camelcase: "off" */
+/* eslint no-var: "off", camelcase: "off", no-undef: "off" */
```
westonruter: I believe this change can now be reverted?

```suggestion
/* eslint no-var: "off", camelcase: "off" */
```

plugins/webp-uploads/hooks.php
```diff
@@ -681,41 +681,58 @@ function webp_uploads_update_featured_image( $html, $post_id, $attachment_id ) {
...
+		sprintf(
+			'( %s )( document, "img", "script", %s, %s )',
+			preg_replace( '/\s+/', ' ', $js_function ),
+			wp_json_encode( $detection_string ),
+			wp_json_encode( plugins_url( '/fallback.js', __FILE__ ) )
+		),
 		array(
 			'id'            => 'webpUploadsFallbackWebpImages',
 			'data-rest-api' => esc_url_raw( trailingslashit( get_rest_url() ) ),
+			'data-output-format' => webp_uploads_get_image_output_format(),
```
westonruter: My bad. I made PHPCS unhappy.

```suggestion
			'id'                 => 'webpUploadsFallbackWebpImages',
			'data-rest-api'      => esc_url_raw( trailingslashit( get_rest_url() ) ),
			'data-output-format' => webp_uploads_get_image_output_format(),
```

phanduynam:
Do the current modifications help convert PNG to Avif/Webp on upload?

adamsilverstein:
> Do the current modifications help convert PNG to Avif/Webp on upload?

@phanduynam - not currently, however you can add this using a simple filter, see this comment for details: https://github.com/WordPress/performance/issues/371#issuecomment-2050294385

adamsilverstein:
Noting that I manually verified the fallback script works as expected (with the caveat that the srcset is not being output correctly when the JPEG generation is enabled.

adamsilverstein:
> when the JPEG generation is enabled.

Also: should we skip doing the detection at all is this is disabled since presumably there will be no fallback?

phanduynam:
Is the Avif image format version you are using the latest Avif 1.1.0 version?

https://aomediacodec.github.io/av1-avif/
https://aomediacodec.github.io/av1-avif/#change-list

plugins/webp-uploads/fallback.js
```diff
@@ -100,8 +112,12 @@ window.wpPerfLab = window.wpPerfLab || {};
 
 			if (
 				node.nodeName !== 'IMG' ||
-				( ! node.src.match( /\.webp$/i ) &&
-					! srcset.match( /\.webp\s+/ ) )
```
westonruter, line 104: This could also have just been:

```
				( ! node.src.match( /\.(webp|avif)$/i ) &&
					! srcset.match( /\.(webp|avif)\s+/ ) )
```

Maybe that would be more "scalable"?

plugins/webp-uploads/fallback.js
```diff
@@ -3,6 +3,9 @@
 window.wpPerfLab = window.wpPerfLab || {};
 
 ( function ( document ) {
+	const webpUploadsOutputFormat = document
+		.getElementById( 'webpUploadsFallbackWebpImages' )
```
westonruter, line 7: Minor point: script element could be stored in a variable and then reused below.

Alternatively, I wonder if we should not rely on attributes on the script tag to pass data from PHP since it could be that an optimization plugin concatenates the script with another and then the ID is no longer in the document. Perhaps some `wp_add_inline_script()` would be a more resilient approach? Doesn't have to be part of this PR.

plugins/webp-uploads/helper.php
```diff
@@ -16,20 +16,26 @@
...
  *
  * @since 1.0.0
+ * @since n.e.x.t Added support for AVIF.
  *
  * @return array<string, array<string>> An array of valid mime types, where the key is the mime type and the value is the extension type.
  */
 function webp_uploads_get_upload_image_mime_transforms(): array {
+
+	// Check the selected output format.
+	$output_format = webp_uploads_mime_type_supported( 'image/avif' ) ? get_option( 'perflab_modern_image_format' ) : 'webp';
```
```suggestion
	$output_format = webp_uploads_mime_type_supported( 'image/avif' ) ? webp_uploads_get_image_output_format() : 'webp';
```

plugins/webp-uploads/settings.php
```diff
@@ -47,24 +75,57 @@ function webp_uploads_add_media_settings_field(): void {
...
 add_action( 'admin_init', 'webp_uploads_add_media_settings_field' );
 
+/**
+ * Renders the settings field for the 'perflab_modern_image_format' setting.
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_generate_avif_webp_setting_callback(): void {
+
+	$selected       = get_option( 'perflab_modern_image_format' );
```
```suggestion
	$selected       = webp_uploads_get_image_output_format();
```

plugins/webp-uploads/settings.php
```diff
@@ -47,24 +75,57 @@ function webp_uploads_add_media_settings_field(): void {
...
+	$avif_supported = webp_uploads_mime_type_supported( 'image/avif' );
+	// Ensure WebP selected if AVIF is not supported.
+	if ( ! $avif_supported ) {
+		$selected = 'webp';
+	}
+	?>
+			<label for="perflab_modern_image_format">
+				<select name="perflab_modern_image_format" id="perflab_modern_image_format" aria-describedby="perflab_modern_image_format_description">
+					<option value="webp"<?php selected( 'webp', $selected ); ?>><?php esc_html_e( 'WebP', 'webp-uploads' ); ?></option>
+					<option value="avif"<?php selected( 'avif', $selected ); ?><?php echo $avif_supported ? '' : ' disabled'; ?>><?php esc_html_e( 'AVIF', 'webp-uploads' ); ?></option>
```
```suggestion
					<option value="avif"<?php selected( 'avif', $selected ); ?><?php disabled( ! $avif_supported ); ?>><?php esc_html_e( 'AVIF', 'webp-uploads' ); ?></option>
```

plugins/webp-uploads/settings.php
```diff
@@ -12,11 +12,28 @@
...
+ * @since n.e.x.t The `perflab_modern_image_format` was added to enable selecting an output format.
+ *                Currently includes AVIF and WebP.
  */
 function webp_uploads_register_media_settings_field(): void {
+	register_setting(
+		'media',
+		'perflab_modern_image_format',
+		array(
+			'sanitize_callback' => static function ( $image_format ): string {
+				return in_array( $image_format, array( 'webp', 'avif' ), true ) ? $image_format : 'webp';
```
westonruter: This same `in_array()` logic is used here and in `webp_uploads_get_image_output_format()`. Perhaps an opportunity to put extract this logic into a common helper to make it easier when a new image format comes out possibly?

plugins/webp-uploads/settings.php
```diff
@@ -12,11 +12,28 @@
...
+			},
+			'type'              => 'string',
+			'default'           => 'avif',                                                                                                                                    // AVIF is the default if the editor supports it.
```
westonruter, line 30: Sites currently using the plugin and getting WebP, when they upgrade to the new version of the plugin will find that `avif` is the new default if they support AVIF, right? Is that OK?

plugins/webp-uploads/fallback.js
adamsilverstein: yeah, I was following the existing pattern here, not sure why this approach was taken. My default would probably be to use `wp_localize_script`.  Let's discuss in a separate Issue.

plugins/webp-uploads/fallback.js
adamsilverstein: yea, that is probably better. I was wondering about handling fallbacks when users generate one format (WebP) for a while then switch to another format (AVIF). 

Currently the fallback will only handle the currently selected output format - it would be better if it were more flexible and handled both even on the same page. I haven't had a chance to work on that and I'm not sure it is even worth the effort. Ultimately, I feel like the fallback script is never going to be anything but a mediocre solution especially compared to using the `picture` element to implement a fallback.

plugins/webp-uploads/settings.php
adamsilverstein: good suggestion

plugins/webp-uploads/settings.php
adamsilverstein: I think it is ok because it is a lab plugin, lets discuss tomorrow at the chat. 

We could also bump the plugin version to 2.0, since this is a pretty major new feature; this helps indicate potentially breaking changes.

plugins/webp-uploads/settings.php
adamsilverstein: right, nice!

plugins/webp-uploads/settings.php
westonruter: Added as agenda topic: https://make.wordpress.org/core/2024/05/13/performance-chat-agenda-14-may-2024/#comment-46672

joemcgill: I think the change to the default option value should only apply if this option isn't already set. I assume that this is a new option, so is there any way we could tell if this feature was already in use, and if so, stick with the WebP option? For all new installs, defaulting to AVIF makes a lot of sense.

Additionally, in the sanitize callback, when an invalid option is passed, I am curious why we're falling back to WebP even though we're setting AVIF as the default. Is shouldn't we fallback to the same default?

westonruter: > I think the change to the default option value should only apply if this option isn't already set. I assume that this is a new option, so is there any way we could tell if this feature was already in use, and if so, stick with the WebP option? For all new installs, defaulting to AVIF makes a lot of sense.

I think the problem is that the option may not be set for existing sites. For example, there is a `perflab_generate_webp_and_jpeg` registered setting, but the option will only exist in the database if a user went to the Media screen and hit the Save Changes button.

> Additionally, in the sanitize callback, when an invalid option is passed, I am curious why we're falling back to WebP even though we're setting AVIF as the default. Is shouldn't we fallback to the same default?

Humm. Good point. I guess that should be made `avif` by default then, as well in `webp_uploads_get_image_output_format()`. If AVIF is not actually available, then it will force the use of WebP anyway.

westonruter: > I think the change to the default option value should only apply if this option isn't already set. I assume that this is a new option, so is there any way we could tell if this feature was already in use, and if so, stick with the WebP option? For all new installs, defaulting to AVIF makes a lot of sense.

Or maybe I misunderstood. This setting is new, yes. So you're saying that we should default to WebP if that setting is not set? But that setting will only get populated if the user goes to the settings screen and hits Save Changes, so in that case the default value wouldn't be used anyway.

joemcgill: Honestly, I'm conflicted about what we should do. I think there's some validity in maintaining the WebP default for any sites that already had the plugin installed—particularly since originally plugin was marketed only as a WebP plugin. But I'm also sympathetic to the fact that this is a feature plugin intended to improve the performance of a site by serving modern image formats and changing the default format to the most performant option that the site supports would also make sense.

I guess if we had a way to determine that the plugin was previously installed prior to AVIF support being added, we could use that to keep the site on WebP and prompt the site owner to opt-in to AVIF, while setting AVIF as the default for any new versions of the plugin, but I'm not sure if there's a good way to make that determinations—or if we even should.

westonruter: Wouldn't bumping the version to 2.0 as well as adding an Upgrade Notice to the readme.txt suffice?

joemcgill: Yes. I think it we're changing the default to AVIF, which will essentially change the format being used by existing sites that are currently using WebP, then those two things are good to do. Can we add a link in the upgrade notice or add an admin pointer after the upgrade is installed to alert users that they can control the format in the settings?

adamsilverstein: > Can we add a link in the upgrade notice or add an admin pointer after the upgrade is installed 

We could add a link to the upgrade notice, but would need to do that ahead of time in a separate release.

An admin pointer after the upgrade should work, I can try that.

westonruter: I thought the Upgrade Notice section of the readme got displayed somehow prior to upgrading:

https://github.com/WordPress/performance/blob/0a9570795dbc602cbc305d2b47b3a5a589f547cd/plugins/performance-lab/readme.txt#L485-L493

But I don't see it displayed anywhere that I can see when I'm downgraded to 2.9.0.

westonruter: Oh, I found it! The Web Stories plugin has an [upgrade notice](https://github.com/GoogleForCreators/web-stories-wp/blob/91460e88f8a2eecd0c8455f9a4f9b4a612677ef0/readme.txt#L141-L145) for the current version.

It gets displayed with the plugin on the Upgrade screen:

image

Honestly, I've never seen this before, and I rarely upgrade plugins from this screen now, as the upgrade from the plugins list table is a better experience. Still, maybe it is sufficient instead of adding an admin pointer?

adamsilverstein: > Still, maybe it is sufficient instead of adding an admin pointer?

Yes, honestly I prefer this approach. I find the admin pointers distracting and most people probably dismiss them without even seeing what they say.

plugins/webp-uploads/fallback.js
adamsilverstein: updated in 2266ea81ad344df77438213b6250877a67aa529f

plugins/webp-uploads/settings.php
adamsilverstein: addressed in afb34fa15e0abc65c2c7f60b4371b69b2f596928

plugins/webp-uploads/settings.php
```diff
@@ -12,11 +12,28 @@
...
+ *                Currently includes AVIF and WebP.
  */
 function webp_uploads_register_media_settings_field(): void {
+	register_setting(
+		'media',
+		'perflab_modern_image_format',
+		array(
+			'sanitize_callback' => static function ( $image_format ): string {
+				return webp_uploads_image_format_is_supported( $image_format ) ? $image_format : 'webp';
+			},
```
```suggestion
			'sanitize_callback' => 'webp_uploads_sanitize_image_format',
```

plugins/webp-uploads/helper.php
```diff
@@ -322,3 +328,59 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
...
+ * Check if an image format type is supported by the plugin.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $image_format The image format to check.
+ * @return bool Whether the image format is supported.
+ */
+function webp_uploads_image_format_is_supported( string $image_format ): bool {
+	return in_array( $image_format, array( 'webp', 'avif' ), true );
+}
```
```suggestion
/**
 * Sanitizes the image format.
 *
 * @since n.e.x.t
 *
 * @param string $image_format The image format to check.
 * @return string Supported image format.
 */
function webp_uploads_sanitize_image_format( string $image_format ): string {
	return in_array( $image_format, array( 'webp', 'avif' ), true ) ? $image_format : 'webp';
}
```

plugins/webp-uploads/helper.php
```diff
@@ -322,3 +328,59 @@ function webp_uploads_should_discard_additional_image_file( array $original, arr
...
+/**
+ * Get the image output format from the option, falling back to webp if unset.
+ *
+ * @since n.e.x.t
+ *
+ * @return string The image output format. One of 'webp' or 'avif'.
+ */
+function webp_uploads_get_image_output_format(): string {
+	$image_format = get_option( 'perflab_modern_image_format' );
+	return webp_uploads_image_format_is_supported( $image_format ) ? $image_format : 'webp';
```
```suggestion
	return webp_uploads_sanitize_image_format( $image_format );
```

westonruter (commented):
Could more of the logic be factored out into the helper function like this?

plugins/webp-uploads/settings.php
adamsilverstein: In 4754d21803a94d7574f0e122e485aac3e56357e0

plugins/webp-uploads/helper.php
adamsilverstein: In f6ba11d0c52e6bd5924cd909439b743822306202

plugins/webp-uploads/settings.php
adamsilverstein: Updated in [`ac925e4` (#1176)](https://github.com/WordPress/performance/pull/1176/commits/ac925e4e0346af4b0d7e35480d9e80cb68ea9b7c)

plugins/webp-uploads/settings.php
```diff
@@ -47,24 +73,56 @@ function webp_uploads_add_media_settings_field(): void {
...
+			if ( ! $avif_supported ) {
+				?>
+				<br />
+				<div class="notice notice-warning is-dismissible inline">
+					<p><b><?php esc_html_e( 'AVIF support is not available.', 'webp-uploads' ); ?></b></p>
+					<p><?php esc_html_e( 'AVIF support can only be enabled by your hosting provider, so contact them for more information.', 'webp-uploads' ); ?></p>
+				</div>
+				<?php
+			}
+			?>
```
westonruter: Looks like the formatting here got messed up a bit:

```suggestion
	<select name="perflab_modern_image_format" id="perflab_modern_image_format" aria-describedby="perflab_modern_image_format_description">
		<option value="webp"<?php selected( 'webp', $selected ); ?>><?php esc_html_e( 'WebP', 'webp-uploads' ); ?></option>
		<option value="avif"<?php selected( 'avif', $selected ); ?><?php disabled( ! $avif_supported ); ?>><?php esc_html_e( 'AVIF', 'webp-uploads' ); ?></option>
	</select>
	<label for="perflab_modern_image_format">
		<?php esc_html_e( 'Generate images in this format', 'webp-uploads' ); ?>
	</label>
	<p class="description" id="perflab_modern_image_format_description"><?php esc_html_e( 'Select the format to use when generating new images from uploaded JPEGs.', 'webp-uploads' ); ?></p>
	<?php if ( ! $avif_supported ) : ?>
		<br />
		<div class="notice notice-warning is-dismissible inline">
			<p><b><?php esc_html_e( 'AVIF support is not available.', 'webp-uploads' ); ?></b></p>
			<p><?php esc_html_e( 'AVIF support can only be enabled by your hosting provider, so contact them for more information.', 'webp-uploads' ); ?></p>
		</div>
	<?php endif;
```

tests/plugins/webp-uploads/load-tests.php
```diff
@@ -39,20 +42,29 @@ static function ( string $filename ) {
 
 	/**
 	 * Don't create the original mime type for JPEG images.
+	 *
+	 * @dataProvider data_provider_supported_image_types
+	 *
+	 * @test
```
westonruter: These `@test` annotations have been removed as of #1246

```suggestion
```

tests/plugins/webp-uploads/load-tests.php
```diff
@@ -95,16 +109,25 @@ public function test_it_should_create_the_original_mime_type_as_well_with_all_th
 	}
 
 	/**
-	 * Create JPEG and WebP for JPEG images, if opted in.
+	 * Create JPEG and output type for JPEG images, if opted in.
+	 *
+	 * @dataProvider data_provider_supported_image_types
+	 *
+	 * @test
```
westonruter:
```suggestion
```

tests/plugins/webp-uploads/load-tests.php
```diff
@@ -119,21 +142,31 @@ public function test_it_should_create_jpeg_and_webp_for_jpeg_images_if_opted_in(
...
 		}
 	}
 
 	/**
-	 * Create JPEG and WebP for JPEG images, if perflab_generate_webp_and_jpeg option set.
+	 * Create JPEG and output format for JPEG images, if perflab_generate_webp_and_jpeg option set.
+	 *
+	 * @dataProvider data_provider_supported_image_types
+	 *
+	 * @test
```
westonruter:
```suggestion
```

tests/plugins/webp-uploads/load-tests.php
```diff
@@ -632,43 +665,86 @@ public function test_it_should_preserve_jpeg_subsizes_using_transform_filter():
...
 
 	/**
-	 * Allow the upload of a WebP image if at least one editor supports the format
+	 * Allow the upload of an image if at least one editor supports the image type
+	 *
+	 * @dataProvider data_provider_supported_image_types
+	 *
+	 * @group current1
+	 *
+	 * @test
```
westonruter:
```suggestion
```

tests/plugins/webp-uploads/load-tests.php
```diff
@@ -632,43 +665,86 @@ public function test_it_should_preserve_jpeg_subsizes_using_transform_filter():
...
+		$this->assertImageHasSizeSource( $attachment_id, 'thumbnail', 'image/' . $image_type );
 	}
 
 	/**
-	 * Replace the featured image to WebP when requesting the featured image
+	 * Replace the featured image to the proper type when requesting the featured image.
+	 *
+	 * @dataProvider data_provider_supported_image_types
+	 *
+	 * @test
```
westonruter:
```suggestion
	 *
```

plugins/webp-uploads/settings.php
```diff
@@ -86,7 +139,6 @@ function webp_uploads_media_setting_style(): void {
 	</style>
 	<?php
 }
-add_action( 'admin_head-options-media.php', 'webp_uploads_media_setting_style' );
```
westonruter, line 89: The `webp_uploads_media_setting_style` function is now no longer used. Should it be moved to `deprecated.php` with a `_deprecated_function()` call?

tests/plugins/webp-uploads/load-tests.php
```diff
@@ -632,43 +665,86 @@ public function test_it_should_preserve_jpeg_subsizes_using_transform_filter():
...
+
+	/**
+	 * Data provider for tests returns the supported image types to run the tests against.
+	 *
+	 * @return array<array<string>> An array of valid image types.
+	 */
+	public function data_provider_supported_image_types(): array {
+		return array(
+			array( 'webp' ),
+			array( 'avif' ),
```
westonruter: Can help when wanting to run some specific test data with `--filter`:


```suggestion
	 * @return array<string, array<string>> An array of valid image types.
	 */
	public function data_provider_supported_image_types(): array {
		return array(
			'webp' => array( 'webp' ),
			'avif' => array( 'avif' ),
```

westonruter (commented):
Last minor requested changes!

joemcgill (approved):
This is testing well for me. Pre-approving pending @weston's feedback

tests/plugins/webp-uploads/load-tests.php
adamsilverstein: oh, thanks. i didn't see that and noticed they were removed in a merge.

plugins/webp-uploads/settings.php
adamsilverstein: ah, right. moved in f02c2d29ba33e019e847cd4bbe42c9a90b723705

westonruter (approved):
🎉

adamsilverstein:
> 🎉

"At least 1 approving review is required to merge this pull request."

adamsilverstein:
> > 🎉
> 
> "At least 1 approving review is required to merge this pull request."

https://github.com/WordPress/performance/pull/1257 should let you approve this one @westonruter

phanduynam:
Is the AVIF image file format version used the latest version 1.1.0 as described below?

https://aomediacodec.github.io/av1-avif/

phanduynam:
Is the AVIF image file format version being used the latest version 1.1.0 as described below?
https://aomediacodec.github.io/av1-avif/

PR #1182, opened 2024-04-25
Move PL plugin under `plugins/performance-lab`

### LINKED ISSUE #1165: Move PL plugin under `plugins/performance-lab`
## Feature Description

As mentioned by @joemcgill in https://github.com/WordPress/performance/issues/1012#issuecomment-1967337536, we should consider moving the PL to plugin to `plugins` dir.

thelovekesh:
## Summary

Fixes https://github.com/WordPress/performance/issues/1165

## Relevant technical choices

- Relocate the plugin to the plugins/performance-lab directory.
- Create a dummy plugin file in the root directory for wp-env mounting purposes and to display an administrative notification if an attempt is made to install it on a production environment.


performance.php
```diff
@@ -0,0 +1,36 @@
+<?php
+/**
+ * Plugin Name: Performance Lab Monorepo (not a real plugin)
+ * Plugin URI: https://github.com/WordPress/performance
+ * Description: The Performance Monorepo is not a plugin rather a collection of performance features as plugins. Download <a href="https://wordpress.org/plugins/performance-lab/" to install performance features instead.
+ * Author: WordPress Performance Team
+ * Author URI: https://make.wordpress.org/performance/
+ * Text Domain: performance-lab
```
thelovekesh, line 8: Initially, I kept it `performance` but tests have `performance-lab` text domain so switched it. But keeping it `performance-lab` doesn't seem right given it's already taken by the PL plugin.

performance.php
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Notify the admin that the Performance Monorepo is not a plugin,
+ * if they tried to install it as one.
+ */
+function perflab_monorepo_is_not_a_plugin() {
+	echo '<div class="notice notice-error"><p>';
+	printf(
+		wp_kses(
+			/* translators: Link to the Performance Lab plugin on WordPress.org */
+			__( 'The Performance Monorepo is not a plugin, and should not be installed as one. Download the <a href="%s">Performance Lab</a> plugin to install performance features instead.', 'performance-lab' ),
```
westonruter:
```suggestion
			__( 'The Performance Monorepo is not a plugin, and should not be installed as one. Download the <a href="%s">Performance Lab</a> plugin to install performance features instead. If you are developing one of the plugins, you can find them in the <code>plugins</code> directory and they should automatically have been added as plugins in wp-env. You may need to restart wp-env for the changes to take effect. If you are not using wp-env, you can add symlinks in your plugins directory to each of the plugins.', 'performance-lab' ),
```

tests/bootstrap.php
```diff
@@ -65,22 +64,19 @@ static function () use ( $plugin_test_path, $plugin_name ) {
...
-	tests_add_filter(
-		'plugins_loaded',
-		static function () {
-			require_once TESTS_PLUGIN_DIR . '/includes/admin/load.php';
-			require_once TESTS_PLUGIN_DIR . '/includes/admin/server-timing.php';
-			require_once TESTS_PLUGIN_DIR . '/includes/admin/plugins.php';
-		},
-		1
-	);
+	if ( 'performance-lab' === $plugin_name ) {
```
westonruter: Doesn't the `--testsuite` need to always be supplied now? If `$plugi_bane` is empty, shouldn't an error occur?

westonruter (commented):
Looking good!

tests/bootstrap.php
thelovekesh: I have set the default test suite to `performance-lab`. See https://github.com/WordPress/performance/pull/1182/commits/7a5dea4b49c09b82bb12158901e12ba39f68960e

westonruter:
Let's prioritize merging this right after the 3.1.0 release.

.wp-env.json
```diff
@@ -6,6 +6,7 @@
     "./plugins/dominant-color-images",
     "./plugins/embed-optimizer",
     "./plugins/optimization-detective",
+    "./plugins/performance-lab",
```
joemcgill, line 8: I think the first line, `".",` can be removed.

bin/plugin/commands/readme.js
```diff
@@ -75,10 +75,6 @@ function detectReadmePath( milestone ) {
 		);
 	}
 
-	if ( 'performance-lab' === slug[ 1 ] ) {
```
joemcgill: We should probably wait and update this PR once #1224 is merged.

bin/plugin/commands/since.js
```diff
@@ -41,10 +41,7 @@ exports.handler = async ( opt ) => {
 		);
 	}
 
-	if (
```
joemcgill, line 35: Similarly, we should wait an update this PR after #1223 is merged.

composer.json
```diff
@@ -63,57 +63,58 @@
     "lint": "phpcs",
     "lint:all": [
       "@lint",
+      "@lint:plugins"
+    ],
+    "lint:plugins": [
```
joemcgill: Is there a reason that we need a `lint:plugins` group at all,  instead of the whole list being consolidated under `lint:all`?

package.json
```diff
@@ -29,8 +29,8 @@
     "since": "./bin/plugin/cli.js since",
     "readme": "./bin/plugin/cli.js readme",
     "build": "wp-scripts build",
-    "build-plugins": "npm-run-all 'build:plugin:!(performance-lab)'",
-    "build:plugin:performance-lab": "rm -rf build/performance-lab && mkdir -p build/performance-lab && git archive HEAD | tar -x -C build/performance-lab",
+    "build-plugins": "npm-run-all 'build:plugin:*'",
```
joemcgill, line 33: This is working great!

.github/workflows/deploy-plugins.yml
joemcgill, line 1: We can't test this workflow in debug mode until after this PR is merged, so let's make sure to check that this works once after merging.

performance.php
joemcgill: I like the idea of providing some feedback to anyone who tries to install the whole repo. I think using the `performance-lab` text domain is likely fine. I wonder if we could even automatically load the root file for the actual Performance Lab plugin if someone ends up in this situation?

phpcs.xml.dist
```diff
@@ -1,16 +1,12 @@
 <?xml version="1.0"?>
-<ruleset name="WPP-PerfLab">
-	<description>WordPress Coding Standards for the Performance Lab Plugin</description>
+<ruleset name="WPP-Performance">
```
joemcgill: With this being the root ruleset for all WPP rules, maybe we just name it "WPP"?
```suggestion
<ruleset name="WPP">
```

plugins/performance-lab/phpcs.xml.dist
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0"?>
+<ruleset name="WPP-PerfLab">
```
joemcgill: This would be a good time to change the ruleset name to `WPP-PerformanceLab`.

joemcgill (commented):
I dig this idea and it's looking really good. I've added some feedback inline.

.wp-env.json
thelovekesh: That is required to mount the root files in the container.

joemcgill: Ah ok, is that so you can run unit tests from inside the container, or are there other reasons?

One of the side effects of adding that line to the `plugins` property of the `wp-env` object is that it will try to find and activate a plugin at the root, which I think we want to avoid. We might be better off using the `mappings` property to place any extra folders we need into the plugin directory root, or optionally, only do this in the test environment where they're needed, e.g.:

```
"env": {
        "tests": {
            "mappings": {
                "/wp-content/plugins/": "."
            },
        }
    }
```

The other thing I noticed while trying to test the unit tests with this setup is that we will need to remove the `$(basename $(pwd))` part of each of the npm scripts for running unit tests now that we're doing so from the plugin root and not from the performance lab plugin root.

thelovekesh: > Ah ok, is that so you can run unit tests from inside the container, or are there other reasons?

Yes, that's correct. Besides that, I don't see any other reasons.

Thanks for suggesting mappings. It's pointless to activate the base plugin as it only adds a confusing notice. I'll give it a try and hopefully, it will resolve this issue.

thelovekesh: > The other thing I noticed while trying to test the unit tests with this setup is that we will need to remove the $(basename $(pwd)) part of each of the npm scripts for running unit tests now that we're doing so from the plugin root and not from the performance lab plugin root.

We have an ongoing discussion on this topic at https://github.com/WordPress/performance/issues/1012. Perhaps we should address this improvement as part of that ticket? Currently, all our tooling is configured to run from the monorepo plugin root.

thelovekesh: Fixed in https://github.com/WordPress/performance/pull/1182/commits/23de67ac27966a97dc752e95ccd0c18eef5423db and https://github.com/WordPress/performance/pull/1182/commits/db957d47717f978c5c0673e6a518b5762d56ef19

composer.json
thelovekesh: Actually, no. Fixed it in https://github.com/WordPress/performance/pull/1182/commits/0596b062c52c7f4c584e69c7af3ea63981b274d4

thelovekesh:
In my opinion, the contents of `.gitattributes` for ignoring files in the archive are no longer relevant.

performance.php
thelovekesh: @joemcgill For users utilizing wp-env, it should function as expected. However, for those with custom setups, we should include documentation on how to work with the monorepo. Simply activating the PL plugin isn't sufficient since we need to provide references to other plugins as well. It could also be confusing if one plugin from the `/plugins` directory is loaded while others are not.

@westonruter Any thoughts here?

westonruter:
> In my opinion, the contents of `.gitattributes` for ignoring files in the archive are no longer relevant.

@thelovekesh Agreed. Let's eliminate all of these lines:

https://github.com/WordPress/performance/blob/be4c00af6f979e3939f365471fcf857cde1d76eb/.gitattributes#L3-L43

performance.php
westonruter: We're currently lacking such documentation for users who want to work on the standalone plugins as well, so I don't see this as a blocker.

westonruter: I've updated notices in 8cbac0c1f2ae3ed6d8c943192b85b1a1457d7b39 with a link to the handbook page. We can add documentation there for how to handle contributing when not using wp-env.

westonruter: image

phpstan.neon.dist
```diff
@@ -4,13 +4,13 @@ parameters:
...
 	treatPhpDocTypesAsCertain: false
 	paths:
-		- load.php
-		- includes/
-		- plugins/
-		- tests/
+		- plugins
+		- tests
+		- performance.php
+		- plugins/performance-lab/load.php
```
westonruter, line 10: This line is redundant since it is already inside the `plugins` directory now:

```suggestion
```

thelovekesh: @westonruter If I remove this, I got the following error:

```
 ------ ------------------------------------------------------------------------------------------------ 
  Line   tests/plugins/performance-lab/load-tests.php                                                    
 ------ ------------------------------------------------------------------------------------------------ 
  :35    Call to method PHPUnit\Framework\Assert::assertFalse() with int will always evaluate to false.  
  :50    Call to method PHPUnit\Framework\Assert::assertFalse() with int will always evaluate to false.  
  :73    Call to method PHPUnit\Framework\Assert::assertFalse() with int will always evaluate to false.  
  :144   Call to method PHPUnit\Framework\Assert::assertFalse() with int will always evaluate to false.  
 ------ ------------------------------------------------------------------------------------------------ 
```

I think it has something to do with the file loading order.

composer.json
```diff
@@ -71,53 +71,51 @@
...
-    "test-multisite": "phpunit -c tests/multisite.xml --verbose --testsuite performance-lab",
-    "test-multisite:all": [
-      "@test-multisite",
-      "@test-multisite:plugins"
-    ],
-    "test-multisite:auto-sizes": "phpunit -c tests/multisite.xml --verbose --testsuite auto-sizes",
-    "test-multisite:dominant-color-images": "phpunit -c tests/multisite.xml --verbose --testsuite dominant-color-images",
-    "test-multisite:embed-optimizer": "phpunit -c tests/multisite.xml --verbose --testsuite embed-optimizer",
-    "test-multisite:optimization-detective": "phpunit -c tests/multisite.xml --verbose --testsuite optimization-detective",
+    "test": "phpunit",
```
westonruter, line 86: `--verbose` is intentionally removed? I suppose there's no need for it to be supplied here, as it can be provided when actually invoking. Perhaps it should then be added in the workflow?

westonruter: Oh, nevermind. It _is_ still present below. Would anyone call `composer run-script test` directly then?

phpstan.neon.dist
westonruter: Interesting. I don't get that error with that line removed.

```
$ npm run phpstan

> phpstan
> composer phpstan

> phpstan analyse --memory-limit=2048M
Note: Using configuration file /home/westonruter/repos/performance/phpstan.neon.dist.
 107/107 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%


                                                                                                                        
 [OK] No errors                                                                                                         
                                                                                                                        
```

composer.json
thelovekesh: It's being called in:

https://github.com/WordPress/performance/blob/d9cd4d4d6e47de59e301228399c8fe8e400cf873/package.json#L46-L47

In case someone wants to use binary their way.

phpcs.xml.dist
```diff
@@ -1,16 +1,12 @@
...
-	<rule ref="WordPress.Files.FileName.NotHyphenatedLowercase">
-		<exclude-pattern>includes/server-timing/object-cache.copy.php</exclude-pattern>
-	</rule>
-
 	<file>.</file>
-	<exclude-pattern>./node_modules/*</exclude-pattern>
+
+	<!--Ignore plugins directory-->
 	<exclude-pattern>./plugins/*</exclude-pattern>
-	<exclude-pattern>./vendor/*</exclude-pattern>
```
westonruter, line 15: The `node_modules` and `vendor` directories are removed because they are excluded by default?

westonruter: Nevermind. They were added to `phpcs.ruleset.xml`

westonruter (approved):
Great work and great changes! 🎉

phpcs.xml.dist
thelovekesh: They are excluded in the global ruleset.

https://github.com/WordPress/performance/blob/d9cd4d4d6e47de59e301228399c8fe8e400cf873/bin/phpcs/phpcs.ruleset.xml#L136-L137

phpstan.neon.dist
thelovekesh: Interesting. Which version are you on?

performance.php
joemcgill: I think this is a good step. Thanks.

phpstan.neon.dist
westonruter: 1.11.1

thelovekesh: Can you try with `--debug` flag once since it disables the cache? Also, it was failing in CI as well - https://github.com/WordPress/performance/actions/runs/9111313232/job/25048167246

westonruter: Humm. No errors when passing `--debug` either.

thelovekesh: 🤔 

No configs override with `phpstan.neon` in the root?

westonruter: No 😞 

```
$ npm run phpstan -- -- --debug

> phpstan
> composer phpstan -- --debug

> phpstan analyse --memory-limit=2048M '--debug'

 ! [NOTE] The Xdebug PHP extension is active, but "--xdebug" is not used.                                               
 !        The process was restarted and it will not halt at breakpoints.                                                
 !        Use "--xdebug" if you want to halt at breakpoints.                                                            

Note: Using configuration file /home/westonruter/repos/performance/phpstan.neon.dist.
/home/westonruter/repos/performance/plugins/optimization-detective/class-od-data-validation-exception.php
...
/home/westonruter/repos/performance/performance.php

                                                                                                                        
 [OK] No errors                                                                                                         
                                                                                                                        

```

thelovekesh: Failing in `wp-env` as well with and without XDebug.

package.json
```diff
@@ -43,10 +43,10 @@
...
     "phpstan": "composer phpstan",
     "lint-php": "composer lint:all",
-    "test-php": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/$(basename $(pwd)) composer test",
-    "test-php-multisite": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/$(basename $(pwd)) composer test-multisite",
-    "test-php-plugins": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/$(basename $(pwd)) composer test:plugins",
-    "test-php-multisite-plugins": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/$(basename $(pwd)) composer test-multisite:plugins",
+    "test-php": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test",
+    "test-php-multisite": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test-multisite",
+    "test-php-plugins": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test:plugins",
+    "test-php-multisite-plugins": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test-multisite:plugins",
```
westonruter: Is `test-php` now obsolete since `performance-lab` is tested as part of `test-php-plugins`?

```suggestion
    "test-php": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test:plugins",
    "test-php-multisite": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test-multisite:plugins",
```

thelovekesh: Might be required in case someone want to play around with different PHPUnit cli options?

westonruter: Why? They can still pass them to to `test-php-plugins` just the same, right?

westonruter: A developer would be confused why there are two sets of `test` commands. At least, I was confused.

thelovekesh: If I just want to run a test in isolation for a single plugin only, it might be required IMO.

thelovekesh: > A developer would be confused why there are two sets of test commands. At least, I was confused.

That is also true.

westonruter: I think for that we should do something like what you've done in #1065. There needs to be an easy way to run tests for a single plugin via `npm`.

thelovekesh: I think we can go with your suggestion for now. In the future, it can be iterated given we have multiple pending discussions regarding such behaviors.

westonruter: I've just added a script locally to facilitate testing one or more plugins in isolation:

```bash
#!/bin/bash
for plugin in $@; do
	wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test:$plugin
done
```

Usage:

```bash
bash bin/test-single-plugin.sh performance-lab auto-sizes
```

joemcgill: I like the consolidation here because I agree that running `npm run test-php` should run the full suite, but since we're now running a set of individual suites, you can't pass other flags to `npm run test-php`. I think we can iterate on this in another followup issue though since as @thelovekesh [mentioned](https://github.com/WordPress/performance/pull/1182#discussion_r1610502152) we haven't aligned on our PHPUnit behavior in general.

joemcgill (approved):
I think we should go ahead and merge this. I think there's some follow-up work we need to do to make it easier to run PHPUnit locally, but that can be handled in a follow-up issue.

Issue #1088, opened 2024-03-25
Split out image-specific aspects of Optimization Detective to dependent plugin


westonruter:
## Feature Description

As discussed at depth in #1046, Optimization Detective is to be an infrastructure/framework/API dependency plugin which isn't promoted directly to end-users. Instead, a dependent plugin like "Image Loading Optimization" (or perhaps "Image Prioritizer") would be developed which extends Optimization Detective with the image-related aspects. 

To do this, we will need to ensure that Optimization Detective exposes an API that allows dependent plugins to hook into the detection and optimization process. For detection, I'm assuming this looks like a PHP filter that allows additional tags to be targeted for visiting by the tag processor. And then for optimization, this could look like an action that fires at each such tag which is visited.

joemcgill:
I like this idea, but wonder what the scope of a new Image Loading Optimization plugin would include and whether using client side data from the Optimization Detective Framework would be an optional feature of this type of plugin or if it would be a base requirement. For example, I could see a plugin that includes some of all of these options:

* Serve modern image formats (e.g., WebP, AVIF, etc.)
* Improve responsive image handling (e.g., auto-sizes, calculation improvements, srcset generation enhancements, etc.)
* Improve resource loading: (fetch-priority, lazy-loading, etc)
* CDN integration
* Use client side optimization enhancements

Alternatively, I could see some confusion if we had a standalone plugin for Image Loading Optimization, which was only optimizations built using the Optimizations Detective API but _also_ had individual feature plugins for things like auto-sizes, sizes calculation ehancements, webp, etc. that were not based on the new API.

felixarntz:
@joemcgill The scope of the "Image Loading Optimization" (which now may be renamed to "Image Prioritizer") would be the same as what Optimization Detective currently includes as a proof of concept: It's focused on applying `fetchpriority="high"` and `loading="lazy"` correctly. Any other image related features should be in separate plugins, such as for instead modern image formats being part of the "Modern Image Formats" plugin (formerly "WebP Uploads").

In other words, the image loading optimization plugin would have "Optimization Detective" as a dependency (i.e. required), since without it all it could do is enhance the limited server-side heuristics that core already comes with.

felixarntz:
[...]

Issue #1139, opened 2024-04-15
Progress indicator is missing when activating a feature


westonruter:
## Feature Description

When activating a feature on the Performance screen, there can be quite a bit of delay between the initial click and the page reloading with the success admin notice:

[Screen recording 2024-04-15 10.18.42.webm](https://github.com/WordPress/performance/assets/134745/eea5a8f6-17af-4792-9123-4a297f1c72f0)

This is because activating a feature may involve installing a plugin first before activating the plugin. There should be some spinner that appears with the Activate button that was clicked. Also, the other features' Activate buttons should also be disabled while this reload process is happening.

Perhaps better would be to use an Ajax process to activate the plugin rather than to do a full page reload. This would allow a user to click to activate multiple features without having to wait for a page reload (e.g. shiny updates).

mukeshpanchal27:
In #1091, I initially used the AJAX approach for installation and activation. However, based on the feedback https://github.com/WordPress/performance/pull/1091#discussion_r1539743622, I changed it to reload. In the AJAX approach, I added a loading button. If we agree to use AJAX, then I'm happy to open a PR that uses that approach.

westonruter:
See video in https://github.com/WordPress/performance/pull/1184 (for features with plugin dependencies) which shows further need for a spinner.

westonruter:
@mukeshpanchal27 Let's start with just the spinner as the quick fix, and then we can go with Ajax next.

PR #1184, opened 2024-04-29
Add plugin dependency support for activating performance features

westonruter:
## Summary

This is needed for the Image Prioritizer plugin which is to be a dependent of Optimization Detective. See #1088.

- Check for whether dependency plugins can be installed and activated when determining whether to show a feature can be activated or not.
- Install and activate dependencies when activating a feature.

## Relevant technical choices

The existing logic to check if a plugin is compatible with the current version of PHP and WP and whether the user can install or activate plugins is moved into a `perflab_get_plugin_availability()` helper function. This function is recursive to check not only the provided plugin slug but all of its dependencies and their dependencies. In the end, the `compatible_php` and `compatible_wp` array items returned will be true if _all_ of the plugin's dependencies and the plugin itself are true. The same is true for the checks for whether the user can install or activate each plugin.

Additionally, the current logic to install and activate a plugin in `perflab_install_activate_plugin_callback()` is extracted into a new helper function `perflab_install_and_activate_plugin()` which is also recursive. Given the provided plugin slug, it will activate and install the provided plugin's dependencies recursively before attempting to install and activate the provided plugin.

## Testing

Given a `perflab_get_standalone_plugin_data()` which is modified to include the `woocommerce-square` plugin, which is dependent on `woocommerce`:

```diff
--- a/load.php
+++ b/load.php
@@ -93,6 +93,10 @@ function perflab_get_standalone_plugin_data() {
 	 * - 'experimental' (boolean, optional)
 	 */
 	return array(
+		'woocommerce-square' => array(
+			'constant' => 'WOOCOMMERCE',
+			'experimental' => false,
+		),
 		'auto-sizes'              => array(
 			'constant'     => 'IMAGE_AUTO_SIZES_VERSION',
 			'experimental' => true,
```

Here is the behavior with this PR:

[Screen recording 2024-04-29 14.00.43.webm](https://github.com/WordPress/performance/assets/134745/a8a73057-8c86-4cc0-857c-8a21f32e0472)


westonruter:
@mukeshpanchal27 FYI: Still a work in progress, but I wanted to get your eyes on this early.

westonruter:
The need for a progress spinner (#1139) is made even more evident here.

includes/admin/plugins.php
```diff
@@ -150,6 +151,126 @@ function perflab_render_plugins_ui() {
...
+
+	$plugin_status = install_plugin_install_status( $plugin_data );
+
+	// The plugin is already installed or the user can install plugins.
+	$availability['can_install'] = (
+		'install' !== $plugin_status['status'] ||
+		current_user_can( 'install_plugins' )
+	);
+
+	// The plugin is installed and the user can activate it, or the use can activate plugins in general.
```
mukeshpanchal27:
```suggestion
	// The plugin is installed and the user can activate it, or the user can activate plugins in general.
```

mukeshpanchal27 (commented):
Thanks @westonruter for the PR!

Overall look good to me. I'm enable to tested old PHP/WP version plugins with dependancy as script get data from plugin repo.

Could we add notice in plugin card similar to WP core? Check screenshot.

Screenshot 2024-05-01 at 2 53 58 PM

westonruter:
@mukeshpanchal27 how did you get into that state where the dependent plugin is active but the dependency plugin is not? Normally this wouldn't be possible, right?

includes/admin/plugins.php
```diff
@@ -150,6 +151,126 @@ function perflab_render_plugins_ui() {
...
+ *
+ * Dependencies are recursively installed and activated as well.
+ *
+ * @since n.e.x.t
+ * @see perflab_get_plugin_availability()
+ *
+ * @param string $plugin_slug Plugin slug.
+ * @return WP_Error|null WP_Error on failure.
+ */
+function perflab_install_and_activate_plugin( string $plugin_slug ): ?WP_Error {
```
adamsilverstein: Should we add protection against infinite recursion here by ensuring we only handle a given plugin slug once at most. In theory if you had a plugin dependency chain that went back to the top plugin, this function would never exit (A->B->C->A). Track which slug has been handled and return early if a handled slug is called.

includes/admin/plugins.php
```diff
@@ -150,6 +151,126 @@ function perflab_render_plugins_ui() {
...
+/**
+ * Checks if a given plugin is available.
+ *
+ * @since n.e.x.t
+ * @see perflab_install_and_activate_plugin()
+ *
+ * @param array{name: string, slug: string, short_description: string, requires_php: string|false, requires: string|false, requires_plugins: string[], version: string} $plugin_data Plugin data from the WordPress.org API.
+ * @return array{compatible_php: bool, compatible_wp: bool, can_install: bool, can_activate: bool} Availability.
+ */
+function perflab_get_plugin_availability( array $plugin_data ): array {
```
adamsilverstein: same comment as below about protecting against infinite recursion.

includes/admin/plugins.php
westonruter: Good question. I was thinking core didn't account for this, but I stand corrected:

> [image](https://make.wordpress.org/core/2024/03/05/introducing-plugin-dependencies-in-wordpress-6-5/#:~:text=What%20happens%20if%20there%20are%20circular%20dependencies%3F)

Since our plugin has to support 6.4 still we can't reuse the validity-checking logic that core has, but it is overkill for our purposes here since we are managing all of the plugins that we list here. For that matter, checking for recursion may also be overkill since again we're managing all of these plugins, so we hopefully won't make such a recursive error when we list out the `Requires Plugins` field 😄

adamsilverstein: Right, it was mostly a reflexive comment as a general rule for recursive functions - make sure they don't infinitely recurse.

I agree since we control all the plugins here we are unlikely to introduce a circular dependency. Still the code change would be relatively small so maybe worth including anyway? I don't feel strongly so I'll leave it up to you.

adamsilverstein: not sure how core does it, maybe something like https://github.com/WordPress/performance/pull/1197/files

westonruter:
> Could we add notice in plugin card similar to WP core? Check screenshot.

@mukeshpanchal27 I've accounted for this in 7a5c9b0. If somehow a site gets into a state where the dependency is deactivated but the dependent is active, then the feature will be listed as if it was not active in the first place. The "Activate" button will then be shown:

Before | After
----|---
Screenshot 2024-05-01 11 18 13 | image

includes/admin/plugins.php
westonruter: @adamsilverstein I had a great conversation with Gemini about this: https://g.co/gemini/share/100242fb5096

I asked about how to address the circular dependency problem and it suggested a dependency tracker as an argument to the function, passed by reference. After some back and forth, I've come up with this: 1362cef.

adamsilverstein: >  had a great conversation with Gemini about this: g.co/gemini/share/100242fb5096

haha, that conversation is quite funny, thanks for sharing. The tips and feedback provided seem pretty through and insightful, I especially like the table with pros and cons of various approaches. Then at then end where it turns out it got some stuff wrong and you admonish it, pure gold.  

>  I've come up with this: 1362cef.

Thats perfect, thanks. Passing the processes list through as a parameter by reference is smart.

adamsilverstein (approved):
Terrific!

mukeshpanchal27 (approved):
Awesome ❤️

PR #1185, opened 2024-04-29
Facilitate embedding Optimization Detective in other plugins/themes

### LINKED ISSUE #1081: Facilitate Optimization Detective to be embedded in other plugins/themes

westonruter:
Fixes #1081

This copies the exact same boilerplate bootstrap logic as was implemented for Speculative Loading in https://github.com/WordPress/performance/pull/1159.

Review with whitespace changes suppressed: https://github.com/WordPress/performance/pull/1185/files?diff=unified&w=1

### DIFF
--- a/plugins/optimization-detective/load.php
+++ b/plugins/optimization-detective/load.php
@@ -64,7 +64,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 		}
 	}
 )(
-	'od_pending_plugin_info',
+	'optimization_detective_pending_plugin',
 	'0.1.1',
 	static function ( string $version ) {
 
--- a/plugins/optimization-detective/load.php
+++ b/plugins/optimization-detective/load.php
@@ -20,52 +20,102 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
-// Define the constant.
-if ( defined( 'OPTIMIZATION_DETECTIVE_VERSION' ) ) {
-	return;
-}
+(
+	/**
+	 * Register this copy of the plugin among other potential copies embedded in plugins or themes.
+	 *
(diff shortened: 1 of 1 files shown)


joemcgill (approved):
This looks right to me.

plugins/optimization-detective/load.php
```diff
@@ -20,52 +20,102 @@
...
+			||
+			// Otherwise, register this copy if it is actually the one installed in the directory for plugins.
+			rtrim( WP_PLUGIN_DIR, '/' ) === dirname( __DIR__ )
+		) {
+			$GLOBALS[ $global_var_name ]['version'] = $version;
+			$GLOBALS[ $global_var_name ]['load']    = $load;
+		}
+	}
+)(
+	'od_pending_plugin_info',
```
westonruter: Just going to tweak the variable name to make it more explicit.

```suggestion
	'optimization_detective_pending_plugin',
```

adamsilverstein (approved):
Excellent!

PR #1189, opened 2024-04-30
Ensure only HTML responses are optimized

westonruter:
I realized in https://github.com/WordPress/gutenberg/pull/61212 that Optimization Detective is currently attempting to optimize all frontend responses, including those which may not be HTML. (This does not apply to feeds or robots.txt, as it only is relevant when `template-loader.php` gets to filtering `template_include`.) For example, a plugin may decide to render an XML response instead for a template. These should be ignored from optimization.

### DIFF
--- a/plugins/optimization-detective/optimization.php
+++ b/plugins/optimization-detective/optimization.php
@@ -33,17 +33,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 function od_buffer_output( string $passthrough ): string {
 	ob_start(
 		static function ( string $output ): string {
-			$is_html_content_type = false;
-			foreach ( headers_list() as $header ) {
-				$header_parts = preg_split( '/\s*[:;]\s*/', strtolower( $header ) );
-				if ( count( $header_parts ) >= 2 && 'content-type' === $header_parts[0] ) {
-					$is_html_content_type = in_array( $header_parts[1], array( 'text/html', 'application/xhtml+xml' ), true );
-				}
-			}
-			if ( ! $is_html_content_type ) {
-				return $output;
-			}
-
 			/**
 			 * Filters the template output buffer prior to sending to the client.
 			 *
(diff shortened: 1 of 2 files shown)


plugins/optimization-detective/optimization.php
```diff
@@ -172,6 +172,31 @@ function od_construct_preload_links( array $lcp_elements_by_minimum_viewport_wid
...
+	$is_html_content_type = false;
+
+	$headers_list = array_merge(
+		array( 'Content-Type: ' . ini_get( 'default_mimetype' ) ),
+		headers_list()
+	);
+	foreach ( $headers_list as $header ) {
+		$header_parts = preg_split( '/\s*[:;]\s*/', strtolower( $header ) );
+		if ( count( $header_parts ) >= 2 && 'content-type' === $header_parts[0] ) {
+			$is_html_content_type = in_array( $header_parts[1], array( 'text/html', 'application/xhtml+xml' ), true );
```
westonruter, line 193: Note the `application/xhtml+xml` may be overkill per https://github.com/GoogleChromeLabs/wpp-research/pull/74, but I added it for completeness.

plugins/optimization-detective/optimization.php
```diff
@@ -172,6 +172,31 @@ function od_construct_preload_links( array $lcp_elements_by_minimum_viewport_wid
...
+ */
+function od_is_response_html_content_type(): bool {
+	$is_html_content_type = false;
+
+	$headers_list = array_merge(
+		array( 'Content-Type: ' . ini_get( 'default_mimetype' ) ),
+		headers_list()
+	);
+	foreach ( $headers_list as $header ) {
+		$header_parts = preg_split( '/\s*[:;]\s*/', strtolower( $header ) );
```
westonruter, line 191: Splitting on `:` and `;` here will result in `Content-Type: text/html; charset=utf-8` being split as `[ 'content-type', 'text/html', 'charset=utf-8' ]`. We then only use the first two items in the array.

swissspidy (approved):

PR #1188, opened 2024-04-29
Improve static analysis in Performance Lab plugin files

westonruter:
This fixes a bunch of static analysis issues identified by PHPStan, as well as a few from PhpStorm. This should go a long way to kickstart #775.

With PHPStan level 6, the following issues are resolved with this PR when running `composer phpstan load.php includes/*`:

<details><summary>PHPStan Output</summary>

```
 ------ ------------------------------------------------------------------------------------------------------------ 
  Line   includes/admin/load.php                                                                                     
 ------ ------------------------------------------------------------------------------------------------------------ 
  18     Function perflab_add_features_page() has no return type specified.                                          
  44     Function perflab_load_features_page() has no return type specified.                                         
  61     Function perflab_render_settings_page() has no return type specified.                                       
  79     Function perflab_admin_pointer() has no return type specified.                                              
  121    Function perflab_render_pointer() has no return type specified.                                             
  121    Function perflab_render_pointer() has parameter $args with no value type specified in iterable type array.  
         💡 See: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type                   
  207    Function perflab_dismiss_wp_pointer_wrapper() has no return type specified.                                 
  222    Function perflab_enqueue_features_page_scripts() has no return type specified.                              
  234    Function perflab_install_activate_plugin_callback() has no return type specified.                           
  276    Function perflab_print_features_page_style() has no return type specified.                                  
  300    Function perflab_plugin_admin_notices() has no return type specified.                                       
 ------ ------------------------------------------------------------------------------------------------------------ 

 ------ --------------------------------------------------------------------- 
  Line   includes/admin/plugins.php                                           
 ------ --------------------------------------------------------------------- 
  84     Function perflab_render_plugins_ui() has no return type specified.   
  286    Function perflab_render_plugin_card() has no return type specified.  
 ------ --------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------------------------------------- 
  Line   includes/admin/server-timing.php                                                                     
 ------ ----------------------------------------------------------------------------------------------------- 
  22     Function perflab_add_server_timing_page() has no return type specified.                              
  45     Function perflab_load_server_timing_page() has no return type specified.                             
  152    Function perflab_render_server_timing_page() has no return type specified.                           
  176    Function perflab_render_server_timing_page_hooks_field() has no return type specified.               
  208    Function perflab_render_server_timing_page_output_buffering_section() has no return type specified.  
[...]

PR #1190, opened 2024-04-30
Add Progress indicator when activating a feature

### LINKED ISSUE #1139: Progress indicator is missing when activating a feature

mukeshpanchal27:
## Summary

Fixes #1139


includes/admin/load.php
```diff
@@ -393,3 +396,30 @@ static function ( $name ) {
...
+				const target = event.target;
+				target.classList.add( 'updating-message' );
+				target.innerHTML = wp.i18n.__( 'Activating…' );
+			}
+		} );
+	} );
+} )( document );
+JS;
+
+	wp_print_inline_script_tag( $js );
```
westonruter: Best practice to not use `innerHTML` but rather `textContent`. This avoids the risk of injected markup.

Also, I don't believe the `'Activating…'` string will get translated here. I believe it will need to be set via PHP, but @swissspidy can confirm.

```suggestion
	$js_function = <<<JS
		function addPluginProgressIndicator( message ) {
			document.addEventListener( 'DOMContentLoaded', function () {
				document.addEventListener( 'click', function ( event ) {
					if (
						event.target.classList.contains(
							'perflab-install-active-plugin'
						)
					) {
						const target = event.target;
						target.classList.add( 'updating-message' );
						target.textContent = message;
					}
				} );
			} );
		}
JS;

	wp_print_inline_script_tag(
		sprintf(
			'( %s )( %s );',
			$js_function,
			wp_json_encode( __( 'Activating…', 'default' ) )
		),
		array( 'type' => 'module' )
	);
```

westonruter (changes requested):
Wow, so nice! 

[Screen recording 2024-04-29 22.15.19.webm](https://github.com/WordPress/performance/assets/134745/d63755e9-2c01-4ddf-a4a3-f126f335a1fc)

includes/admin/load.php
mukeshpanchal27: Thanks for feedback. Yes that string is not translated in inline script so i update in https://github.com/WordPress/performance/pull/1190/commits/59dc0945245dfecac2e99f87eee0daf57590f37a

westonruter (approved):

westonruter:
@mukeshpanchal27 this works in WP 6.4 as well, yes?

includes/admin/load.php
```diff
@@ -393,3 +396,37 @@ static function ( $name ) {
...
+			document.addEventListener( 'DOMContentLoaded', function () {
+				document.addEventListener( 'click', function ( event ) {
+					if (
+						event.target.classList.contains(
+							'perflab-install-active-plugin'
+						)
+					) {
+						const target = event.target;
+						target.classList.add( 'updating-message' );
+						target.textContent = message;
```
swissspidy, line 420: Ideally we would also use wp.a11y.speak

mukeshpanchal27:
> @mukeshpanchal27 this works in WP 6.4 as well, yes?

Yes, it works fine in 6.4.

includes/admin/load.php
mukeshpanchal27: Added in https://github.com/WordPress/performance/pull/1190/commits/4229c4fa4ee1da0763879dd8f10685ada7c27276

includes/admin/load.php
```diff
@@ -393,3 +396,39 @@ static function ( $name ) {
...
+
+						wp.a11y.speak(message);
```
swissspidy, line 422: Where are we enqueueing the a11y script?

mukeshpanchal27: Missed it. Added in https://github.com/WordPress/performance/pull/1190/commits/b5e675105396de0e655eeb29903a2d72ebdcad23

swissspidy (approved):

includes/admin/load.php
```diff
@@ -393,3 +399,39 @@ static function ( $name ) {
...
+				} );
+			} );
+		}
+JS;
+
+	wp_print_inline_script_tag(
+		sprintf(
+			'( %s )( %s );',
+			$js_function,
+			wp_json_encode( __( 'Activating…', 'default' ) )
```
westonruter, line 433: I just realized that this is `'Activating…'` when the core string appears to rather be `'Activating...'` (not using the ellipsis):

https://github.com/WordPress/wordpress-develop/blob/f9e5c6fb22a331746a1821e393c9ea6a6e6d70a0/src/js/_enqueues/wp/updates.js#L1090

Does this need to be amended?

includes/admin/load.php
```diff
@@ -393,3 +399,39 @@ static function ( $name ) {
...
+					if (
+						event.target.classList.contains(
+							'perflab-install-active-plugin'
+						)
+					) {
+						const target = event.target;
+						target.classList.add( 'updating-message' );
+						target.textContent = message;
+
+						wp.a11y.speak(message);
```
westonruter, line 422: Nit: Add spacing around the `message` param

includes/admin/load.php
```diff
@@ -393,3 +399,39 @@ static function ( $name ) {
 		);
 	}
 }
+
+/**
+ * Callback function that print plugin progress indicator script.
+ *
+ * @since n.e.x.t
+ */
+function perflab_print_plugin_progress_indicator_script() {
```
westonruter, line 408: Also, per #1188, might as well explicitly make this return `void`:

```php
function perflab_print_plugin_progress_indicator_script(): void {
```

PR #1192, opened 2024-04-30
Hide post embed iframes with visibility:hidden instead of clipping

### LINKED ISSUE #1114: Embed Optimizer breaks loading of Post Embeds
## Bug Description

When the Embed Optimizer plugin is active, it adds `loading=lazy` to a Post Embed `iframe`. Apparently because WordPress styles the initial iframe as pseudo-hidden with `clip: rect(1px, 1px, 1px, 1px)`, Chrome (at least) is interpreting this to mean that the iframe is hidden and thus shouldn't be loaded yet. However, since the iframe never loads, the JS inside the iframe never fires and doesn't communicate with the parent window to show the iframe.

## Steps to reproduce

1. Enable Embed Optimizer
2. Add a Post Embed
3. View the post with the embed on the frontend

## Screenshots

image

westonruter:
## Summary

Fixes #1114

## Relevant technical choices

When a lazy-loaded `iframe` is styled with `clip: rect(1px, 1px, 1px, 1px)`, it seems Chrome does not reliably load it. Interestingly, the iframe loads when navigating to the page from another page, but it does not load when doing an iframe.

If the `clip` is replaced with `visibility:hidden` (or other options per https://github.com/WordPress/performance/pull/1192#issuecomment-2096642854) then it does load reliably. This also fixes an accessibility problem whereby if a post embed iframe does not get displayed (such as due to JavaScript being disabled), when the iframe is hidden with `clip` then the user can still tab through links in the iframe which is confusing since the links are not shown (for sighted users); when `visibility:hidden` is used, then content is skipped from being navigated until it is made visible.

### Before

[Screen recording 2024-04-30 12.22.47.webm](https://github.com/WordPress/performance/assets/134745/8a232938-5630-46df-af60-2bd59910c7f9)

### After

[Screen recording 2024-04-30 12.23.41.webm](https://github.com/WordPress/performance/assets/134745/ccac72d9-f187-4a3f-a53e-477a2e69cb9f)


swissspidy:
As per my comment on the issue, this also needs to be tested with an iframe containing an image (post thumbnail) to ensure that is loading as expected too.

westonruter:
@swissspidy oh, got it. Here's a test page: https://retractable-crane-47a33d.instawp.xyz/pascal-birchler-on-democratizing-performance/

✅ Behavior in Chrome 123 on ChromeOS:

[Screen recording 2024-05-01 14.44.11.webm](https://github.com/WordPress/performance/assets/134745/7d2abcbf-8895-468f-be22-0bac20d53d7f)

westonruter:
✅ Behavior in Safari 17.4.1 on macOS 14.4.1:

https://github.com/WordPress/performance/assets/134745/45000c8e-fa5e-43d9-82bf-363f9eb8921d

westonruter:
✅ Behavior in Firefox 124.0.2 on macOS 14.4.1:

https://github.com/WordPress/performance/assets/134745/c475d195-037e-4cbc-a9d6-61d7b17aa5b7

plugins/embed-optimizer/hooks.php
```diff
@@ -75,6 +75,21 @@ function embed_optimizer_filter_oembed_html( string $html ): string {
 	if ( 1 === $iframe_count && $html_processor->has_bookmark( 'iframe' ) ) {
 		if ( $html_processor->seek( 'iframe' ) ) {
 			$html_processor->set_attribute( 'loading', 'lazy' );
+
+			// For post embeds, let them be transparent instead of almost completely clipped, since Chromium at least
+			// loads transparent lazy-loaded iframes while it does not load clipped lazy-loaded iframes.
```
westonruter:
```suggestion
			// loads transparent lazy-loaded iframes while it does not load clipped lazy-loaded iframes reliably.
```

westonruter:
The field guide posts on Make/Core are probably the most extensive uses of post embeds (that I've seen at least). So I tried applying these changes to the [6.4 field guide post](https://make.wordpress.org/core/2023/10/23/wordpress-6-4-field-guide/) with Chrome Local Overrides in DevTools. The changes are significant when testing with 6x CPU slowdown.

[Before trace](https://trace.cafe/t/gWHrPFoB21):

image

[After trace](https://trace.cafe/t/tRLN4J8qcB):

image

Before | After
--|--
image | image

LCP is reduced from 6932.4 ms to 3990.1 ms, so a **42.44% improvement**.

westonruter:
Actually, I can't confirm whether Local Overrides were working correctly when I did that test in https://github.com/WordPress/performance/pull/1192#issuecomment-2089236791. So I've copied the post content and put it on a test site: https://retractable-crane-47a33d.instawp.xyz/wordpress-6-4-field-guide/

[Trace before](https://trace.cafe/t/uU3gE6pSoq) | [Trace after](https://trace.cafe/t/OSTEMrWlbA)
--|--
image | image

LCP is reduced from 438.5 ms to 411.9 ms so a 6% improvement, at least for this single run.

westonruter:
I've opened https://github.com/WordPress/wordpress-develop/pull/6474 to apply this patch to core as well: https://core.trac.wordpress.org/ticket/58773#comment:26

swissspidy (approved):

adamsilverstein (approved):
Great!

westonruter:
After [discussing](https://wordpress.slack.com/archives/C02RP4X03/p1715011007477499) with @joedolson, he also [suggested](https://wordpress.slack.com/archives/C02RP4X03/p1715013929363499?thread_ts=1715011007.477499&cid=C02RP4X03) `visibility:hidden` as an alternative. I assumed this would be even more likely for browsers to consider hidden and avoid lazy-loading, but in fact Chrome, Safari, and Firefox all load such iframes. The other alternative which I've tested that works is to add a `width` and `height` of zero. Test cases: https://codepen.io/westonruter/pen/rNbXPgq?editors=1111

It seems going with `visibility:hidden` is a simpler approach than `opacity:0; pointer-events:none`. Also, it seems better than setting a zero `width`/`height` because then you have to worry about setting a zero `border` potentially. Also, since there is no need for screen readers to have access to the iframe prior to it being loaded (as regular visitors have no access to such an iframe either), I think it is preferable to go with `visibility:hidden`. 

Finally, I just discovered that `visibility:hidden` seems to also be an accessibility fix here because if a post embed iframe fails to be shown, such as when JavaScript is turned off, tabbing through a document with a post iframe hidden with `clip` (or with `opacity`) will result in a few tab stops occurring inside the hidden iframe which is confusing to sighted users navigating with the keyboard. If the iframe has `visibility:hidden` then no tab stops occur in the iframe.

PR #1193, opened 2024-05-01
Add optimization_detective_disabled query var to disable behavior

### LINKED ISSUE #1069: Provide mechanism to easily turn off optimizations applied by Optimization Detective for lab testing
## Feature Description

As sites start to adopt Optimization Detective and their pages are optimized, it would be very useful to be able to compare how the page loads with the optimizations applied as well as with the optimizations absent. This would allow us to do lab testing for live sites to see the impact. This would avoid us having to find sites in CrUX or HTTP Archive that used to not have the plugin and then got the plugin installed, in order to do a before/after comparison. (Granted, this should still be done with CrUX to get actual field impact.)

This could be done simply by adding a query parameter to the URL, for example `?disable-optimization-detective=1`.

westonruter:
Fixes #1069

This will allow us to add `?optimization_detective_disabled` to any URL running with Optimization Detective in order to generate the page without the optimizations. This will be useful for us to observe in the wild the performance gains by comparing a page with the non-optimized version.

### DIFF
--- a/plugins/optimization-detective/optimization.php
+++ b/plugins/optimization-detective/optimization.php
@@ -54,7 +54,7 @@ function od_buffer_output( string $passthrough ): string {
  * @access private
  */
 function od_maybe_add_template_output_buffer_filter(): void {
-	if ( ! od_can_optimize_response() ) {
+	if ( ! od_can_optimize_response() || isset( $_GET['optimization_detective_disabled'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 		return;
 	}
 	$callback = 'od_optimize_template_output_buffer';
--- a/tests/plugins/optimization-detective/optimization-tests.php
+++ b/tests/plugins/optimization-detective/optimization-tests.php
@@ -83,6 +83,20 @@ class OD_Optimization_Tests extends WP_UnitTestCase {
 		od_maybe_add_template_output_buffer_filter();
 		$this->assertTrue( has_filter( 'od_template_output_buffer' ) );
 	}
+	/**
+	 * Test od_maybe_add_template_output_buffer_filter().
+	 *
+	 * @covers ::od_maybe_add_template_output_buffer_filter
+	 */
+	public function test_od_maybe_add_template_output_buffer_filter_with_query_var_to_disable() {
+		$this->assertFalse( has_filter( 'od_template_output_buffer' ) );
+
+		add_filter( 'od_can_optimize_response', '__return_true' );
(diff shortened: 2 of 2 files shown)


mukeshpanchal27 (approved):

westonruter:
@joemcgill What do you think of this being a common pattern we use across other Performance Lab plugins? Being able to disable the functionality of a plugin in the wild could be a great way to compare how effective it is at improving performance. Naturally this wouldn't make sense for core, but for a feature plugin it could be valuable.

joemcgill:
> What do you think of this being a common pattern we use across other Performance Lab plugins?

Are you proposing that we would be able to disable all of our feature plugins by passing a query param for each feature, or are you thinking of a single param that would disable all of the features? I guess as a way to collect some A/B data, this could be useful, though we wouldn't really be able to use those params in collected user metrics, like CrUX.

Also, something to keep in mind is that even though most full-page caches are likely already handling GET requests with query params separately, there may be some where the addition of the query param may still result in a previous cached result being served.

westonruter:
@joemcgill I was thinking each feature would have their own query parameter, so we can granularly turn off optimizations to see their relative impact. This would only help in a lab context, as you note, as we wouldn't have field metrics for any A/B comparison.

And yes, full page caches may interfere with this, so when attempting to disable a feature we'll need to make sure that it is actually disabled when we load the page.

joemcgill:
Makes sense to me. I think this is one of those cases where having a single implementation that automatically gets incorporated into all our plugins would be useful. We don't currently have anything like that set up for this repo but maybe it's time to set that up so we can handle use cases like this one and the need to support embedded plugins, like #1081. What do you think?

westonruter:
@joemcgill Yeah, I was thinking about something like that too. However, doing something as low level as #1081 may be too low. A plugin may have multiple aspects to the optimizations it performs (e.g. fetchpriority and lazy loading) so it would be useful to be able to target each feature discretely, rather than be able to turn off the entire plugin altogether. So this makes me think more like "feature flag" query parameters are the most useful here. Given that each plugin will have their own set of features, perhaps it doesn't make sense to have a unified API for managing them? Or perhaps it could just be a common query var that we use, for example: 

```
?perflab_feature_disabled[]=foo&perflab_feature_disabled[]=bar&perflab_feature_disabled[]=baz
```

Then `$_GET['perflab_feature_disabled']` would be `['foo', 'bar', 'baz']`.

joemcgill:
I like the idea of a common pattern for feature flags. Let's try to be observant of when we see this type of pattern popping up in multiple plugins and open a new issue for consolidating logic/architecture when it makes sense.

PR #1197, opened 2024-05-01
Guard against infinite recursion

adamsilverstein:
silently succed when a circular dependency is encountered

## Summary

Fixes #

### DIFF
--- a/includes/admin/plugins.php
+++ b/includes/admin/plugins.php
@@ -151,6 +151,8 @@ function perflab_render_plugins_ui() {
 	<?php
 }
 
+global $perflab_get_plugin_availability_plugins_checked;
+
 /**
  * Checks if a given plugin is available.
  *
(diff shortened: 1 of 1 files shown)


includes/admin/plugins.php
```diff
@@ -161,6 +163,12 @@ function perflab_render_plugins_ui() {
  * @return array{compatible_php: bool, compatible_wp: bool, can_install: bool, can_activate: bool} Availability.
  */
 function perflab_get_plugin_availability( array $plugin_data ): array {
+	if ( empty( $perflab_get_plugin_availability_plugins_checked ) ) {
```
westonruter, line 167: Is this missing a `global`? AFAIK otherwise the `empty()` check here will always be `false` otherwise.

```suggestion
	global $perflab_get_plugin_availability_plugins_checked;
	if ( empty( $perflab_get_plugin_availability_plugins_checked ) ) {
```

includes/admin/plugins.php
```diff
@@ -213,6 +225,9 @@ function perflab_get_plugin_availability( array $plugin_data ): array {
  * @return WP_Error|null WP_Error on failure.
  */
 function perflab_install_and_activate_plugin( string $plugin_slug ): ?WP_Error {
+	if ( isset( $perflab_install_and_activate_plugin_plugins_activated[ $plugin_slug ] ) ) {
```
westonruter, line 230:
```suggestion
	global $perflab_install_and_activate_plugin_plugins_activated;
	if ( isset( $perflab_install_and_activate_plugin_plugins_activated[ $plugin_slug ] ) ) {
```

PR #1199, opened 2024-05-03
Update PHPStan level to 2

westonruter:
See #775.

This fixes the following PHPStan issues:

```
 ------ -------------------------------------------------------------------- 
  Line   plugins/dominant-color-images/helper.php                            
 ------ -------------------------------------------------------------------- 
  76     Call to an undefined method WP_Image_Editor::has_transparency().    
  82     Call to an undefined method WP_Image_Editor::get_dominant_color().  
 ------ -------------------------------------------------------------------- 
```

Note that PHPStan is correctly identifying a bug here which could cause a fatal error in case another plugin is filtering `wp_image_editors` so that our specific subclasses are not used.

### DIFF
--- a/plugins/dominant-color-images/helper.php
+++ b/plugins/dominant-color-images/helper.php
@@ -24,8 +24,8 @@ function dominant_color_set_image_editors( $editors ) {
 	}
 
 	$replaces = array(
-		'WP_Image_Editor_GD'      => 'Dominant_Color_Image_Editor_GD',
-		'WP_Image_Editor_Imagick' => 'Dominant_Color_Image_Editor_Imagick',
+		WP_Image_Editor_GD::class      => Dominant_Color_Image_Editor_GD::class,
+		WP_Image_Editor_Imagick::class => Dominant_Color_Image_Editor_Imagick::class,
 	);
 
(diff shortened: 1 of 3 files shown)


plugins/dominant-color-images/helper.php
```diff
@@ -58,6 +58,13 @@ function dominant_color_get_dominant_color_data( $attachment_id ) {
 		$file = get_attached_file( $attachment_id );
 	}
 	add_filter( 'wp_image_editors', 'dominant_color_set_image_editors' );
+
+	/**
```
mukeshpanchal27, line 62:
```suggestion
	/*
```
Seems it's local variable so do we needs to use multiline comment per https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-inline-comments

mukeshpanchal27 (approved):

plugins/dominant-color-images/helper.php
westonruter: I don't think that's right. Without `/**` then the phpdoc is not parsed.

westonruter: The docs you link to are talking about a regular multiline comment, not one that serves as a doc block.

PR #1206, opened 2024-05-07
Added server timing to REST API response

### LINKED ISSUE #1149: Server-Timing: Add in REST API
## Feature Description

Related: #770 

Here's how I did it manually by filtering `rest_post_dispatch`:

https://github.com/swissspidy/media-experiments/blob/c3b9d0c76734b215e1ac7015c0bc7a9dd96381ab/inc/functions.php#L1019-L1046

Related: https://core.trac.wordpress.org/ticket/49634

ashwinparthasarathi:
## Summary

Fixes #1149

## Relevant technical choices

Updated the core plugin to add ServerTiming to WP REST API end-points as well

## How to test
1. On the site, visit the REST endpoint http://localhost:8888/wp-json/ with the Dev Tools open on the Network Tab
2. Record the data in the Timing tab as seen [here](https://make.wordpress.org/performance/files/2023/06/image.png) before and after the change.
4. The difference is documented as follows,
- Before the change is introduced,
before-change-rest-api

- After the change,
after-change-rest-api

props: @swissspidy for the code.

### DIFF
--- a/includes/server-timing/hooks.php
+++ b/includes/server-timing/hooks.php
@@ -15,6 +15,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 /**
  * Adds server timing to REST API response.
  *
+ * @since n.e.x.t
+ *
  * @param WP_REST_Response|WP_Error $response Result to send to the client. Usually a `WP_REST_Response`.
  * @return WP_REST_Response|WP_Error Filtered response.
  */
--- a/includes/server-timing/hooks.php
+++ b/includes/server-timing/hooks.php
@@ -4,7 +4,7 @@
  *
  * @package performance-lab
  *
- * @since 1.0.0
+ * @since n.e.x.t
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
--- a/includes/server-timing/hooks.php
+++ b/includes/server-timing/hooks.php
@@ -11,7 +11,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly.
 }
 
-
 /**
  * Adds server timing to REST API response.
  *
@@ -20,18 +19,23 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @param WP_REST_Response|WP_Error $response Result to send to the client. Usually a `WP_REST_Response`.
  * @return WP_REST_Response|WP_Error Filtered response.
  */
-function rest_post_dispatch_add_server_timing( $response ) {
-	if ( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST || ! function_exists( 'perflab_server_timing' ) || ! $response instanceof WP_REST_Response ) {
+function perflab_rest_post_dispatch_add_server_timing( $response ) {
+	// TODO: Change condition to use wp_is_rest_endpoint() once minimum-supported version is WordPress 6.5.
+	if ( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST || ! $response instanceof WP_REST_Response ) {
 		return $response;
 	}
 
 	$server_timing = perflab_server_timing();
 
 	/** This filter is documented in includes/server-timing/class-perflab-server-timing.php */
-	do_action( 'perflab_server_timing_send_header' ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
+	do_action( 'perflab_server_timing_send_header' );
 
-	$response->header( 'Server-Timing', $server_timing->get_header() );
+	$header_value = $server_timing->get_header();
+
+	if ( $header_value ) {
+		$response->header( 'Server-Timing', $header_value );
+	}
 
 	return $response;
 }
-add_filter( 'rest_post_dispatch', 'rest_post_dispatch_add_server_timing', PHP_INT_MAX );
+add_filter( 'rest_post_dispatch', 'perflab_rest_post_dispatch_add_server_timing', PHP_INT_MAX );
--- a/includes/server-timing/hooks.php
+++ b/includes/server-timing/hooks.php
@@ -2,9 +2,9 @@
 /**
  * Hook callbacks used for Server Timing.
  *
- * @package dominant-color-images
+ * @package performance-lab
  *
- * @since 1.0.0
+ * @since n.e.x.t
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
@@ -19,20 +19,17 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @return WP_REST_Response|WP_Error Filtered response.
  */
 function rest_post_dispatch_add_server_timing( $response ) {
-	if ( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST ) {
-		return $response;
-	}
-
-	if ( ! function_exists( 'perflab_server_timing' ) || ! $response instanceof WP_REST_Response ) {
+	if ( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST || ! function_exists( 'perflab_server_timing' ) || ! $response instanceof WP_REST_Response ) {
 		return $response;
 	}
 
 	$server_timing = perflab_server_timing();
 
+	/** This filter is documented in includes/server-timing/class-perflab-server-timing.php */
 	do_action( 'perflab_server_timing_send_header' ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
 
 	$response->header( 'Server-Timing', $server_timing->get_header() );
 
 	return $response;
 }
-add_filter( 'rest_post_dispatch', 'rest_post_dispatch_add_server_timing' );
+add_filter( 'rest_post_dispatch', 'rest_post_dispatch_add_server_timing', PHP_INT_MAX );
--- a/includes/server-timing/hooks.php
+++ b/includes/server-timing/hooks.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * Hook callbacks used for Server Timing.
+ *
+ * @package dominant-color-images
+ *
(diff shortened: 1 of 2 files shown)


includes/server-timing/hooks.php
```diff
@@ -0,0 +1,38 @@
+<?php
+/**
+ * Hook callbacks used for Server Timing.
+ *
+ * @package dominant-color-images
```
westonruter:
```suggestion
 * @package performance-lab
```

includes/server-timing/hooks.php
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @since 1.0.0
```
westonruter:
```suggestion
 * @since n.e.x.t
```

includes/server-timing/hooks.php
```diff
@@ -0,0 +1,38 @@
...
+ * @return WP_REST_Response|WP_Error Filtered response.
+ */
+function rest_post_dispatch_add_server_timing( $response ) {
+	if ( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST ) {
+		return $response;
+	}
+
+	if ( ! function_exists( 'perflab_server_timing' ) || ! $response instanceof WP_REST_Response ) {
+		return $response;
+	}
```
westonruter: Perhaps these two conditionals could be combined?

includes/server-timing/hooks.php
```diff
@@ -0,0 +1,38 @@
...
+
+	$server_timing = perflab_server_timing();
+
+	do_action( 'perflab_server_timing_send_header' ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
```
westonruter:
```suggestion
	/** This filter is documented in includes/server-timing/class-perflab-server-timing.php */
	do_action( 'perflab_server_timing_send_header' ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
```

includes/server-timing/hooks.php
```diff
@@ -0,0 +1,38 @@
...
+
+	$response->header( 'Server-Timing', $server_timing->get_header() );
+
+	return $response;
+}
+add_filter( 'rest_post_dispatch', 'rest_post_dispatch_add_server_timing' );
```
westonruter: So that we account for other filters that run here.

```suggestion
add_filter( 'rest_post_dispatch', 'rest_post_dispatch_add_server_timing', PHP_INT_MAX );
```

westonruter (changes requested):

includes/server-timing/hooks.php
```diff
@@ -0,0 +1,36 @@
...
+ * @return WP_REST_Response|WP_Error Filtered response.
+ */
+function rest_post_dispatch_add_server_timing( $response ) {
+	if ( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST || ! function_exists( 'perflab_server_timing' ) || ! $response instanceof WP_REST_Response ) {
+		return $response;
+	}
+
+	$server_timing = perflab_server_timing();
+
+	/** This filter is documented in includes/server-timing/class-perflab-server-timing.php */
```
mukeshpanchal27, line 30:
```suggestion
```

Remove duplicate doc.

includes/server-timing/hooks.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+
+/**
+ * Adds server timing to REST API response.
+ *
```
mukeshpanchal27, line 16:
```suggestion
 * Adds server timing to REST API response.
 *
 * @since n.e.x.t
 *
```

Add `@since` for new function.

includes/server-timing/hooks.php
```diff
@@ -0,0 +1,37 @@
...
+ * @since n.e.x.t
+ *
+ * @param WP_REST_Response|WP_Error $response Result to send to the client. Usually a `WP_REST_Response`.
+ * @return WP_REST_Response|WP_Error Filtered response.
+ */
+function rest_post_dispatch_add_server_timing( $response ) {
+	if ( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST || ! function_exists( 'perflab_server_timing' ) || ! $response instanceof WP_REST_Response ) {
```
westonruter: Note, once WordPress 6.5 is the minimum-supported version, this code:

```php
! defined( 'REST_REQUEST' ) || ! REST_REQUEST
```

I believe can be replaced with:

```php
! wp_is_rest_endpoint()
```

includes/server-timing/hooks.php
```diff
@@ -0,0 +1,37 @@
...
+ * @package performance-lab
+ *
+ * @since n.e.x.t
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+
```
westonruter: Extra line break can be removed.

```suggestion

```

includes/server-timing/hooks.php
```diff
@@ -0,0 +1,37 @@
...
+ */
+function rest_post_dispatch_add_server_timing( $response ) {
+	if ( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST || ! function_exists( 'perflab_server_timing' ) || ! $response instanceof WP_REST_Response ) {
+		return $response;
+	}
+
+	$server_timing = perflab_server_timing();
+
+	/** This filter is documented in includes/server-timing/class-perflab-server-timing.php */
+	do_action( 'perflab_server_timing_send_header' ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
```
westonruter: Why is suppressing `WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound` needed here? It's not being triggered by our PHPCS config. I think it should be removed:

```suggestion
	do_action( 'perflab_server_timing_send_header' );
```

Perhaps it's due to a local PHPCS config you have? Perhaps we should add the `WordPress.NamingConventions.PrefixAllGlobals` sniff to this repo and configure it to require the plugin-specific prefix(es).

westonruter (changes requested):
Almost there.

includes/server-timing/hooks.php
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Adds server timing to REST API response.
+ *
+ * @since n.e.x.t
+ *
+ * @param WP_REST_Response|WP_Error $response Result to send to the client. Usually a `WP_REST_Response`.
+ * @return WP_REST_Response|WP_Error Filtered response.
+ */
+function rest_post_dispatch_add_server_timing( $response ) {
```
joemcgill: This function is essentially part of the performance lab plugin, so should use the proper prefix for the plugin. The `add_filter` call below will need to be updated as well.
```suggestion
function perflab_rest_post_dispatch_add_server_timing( $response ) {
```

includes/server-timing/hooks.php
joemcgill: This line is fairly long, consider breaking into multiple lines. I'm also not certain that the `function_exists()` check is needed, since this function is defined in the same plugin.

```suggestion
if ( 
    ! defined( 'REST_REQUEST' ) || ! REST_REQUEST
    || ! $response instanceof WP_REST_Response
) {
```

includes/server-timing/hooks.php
```diff
@@ -0,0 +1,37 @@
...
+	if ( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST || ! function_exists( 'perflab_server_timing' ) || ! $response instanceof WP_REST_Response ) {
+		return $response;
+	}
+
+	$server_timing = perflab_server_timing();
+
+	/** This filter is documented in includes/server-timing/class-perflab-server-timing.php */
+	do_action( 'perflab_server_timing_send_header' ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
+
+	$response->header( 'Server-Timing', $server_timing->get_header() );
```
joemcgill: The `Perflab_Server_Timing::send_header()` method checks to see if there are any headers returned by the `get_header()` method prior to sending header info. Doing something similar here probably makes sense.

```suggestion
	$header_value = $server_timing->get_header();
	
	if ( $header_value ) {
		$response->header( 'Server-Timing', $header_value );
	}
```

Alternately, rather than duplicating a lot of the logic from the `send_header()` method to this filter, I wonder if we should add a `set_rest_header()` method to the `Perflab_Server_Timing` class that we could all in this filter instead?

joemcgill (changes requested):
Thanks! I've left a lot of feedback, but the only real blocking issue I see is the function name being introduced.

ashwinparthasarathi (commented):
Changed code based on review comments.

includes/server-timing/hooks.php
ashwinparthasarathi: noted and added a TODO for future. I hope that is the best way to get it documented.

includes/server-timing/hooks.php
```diff
@@ -0,0 +1,41 @@
...
+
+	$server_timing = perflab_server_timing();
+
+	/** This filter is documented in includes/server-timing/class-perflab-server-timing.php */
+	do_action( 'perflab_server_timing_send_header' );
+
+	$header_value = $server_timing->get_header();
+
+	if ( $header_value ) {
+		$response->header( 'Server-Timing', $header_value );
```
westonruter: Just realized that this would override any existing `Server-Timing` header. Setting `$replace` to `false` will prevent that:

```suggestion
		$response->header( 'Server-Timing', $header_value, false );
```

adamsilverstein (approved):
Nice work!

westonruter (approved):

PR #1208, opened 2024-05-07
Updated code to display plugin settings link in the features screen and fix responsive layout for mobile

### LINKED ISSUE #1137: Add settings links to Performance features screen
## Feature Description

Certain plugins have settings (e.g. Modern Image Formats and Speculative Loading). However, discovery of how to access these plugins' settings is not facilitated on the Performance features screen:

image

Just as we've added plugin row action links for these plugins (#1131, #1036), we should facilitate discovery on the Performance screen as well.

ashwinparthasarathi:
## Summary

Fixes #1137

## Relevant technical choices

1. Made changes in the Performance Features admin page code.
2. Updated the plugin card rendering to display the plugin settings link if the plugin is active and if it has a settings page.

## Result
1. Currently shows the Settings link for the 2 modules that have it, Speculative Loading & Modern Image Formats
image

**Update:** See https://github.com/WordPress/performance/pull/1208#issuecomment-2115966001 for how the responsive layout is also fixed for mobile.

### DIFF
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -333,7 +333,7 @@ function perflab_plugin_admin_notices(): void {
 		$plugin_settings_link = perflab_get_plugin_settings_link( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 		wp_admin_notice(
 			/* translators: %s: Settings link of the plugin */
-			sprintf( esc_html__( 'Feature activated. Review %s.', 'performance-lab' ), $plugin_settings_link ),
+			wp_kses( sprintf( __( 'Feature activated. Review %s.', 'performance-lab' ), $plugin_settings_link ), array( 'a' => array( 'href' => array() ) ) ),
 			array(
 				'type'        => 'success',
 				'dismissible' => true,
@@ -382,12 +382,11 @@ JS;
  * Callback function that returns plugin settings link.
  *
  * @param string $plugin_slug Plugin slug passed to generate the settings link.
- * @return string Either the plugin settings link or empty.
+ * @return string|null Either the plugin settings link or null if not available.
  * @since n.e.x.t
  */
-function perflab_get_plugin_settings_link( string $plugin_slug ): string {
-	$plugin_file  = '';
-	$plugin_links = '';
+function perflab_get_plugin_settings_link( string $plugin_slug ): ?string {
+	$plugin_file = null;
 
 	foreach ( array_keys( get_plugins() ) as $file ) {
 		if ( strtok( $file, '/' ) === $plugin_slug ) {
@@ -396,14 +395,16 @@ function perflab_get_plugin_settings_link( string $plugin_slug ): string {
 		}
 	}
 
-	if ( ! empty( $plugin_file ) ) {
-		/** This filter is documented in wp-admin/includes/class-wp-plugins-list-table.php */
-		$plugin_links = apply_filters( "plugin_action_links_{$plugin_file}", array() );
+	if ( null === $plugin_file ) {
+		return null;
 	}
 
-	if ( array_key_exists( 'settings', $plugin_links ) ) {
+	/** This filter is documented in wp-admin/includes/class-wp-plugins-list-table.php */
+	$plugin_links = apply_filters( "plugin_action_links_{$plugin_file}", array() );
+
+	if ( is_array( $plugin_links ) && array_key_exists( 'settings', $plugin_links ) ) {
 		return $plugin_links['settings'];
 	}
 
-	return '';
+	return null;
 }
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -332,11 +332,11 @@ function perflab_plugin_admin_notices(): void {
 	if ( isset( $_GET['activate'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 		$message = __( 'Feature activated.', 'performance-lab' );
 
-		$plugin_settings_link = perflab_get_plugin_settings_link( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$plugin_settings_url = perflab_get_plugin_settings_url( wp_unslash( $_GET['activate'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 
-		if ( $plugin_settings_link ) {
+		if ( $plugin_settings_url ) {
 			/* translators: %s is the settings URL */
-			$message .= ' ' . sprintf( __( 'Review <a href="%s">settings</a>.', 'performance-lab' ), esc_url( $plugin_settings_link ) );
+			$message .= ' ' . sprintf( __( 'Review <a href="%s">settings</a>.', 'performance-lab' ), esc_url( $plugin_settings_url ) );
 		}
 
 		wp_admin_notice(
@@ -393,14 +393,14 @@ JS;
 }
 
 /**
- * Callback function that returns plugin settings link.
+ * Gets the URL to the plugin settings screen if one exists.
  *
  * @since n.e.x.t
  *
  * @param string $plugin_slug Plugin slug passed to generate the settings link.
- * @return string|null Either the plugin settings link or null if not available.
+ * @return string|null Either the plugin settings URL or null if not available.
  */
-function perflab_get_plugin_settings_link( string $plugin_slug ): ?string {
+function perflab_get_plugin_settings_url( string $plugin_slug ): ?string {
 	$plugin_file = null;
 
 	foreach ( array_keys( get_plugins() ) as $file ) {
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -330,10 +330,27 @@ function perflab_plugin_admin_notices(): void {
 	}
 
 	if ( isset( $_GET['activate'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$message = __( 'Feature activated.', 'performance-lab' );
+
 		$plugin_settings_link = perflab_get_plugin_settings_link( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+
+		if ( $plugin_settings_link ) {
+			$p = new WP_HTML_Tag_Processor( $plugin_settings_link );
+			if ( ! empty( $p->next_tag( array( 'tag_name' => 'A' ) ) ) && ! empty( $p->get_attribute( 'href' ) ) && ! is_bool( $p->get_attribute( 'href' ) ) ) {
+				/* translators: %s is the settings URL */
+				$message .= ' ' . sprintf( __( 'Review <a href="%s">settings</a>.', 'performance-lab' ), esc_url( $p->get_attribute( 'href' ) ) );
+			}
+		}
+
 		wp_admin_notice(
-			/* translators: %s: Settings link of the plugin */
-			wp_kses( sprintf( __( 'Feature activated. Review %s.', 'performance-lab' ), $plugin_settings_link ), array( 'a' => array( 'href' => array() ) ) ),
+			wp_kses(
+				$message,
+				array(
+					'a' => array(
+						'href' => array(),
+					),
+				)
+			),
 			array(
 				'type'        => 'success',
 				'dismissible' => true,
@@ -381,9 +398,10 @@ JS;
 /**
  * Callback function that returns plugin settings link.
  *
+ * @since n.e.x.t
+ *
  * @param string $plugin_slug Plugin slug passed to generate the settings link.
  * @return string|null Either the plugin settings link or null if not available.
- * @since n.e.x.t
  */
 function perflab_get_plugin_settings_link( string $plugin_slug ): ?string {
 	$plugin_file = null;
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -287,7 +287,8 @@ function perflab_print_features_page_style(): void {
 		margin-left: 0;
 	}
 	.plugin-card-top {
-		min-height: auto;
+		/* This is required to ensure the Settings link does not extend below the bottom of a plugin card on a wide screen */
+		min-height: 90px;
 	}
 	.plugin-card .perflab-plugin-experimental {
 		font-size: 80%;
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -335,11 +335,8 @@ function perflab_plugin_admin_notices(): void {
 		$plugin_settings_link = perflab_get_plugin_settings_link( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 
 		if ( $plugin_settings_link ) {
-			$p = new WP_HTML_Tag_Processor( $plugin_settings_link );
-			if ( ! empty( $p->next_tag( array( 'tag_name' => 'A' ) ) ) && ! empty( $p->get_attribute( 'href' ) ) && ! is_bool( $p->get_attribute( 'href' ) ) ) {
-				/* translators: %s is the settings URL */
-				$message .= ' ' . sprintf( __( 'Review <a href="%s">settings</a>.', 'performance-lab' ), esc_url( $p->get_attribute( 'href' ) ) );
-			}
+			/* translators: %s is the settings URL */
+			$message .= ' ' . sprintf( __( 'Review <a href="%s">settings</a>.', 'performance-lab' ), esc_url( $plugin_settings_link ) );
 		}
 
 		wp_admin_notice(
@@ -420,8 +417,17 @@ function perflab_get_plugin_settings_link( string $plugin_slug ): ?string {
 	/** This filter is documented in wp-admin/includes/class-wp-plugins-list-table.php */
 	$plugin_links = apply_filters( "plugin_action_links_{$plugin_file}", array() );
 
-	if ( is_array( $plugin_links ) && array_key_exists( 'settings', $plugin_links ) ) {
-		return $plugin_links['settings'];
+	if ( ! is_array( $plugin_links ) || ! array_key_exists( 'settings', $plugin_links ) ) {
+		return null;
+	}
+
+	$p = new WP_HTML_Tag_Processor( $plugin_links['settings'] );
+	if ( ! $p->next_tag( array( 'tag_name' => 'A' ) ) ) {
+		return null;
+	}
+	$href = $p->get_attribute( 'href' );
+	if ( $href && is_string( $href ) ) {
+		return $href;
 	}
 
 	return null;
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -329,11 +329,29 @@ function perflab_plugin_admin_notices(): void {
 		}
 	}
 
-	if ( isset( $_GET['activate'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+	/**
+	 * Sanitizes plugin slug.
+	 *
+	 * @param string $unsanitized_plugin_slug Unsanitized plugin slug.
+	 * @return string|null Sanitized plugin slug or null if invalid.
+	 */
+	$sanitize_plugin_slug = static function ( string $unsanitized_plugin_slug ): ?string {
+		if ( in_array( $unsanitized_plugin_slug, perflab_get_standalone_plugins(), true ) ) {
+			return $unsanitized_plugin_slug;
+		} else {
+			return null;
+		}
+	};
+
+	$activated_plugin_slug = null;
+	if ( isset( $_GET['activate'] ) && is_string( $_GET['activate'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$activated_plugin_slug = $sanitize_plugin_slug( (string) wp_unslash( $_GET['activate'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+	}
+
+	if ( $activated_plugin_slug ) {
 		$message = __( 'Feature activated.', 'performance-lab' );
 
-		$plugin_settings_url = perflab_get_plugin_settings_url( wp_unslash( $_GET['activate'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-
+		$plugin_settings_url = perflab_get_plugin_settings_url( $activated_plugin_slug );
 		if ( $plugin_settings_url ) {
 			/* translators: %s is the settings URL */
 			$message .= ' ' . sprintf( __( 'Review <a href="%s">settings</a>.', 'performance-lab' ), esc_url( $plugin_settings_url ) );
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -262,7 +262,7 @@ function perflab_install_activate_plugin_callback(): void {
 	$url = add_query_arg(
 		array(
 			'page'     => PERFLAB_SCREEN,
-			'activate' => 'true',
+			'activate' => $plugin_slug,
 		),
 		admin_url( 'options-general.php' )
 	);
@@ -330,8 +330,10 @@ function perflab_plugin_admin_notices(): void {
 	}
 
 	if ( isset( $_GET['activate'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$plugin_settings_link = perflab_get_plugin_settings_link( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 		wp_admin_notice(
-			esc_html__( 'Feature activated.', 'performance-lab' ),
+			/* translators: %s: Settings link of the plugin */
+			sprintf( esc_html__( 'Feature activated. Review %s.', 'performance-lab' ), $plugin_settings_link ),
 			array(
 				'type'        => 'success',
 				'dismissible' => true,
@@ -374,4 +376,34 @@ JS;
 		),
 		array( 'type' => 'module' )
 	);
+}
+
+/**
+ * Callback function that returns plugin settings link.
+ *
+ * @param string $plugin_slug Plugin slug passed to generate the settings link.
+ * @return string Either the plugin settings link or empty.
+ * @since n.e.x.t
+ */
+function perflab_get_plugin_settings_link( string $plugin_slug ): string {
+	$plugin_file  = '';
+	$plugin_links = '';
+
+	foreach ( array_keys( get_plugins() ) as $file ) {
+		if ( strtok( $file, '/' ) === $plugin_slug ) {
+			$plugin_file = $file;
+			break;
+		}
+	}
+
+	if ( ! empty( $plugin_file ) ) {
+		/** This filter is documented in wp-admin/includes/class-wp-plugins-list-table.php */
+		$plugin_links = apply_filters( "plugin_action_links_{$plugin_file}", array() );
+	}
+
+	if ( array_key_exists( 'settings', $plugin_links ) ) {
+		return $plugin_links['settings'];
+	}
+
+	return '';
 }
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -232,6 +232,22 @@ function perflab_enqueue_features_page_scripts(): void {
 }
 
 /**
+ * Sanitizes a plugin slug.
+ *
+ * @since 3.1.0
+ *
+ * @param mixed $unsanitized_plugin_slug Unsanitized plugin slug.
+ * @return string|null Validated and sanitized slug or else null.
+ */
+function perflab_sanitize_plugin_slug( $unsanitized_plugin_slug ): ?string {
+	if ( in_array( $unsanitized_plugin_slug, perflab_get_standalone_plugins(), true ) ) {
+		return $unsanitized_plugin_slug;
+	} else {
+		return null;
+	}
+}
+
+/**
  * Callback for handling installation/activation of plugin.
  *
  * @since 3.0.0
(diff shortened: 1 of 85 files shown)


includes/admin/plugins.php
```diff
@@ -325,6 +324,12 @@ function perflab_render_plugin_card( array $plugin_data ): void {
 			'<button type="button" class="button button-disabled" disabled="disabled">%s</button>',
 			esc_html( _x( 'Active', 'plugin', 'default' ) )
 		);
+		$plugin_path    = $plugin_data['slug'] . '/load.php';
```
westonruter: Note that not every plugin uses `load.php` as the plugin file. Better to not hard-code this.

ashwinparthasarathi: I agree, what would be the best way to not hard code this in this situation?

westonruter: You can use `array_keys( get_plugins() )` to get a list of all the plugin files, and then you can get the dirname of each to find the slugs to come up with a mapping of slug to file (e.g. `auto-sizes` to `auto-sizes/auto-sizes.php`)

ashwinparthasarathi: Made the changes by fetching the path from the installed plugins.

includes/admin/plugins.php
```diff
@@ -325,6 +324,20 @@ function perflab_render_plugin_card( array $plugin_data ): void {
...
+		$all_plugins    = array_keys( get_plugins() );
+		$plugin_slug    = $plugin_data['slug'];
+		$plugin_path    = '';
+		foreach ( $all_plugins as $path ) {
+			if ( strpos( $path, $plugin_slug ) !== false ) {
+				$plugin_path = $path;
+				break;
+			}
+		}
+		$plugin_links = apply_filters( "plugin_action_links_{$plugin_path}", array() );
```
westonruter:
```suggestion
		/** This filter is documented in wp-admin/includes/class-wp-plugins-list-table.php */
		$plugin_links = apply_filters( "plugin_action_links_{$plugin_path}", array() );
```

includes/admin/plugins.php
```diff
@@ -325,6 +324,20 @@ function perflab_render_plugin_card( array $plugin_data ): void {
 			'<button type="button" class="button button-disabled" disabled="disabled">%s</button>',
 			esc_html( _x( 'Active', 'plugin', 'default' ) )
 		);
+		$all_plugins    = array_keys( get_plugins() );
+		$plugin_slug    = $plugin_data['slug'];
+		$plugin_path    = '';
+		foreach ( $all_plugins as $path ) {
+			if ( strpos( $path, $plugin_slug ) !== false ) {
```
westonruter: Safer to do this:

```suggestion
			if ( strtok( $file, '/' ) === $plugin_slug ) {
```

This is because the multiple plugins may use the slug as a prefix, like `woocommerce` as well as `woocommerce-gateway-stripe`.

Instead of `strtok()` this could also use `dirname()`.

includes/admin/plugins.php
```diff
@@ -325,6 +324,20 @@ function perflab_render_plugin_card( array $plugin_data ): void {
 			'<button type="button" class="button button-disabled" disabled="disabled">%s</button>',
 			esc_html( _x( 'Active', 'plugin', 'default' ) )
 		);
+		$all_plugins    = array_keys( get_plugins() );
+		$plugin_slug    = $plugin_data['slug'];
+		$plugin_path    = '';
```
westonruter: Let's call this variable `$plugin_file` for consistency with how WordPress calls it.

includes/admin/plugins.php
```diff
@@ -325,6 +324,20 @@ function perflab_render_plugin_card( array $plugin_data ): void {
...
+		foreach ( $all_plugins as $path ) {
```
westonruter: Might as well just do this instead of adding the `$all_plugins` variable.

```suggestion
		foreach ( array_keys( get_plugins() ) as $file ) {
```

includes/admin/plugins.php
westonruter: Shouldn't this be wrapped in a condition so that it only applies the filters if `$plugin_path` (er, `$plugin_file`) is found?

includes/admin/plugins.php
```diff
@@ -325,6 +324,20 @@ function perflab_render_plugin_card( array $plugin_data ): void {
 			'<button type="button" class="button button-disabled" disabled="disabled">%s</button>',
 			esc_html( _x( 'Active', 'plugin', 'default' ) )
 		);
+		$all_plugins    = array_keys( get_plugins() );
+		$plugin_slug    = $plugin_data['slug'];
```
westonruter: Maybe just reuse `$plugin_data['slug']` instead of introducing a new variable?

westonruter (changes requested):
Could the Settings link be moved after the Learn More link? Since not all plugins have Settings links, this will allow the Learn More link to be consistently after the button.

westonruter:
Something else to ponder: upon activating a plugin which has Settings, it would be a nice touch to do some yellow-fade or some other styling effect to highlight the fact that there is a Settings link now after the page reloaded which wasn't there before. Not sure what would be the most in-line with existing core designs.

ashwinparthasarathi:
> Could the Settings link be moved after the Learn More link? Since not all plugins have Settings links, this will allow the Learn More link to be consistently after the button.

Understood, made changes to move the link down as the last in the action items.

> Something else to ponder: upon activating a plugin which has Settings, it would be a nice touch to do some yellow-fade or some other styling effect to highlight the fact that there is a Settings link now after the page reloaded which wasn't there before. Not sure what would be the most in-line with existing core designs.

If you could point me to a reference, I can figure out an appropriate design cue to highlight the settings link freshly added.

westonruter:
Screencast of activation currently:

[Screen recording 2024-05-13 12.00.54.webm](https://github.com/WordPress/performance/assets/134745/cf47137d-cae1-4ede-9022-6dcbc7a400bd)

Looks good!

> > Something else to ponder: upon activating a plugin which has Settings, it would be a nice touch to do some yellow-fade or some other styling effect to highlight the fact that there is a Settings link now after the page reloaded which wasn't there before. Not sure what would be the most in-line with existing core designs.
> 
> If you could point me to a reference, I can figure out an appropriate design cue to highlight the settings link freshly added.

I don't have anything to reference. But actually, I have another idea which may be better. There is that "Feature activated" admin notice which is displayed when the plugin is activated. What about adding the settings link there? When there is a settings link available, the notice could say, where settings is a link:

> Feature activated. Review <u>settings</u>.

To implement this, here is where the admin notice is output:

https://github.com/WordPress/performance/blob/eb5c8a38ca79a965006803cd84b2a049e224330b/includes/admin/load.php#L332-L340

Notice that at this point it doesn't know which plugin was activated. It just has `activate=true` as the query parameter. But this can be easily changed in `perflab_install_activate_plugin_callback()`. Namely, see here:

https://github.com/WordPress/performance/blob/eb5c8a38ca79a965006803cd84b2a049e224330b/includes/admin/load.php#L262-L268

It can add the `$plugin_slug` to that array. In fact, instead of setting the value of `activate` to `true` it can instead set the value to `$plugin_slug`. Then in `perflab_plugin_admin_notices()` it can get `$_GET['activate']` to see if it is among `$installed_plugin_slugs`, and if so, find the settings link and add it to the admin notice.

westonruter:
@ashwinparthasarathi Will you be able to get to this today or tomorrow? We're about to the deadline for 3.1.0.

ashwinparthasarathi:
Yes, today will update it.

ashwinparthasarathi:
@westonruter 

1. The code has been updated to include the Settings link in the successful plugin activation admin notice only when it gets activated from the Features screen.
Here's how it looks now,
image

2. Also since the plugin link is being generated at 2 places, have created a new function to handle the settings link generation and updated it in the plugin card render function as well.

3. For consistency, I believe this should also be updated by the individual plugin when activated from the WP plugins page.
Here's how it looks now,
image
Do let me know how to tackle this part.

includes/admin/load.php
```diff
@@ -375,3 +377,33 @@ function addPluginProgressIndicator( message ) {
 		array( 'type' => 'module' )
 	);
 }
+
+/**
+ * Callback function that returns plugin settings link.
+ *
+ * @param string $plugin_slug Plugin slug passed to generate the settings link.
+ * @return string Either the plugin settings link or empty.
```
```suggestion
 * @return string|null Either the plugin settings link or null if not available.
```

includes/admin/load.php
```diff
@@ -375,3 +377,33 @@ function addPluginProgressIndicator( message ) {
...
+ * @since n.e.x.t
+ */
+function perflab_get_plugin_settings_link( string $plugin_slug ): string {
```
```suggestion
function perflab_get_plugin_settings_link( string $plugin_slug ): ?string {
```

includes/admin/load.php
```diff
@@ -375,3 +377,33 @@ function addPluginProgressIndicator( message ) {
...
+	if ( ! empty( $plugin_file ) ) {
+		/** This filter is documented in wp-admin/includes/class-wp-plugins-list-table.php */
+		$plugin_links = apply_filters( "plugin_action_links_{$plugin_file}", array() );
+	}
+
+	if ( array_key_exists( 'settings', $plugin_links ) ) {
+		return $plugin_links['settings'];
+	}
+
+	return '';
```
```suggestion
	return null;
```

includes/admin/load.php
```diff
@@ -375,3 +377,33 @@ function addPluginProgressIndicator( message ) {
...
+			break;
+		}
+	}
+
+	if ( ! empty( $plugin_file ) ) {
+		/** This filter is documented in wp-admin/includes/class-wp-plugins-list-table.php */
+		$plugin_links = apply_filters( "plugin_action_links_{$plugin_file}", array() );
+	}
+
+	if ( array_key_exists( 'settings', $plugin_links ) ) {
```
```suggestion
	if ( is_array( $plugin_links ) && array_key_exists( 'settings', $plugin_links ) ) {
```

includes/admin/load.php
```diff
@@ -375,3 +377,33 @@ function addPluginProgressIndicator( message ) {
...
+		if ( strtok( $file, '/' ) === $plugin_slug ) {
+			$plugin_file = $file;
+			break;
+		}
+	}
+
+	if ( ! empty( $plugin_file ) ) {
+		/** This filter is documented in wp-admin/includes/class-wp-plugins-list-table.php */
+		$plugin_links = apply_filters( "plugin_action_links_{$plugin_file}", array() );
+	}
```
```suggestion
	if ( null === $plugin_file ) {
		return null;
	}

	/** This filter is documented in wp-admin/includes/class-wp-plugins-list-table.php */
	$plugin_links = apply_filters( "plugin_action_links_{$plugin_file}", array() );
```

includes/admin/load.php
```diff
@@ -375,3 +377,33 @@ function addPluginProgressIndicator( message ) {
...
+/**
+ * Callback function that returns plugin settings link.
+ *
+ * @param string $plugin_slug Plugin slug passed to generate the settings link.
+ * @return string Either the plugin settings link or empty.
+ * @since n.e.x.t
+ */
+function perflab_get_plugin_settings_link( string $plugin_slug ): string {
+	$plugin_file  = '';
+	$plugin_links = '';
```
```suggestion
	$plugin_file  = null;
```

includes/admin/load.php
```diff
@@ -330,8 +330,10 @@ static function ( $name ) {
 	}
 
 	if ( isset( $_GET['activate'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$plugin_settings_link = perflab_get_plugin_settings_link( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 		wp_admin_notice(
-			esc_html__( 'Feature activated.', 'performance-lab' ),
+			/* translators: %s: Settings link of the plugin */
+			sprintf( esc_html__( 'Feature activated. Review %s.', 'performance-lab' ), $plugin_settings_link ),
```
```suggestion
			wp_kses( sprintf( __( 'Feature activated. Review %s.', 'performance-lab' ), $plugin_settings_link ), array( 'a' => array( 'href' => array() ) ) ),
```

includes/admin/plugins.php
```diff
@@ -398,6 +400,13 @@ function perflab_render_plugin_card( array $plugin_data ): void {
 			esc_html__( 'Visit plugin site', 'default' )
 		);
 	}
+
+	if ( $show_settings_link ) {
+		$plugin_link = perflab_get_plugin_settings_link( $plugin_data['slug'] );
+		if ( ! empty( $plugin_link ) ) {
+			$action_links[] = $plugin_link;
+		}
+	}
```
westonruter: Is the `$show_settings_link` variable needed? Couldn't this code be put right up above where `$show_settings_link = true;` is currently?

westonruter (changes requested):
Overall looks good! Just some minor feedback.

westonruter:
> 3\. For consistency, I believe this should also be updated by the individual plugin when activated from the WP plugins page.

That would be nice. In `wp-admin/plugins.php` there is this code:

```php
} elseif ( isset( $_GET['activate'] ) ) {
	wp_admin_notice( __( 'Plugin activated.' ), $updated_notice_args );
```

That is where the message is coming from. So the question is how do we extend it. In this context, `$_GET['activate']` is again just `true`. So it doesn't have the information we need for which plugin specifically was activated. 

When first activating a single plugin, the initial URL the user is directed to is like:

```
/wp-admin/plugins.php?action=activate&plugin=speculation-rules%2Fload.php&plugin_status=all&paged=1&s&_wpnonce=4f6ef90cfa
```

And then if successful, there is a redirect to:

```
/wp-admin/plugins.php?activate=true&plugin_status=all&paged=1&s=
```

The redirect code for this is:

```php
			} else {
				// Overrides the ?error=true one above.
				wp_redirect( self_admin_url( "plugins.php?activate=true&plugin_status=$status&paged=$page&s=$s" ) );
			}
```

So it seems like what could be done is to add a filter to override `activate=true` to instead have the value of the plugin slug. This code would do it:

```php
add_filter( 'wp_redirect', static function ( $url ) {
	global $pagenow;
	if ( ! is_string( $url ) ) {
		return $url;
	}

	if ( ! (
		isset( $pagenow )
		&& 'plugins.php' === $pagenow
		&& isset( $_GET['action'], $_GET['plugin'] )
		&& 'activate' === $_GET['action']
	) ) {
		return $url;
	}

	$parsed_redirect_url = wp_parse_url( $url );
	if ( ! (
		is_array( $parsed_redirect_url )
		&&
		isset( $parsed_redirect_url['path'], $parsed_redirect_url['query'] )
		&&
		basename( $parsed_redirect_url['path'] ) === 'plugins.php'
		&&
		preg_match( '/(^|&)activate=true(&|$)/', $parsed_redirect_url['query'] )
	) ) {
		return $url;
	}

	$plugin_slug = dirname( wp_unslash( $_GET['plugin'] ) );

	// TODO: Validate that $plugin_slug is among the slugs for Performance Lab.

	return add_query_arg( 'activate', $plugin_slug, $url );
} );
```

But then the question is how to override the message it gets even messier:

```php
add_filter(
	'wp_admin_notice_markup',
	static function ( $markup, string $message ) {
		global $pagenow;
		if (
			! is_string( $markup )
			|| __( 'Plugin activated.', 'default' ) !== strip_tags( $message )
			|| ! isset( $pagenow )
			|| 'plugins.php' !== $pagenow
			|| ! isset( $_GET['activate'] )
		) {
			return $markup;
		}

		// TODO: Validate that this is one of our plugin slugs.
		$plugin_slug = wp_unslash( $_GET['activate'] );

		// TODO: Obtain the actual settings link for $plugin_slug.
		$settings_link = '<a href="#settings">Settings</a>';

		return str_replace(
			'</p>',
			' ' . wp_kses( sprintf( __( 'Review %s', 'performance-lab' ), $settings_link ), array( 'a' => array( 'href' => array() ) ) ) . '</p>',
			$markup
		);
	},
	10,
	2
);
```

Maybe there's a more elegant way to do this.

But I don't think this should be done as part of this PR.

includes/admin/load.php
westonruter: Oh wait, there's a more _important_ point here. Many of the plugins don't have settings links. So this needs to be made conditional. Also, ideally "Settings" would rather be "settings". So for example:

```php
$message = __( 'Feature activated.', 'performance-lab' );

$plugin_settings_link = perflab_get_plugin_settings_link( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
if ( $plugin_settings_link ) {
	$p = new WP_HTML_Tag_Processor( $plugin_settings_link );
	if ( $p->next_tag( array( 'tag_name' => 'A' ) ) && $p->get_attribute( 'href' ) ) {
		/* translators: %s is the settings URL */
		$message .= ' ' . sprintf( __( 'Review <a href="%s">settings</a>.', 'performance-lab' ), esc_url( $p->get_attribute( 'href' ) ) )
	}
}
wp_admin_notice(
	wp_kses( $message, array( 'a' => array( 'href' => array() ) ),
	array(
		'type'        => 'success',
		'dismissible' => true,
	)
);
```

includes/admin/load.php
```diff
@@ -375,3 +377,33 @@ function addPluginProgressIndicator( message ) {
 		array( 'type' => 'module' )
 	);
 }
+
+/**
+ * Callback function that returns plugin settings link.
+ *
+ * @param string $plugin_slug Plugin slug passed to generate the settings link.
+ * @return string Either the plugin settings link or empty.
+ * @since n.e.x.t
```
mukeshpanchal27:
```suggestion
 * @since n.e.x.t
 *
 * @param string $plugin_slug Plugin slug passed to generate the settings link.
 * @return string Either the plugin settings link or empty.
```

mukeshpanchal27 (changes requested):
@ashwinparthasarathi @westonruter, I believe we should postpone this to the next release. In issue #1032, we're investigating the activation of all active plugins (whether stable or experimental). Consequently, we would need to refactor this code significantly. Do these code changes support the activation of multiple plugins?

includes/admin/plugins.php
ashwinparthasarathi: @westonruter yes the variable is needed as the suggestion to move the Settings links to the last in the list made the "Learn more" link consistent over all the plugin cards.

I set a variable in the if condition checking if the plugin is active, and at the end of the else condition, added the Settings link into the action links array.

westonruter:
> In issue #1032, we're investigating the activation of all active plugins (whether stable or experimental). Consequently, we would need to refactor this code significantly. Do these code changes support the activation of multiple plugins?

@mukeshpanchal27 No, it doesn't support multiple plugin activation, since this screen doesn't support that yet. But I don't think it would be so relevant as then all of the settings links would appear together all at once. I think we can land this in 3.1.0 and then refactor as part of the bulk-activation and also Ajax activation.

includes/admin/load.php
```diff
@@ -330,8 +330,27 @@ static function ( $name ) {
...
 	if ( isset( $_GET['activate'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$message = __( 'Feature activated.', 'performance-lab' );
+
+		$plugin_settings_link = perflab_get_plugin_settings_link( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+
+		if ( $plugin_settings_link ) {
+			$p = new WP_HTML_Tag_Processor( $plugin_settings_link );
+			if ( ! empty( $p->next_tag( array( 'tag_name' => 'A' ) ) ) && ! empty( $p->get_attribute( 'href' ) ) && ! is_bool( $p->get_attribute( 'href' ) ) ) {
+				/* translators: %s is the settings URL */
+				$message .= ' ' . sprintf( __( 'Review <a href="%s">settings</a>.', 'performance-lab' ), esc_url( $p->get_attribute( 'href' ) ) );
```
westonruter: I think this this would be a bit simpler for the necessary type checks:

```suggestion
			if ( $p->next_tag( array( 'tag_name' => 'A' ) ) && $p->get_attribute( 'href' ) ) {
				/* translators: %s is the settings URL */
				$message .= ' ' . sprintf( __( 'Review <a href="%s">settings</a>.', 'performance-lab' ), esc_url( (string) $p->get_attribute( 'href' ) ) );
```

includes/admin/plugins.php
```diff
@@ -398,6 +400,13 @@ function perflab_render_plugin_card( array $plugin_data ): void {
 			esc_html__( 'Visit plugin site', 'default' )
 		);
 	}
+
+	if ( $show_settings_link ) {
+		$plugin_link = perflab_get_plugin_settings_link( $plugin_data['slug'] );
+		if ( ! empty( $plugin_link ) ) {
```
westonruter: Let's avoid using `empty()`, for example:

```suggestion
		if ( $plugin_link ) {
```

includes/admin/load.php
```diff
@@ -375,3 +394,35 @@ function addPluginProgressIndicator( message ) {
...
+
+/**
+ * Callback function that returns plugin settings link.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $plugin_slug Plugin slug passed to generate the settings link.
+ * @return string|null Either the plugin settings link or null if not available.
+ */
+function perflab_get_plugin_settings_link( string $plugin_slug ): ?string {
```
westonruter: Ah, let's move the WP_HTML_Processor code from above down into here. Then this function can just return the settings _URL_. That will simplify the code in `perflab_plugin_admin_notices()` and move that complexity into this function.

includes/admin/load.php
```diff
@@ -375,3 +394,35 @@ function addPluginProgressIndicator( message ) {
...
+	if ( null === $plugin_file ) {
+		return null;
+	}
+
+	/** This filter is documented in wp-admin/includes/class-wp-plugins-list-table.php */
+	$plugin_links = apply_filters( "plugin_action_links_{$plugin_file}", array() );
+
+	if ( is_array( $plugin_links ) && array_key_exists( 'settings', $plugin_links ) ) {
+		return $plugin_links['settings'];
+	}
```
westonruter: Per https://github.com/WordPress/performance/pull/1208/files#r1603609835:

```suggestion
	if ( ! is_array( $plugin_links ) || ! array_key_exists( 'settings', $plugin_links ) ) {
		return null;
	}
	
	$p = new WP_HTML_Tag_Processor( $plugin_links['settings'] );
	if ( ! $p->next_tag( array( 'tag_name' => 'A' ) ) ) {
		return null;
	}
	$href = $p->get_attribute( 'href' );
	if ( $href && is_string( $href ) ) {
		return $href;
	}
```

includes/admin/load.php
```diff
@@ -330,8 +330,27 @@ static function ( $name ) {
...
+		$message = __( 'Feature activated.', 'performance-lab' );
+
+		$plugin_settings_link = perflab_get_plugin_settings_link( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+
+		if ( $plugin_settings_link ) {
+			$p = new WP_HTML_Tag_Processor( $plugin_settings_link );
+			if ( ! empty( $p->next_tag( array( 'tag_name' => 'A' ) ) ) && ! empty( $p->get_attribute( 'href' ) ) && ! is_bool( $p->get_attribute( 'href' ) ) ) {
+				/* translators: %s is the settings URL */
+				$message .= ' ' . sprintf( __( 'Review <a href="%s">settings</a>.', 'performance-lab' ), esc_url( $p->get_attribute( 'href' ) ) );
+			}
```
westonruter: Per below, this could then just be:

```suggestion
			/* translators: %s is the settings URL */
			$message .= ' ' . sprintf( __( 'Review <a href="%s">settings</a>.', 'performance-lab' ), esc_url( $plugin_settings_link ) );
```

includes/admin/plugins.php
```diff
@@ -398,6 +400,13 @@ function perflab_render_plugin_card( array $plugin_data ): void {
 			esc_html__( 'Visit plugin site', 'default' )
 		);
 	}
+
+	if ( $show_settings_link ) {
```
westonruter: Alternatively, what about getting rid of the `$show_settings_link` and just reuse `$availability['activated']`?

```suggestion
	if ( $availability['activated'] ) {
```

includes/admin/plugins.php
```diff
@@ -318,18 +318,20 @@ function perflab_render_plugin_card( array $plugin_data ): void {
...
-	$availability = perflab_get_plugin_availability( $plugin_data );
-
+	$availability   = perflab_get_plugin_availability( $plugin_data );
 	$compatible_php = $availability['compatible_php'];
 	$compatible_wp  = $availability['compatible_wp'];
 
 	$action_links = array();
 
+	$show_settings_link = false;
+
```
westonruter:
```suggestion
```

includes/admin/plugins.php
```diff
@@ -318,18 +318,20 @@ function perflab_render_plugin_card( array $plugin_data ): void {
...
 	if ( $availability['activated'] ) {
-		$action_links[] = sprintf(
+		$action_links[]     = sprintf(
 			'<button type="button" class="button button-disabled" disabled="disabled">%s</button>',
 			esc_html( _x( 'Active', 'plugin', 'default' ) )
 		);
+		$show_settings_link = true;
```
westonruter:
```suggestion
```

includes/admin/load.php
```diff
@@ -375,3 +391,44 @@ function addPluginProgressIndicator( message ) {
 		array( 'type' => 'module' )
 	);
 }
+
+/**
+ * Callback function that returns plugin settings link.
```
```suggestion
 * Gets the URL to the plugin settings screen if one exists.
```

includes/admin/load.php
```diff
@@ -330,8 +330,24 @@ static function ( $name ) {
 	}
 
 	if ( isset( $_GET['activate'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$message = __( 'Feature activated.', 'performance-lab' );
+
+		$plugin_settings_link = perflab_get_plugin_settings_link( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+
+		if ( $plugin_settings_link ) {
+			/* translators: %s is the settings URL */
+			$message .= ' ' . sprintf( __( 'Review <a href="%s">settings</a>.', 'performance-lab' ), esc_url( $plugin_settings_link ) );
```
westonruter: Minor point: Let's rename the function to specify that it is returning the URL:

```suggestion
		$plugin_settings_url = perflab_get_plugin_settings_url( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
		if ( $plugin_settings_url ) {
			/* translators: %s is the settings URL */
			$message .= ' ' . sprintf( __( 'Review <a href="%s">settings</a>.', 'performance-lab' ), esc_url( $plugin_settings_url ) );
```

includes/admin/load.php
```diff
@@ -375,3 +391,44 @@ function addPluginProgressIndicator( message ) {
...
+
+/**
+ * Callback function that returns plugin settings link.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $plugin_slug Plugin slug passed to generate the settings link.
+ * @return string|null Either the plugin settings link or null if not available.
+ */
+function perflab_get_plugin_settings_link( string $plugin_slug ): ?string {
```
```suggestion
 * @return string|null Either the plugin settings URL or null if not available.
 */
function perflab_get_plugin_settings_url( string $plugin_slug ): ?string {
```

includes/admin/plugins.php
```diff
@@ -398,6 +397,14 @@ function perflab_render_plugin_card( array $plugin_data ): void {
 			esc_html__( 'Visit plugin site', 'default' )
 		);
 	}
+
+	if ( $availability['activated'] ) {
+		$plugin_link = perflab_get_plugin_settings_link( $plugin_data['slug'] );
```
westonruter:
```suggestion
		$plugin_link = perflab_get_plugin_settings_url( $plugin_data['slug'] );
```

includes/admin/plugins.php
```diff
@@ -398,6 +397,14 @@ function perflab_render_plugin_card( array $plugin_data ): void {
...
+		if ( $plugin_link ) {
+			/* translators: %s is the settings URL */
+			$action_links[] = sprintf( __( '<a href="%s">Settings</a>', 'performance-lab' ), esc_url( $plugin_link ) );
```
westonruter: This isn't currently getting escaped. This should do it:

```suggestion
			$action_links[] = sprintf( '<a href="%s">%s</a>', esc_url( $plugin_link ), esc_html__( 'Settings', 'performance-lab'  ) );
```

includes/admin/load.php
```diff
@@ -330,8 +330,24 @@ static function ( $name ) {
 	}
 
 	if ( isset( $_GET['activate'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$message = __( 'Feature activated.', 'performance-lab' );
+
+		$plugin_settings_link = perflab_get_plugin_settings_link( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
westonruter: Technically since all input vars are slashed in WP, this should get unslashed (even though it is highly unlikely to ever have slashes):

```suggestion
		$plugin_settings_link = perflab_get_plugin_settings_link( wp_unslash( $_GET['activate'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```

includes/admin/plugins.php
westonruter: Oh, and then `$plugin_link` should rather be `$settings_url`, right?

westonruter:
Just noticing this styling issue, where the Settings link is overflowing the card container:

image

westonruter:
Fixed styling issue in c7ba02c:

[Screen recording 2024-05-16 10.59.14.webm](https://github.com/WordPress/performance/assets/134745/74f781b0-e221-437d-9395-d80f0a577f61)

~I also noticed a separate issue where the plugin cards fail to render properly on mobile, but I'll address that in a follow-up PR.~ Actually, I went ahead and did it here: 57e0f8e

westonruter (approved):
Great! This will really improve discoverability of the features' settings.

westonruter:
With 57e0f8e, compare the responsive layout of the feature cards:

# Before

[Screen recording 2024-05-16 11.45.09.webm](https://github.com/WordPress/performance/assets/134745/a62e6819-d2ed-47f7-bbd0-a41c9865b37d)

# After

[Screen recording 2024-05-16 11.45.57.webm](https://github.com/WordPress/performance/assets/134745/f6335cf2-ce2d-401d-b835-7290673bb1f2)

Take note of the addition of the Settings link here, including how a divider appears when the links are presented horizontally.

joemcgill (approved):
This looks good to me. Nice job.

PR #1213, opened 2024-05-13
Update PHPStan to 1.11.1

### LINKED ISSUE #1209: PHPStan errors being ignored should specify the error identifiers
This is part of https://github.com/WordPress/performance/issues/775.

## Feature Description

A new feature coming in PHPStan 1.11 is the ability to ignore specific errors rather than all errors on a given line:

> [image](https://phpstan.org/user-guide/ignoring-errors#:~:text=If%20you%20want%20to%20ignore%20only%20a%20specific%20error%2C%20you%20can%20take%20advantage%20of%20error%20identifiers%20and%20use%20the%20new%20%40phpstan%2Dignore%20tag%20that%20comes%20with%20two%20features%3A)

We should implement this for the few places we have ignores in place:

https://github.com/WordPress/performance/blob/8b742ee64318dcd63fac955a419e6d22357116fd/tests/includes/server-timing/perflab-server-timing-tests.php#L114

https://github.com/WordPress/performance/blob/8b742ee64318dcd63fac955a419e6d22357116fd/tests/includes/server-timing/perflab-server-timing-tests.php#L126

https://github.com/WordPress/performance/blob/8b742ee64318dcd63fac955a419e6d22357116fd/tests/plugins/webp-uploads/helper-tests.php#L382

https://github.com/WordPress/performance/blob/8b742ee64318dcd63fac955a419e6d22357116fd/tests/plugins/webp-uploads/image-edit-tests.php#L402

This will ensure we're only ignoring the error we expect. This is [coming out next week](https://phpc.social/@OndrejMirtes/112393186034803177)!

westonruter:
Fixes #1209 
See https://github.com/WordPress/performance/issues/775

* Leverage the new PHPStan error identifiers to ignore errors.
* Remove the need to ignore some errors.

### DIFF
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -277,7 +277,6 @@ add_action( 'admin_action_perflab_install_activate_plugin', 'perflab_install_act
  * Callback function to handle admin inline style.
  *
  * @since 3.0.0
- * @phpstan-ignore void.pure (PHPStan isn't apparently aware that this function is printing anything. See https://github.com/phpstan/phpstan/issues/11008)
  */
 function perflab_print_features_page_style(): void {
 	?>
--- a/includes/admin/load.php
+++ b/includes/admin/load.php
@@ -277,6 +277,7 @@ add_action( 'admin_action_perflab_install_activate_plugin', 'perflab_install_act
  * Callback function to handle admin inline style.
  *
  * @since 3.0.0
+ * @phpstan-ignore void.pure (PHPStan isn't apparently aware that this function is printing anything. See https://github.com/phpstan/phpstan/issues/11008)
  */
 function perflab_print_features_page_style(): void {
 	?>
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -26,38 +26,35 @@ parameters:
 	ignoreErrors:
 		-
 			# Ignore error related to PHP 8 using a GdImage but PHP 7 using a resource.
-			message: '/^Parameter .+? of function .+? expects resource, GdImage\|resource given\.$/'
+			identifier: argument.type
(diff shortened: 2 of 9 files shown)


westonruter:
@thelovekesh Sorry, there is _one_ more PR for PHPStan.

phpstan.neon.dist
```diff
@@ -30,34 +30,30 @@ parameters:
 			path: plugins/dominant-color-images/*.php
 
 		# The following ignored errors are all for tests. They are likely excessively strict and do not warrant being fixed.
-		# TODO: Use error identifiers instead of messages once PHPStan 1.11 is out: <https://github.com/WordPress/performance/issues/1209>.
 		-
-			message: '/Parameter #\d.+given./'
+			identifier: argument.type
```
westonruter, line 34: Note: I tried originally to use `identifier: argument.type` with the ignored error above in `dominant-color-images`. However, it seems you can only use an `identifier` once in the `ignoreErrors`, or else the ignore is ignored (such _ignorance_!) But it is better anyway to use the `message` above since it is explicitly about `GdImage|resource` being used.

includes/admin/load.php
```diff
@@ -277,6 +277,7 @@ function perflab_install_activate_plugin_callback(): void {
  * Callback function to handle admin inline style.
  *
  * @since 3.0.0
+ * @phpstan-ignore void.pure (PHPStan isn't apparently aware that this function is printing anything. See https://github.com/phpstan/phpstan/issues/11008)
```
swissspidy: PHPStan 1.11.1 was just released, fixing this.

swissspidy (approved):
Let's update to 1.11.1 and then it should be good to go

PR #1215, opened 2024-05-14
Fix Imagick detecting partial transparency

### LINKED ISSUE #1214: Images with partial transpency return "false" for "has-transparency"
## Bug Description

Images that have partial transparency return `false` for `has-transparency`

## Steps to reproduce

1. Upload a partially transparent image, such as [this one](https://github.com/WordPress/performance/assets/1123997/be3cea2a-f917-4f7c-b84b-0240b645dd2b)
2. View the source code on the front end
3. Observer `data-transparency="false"` is added to the image

## Screenshots

image

image

## Additional Context

 - PHP Version: 8.2.18
 - OS: All
 - Browser: All
 - Plugin Version: Performance Lab 3.0.0, Image Placeholders v1.1.0
 - Device: All

westonruter:
Fixes #1214

By default [`ImagickPixel::getColor()`](https://www.php.net/manual/en/imagickpixel.getcolor.php) returns an array with the 4th alpha array item being a binary `0` or `1`. This doesn't work when the image has partial transparency. To detect that, you have to pass a `$normalized` parameter to this method, either `1` (to return a 0.0-1.0 float) or `2` (to return a 0-1255 int). So this passes 2 so we can correctly compare if the alpha value is greater than 0.

This PR also improves the test cases by removing duplicate tests and ensuring that the GD and Imagick code paths are both tested. Previously it was doing the same tests twice for GD since the tests weren't forcing either the GD or Imagick code path. As part of this, the Imagick tests now show up as being skipped since wp-env appears to have a misconfigured Imagick as [detailed](https://github.com/WordPress/performance/pull/1215#issuecomment-2110857517) by @thelovekesh.


plugins/dominant-color-images/class-dominant-color-image-editor-imagick.php
```diff
@@ -82,7 +82,7 @@ public function has_transparency() {
 			for ( $x = 0; $x < $w; $x++ ) {
 				for ( $y = 0; $y < $h; $y++ ) {
 					$pixel = $this->image->getImagePixelColor( $x, $y );
-					$color = $pixel->getColor();
+					$color = $pixel->getColor( 2 );
```
westonruter, line 85: This is the crux of the fix.

westonruter:
I discovered that the tests were not testing both of the Imagick and GD codepaths. It was rather just testing the GD codepath twice, since it was letting WP select the implementation that was available in both the GD and Imagick test cases.

As part of this, I found that Imagick seems to not be configured properly in wp-env because `wp_image_editor_supports()` is returning `false` for every image when testing Imagick. @thelovekesh was this what you discovered as well?

thelovekesh:
Yes, @westonruter. I have explained it in - https://github.com/WordPress/performance/pull/1013#issuecomment-1962589426 and https://github.com/WordPress/performance/pull/1013#issuecomment-1962731568

plugins/dominant-color-images/class-dominant-color-image-editor-imagick.php
westonruter: Because of https://github.com/WordPress/performance/pull/1215#issuecomment-2110853741, however, this is not currently getting tested in PHPUnit in wp-env.

composer.json
```diff
@@ -15,6 +15,10 @@
     "php": "^7.2 || ^8.0",
     "ext-json": "*"
   },
+  "suggest": {
+    "ext-imagick": "*",
+    "ext-gd": "*"
```
thelovekesh: I think we should consider adding a reason for why these are suggested.

thelovekesh (approved):

composer.json
westonruter: Is there a way to provide a reason?

westonruter: I added them because PhpStorm was complaining about them not being in `composer.json`. Granted, it still complains, although less loudly, that they are just suggested and not required.

thelovekesh: As per docs, yes - https://getcomposer.org/doc/04-schema.md#suggest

westonruter: Oh, interesting:

image

I guess you can't suggest specific versions of things then?

westonruter: How's this?

```suggestion
    "ext-imagick": "Required to use Modern Image Format's Dominant_Color_Image_Editor_Imagick class",
    "ext-gd": "Required to use Modern Image Format's Dominant_Color_Image_Editor_GD class"
```

thelovekesh: Much better. I think we can tweak it to include the version info in the text itself. Something like:
```
"ext-imagick": "Required at least {version constraint} to use Modern Image Format's Dominant_Color_Image_Editor_Imagick class",
```

westonruter: Although here we don't need to specify the version anyway.

westonruter: What I meant is that the freeform nature of `suggest` means that composer wouldn't be able to install the suggested dependencies automatically.

PR #1223, opened 2024-05-17
Automatically detect version when running since command and allow updating all plugins at once

westonruter:
The current instructions instructions for the Performance Lab plugin instruct:

> `npm run since -- -r $VERSION_NUMBER`

And for standalone plugins instruct:

> Make sure that any n.e.x.t version references have been updated via `npm run since -- --plugin $plugin_slug --release $plugin_version`.

This quickly gets tedious because you have to look up whatever the plugin version is and use it with each command. However, this is entirely unnecessary because the release instructions already instruct to update the `Stable tag` for each plugin first. Therefore, the `since` command can simply reuse the version in the `Stable tag` to use in the `n.e.x.t` version replacement. 

So this PR eliminates the `-r`/`--release` option, letting the version be automatically detected instead. It also removes `performance-lab` from being the default value for the `-p`/`--plugin` option, in which case all plugins get their `n.e.x.t` versions updated. For example:

image

Where the output is 142c7813f31af3660f7c044c734f9a8753c9cceb.

### DIFF
--- a/bin/plugin/commands/since.js
+++ b/bin/plugin/commands/since.js
@@ -4,6 +4,7 @@
 const fs = require( 'fs' );
 const path = require( 'path' );
 const glob = require( 'fast-glob' );
+const { log, formats } = require( '../lib/logger' );
 
 /**
  * Internal dependencies
@@ -13,19 +14,13 @@ const { plugins } = require( '../../../plugins.json' );
 /**
  * @typedef WPSinceCommandOptions
  *
- * @property {string} plugin  Plugin slug.
- * @property {string} release Release version number.
+ * @property {string=} plugin Plugin slug.
  */
 
 exports.options = [
 	{
 		argname: '-p, --plugin <plugin>',
-		description: 'Plugin slug',
-		defaults: 'performance-lab',
-	},
-	{
-		argname: '-r, --release <release>',
-		description: 'Release version number',
+		description: 'Plugin slug. Defaults to update all.',
 	},
 ];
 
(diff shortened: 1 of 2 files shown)


bin/plugin/commands/since.js
```diff
@@ -13,19 +14,13 @@ const { plugins } = require( '../../../plugins.json' );
 /**
  * @typedef WPSinceCommandOptions
  *
- * @property {string} plugin  Plugin slug.
- * @property {string} release Release version number.
+ * @property {?string} plugin Plugin slug.
```
westonruter:
```suggestion
 * @property {string=} plugin Plugin slug.
```

bin/plugin/commands/since.js
```diff
@@ -13,19 +14,13 @@ const { plugins } = require( '../../../plugins.json' );
...
 exports.options = [
 	{
 		argname: '-p, --plugin <plugin>',
-		description: 'Plugin slug',
-		defaults: 'performance-lab',
-	},
-	{
-		argname: '-r, --release <release>',
-		description: 'Release version number',
+		description: 'Plugin slug. Defaults to update all.',
```
joemcgill, line 23: 👍🏻

joemcgill (approved):
Lovely 💖 

Once this is merged, we need to update [this section](https://make.wordpress.org/performance/handbook/performance-lab/releasing-the-plugin/#update-the-version-number) of the handbook.

PR #1224, opened 2024-05-17
Improve readme command

westonruter:
_This is a sub-PR of #1223. Merge it before merging this one._

As part of the 3.1.0 release, the readme changelogs need to be updated for the plugins. For the Performance Lab plugin, the instructions are:

<blockquote>Run <code>npm run readme -- -m "performance-lab $VERSION_NUMBER"</code> to update the <code>readme.txt</code> file with the release changelog, where the quoted string is the name of the milestone. For example, if the milestone is <code>performance-lab 3.0.0</code>, the command needs to be <code>npm run readme -- -m "performance-lab 3.0.0"</code>.</blockquote>

And for standalone plugins:

<blockquote>Ensure the plugin’s <code>readme.txt</code> already contains the changelog section with all entries for the new version. You can use <code>npm run changelog -- --milestone="$plugin_slug $plugin_version"</code>.</blockquote>

This is tedious for couple reasons (similarly to #1223 for the `since` command):

1. You have to manually look up the current version for each plugin.
2. You have to constrict the milestone name by concatenating the plugin slug with the version.
3. You have to do this for each plugin separately.

This PR makes this process much easier by allowing you to just run `npm run readme` in one go to update all plugins in one go. The version is automatically pulled from the `Stable tag` in the `readme.txt`, and the list of all of the plugin slugs is pulled from `plugins.json`. If you want to just update a single plugin, you can do so with the `--plugin` argument.

Additionally, the behavior is improved in the following ways:

1. It does not get confused when there is an `Upgrade Notice` with a heading for the stable tag.
2. It merges any existing changelog content for the stable tag, putting it at the end in an **Other** group.
3. It only considers _merged_ pull requests as opposed to non-merged _closed_ pull requests which may still be in the milestone.
4. When all of the pull requests in a milestone have the `skip changelog` label, then this is now reflected separately in the output, as opposed to there being no merged pull requests at all.
5. Different success messages are shown based on whether the changelog was added as an entirely new section, or it was merged with an existing changelog section

image

Example output: 4023559d

### DIFF
--- a/bin/plugin/commands/readme.js
+++ b/bin/plugin/commands/readme.js
@@ -48,6 +48,7 @@ exports.handler = async ( opt ) => {
  *
  * @param {string} readmeFile Readme file path.
  * @param {string} changelog  Changelog in markdown.
+ * @return {string} Success status message.
  */
 function updateReadmeChangelog( readmeFile, changelog ) {
 	const fileContent = fs.readFileSync( readmeFile, 'utf-8' );
--- a/bin/plugin/commands/readme.js
+++ b/bin/plugin/commands/readme.js
@@ -10,34 +10,20 @@ const fs = require( 'fs' );
 const { log, formats } = require( '../lib/logger' );
 const config = require( '../config' );
 const { getChangelog } = require( './changelog' );
+const { plugins } = require( '../../../plugins.json' );
 
 /**
  * @typedef WPReadmeCommandOptions
  *
- * @property {string=} milestone Optional milestone title, to update the changelog in the readme.
- * @property {string=} path      Optional path to the readme.txt file to update. If omitted, it will be detected via --milestone.
- * @property {string=} token     Optional personal GitHub access token, only relevant for changelog updates.
- */
-
-/**
- * @typedef WPReadmeSettings
- *
- * @property {string}  owner     GitHub repository owner.
- * @property {string}  repo      GitHub repository name.
- * @property {string=} milestone Optional milestone title, to update the changelog in the readme.
- * @property {string=} path      Optional path to the readme.txt file to update.
- * @property {string=} token     Optional personal GitHub access token, only relevant for changelog updates.
+ * @property {string=} plugin Optional plugin slug to update one a single plugin's readme. If omitted, all plugins are updated.
+ * @property {string=} token  Optional personal GitHub access token.
  */
 
 exports.options = [
 	{
-		argname: '-m, --milestone <milestone>',
-		description: 'Milestone title, to update the changelog',
-	},
-	{
-		argname: '-p, --path <path>',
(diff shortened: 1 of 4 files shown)


bin/plugin/commands/readme.js
```diff
@@ -52,112 +38,136 @@ exports.options = [
...
+
+	// Try to merge the new changelog with the existing changelog.
+	let newContent = fileContent.replace(
+		regex,
+		( match, changelogHeading, _versionHeading, existingChangelog ) => {
+			status =
+				'Merged existing changelog with the new changelog in an Other section.';
+			return `${ changelogHeading }${ _versionHeading }${ normalizedChangelog }\n**Other**\n\n${ existingChangelog }`;
+		}
+	);
```
joemcgill, line 81: Does this assume that any existing changelog entries should be classified as "Other"? If you have an existing `**Enhancements**` section in the changelog and then run this command again, you would end up with something like this:

```md
**Enhancements**

* Avoid needless array allocation in rgb to hex conversion. ([1104](https://github.com/WordPress/performance/pull/1104))

**Bug Fixes**

* Fix Imagick detecting partial transparency. ([1215](https://github.com/WordPress/performance/pull/1215))

**Other**

**Enhancements**

* Improve overall code quality with stricter static analysis checks. ([775](https://github.com/WordPress/performance/issues/775))
```

This is probably fine, so long as people are paying attention to the output and not merging it uncritically, just something I noticed. I also noticed that since there were no _additional_ PRs to add to the changelog, the auto-sizes changelog didn't get the `**Other**` heading added above the existing `Improve overall code quality with stricter static analysis checks.`. Also, probably fine, but wonder if we should try to be consistent—even if there are no other merged PRs to add to the changelong.

joemcgill (approved):
One comment, but I think this is a great improvement. We'll need to make sure and update [this section in the handbook](https://make.wordpress.org/performance/handbook/performance-lab/releasing-the-plugin/#update-readme-txt) once this is merged.

bin/plugin/commands/readme.js
westonruter: Yeah, it will still require the user to review the output. But the important thing here is that it doesn't blow away the existing changelog entries already in the readme, so at least it is easier for the developer to review the changes and merge them.

felixarntz: @westonruter Chiming in late here, but I'm not sure I understand the purpose of this. What would "Other" be for? Shouldn't we be good with features, enhancements, and bug fixes?

westonruter: @felixarntz It's whatever was originally added to that section in the readme, as opposed to that original content being entirely overwritten. It's not intended to be left as-is, so when reviewing the changes the other items can be moved into the other sections.

PR #1232, opened 2024-05-20
Check if rel contains nofollow instead of being just nofollow when excluding speculative loading

tunetheweb:
## Summary

Fixes issue raised here: https://wordpress.org/support/topic/arelnofollow-exact-match-only/

## Relevant technical choices

Made the selector wider as [detailed here](https://stackoverflow.com/a/8714421/2144578).

There were no unit tests for this one. Maybe should add some but didn't see an easy one I could copy and paste as none of the existing ones are attribute based.

### DIFF
--- a/plugins/speculation-rules/helper.php
+++ b/plugins/speculation-rules/helper.php
@@ -103,7 +103,7 @@ function plsr_get_speculation_rules(): array {
 					// Also exclude rel=nofollow links, as plugins like WooCommerce use that on their add-to-cart links.
 					array(
 						'not' => array(
-							'selector_matches' => 'a[rel=nofollow]',
+							'selector_matches' => 'a[rel*=nofollow]',
 						),
 					),
 				),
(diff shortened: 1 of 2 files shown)


plugins/speculation-rules/helper.php
```diff
@@ -103,7 +103,7 @@ static function ( string $href_exclude_path ) use ( $prefixer ): string {
 					// Also exclude rel=nofollow links, as plugins like WooCommerce use that on their add-to-cart links.
 					array(
 						'not' => array(
-							'selector_matches' => 'a[rel=nofollow]',
+							'selector_matches' => 'a[rel*=nofollow]',
```
westonruter: I think the "attribute contains" selector is more appropriate here since we want to match the entire token. Also, might as well quote it.

```suggestion
							'selector_matches' => 'a[rel~="nofollow"]',
```

westonruter (approved):

PR #1233, opened 2024-05-20
Fix selector for blocking speculative loading when rel attribute contains more than nofollow link type token

westonruter:
As reported in a [support topic](https://wordpress.org/support/topic/arelnofollow-exact-match-only/), the CSS selector for preventing speculative loading for `rel=nofollow` links is overly constrictive. It doesn't allow for the fact that other link type tokens. So it currently fails to prevent speculative loading of `<a href="/our-sponsor/" rel="nofollow sponsored">`. To fix this, the attribute contains selector should be used instead.

### DIFF
--- a/plugins/speculation-rules/helper.php
+++ b/plugins/speculation-rules/helper.php
@@ -103,7 +103,7 @@ function plsr_get_speculation_rules(): array {
 					// Also exclude rel=nofollow links, as plugins like WooCommerce use that on their add-to-cart links.
 					array(
 						'not' => array(
-							'selector_matches' => 'a[rel=nofollow]',
+							'selector_matches' => 'a[rel~="nofollow"]',
 						),
 					),
 				),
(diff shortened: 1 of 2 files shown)


westonruter:
@tunetheweb Hah, you [beat me](https://github.com/WordPress/performance/pull/1232) by a hair

plugins/speculation-rules/helper.php
```diff
@@ -103,7 +103,7 @@ static function ( string $href_exclude_path ) use ( $prefixer ): string {
 					// Also exclude rel=nofollow links, as plugins like WooCommerce use that on their add-to-cart links.
 					array(
 						'not' => array(
-							'selector_matches' => 'a[rel=nofollow]',
+							'selector_matches' => 'a[rel~="nofollow"]',
```
tunetheweb, line 106: I went with `*-` in #1232 - any ideas what the difference is?:

```suggestion
							'selector_matches' => 'a[rel*="nofollow"]',
```

tunetheweb: Ah found it. `~=` is correct and I'm wrong:

https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors

westonruter: Yeah, `*=` would erroneously match `rel="not-nofollow"`, whereas `~=` would only match the entire token.

PR #73, opened 2021-12-31
Add `picture` element support

adamsilverstein:
Fixes https://github.com/WordPress/performance/issues/21
Closes #1236

* Filters on `wp_content_img_tag` to potentially wrap (media) images in a picture element.
* Works best with themes that support responsive images, for example `twentysixteen`.
* Check the network panel to see which images the browser actually loads.

## Picture element and AVIF support by version
AVIF support is [being worked on](https://github.com/WordPress/performance/pull/1176) and `picture` element support can be especially useful for adding AVIFs to your site if you have users on browsers that don't support AVIF.

Browser | `picture` Support (Version, Date) | AVIF Support (Version, Date)
-- | -- | --
Chrome | 4 (January 2010) | 85 (August 2020)
Edge | 12 (July 2015) | 121 (January 2024)
Safari | 9 (September 2015) | 16.1 (October 2023)
Firefox | 38 (May 2015) | 93 (October 2021)

PR #1235, opened 2024-05-21
Add Image Prioritizer plugin

### LINKED ISSUE #1088: Split out image-specific aspects of Optimization Detective to dependent plugin

westonruter:
This splits out the image-specific optimizations from Optimization Detective into a separate dependent plugin: Image Prioritizer.

Please review the commits one-by-one as they have been tailored to be as atomic as possible. Also, suppressing whitespace changes will help with reviewing methods for which caching was added.

The image optimization logic previously located in the `od_optimize_template_output_buffer()` function has been heavily refactored in the following ways:

A new `OD_Tag_Visitor_Registry` class has been introduced which allows plugins at the `od_register_tag_visitors` action to register callbacks that are invoked for each tag while iterating over the document. This callback is passed the `OD_HTML_Tag_Walker` instance, allowing the callback to inspect the current tag and attributes, as well as to obtain the XPath. With this information in hand, it is able to look up the element in the provided `OD_URL_Metrics_Group_Collection` instance to see if it is an LCP element (or in future, whether it was in the viewport, etc). It can also add preload links via the provided `OD_Preload_Link_Collection` instance. If the callback returns `true` then this indicates that the tag is relevant for the visitor. Such visitor-related tags will automatically get the `data-od-xpath` attributes added to them when the collection of URL metric groups is not complete; this ensures that the element is captured in the URL metrics.

Here's a basic example of how a visitor could be implemented with a closure:

```php
<?php
add_action(
	'od_register_tag_visitors',
	function ( OD_Tag_Visitor_Registry $tag_visitor_registry ): void {
		$tag_visitor_registry->register(
			'img-dimensions',
			function ( OD_HTML_Tag_Walker $walker ): bool {
				if ( $walker->get_tag() !== 'IMG' ) {
					return false;
				}

				$src = trim( (string) $walker->get_attribute( 'src' ) );
				if ( ! str_starts_with( $src, 'http://' ) && ! str_starts_with( $src, 'https://' ) ) {
					return false;
				}

				$width  = (int) $walker->get_attribute( 'width' );
				$height = (int) $walker->get_attribute( 'height' );
				if ( 0 !== $width && 0 !== $height ) {
					return false;
				}

				// Example code to fetch dimensions (but without transient caching 🙈).
				$client = new \FasterImage\FasterImage();
				$images = $client->batch( array( $src ) );
				list( $width, $height ) = current( $images )['size'];

				$walker->set_attribute( 'width', (string) $width );
				$walker->set_attribute( 'height', (string) $height );

				return true;
			}
		);
	}
);
```

The image optimization logic is further refactored into two classes:

* `IP_Img_Tag_Visitor` handles optimizing LCP `img` tags
* `IP_Background_Image_Styled_Tag_Visitor` handles optimizing LCP elements with a `background-image` style.

Both of these classes inherit from an `IP_Tag_Visitor` abstract class that sets up the common interface. Note that these classes implement an `__invoke()` method, allowing them to be passed to the tag visitor registry to be called the same way as a closure.

Fixes #1088


.wp-env.json
```diff
@@ -2,10 +2,11 @@
   "core": null,
   "plugins": [
     ".",
+    "./plugins/optimization-detective",
```
westonruter, line 4: This has to appear before `image-prioritizer` or else an error occurs since it is a dependency.

plugins/optimization-detective/class-od-url-metrics-group-collection.php
```diff
@@ -188,20 +210,29 @@ public function add_url_metric( OD_URL_Metric $new_url_metric ): void {
 	 * @return OD_URL_Metrics_Group URL metrics group for the viewport width.
 	 */
 	public function get_group_for_viewport_width( int $viewport_width ): OD_URL_Metrics_Group {
-		foreach ( $this->groups as $group ) {
-			if ( $group->is_viewport_width_in_range( $viewport_width ) ) {
-				return $group;
-			}
+		if ( array_key_exists( __FUNCTION__, $this->result_cache ) && array_key_exists( $viewport_width, $this->result_cache[ __FUNCTION__ ] ) ) {
+			return $this->result_cache[ __FUNCTION__ ][ $viewport_width ];
 		}
```
westonruter, line 215: I'm not entirely happy with this. I wanted to instead have a `get_cached_result()` helper method that relied on generics for PHPStan strict type checks. However, I wasn't able to get it to work. See https://phpstan.org/r/b203ec8c-2d70-469c-b24f-9e7143cff72d

plugins/optimization-detective/class-od-url-metrics-group-collection.php
```diff
@@ -217,12 +248,21 @@ public function get_group_for_viewport_width( int $viewport_width ): OD_URL_Metr
 	 * @return bool Whether all groups have some URL metrics.
 	 */
 	public function is_every_group_populated(): bool {
-		foreach ( $this->groups as $group ) {
-			if ( count( $group ) === 0 ) {
-				return false;
-			}
+		if ( array_key_exists( __FUNCTION__, $this->result_cache ) ) {
+			return $this->result_cache[ __FUNCTION__ ];
 		}
```
westonruter, line 253: I'm not entirely happy with this. There's too much boilerplate. I wanted to instead have a `get_cached_result()` helper method that relied on generics for PHPStan strict type checks. However, I wasn't able to get it to work. See https://phpstan.org/r/b203ec8c-2d70-469c-b24f-9e7143cff72d

westonruter: Granted, it would also have to be updated to support caching by the method argument as in the `get_group_for_viewport_width` example above.

westonruter:
Note the linting issues will be resolved once https://github.com/WordPress/performance/pull/73 is merged and this PR is rebased/merged back onto `trunk`.

plugins/optimization-detective/class-od-url-metrics-group-collection.php
```diff
@@ -217,12 +248,21 @@ public function get_group_for_viewport_width( int $viewport_width ): OD_URL_Metr
 	 * @return bool Whether all groups have some URL metrics.
 	 */
 	public function is_every_group_populated(): bool {
-		foreach ( $this->groups as $group ) {
-			if ( count( $group ) === 0 ) {
-				return false;
-			}
+		if ( array_key_exists( __FUNCTION__, $this->result_cache ) ) {
```
westonruter, line 251: See https://github.com/WordPress/performance/pull/1235#discussion_r1619533970

westonruter:
An area for future optimization would be to consider whether we can avoid invoking all tag visitor callbacks for every tag.

.wp-env.json
adamsilverstein: please add an inline comment explaining this so it doesn't get moved

westonruter: Unfortunately inline comments are not allowed in JSON

westonruter: We'd have to hack it with something like this:

```json
{
  "core": null,
  "// Notice": "The optimization-detective plugin must be listed first since it is a dependency.",
  "plugins": [
    "./plugins/optimization-detective",
    "./plugins/auto-sizes",
    "./plugins/dominant-color-images",
    "./plugins/embed-optimizer",
    "./plugins/image-prioritizer",
    "./plugins/performance-lab",
    "./plugins/speculation-rules",
    "./plugins/webp-uploads"
  ],
  "env": {
    "tests": {
      "config": {
        "FS_METHOD": "direct"
      },
      "mappings": {
        "/wp-content/plugins/performance": "."
      }
    }
  }
}
```

plugins/image-prioritizer/class-ip-img-tag-visitor.php
```diff
@@ -0,0 +1,99 @@
...
+			// Never include loading=lazy on the LCP image common across all breakpoints.
+			if ( 'lazy' === $walker->get_attribute( 'loading' ) ) {
+				$walker->set_attribute( 'data-od-removed-loading', $walker->get_attribute( 'loading' ) );
+				$walker->remove_attribute( 'loading' );
+			}
+		} elseif ( is_string( $walker->get_attribute( 'fetchpriority' ) ) && $this->url_metrics_group_collection->is_every_group_populated() ) {
+			// Note: The $all_breakpoints_have_url_metrics condition here allows for server-side heuristics to
+			// continue to apply while waiting for all breakpoints to have metrics collected for them.
+			$walker->set_attribute( 'data-od-removed-fetchpriority', $walker->get_attribute( 'fetchpriority' ) );
+			$walker->remove_attribute( 'fetchpriority' );
```
adamsilverstein: could we set fetchpriority to low here instead of removing?

.wp-env.json
westonruter: Actually, that doesn't work. I get an error:

```
✖ Invalid /home/westonruter/repos/performance/.wp-env.override.json: "// Note" is not a configuration option.
```

swissspidy: Yeah I don‘t think we need a clever comment here. If you try to move it, there will be an error anyway.

plugins/image-prioritizer/class-ip-img-tag-visitor.php
westonruter: I see the comment here explaining the removal is not fleshed out enough. I've elaborated in 489d1f6b. It may be that the element here is actually the LCP element for _some_ of the viewport groups. In this case, `fetchpriority=high` is removed from the `img` but it is _added_ to the preload link below which also includes the media query to ensure it is only prioritized for the right viewports.

Adding `fetchpriority=low` is something I want to explore in the context of images that are above the fold but obscured, such as in a mega menu or as subsequent slides in a carousel.

.wp-env.json
adamsilverstein: right - I didn't realize this was in JSON when suggesting a comment.

plugins/image-prioritizer/readme.txt
```diff
@@ -0,0 +1,58 @@
+=== Image Prioritizer ===
+
+Contributors:      wordpressdotorg
+Requires at least: 6.4
+Tested up to:      6.5
+Requires PHP:      7.2
+Stable tag:        0.1.0
```
adamsilverstein, line 7: I'd go with 1.0.0 - what have we done n the past for new plugins?

westonruter: Optimization Detective started with 0.1.0, as did Embed Optimizer. I'd keep it below 1.x since it's still experimental. Once something graduates from being experimental, then I think it makes sense to go 1.x.

plugins/optimization-detective/class-od-url-metrics-group-collection.php
westonruter: I've reported this as a possible bug to PHPStan: https://github.com/phpstan/phpstan/issues/11138

tests/bootstrap.php
```diff
@@ -51,21 +51,34 @@
 	$plugin_name = 'performance-lab';
 }
 
-$plugin_test_path = TESTS_PLUGIN_DIR . '/plugins/' . $plugin_name;
+/**
+ * Load plugin bootstrap and any dependencies.
+ *
+ * @param string $plugin_name Plugin slug to load.
+ */
+$load_plugin = static function ( string $plugin_name ) use ( &$load_plugin ): void {
```
swissspidy, line 59: Clever!

plugins/optimization-detective/tests/test-class-od-tag-visitor-registry.php
```diff
@@ -0,0 +1,73 @@
...
+		);
+
+		// Unregister a visitor.
+		$this->assertTrue( $registry->unregister( 'video' ) );
+		$this->assertFalse( $registry->unregister( 'video' ) );
+		$this->assertEqualSets(
+			array( $video_visitor ),
+			iterator_to_array( $registry )
+		);
+		$this->assertCount( 1, $registry );
```
swissspidy: This assertions is a bit unnecessary since the `assertEqualSets` already verifies that it's an array with 1 element.

plugins/optimization-detective/readme.txt
```diff
@@ -161,3 +159,8 @@ The [plugin source code](https://github.com/WordPress/performance/tree/trunk/plu
 = 0.1.0 =
 
 * Initial release.
+
+== Upgrade Notice ==
+
+= 0.3.0 =
+Image loading optimizations have been moved to a new dependent plugin called Image Prioritizer. The Optimization Detective plugin is now serves as a dependency.
```
swissspidy:
```suggestion
Image loading optimizations have been moved to a new dependent plugin called Image Prioritizer. The Optimization Detective plugin now serves as a dependency.
```

plugins/image-prioritizer/tests/test-helper.php
```diff
@@ -0,0 +1,900 @@
+<?php
+/**
+ * Tests for optimization-detective plugin helper.php.
+ *
+ * @package optimization-detective
```
swissspidy:
```suggestion
 * Tests for image-prioritizer plugin helper.php.
 *
 * @package image-prioritizer
```

plugins/image-prioritizer/readme.txt
```diff
@@ -0,0 +1,58 @@
...
+
+Optimizes the loading of the LCP image by leveraging client-side detection with real user metrics.
+
+== Description ==
+
+Optimizes the loading of the LCP image by leveraging client-side detection with real user metrics. Currently, this involves adding `fetchpriority=high` to the LCP image and adding preload links for the LCP `img` as well as any LCP `background-image`.
+
+This plugin requires the [Optimization Detective](https://wordpress.org/plugins/optimization-detective/) plugin as a dependency.
+
+TODO: Flesh out description.
```
swissspidy, line 20: Let's not forget this before release

plugins/image-prioritizer/helper.php
```diff
@@ -0,0 +1,41 @@
...
+}
+
+/**
+ * Displays the HTML generator meta tag for the Image Prioritizer plugin.
+ *
+ * See {@see 'wp_head'}.
+ *
+ * @since n.e.x.t
+ */
+function ip_render_generator_meta_tag(): void {
```
swissspidy: We should consider using `image_prioritzer_` as the prefix.

It's very likely that many other plugins (public or private) use `ip_` already.

The plugin review guidelines also recommend using longer prefixes (or namespaces) nowadays because conflicts are less likely.

plugins/image-prioritizer/helper.php
```diff
@@ -0,0 +1,41 @@
...
+ * Registers tag visitors for images.
+ *
+ * @since n.e.x.t
+ *
+ * @param OD_Tag_Visitor_Registry         $registry                     Tag visitor registry.
+ * @param OD_URL_Metrics_Group_Collection $url_metrics_group_collection URL Metrics Group Collection.
+ * @param OD_Preload_Link_Collection      $preload_links_collection     Preload Links Collection.
+ */
+function ip_register_tag_visitors( OD_Tag_Visitor_Registry $registry, OD_URL_Metrics_Group_Collection $url_metrics_group_collection, OD_Preload_Link_Collection $preload_links_collection ): void {
+	// Note: The class is invocable (it as an __invoke() method).
```
swissspidy:
```suggestion
	// Note: The class is invocable (it has an __invoke() method).
```

plugins/image-prioritizer/load.php
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+(
+	/**
+	 * Register this copy of the plugin among other potential copies embedded in plugins or themes.
```
swissspidy, line 26: I thought this whole logic was only relevant for allowing Optimization Detective to be embedded. Are we assuming plugins will embed this plugin as well?

plugins/optimization-detective/class-od-url-metrics-group-collection.php
swissspidy: Hmm I thought I had a similar issue recently but now I can't find the code anymore 😅  Curious to hear the response on that issue 👍

swissspidy (approved):
Nice work! 🚀 

Just a few minor thoughts

plugins/image-prioritizer/readme.txt
westonruter: Indeed! I've fleshed it out in the sub-PR: https://github.com/WordPress/performance/pull/1261

plugins/image-prioritizer/load.php
westonruter: Yes, if a plugin wants to embed image optimization logic then they'd have to embed both Optimization Detective and Image Prioritizer.

plugins/image-prioritizer/helper.php
westonruter: Oh, namespaces? 😋

westonruter: Used longer prefix in d438024a.

Using namespaces would be very nice, but probably deserves more discussion first.

plugins/optimization-detective/tests/test-class-od-tag-visitor-registry.php
westonruter: Good point. Removed in 48b930b481259e8801502c1ab23fb1108e8d4d2c

PR #1236, opened 2024-05-21
Ensure content filter runs after `wp_filter_content_tags`

### LINKED ISSUE #1207: `srcset` not created when "Generate JPEG files in addition to WebP" enabled
## Bug Description

When the "Generate JPEG files in addition to WebP" option is enabled, the HTML output does not include the expected `srcset`, instead, only the main image is referenced.

## Steps to reproduce

1. Install the modern images feature
2. enabled the Generate JPEG option under settings->media
3. upload a jpeg image to a post
4. check the source code

Expected results: in `img` tag uses `srcset`
Actual results: the `img` tag lags a `srcset` attribute

## Screenshots

Setting:
image

With setting **disabled**:
image

With setting **enabled**:
image

Note: I discovered this issue when working to update the plugin to support AVIF: https://github.com/WordPress/performance/pull/1176

adamsilverstein:
## Summary

Ensures `srcset` is generated when JPEG generation is enabled

Fixes #1207

## Relevant technical choices

Delay filtering`the_content` so `srcset` is properly generated when "Generate JPEG files in addition to WebP" is enabled.


adamsilverstein:
I tested this fix with the image from the issue and also verified the fallback script still works correctly.

joemcgill (approved):
This will work, but I'd suggest using the `wp_calculate_image_srcset` filter for this instead so you don't have to rerun a `preg_match` on the final content. Approving for now since this addresses the issue.

adamsilverstein:
> This will work, but I'd suggest using the `wp_calculate_image_srcset` filter for this instead so you don't have to rerun a `preg_match` on the final content. Approving for now since this addresses the issue.

Great suggestion, I will open an issue to follow up with this improvement.

westonruter:
Otherwise I was goin to suggest that `webp_uploads_update_image_references()` would do well to incorporate `WP_HTML_Tag_Processor`. But I guess the function would go away entirely.

adamsilverstein:
> This will work, but I'd suggest using the `wp_calculate_image_srcset` filter for this instead so you don't have to rerun a `preg_match` on the final content. Approving for now since this addresses the issue.

I created a follow up [Issue](https://github.com/WordPress/performance/issues/1259) to consider this improvement.

PR #1241, opened 2024-05-23
Add PHPStan strict rules (except for empty.notAllowed)

### LINKED ISSUE #1219: Add checks to disallow use of empty()
## Feature Description

The use of `empty()` and `isset( $var )` can mask code problems (e.g. typos). And they rarely need to be used, for example:

* https://dev.to/klnjmm/never-use-empty-function-in-php-4pb0
* https://localheinz.com/articles/2023/05/10/avoiding-empty-in-php/

cf. comment by @felixarntz in https://github.com/WordPress/performance/pull/1091#discussion_r1543908524

We should consider adding sniffs to warn against their use. (Granted, PHPStan should catch the problematic uses of `isset()` and `empty()`.)

westonruter:
Part of #1219
See #775

Note: This does not fix the `empty.notAllowed` issues since they are more complicated and deserve a separate PR.

Just adding `phpstan-strict-rules` results in the following 863 errors in the codebase:

```
    712 staticMethod.dynamicCall
     60 empty.notAllowed
     36 booleanNot.exprNotBoolean
     21 if.condNotBoolean
     11 booleanAnd.leftNotBoolean
      5 cast.useless
      4 booleanOr.rightNotBoolean
      4 arrayFilter.strict
      3 method.childReturnType
      2 ternary.condNotBoolean
      2 booleanAnd.rightNotBoolean
      1 variable.implicitArray
      1 plus.leftNonNumeric
      1 elseif.condNotBoolean
```

This report was generated via:

```bash
npm run phpstan -- -- -v | grep 🪪 | cut -c16- | sed 's/ *$//' | sort | uniq -c | sort -nr
```

When ignoring `staticMethod.dynamicCall`, `booleanNot.exprNotBoolean`, and `if.condNotBoolean` in tests, this goes down to just 144 errors:

```
     60 empty.notAllowed
     34 booleanNot.exprNotBoolean
     16 if.condNotBoolean
     11 booleanAnd.leftNotBoolean
      5 cast.useless
      4 booleanOr.rightNotBoolean
      4 arrayFilter.strict
      3 method.childReturnType
      2 ternary.condNotBoolean
      2 booleanAnd.rightNotBoolean
      1 variable.implicitArray
      1 plus.leftNonNumeric
      1 elseif.condNotBoolean
```

The following remain to be addressed:

- [ ] `empty.notAllowed`                 (60) 
- [x] `booleanNot.exprNotBoolean`        (34) 
- [x] `if.condNotBoolean`                (16) 
- [x] `booleanAnd.leftNotBoolean`        (11) 
- [x] `cast.useless`                      (5) 
- [x] `booleanOr.rightNotBoolean`         (4) 
- [x] ~`arrayFilter.strict`                (4)~ (rule disabled)
- [x] `method.childReturnType`            (3) 
- [x] `ternary.condNotBoolean`            (2) 
- [x] `booleanAnd.rightNotBoolean`        (2) 
- [x] `variable.implicitArray`            (1) 
- [x] `plus.leftNonNumeric`               (1) 
- [x] `elseif.condNotBoolean`             (1)


westonruter:
The one last PHPStan error remaining is `empty.notAllowed`. There are 60 instances of this.

phpstan.neon.dist
```diff
@@ -57,3 +59,6 @@ parameters:
 		-
 			identifier: cast.string
 			path: */tests/*
+		-
+			identifier: staticMethod.dynamicCall
```
westonruter, line 64: I'm not sure why this issue is being reported for assertion method calls in tests.

plugins/embed-optimizer/hooks.php
```diff
@@ -48,7 +48,7 @@ function embed_optimizer_filter_oembed_html( string $html ): string {
 				}
 			}
 		} elseif ( 'SCRIPT' === $html_processor->get_tag() ) {
-			if ( ! $html_processor->get_attribute( 'src' ) ) {
+			if ( ! is_string( $html_processor->get_attribute( 'src' ) ) ) {
```
joemcgill: We can avoid the `is_string` call and confirm that the result is not `null`.

```suggestion
			if ( null !== $html_processor->get_attribute( 'src' ) ) {
```

🔢

plugins/embed-optimizer/hooks.php
```diff
@@ -63,7 +63,7 @@ function embed_optimizer_filter_oembed_html( string $html ): string {
 	if ( 1 === $script_count && ! $has_inline_script && $html_processor->has_bookmark( 'script' ) ) {
 		add_action( 'wp_footer', 'embed_optimizer_lazy_load_scripts' );
 		if ( $html_processor->seek( 'script' ) ) {
-			if ( $html_processor->get_attribute( 'type' ) ) {
+			if ( is_string( $html_processor->get_attribute( 'type' ) ) ) {
```
joemcgill: Same as above.
```suggestion
			if ( null !== $html_processor->get_attribute( 'type' ) ) ) {
```

plugins/performance-lab/includes/admin/load.php
```diff
@@ -465,7 +465,7 @@ function perflab_get_plugin_settings_url( string $plugin_slug ): ?string {
 		return null;
 	}
 	$href = $p->get_attribute( 'href' );
-	if ( $href && is_string( $href ) ) {
+	if ( is_string( $href ) && '' !== $href ) {
```
joemcgill, line 468: I could go either way here, but for consistency...
```suggestion
	if ( null !== $href && '' !== $href ) {
```

plugins/performance-lab/includes/server-timing/class-perflab-server-timing.php
```diff
@@ -67,7 +67,7 @@ public function register_metric( string $metric_slug, array $args ): void {
 			return;
 		}
 
-		if ( did_action( 'perflab_server_timing_send_header' ) && ! doing_action( 'perflab_server_timing_send_header' ) ) {
+		if ( 0 !== did_action( 'perflab_server_timing_send_header' ) && ! doing_action( 'perflab_server_timing_send_header' ) ) {
```
joemcgill, line 70: This is such a common pattern in WP that I wouldn't even think of needing to specify, but makes sense.

plugins/performance-lab/includes/site-health/audit-enqueued-assets/helper.php
```diff
@@ -287,7 +287,7 @@ function perflab_aea_get_total_size_bytes_enqueued_styles() {
  * @return string Returns absolute path to the resource.
  */
 function perflab_aea_get_path_from_resource_url( string $resource_url ): string {
-	if ( ! $resource_url ) {
+	if ( '' === $resource_url ) {
 		return '';
```
joemcgill, line 291: Could just return the value early in this case.
```suggestion
	if ( '' === $resource_url ) {
		return $resource_url;
```

joemcgill (approved):
This looks good to me. I've added some suggestions, but none blocking so I'm pre-approving. Feel free to apply or ignore.

plugins/embed-optimizer/hooks.php
westonruter: The return value of `get_attribute()` is either `string`, `null`, or `true`.  So if we just checked if not `null` then this could actually be a case where `<script src>` which wouldn't be right. This is a script which isn't inline but also its external URL is void so it doesn't make any request. So that's somewhat of a unique case, unlikely to be encountered. So there are three possible states:

1. `<script src="script.js"></script>`
2. `<script>document.write('hi');</script>`
3. `<script src></script>`

The third case is not treated like either an inline script or an external script. But since it doesn't make sense for Embed Optimizer to try to lazy-load a void `src`, then I think it makes sense to treat it as if it is an inline script.

plugins/embed-optimizer/hooks.php
westonruter: In this case, I think the code should stay as-is.  Since `get_attribute` can return `true` in the case of a boolean attribute, we wouldn't want to try passing a `true` to `set_attrbute()`. I believe PHPStan would complain about that anyway.

plugins/performance-lab/includes/admin/load.php
westonruter: This should stay as is because if `<a href>` is encountered, then `$href` here would be `true` which would be a type error since the function returns a `string` or `null`.

swissspidy (approved):
LGTM, pending the merge conflict resolution

plugins/webp-uploads/hooks.php
```diff
@@ -534,22 +534,26 @@ function webp_uploads_update_image_references( $content ): string {
...
 		}
 
 		// Find the ID of each image by the class.
 		// TODO: It would be preferable to use the $processor->class_list() method but there seems to be some typing issues with PHPStan.
 		$class_name = $processor->get_attribute( 'class' );
-		if ( ! is_string( $class_name ) || ! preg_match( '/(?:^|\s)wp-image-([1-9]\d*)(?:\s|$)/i', $class_name, $matches ) ) {
+		if (
+			! is_string( $class_name )
+			||
+			1 !== preg_match( '/(?:^|\s)wp-image-([1-9]\d*)(?:\s|$)/i', $class_name, $matches )
```
mukeshpanchal27, line 555:
```suggestion
			1 !== (int) preg_match( '/(?:^|\s)wp-image-([1-9]\d*)(?:\s|$)/i', $class_name, $matches )
```

Do we needs to cast `int` here?

westonruter: No, because it can either be an `int` or `false`, and we're checking to see if it is not identical to the value.

plugins/webp-uploads/hooks.php
```diff
@@ -534,22 +534,26 @@ function webp_uploads_update_image_references( $content ): string {
 
 	// This content does not have any tag on it, move forward.
 	// TODO: Eventually this should use the HTML API to parse out the image tags and then update them.
-	if ( ! preg_match_all( '/<(img)\s[^>]+>/', $content, $img_tags, PREG_SET_ORDER ) ) {
+	if ( 0 === (int) preg_match_all( '/<(img)\s[^>]+>/', $content, $img_tags, PREG_SET_ORDER ) ) {
```
westonruter, line 537: The use of `(int)` isn't needed here either...

westonruter: Well, actually it is. Because it's `preg_match_all()` so we don't know the number of image tags being matched. Otherwise, it could be `1 === ...`. So by casting the return value to `(int)`, it will convert both `0` and `false` to `0`. Alternatively, this could be:

```suggestion
	if ( false === (bool) preg_match_all( '/<(img)\s[^>]+>/', $content, $img_tags, PREG_SET_ORDER ) ) {
```

But `false` indicates the error state. The integer return value is the normal case, the number of matches.

plugins/webp-uploads/hooks.php
```diff
@@ -534,22 +534,26 @@ function webp_uploads_update_image_references( $content ): string {
...
 		return $content;
 	}
 
 	$images = array();
 	foreach ( $img_tags as list( $img ) ) {
 		$processor = new WP_HTML_Tag_Processor( $img );
-		if ( ! $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
+		if ( 0 === (int) $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
```
westonruter: This change was not warranted.

```suggestion
		if ( ! $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
```

tests/bootstrap.php
```diff
@@ -74,7 +74,7 @@
 		exit( 1 );
 	}
 
-	if ( preg_match( '/^ \* Requires Plugins:\s*(.+)$/m', (string) file_get_contents( $plugin_file ), $matches ) ) {
+	if ( 0 < (int) preg_match( '/^ \* Requires Plugins:\s*(.+)$/m', (string) file_get_contents( $plugin_file ), $matches ) ) {
```
westonruter: This is better:

```suggestion
	if ( 1 !== preg_match( '/^ \* Requires Plugins:\s*(.+)$/m', (string) file_get_contents( $plugin_file ), $matches ) ) {
```

mukeshpanchal27 (approved):
Thanks @westonruter, LGTM!

mukeshpanchal27:
@westonruter Could you please merge as it not allow me. Shows that `Merging can be performed automatically with 1 approving review.`

PR #1247, opened 2024-05-24
Add web worker offloading plugin

### LINKED ISSUE #176: Partytown support

thelovekesh:
## Summary

Fixes https://github.com/WordPress/performance/issues/176

Previously: https://github.com/WordPress/performance/pull/556

## Relevant technical choices

- Add Web Worker Offloading standalone plugin which allows running scripts within the worker thread with the help of the Partytown library.
- Add the `wwo_configuration` filter which can be used to update web worker [configurations](https://partytown.builder.io/configuration).

### Testing steps
- Enable this performance plugin or use `wp-env` in which this plugin is enabled by default on this branch.
- Add a JS script with `web-worker-offloading` as a dependency. When the page is loaded, the JS script should be loaded within the web worker.

  ```php
  function enqueue_non_critical_js() {
	  wp_enqueue_script(
		  'non-critical-js',
		  '/path/to/non-critical.js',
		  array( 'web-worker-offloading' ), // Note the `web-worker-offloading` being added as a dep
		  '1.0.0',
		  true
	  );
  }
  add_action( 'wp_enqueue_scripts', 'enqueue_non_critical_js' );
  ```
- Use `wwo_configuration` hook to update web worker [configurations](https://partytown.builder.io/configuration).
    ```php
    function update_wwo_config( $config ) {
      $config["debug"] = true;
      return $config;
    }
    add_filter( 'wwo_configuration', 'update_wwo_config' );
    ```


thelovekesh:
@westonruter @adamsilverstein I have converted this into a standalone plugin. Next, I'll add a settings page to opt-in offloading for scripts listed in the Partytown documentation (https://partytown.builder.io/common-services). This way, the WordPress plugin team won't classify it as a "framework" plugin, as @westonruter previously mentioned.

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,146 @@
...
+		if ( ! empty( $before_script ) || ! empty( $after_script ) ) {
+			_doing_it_wrong(
+				'wp_add_inline_script',
+				sprintf(
+					/* translators: %s: script handle */
+					esc_html__( 'Cannot add inline script "%s" to scripts with a "partytown" dependency. Script will continue to load in the main thread.', 'web-worker-offloading' ),
+					'<a href="https://developer.wordpress.org/reference/functions/wp_add_inline_script/">wp_add_inline_script()</a>'
+				),
+				esc_html( WEB_WORKER_OFFLOADING_VERSION )
+			);
```
thelovekesh: Can I get some feedback on this part? It doesn't offload a script to a web worker if it has dependent scripts that load before or after it. My concern is that doing so might disrupt the loading order or break some dependencies.

westonruter: It should be fine to have a before script, but an after script would indeed be a problem.

thelovekesh: What happens if the `before` script is async? Partytown searches for scripts with `type="text/partytown"` once `DOMContentLoaded` is triggered. If the before script is `async` and still downloading, it won't block `DOMContentLoaded`, allowing Partytown to start executing the scripts in the worker thread.

westonruter: Inline scripts cannot be `async` in general. The only way to _defer_ an inline script would be to make it a module, but this isn't supported by the `wp_add_inline_script()` API.

With the introduction of the script loading strategies, there is this [`filter_eligible_strategies()`](https://github.com/WordPress/wordpress-develop/blob/2c818502ea70b27c36f420112152b26bd0ebe8fe/src/wp-includes/class-wp-scripts.php#L925C19-L980) method which is useful to refer to. 

Note this logic:

```php
// Handles with inline scripts attached in the 'after' position cannot be delayed.
if ( $this->has_inline_script( $handle, 'after' ) ) {
	return array();
}
```

So an enqueued script which is `async` falls back to blocking if it has an inline `after` script. If it has an inline `before` script, it can still be async.

Partytown scripts should essentially behave as if they are `async` scripts. In fact, it would be good to set the `async` strategy early on in addition to any Partytown-specific data flag on all Partytown scripts. If this `async` strategy "survives" the checks done by `filter_eligible_strategies` then the `script_loader_tag` filter can go ahead and replace the `async` attribute with `type="text/partytown"`. In this way, the `$before_script` and `$after_script` logic here can be eliminated.

thelovekesh: Thank you for the detailed explanation. It makes everything clear.

If a script is designed to run in a worker thread but is forced to remain asynchronous, should we consider implementing something to provide feedback about its behavior? Else it can confuse the user about the implementation.

westonruter: How about something like this:

```php

// Make sure that all partytown scripts are treated as if they are async.
add_filter(
	'print_scripts_array',
	static function ( array $script_handles ): array {
		foreach ( array_intersect( $script_handles, wwo_get_partytown_handles() ) as $handle ) {
			wp_script_add_data( $handle, 'strategy', 'async' );
		}
		return $script_handles;
	}
);

function wwo_update_script_type( string $tag, string $handle, string $src ): string {
	$partytown_handles = wwo_get_partytown_handles();

	if ( in_array( $handle, $partytown_handles, true ) ) {
		$p = new WP_HTML_Tag_Processor( $tag );
		if ( $p->next_tag( array( 'tag_name' => 'SCRIPT' ) ) ) {
			if ( ! $p->get_attribute( 'async' ) ) {
				_doing_it_wrong(
					'wwo_...something_something',
					esc_html(
						sprintf(
							/* translators: %s: script handle */
							__( 'Unable to offload "%s" script to a worker. Script will continue to load in the main thread.', 'web-worker-offloading' )
						)
					)
				);
			} else {
				$p->set_attribute( 'type', 'text/partytown' );
				$p->remove_attribute( 'async' );
				$tag = $p->get_updated_html();
			}
		}
	}

	return $tag;
}
```

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,146 @@
...
+			wp_json_encode( wwo_configuration() )
+		),
+		'before'
+	);
+
+	wp_add_inline_script(
+		'partytown',
+		$partytown_js,
+		'after'
+	);
```
westonruter: Idea: Let's not reference `partytown` specifically. Let's call this script something like `web-worker-offloader`. That will keep us from having to be too bound up with the Partytown implementation.

thelovekesh: Done in https://github.com/WordPress/performance/pull/1247/commits/edaa30507ef2a9f7950d82cf8a6344e6a5e31a1a

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,163 @@
...
+	/**
+	 * Array of script handles which has `web-worker-offloading` dependency.
+	 *
+	 * @var string[] $web_worker_offloading_handles
+	 */
+	static $web_worker_offloading_handles = array();
+
+	if ( count( $web_worker_offloading_handles ) > 0 ) {
+		return $web_worker_offloading_handles;
+	}
```
westonruter: This makes me a bit nervous because what the scripts are printed in head, and then in the body a new script is registered for printing in the footer? Then it would never be picked up.

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,163 @@
...
+ * @since n.e.x.t
+ *
+ * @param string $tag    Script tag.
+ * @param string $handle Script handle.
+ * @return string Script tag with type="text/partytown" for eligible scripts.
+ */
+function wwo_update_script_type( string $tag, string $handle ): string {
+	$web_worker_offloading_handles = wwo_get_web_worker_offloading_handles();
+
+	if ( in_array( $handle, $web_worker_offloading_handles, true ) ) {
```
westonruter: Wouldn't this work just as well?

```suggestion
	if ( in_array( 'web-worker-offloading', wp_scripts()->registered[ $handle ]->deps, true ) ) {
```

plugins/web-worker-offloading/hooks.php
westonruter: Per the below suggestion, this entire block of code could be removed since it is only executed a couple times when the `print_scripts_array` filter applies.

```suggestion
	$web_worker_offloading_handles = array();
```

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,163 @@
...
+ * Scripts offloaded to a worker thread can be considered async. However, they may include `before` and `after` inline
+ * scripts that need sequential execution. Once marked as async, `filter_eligible_strategies()` determines if the
+ * script is eligible for async execution. If so, it will be offloaded to the worker thread.
+ *
+ * @since n.e.x.t
+ *
+ * @param string[] $script_handles Array of script handles.
+ * @return string[] Array of script handles.
+ */
+function wwo_update_script_strategy( array $script_handles ): array {
```
westonruter: Could you add a test for this?

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * Update script type for handles having `web-worker-offloading` as dependency.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $tag    Script tag.
+ * @param string $handle Script handle.
+ * @return string Script tag with type="text/partytown" for eligible scripts.
+ */
+function wwo_update_script_type( string $tag, string $handle ): string {
```
westonruter: This needs tests added.

plugins/web-worker-offloading/tests/test-web-worker-offloading.php
```diff
@@ -0,0 +1,105 @@
...
+		$config = wwo_get_configuration();
+
+		$this->assertArrayHasKey( 'forward', $config );
+		$this->assertArrayHasKey( 'debug', $config );
+		$this->assertNotEmpty( $config['forward'] );
+		$this->assertIsArray( $config['forward'] );
+		$this->assertTrue( $config['debug'] );
+		$this->assertContains( 'datalayer.push', $config['forward'] );
+
+		remove_all_filters( 'wwo_configuration' );
```
westonruter: Filters are automatically removed after each test.

```suggestion
```

plugins/web-worker-offloading/tests/test-web-worker-offloading.php
```diff
@@ -0,0 +1,105 @@
...
+
+		// Ensure that Partytown is enqueued when a script depends on it.
+		wp_enqueue_script( 'partytown-test', 'https://example.com/test.js', array( 'web-worker-offloading' ) );
+
+		$this->assertTrue( wp_script_is( 'web-worker-offloading', 'enqueued' ) );
+		$this->assertTrue( wp_script_is( 'partytown-test', 'enqueued' ) );
+
+		// Reset the state.
+		wp_scripts()->remove( 'web-worker-offloading' );
+		wp_scripts()->remove( 'partytown-test' );
```
westonruter: I suggest unsetting `$wp_scripts` in a `tear_down` method.

```suggestion
```

plugins/web-worker-offloading/tests/test-web-worker-offloading.php
```diff
@@ -0,0 +1,105 @@
...
+		$handles = wwo_get_web_worker_offloading_handles();
+
+		$this->assertNotEmpty( $handles );
+		$this->assertContains( 'partytown-test', $handles );
+		$this->assertTrue( wp_script_is( 'web-worker-offloading', 'enqueued' ) );
+		$this->assertTrue( wp_script_is( 'partytown-test', 'enqueued' ) );
+
+		// Reset the state.
+		wp_scripts()->remove( 'web-worker-offloading' );
+		wp_scripts()->remove( 'partytown-test' );
```
westonruter: Ditto above.

```suggestion
```

plugins/web-worker-offloading/hooks.php
westonruter: Actually, for both `wwo_update_script_strategy()` and `wwo_update_script_type()` it would be good to have a test that doesn't necessarily check the specific functions themselves, but rather does an end-to-end test building on top of what you've done in `test_wwo_get_web_worker_offloading_handles`. What it is missing is actually printing the scripts and then examining the output to see if it has the expected markup.

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * @param string[] $script_handles Array of script handles.
+ * @return string[] Array of script handles.
+ */
+function wwo_update_script_strategy( array $script_handles ): array {
+	$web_worker_offloading_handles = wwo_get_web_worker_offloading_handles();
+
+	foreach ( array_intersect( $script_handles, $web_worker_offloading_handles ) as $handle ) {
+		wp_script_add_data( $handle, 'strategy', 'async' );
+	}
```
westonruter: This would eliminate the need for `wwo_get_web_worker_offloading_handles()`:

```suggestion
	foreach ( $script_handles as $handle ) {
		if ( in_array( 'web-worker-offloading', wp_scripts()->registered[ $handle ]->deps, true ) ) {
			wp_script_add_data( $handle, 'strategy', 'async' );
		}
	}
```

westonruter (changes requested):
So sorry for the very tardy review!

I've fixed the merge conflicts and updated some code primarily to fix PHPStan strict rule compliance.

westonruter:
The other thing is that it would be great if this plugin actually did something out-of-the-box without requiring developers to extend it.

These are the common 3rd party services known to work with Partytown: https://partytown.builder.io/common-services

For Google Tag Manager, for example, could it automatically locate the script and add `text/partytown` to it? Probably can't be done reliably without output buffering and the HTML API in WP 6.7-alpha. But this can be done in a separate PR.

thelovekesh:
@westonruter Yes that's the plan. Once we have a few stable things, I will work on this.

plugins/web-worker-offloading/hooks.php
thelovekesh: @westonruter Can you please confirm if this is what you expected - https://github.com/WordPress/performance/pull/1247/commits/3ca3314cdf9ef1683f0471b56fdb840145168b78?

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,133 @@
...
+	return apply_filters( 'wwo_configuration', $config );
+}
+
+/**
+ * Initialize Web Worker Offloading.
+ *
+ * @since n.e.x.t
+ */
+function wwo_init(): void {
+	$partytown_js = file_get_contents( __DIR__ . '/build/partytown.js' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents -- It's a local filesystem path not a remote request.
```
adamsilverstein, line 44: Did you consider using WP_Filesystem? (and remove the phpcs warning)

plugins/web-worker-offloading/readme.txt
```diff
@@ -0,0 +1,16 @@
+=== Web Worker Offloading ===
+
+Contributors:      wordpressdotorg
+Requires at least: 6.4
+Tested up to:      6.5
```
adamsilverstein:
```suggestion
Tested up to:      6.6
```

adamsilverstein (commented):
Excellent! Lets some small feedback/questions.

plugins/web-worker-offloading/hooks.php
westonruter: That may be better, but core uses `file_get_contents()` for code like this:

https://github.com/WordPress/wordpress-develop/blob/9ed3553da6a61a38a1d4c5585e5ac3c9d13d351a/src/wp-includes/embed.php#L521
https://github.com/WordPress/wordpress-develop/blob/9ed3553da6a61a38a1d4c5585e5ac3c9d13d351a/src/wp-includes/formatting.php#L5962

plugins/web-worker-offloading/tests/test-web-worker-offloading.php
```diff
@@ -0,0 +1,192 @@
...
+
+	/**
+	 * Data provider for testing `wwo_update_script_type`.
+	 *
+	 * @return array<string, mixed> Data.
+	 */
+	public static function data_update_script_types(): array {
+		return array(
+			'add-script'                           => array(
+				'setup'          => static function (): void {
```
westonruter: _Tiny_ nit: let's replace `setup` with `set_up`, given that the former is a noun and the latter is a verb (and is used as method name in unit tests).

plugins/web-worker-offloading/hooks.php
westonruter: Perfect!

.eslintignore
westonruter, line 1: Apparently the `.eslintignore` file is deprecated: https://eslint.org/docs/latest/use/configure/ignore-deprecated#the-eslintignore-file

It looks like now they want you to put it in in `.eslintrc.js`, and `eslint.config.js` is now preferred: https://eslint.org/docs/latest/use/configure/ignore

I see that we already have an `ignorePatterns` defined in our `.eslintrc.js`: 

https://github.com/WordPress/performance/blob/08eba80fd8609add2df44d9960dc011868dc5d93/.eslintrc.js#L21

So these new entries should be added there instead.

.gitignore
```diff
@@ -31,7 +32,6 @@ build
 
 node_modules/
 vendor/
-plugins/optimization-detective/detection/web-vitals*
```
westonruter, line 34: Good catch.

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,133 @@
...
+					_doing_it_wrong(
+						'wwo_update_script_type',
+						esc_html(
+							sprintf(
+								/* translators: %s: script handle */
+								__( 'Unable to offload "%s" script to a worker. Script will continue to load in the main thread.', 'web-worker-offloading' ),
+								$handle
+							)
+						),
+						esc_html( WEB_WORKER_OFFLOADING_VERSION )
```
westonruter:
```suggestion
						'Web Worker Offloading n.e.x.t'
```

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,133 @@
...
+add_action( 'wp_enqueue_scripts', 'wwo_init' );
+
+/**
+ * Mark scripts with `web-worker-offloading` dependency as async.
+ *
+ * Why this is needed?
+ *
+ * Scripts offloaded to a worker thread can be considered async. However, they may include `before` and `after` inline
+ * scripts that need sequential execution. Once marked as async, `filter_eligible_strategies()` determines if the
+ * script is eligible for async execution. If so, it will be offloaded to the worker thread.
```
westonruter:
```suggestion
 * This is needed because scripts offloaded to a worker thread can be considered async. However, they may include `before` and `after` inline
 * scripts that need sequential execution. Once marked as async, `filter_eligible_strategies()` determines if the
 * script is eligible for async execution. If so, it will be offloaded to the worker thread.
```

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,133 @@
...
+		),
+		'before'
+	);
+
+	wp_add_inline_script( 'web-worker-offloading', $partytown_js );
+}
+add_action( 'wp_enqueue_scripts', 'wwo_init' );
+
+/**
+ * Mark scripts with `web-worker-offloading` dependency as async.
```
westonruter:
```suggestion
 * Marks scripts with `web-worker-offloading` dependency as async.
```

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,133 @@
...
+	 * @since n.e.x.t
+	 * @link https://partytown.builder.io/configuration
+	 *
+	 * @param array{ debug?: bool, forward?: non-empty-string[], lib: non-empty-string, loadScriptsOnMainThread?: non-empty-string[], nonce?: non-empty-string } $config Configuration for Partytown.
+	 */
+	return apply_filters( 'wwo_configuration', $config );
+}
+
+/**
+ * Initialize Web Worker Offloading.
```
westonruter:
```suggestion
 * Initializes Web Worker Offloading.
```

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,133 @@
...
+			wp_script_add_data( $handle, 'strategy', 'async' );
+		}
+	}
+
+	return $script_handles;
+}
+add_filter( 'print_scripts_array', 'wwo_update_script_strategy' );
+
+/**
+ * Update script type for handles having `web-worker-offloading` as dependency.
```
westonruter:
```suggestion
 * Updates script type for handles having `web-worker-offloading` as dependency.
```

plugins/web-worker-offloading/readme.txt
```diff
@@ -0,0 +1,16 @@
...
+Stable tag:        n.e.x.t
+License:           GPLv2 or later
+License URI:       https://www.gnu.org/licenses/gpl-2.0.html
+Tags:              performance, JavaScript, web worker, partytown
+
+Offload JavaScript execution to a Web Worker.
+
+== Description ==
+
+This plugin offloads JavaScript execution to a Web Worker, improving performance by freeing up the main thread.
```
westonruter, line 15:
```suggestion
This plugin offloads JavaScript execution to a Web Worker, improving performance by freeing up the main thread.

In order to opt-in a script to be loaded in a worker, simply add the `web-worker-offloading` script as a dependency.
```

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,133 @@
...
+ * script is eligible for async execution. If so, it will be offloaded to the worker thread.
+ *
+ * @since n.e.x.t
+ *
+ * @param string[] $script_handles Array of script handles.
+ * @return string[] Array of script handles.
+ */
+function wwo_update_script_strategy( array $script_handles ): array {
+	foreach ( $script_handles as $handle ) {
+		if ( in_array( 'web-worker-offloading', wp_scripts()->registered[ $handle ]->deps, true ) ) {
```
westonruter: Some hardening in case (1) another filter fails to return an array, and (2) a filter returns a handle for a script that was unregistered (or was never registered):

```suggestion
 * @param string[]|mixed $script_handles Array of script handles.
 * @return string[] Array of script handles.
 */
function wwo_update_script_strategy( $script_handles ): array {
	$script_handles = array_intersect( (array) $script_handles, array_keys( wp_scripts()->registered ) );
	foreach ( $script_handles as $handle ) {
		if ( in_array( 'web-worker-offloading', wp_scripts()->registered[ $handle ]->deps, true ) ) {
```

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,133 @@
...
+ * Update script type for handles having `web-worker-offloading` as dependency.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $tag    Script tag.
+ * @param string $handle Script handle.
+ * @return string Script tag with type="text/partytown" for eligible scripts.
+ */
+function wwo_update_script_type( string $tag, string $handle ): string {
+	if ( in_array( 'web-worker-offloading', wp_scripts()->registered[ $handle ]->deps, true ) ) {
```
westonruter: More hardening:

```suggestion
 * @param string|mixed $tag Script tag.
 * @param string $handle    Script handle.
 * @return string Script tag with type="text/partytown" for eligible scripts.
 */
function wwo_update_script_type( $tag, string $handle ): string {
	if (
		is_string( $tag ) && 
		array_key_exists( $handle, wp_scripts()->registered ) && 
		in_array( 'web-worker-offloading', wp_scripts()->registered[ $handle ]->deps, true ) 
	) {
```

.eslintignore
thelovekesh: Humm yes but with ESLint 9+ I think. I have removed the ignore file and added the ignore pattern in the config file instead. Once WP packages support ESLint 9+, it will require some more updates though.

plugins/web-worker-offloading/hooks.php
thelovekesh: The `$tag` is always string here as per phpdoc. Should we consider those phpdoc as uncertain?

thelovekesh: Oh, it's about if some other filter has made some updated. In that case, shouldn't there be an error to let the dev know they are doing something wrong?

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Updates script type for handles having `web-worker-offloading` as dependency.
+ *
+ * @since n.e.x.t
+ *
+ * @param string|mixed $tag    Script tag.
+ * @param string       $handle Script handle.
+ *
+ * @return string Script tag with type="text/partytown" for eligible scripts.
```
westonruter: Since we pass through the value of it isn't a string.

```suggestion
 * @return string|mixed Script tag with type="text/partytown" for eligible scripts.
```

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,138 @@
...
+ */
+function wwo_update_script_type( $tag, string $handle ): string {
```
westonruter:
```suggestion
function wwo_update_script_type( $tag, string $handle ) {
```

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,138 @@
...
+ * Marks scripts with `web-worker-offloading` dependency as async.
+ *
+ * This is needed because scripts offloaded to a worker thread can be considered async. However, they may include `before` and `after` inline
+ * scripts that need sequential execution. Once marked as async, `filter_eligible_strategies()` determines if the
+ * script is eligible for async execution. If so, it will be offloaded to the worker thread.
+ *
+ * @since n.e.x.t
+ *
+ * @param string[]|mixed $script_handles Array of script handles.
+ *
```
westonruter:
```suggestion
```

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,138 @@
...
+add_filter( 'print_scripts_array', 'wwo_update_script_strategy' );
+
+/**
+ * Updates script type for handles having `web-worker-offloading` as dependency.
+ *
+ * @since n.e.x.t
+ *
+ * @param string|mixed $tag    Script tag.
+ * @param string       $handle Script handle.
+ *
```
westonruter:
```suggestion
```

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,138 @@
...
+							)
+						),
+						'Web Worker Offloading n.e.x.t'
+					);
+				} else {
+					$html_processor->set_attribute( 'type', 'text/partytown' );
+					$html_processor->remove_attribute( 'async' );
+					$tag = $html_processor->get_updated_html();
+				}
+			}
```
westonruter: Minor suggestion to reduce nesting:

```suggestion
			if ( $html_processor->get_attribute( 'id' ) !== "{$handle}-js" ) {
				continue;
			}
			if ( null === $html_processor->get_attribute( 'async' ) ) {
				_doing_it_wrong(
					'wwo_update_script_type',
					esc_html(
						sprintf(
							/* translators: %s: script handle */
							__( 'Unable to offload "%s" script to a worker. Script will continue to load in the main thread.', 'web-worker-offloading' ),
							$handle
						)
					),
					'Web Worker Offloading n.e.x.t'
				);
			} else {
				$html_processor->set_attribute( 'type', 'text/partytown' );
				$html_processor->remove_attribute( 'async' );
				$tag = $html_processor->get_updated_html();
			}
```

plugins/web-worker-offloading/readme.txt
```diff
@@ -0,0 +1,18 @@
+=== Web Worker Offloading ===
+
+Contributors:      wordpressdotorg
+Requires at least: 6.4
+Tested up to:      6.6
+Requires PHP:      7.2
```
westonruter: This is exclusively in the bootstrap PHP file now.

```suggestion
```

westonruter (approved):
Just some trivial suggestions left from me. Pre-approving so that once @adamsilverstein approves this can finally be merged!

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,137 @@
...
+ * @param string       $handle Script handle.
+ * @return string|mixed Script tag with type="text/partytown" for eligible scripts.
+ */
+function wwo_update_script_type( $tag, string $handle ) {
+	if (
+		is_string( $tag ) &&
+		array_key_exists( $handle, wp_scripts()->registered ) &&
+		in_array( 'web-worker-offloading', wp_scripts()->registered[ $handle ]->deps, true )
+	) {
+		$html_processor = new WP_HTML_Tag_Processor( $tag );
```
adamsilverstein, line 112: Nice!

adamsilverstein:
Code looks good to me, doing some manual testing next.

adamsilverstein:
Testing this locally with the sample code from the ticket description, I noticed the script handle is different, I had to use 
'web-worker-offloading' while the instructions for testing in the description use 'web-worker-offloader'.  Can you update that in the testing instructions for posterity?

Once I did that, the tag got the partytown type added:

image

question: do we want to remove the `data-wp-strategy` attribute as well?

I didn't test actual Partytown functionality yet, that will be easier to do once we ship the plugin. Are we planning add script specific handling (eg analytics) in the future?

adamsilverstein (approved):
Tremendous!

westonruter:
> Testing this locally with the sample code from the ticket description, I noticed the script handle is different, I had to use
> 'web-worker-offloading' while the instructions for testing in the description use 'web-worker-offloader'. Can you update that in the testing instructions for posterity?

Ah, that's my bad. I updated the slug in the codebase but didn't do it in the description. Updated now.

> question: do we want to remove the `data-wp-strategy` attribute as well?

You're right. This can be removed as well. Or the `async` attribute can be left as-is. I think it may make sense to keep `async` since this in effect what offloading to Partytown results in. (Or maybe `defer` actually. I'm not entirely sure of the execution model.) But ultimately, on second thought, let's remove it because if the user didn't add `async` to begin with, it could be confusing why it is showing up.

plugins/web-worker-offloading/hooks.php
```diff
@@ -0,0 +1,137 @@
...
+							/* translators: %s: script handle */
+							__( 'Unable to offload "%s" script to a worker. Script will continue to load in the main thread.', 'web-worker-offloading' ),
+							$handle
+						)
+					),
+					'Web Worker Offloading n.e.x.t'
+				);
+			} else {
+				$html_processor->set_attribute( 'type', 'text/partytown' );
+				$html_processor->remove_attribute( 'async' );
```
westonruter, line 134:
```suggestion
				$html_processor->remove_attribute( 'async' );
				$html_processor->remove_attribute( 'data-wp-strategy' );
```

westonruter:
In 44c469c I addressed a case where a script may already have the `defer` or `async` strategy and then its opted-into offloading to web worker. I've added the plugin banner and icon asset. 

✅ Good to merge.

westonruter:
Follow up: https://github.com/WordPress/performance/issues/1468

PR #1250, opened 2024-05-27
Initial implementation of improved image `sizes` algorithm

### LINKED ISSUE #1187: Initial implementation of improved image `sizes` algorithm
## Overview

This issue cover adding the first set of enhancement to the auto-sizes plugin, that improve the default image `sizes` algorithm by constraining the maximum `sizes` width by layout settings from `theme.json` and the content width setting in classic themes for default aligned images.

## Approach

To kick off this effort, I’ve created [a proof of concept implementation](https://gist.github.com/joemcgill/7b4549fddf17f9d201b1f152d26dcc26), which adds the `sizes` attribute to images in blocks during the block rendering process, via block render filters. This approach is based on the following initial proposal for new algorithms based on alignment values:

- **default:** `(max-width: contentSize + padding) calc(100vw - padding), contentSize`
- **wide:** `(max-width: wideSize + padding) calc(100vw - padding), widesize`
- **full:** 
  - With `useRootPaddingAwareAlignments`: `100vw`
  - Without `useRootPaddingAwareAlignments`: `calc(100vw - padding)`

## Goal

An initial implementation should meet the following requirements:

- [x] `default` aligned image blocks should incorporate the `contentSize` setting from theme.json
- [x] `wide` aligned image blocks should incorporate the `wideSize` setting, if available
- [x] `full` aligned images should subtract the padding setting when `useRootPaddingAwareAlignments` is not `true`, otherwise be `100vw`

mukeshpanchal27:
## Summary

Fixes #1187


plugins/auto-sizes/hooks.php
```diff
@@ -90,3 +90,50 @@ function auto_sizes_render_generator(): void {
...
+ *
+ * @since n.e.x.t
+ *
+ * @param string               $content      The block content about to be rendered.
+ * @param array<string, mixed> $parsed_block The parsed block.
+ * @return string The updated block content.
+ */
+function auto_sizes_improve_image_sizes_attribute( string $content, array $parsed_block ): string {
+
+	$tags   = new WP_HTML_Tag_Processor( $content );
```
westonruter: Let's call this `$tags` variable `$processor` instead.

plugins/auto-sizes/hooks.php
```diff
@@ -90,3 +90,50 @@ function auto_sizes_render_generator(): void {
...
+	$image  = $tags->next_tag( array( 'tag_name' => 'img' ) );
```
westonruter: Instead of `$image` I think `$has_image` is more clear.

plugins/auto-sizes/hooks.php
```diff
@@ -90,3 +90,50 @@ function auto_sizes_render_generator(): void {
...
+			default:
+				$sizes = sprintf( '(max-width: %1$s) 100vw, %1$s', $layout['contentSize'] );
+				break;
+		}
+
+		if ( $sizes ) {
+			$tags->set_attribute( 'sizes', $sizes );
+		}
+
+		$content = $tags->get_updated_html();
```
westonruter: Minor variable naming improvements:

```suggestion
	$processor = new WP_HTML_Tag_Processor( $content );
	$has_image = $processor->next_tag( array( 'tag_name' => 'img' ) );
	$layout    = wp_get_global_settings( array( 'layout' ) );

	// Only update the markup if an image is found and we have layout settings.
	if ( $has_image && isset( $layout['wideSize'] ) && $layout['contentSize'] ) {

		$align = $parsed_block['attrs']['align'] ?? null;

		// Handle different alignment use cases.
		switch ( $align ) {
			case 'full':
				$sizes = '100vw';
				break;

			case 'wide':
				$sizes = sprintf( '(max-width: %1$s) 100vw, %1$s', $layout['wideSize'] );
				break;

			// @todo: handle left/right alignments.
			default:
				$sizes = sprintf( '(max-width: %1$s) 100vw, %1$s', $layout['contentSize'] );
				break;
		}

		if ( $sizes ) {
			$processor->set_attribute( 'sizes', $sizes );
		}

		$content = $processor->get_updated_html();
```

plugins/auto-sizes/hooks.php
```diff
@@ -90,3 +90,50 @@ function auto_sizes_render_generator(): void {
...
+
+			case 'wide':
+				$sizes = sprintf( '(max-width: %1$s) 100vw, %1$s', $layout['wideSize'] );
+				break;
+
+			// @todo: handle left/right alignments.
+			default:
+				$sizes = sprintf( '(max-width: %1$s) 100vw, %1$s', $layout['contentSize'] );
+				break;
+		}
```
joemcgill: I think we can support `contentSize` even if `wideSize` isn't present. Perhaps it's better to move those checks to the switch cases where they are used instead.

```suggestion
	if ( $image ) {

		$align  = $parsed_block['attrs']['align'] ?? null;
		$sizes = null;

		// Handle different alignment use cases.
		switch ( $align ) {
			case 'full':
				$sizes = '100vw';
				break;

			case 'wide':
				if ( isset( $layout['wideSize'] ) ) {
					$sizes = sprintf( '(max-width: %1$s) 100vw, %1$s',  $layout['wideSize'] );
				}
				
				break;

			// @todo: handle left/right alignments.
			default:
				if ( isset( $layout['contentSize'] ) ) {
					$sizes = sprintf( '(max-width: %1$s) 100vw, %1$s', $layout['contentSize'] );
				}
				
				break;
		}
```

joemcgill (changes requested):
Good start! I think it would be good to add some tests for these early on so we can make sure we're getting the output we expect.

plugins/auto-sizes/tests/improve-sizes-test.php
```diff
@@ -0,0 +1,91 @@
...
+class Tests_Improve_Sizes extends WP_UnitTestCase {
+
+	/**
+	 * Set up the environment for the tests.
+	 */
+	public function setUp(): void {
+		parent::setUp();
+
+		switch_theme( 'twentytwentyfour' );
+	}
```
joemcgill: We can set up the theme once before all tests rather than before each method.
```suggestion
	public static function set_up_before_class(): void {
		parent::set_up_before_class();

		switch_theme( 'twentytwentyfour' );
	}
```

plugins/auto-sizes/tests/improve-sizes-test.php
```diff
@@ -0,0 +1,91 @@
...
+
+	/**
+	 * Test the function with full alignment.
+	 *
+	 * @covers ::auto_sizes_improve_image_sizes_attribute
+	 */
+	public function test_full_alignment(): void {
+		$content      = '<img src="test.jpg" />';
```
joemcgill: This works, but it is probably more accurate to try to crate content that actually mimics what the expected markup will be for an image block. You can review how I've set up fixtures in the `Test_AutoSizes` class to generate a test attachment and get the expected markup to test against.

joemcgill (changes requested):
This looks like a good first implementation and tests well. I've left a few suggestions about improving the tests, but I think this is pretty close to merge so we have a base in the feature branch to work against.

plugins/auto-sizes/hooks.php
```diff
@@ -90,3 +90,54 @@ function auto_sizes_render_generator(): void {
...
+		$align = $parsed_block['attrs']['align'] ?? null;
+		$sizes = null;
+		// Handle different alignment use cases.
+		switch ( $align ) {
+			case 'full':
+				$sizes = '100vw';
+				break;
+
+			case 'wide':
+				if ( isset( $layout['wideSize'] ) ) {
```
westonruter: Per https://github.com/WordPress/performance/issues/1219:

```suggestion
				if ( array_key_exists( 'wideSize', $layout ) ) {
```

plugins/auto-sizes/hooks.php
```diff
@@ -90,3 +90,54 @@ function auto_sizes_render_generator(): void {
...
+					$sizes = sprintf( '(max-width: %1$s) 100vw, %1$s', $layout['wideSize'] );
+				}
+				break;
+
+			// @todo: handle left/right alignments.
+			default:
+				if ( isset( $layout['contentSize'] ) ) {
```
westonruter: Per https://github.com/WordPress/performance/issues/1219:

```suggestion
				if ( array_key_exists( 'contentSize', $layout ) ) {
```

plugins/auto-sizes/tests/improve-sizes-test.php
mukeshpanchal27: Thanks @joemcgill. Unit tests updated in https://github.com/WordPress/performance/pull/1250/commits/a0e4e264c046b169d9921a6bc902b6ab14251e7e

plugins/auto-sizes/hooks.php
```diff
@@ -90,3 +90,54 @@ function auto_sizes_render_generator(): void {
...
+ *
+ * @param string               $content      The block content about to be rendered.
+ * @param array<string, mixed> $parsed_block The parsed block.
+ * @return string The updated block content.
+ */
+function auto_sizes_improve_image_sizes_attribute( string $content, array $parsed_block ): string {
+
+	$processor = new WP_HTML_Tag_Processor( $content );
+	$has_image = $processor->next_tag( array( 'tag_name' => 'img' ) );
+	$layout    = wp_get_global_settings( array( 'layout' ) );
```
westonruter: Minor point: this variable definition can be moved inside of the if statement

mukeshpanchal27: Thanks. Update in https://github.com/WordPress/performance/pull/1250/commits/e2672f441e889511e032d1c618c2d5a880b06bbb

westonruter (approved):

plugins/auto-sizes/tests/improve-sizes-test.php
```diff
@@ -8,32 +8,54 @@
...
 
 		switch_theme( 'twentytwentyfour' );
 	}
 
+	/**
+	 * Setup shared fixtures.
+	 */
+	public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ): void {
+		self::$image_id = $factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/data/images/leaves.jpg' );
+	}
```
joemcgill: Nice! I think this can be combined with the `set_up_before_class()` method above, but you'll need to call `self::factory()->attachment->{etc)`.

joemcgill (approved):
One minor suggestion, but I think this is good. Pre-approving.

plugins/auto-sizes/tests/improve-sizes-test.php
mukeshpanchal27: Fixed in https://github.com/WordPress/performance/pull/1250/commits/d866befe667846653cd5eff35e36f21e19d65979

PR #1252, opened 2024-05-28
Use correct sizes for small images

### LINKED ISSUE #1251: Parse width presets for use in `sizes` values
As described in https://github.com/WordPress/performance/issues/1187#issuecomment-2133426871, If the image is smaller than the content or wide size, it returns the wrong (bigger) size. In such cases, we need to return the smaller value among the image width, content size, or wide size.

The different widths used in theme.json for content and wide size.

- 620px
- min(640px, 90vw)
- 90vw

We should parse the width so we can use proper width in sizes to get proper image size.

mukeshpanchal27:
## Summary

Fixes: #1251

For initial work:

- Only return a smaller image size if the content or wide size is larger than the image size.
   - If  content size is 620px and content use thumbnail the the size will use thumbnail 150px image size instead of 620px.
- Parse the width value for "px"; for other types( vw, min() ), return the image width according to the default WordPress behaviour.


joemcgill:
Thanks @mukeshpanchal27. It looks like the intent of this PR to constrain the `width` attribute value by the layout width. Is that correct? Is this related to an existing issue, or should a new issue be created to associate this PR with?

mukeshpanchal27:
@joemcgill The PR is ready for review.

plugins/auto-sizes/hooks.php
```diff
@@ -91,6 +91,25 @@ function auto_sizes_render_generator(): void {
 }
 add_action( 'wp_head', 'auto_sizes_render_generator' );
 
+/**
+ * Gets the smaller image size if the layout width id bigger.
```
westonruter:
```suggestion
 * Gets the smaller image size if the layout width is bigger.
```

plugins/auto-sizes/hooks.php
```diff
@@ -91,6 +91,25 @@ function auto_sizes_render_generator(): void {
...
+ *
+ * It will return the smaller image size and return "px" if the layout width
+ * id something else. ex. min(640px, 90vw) or 90vw.
```
westonruter:
```suggestion
 * is something else, e.g. min(640px, 90vw) or 90vw.
```

plugins/auto-sizes/hooks.php
```diff
@@ -91,6 +91,25 @@ function auto_sizes_render_generator(): void {
...
+ *
+ * @since n.e.x.t
+ *
+ * @param string $layout_width The layout width.
+ * @param int    $image_width  The image width.
+ * @return string The proper width after some calculations.
+ */
+function auto_sizes_get_width( string $layout_width, int $image_width ): string {
+	if ( 'px' === substr( $layout_width, -2 ) ) {
```
westonruter:
```suggestion
	if ( str_ends_with( $layout_width, 'px' ) {
```

plugins/auto-sizes/hooks.php
```diff
@@ -107,9 +126,18 @@ function auto_sizes_improve_image_sizes_attribute( string $content, array $parse
...
 	if ( $has_image ) {
-		$layout = wp_get_global_settings( array( 'layout' ) );
-		$align  = $parsed_block['attrs']['align'] ?? null;
-		$sizes  = null;
+		$layout     = wp_get_global_settings( array( 'layout' ) );
+		$align      = $parsed_block['attrs']['align'] ?? null;
+		$image_id   = $parsed_block['attrs']['id'] ?? '';
+		$image_size = $parsed_block['attrs']['sizeSlug'] ?? '';
+
+		$image_attributes = wp_get_attachment_image_src( $image_id, $image_size );
```
joemcgill: Instead of calling `wp_get_attachment_image_src()`, we should be able to parse the width attribute directly from the image markup using the existing `WP_HTML_Tag_Processor` instance. 

We should also factor in the `width` attribute from the parsed block, to cover cases where the image has been resized manually in the editor, which is another natural layout constraint.

mukeshpanchal27: @joemcgill, the editor does not add a width attribute in the image markup, so we can't retrieve it through `WP_HTML_Tag_Processor`.

Check the markup:
```
<!-- wp:image {"id":20,"sizeSlug":"large","linkDestination":"none"} -->
<figure class="wp-block-image size-large"><img src="http://localhost:8889/wp-content/uploads/2024/06/car-1024x576.jpeg" alt="" class="wp-image-20"/></figure>
<!-- /wp:image -->
```

In [this commit](https://github.com/WordPress/performance/pull/1252/commits/2656a0d355e8683ddb047cf8f5c5f06d1311aed0), I added support for image resizing (resizing allowed only in the default alignment).

joemcgill (changes requested):
After discussion with @mukeshpanchal27 and reviewing this PR in more detail, I've left the [following comment](https://github.com/WordPress/performance/issues/1251#issuecomment-2161443295) on the original issue with suggestions for a different approach that we should take to resolve this issue.

mukeshpanchal27:
@joemcgill @westonruter The PR is ready for review.

As discussed in [this comment](https://github.com/WordPress/performance/issues/1251#issuecomment-2161443295), I have updated the PR to add custom attributes to the `image` and `cover` blocks. These custom attributes are then removed in the `wp_content_img_tag` filter, which allows us to get both the alignment and image width.

@joemcgill  In [this commit](https://github.com/WordPress/performance/pull/1252/commits/4e56425dce141ae744424c32fb69b929a0f82e19), I have removed the multiple parsing and added an early check to skip if the custom attribute is not present.

plugins/auto-sizes/hooks.php
```diff
@@ -100,16 +119,64 @@ function auto_sizes_render_generator(): void {
...
+ * @return string The updated block content.
+ */
+function auto_sizes_improve_image_sizes_attributes( string $content ): string {
+	$processor = new WP_HTML_Tag_Processor( $content );
+	while ( $processor->next_tag( array( 'tag_name' => 'img' ) ) ) {
+
+		$has_auto_sizes = false;
+		if ( str_contains( $content, 'auto,' ) ) {
+			$has_auto_sizes = true;
+		}
```
joemcgill: Good idea, but we should be more specific here and only check the `sizes` attribute. I wonder if we avoid the need for this check by making sure this filter runs prior to the `auto_sizes_update_content_img_tag()` callback?

```suggestion
		$old_sizes      = $processor->get_attribute( 'sizes' );
		$has_auto_sizes = str_starts_with( $old_sizes, 'auto' );
```

plugins/auto-sizes/hooks.php
```diff
@@ -100,16 +119,64 @@ function auto_sizes_render_generator(): void {
...
+ * Filter the sizes attribute for images to improve the default calculation.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $content The block content about to be rendered.
+ * @return string The updated block content.
+ */
+function auto_sizes_improve_image_sizes_attributes( string $content ): string {
+	$processor = new WP_HTML_Tag_Processor( $content );
+	while ( $processor->next_tag( array( 'tag_name' => 'img' ) ) ) {
```
joemcgill: I think an `if` statement would be better here, rather than a while loop. I wouldn't ever expect the `wp_content_img_tag` filter to pass more than one `img` tag at a time, so we can bail early if `$processor->next_tag( array( 'tag_name' => 'img' )` returns `false`. Then you could return `$content` early if conditions aren't met.

Is there a particular use case you're thinking of here that I may be missing?

plugins/auto-sizes/hooks.php
```diff
@@ -100,16 +119,64 @@ function auto_sizes_render_generator(): void {
...
 	$processor = new WP_HTML_Tag_Processor( $content );
 	$has_image = $processor->next_tag( array( 'tag_name' => 'img' ) );
 
 	// Only update the markup if an image is found.
 	if ( $has_image ) {
+		$align = $parsed_block['attrs']['align'] ?? 'default';
+
+		if ( $align ) {
+			$processor->set_attribute( 'data-align', $align );
+		}
```
westonruter: ~Would `$align` ever be empty due to the `??`? If not, then this is simpler:~

```suggestion
		$align = parsed_block['attrs']['align'] ?? 'default';
		$processor->set_attribute( 'data-align', $align );
```

plugins/auto-sizes/hooks.php
```diff
@@ -100,16 +119,64 @@ function auto_sizes_render_generator(): void {
...
+		// Skips second time parsing if custom IMG attribute removed.
+		$image_alignment = $processor->get_attribute( 'data-align' );
+		if ( ! $image_alignment ) {
+			continue;
+		}
+
 		$layout = wp_get_global_settings( array( 'layout' ) );
-		$align  = $parsed_block['attrs']['align'] ?? null;
-		$sizes  = null;
+		$align  = $processor->get_attribute( 'data-align' );
```
joemcgill: This duplicates the `$image_alignment` variable above and can be removed.

plugins/auto-sizes/hooks.php
westonruter: Alternatively, I see that this `data-align` attribute is serving two purposes: (1) exposing the `align` block attribute to the HTML tag processor and (2) indicating whether the tag has been processed. What if instead the tag processor added a unique attribute like `data-needs-sizes-update` and that this attribute is then looked for when avoiding redundant optimization. So something like this:

```suggestion
		$processor->set_attribute( 'data-needs-sizes-update', true );

		$align = $parsed_block['attrs']['align'] ?? '';
		if ( $align ) {
			$processor->set_attribute( 'data-align', $align );
		}
```

And then see suggestions below.

plugins/auto-sizes/hooks.php
```diff
@@ -100,16 +119,64 @@ function auto_sizes_render_generator(): void {
...
+		$image_width = $processor->get_attribute( 'width' );
+		if ( ! $image_width ) {
+			continue; // Skip if width attribute is not found.
+		}
+
+		// Skips second time parsing if custom IMG attribute removed.
+		$image_alignment = $processor->get_attribute( 'data-align' );
+		if ( ! $image_alignment ) {
+			continue;
+		}
```
westonruter:
```suggestion
		// Skips second time parsing if already processed.
		if ( false === $processor->get_attribute( 'data-needs-sizes-update' ) ) {
			continue;
		}
```

plugins/auto-sizes/hooks.php
```diff
@@ -125,18 +192,23 @@ function auto_sizes_improve_image_sizes_attribute( string $content, array $parse
...
 				break;
 		}
 
 		if ( $sizes ) {
+			$sizes = $has_auto_sizes ? 'auto, ' . $sizes : $sizes;
 			$processor->set_attribute( 'sizes', $sizes );
 		}
 
-		$content = $processor->get_updated_html();
+		$processor->remove_attribute( 'data-align' );
```
westonruter:
```suggestion
		$processor->remove_attribute( 'data-needs-sizes-update' );
		$processor->remove_attribute( 'data-align' );
```

plugins/auto-sizes/hooks.php
```diff
@@ -100,16 +119,64 @@ function auto_sizes_render_generator(): void {
...
+		$has_auto_sizes = false;
+		if ( str_contains( $content, 'auto,' ) ) {
+			$has_auto_sizes = true;
+		}
+
+		// Retrieve width from the image tag itself.
+		$image_width = $processor->get_attribute( 'width' );
+		if ( ! $image_width ) {
+			continue; // Skip if width attribute is not found.
+		}
```
joemcgill: `$image_width` is only a requirement for default aligned images. We could still update full/wide aligned images even if no `width` attribute is present.

plugins/auto-sizes/tests/improve-sizes-test.php
```diff
@@ -24,85 +24,198 @@ public static function set_up_before_class(): void {
 		switch_theme( 'twentytwentyfour' );
 
 		self::$image_id = self::factory()->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/data/images/leaves.jpg' );
+
+		// Disable lazy loading attribute.
+		add_filter( 'wp_img_tag_add_loading_attr', '__return_false' );
```
joemcgill: This is a risky filter to disable while running these tests, because we'll be missing scenarios where the loading attribute changes the way the markup gets created. I assume you did this to avoid having auto sizes added while running these tests? If so, instead of removing the `wp_img_tag_add_loading_attr` filter, I would suggest running these without the auto-sizes filter, which seems to be what is causing potential failures here.
```suggestion
		// Disable auto sizes in these tests.
		remove_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
```

However, since the hooks global won't get reset after this class runs, you'll also need to add this hook back after the class runs, or else it will affect any tests that run afterwards. Ex:

```php
	public static function tear_down_after_class(): void {
		parent::tear_down_after_class();

		add_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
	}
```

If there are specific tests where you need to disable lazy-loading, it would be better to modify hooks directly in that test, because then they will automatically get reset during the `WP_UnitTestCase_Base::tear_down()` method that runs after each test.

plugins/auto-sizes/tests/improve-sizes-test.php
```diff
@@ -24,85 +24,198 @@ public static function set_up_before_class(): void {
...
+				'medium',
+			),
+			'Return wideSize 1280px instead of large size 1024px'  => array(
+				'large',
+			),
+			'Return wideSize 1280px instead of full size 1080px'  => array(
+				'full',
 			),
 		);
+	}
```
joemcgill: This data provider is essentially returning the same values as `data_image_sizes()`. Do we need both?

plugins/auto-sizes/tests/improve-sizes-test.php
```diff
@@ -24,85 +24,198 @@ public static function set_up_before_class(): void {
...
+	/**
+	 * Test the image block with different image sizes and default alignment (contentSize).
+	 *
+	 * @dataProvider data_image_sizes_for_default_alignment
+	 *
+	 * @param string $image_size Image size.
+	 * @param string $expected   Expected output.
+	 * @param string $is_resize  Whether resize or not.
+	 */
+	public function test_image_block_with_default_alignment( string $image_size, string $expected, string $is_resize = '' ): void {
```
joemcgill: I'd pass a `bool` to `$is_resize`, rather than a string. Will need to update the data provider cases to pass `true` instead of 'yes'.

joemcgill (changes requested):
I've tested locally, and this seems to be working as expected for most of the use cases we're trying to solve with this PR. I've left some code quality feedback that would be good to address before merging.

plugins/auto-sizes/hooks.php
mukeshpanchal27: PR updated with the changes.

plugins/auto-sizes/hooks.php
mukeshpanchal27: Make sense. I have added higher priority for sizes filter and remove these code as now it not needed

plugins/auto-sizes/tests/improve-sizes-test.php
mukeshpanchal27: Thanks @joemcgill for suggestion. This will not working as the filter is re-added through `the_content` filter so it show error in tests.

```
1) Tests_Improve_Sizes::test_image_block_with_full_alignment with data set "Return wideSize 1280px instead of medium size 300px" ('medium')
Failed asserting that '<figure class="wp-block-image size-medium"><img loading="lazy" decoding="async" width="300" height="200"   src="http://localhost:8889/wp-content/uploads/2024/06/leaves-283-300x200.jpg" alt="" class="wp-image-5" srcset="http://localhost:8889/wp-content/uploads/2024/06/leaves-283-300x200.jpg 300w, http://localhost:8889/wp-content/uploads/2024/06/leaves-283-1024x683.jpg 1024w, http://localhost:8889/wp-content/uploads/2024/06/leaves-283-768x512.jpg 768w, http://localhost:8889/wp-content/uploads/2024/06/leaves-283.jpg 1080w" sizes="auto, 100vw" /></figure>' contains "sizes="100vw" ".

/var/www/html/wp-content/plugins/performance/plugins/auto-sizes/tests/test-improve-sizes.php:53

2) Tests_Improve_Sizes::test_image_block_with_full_alignment with data set "Return wideSize 1280px instead of large size 1024px" ('large')
Failed asserting that '<figure class="wp-block-image size-large"><img loading="lazy" decoding="async" width="1024" height="683"   src="http://localhost:8889/wp-content/uploads/2024/06/leaves-283-1024x683.jpg" alt="" class="wp-image-5" srcset="http://localhost:8889/wp-content/uploads/2024/06/leaves-283-1024x683.jpg 1024w, http://localhost:8889/wp-content/uploads/2024/06/leaves-283-300x200.jpg 300w, http://localhost:8889/wp-content/uploads/2024/06/leaves-283-768x512.jpg 768w, http://localhost:8889/wp-content/uploads/2024/06/leaves-283.jpg 1080w" sizes="auto, 100vw" /></figure>' contains "sizes="100vw" ".

/var/www/html/wp-content/plugins/performance/plugins/auto-sizes/tests/test-improve-sizes.php:53

3) Tests_Improve_Sizes::test_image_block_with_full_alignment with data set "Return wideSize 1280px instead of full size 1080px" ('full')
Failed asserting that '<figure class="wp-block-image size-full"><img loading="lazy" decoding="async" width="1080" height="720"   src="http://localhost:8889/wp-content/uploads/2024/06/leaves-283.jpg" alt="" class="wp-image-5" srcset="http://localhost:8889/wp-content/uploads/2024/06/leaves-283.jpg 1080w, http://localhost:8889/wp-content/uploads/2024/06/leaves-283-300x200.jpg 300w, http://localhost:8889/wp-content/uploads/2024/06/leaves-283-1024x683.jpg 1024w, http://localhost:8889/wp-content/uploads/2024/06/leaves-283-768x512.jpg 768w" sizes="auto, 100vw" /></figure>' contains "sizes="100vw" ".

/var/www/html/wp-content/plugins/performance/plugins/auto-sizes/tests/test-improve-sizes.php:53

```

mukeshpanchal27:
@joemcgill @westonruter Thanks for the feedback. I have updated the PR as per your directions. It is now ready for review.

westonruter (approved):
LGTM

plugins/auto-sizes/hooks.php
```diff
@@ -100,43 +119,93 @@ function auto_sizes_render_generator(): void {
...
+		$processor->set_attribute( 'sizes', $sizes );
+	}
+
+	$processor->remove_attribute( 'data-needs-sizes-update' );
+	$processor->remove_attribute( 'data-align' );
+	$processor->remove_attribute( 'data-resize-width' );
+
+	return $processor->get_updated_html();
+}
+// Run filter prior to auto sizes "auto_sizes_update_content_img_tag" filter.
```
joemcgill, line 210: 💖

joemcgill (approved):
This is working as expected. I reverted the changes to the `composer.lock` file that were added because I think they were unintentional and doesn't seem to be related to this change. If that was an incorrect assumption we can update composer dependencies in a follow-up task.

PR #1254, opened 2024-05-28
Fix linting of PHP files in tests/plugins

westonruter:
I discovered in https://github.com/WordPress/performance/pull/73#discussion_r1617965728 that PHPCS wasn't being run on files in `tests/plugins`. This is apparently because of this:

```xml
```

It seems to match not only `./plugins` but also `./tests/plugins`. To address this, when linting a specific plugin we can simply pass in the path to the plugin's tests in addition to the main plugin files. This will ensure that things like the plugin-specific PHPCS configuration, like the text domain, are properly applied in the test files.


composer.json
```diff
@@ -75,13 +75,13 @@
...
       "@lint:webp-uploads"
     ],
-    "lint:auto-sizes": "@lint -- ./plugins/auto-sizes --standard=./plugins/auto-sizes/phpcs.xml.dist",
-    "lint:dominant-color-images": "@lint -- ./plugins/dominant-color-images --standard=./plugins/dominant-color-images/phpcs.xml.dist",
-    "lint:embed-optimizer": "@lint -- ./plugins/embed-optimizer --standard=./plugins/embed-optimizer/phpcs.xml.dist",
-    "lint:optimization-detective": "@lint -- ./plugins/optimization-detective --standard=./plugins/optimization-detective/phpcs.xml.dist",
-    "lint:performance-lab": "@lint -- ./plugins/performance-lab --standard=./plugins/performance-lab/phpcs.xml.dist",
-    "lint:speculation-rules": "@lint -- ./plugins/speculation-rules --standard=./plugins/speculation-rules/phpcs.xml.dist",
-    "lint:webp-uploads": "@lint -- ./plugins/webp-uploads --standard=./plugins/webp-uploads/phpcs.xml.dist",
+    "lint:auto-sizes":             "@lint -- ./plugins/auto-sizes             ./tests/plugins/auto-sizes             --standard=./plugins/auto-sizes/phpcs.xml.dist",
```
mukeshpanchal27, line 78: Does this space needed? Can we exclude it?

adamsilverstein (approved):
thanks!

composer.json
westonruter: No, but it makes it much easier to see.

PR #1264, opened 2024-05-31
Fix lint-staged config for plugins

thelovekesh:
## Summary

See https://github.com/WordPress/performance/pull/1089

## Relevant technical choices

- Move lint-staged config from `package.json` to `lint-staged.config.js`.
- Add dynamic config based on staged files.

### DIFF
--- a/lint-staged.config.js
+++ b/lint-staged.config.js
@@ -1,0 +1,60 @@
+/**
+ * External dependencies
+ */
+const path = require( 'path' );
+const micromatch = require( 'micromatch' );
+
+/**
+ * Internal dependencies
+ */
+const { plugins } = require( './plugins.json' );
+
+/**
+ * Join and escape filenames for shell.
+ *
+ * @param {string[]} files Files to join.
+ *
+ * @return {string} Joined files.
+ */
+const joinFiles = ( files ) => {
+	return files.map( ( file ) => `'${ file }'` ).join( ' ' );
+};
+
+// Get plugin base name to match regex more accurately.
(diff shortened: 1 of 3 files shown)


westonruter (approved):
Thank you!

westonruter:
The milestone check seems to be stuck, as I _did_ add a milestone. I'll merge anyway.

I don't think we should use the `no-milestone` label anymore. We should always assign a milestone so issues/PRs get tracked for releases, but we can just add the `skip-changelog` label to avoid it being added to the changelog.

lint-staged.config.js
```diff
@@ -0,0 +1,58 @@
...
+				files,
+				`**/${ PLUGIN_BASE_NAME }/plugins/${ plugin }/**`,
+				{ dot: true }
+			);
+
+			if ( pluginFiles.length ) {
+				commands.push(
+					`composer lint:${ plugin } ${ joinFiles( pluginFiles ) }`
+				);
+			}
```
westonruter, line 43: I just realized that this should technically also include `**/${ PLUGIN_BASE_NAME }/tests/plugins/${ plugin }/*` at least until #1262 is merged (although hopefully that will be done momentarily and this won't be needed).

lint-staged.config.js
```diff
@@ -0,0 +1,58 @@
...
+		} );
+
+		const otherFiles = micromatch(
+			files,
+			`!**/${ PLUGIN_BASE_NAME }/plugins/**`,
```
westonruter, line 48: Ditto, if the above is done, then this would also omit `!**/${ PLUGIN_BASE_NAME }/tests/plugins/**`

PR #1265, opened 2024-05-31
Add install note after each PerfLab feature plugin in the plugin list table

### LINKED ISSUE #1222: Indicate plugins in list table which are installed by Performance Lab
## Feature Description

Currently when a user activates a feature on the Performance screen, this installs a plugin. The user doesn't know this, however. When they go to their plugins list table and see additional plugins unexpectedly, it's possible they could get alarmed. In #1046 we discussed some possibilities for improving the way that plugins are bundled together under the Performance Lab brand, with one possibility being to add a "Performance Lab" prefix to each plugin. This seems to not be moving forward, however. But here's an alternative: what if as part of the plugin row meta there is a "From Performance Lab" bit that is added like as follows:

image

Clicking on "Performance Lab" could take the user to the Performance screen. This plugin row meta could be added by the Performance Lab plugin when it is active (cf. #1052 which dynamically prefixes feature plugin names with "PL" when the Performance Lab plugin is active).

Since none of the plugin assets (#1136) are displayed on the plugin list table, having this additional metadata may be helpful. Granted, it does currently say "By WordPress Performance Team", but that is not an explicit connection to the Performance Lab plugin.

westonruter:
Fixes #1222

As suggested by @justlevine in https://github.com/WordPress/performance/issues/1222#issuecomment-2126933903:

> Has adopting the design language from the `Plugin Dependencies` feature already been discussed and discarded?
> 
> A standard _`Note: This plugin was installed by...`_ that's _below_ the plugin meta, similar in style to the new `Note: this plugin cannot be activated or deactivated...`, would provide a straightforward solve without worrying about prefixes, slug lengths, metadata density, or over/underengineering

Here's what that looks like in WP 6.5+:

Before | After
--|--
Screenshot 2024-05-31 12 19 31 | Screenshot 2024-05-31 12 19 46

The Performance Lab link goes to the Performance settings screen.

Note that I chose to say "This plugin is installed" as opposed to "This plugin was installed" because we don't keep track of whether or not a given feature plugin was indeed installed by the Performance Lab plugin. So it is intended to say that the Performance Lab _in general_ installs a given feature plugin, but that's not to say it couldn't be installed directly.

### DIFF
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -485,7 +485,7 @@ function perflab_print_row_meta_install_notice( string $plugin_file ): void {
 	}
 
 	$message = sprintf(
-		/* translators: placeholder is link to Performance Lab screen */
+		/* translators: %s: link to Performance Lab settings screen */
 		__( 'This plugin is installed by %s.', 'performance-lab' ),
 		sprintf(
 			'<a href="%s">%s</a>',
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -487,7 +487,7 @@ function perflab_print_row_meta_install_notice( string $plugin_file ): void {
 	$message = sprintf(
 		/* translators: %s: link to Performance Lab settings screen */
 		__( 'This plugin is installed by <a href="%s">Performance Lab</a>', 'performance-lab' ),
-		esc_url( add_query_arg( 'page', PERFLAB_SCREEN, admin_url( 'options-general.php' )
+		esc_url( add_query_arg( 'page', PERFLAB_SCREEN, admin_url( 'options-general.php' ) ) )
 	);
 
 	printf(
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -471,3 +471,28 @@ function perflab_get_plugin_settings_url( string $plugin_slug ): ?string {
 
 	return null;
 }
+
+/**
+ * Prints the Performance Lab install notice after each feature plugin's row meta.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $plugin_file Plugin file.
+ */
+function perflab_print_row_meta_install_notice( string $plugin_file ): void {
+	if ( ! in_array( strtok( $plugin_file, '/' ), perflab_get_standalone_plugins(), true ) ) {
+		return;
+	}
+
+	$message = sprintf(
+		/* translators: placeholder is URL to Performance Lab screen */
+		__( 'This plugin is installed by <a href="%s">Performance Lab</a>.', 'performance-lab' ),
+		esc_url( add_query_arg( 'page', PERFLAB_SCREEN, admin_url( 'options-general.php' ) ) )
+	);
+
+	printf(
+		'<div class="requires"><p>%1$s</p></div>',
+		wp_kses( $message, array( 'a' => array( 'href' => array() ) ) )
+	);
+}
+add_action( 'after_plugin_row_meta', 'perflab_print_row_meta_install_notice' );
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -486,12 +486,8 @@ function perflab_print_row_meta_install_notice( string $plugin_file ): void {
 
 	$message = sprintf(
 		/* translators: %s: link to Performance Lab settings screen */
-		__( 'This plugin is installed by %s.', 'performance-lab' ),
-		sprintf(
-			'<a href="%s">%s</a>',
-			esc_url( add_query_arg( 'page', PERFLAB_SCREEN, admin_url( 'options-general.php' ) ) ),
-			__( 'Performance Lab', 'performance-lab' )
-		)
+		__( 'This plugin is installed by <a href="%s">Performance Lab</a>', 'performance-lab' ),
+		esc_url( add_query_arg( 'page', PERFLAB_SCREEN, admin_url( 'options-general.php' )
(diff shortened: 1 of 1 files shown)


plugins/performance-lab/includes/admin/load.php
```diff
@@ -471,3 +471,32 @@ function perflab_get_plugin_settings_url( string $plugin_slug ): ?string {
...
+ *
+ * @param string $plugin_file Plugin file.
+ */
+function perflab_print_row_meta_install_notice( string $plugin_file ): void {
+	if ( ! in_array( strtok( $plugin_file, '/' ), perflab_get_standalone_plugins(), true ) ) {
+		return;
+	}
+
+	$message = sprintf(
+		/* translators: placeholder is link to Performance Lab screen */
```
swissspidy:
```suggestion
		/* translators: %s: link to Performance Lab settings screen */
```

plugins/performance-lab/includes/admin/load.php
```diff
@@ -471,3 +471,32 @@ function perflab_get_plugin_settings_url( string $plugin_slug ): ?string {
...
+		__( 'This plugin is installed by %s.', 'performance-lab' ),
+		sprintf(
+			'<a href="%s">%s</a>',
+			esc_url( add_query_arg( 'page', PERFLAB_SCREEN, admin_url( 'options-general.php' ) ) ),
+			__( 'Performance Lab', 'performance-lab' )
+		)
```
swissspidy: An alternative:

```suggestion
		/* translators: %s: link to Performance Lab settings screen */
		__( 'This plugin is installed by <a href="%s">Performance Lab</a>', 'performance-lab' ),
		esc_url( add_query_arg( 'page', PERFLAB_SCREEN, admin_url( 'options-general.php' ) ) ),
```

westonruter: I had done this originally but then I refactored it to extract the plugin name into a separate string since it is already a standalone translation string. But you know best

swissspidy: The effort required for translators is the same (1 new string), plus plugin names are usually not translated.
I‘d prefer this version here

westonruter: Ok cool. Could you re-suggest since there is a merge conflict now?

plugins/performance-lab/includes/admin/load.php
```diff
@@ -471,3 +471,32 @@ function perflab_get_plugin_settings_url( string $plugin_slug ): ?string {
...
+	}
+
+	$message = sprintf(
+		/* translators: %s: link to Performance Lab settings screen */
+		__( 'This plugin is installed by %s.', 'performance-lab' ),
+		sprintf(
+			'<a href="%s">%s</a>',
+			esc_url( add_query_arg( 'page', PERFLAB_SCREEN, admin_url( 'options-general.php' ) ) ),
+			__( 'Performance Lab', 'performance-lab' )
+		)
```
swissspidy:
```suggestion
		/* translators: %s: link to Performance Lab settings screen */
		__( 'This plugin is installed by <a href="%s">Performance Lab</a>', 'performance-lab' ),
		esc_url( add_query_arg( 'page', PERFLAB_SCREEN, admin_url( 'options-general.php' )
```

plugins/performance-lab/includes/admin/load.php
```diff
@@ -471,3 +471,28 @@ function perflab_get_plugin_settings_url( string $plugin_slug ): ?string {
...
+ */
+function perflab_print_row_meta_install_notice( string $plugin_file ): void {
+	if ( ! in_array( strtok( $plugin_file, '/' ), perflab_get_standalone_plugins(), true ) ) {
+		return;
+	}
+
+	$message = sprintf(
+		/* translators: %s: link to Performance Lab settings screen */
+		__( 'This plugin is installed by <a href="%s">Performance Lab</a>', 'performance-lab' ),
+		esc_url( add_query_arg( 'page', PERFLAB_SCREEN, admin_url( 'options-general.php' )
```
mukeshpanchal27:
```suggestion
		esc_url( add_query_arg( 'page', PERFLAB_SCREEN, admin_url( 'options-general.php' ) ) )
```

plugins/performance-lab/includes/admin/load.php
```diff
@@ -471,3 +471,28 @@ function perflab_get_plugin_settings_url( string $plugin_slug ): ?string {
...
+	}
+
+	$message = sprintf(
+		/* translators: %s: link to Performance Lab settings screen */
+		__( 'This plugin is installed by <a href="%s">Performance Lab</a>', 'performance-lab' ),
+		esc_url( add_query_arg( 'page', PERFLAB_SCREEN, admin_url( 'options-general.php' ) ) )
+	);
+
+	printf(
+		'<div class="requires"><p>%1$s</p></div>',
```
mukeshpanchal27, line 494:
```suggestion
		'<div class="requires">%s</div>',
```

The `<p>` tag added bottom margin that is not needed.

mukeshpanchal27 (approved):
Thanks for the PR. The code shows error so i have push https://github.com/WordPress/performance/pull/1265/commits/5019dd516fae24be424c85927037b3bcaa77883f that fixed the issue.

ThierryA (approved):

plugins/performance-lab/includes/admin/load.php
westonruter: @mukeshpanchal27 That's true, but core is [using](https://github.com/WordPress/wordpress-develop/blob/27151ba0b9693f95717785c164e2f30e64a92627/src/wp-admin/includes/class-wp-plugins-list-table.php#L1596-L1600) a paragraph in such cases:

image

So it seems best to use it here for consistency.

PR #1267, opened 2024-05-31
Try excluding unstaged files from phpstan in lint-staged

westonruter:
Similarly to #1266, I found that phpstan is checking unstaged files. I had an idea to work around this by passing _all_ of the files tracked by git to `composer phpstan`. This works, but there is a big down side: the PHPStan result cache is not used when passing individual files to PHPStan. 

Maybe there is a better way to exclude paths from being checked.

Use case:

Sometimes I have stray PHP files not yet committed which aren't ready, work-in-progress PHP files which I haven't committed yet which stick around as I switch branches. These should not be checked by lint-staged.

### DIFF
--- a/lint-staged.config.js
+++ b/lint-staged.config.js
@@ -3,6 +3,7 @@
  */
 const path = require( 'path' );
 const micromatch = require( 'micromatch' );
+const { execSync } = require( 'child_process' );
 
 /**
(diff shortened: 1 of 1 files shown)


westonruter:
Nevertheless, this probably isn't a common enough of an issue to warrant fixing.

PR #1269, opened 2024-06-03
Remove obsolete fallback script now that picture element is supported

### LINKED ISSUE #996: Fallback script obsolete - remove it

Anuj-Rathore24:
## Summary
Remove Fallback Script which was used to replace webp images with jpeg alternatives on older browsers.
Fixes #996

## Relevant technical choices
- Remove the fallback File.
- Remove Function webpUploadsDetectFallback and the action associate with this.


westonruter:
Looks like this test case needs to be removed as well: `Test_WebP_Uploads_Load::test_it_should_add_fallback_script_if_content_has_updated_images`

westonruter (approved):
Thank you!

PR #1272, opened 2024-06-04
Update plugins with new banners and icons

### LINKED ISSUE #1136: Update standalone plugin assets
## Bug Description

Following #1101, the icon and banner for some feature plugins no longer match the names. Specifically, "WebP Uploads" was renamed "Modern Image Formats" and "Dominant Color Images" was renamed "Image Placeholders".

## For Discussion
We could create new assets using the previous practice of using the first two letters of the plugin name, but this also could be a good time to rethink the design strategy for these assets. To consider:

- Should we use icons that are not "alpha" based at all, so that changing a feature name doesn't require new assets?
- Should we keep using the current "alpha" strategy, but make them based on feature initials, rather than first two letters (e.g., Auto-Sizes becomes "AS" instead of "AU"?
- Should we use a generic icon that can be applied to all features so we don't have to create assets each time we want to create a new feature plugin?

## Steps to reproduce

Visit the WP.org plugin pages:
- https://wordpress.org/plugins/dominant-color-images/
- https://wordpress.org/plugins/webp-uploads/

## Screenshots

image

image

westonruter:
Fixes #1136

All PNG assets have been run through Squoosh, and all SVG assets have been run through SVGOMG.


joemcgill (approved):
These are looking good 👍🏻

adamsilverstein (commented):
Thanks!

PR #1273, opened 2024-06-04
Improve Settings->Media controls for Modern Image Formats

### LINKED ISSUE #1258: Improve Settings for Modern Image Formats
## Feature Description

Since adding AVIF and `<picture>` element support, some potential improvements to the settings screen became apparent.

**Potential Improvements**

* As @joemcgill mentioned in https://github.com/WordPress/performance/pull/73#discussion_r1619478285 we should properly handle sites that lack both AVIF and WebP support (~2% of sites).
* `<picture>` element support is only useful when there is both a modern format (WebP/AVIF) and a JPEG fallback available for each sub size, eg. when the "Also output JPEG images" setting is enabled. Therefore, maybe the picture element support toggle should **only be enabled when JPEGs are being generated?** (and otherwise displayed in a greyed out "disabled" state)

adamsilverstein:
## Summary

* Improve settings when WebP support missing, or no when modern image support is available
* Disable picture element form/description (at load) when JPEG output disabled
* Dynamically toggle picture support form when JPEG output checkbox is toggled
* Add a webp_uploads_is_generate_webp_and_jpeg_enabled helper

Fixes #1258


adamsilverstein:
Adding some screenshots and a screencast:

No modern image formats supported:
no format supported

only avif supported:
only avif available

only WebP supported:
only webp available

screencast - toggling the JPEG output settings on and off
settings as gif

plugins/webp-uploads/helper.php
```diff
@@ -395,3 +395,14 @@ function webp_uploads_sanitize_image_format( string $image_format ): string {
...
 }
+
+/**
+ * Checks if the `perflab_generate_webp_and_jpeg` option is enabled.
+ *
+ * @since n.e.x.t
+ *
+ * @return bool True if the option is enabled, false otherwise.
+ */
+function webp_uploads_is_generate_webp_and_jpeg_enabled(): bool {
```
westonruter: Maybe this should be named to take into account AVIF now? Maybe like:

```suggestion
function webp_uploads_is_jpeg_fallback_enabled(): bool {
```

plugins/webp-uploads/settings.php
```diff
@@ -113,13 +118,29 @@ function webp_uploads_generate_avif_webp_setting_callback(): void {
...
 	$avif_supported = webp_uploads_mime_type_supported( 'image/avif' );
-	// Ensure WebP selected if AVIF is not supported.
-	if ( ! $avif_supported ) {
+	$webp_supported = webp_uploads_mime_type_supported( 'image/webp' );
+
+	// If neither format is support, the entire field is not shown.
+	if ( ! $avif_supported && ! $webp_supported ) {
+		?>
+		<br />
+		<div class="notice notice-warning is-dismissible inline">
```
westonruter: Probably shouldn't be dismissible?

```suggestion
		<div class="notice notice-warning inline">
```

plugins/webp-uploads/settings.php
```diff
@@ -113,13 +118,29 @@ function webp_uploads_generate_avif_webp_setting_callback(): void {
...
+			<p><b><?php esc_html_e( 'Modern Image support is not available.', 'webp-uploads' ); ?></b></p>
+			<p><?php esc_html_e( 'WebP of AVIF support can only be enabled by your hosting provider, so contact them for more information.', 'webp-uploads' ); ?></p>
```
```suggestion
			<p><?php esc_html_e( 'WebP or AVIF support can only be enabled by your hosting provider, so contact them for more information.', 'webp-uploads' ); ?></p>
```

plugins/webp-uploads/settings.php
```diff
@@ -113,13 +118,29 @@ function webp_uploads_generate_avif_webp_setting_callback(): void {
...
+		</div>
+		<?php
+		return;
+	}
+
+	// If only one of the two formats is supported, the dropdown defaults to that type and the other type is disabled.
+	if ( $avif_supported && ! $webp_supported ) {
+		$selected = 'avif';
+	} elseif ( ! $avif_supported && $webp_supported ) {
```
westonruter: Since at this point either `$avif_supported` or `$webp_supported` will always be true, could this be simplified to:

```suggestion
	// If only one of the two formats is supported, the dropdown defaults to that type and the other type is disabled.
	if ( $avif_supported ) {
		$selected = 'avif';
	} else {
```

plugins/webp-uploads/settings.php
```diff
@@ -133,6 +154,13 @@ function webp_uploads_generate_avif_webp_setting_callback(): void {
 			<p><?php esc_html_e( 'AVIF support can only be enabled by your hosting provider, so contact them for more information.', 'webp-uploads' ); ?></p>
 		</div>
 	<?php endif; ?>
+	<?php if ( ! $webp_supported ) : ?>
+		<br />
+		<div class="notice notice-warning is-dismissible inline">
```
```suggestion
		<div class="notice notice-warning inline">
```

plugins/webp-uploads/settings.php
```diff
@@ -149,6 +177,30 @@ function webp_uploads_generate_webp_jpeg_setting_callback(): void {
...
 		</label>
 		<p class="description" id="perflab_generate_webp_and_jpeg_description"><?php esc_html_e( 'Enabling JPEG output can improve compatibility, but will increase the filesystem storage use of your images.', 'webp-uploads' ); ?></p>
+		<script>
+			// Listen for clicks on the JPEG output checkbox, enabling/disabling the
+			// picture element checkbox accordingly.
+			document.addEventListener( 'DOMContentLoaded', function() {
+				var jpegCheckbox               = document.getElementById( 'perflab_generate_webp_and_jpeg' );
+				var pictureCheckbox            = document.getElementById( 'webp_uploads_use_picture_element' );
+				var pictureCheckboxLabel       = document.getElementById( 'webp_uploads_use_picture_element_label' );
+				var pictureCheckboxDescription = document.getElementById( 'webp_uploads_use_picture_element_description' );
```
```suggestion
				const jpegCheckbox               = document.getElementById( 'perflab_generate_webp_and_jpeg' );
				const pictureCheckbox            = document.getElementById( 'webp_uploads_use_picture_element' );
				const pictureCheckboxLabel       = document.getElementById( 'webp_uploads_use_picture_element_label' );
				const pictureCheckboxDescription = document.getElementById( 'webp_uploads_use_picture_element_description' );
```

plugins/webp-uploads/settings.php
```diff
@@ -149,6 +177,30 @@ function webp_uploads_generate_webp_jpeg_setting_callback(): void {
...
+				function togglePictureCheckbox() {
+					if ( jpegCheckbox.checked ) {
+						pictureCheckbox.removeAttribute( 'disabled' );
+						pictureCheckboxLabel.classList.remove( 'webp-uploads-disabled' );
+						pictureCheckboxDescription.classList.remove( 'webp-uploads-disabled' );
+					} else {
+						pictureCheckbox.setAttribute( 'disabled', 'disabled' );
+						pictureCheckboxLabel.classList.add( 'webp-uploads-disabled' );
+						pictureCheckboxDescription.classList.add( 'webp-uploads-disabled' );
+					}
```
```suggestion
					pictureCheckbox.disabled = ! jpegCheckbox.checked;
					pictureCheckboxLabel.classList.toggle( 'webp-uploads-disabled', ! jpegCheckbox.checked );
					pictureCheckboxDescription.classList.toggle( 'webp-uploads-disabled', ! jpegCheckbox.checked );
```

plugins/webp-uploads/settings.php
```diff
@@ -158,13 +210,21 @@ function webp_uploads_generate_webp_jpeg_setting_callback(): void {
...
 	?>
-		<label for="webp_uploads_use_picture_element">
-			<input name="webp_uploads_use_picture_element" type="checkbox" id="webp_uploads_use_picture_element" aria-describedby="webp_uploads_use_picture_element_description" value="1"<?php checked( webp_uploads_is_picture_element_enabled() ); ?> />
+		<style>
+			.webp-uploads-disabled {
+				color #a7aaad;
+				opacity: 0.5;
+			}
+		</style>
+		<label for="webp_uploads_use_picture_element" class="<?php echo $jpeg_disabled ? 'webp-uploads-disabled' : ''; ?>" id="webp_uploads_use_picture_element_label">
```
westonruter: My suggestion here is unnecessary, but it looks safer 😄 

```suggestion
		<label for="webp_uploads_use_picture_element" class="<?php echo esc_attr( $jpeg_disabled ? 'webp-uploads-disabled' : '' ); ?>" id="webp_uploads_use_picture_element_label">
```

westonruter:
@adamsilverstein As for disabling the picture element checkbox when the JPEG output checkbox is disabled, what do you think about hiding the picture element row instead? See https://github.com/adamsilverstein/performance/pull/1

plugins/webp-uploads/helper.php
adamsilverstein: sure, that makes sense. we just can't easily change the option name.

plugins/webp-uploads/settings.php
adamsilverstein: right

plugins/webp-uploads/settings.php
adamsilverstein: Does this really need escaping since it always outputs a predictable, untranslated string?

adamsilverstein:
> @adamsilverstein As for disabling the picture element checkbox when the JPEG output checkbox is disabled, what do you think about hiding the picture element row instead? See [adamsilverstein#1](https://github.com/adamsilverstein/performance/pull/1)

I have mixed feelings about this. I get the point about the disabled checked checkbox and not knowing how to disable it. and I like how that version is simpler. Whats lost is the easy discoverability of the picture element feature. If users don't have the checkbox enabled currently, or its a new install where it isn't enabled by default, they won't see the picture element option at all.

Maybe we can hide just the right column (checkbox/label/description) and replace it with a bit of text that says something like "JPEG fallback required to enable picture element" or something like that?

plugins/webp-uploads/settings.php
westonruter: No, it's not necessary. It just "looks" safer, but feel free to ignore.

plugins/webp-uploads/settings.php
```diff
@@ -113,13 +118,29 @@ function webp_uploads_generate_avif_webp_setting_callback(): void {
...
+		return;
+	}
+
+	// If only one of the two formats is supported, the dropdown defaults to that type and the other type is disabled.
+	if ( $avif_supported ) {
+		$selected = 'avif';
+	} else {
 		$selected = 'webp';
 	}
 	?>
```
westonruter: Oops. I made a bad suggestion here. Now the actual `$selected` option from above is never shown in the UI. If you try to change the format then it just goes back to AVIF if that format is supported.

westonruter:
> Maybe we can hide just the right column (checkbox/label/description) and replace it with a bit of text that says something like "JPEG fallback required to enable picture element" or something like that?

I like that idea. Like this? https://github.com/adamsilverstein/performance/pull/1#issuecomment-2148336125

plugins/webp-uploads/settings.php
```diff
@@ -113,13 +118,29 @@ function webp_uploads_generate_avif_webp_setting_callback(): void {
...
+			<p><?php esc_html_e( 'WebP or AVIF support can only be enabled by your hosting provider, so contact them for more information.', 'webp-uploads' ); ?></p>
+		</div>
+		<?php
+		return;
+	}
+
+	// If only one of the two formats is supported, the dropdown defaults to that type and the other type is disabled.
+	if ( $avif_supported ) {
+		$selected = 'avif';
+	} else {
```
westonruter: I guess this is what I should have suggested instead.

```suggestion
	if ( ! $avif_supported && 'avif' === $selected ) {
		$selected = 'webp';
```

plugins/webp-uploads/settings.php
adamsilverstein: once you merge in your other PR I can give it another test.

westonruter: I don't have permission to merge my PR into your fork.

plugins/webp-uploads/helper.php
```diff
@@ -395,3 +395,14 @@ function webp_uploads_sanitize_image_format( string $image_format ): string {
...
+
+/**
+ * Checks if the `perflab_generate_webp_and_jpeg` option is enabled.
+ *
+ * @since n.e.x.t
+ *
+ * @return bool True if the option is enabled, false otherwise.
+ */
+function webp_uploads_is_jpeg_fallback_enabled(): bool {
+	return true === (bool) get_option( 'perflab_generate_webp_and_jpeg' );
```
westonruter: Nit to make this less verbose.

```suggestion
	return (bool) get_option( 'perflab_generate_webp_and_jpeg' );
```

plugins/webp-uploads/settings.php
```diff
@@ -37,7 +37,7 @@ function webp_uploads_register_media_settings_field(): void {
 		'perflab_generate_webp_and_jpeg',
 		array(
 			'type'         => 'boolean',
-			'default'      => false,
+			'default'      => current_theme_supports( 'html5', 'picture' ), // Use picture element by default if the theme declares support for it.
```
```suggestion
			'default'      => current_theme_supports( 'html5', 'picture' ), // If picture element is supported by the theme, default to enabling the JPEG fallback.
```

plugins/webp-uploads/settings.php
```diff
@@ -113,13 +118,27 @@ function webp_uploads_generate_avif_webp_setting_callback(): void {
...
-	// Ensure WebP selected if AVIF is not supported.
-	if ( ! $avif_supported ) {
+	$webp_supported = webp_uploads_mime_type_supported( 'image/webp' );
+
+	// If neither format is support, the entire field is not shown.
+	if ( ! $avif_supported && ! $webp_supported ) {
+		?>
+		<br />
+		<div class="notice notice-warning inline">
+			<p><b><?php esc_html_e( 'Modern Image support is not available.', 'webp-uploads' ); ?></b></p>
```
```suggestion
			<p><b><?php esc_html_e( 'Modern image support is not available.', 'webp-uploads' ); ?></b></p>
```

westonruter (approved):
Just a couple more nits, but I'm pre-approving otherwise!

adamsilverstein:
> Just a couple more nits, but I'm pre-approving otherwise!

Great, thanks for the feedback.

joemcgill (approved):
Untested, but the code looks right to me. Nice update!

PR #1275, opened 2024-06-05
Add `roots/wordpress` for WordPress code symbol discovery by IDEs/editors

thelovekesh:
## Summary

When working with `wp-env`, code editors and IDEs often struggle to locate WordPress code symbols. This PR introduces `roots/wordpress`, which dumps core files into the `vendor/roots/wordpress` directory. This enhancement improves developer experience (DX) by enabling IDEs and editors to resolve function and class symbols more effectively.

## Relevant technical choices

- Add `roots/wordpress` package.
- Add config to add `WordPress` files in `vendor/roots/wordpress`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,6 +44,9 @@
       "php": "7.2"
     }
   },
+  "extra": {
+    "wordpress-install-dir": "vendor/roots/wordpress"
+  },
   "scripts": {
     "format": "phpcbf --report-summary --report-source",
     "format:all": [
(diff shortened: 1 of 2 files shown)


westonruter:
Humm. We already have core files locally, however, when wp-env had started. Personally I symlink `vendor/wp-env/wp-core` to the path returned by `wp-env install-path`. This gets me the the symbols as well as when I jump to a definition for some symbol in core, I can actually modify the core source code and it is the actual code that is being used in the running environment. If we added the core files to `vendor/roots/wordpress` then there would be another copy that isn't doing anything.

swissspidy:
We also already have the WordPress stubs from PHPStan. If we added this then there would be another copy that isn't doing anything.

westonruter:
Maybe later

PR #1287, opened 2024-06-07
I18N: Add context to Modern Image Formats section title

alexclassroom:
## Summary

The plugin name and the Media setting page section name are the same, "Modern Image Formats," so they are the same string, but they should be disambiguated by context.

## Relevant technical choices

We don't translate plugin names, but even if the "Modern Image Formats" string in the Media setting page is the same as the plugin name, I think it should be a localizable setting page section name.

### DIFF
--- a/plugins/webp-uploads/settings.php
+++ b/plugins/webp-uploads/settings.php
@@ -62,7 +62,7 @@ add_action( 'init', 'webp_uploads_register_media_settings_field' );
 function webp_uploads_add_media_settings_fields(): void {
 	add_settings_section(
 		'perflab_modern_image_format_settings',
-		_x( 'Modern Image Formats', 'Setting Page Section Name', 'webp-uploads' ),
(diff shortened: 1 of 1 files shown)


plugins/webp-uploads/settings.php
```diff
@@ -62,7 +62,7 @@ function webp_uploads_register_media_settings_field(): void {
 function webp_uploads_add_media_settings_fields(): void {
 	add_settings_section(
 		'perflab_modern_image_format_settings',
-		__( 'Modern Image Formats', 'webp-uploads' ),
+		_x( 'Modern Image Formats', 'Setting Page Section Name', 'webp-uploads' ),
```
```suggestion
		_x( 'Modern Image Formats', 'settings page section name', 'webp-uploads' ),
```

swissspidy (approved):

mukeshpanchal27 (approved):

westonruter (approved):

PR #1290, opened 2024-06-10
improved image `sizes` for left/right/center alignment

### LINKED ISSUE #1289: improved image `sizes` for left/right/center alignment
## Overview

Follow-up: #1187
This issue cover calculate `sizes` for left/right/center alignment.

## Approach

For initial work:

- Only return a smaller image size if the content or wide size is larger than the image size.
   - If  content size is 620px and content use thumbnail the the size will use thumbnail 150px image size instead of 620px.
- Parse the width value for "px"; for other types( vw, min() ), return the image width according to the default WordPress behaviour.

## Core editor issues:

- The image markup is missing default width and height attributes. These attributes are set when the image is resized.
- The Cover block does not add the image width attribute.
   - When adding a new Cover block, it defaults to using the full image size.
   - When adding an Image block, selecting a thumbnail, and transforming it to a Cover block, the thumbnail image is used instead of the full image. There should be an image size attribute in the Cover block to display this information.

mukeshpanchal27:
## Summary

Fixes #1289


joemcgill (dismissed):
This is currently blocked until [#1251](https://github.com/WordPress/performance/issues/1251) is resolved, then will need to be updated based on the approach taken to fix that issue.

mukeshpanchal27:
@joemcgill The commit https://github.com/WordPress/performance/pull/1290/commits/7cbb7ae6fefad391eecdea881fece8b7710f70be addresses center alignment.

plugins/auto-sizes/hooks.php
```diff
@@ -159,7 +159,7 @@ function auto_sizes_improve_image_sizes_attributes( string $content ): string {
 	}
 
 	// Skips second time parsing if already processed.
-	if ( false === $processor->get_attribute( 'data-needs-sizes-update' ) ) {
+	if ( null === $processor->get_attribute( 'data-needs-sizes-update' ) ) {
```
mukeshpanchal27, line 162: Per [WP_HTML_Tag_Processor::get_attribute](https://developer.wordpress.org/reference/classes/wp_html_tag_processor/get_attribute/) documentation, It return null instead of `false` when element is not presented.

mukeshpanchal27:
@joemcgill @westonruter The PR is ready for review.

Joe, In https://github.com/WordPress/performance/pull/1290/commits/892b0dc7cce7614e9e17ccf23f4488dec6c38359, I used `set_up` and `tear_down` methods to add/remove the `autosizes` filter and removed the individual test filters.

plugins/auto-sizes/hooks.php
westonruter: Yes, good catch.

plugins/auto-sizes/hooks.php
```diff
@@ -186,7 +186,15 @@ function auto_sizes_improve_image_sizes_attributes( string $content ): string {
 			}
 			break;
 
-		// @todo: handle left/right alignments.
+		case 'left':
+		case 'right':
+		case 'center':
+			// Resize image width.
+			$image_width = $processor->get_attribute( 'data-resize-width' ) ?? $image_width;
+			$width       = auto_sizes_get_width( '', (int) $image_width );
```
westonruter: I'm looking at this function:

https://github.com/WordPress/performance/blob/892b0dc7cce7614e9e17ccf23f4488dec6c38359/plugins/auto-sizes/hooks.php#L106-L111

When the first argument doesn't end in `px` it then appends `px`. Therefore, why wouldn't this just be:

```suggestion
			$width       = sprintf( '%dpx', $image_width );
```

plugins/auto-sizes/hooks.php
```diff
@@ -186,7 +186,15 @@ function auto_sizes_improve_image_sizes_attributes( string $content ): string {
...
+			$sizes       = sprintf( '(max-width: %1$s) 100vw, %1$s', $width );
```
westonruter: Alternatively:

```suggestion
			$sizes       = sprintf( '(max-width: %1$dpx) 100vw, %1$dpx', $image_width );
```

plugins/auto-sizes/hooks.php
mukeshpanchal27: Nice catch.

plugins/auto-sizes/tests/test-improve-sizes.php
```diff
@@ -26,6 +26,21 @@ public static function set_up_before_class(): void {
...
+	}
+
+	public function tear_down(): void {
+		// Clean up specific to the tests within this class.
+		parent::tear_down();
+
+		// Enable auto sizes.
+		add_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
+	}
+
```
westonruter: This shouldn't be necessary because WordPress restores all default filters and actions with each test already.

```suggestion
```

westonruter (approved):
LGTM, but with the `tear_down` not being necessary

PR #1297, opened 2024-06-11
Implement speculative loading of the search form

### LINKED ISSUE #1291: Kick off speculative preload/prerender when searching
## Feature Description

See https://x.com/tunetheweb/status/1800104005873655924 from @tunetheweb

westonruter:
## Summary

Fixes #1291

This speeds up searches by doing speculative loading of the search results page when the user has entered a search query and is signaling they are about to perform a search.

Here's a test site with this patched version of the plugin installed: https://speculative-loading-test.instawp.xyz/search/

[Screen recording 2024-06-11 03.22.30.webm](https://github.com/WordPress/performance/assets/134745/6e0539fb-82ee-4786-aa55-4b5454893f6d)

(Note the TTFB of 0 ms. This was recorded on airplane wifi!)

## Relevant technical choices

The Interactivity API is leveraged to listen for events which signal that a pending search has been entered, namely on the `change` event as well as whenever a submit button gets a `focus` event or `pointerover`. Additionally, when hitting <kbd>Enter</kbd> in the search `input` field it will start to speculatively-load at `keydown` which could in theory save a few milliseconds. (Alternatively to this, a debounced `input` event may make more sense.)

The HTML Tag Processor is used to inject the Interactivity API directives on the search form, both in the Search block variant as well as in the classic variant returned by `get_search_form()`. 

It seems the key is how the form `submit` event is handled. If the default form submission handling is done, then the speculative prerender is not reused. However, if the `submit` event handler does `preventDefault` and then manually navigates to the URL that the search submission was going to go to anyway, then it works!

https://github.com/WordPress/performance/blob/d45eb1c9455915e1fa32aeb29da5d7b7679a9453/plugins/speculation-rules/search-form.js#L65-L72

This seems like a hacky workaround for something the browser should be doing automatically for `GET` form submissions.

### DIFF
--- a/plugins/speculation-rules/hooks.php
+++ b/plugins/speculation-rules/hooks.php
@@ -101,6 +101,12 @@ function plsr_filter_searchform( $form ): string {
 		) {
 			$p->set_attribute( "{$directive_prefix}--focus--{$namespace}", "{$namespace}::actions.updateSpeculativeLoadUrl" );
 			$p->set_attribute( "{$directive_prefix}--pointerover--{$namespace}", "{$namespace}::actions.updateSpeculativeLoadUrl" );
+		} elseif (
+			'INPUT' === $p->get_tag()
+			&&
+			's' === $p->get_attribute( 'name' )
+		) {
+			$p->set_attribute( "{$directive_prefix}--keydown--{$namespace}", "{$namespace}::actions.handleInputKeydown" );
 		}
 	}
 
--- a/plugins/speculation-rules/hooks.php
+++ b/plugins/speculation-rules/hooks.php
@@ -86,7 +86,7 @@ function plsr_filter_searchform( $form ): string {
 			if ( ! $p->get_attribute( 'data-wp-context' ) ) {
 				$p->set_attribute( 'data-wp-context', '{}' );
 			}
-			// TODO: Shouldn't we just watch changes to the one context key?
+			// Note: This directive only subscribes to the properties that are accessed during its execution.
 			$p->set_attribute( "data-wp-watch--{$namespace}", "{$namespace}::callbacks.doSpeculativeLoad" );
 			$p->set_attribute( "data-wp-on--submit--{$namespace}", "{$namespace}::actions.handleFormSubmit" );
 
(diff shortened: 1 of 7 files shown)


plugins/speculation-rules/search-form.js
```diff
@@ -0,0 +1,74 @@
...
+			} else {
+				const url = new URL( form.action );
+				url.search = new URLSearchParams( formData ).toString();
+				context.speculativeLoadUrl = url.href;
+			}
+		},
+		handleInputKeydown: ( event ) => {
+			// Eke out a few milliseconds when hitting enter on the input to submit.
+			if ( event.key === 'Enter' ) {
+				actions.updateSpeculativeLoadUrl();
```
westonruter, line 62: An action calling another action. Is this bad?

luisherranz: That's fine. Even expected 🙂

plugins/speculation-rules/hooks.php
```diff
@@ -58,3 +58,75 @@ function plsr_render_generator_meta_tag(): void {
...
+	while ( $p->next_tag() ) {
+		if ( 'FORM' === $p->get_tag() ) {
+			if ( ! $p->get_attribute( 'data-wp-interactive' ) ) {
+				$p->set_attribute( 'data-wp-interactive', $namespace );
+			}
+			// Create context if not already present.
+			// TODO: Should there be namespaced context?
+			if ( ! $p->get_attribute( 'data-wp-context' ) ) {
+				$p->set_attribute( 'data-wp-context', '{}' );
+			}
```
westonruter, line 88: See `wp_interactivity_data_wp_context()` which has a `$store_namepace` argument. The logic in that function could be replicated here. However, it seems this would conflict when there is already a `data-wp-context` attribute present.

luisherranz: I have created a discussion to see how we can solve that problem in a general way for all the directives:
- https://github.com/WordPress/gutenberg/discussions/62720

plugins/speculation-rules/hooks.php
```diff
@@ -58,3 +58,75 @@ function plsr_render_generator_meta_tag(): void {
...
+			// TODO: Shouldn't we just watch changes to the one context key?
```
luisherranz: I believe this is solved, isn't it?

https://wordpress.slack.com/archives/C071CRKGKUP/p1718106180146629

westonruter: Yes, thanks for following up.

```suggestion
			// Note: This directive only subscribes to the properties that are accessed during its execution.
```

phanduynam:

PR #1298, opened 2024-06-11
Extend core's Autoloaded Options Site Health test if present (in WP 6.6)

### LINKED ISSUE #1274: Audit Autoloaded Options Site Health should extend Core's check if available
## Bug Description

A simplified version of the Audit Autoloaded Options Site Health from this plugin has been merged into WP trunk ([see issue](https://core.trac.wordpress.org/ticket/61276)). The Site Health Check in this plugin still works when the plugin is installed on a site running WP nightlies, but instead of extending the check, it replaces it entirely, which means that core filters for things like the `site_status_autoloaded_options_size_limit` no longer work, and instead need to use the plugin specific `perflab_aao_autoloaded_options_limit_size_in_bytes` version.

## Steps to reproduce

1. Upgrade core local development environment using `https://wordpress.org/wordpress-6.6-beta1.zip` or a nightly version.
2. Install and activate the Performance Lab plugin
3. Try to filter the autoloaded options size limit, like `add_filter( 'site_status_autoloaded_options_size_limit', '__return_zero' );`
4. Observe that no issues are visible from the Site Health screen
5. Deactivate the plugin with the same filter still registered
6. Observe that a critical issue is now visible

## Suggested approach

The Site Health check in this plugin should still support all supported versions of WordPress (currently 6.4+) so we'll need to do a check to see if there is already a `autoloaded_options` check returned from `WP_Site_Health::get_tests()` before adding our own. If it already exists, then our enhancements to the test like the description added via `perflab_aao_get_autoloaded_options_table()` should hook into various filters like the `site_status_autoloaded_options_limit_description` filter.

Optionally, we could already shim in support for anyone using the filters in the 6.6 checks so that they wouldn't have to update their code when upgrading from 6.5 to 6.6 with the Performance Lab plugin running.

mukeshpanchal27:
## Summary

Fixes #1274

### DIFF
--- a/plugins/performance-lab/includes/site-health/audit-autoloaded-options/hooks.php
+++ b/plugins/performance-lab/includes/site-health/audit-autoloaded-options/hooks.php
@@ -19,6 +19,11 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @return array{direct: array<string, array{label: string, test: string}>} Amended tests.
  */
 function perflab_aao_add_autoloaded_options_test( array $tests ): array {
+	// Bail early if check already registered in WordPress core version 6.6.
+	if ( isset( $tests['direct']['autoloaded_options'] ) ) {
+		return $tests;
+	}
+
 	$tests['direct']['autoloaded_options'] = array(
 		'label' => __( 'Autoloaded options', 'performance-lab' ),
 		'test'  => 'perflab_aao_autoloaded_options_test',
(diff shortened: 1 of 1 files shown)


plugins/performance-lab/includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -111,3 +116,15 @@ function perflab_aao_admin_notices(): void {
 	}
 }
 add_action( 'admin_notices', 'perflab_aao_admin_notices' );
+
+/**
+ * Extends the health check description that merged in WordPress 6.6.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $description Description message when autoloaded options bigger than threshold.
```
westonruter, line 125:
```suggestion
 * @param string $description Description message when autoloaded options bigger than threshold.
 * return string Extended health check description.
```

westonruter (approved):
Minor suggestion but otherwise LGTM.

plugins/performance-lab/includes/site-health/audit-autoloaded-options/hooks.php
mukeshpanchal27: Missing `@` before return. In https://github.com/WordPress/performance/pull/1298/commits/e4775cfa8834dd4723d41d52da1d7f573d45b7b2 i added it.

PR #1302, opened 2024-06-13
Leverage Optimization Detective to optimize embeds in Embed Optimizer

### LINKED ISSUE #1054: Integrate Embed Optimizer with Optimization Detective to skip lazy-loading LCP embeds
## Feature Description

The Embed Optimizer (https://github.com/WordPress/performance/issues/888) plugin improves performance for the majority of embeds by lazy-loading them so they don't impede page loading when not in the viewport. However, if they _are_ in the initial viewport the lazy-loading will actually hurt performance by potentially delaying LCP. Any embeds which do not appear in the initial viewport should be excluded from lazy-loading. This can't be done reliably server-side, for example core uses a counter to skip lazy-loading the first N instances. However, Optimization Detective (#869) _is_ built to be able to detect which elements are in the initial viewport so that they can be properly optimized. What is needed is for Optimization Detective to expose an interface for other plugins to hook into the optimization process so that, in the case of Embed Optimizer, it can actually reverse the lazy-loading which it blindly applied server-side for the in-viewport embeds. 

Alternatively, the server-side logic for Embed Optimizer could be modified when Optimization Detective is also running to not immediately make the embeds lazy-loaded, but rather to instead mark the embeds with an attribute for the lazy-loading to be applied during post-processing, that is, on the output-buffered page which the HTML Tag Processor iterates over.

### LINKED ISSUE #1058: Preconnect to domains of resources for LCP elements known to be in embeds when in initial viewport
## Feature Description

When an embed is the LCP element, we can optimize its loading (beyond omitting `loading=lazy` cf. https://github.com/WordPress/performance/issues/1054) by adding preconnect links to domains for resources which are known to be present in the iframe, especially for the LCP element in the iframe. 

# YouTube Embeds

A [YouTube embed](https://www.youtube.com/embed/3KtWfp0UopM?si=8vvS-njzIsPoSc6W) loads resources from the following domains (see on [trace.cafe](https://trace.cafe/t/EVURnSRzyc)):

* `fonts.gstatic.com` (for the fonts used in the title)
* `yt3.ggpht.com` (for the channel logo)
* `i.ytimg.com` (for the poster image)
* `www.youtube.com` (for JS and CSS assets)
* `www.gstatic.com` (for more JS assets)

Note that the actual LCP element and the perceived LCP element on YouTube are not the same due to limitations in the [definition of LCP heuristics](https://web.dev/articles/lcp#:~:text=Elements%20that%20cover%20the%20full%20viewport%2C%20that%20are%20likely%20to%20be%20background%20elements.) (h/t @tunetheweb):

<blockquote>
<p>As well as only considering some elements, LCP measurements use heuristics to
exclude certain elements that users are likely to see as "non-contentful". For
Chromium-based browsers, these include:</p>
<ul>
<li>Elements with an opacity of 0, making them invisible to the user.</li>
<li><mark>Elements that cover the full viewport, that are likely to be background elements.</mark></li>
<li>Placeholder images or other images with a low entropy, that likely don't
reflect the true content of the page.</li>
</ul>
<p>Browsers are likely to continue to improve these heuristics to ensure we match
user expectations of what the largest <em>contentful</em> element is.</p>
</blockquote>

Bullet point two is problematic for YouTube embeds. Chrome is detecting this element to be LCP:

```html
<a class="ytp-title-link yt-uix-sessionlink" target="_blank" data-sessionlink="feature=player-title" tabindex="0" href="https://www.youtube.com/watch?v=3KtWfp0UopM">
    Google — 25 Years in Search: The Most Searched
</a>
```

But the _percieved_ LCP element is the poster image which is background:

```html
<div class="ytp-cued-thumbnail-overlay-image" style="background-image: url(&quot;https://i.ytimg.com/vi_webp/3KtWfp0UopM/maxresdefault.webp&quot;);"></div>
```

However, since it is a background image which covers the full viewport, it is excluded from being considered the LCP element.

youtube-embed

So this introduces a bit of a conundrum. For the sake of YouTube embeds, do we add a preconnect for `fonts.gstatic.com` so the actual LCP element renders sooner, or do we preconnect `i.ytimg.com` so that the perceived LCP element renders sooner? I guess for now the answer is both, at least until browsers improve on these heuristics to match user expectations.

# Other Embeds

A similar analysis should be done for other embeds.

# Implementation

This issue is tagged for both Embed Optimizer and the Optimization Detective plugins. The latter should expose an API so that the former can discover an initial viewport embed and then be able to add the preconnects.

See also https://github.com/ampproject/amp-toolbox-php/issues/10 (from 5+ years ago!) which proposed doing this similarly for AMP components for embeds. Note that AMP components are designed to be able to preconnect to relevant domains when initialized. For example, the `amp-youtube` component [preconnects](https://github.com/ampproject/amphtml/blob/dccbba6692d322f9843c856bd21d9dcc693291a1/extensions/amp-youtube/0.1/amp-youtube.js#L110-L122) to `https://s.ytimg.com` and `https://i.ytimg.com` (the former seems obsolete which I've proposed removing in https://github.com/ampproject/amphtml/pull/39903). Nevertheless, when the embed is in the initial viewport, preconnecting when the component is rendered is almost too late, and the preconnect links should have rather been added to the HTML so they don't depend on JS to execute. This is what the amp-toolbox-php issue was for, and we can do the same for plain HTML.

westonruter:
## Summary

* When an embed appears in the initial viewport, preconnect links are added for the domains for assets known to be loaded when that embed is on the page. Initial support has been added for YouTube, Twitter, and WordPress TV. Fixes #1058.
* When embeds are in the initial viewport they now do _not_ get lazy-loaded. Only embeds which do not appear in any initial viewport will get lazy-loaded. Fixes #1054.

## Relevant technical choices

* The `OD_HTML_Tag_Walker` class has been eliminated. This was a wrapper around the `WP_HTML_Tag_Processor` subclass to ensure that any of the visitors would not call `next_tag` or `seek` and thus cause the processor to lose track of where it was in the document which would cause problems with generating XPaths for the tags. For Image Optimizer, iterating over the document tag-by-tag in a forward direction was sufficient. For Embed Optimizer, however, it was too limiting. Embed Optimizer needs to be able to iterate over the tags in an embed, and then based on what it discovers, go back and apply the lazy-loading optimizations. Extenders who write tag visitors will have the full `WP_HTML_Tag_Processor` API available to them. _Many of the changes you'll see are just swapping out `walker` for `processor`._
* The `OD_HTML_Tag_Processor::append_head_html()` and `OD_HTML_Tag_Processor::append_body_html()` no longer require that the closing `</head>` and `</body>` tag to have first been visited to be successful. The supplied HTML is buffered until the time that `get_updated_html()` is called.
* The existing logic to apply lazy-loading to embeds has been refactored so that it can apply to both the cases in which Optimization Detective is or is not active.
* Embed Optimizer now adds a notice to the plugin row meta to suggest installing Optimization Detective and conversely that Optimization Detective is recommended by Embed Optimizer:

Screenshot 2024-07-09 09 55 32

Screenshot 2024-07-09 09 54 41

Screenshot 2024-07-09 09 56 36

* The `OD_Preload_Link_Collection` class has been made more generic as an `OD_Link_Collection` class, allowing tag visitors to register links of any relation type. For Embed Optimizer, this is leveraged to add `preconnect` links. Support for additional link attributes has also been added.
* Enhanced static analysis with better typing in a few places.
* The PHPUnit bootstrap logic now looks for plugin-specific bootstrap file in `plugins/$slug/tests/bootstrap.php`. This is needed for Embed Optimizer to load Optimization Detective to test its integration.


sstopfer:
@westonruter I added this PR to the project board (In prgress column). Could you move it to CR when ready and tag other for review?

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -0,0 +1,111 @@
...
+				$preconnect_hrefs[] = 'https://syndication.twitter.com';
+				$preconnect_hrefs[] = 'https://pbs.twimg.com';
+			} elseif ( $processor->has_class( 'wp-block-embed-wordpress-tv' ) ) {
+				$preconnect_hrefs[] = 'https://video.wordpress.com';
+				$preconnect_hrefs[] = 'https://public-api.wordpress.com';
+				$preconnect_hrefs[] = 'https://videos.files.wordpress.com';
+			}
+
+			foreach ( $preconnect_hrefs as $preconnect_href ) {
+				$this->link_collection->add_link(
```
westonruter, line 131: It would be slightly better if the viewport widths were also taken into account, so that an embed only visible on desktop would not get preconnected on mobile. This can be addressed in a follow-up PR.

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -0,0 +1,111 @@
...
+				$preconnect_hrefs[] = 'https://www.youtube.com';
+				$preconnect_hrefs[] = 'https://i.ytimg.com';
+			} elseif ( $processor->has_class( 'wp-block-embed-twitter' ) ) {
+				$preconnect_hrefs[] = 'https://syndication.twitter.com';
+				$preconnect_hrefs[] = 'https://pbs.twimg.com';
+			} elseif ( $processor->has_class( 'wp-block-embed-wordpress-tv' ) ) {
+				$preconnect_hrefs[] = 'https://video.wordpress.com';
+				$preconnect_hrefs[] = 'https://public-api.wordpress.com';
+				$preconnect_hrefs[] = 'https://videos.files.wordpress.com';
+			}
```
westonruter: @adamsilverstein Any other embeds you'd recommend for the initial merge of this?

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -0,0 +1,111 @@
...
+			$processor->append_body_html( wp_get_inline_script_tag( embed_optimizer_get_lazy_load_script(), array( 'type' => 'module' ) ) );
+			$this->added_lazy_script = true;
+		}
+
+		// Since there is a limit to the number of bookmarks we can add, make sure any new ones we add get removed.
+		// TODO: Instead of this consider throwing an exception inside embed_optimizer_update_markup() and clear out the bookmarks in the catch() block or else when the function returns.
+		$new_bookmarks = array_diff( $original_bookmarks, $processor->get_bookmark_names() );
+		foreach ( $new_bookmarks as $new_bookmark ) {
+			$processor->release_bookmark( $new_bookmark );
+		}
```
westonruter: It's possible this could clear bookmarks for other tag visitors that possibly run on tags inside of the embed. Therefore, this should only clear the bookmarks that we set.

plugins/embed-optimizer/hooks.php
```diff
@@ -10,6 +10,36 @@
...
+ * Registers the tag visitor for embeds.
+ *
+ * @since n.e.x.t
+ *
+ * @param OD_Tag_Visitor_Registry         $registry                     Tag visitor registry.
+ * @param OD_URL_Metrics_Group_Collection $url_metrics_group_collection URL Metrics Group Collection.
+ * @param OD_Link_Collection              $link_collection              Link Collection.
+ */
+function embed_optimizer_register_tag_visitors( OD_Tag_Visitor_Registry $registry, OD_URL_Metrics_Group_Collection $url_metrics_group_collection, OD_Link_Collection $link_collection ): void {
+	require_once __DIR__ . '/class-embed-optimizer-tag-visitor.php';
```
westonruter, line 40: This `require_once` could go into `load.php`. But there's no need to load it if it won't be used. I'm happy to move it however.

plugins/embed-optimizer/hooks.php
```diff
@@ -158,4 +317,3 @@ function embed_optimizer_render_generator(): void {
 	// Use the plugin slug as it is immutable.
 	echo '<meta name="generator" content="embed-optimizer ' . esc_attr( EMBED_OPTIMIZER_VERSION ) . '">' . "\n";
 }
-add_action( 'wp_head', 'embed_optimizer_render_generator' );
```
westonruter, line 161: Moved to `embed_optimizer_add_hooks()`

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
```diff
@@ -35,11 +35,11 @@ public function __invoke( OD_HTML_Tag_Walker $walker ): bool {
 		 * parser of the `url()` functions from the property value.
 		 */
 		$background_image_url = null;
-		$style                = $walker->get_attribute( 'style' );
+		$style                = $processor->get_attribute( 'style' );
 		if (
 			is_string( $style )
 			&&
-			1 === preg_match( '/background(?:-image)?\s*:[^;]*?url\(\s*[\'"]?\s*(?<background_image>.+?)\s*[\'"]?\s*\)/', $style, $matches ) // Not compatible with strict rules. See <https://github.com/phpstan/phpstan/issues/11262>.
```
westonruter, line 42: This comment was supposed to have been removed in https://github.com/WordPress/performance/pull/1325.

plugins/optimization-detective/class-od-link-collection.php
```diff
@@ -139,15 +177,22 @@ static function ( array $carry, array $link ): array {
 		// Add media attributes to the deduplicated links.
 		return array_map(
 			static function ( array $link ): array {
-				$media_attributes = array( 'screen' );
```
westonruter, line 142: The `screen` media type is no longer considered the default because arbitrary links can now be added, not just preloads. And consider for Embed Optimizer it may be that (hypothetically) an audio-only browser could load a page which includes a YouTube embed. This embed should still get preconnected even if there isn't anything visual on the screen. Granted, this is highly hypothetical 😄

plugins/optimization-detective/class-od-link-collection.php
```diff
@@ -215,9 +260,9 @@ public function get_response_header(): ?string {
 	/**
 	 * Counts the links.
 	 *
-	 * @return int Link count.
+	 * @return int<0, max> Link count.
 	 */
 	public function count(): int {
-		return count( $this->links );
+		return count( $this->links_by_rel );
```
westonruter: Humm. This actually is wrong. It should rather count recursively?

plugins/optimization-detective/tests/test-optimization.php
```diff
@@ -368,44 +368,6 @@ function ( OD_HTML_Tag_Walker $walker, OD_URL_Metrics_Group_Collection $url_metr
...
-				'crossorigin'   => 'anonymous',
-			),
-			600,
-			1200
-		);
-
-		$expected_header = 'Link: <https://example.com/foo.jpg>; rel="preload"; as="image"; fetchpriority="high"; imagesrcset="https://example.com/foo-480w.jpg 480w, https://example.com/foo-800w.jpg 800w"; imagesizes="(max-width: 600px) 480px, 800px"; crossorigin="anonymous"; media="screen", <https://example.com/bar.jpg>; rel="preload"; as="image"; fetchpriority="high"; imagesrcset="https://example.com/\"bar\"-480w.jpg 480w, https://example.com/\"bar\"-800w.jpg 800w"; imagesizes="(max-width: 600px) 480px, 800px"; crossorigin="anonymous"; media="screen and (min-width: 600px) and (max-width: 1200px)"';
-		$this->assertSame( $expected_header, $collection->get_response_header() );
-	}
-
```
westonruter, line 408: Assertions were incorporated into an existing test in `test-class-od-html-tag-processor.php`

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
westonruter: Actually, other tag visitors' bookmarks would not have been introduced here because this tag visitors do not iterate concurrently. In any case, the logic has been made more robust in bc6db155c583f78d37aed611c34bfc0988bb4e40

plugins/optimization-detective/class-od-html-tag-processor.php
```diff
@@ -12,32 +12,563 @@
...
+		}
+
+		if ( $this->previous_tag_without_closer ) {
+			array_pop( $this->open_stack_tags );
+		}
+
+		if ( ! $this->is_tag_closer() ) {
+
+			// Close an open P tag when a P-closing tag is encountered.
+			// TODO: There are quite a few more cases of optional closing tags: https://html.spec.whatwg.org/multipage/syntax.html#optional-tags
```
westonruter, line 268: This will be resolved when this class makes use of `WP_HTML_Processor` instead, as all of this will be handled automatically. But this will be WordPress 6.7.

plugins/optimization-detective/class-od-link-collection.php
```diff
@@ -215,9 +260,16 @@ public function get_response_header(): ?string {
 	/**
 	 * Counts the links.
 	 *
-	 * @return int Link count.
+	 * @return int<0, max> Link count.
 	 */
 	public function count(): int {
-		return count( $this->links );
+		return array_sum(
```
swissspidy, line 281: Use case for `array_reduce` perhaps?

plugins/optimization-detective/class-od-url-metrics-group.php
```diff
@@ -330,7 +330,7 @@ public function getIterator(): ArrayIterator {
 	/**
 	 * Counts the URL metrics in the group.
 	 *
-	 * @return int URL metric count.
+	 * @return int<0, max> URL metric count.
```
swissspidy, line 333: There's also `positive-int` in PHPStan

plugins/optimization-detective/class-od-link-collection.php
```diff
@@ -52,7 +57,15 @@ final class OD_Preload_Link_Collection implements Countable {
...
 	 */
-	public function add_link( array $attributes, ?int $minimum_viewport_width, ?int $maximum_viewport_width ): void {
+	public function add_link( array $attributes, ?int $minimum_viewport_width = null, ?int $maximum_viewport_width = null ): void {
+		if ( ! array_key_exists( 'rel', $attributes ) ) {
+			throw new InvalidArgumentException( esc_html__( 'The rel attribute must be provided.', 'optimization-detective' ) );
+		}
+		if ( 'preload' === $attributes['rel'] && ! array_key_exists( 'as', $attributes ) ) {
+			throw new InvalidArgumentException( esc_html__( 'A preload link must include an as attribute.', 'optimization-detective' ) );
+		} elseif ( 'preconnect' === $attributes['rel'] && ! array_key_exists( 'href', $attributes ) ) {
+			throw new InvalidArgumentException( esc_html__( 'A preconnect link must include an href attribute.', 'optimization-detective' ) );
```
swissspidy: Nit, I would use sprintf and placeholders for `preconnect`, `preload`, `as` and `href` here, so that they aren't translated. Then this becomes the same string for both, which is less work for translators too.

plugins/embed-optimizer/hooks.php
```diff
@@ -23,90 +53,156 @@
...
+						throw new Exception( __( 'Embed Optimizer unable to set iframe bookmark.', 'embed-optimizer' ) );
+					}
+				}
+			} elseif ( 'SCRIPT' === $html_processor->get_tag() ) {
+				if ( ! $html_processor->get_attribute( 'src' ) ) {
+					$has_inline_script = true;
+				} else {
+					++$script_count;
+					if ( ! $html_processor->set_bookmark( $bookmark_names['script'] ) ) {
+						throw new Exception( __( 'Embed Optimizer unable to set script bookmark.', 'embed-optimizer' ) );
```
swissspidy: Use sprintf placeholders for `script` and `iframe` here so that this becomes the same string for both.

plugins/embed-optimizer/hooks.php
```diff
@@ -23,90 +53,156 @@
...
+		// If there was only one non-inline script, make it lazy.
+		if ( 1 === $script_count && ! $has_inline_script && $html_processor->has_bookmark( $bookmark_names['script'] ) ) {
+			$needs_lazy_script = true;
+			if ( $html_processor->seek( $bookmark_names['script'] ) ) {
+				if ( $html_processor->get_attribute( 'type' ) ) {
+					$html_processor->set_attribute( 'data-original-type', $html_processor->get_attribute( 'type' ) );
+				}
+				$html_processor->set_attribute( 'type', 'application/vnd.embed-optimizer.javascript' );
+			} else {
+				wp_trigger_error( __FUNCTION__, esc_html__( 'Embed Optimizer unable to seek to script bookmark.', 'embed-optimizer' ) );
```
swissspidy: Ditto here for using a placeholder for `script`

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -0,0 +1,102 @@
...
+	 * @param OD_URL_Metrics_Group_Collection $url_metrics_group_collection URL Metrics Group Collection.
+	 * @param OD_Link_Collection              $link_collection              Link Collection.
+	 */
+	public function __construct( OD_URL_Metrics_Group_Collection $url_metrics_group_collection, OD_Link_Collection $link_collection ) {
+		$this->url_metrics_group_collection = $url_metrics_group_collection;
+		$this->link_collection              = $link_collection;
+	}
+
+	/**
+	 * Visits a tag.
```
swissspidy, line 55: Missing `@version`

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -0,0 +1,102 @@
...
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Tag visitor that optimizes embeds.
+ *
+ * @since 0.1.0
```
swissspidy:
```suggestion
 * @since n.e.x.t
```

plugins/optimization-detective/class-od-link-collection.php
westonruter: I tried this:

```php
		return array_reduce(
			array_values( $this->links_by_rel ),
			static function ( int $accumulator, array $links ): int {
				return $accumulator + count( $links );
			},
			0
		);
```

But then PHPStan complains that it returns an `int` instead of a `int<0, max>`.

plugins/optimization-detective/class-od-url-metrics-group.php
westonruter: I think that is `int<1, max>`, isn't it?

swissspidy: Right. `non-negative-int` is `0,max`

plugins/optimization-detective/class-od-link-collection.php
westonruter: Done in 7cc5e19a4b7531003fbd416e91771a656b8aec44

plugins/embed-optimizer/hooks.php
westonruter: Done in d8ac531

swissspidy (approved):

plugins/optimization-detective/optimization.php
```diff
@@ -171,38 +169,45 @@ function od_optimize_template_output_buffer( string $buffer ): string {
...
-	do_action( 'od_register_tag_visitors', $tag_visitor_registry, $group_collection, $preload_links );
+	do_action( 'od_register_tag_visitors', $tag_visitor_registry, $group_collection, $link_collection );
+
+	$current_tag_bookmark = 'optimization_detective_current_tag';
 
-	$visitors  = iterator_to_array( $tag_visitor_registry );
-	$generator = $walker->open_tags();
-	while ( $generator->valid() ) {
+	$visitors = iterator_to_array( $tag_visitor_registry );
+	while ( $processor->next_open_tag() ) {
```
westonruter, line 179: Initially I had this as `$processor->next_tag()` which was not correct as it was resulting in visits to closing tags. Then I added a `$processor->is_tag_closer()` check which would `continue` if `true`. But then I remembered this subclass's `next_open_tag()` helper method which "works a treat".

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
westonruter: To be updated in subsequent PR based on the Colab data on WP embeds.

adamsilverstein: The most common 10 embed types used on WordPress sites and the percent of sites with embeds hat use that type:

image

data from httparchive, queries in [colab](https://colab.sandbox.google.com/drive/1nSpg3qoCLY-cBTV2zOUkgUCU7R7X2f_R?resourcekey=0-MgT7Ur0pT__vw-5_AHjgWQ#scrollTo=utZv59sXzXvS)

westonruter: Awesome! What is `embed-handler` and `amazon`?

swissspidy: `amazon` might refer to Amazon Kindle embeds.

`embed-handler` sounds odd though 🤔 Maybe coming from custom embed handlers added via `wp_embed_register_handler` where the developer forgot to change the class name?

adamsilverstein: yes, we support amazon embeds! `embed` is when you embed one WordPress post in another post (example - https://oembeds-test.instawp.xyz/test-wordpress-post/)

adamsilverstein: Full [list of oembeds](https://github.com/WordPress/wordpress-develop/blob/9cff1de6b3adcf8767bfc60c4e71b5cce81f0e1b/src/wp-includes/class-wp-oembed.php#L53) in core.

plugins/embed-optimizer/hooks.php
adamsilverstein: maybe add an inline comment explaining why you are loading it here (to avoid loading it unnecessarily)?

plugins/embed-optimizer/hooks.php
```diff
@@ -23,90 +53,171 @@
...
+		$iframe_count      = 0;
+		$script_count      = 0;
+		$needs_lazy_script = false;
+		$has_inline_script = false;
+		$figure_depth      = 0;
+		// Locate the iframes and scripts.
+		do {
+			// This condition ensures that when iterating over an embed inside a larger document that we stop once we reach
+			// closing </figure> tag. The $processor is an OD_HTML_Tag_Processor when Optimization Detective is iterating
+			// over all tags in the document, and this embed_optimizer_update_markup() is usd as part of the tag visitor
```
adamsilverstein:
```suggestion
			// over all tags in the document, and this embed_optimizer_update_markup() is used as part of the tag visitor
```

plugins/embed-optimizer/hooks.php
```diff
@@ -23,90 +53,171 @@
...
+			// from Embed Optimizer. On the other hand, if $html_processor is not an OD_HTML_Tag_Processor then this is
+			// iterating over the tags of the embed markup alone as is passed into the embed_oembed_html filter.
+			if ( $html_processor instanceof OD_HTML_Tag_Processor ) {
```
adamsilverstein: this feels a bit hacky to determine the context this way. could you add an additional parameter to `embed_optimizer_update_markup` instead (and then use a wrapper for the filter callback). not a blocker.

plugins/embed-optimizer/readme.txt
```diff
@@ -13,6 +13,10 @@ Optimizes the performance of embeds by lazy-loading iframes and scripts.
...
 
+This plugin also recommends that you install and activate the [Optimization Detective](https://wordpress.org/plugins/optimization-detective/) plugin. When it is active, it will start learning about which embeds appear in the initial viewport based on actual visitors to your site. With this information in hand, Embed Optimizer will then avoid lazy-loading embeds which appear in the initial viewport (above the fold). This is important because lazy-loading adds a delay which can hurt the user experience and even degrade the Largest Contentful Paint (LCP) score for the page. In addition to not lazy-loading such above-the-fold embeds, Embed Optimizer will add preconnect links for the hosts of network resources known to be required for embeds (e.g. YouTube, Twitter, and WordPress TV); this can further speed up the loading of critical embeds. Again, these performance enhancements are only enabled when Optimization Detective is active.
```
adamsilverstein:
```suggestion
This plugin also recommends that you install and activate the [Optimization Detective](https://wordpress.org/plugins/optimization-detective/) plugin. When it is active, it will start recording which embeds appear in the initial viewport based on actual visitors to your site. With this information in hand, Embed Optimizer will then avoid lazy-loading embeds which appear in the initial viewport (above the fold). This is important because lazy-loading adds a delay which can hurt the user experience and even degrade the Largest Contentful Paint (LCP) score for the page. In addition to not lazy-loading such above-the-fold embeds, Embed Optimizer will add preconnect links for the hosts of network resources known to be required for embeds (e.g. YouTube, Twitter, and WordPress TV); this can further speed up the loading of critical embeds. Again, these performance enhancements are only enabled when Optimization Detective is active.
```

adamsilverstein: I notice you don't mention breakpoints, are you planning to add that in a future PR?

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
westonruter: Excellent. I've fleshed out the list of embeds to preconnect in https://github.com/WordPress/performance/pull/1302/commits/6c3a9b431b368e51a1bba04da5d11e3741ca6a53. I went down to Soundcloud.

plugins/embed-optimizer/hooks.php
westonruter: Added in e605df59

plugins/embed-optimizer/hooks.php
westonruter: OK, fair enough. Changed in 357364ed773140610c3d86babc4f7b93e551ed06.

adamsilverstein (approved):
Looks good, left some small feedback/questions

plugins/embed-optimizer/readme.txt
westonruter: I believe you're referring to https://github.com/WordPress/performance/pull/1302/files#r1670865833 ?

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
westonruter: See https://github.com/WordPress/performance/issues/1341

PR #1307, opened 2024-06-19
Improve compatibility of styling picture elements

### LINKED ISSUE #1300: Modern Image Formats `picture` support breaks gallery block cropping
## Bug Description

When using the new Modern Image Formats `picture` element support on my personal site, I noticed that the markup conflicted with the style for an image gallery created using the `wp:gallery` block when the "Crop images to fit" setting was enabled for the block.

I've left the feature enabled [on my site](https://joemcgill.net/2024/01/a-2023-retrospective/) for now so the behavior can be observed.

## Steps to reproduce

1. Activate the Modern Image Formats plugin and enable `picture` support from the Media Settings screen.
2. Create an image gallery using the core gallery block that includes images with different aspect ratios
3. Turn on the "Crop images to fit" setting on the Gallery block settings panel (screenshot below)
4. Preview the page in the front end

### Screenshot of the settings panel

image

### Before using `picture`

image

### With `picture` support enabled

image

devansh016:
## Summary
Fixes #1300 

## Relevant technical choices
Added following style to picture tag and updated the test for the same.
```html
style="display: contents;"
```

The `wp:gallery` style was not getting applying due to the `img` getting inside the `picture` tag.

### DIFF
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -146,7 +146,7 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	remove_filter( 'wp_calculate_image_srcset_meta', $filter );
 
 	return sprintf(
-		'<picture class="%s">%s%s</picture>',
+		'<picture class="%s" style="display: contents;">%s%s</picture>',
 		esc_attr( 'wp-picture-' . $attachment_id ),
 		$picture_sources,
 		$original_image_without_srcset
--- a/plugins/webp-uploads/tests/test-picture-element.php
+++ b/plugins/webp-uploads/tests/test-picture-element.php
@@ -43,7 +43,7 @@ class Test_WebP_Uploads_Picture_Element extends TestCase {
 		$the_image = apply_filters( 'wp_content_img_tag', $the_image, 'the_content', $attachment_id );
 
 		// Check that the image is wrapped in a picture element with the correct class.
-		$picture_element = sprintf( '<picture class="wp-picture-%s">', $attachment_id );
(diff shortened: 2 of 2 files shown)


plugins/webp-uploads/picture-element.php
```diff
@@ -146,7 +146,7 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	remove_filter( 'wp_calculate_image_srcset_meta', $filter );
 
 	return sprintf(
-		'<picture class="%s">%s%s</picture>',
+		'<picture class="%s" style="display: contents;">%s%s</picture>',
```
mukeshpanchal27, line 149: Instead of adding style in all picture elements can we add custom CSS for element for gallery block?

something like: `.block picture { display: contents; }`

devansh016: Thanks @mukeshpanchal27. Can you please provide more context for making these changes to `wp:gallery` and not here? Any specific reason?

westonruter: I believe it makes sense to apply this globally to all `picture` elements specifically generated by this plugin, as done here, as it will essentially ignore the `picture` for styling purposes. This is what we want because we only care about the `img`.

westonruter (approved):

adamsilverstein (approved):

plugins/webp-uploads/picture-element.php
mukeshpanchal27: Gotcha!

mukeshpanchal27 (approved):

PR #1314, opened 2024-06-23
Add error for incorrect test command usage

### LINKED ISSUE #1256: `composer test:plugins` failed with error
## Bug Description

Run `composer test:plugins` command.

```
> phpunit '--verbose' '--testsuite' 'auto-sizes'
Error: The following required constants are not defined: WP_TESTS_DOMAIN, WP_TESTS_EMAIL, WP_TESTS_TITLE, WP_PHP_BINARY.
Please check out `wp-tests-config-sample.php` for an example.
Script phpunit handling the test event returned with error code 1
Script @test --verbose --testsuite auto-sizes was called via test:auto-sizes
Script @test:auto-sizes was called via test:plugins
```

## Screenshots

Screenshot 2024-05-29 at 10 57 27 AM

AhmarZaidi:
## Summary

Fixes #1256

This PR addresses an issue where running Composer test commands directly shows error related to missing constants but doesn't show that the user should run the npm command to run tests.

## Relevant technical choices

Added missing environment variables check and file exists check for php unit test `functions.php` in the test `bootstrap.php`.

Now, running `composer test:plugins` command directly:
```bash
> phpunit '--verbose' '--testsuite' 'auto-sizes'
Please use 'npm run test-php' instead of running the composer test command directly.
If you are intending to run PHPUnit directly, the following environment variables are missing: WORDPRESS_DB_PASSWORD, WORDPRESS_DB_HOST, WORDPRESS_DB_USER, WORDPRESS_DB_NAME, PHP_VERSION, WP_TESTS_DIR
Script phpunit handling the test event returned with error code 1
Script @test --verbose --testsuite auto-sizes was called via test:auto-sizes
Script @test:auto-sizes was called via test:plugins
```


AhmarZaidi:
Please add relevant [Type] label and milestone to this PR.

pretest.php
```diff
@@ -0,0 +1,12 @@
+<?php
+/**
+ * This script runs before the test command is executed.
+ *
+ * @package performance
+ */
+
+// On local environment.
+if ( ! getenv( 'WORDPRESS_DB_USER' ) ) {
```
westonruter: Or what about checking for all of the required environment variables, and if they aren't all set, then the error message could mention which one(s) are missing in case someone is wanting to run the tests from their machine without Docker. For example, the error message could be like:

> Please use 'npm run test-php' instead of running the composer test command directly.
>
> If you are intending to run PHPUnit directly, the following environment variables are missing: WORDPRESS_DB_USER, WORDPRESS_DB_PASSWORD.

Or alternatively, maybe it would make more sense to put this error message in the [`bootstrap.php`](https://github.com/WordPress/performance/blob/trunk/tests/bootstrap.php) instead of introducing this `pretest.php` file?

For example, before this line:

https://github.com/WordPress/performance/blob/2b2bebcb9a6285fac8653d078c1d49588f3c6b70/tests/bootstrap.php#L35

It could do:

```php
if ( ! file_exists( $_test_root . '/includes/functions.php' ) ) {
    echo "It seems your environment is not set up for running tests. (Unable to determine the test root.) Please use 'npm run test-php' instead.\n";
    exit( 1 );
}
```

AhmarZaidi: Hey, thanks for the input.

The first method seems easy enough, I can check for all required environment variables and show the missing error accordingly in `pretest.php`.

But I think the second approach would be better since it would integrate in the existing code without creating a new file/command.

I noticed that `file_exists( $_test_root . '/includes/functions.php' )` is true even on local, so I was thinking of adding all required environment variables check in the `bootstrap.php`.

Something like this:
```php
$required_constants = array(
    // required env variables.
);

$missing_constants = array();

foreach ( $required_constants as $constant ) {
    // Check if the environment variable is not set
    // if not add it to missing_constants array
}

if ( $missing_constants ) {
    // If there are missing constants print error
    exit( 1 );
}
```

The the error would be like this:
```bash
> phpunit '--verbose' '--testsuite' 'auto-sizes'
Please use 'npm run test-php' instead of running the composer test command directly.
If you are intending to run PHPUnit directly, the following environment variables are missing: WORDPRESS_DB_PASSWORD, WORDPRESS_DB_HOST, WORDPRESS_DB_USER, WORDPRESS_DB_NAME, PHP_VERSION, WP_TESTS_DIR
Script phpunit handling the test event returned with error code 1
Script @test --verbose --testsuite auto-sizes was called via test:auto-sizes
Script @test:auto-sizes was called via test:plugins
```

westonruter: It looks like on local I have the `WP_PHPUNIT__DIR` environment variable being set by Composer. This is due to the `wp-phpunit/wp-phpunit` package which has a `vendor/wp-phpunit/wp-phpunit/__loaded.php` file that is doing:

```php
putenv( sprintf( 'WP_PHPUNIT__DIR=%s', dirname( __FILE__ ) ) );
```

So yeah, that `file_exists()` check won't work _if_ a user had first done `composer install`. Still, I think it should include both that `file_exists()` check in addition to the check for the required environment variables.

AhmarZaidi: Agreed, added both checks in the [latest commit](https://github.com/WordPress/performance/blob/48a17f5bddfd785d02e746795d0b80922cf9df45/tests/bootstrap.php#L35-L66).

thelovekesh:
I suggest we handle the exit code from PHPUnit to detect if it exits due to the absence of the WP tests setup. This way, we can wrap the test command in a bash or PHP script, avoiding the need to check global variables that might be initialized at multiple points during the WP test suite lifecycle.

tests/bootstrap.php
```diff
@@ -32,6 +32,39 @@
...
+$required_constants = array(
+	'WORDPRESS_DB_PASSWORD',
+	'WORDPRESS_DB_HOST',
+	'WORDPRESS_DB_USER',
+	'WORDPRESS_DB_NAME',
+	'PHP_VERSION',
+	'COMPOSER_BINARY',
+	'WP_TESTS_DIR',
+	'WP_PHPUNIT__DIR',
+);
```
thelovekesh: Are these env vars or constants defined at this point?

AhmarZaidi: These are env variables set by `wp-env` in `db-env.js` and used in `build-docker-compose-config.js`.
```
'WORDPRESS_DB_PASSWORD',
'WORDPRESS_DB_HOST',
'WORDPRESS_DB_USER',
'WORDPRESS_DB_NAME',
```

These are also env variables:
```
'PHP_VERSION',
'WP_TESTS_DIR',
```

`WP_PHPUNIT__DIR` is an env var that is set even on local when user runs `composer install`
> It looks like on local I have the `WP_PHPUNIT__DIR` environment variable being set by Composer. 

So, my understanding was that if these env vars are not set then that means the user is running the `composer test:plugin` command directly instead of running it using `npm`.

thelovekesh: And what about the scenario when this bootstrap file will be executed in a non `wp-env` environment?

AhmarZaidi: In case of local environment it would show error like this:
```bash
➜ composer test:plugins
> phpunit '--verbose' '--testsuite' 'auto-sizes'
Please use 'npm run test-php' instead of running the composer test command directly.
If you are intending to run PHPUnit directly, the following environment variables are missing: WORDPRESS_DB_PASSWORD, WORDPRESS_DB_HOST, WORDPRESS_DB_USER, WORDPRESS_DB_NAME, PHP_VERSION, WP_TESTS_DIR
Script phpunit handling the test event returned with error code 1
Script @test --verbose --testsuite auto-sizes was called via test:auto-sizes
Script @test:auto-sizes was called via test:plugins
```

AhmarZaidi:
> I suggest we handle the exit code from PHPUnit to detect if it exits due to the absence of the WP tests setup. This way, we can wrap the test command in a bash or PHP script, avoiding the need to check global variables that might be initialized at multiple points during the WP test suite lifecycle.

Makes sense but I just have one doubt. 

Currently this is the error we are getting on running the composer test commands directly:
```bash
➜  composer test:plugins
> phpunit '--verbose' '--testsuite' 'auto-sizes'
Error: The following required constants are not defined: WP_TESTS_DOMAIN, WP_TESTS_EMAIL, WP_TESTS_TITLE, WP_PHP_BINARY.
Please check out `wp-tests-config-sample.php` for an example.
Script phpunit handling the test event returned with error code 1
Script @test --verbose --testsuite auto-sizes was called via test:auto-sizes
Script @test:auto-sizes was called via test:plugins
```

This error `code 1` is from `vendor/wp-phpunit/wp-phpunit/includes/bootstrap.php`. A lot of other errors produce error code 1 so how would we determine that this is from incorrect WP test setup without checking for missing environment variables?

tests/bootstrap.php
thelovekesh: This message is irrelevant if someone tries to run tests from their custom setup. Also if the test setup is non-docker based, then I doubt if these globals/constants would ever be defined at this place in the bootstrap file.

thelovekesh:
Humm yes, this is what I was concerned about. A better exit code between 0 to 255 can be used apart from the reserved ones to communicate the error type. But I am not sure if it can be added to the core tests bootstrap file. 

Also, I think adding an exit with the proper message above this line should work in most of cases.

https://github.com/WordPress/performance/blob/d300b2ce48db3f1bcf575a7f96347d5fa007116b/tests/bootstrap.php#L35

It can check if `$_test_root` exists or not and then show the message.

thelovekesh:
@westonruter Any thoughts on if we can propose to add a meaningful error code if the test process is exiting prematurely in case dependencies not met?

westonruter:
I guess any error code other than what `phpunit` returns in the case of a failure. I think it returns `1`, so if you want a different exit code to be able to determine it failed for some other reason, then `2` seems fine.

AhmarZaidi:
> Also, I think adding an exit with the proper message above this line should work in most of cases.
> 
> https://github.com/WordPress/performance/blob/d300b2ce48db3f1bcf575a7f96347d5fa007116b/tests/bootstrap.php#L35
> 
> It can check if `$_test_root` exists or not and then show the message.


The `$_test_root` is also present on local without proper test env setup because it is being set by `vendor/wp-phpunit/wp-phpunit/__loaded.php` on `composer install`.
```php
putenv( sprintf( 'WP_PHPUNIT__DIR=%s', dirname( __FILE__ ) ) );
```
```php
elseif ( false !== getenv( 'WP_PHPUNIT__DIR' ) ) {
	$_test_root = getenv( 'WP_PHPUNIT__DIR' );
}
```

So, I don't think we can use that to determine if we should show error.


Also, the [original issue](https://github.com/WordPress/performance/issues/1256) was that if the user runs `composer test:plugins` command then they get a non descriptive error but if I'm not wrong, we would need to run `composer test:plugins` command from external script (bash or php) in order to get the exit code so that wouldn't solve the issue.

Please feel free correct me if I have misunderstood something here.

thelovekesh:
> The $_test_root is also present on local without proper test env setup because it is being set by vendor/wp-phpunit/wp-phpunit/__loaded.php on composer install.

Well, that's another case, I am not aware of. Also, I am not sure if we use `wp-phpunit` apart from symbol identification for IDEs. Can we remove it? I am not fully sure why it was added.

> Also, the https://github.com/WordPress/performance/issues/1256 was that if the user runs composer test:plugins command then they get a non descriptive error but if I'm not wrong, we would need to run composer test:plugins command from external script (bash or php) in order to get the exit code so that wouldn't solve the issue.Also, the https://github.com/WordPress/performance/issues/1256 was that if the user runs composer test:plugins command then they get a non descriptive error but if I'm not wrong, we would need to run composer test:plugins command from external script (bash or php) in order to get the exit code so that wouldn't solve the issue.

That doesn't mean we can't add a bash script to run the tests inside this command... right? But it's not doable right now because of the existing return code.

thelovekesh:
The issue with modifying the bootstrap file is that it can be loaded by PHPUnit in various environments, whereas this change only considers `wp-env` adjustments. Currently, if I run `composer test:plugin` in Lando, I encounter the following error:

```
➜  performance git:(trunk) ✗ lando composer test:plugins
> phpunit '--verbose' '--testsuite' 'auto-sizes'
Please use 'npm run test-php' instead of running the composer test command directly.
If you are intending to run PHPUnit directly, the following environment variables are missing: WORDPRESS_DB_PASSWORD, WORDPRESS_DB_HOST, WORDPRESS_DB_USER, WORDPRESS_DB_NAME
Script phpunit handling the test event returned with error code 1
Script @test --verbose --testsuite auto-sizes was called via test:auto-sizes
Script @test:auto-sizes was called via test:plugins
```
Without this change, my tests run as expected in the Lando environment.

The correct behavior should be: We need to detect when tests are failing due to a missing WP test suite setup and suggest using wp-env to run their tests.

thelovekesh:
> Also, I think adding an exit with the proper message above this line should work in most of cases.

> https://github.com/WordPress/performance/blob/d300b2ce48db3f1bcf575a7f96347d5fa007116b/tests/bootstrap.php#L35

> It can check if `$_test_root` exists or not and then show the message. 

We can still achieve this. If `$_test_root` contains `wp-phpunit/wp-phpunit`, then check if the `WP_PHPUNIT__TESTS_CONFIG` environment variable or `WP_TESTS_CONFIG_FILE_PATH` exists. If it doesn't, we can exit with an error message like this:

```php
echo 'Error: wp-tests-config.php is missing! Please use wp-tests-config-sample.php to create a config file.' . PHP_EOL;
echo 'Info: Alternatively, you can run `npm run wp-env start && npm run test-php` to execute tests in the wp-env environment.' . PHP_EOL;
exit(1);
```

In the rest of the cases, WP tests bootstrap will handle on its own:

https://github.com/WordPress/wordpress-develop/blob/df256e414c183ff001cdf26ebf37b19b960180de/tests/phpunit/includes/bootstrap.php#L25-L28

It will only jump to other cases if the user is using `wp-env` or some custom setup, so they know what they are doing :)

thelovekesh:
Given what @westonruter noted in https://github.com/WordPress/performance/pull/1314#discussion_r1651299256, I doubt if these lines are relevant anymore:

https://github.com/WordPress/performance/blob/49beab84e69d05e3eaa43b8ea96d2aa8587e4142/tests/bootstrap.php#L29-L33

because `wp-phpunit` always defines the `WP_PHPUNIT__DIR` env variable.

tests/bootstrap.php
```diff
@@ -32,6 +32,15 @@
 	$_test_root = '/tmp/wordpress-tests-lib';
 }
 
+// Check if tests are run locally without configuring test environment.
+if ( false !== strpos( $_test_root, 'wp-phpunit/wp-phpunit' ) ) {
+	if ( ! getenv( 'WP_PHPUNIT__TESTS_CONFIG' ) || ! getenv( 'WP_TESTS_CONFIG_FILE_PATH' ) ) {
```
thelovekesh: `WP_TESTS_CONFIG_FILE_PATH` is a constant.

AhmarZaidi: Thanks, I've updated it in the latest commit.

tests/bootstrap.php
```diff
@@ -32,6 +32,15 @@
...
 }
 
+// Check if tests are run locally without configuring test environment.
+if ( false !== strpos( $_test_root, 'wp-phpunit/wp-phpunit' ) ) {
+	if ( ! getenv( 'WP_PHPUNIT__TESTS_CONFIG' ) || ! defined( 'WP_TESTS_CONFIG_FILE_PATH' ) ) {
+		echo 'Error: wp-tests-config.php is missing! Please use wp-tests-config-sample.php to create a config file.' . PHP_EOL;
+		echo 'Info: Alternatively, you can run `npm run wp-env start && npm run test-php` to execute tests in the wp-env environment.' . PHP_EOL;
+		exit( 1 );
+	}
+}
```
westonruter:
```suggestion
if ( false !== strpos( $_test_root, 'wp-phpunit/wp-phpunit' ) && ! ( getenv( 'WP_PHPUNIT__TESTS_CONFIG' ) || defined( 'WP_TESTS_CONFIG_FILE_PATH' ) ) ) {
	echo 'Error: wp-tests-config.php is missing! Please use wp-tests-config-sample.php to create a config file.' . PHP_EOL;
	echo 'Info: Alternatively, you can run `npm run wp-env start && npm run test-php` to execute tests in the wp-env environment.' . PHP_EOL;
	exit( 2 );
}
```

thelovekesh: @westonruter `str_contains` is only PHP 8+ and not sure if polyfills will be loaded at this point of time.

westonruter: OK, right. I've updated my suggestion.

westonruter (approved):

PR #1318, opened 2024-06-24
Update PHPStan to 1.11.5

westonruter:
The following PHPStan failures are occurring in PHPStan 1.11.5:

```
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   plugins/dominant-color-images/class-dominant-color-image-editor-gd.php                                                                             
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------- 
  37     Strict comparison using === between false and int<1, max> will always evaluate to false.                                                           
         🪪  identical.alwaysFalse                                                                                                                          
  37     Strict comparison using === between false and int<1, max> will always evaluate to false.                                                           
         🪪  identical.alwaysFalse                                                                                                                          
  81     Call to function is_array() with array{red: int<0, 255>, green: int<0, 255>, blue: int<0, 255>, alpha: int<0, 127>} will always evaluate to true.  
         🪪  function.alreadyNarrowedType                                                                                                                   
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------------------------------- 
  Line   plugins/webp-uploads/tests/test-image-edit.php                                              
 ------ -------------------------------------------------------------------------------------------- 
  53     Strict comparison using === between 'full-orig' and *NEVER* will always evaluate to false.  
         🪪  identical.alwaysFalse                                                                   
 ------ -------------------------------------------------------------------------------------------- 
```

### DIFF
--- a/plugins/dominant-color-images/class-dominant-color-image-editor-gd.php
+++ b/plugins/dominant-color-images/class-dominant-color-image-editor-gd.php
@@ -35,10 +35,10 @@ class Dominant_Color_Image_Editor_GD extends WP_Image_Editor_GD {
 		if ( false === $shorted_image ) {
 			return new WP_Error( 'image_editor_dominant_color_error', __( 'Dominant color detection failed.', 'dominant-color-images' ) );
 		}
-		// Note: These two functions only return integers, but PHPStan thinks they return int|false. PhpStorm's stubs also think this. However, from looking at
-		// <https://github.com/php/php-src/blob/5db847e/ext/gd/gd.stub.php#L716-L718> these functions only ever returned integers. Nevertheless, the PHP docs
-		// did used to have int|false as the return type but this was removed for PHP 8: <https://github.com/php/doc-en/commit/0462f49>.
-		// So this is why there is type casting to integer, just for the sake of PHPStan, and possibly for PHP<8.
+		// Note: These two functions only return integers, but they used to return int|false in PHP<8. This was changed in the PHP documentation in
+		// <https://github.com/php/doc-en/commit/0462f49> and <https://github.com/php/doc-en/commit/37f858a>. However, PhpStorm's stubs still think
+		// they return int|false. However, from looking at <https://github.com/php/php-src/blob/5db847e/ext/gd/gd.stub.php#L716-L718> these functions
+		// apparently only ever returned integers. So the type casting is here for the possible sake PHP<8.
 		$image_width  = (int) imagesx( $this->image );
 		$image_height = (int) imagesy( $this->image );
 		imagecopyresampled( $shorted_image, $this->image, 0, 0, 0, 0, 1, 1, $image_width, $image_height );
--- a/plugins/dominant-color-images/class-dominant-color-image-editor-gd.php
+++ b/plugins/dominant-color-images/class-dominant-color-image-editor-gd.php
@@ -32,11 +32,11 @@ class Dominant_Color_Image_Editor_GD extends WP_Image_Editor_GD {
 		}
 		// The logic here is resize the image to 1x1 pixel, then get the color of that pixel.
 		$shorted_image = imagecreatetruecolor( 1, 1 );
-		$image_width   = imagesx( $this->image );
-		$image_height  = imagesy( $this->image );
-		if ( false === $shorted_image || false === $image_width || false === $image_height ) {
+		if ( false === $shorted_image ) {
 			return new WP_Error( 'image_editor_dominant_color_error', __( 'Dominant color detection failed.', 'dominant-color-images' ) );
 		}
+		$image_width  = imagesx( $this->image );
+		$image_height = imagesy( $this->image );
 		imagecopyresampled( $shorted_image, $this->image, 0, 0, 0, 0, 1, 1, $image_width, $image_height );
 
 		$rgb = imagecolorat( $shorted_image, 0, 0 );
@@ -78,9 +78,6 @@ class Dominant_Color_Image_Editor_GD extends WP_Image_Editor_GD {
 					return new WP_Error( 'unable_to_obtain_rgb_via_imagecolorat' );
 				}
 				$rgba = imagecolorsforindex( $this->image, $rgb );
-				if ( ! is_array( $rgba ) ) {
-					return new WP_Error( 'unable_to_obtain_rgba_via_imagecolorsforindex' );
-				}
 				if ( $rgba['alpha'] > 0 ) {
 					return true;
 				}
--- a/plugins/dominant-color-images/class-dominant-color-image-editor-gd.php
+++ b/plugins/dominant-color-images/class-dominant-color-image-editor-gd.php
@@ -35,7 +35,10 @@ class Dominant_Color_Image_Editor_GD extends WP_Image_Editor_GD {
 		if ( false === $shorted_image ) {
 			return new WP_Error( 'image_editor_dominant_color_error', __( 'Dominant color detection failed.', 'dominant-color-images' ) );
 		}
-		// Note: These two functions only return integers, but PHPStan thinks they return int|false. PhpStorm's stubs also think this.
+		// Note: These two functions only return integers, but PHPStan thinks they return int|false. PhpStorm's stubs also think this. However, from looking at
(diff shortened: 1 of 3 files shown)


plugins/dominant-color-images/class-dominant-color-image-editor-gd.php
```diff
@@ -32,11 +32,11 @@ public function get_dominant_color() {
 		}
 		// The logic here is resize the image to 1x1 pixel, then get the color of that pixel.
 		$shorted_image = imagecreatetruecolor( 1, 1 );
-		$image_width   = imagesx( $this->image );
-		$image_height  = imagesy( $this->image );
-		if ( false === $shorted_image || false === $image_width || false === $image_height ) {
```
westonruter, line 37: In PhpStorm's stubs, it defines:

```php
/**
 * Create a new true color image
 * @link https://php.net/manual/en/function.imagecreatetruecolor.php
 * @param int $width <p>
 * Image width.
 * </p>
 * @param int $height <p>
 * Image height.
 * </p>
 * @return resource|GdImage|false an image resource identifier on success, false on errors.
 */
#[Pure]
function imagecreatetruecolor(int $width, int $height): GdImage|false {}

/**
 * Get image width
 * @link https://php.net/manual/en/function.imagesx.php
 * @param resource|GdImage $image
 * @return int|false Return the width of the image or false on
 * errors.
 */
#[Pure]
function imagesx(GdImage $image): int {}

/**
 * Get image height
 * @link https://php.net/manual/en/function.imagesy.php
 * @param resource|GdImage $image
 * @return int|false Return the height of the image or false on
 * errors.
 */
#[Pure]
function imagesy(GdImage $image): int {}
```

All three have phpdoc that say they can return `false`, but only `imagecreatetruecolor()` has the PHP type hint which says it can return `false`, and this is confirmed in the PHP docs: https://www.php.net/manual/en/function.imagecreatetruecolor.php

In contrast, the PHP docs for the other functions say nothing about them returning false:

https://www.php.net/manual/en/function.imagesx.php
https://www.php.net/manual/en/function.imagesy.php

So it seems PhpStorm's docs are incorrect.

plugins/dominant-color-images/class-dominant-color-image-editor-gd.php
```diff
@@ -78,9 +78,6 @@ public function has_transparency() {
 					return new WP_Error( 'unable_to_obtain_rgb_via_imagecolorat' );
 				}
 				$rgba = imagecolorsforindex( $this->image, $rgb );
-				if ( ! is_array( $rgba ) ) {
-					return new WP_Error( 'unable_to_obtain_rgba_via_imagecolorsforindex' );
-				}
```
westonruter: PHP docs say that this function only returns an array: https://www.php.net/manual/en/function.imagecolorsforindex.php

But PhpStorm's stubs say it can return `false`:

```php
/**
 * Get the colors for an index
 * @link https://php.net/manual/en/function.imagecolorsforindex.php
 * @param resource|GdImage $image
 * @param int $color <p>
 * The color index.
 * </p>
 * @return array|false an associative array with red, green, blue and alpha keys that
 * contain the appropriate values for the specified color index or <b>FALSE</b> on failure
 */
#[Pure]
#[LanguageLevelTypeAware(['8.0' => 'array'], default: 'array|false')]
#[ArrayShape(["red" => "int", "green" => "int", "blue" => "int", "alpha" => "int"])]
function imagecolorsforindex(GdImage $image, int $color) {}
```

Again, it seems PhpStorm is wrong.

plugins/webp-uploads/tests/test-image-edit.php
```diff
@@ -47,18 +47,12 @@ public function test_it_should_backup_the_sources_structure_alongside_the_full_s
 		$this->assertArrayHasKey( 'full-orig', $backup_sources );
 		$this->assertSame( $metadata['sources'], $backup_sources['full-orig'] );
 
+		unset( $backup_sizes['full-orig'] );
 		foreach ( $backup_sizes as $size => $properties ) {
 			$size_name = str_replace( '-orig', '', $size );
-
-			if ( 'full-orig' === $size ) {
```
westonruter, line 53: I have no idea why PHPStan is saying that this is an error:

> Strict comparison using === between 'full-orig' and *NEVER* will always evaluate to false.

In any case, it seems better to just unset the array key rather than to check if with each iteration.

plugins/webp-uploads/tests/test-image-edit.php
```diff
@@ -47,18 +47,12 @@ public function test_it_should_backup_the_sources_structure_alongside_the_full_s
...
-				continue;
-			}
-
 			$this->assertArrayHasKey( 'sources', $properties );
 			$this->assertSame( $metadata['sizes'][ $size_name ]['sources'], $properties['sources'] );
 		}
-
-		$metadata = wp_get_attachment_metadata( $attachment_id );
```
westonruter, line 61: This appears to be dead code, perhaps code which was going to have assertions added to it, but they never were written.

plugins/webp-uploads/tests/test-image-edit.php
```diff
@@ -136,7 +130,7 @@ public function test_it_should_prevent_to_back_up_the_sources_when_the_sources_a
 		$backup_sizes = get_post_meta( $attachment_id, '_wp_attachment_backup_sizes', true );
 		$this->assertIsArray( $backup_sizes );
 
-		foreach ( $backup_sizes as $size_name => $properties ) {
```
westonruter, line 139: Removing unused variable.

plugins/webp-uploads/tests/test-image-edit.php
```diff
@@ -281,7 +275,7 @@ static function () {
 		foreach ( $metadata['sizes'] as $size_name => $properties ) {
 			$this->assertImageNotHasSizeSource( $attachment_id, $size_name, 'image/jpeg' );
 			$this->assertImageHasSizeSource( $attachment_id, $size_name, 'image/webp' );
-			foreach ( $properties['sources'] as $mime_type => $values ) {
```
westonruter, line 284: Removing unused variable.

plugins/dominant-color-images/class-dominant-color-image-editor-gd.php
joemcgill: According to the [changelog](https://www.php.net/manual/en/function.imagecolorsforindex.php#refsect1-function.imagecolorsforindex-changelog):

> imagecolorsforindex() now throws a [ValueError](https://www.php.net/manual/en/class.valueerror.php) exception if color is out of range; previously, [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) was returned instead.

I wonder what versions of GD we need to support?

joemcgill (approved):
Looks good to me.

plugins/dominant-color-images/class-dominant-color-image-editor-gd.php
westonruter: Ah, then we should probably wrap this in a try/catch block to account for this.

mukeshpanchal27 (approved):
The changes look good to me, and I pre-approve the PR. There is one piece of feedback that needs to be addressed: https://github.com/WordPress/performance/pull/1318#discussion_r1651895825

mukeshpanchal27:
@westonruter @joemcgill  I recommend that we stick to a specific PHPStan version for now, as it blocks the progress of other PRs. When I worked on #1252, I encountered similar errors, and it didn't allow me to push changes without resolving these PHPStan errors.

Sticking to the current version will prevent blocking other PRs and we can plan to update eventually when a new version comes up, ensuring a smooth workflow.

westonruter:
@mukeshpanchal27 I don't know if I agree. When would we upgrade then? Resolving PHPStan errors should in general just make the code better. In your PR https://github.com/WordPress/performance/pull/1252 you may have gotten PHPStan errors if you switched to this branch, ran `composer install`, and then switched back to your other branch without doing `composer install` again. However, GitHub Actions shouldn't have been reporting any errors against `trunk` since it is still pinned on the older version of PHPStan.

PR #1319, opened 2024-06-25
Picture element images: Add missing alt text

### LINKED ISSUE #1315: `Picture` element images: Missing alt text
## Bug Description

When using the new Modern Image Formats `picture` element support, I noticed that the markup didn't add the alt text that supply by the editor.

## Steps to reproduce

1. Activate the Modern Image Formats plugin and enable `picture` support from the Media Settings screen.
2. Add an image block and add alt text in Block settings > ALTERNATIVE TEXT
3. Preview the page in the front end

### Editor
Screenshot 2024-06-24 at 4 54 56 PM

### Frontend
Screenshot 2024-06-24 at 4 55 08 PM

devansh016:
## Summary
- Add alt text to original_image_without_srcset variable which was missing.

Fixes #1315

## Relevant technical choices

For maintainers only, please make sure:

- PR has a `[Type]` label.
- PR has a plugin-specific milestone, or the `no milestone` label if it does not apply to any specific plugin.
- PR has a changelog-friendly title, or the `skip changelog` label if it should not be mentioned in the plugin's changelog.
-->

### DIFF
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -81,10 +81,12 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	// Extract sizes using regex to parse image tag, then use to retrieve tag.
 	$width     = 0;
 	$height    = 0;
+	$alt       = '';
 	$processor = new WP_HTML_Tag_Processor( $image );
 	if ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
 		$width  = (int) $processor->get_attribute( 'width' );
 		$height = (int) $processor->get_attribute( 'height' );
+		$alt    = (string) $processor->get_attribute( 'alt' );
 	}
 	$size_to_use = ( $width > 0 && $height > 0 ) ? array( $width, $height ) : 'full';
 
@@ -124,10 +126,11 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 			continue;
 		}
 		$picture_sources .= sprintf(
-			'<source type="%s" srcset="%s" sizes="%s">',
+			'<source type="%s" srcset="%s" sizes="%s" alt="%s">',
 			esc_attr( $image_mime_type ),
 			esc_attr( $image_srcset ),
-			esc_attr( $sizes )
+			esc_attr( $sizes ),
+			esc_attr( $alt )
 		);
 	}
 
@@ -142,7 +145,15 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 		return false;
 	};
 	add_filter( 'wp_calculate_image_srcset_meta', $filter );
-	$original_image_without_srcset = wp_get_attachment_image( $attachment_id, $original_sizes, false, array( 'src' => $original_image ) );
+	$original_image_without_srcset = wp_get_attachment_image(
+		$attachment_id,
+		$original_sizes,
+		false,
+		array(
+			'src' => $original_image,
+			'alt' => $alt,
+		)
+	);
 	remove_filter( 'wp_calculate_image_srcset_meta', $filter );
 
 	return sprintf(
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -81,12 +81,10 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	// Extract sizes using regex to parse image tag, then use to retrieve tag.
 	$width     = 0;
 	$height    = 0;
-	$alt       = '';
 	$processor = new WP_HTML_Tag_Processor( $image );
 	if ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
 		$width  = (int) $processor->get_attribute( 'width' );
 		$height = (int) $processor->get_attribute( 'height' );
-		$alt    = (string) $processor->get_attribute( 'alt' );
 	}
 	$size_to_use = ( $width > 0 && $height > 0 ) ? array( $width, $height ) : 'full';
 
@@ -126,11 +124,10 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 			continue;
 		}
 		$picture_sources .= sprintf(
-			'<source type="%s" srcset="%s" sizes="%s" alt="%s">',
+			'<source type="%s" srcset="%s" sizes="%s">',
 			esc_attr( $image_mime_type ),
 			esc_attr( $image_srcset ),
-			esc_attr( $sizes ),
-			esc_attr( $alt )
+			esc_attr( $sizes )
 		);
 	}
 
@@ -145,15 +142,7 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 		return false;
 	};
 	add_filter( 'wp_calculate_image_srcset_meta', $filter );
-	$original_image_without_srcset = wp_get_attachment_image(
-		$attachment_id,
-		$original_sizes,
-		false,
-		array(
-			'src' => $original_image,
-			'alt' => $alt,
-		)
-	);
+	$original_image_without_srcset = wp_get_attachment_image( $attachment_id, $original_sizes, false, array( 'src' => $original_image ) );
 	remove_filter( 'wp_calculate_image_srcset_meta', $filter );
 
 	return sprintf(
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -131,16 +131,6 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 		);
 	}
 
-	// Fall back to the original image without a srcset.
-	$original_image = wp_get_original_image_url( $attachment_id );
-	// Fail gracefully if the original image is not found.
-	if ( false === $original_image ) {
-		return $image;
-	}
-	$filter = static function (): bool {
-		return false;
-	};
-
 	return sprintf(
 		'<picture class="%s" style="display: contents;">%s%s</picture>',
 		esc_attr( 'wp-picture-' . $attachment_id ),
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -132,7 +132,6 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	}
 
 	// Fall back to the original image without a srcset.
-	$original_sizes = array( $image_src[1], $image_src[2] );
 	$original_image = wp_get_original_image_url( $attachment_id );
 	// Fail gracefully if the original image is not found.
 	if ( false === $original_image ) {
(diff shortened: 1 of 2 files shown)


plugins/webp-uploads/picture-element.php
thelovekesh, line 1: It might need some test cases to be written/updated?

plugins/webp-uploads/picture-element.php
```diff
@@ -124,10 +126,11 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 			continue;
 		}
 		$picture_sources .= sprintf(
-			'<source type="%s" srcset="%s" sizes="%s">',
+			'<source type="%s" srcset="%s" sizes="%s" alt="%s">',
```
westonruter: I don't think this is right. The `alt` should still only go on the `img` shouldn't it? The bug seems to be that it is not being added to the `img` not that it is absent from the `source`. You can see from the [examples on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture#attributes) that only the underlying `img` tag gets the `alt` attribute.

westonruter (changes requested):

plugins/webp-uploads/picture-element.php
```diff
@@ -142,7 +145,15 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
...
+	$original_image_without_srcset = wp_get_attachment_image(
+		$attachment_id,
+		$original_sizes,
+		false,
+		array(
+			'src' => $original_image,
+			'alt' => $alt,
+		)
+	);
 	remove_filter( 'wp_calculate_image_srcset_meta', $filter );
```
westonruter: I think there is a better approach here, one critically which will preserve all of the other attributes. The HTML tag processor can be leveraged instead like as follows:

```php
$processor = new WP_HTML_Tag_Processor( $image );
if ( $processor->next_tag( array( 'tag_name' => 'img' ) ) ) {
    $processor->remove_attribute( 'srcset' );
    $image = $processor->get_updated_html();
}
```

Nevertheless, why is the `srcset` being removed in the first place? Shouldn't it be retained for the sake of browsers that don't support `picture`?

plugins/webp-uploads/picture-element.php
devansh016: gotcha

plugins/webp-uploads/picture-element.php
devansh016: Yes it will be a  good practice to add srcset to both `<source>` and `` elements. I couldn't find why it is being removed.

On `<source>` elements:
- Provides options for different viewport sizes or device pixel ratios.
- Allows browsers to choose the most appropriate image for the current conditions.

On the `` element:
- Acts as a fallback for browsers that don't support <picture> or <source>.
- Ensures responsiveness even if <source> elements are ignored.

plugins/webp-uploads/picture-element.php
westonruter:
```suggestion
			'<source type="%s" srcset="%s" sizes="%s">',
```

plugins/webp-uploads/picture-element.php
westonruter: So yeah, then this entire block of code can be removed and instead of using `$original_image_without_srcset` we can just use `$image` as was passed into this function originally.

plugins/webp-uploads/picture-element.php
```diff
@@ -141,14 +140,11 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	$filter = static function (): bool {
 		return false;
 	};
```
westonruter:
```suggestion
```

plugins/webp-uploads/picture-element.php
```diff
@@ -141,14 +140,11 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
...
-	add_filter( 'wp_calculate_image_srcset_meta', $filter );
-	$original_image_without_srcset = wp_get_attachment_image( $attachment_id, $original_sizes, false, array( 'src' => $original_image ) );
-	remove_filter( 'wp_calculate_image_srcset_meta', $filter );
 
 	return sprintf(
 		'<picture class="%s" style="display: contents;">%s%s</picture>',
 		esc_attr( 'wp-picture-' . $attachment_id ),
 		$picture_sources,
-		$original_image_without_srcset
+		$image
```
adamsilverstein, line 138: does this `$image` include a `srcset`?

devansh016: Yes, `$image` include a `srcset`.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -50,8 +50,22 @@ public function test_maybe_wrap_images_in_picture_element( bool $jpeg_and_webp,
...
+			$this->assertEquals( 1, substr_count( $the_image, 'srcset=' ) );
+		}
+		// When only jpeg_and_webp is enabled, the picture element will contain 1 srcset elements.
+		if ( $jpeg_and_webp && ! $expect_picture_element ) {
+			$this->assertEquals( 1, substr_count( $the_image, 'srcset=' ) );
+		}
+		// When only picture_element is enabled, the picture element will contain 2 srcset elements.
+		if ( ! $jpeg_and_webp && $expect_picture_element ) {
+			$this->assertEquals( 2, substr_count( $the_image, 'srcset=' ) );
+		}
```
westonruter: Instead of `$expect_picture_element` I think it would perhaps be helpful if the actual HTML for the `picture` element is passed into this test form the `data_provider_it_should_maybe_wrap_images_in_picture_element` data provider. For example, it could be `$expected_html` and then this here can just do:

```php
$this->assertEquals( $expected_html, $the_image );
```

By doing such snapshot comparisons, we'll be able to check for more expected markup than just the `srcset`, for example whether `alt` appears on the `img`.

plugins/webp-uploads/picture-element.php
mukeshpanchal27:
```suggestion
```

Remove un-used code.

plugins/webp-uploads/picture-element.php
mukeshpanchal27, line 138: @adamsilverstein As a fallback image, we should return the original image. In [this MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture#the_type_attribute), they return the original image instead of mime-type-specific images.

I couldn't find proper documentation for this. What do you think?

mukeshpanchal27 (changes requested):
Thanks @devansh016 for working on it. Left one question and code suggestion.

plugins/webp-uploads/picture-element.php
```diff
@@ -132,7 +132,6 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	}
 
 	// Fall back to the original image without a srcset.
-	$original_sizes = array( $image_src[1], $image_src[2] );
 	$original_image = wp_get_original_image_url( $attachment_id );
 	// Fail gracefully if the original image is not found.
 	if ( false === $original_image ) {
```
westonruter: This code also unused and can be removed.

plugins/webp-uploads/picture-element.php
westonruter: Yes, I think this is right. 

Also, I think the `source` being generated for `'image/jpeg'` can also be removed as well, because if the browser doesn't support AVIF/WebP then it can just fall back to the `img` which is a JPEG.

plugins/webp-uploads/tests/test-picture-element.php
devansh016: I was trying to implement this but the `src` and `srcset` keep on changing whenever we run the test case. Therefore, `$this->assertEquals( $expected_html, $the_image );` will may not work here.

@westonruter

plugins/webp-uploads/picture-element.php
devansh016: https://github.com/WordPress/performance/blob/1ed2fbdab5fa47e2b174b813f6d5d3273b589482/plugins/webp-uploads/picture-element.php#L56-L64

Should we remove `line 61` for the same.

plugins/webp-uploads/tests/test-picture-element.php
thelovekesh: @devansh016 Since data providers are evaluated before test cases, and the generated attachments are dynamic, you can use placeholders for image src in the expected content like `{{img-src}}`, which can be search and replaced during the actual test.

devansh016: I have updated the tests.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -37,49 +38,90 @@ public function test_maybe_wrap_images_in_picture_element( bool $jpeg_and_webp,
...
+		// Prepare the expected HTML by replacing placeholders with expected values.
+		$expected_html = str_replace( '{{img-width}}', $width, $expected_html );
+		$expected_html = str_replace( '{{img-height}}', $height, $expected_html );
+		$expected_html = str_replace( '{{img-src}}', $img_src, $expected_html );
+		$expected_html = str_replace( '{{img-attachment-id}}', $attachment_id, $expected_html );
+		$expected_html = str_replace( '{{img-alt}}', $alt, $expected_html );
+		$expected_html = str_replace( '{{img-srcset}}', $image_srcset, $expected_html );
+		$expected_html = str_replace( '{{img-sizes}}', $sizes, $expected_html );
+		$expected_html = str_replace( '{{jpeg-srcset}}', $jpeg_srcset, $expected_html );
+		$expected_html = str_replace( '{{webp-srcset}}', $webp_srcset, $expected_html );
```
westonruter: This can be made less verbose like as follows:

```php
$replacements = array(
  '{{img-width}}' => $width,
  '{{img-height}}' => $height,
  // ...
)
$expected_html = str_replace( array_keys( $replacements ), array_values( $replacements ), $expected_html );
```

devansh016: Yes, this seems better.

plugins/webp-uploads/picture-element.php
mukeshpanchal27: I think we can open a separate PR for this issue to avoid complicating the current one. This will allow us to address it more effectively.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -37,49 +38,94 @@ public function test_maybe_wrap_images_in_picture_element( bool $jpeg_and_webp,
...
-		$the_image = apply_filters( 'wp_content_img_tag', $the_image, 'the_content', $attachment_id );
+		$processor = new WP_HTML_Tag_Processor( $the_image );
+		$width     = 0;
+		$height    = 0;
+		$alt       = '';
+		if ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
+			$width  = (int) $processor->get_attribute( 'width' );
+			$height = (int) $processor->get_attribute( 'height' );
+			$alt    = (string) $processor->get_attribute( 'alt' );
+		}
```
westonruter:
```suggestion
		$this->assertTrue( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) );
		$width  = (int) $processor->get_attribute( 'width' );
		$height = (int) $processor->get_attribute( 'height' );
		$alt    = (string) $processor->get_attribute( 'alt' );
```

westonruter (approved):
LGTM

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -37,49 +38,94 @@ public function test_maybe_wrap_images_in_picture_element( bool $jpeg_and_webp,
...
+				'alt'   => 'Green Leaves',
+			)
+		);
 
-		// Apply the wp_content_img_tag filter.
-		$the_image = apply_filters( 'wp_content_img_tag', $the_image, 'the_content', $attachment_id );
+		$processor = new WP_HTML_Tag_Processor( $the_image );
+		$width     = 0;
+		$height    = 0;
+		$alt       = '';
```
westonruter:
```suggestion
```

mukeshpanchal27 (approved):

PR #1320, opened 2024-06-25
Set fetchpriority to low if image is in the initial viewport

### LINKED ISSUE #1309: Occluded initial-viewport images should get fetchpriority=low
## Feature Description

When an image is in the initial viewport but is hidden, it should get `fetchpriority=low`. For example, this would apply to images which appear inside of a mega menu or images which are part of hidden carousel slides. Such images should not get `loading=lazy` since they should still be loaded and ready to go the moment the user interacts with the page to reveal it, but they should not get the same priority as images which are visible when first loading the page.

See https://web.dev/articles/fetch-priority#lower-carousel-priority

devansh016:
## Summary
Set fetchpriority to low if image is in the initial viewport but is not visible

Fixes #1309 

## Relevant technical choices

```php
if ( ! $is_visible ) {
	$walker->set_attribute( 'fetchpriority', 'low' );
}
```


westonruter (changes requested):
Thank you for the PR. See feedback in https://github.com/WordPress/performance/issues/1309#issuecomment-2189404341

PR #1322, opened 2024-06-25
Integrate Auto Sizes with Image Prioritizer to ensure correct sizes=auto

### LINKED ISSUE #1099: Auto Sizes should extend Optimization Detective to add sizes=auto when also lazy-loading images
## Feature Description

The Auto Sizes plugin adds `sizes=auto` to all images with `loading=lazy`. This is all done server-side without knowledge of which images are actually in the initial viewport across breakpoints. This is knowledge that Optimization Detective has from user visits. In #1077 images which are not in any breakpoint group's initial viewport will be lazy-loaded. However, the addition of `loading=lazy` to these images should also include `sizes=auto`. This could be done in the same "Image Loading Optimization" (Image Prioritizer?) dependent plugin of Optimization Detective, or better it could be incorporated as part of the Auto Sizes plugin instead.

westonruter:
This integrates the Auto Sizes plugin with Image Prioritizer (and Optimization Detective) to ensure that `sizes=auto` is added to all of the images which actually end up lazy-loaded. Conversely, if Auto Sizes added `sizes=auto` to an initially lazy-loaded image, but then Image Prioritizer removed `loading=lazy` since the image is actually in the initial viewport, then remove `sizes=auto`.

Fixes #1099

### DIFF
--- a/plugins/auto-sizes/optimization-detective.php
+++ b/plugins/auto-sizes/optimization-detective.php
@@ -13,6 +13,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 /**
  * Visits responsive lazy-loaded IMG tags to ensure they include sizes=auto.
  *
+ * @since n.e.x.t
+ *
  * @param OD_Tag_Visitor_Context $context Tag visitor context.
  * @return bool Whether visited.
  */
--- a/plugins/auto-sizes/optimization-detective.php
+++ b/plugins/auto-sizes/optimization-detective.php
@@ -24,13 +24,27 @@ function auto_sizes_visit_tag( OD_Tag_Visitor_Context $context ): bool {
 	}
 
 	$sizes = $context->processor->get_attribute( 'sizes' );
-	if ( ! is_string( $sizes ) || 'lazy' !== $context->processor->get_attribute( 'loading' ) ) {
+	if ( ! is_string( $sizes ) ) {
 		return false;
 	}
 
 	$sizes = preg_split( '/\s*,\s*/', $sizes );
-	if ( is_array( $sizes ) && ! in_array( 'auto', $sizes, true ) ) {
+	if ( ! is_array( $sizes ) ) {
+		return false;
+	}
+
+	$is_lazy_loaded = ( 'lazy' === $context->processor->get_attribute( 'loading' ) );
+	$has_auto_sizes = in_array( 'auto', $sizes, true );
+
+	$changed = false;
+	if ( $is_lazy_loaded && ! $has_auto_sizes ) {
 		array_unshift( $sizes, 'auto' );
+		$changed = true;
+	} elseif ( ! $is_lazy_loaded && $has_auto_sizes ) {
+		$sizes   = array_diff( $sizes, array( 'auto' ) );
+		$changed = true;
+	}
+	if ( $changed ) {
 		$context->processor->set_attribute( 'sizes', join( ', ', $sizes ) );
 	}
 
--- a/plugins/auto-sizes/optimization-detective.php
+++ b/plugins/auto-sizes/optimization-detective.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Optimization Detective extensions by Auto Sizes.
+ *
+ * @since n.e.x.t
+ * @package auto-sizes
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * Visits responsive lazy-loaded IMG tags to ensure they include sizes=auto.
+ *
+ * @param OD_Tag_Visitor_Context $context Tag visitor context.
+ * @return bool Whether visited.
+ */
+function auto_sizes_visit_tag( OD_Tag_Visitor_Context $context ): bool {
+	if ( 'IMG' !== $context->processor->get_tag() ) {
+		return false;
+	}
+
+	$sizes = $context->processor->get_attribute( 'sizes' );
+	if ( ! is_string( $sizes ) || 'lazy' !== $context->processor->get_attribute( 'loading' ) ) {
+		return false;
+	}
+
+	$sizes = preg_split( '/\s*,\s*/', $sizes );
+	if ( is_array( $sizes ) && ! in_array( 'auto', $sizes, true ) ) {
+		array_unshift( $sizes, 'auto' );
+		$context->processor->set_attribute( 'sizes', join( ', ', $sizes ) );
+	}
+
+	return true;
+}
+
+/**
+ * Registers the tag visitor for image tags.
+ *
+ * @since n.e.x.t
+ *
+ * @param OD_Tag_Visitor_Registry $registry Tag visitor registry.
+ */
+function auto_sizes_register_tag_visitors( OD_Tag_Visitor_Registry $registry ): void {
+	$registry->register( 'auto-sizes', 'auto_sizes_visit_tag' );
+}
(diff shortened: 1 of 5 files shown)


plugins/auto-sizes/optimization-detective.php
```diff
@@ -0,0 +1,50 @@
...
+ * @package auto-sizes
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * Visits responsive lazy-loaded IMG tags to ensure they include sizes=auto.
+ *
```
mukeshpanchal27, line 15:
```suggestion
 *
 * @since n.e.x.t
 *
```

plugins/auto-sizes/optimization-detective.php
```diff
@@ -0,0 +1,52 @@
...
+		return false;
+	}
+
+	$sizes = preg_split( '/\s*,\s*/', $sizes );
+	if ( is_array( $sizes ) && ! in_array( 'auto', $sizes, true ) ) {
+		array_unshift( $sizes, 'auto' );
+		$context->processor->set_attribute( 'sizes', join( ', ', $sizes ) );
+	}
+
+	return true;
```
westonruter: In reality this can return `false` per #1342.

plugins/auto-sizes/optimization-detective.php
```diff
@@ -0,0 +1,52 @@
...
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * Visits responsive lazy-loaded IMG tags to ensure they include sizes=auto.
+ *
+ * @since n.e.x.t
+ *
+ * @param OD_Tag_Visitor_Context $context Tag visitor context.
+ * @return bool Whether visited.
```
westonruter:
```suggestion
 * @return false Whether the tag should be recorded in URL metrics.
```

See https://github.com/WordPress/performance/issues/1342.

joemcgill:
Thanks for this, @westonruter. I assume that making use of OD functionality directly when enabled is needed because there is not a better way to make sure that the auto-sizes functionality runs late enough that any alterations that OD makes to the `loading` attribute can be taken into account before we add `auto` to the front of the `sizes` attribute?

plugins/auto-sizes/optimization-detective.php
joemcgill: Happy to defer to whatever you think here.

westonruter:
```suggestion
	return false; // Since this tag visitor does not require this tag to be included in the URL Metrics.
```

westonruter:
@joemcgill Here OD is being leveraged as a way to make modifications to the output-buffered page, which Image Prioritizer may have also modified to add/remove `loading=lazy` from `img` elements based on whether they appear in the initial viewport or not. So yeah, Auto Sizes may be missing opportunities to add `sizes=auto` since it doesn't know for sure whether an image should be lazy-loaded or not, since the server-side heuristics are limited. With the URL metrics in hand, Image Prioritizer can be more accurate about adding or removing lazy-loading, and then this integration here follows up to amend `sizes=auto` where appropriate.

swissspidy (approved):

joemcgill:
@westonruter, thanks for clarifying. I think there is a slight by important difference between:

a) This plugin uses OD when available to more accurately apply `sizes=auto`, and
b) This plugin uses OD to fix a potential bug when Image Optimizer (or another OD dependent plugin) modifies the lazy loading attribute of an image after `sizes=auto` has been applied.

I assumed that the goal of this was to address the latter, which is why I was asking about wether there is an alternate way that this plugin could account for any cases (including but not limited to OD) where the loading attribute is modified later than we currently apply `sizes=auto`.

westonruter:
@joemcgill actually it's both, isn't it? It does more accurately add `sizes=auto`, while at the same time it will fix up any `sizes=auto` being applied to images which have had their lazy-loading updated after initial template rendering with server-side heuristics.

westonruter:
@joemcgill I fixed the merge conflict. (Difficult to do on mobile!)

joemcgill:
> @joemcgill actually it's both, isn't it?

Right now this PR is addressing both, but what I'm trying to better understand is what the _primary_ objective of this PR is, since it isn't related to an open issue.

While I see OD being very valuable tool in potentially fixing various types of bugs where third party plugins (like Image Prioritizer) modify the `loading` attribute of an image after we've applied `sizes=auto`, I want to avoid having OD be a dependency of this functionality so we can keep the auto-sizes improvements decoupled from OD while working towards including that functionality directly in WP Core.

The two ways that I'd prefer to address the Image Prioritizer issue would be to either:

1. Find an OD agnostic fix that would make sure we're modifying `sizes` late enough where the Image Prioritizer modifications run first.
2. Handle the compatibility issue in Image Prioritizer, since it is the plugin that would be responsible for the incompatibility if auto-sizes were in Core.

westonruter:
@joemcgill 

> Right now this PR is addressing both, but what I'm trying to better understand is what the _primary_ objective of this PR is, since it isn't related to an open issue.

Oops. I failed to link this PR to the open issue. See #1099.

> 1. Find an OD agnostic fix that would make sure we're modifying `sizes` late enough where the Image Prioritizer modifications run first.

Since Image Prioritizer always applies changes to lazy-loading (either adding `loading=lazy` or removing the `loading` attribute) after the template has rendered during output-buffer processing, I don't see how there is any alternative to introducing a new OD tag visitor as this PR proposes.

> I want to avoid having OD be a dependency of this functionality so we can keep the auto-sizes improvements decoupled from OD while working towards including that functionality directly in WP Core.

OD is not made a dependency with this PR. Rather, it integrates with OD if it is active. I kept all the OD functionality in a separate file to ensure it doesn't trip up core merge.

> 2\. Handle the compatibility issue in Image Prioritizer, since it is the plugin that would be responsible for the incompatibility if auto-sizes were in Core.

The inclusion of `sizes=auto` could be applied in Image Prioritizer directly instead of doing it in the Auto Sizes plugin. But from https://github.com/WordPress/performance/issues/1088#issuecomment-2035659201 I understood `sizes`-specific optimizations should be out of scope for Image Prioritizer.

plugins/auto-sizes/optimization-detective.php
```diff
@@ -0,0 +1,52 @@
...
+ * @return false Whether the tag should be recorded in URL metrics.
+ */
+function auto_sizes_visit_tag( OD_Tag_Visitor_Context $context ): bool {
+	if ( 'IMG' !== $context->processor->get_tag() ) {
+		return false;
+	}
+
+	$sizes = $context->processor->get_attribute( 'sizes' );
+	if ( ! is_string( $sizes ) || 'lazy' !== $context->processor->get_attribute( 'loading' ) ) {
+		return false;
```
westonruter, line 28: Oh, what if `sizes=auto` had been added by Auto Sizes but Image Prioritizer removed `loading=lazy` since the image actually appears in the initial viewport? Should this make sure that `auto` is removed from `sizes` as well? Is it bad if a non-lazy loaded image includes `sizes=auto`?

westonruter: Addressed in f5e4231c

joemcgill (approved):
Tested the latest updates locally and everything looks good. I still have reservations about using OD to make late adjustments to the `sizes` attribute, but this at least fixes the Image Prioritizer problem until we have a better option. Let's land it.

PR #1323, opened 2024-06-28
Send preload links via HTTP Link headers in addition to LINK tags

### LINKED ISSUE #1321: Preload links can be sent via `Link` HTTP response headers over HTML `link` tags
As noted in the web.dev article, [Preload critical assets to improve loading speed](https://web.dev/articles/preload-critical-assets#how_to_implement_relpreload:~:text=You%20can%20also%20preload%20any%20type%20of%20resource%20via%20the%20Link%20HTTP%20header%3A), sending the preload links via `Link` HTTP response headers has a slight performance advantage over HTML `link` tags:

> You can also preload any type of resource via the [Link HTTP header](https://developer.mozilla.org/docs/Web/HTTP/Headers/Link):
> 
> ```http
> Link: </css/style.css>; rel="preload"; as="style"
> ```
> 
> A benefit of specifying preload in the HTTP Header is that the browser doesn't need to parse the document to discover it, which can offer small improvements in some cases.

Since Optimization Detective is output-buffering the page, this can be very straightforward to implement. Instead of (or rather _in addition to_) doing this:

https://github.com/WordPress/performance/blob/2f2691a153cf1819a907a7cb54205d44c008bb62/plugins/optimization-detective/optimization.php#L192-L195

A new method can be added to `OD_Preload_Link_Collection` like `send_headers()` which can construct the `Link` response header and send it via `header()`.

AhmarZaidi:
## Summary

Fixes #1321

## Relevant technical choices

- Added a new method `get_headers()` to `OD_Preload_Link_Collection` to construct the Link HTTP response headers for preloading resources.
- Updated `od_optimize_template_output_buffer` to call `headers()` if there are preload links available and headers have not been sent already.
- Sent HTTP Link headers separately (above appending the link tags) to prevent [`Cannot modify header information - headers already sent`](https://stackoverflow.com/a/8028987/19726929) error due to output before sending headers.


AhmarZaidi:
Please add type label and milestone.

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -166,6 +166,37 @@ public function get_html(): string {
 		return implode( '', $link_tags );
 	}
 
+	/**
+	 * Gets the HTTP Link header string.
+	 *
+	 * @return string HTTP Link header.
+	 */
+	public function get_headers(): string {
```
westonruter, line 174: Let's call this `get_response_header`

westonruter: Let's call this `get_response_header`

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -166,6 +166,37 @@ public function get_html(): string {
 		return implode( '', $link_tags );
 	}
 
+	/**
+	 * Gets the HTTP Link header string.
```
westonruter:
```suggestion
	 * Constructs the Link HTTP response header.
```

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -166,6 +166,37 @@ public function get_html(): string {
...
+	 *
+	 * @return string HTTP Link header.
```
westonruter:
```suggestion
	 * @return string|null Link HTTP response header, or null if there are none.
```

plugins/optimization-detective/class-od-preload-link-collection.php
westonruter:
```suggestion
	public function get_response_header(): ?string {
```

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -166,6 +166,37 @@ public function get_html(): string {
...
+			$media_features = array( 'screen' );
+			if ( null !== $link['minimum_viewport_width'] && $link['minimum_viewport_width'] > 0 ) {
+				$media_features[] = sprintf( '(min-width: %dpx)', $link['minimum_viewport_width'] );
+			}
+			if ( null !== $link['maximum_viewport_width'] && PHP_INT_MAX !== $link['maximum_viewport_width'] ) {
+				$media_features[] = sprintf( '(max-width: %dpx)', $link['maximum_viewport_width'] );
+			}
+			$link['attributes']['media'] = implode( ' and ', $media_features );
+
+			$link_header = '<' . esc_url( $link['attributes']['href'] ?? '' ) . '>; rel="preload"';
```
westonruter: The `esc_url()` function includes HTML entities. So this should rather be:

```suggestion
			$link_header = '<' . esc_url_raw( $link['attributes']['href'] ?? '' ) . '>; rel="preload"';
```

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -166,6 +166,37 @@ public function get_html(): string {
...
+			foreach ( $link['attributes'] as $name => $value ) {
+				if ( 'href' !== $name ) {
+					$link_header .= sprintf( '; %s="%s"', $name, esc_attr( $value ) );
+				}
+			}
+
+			$link_headers[] = $link_header;
+		}
+
```
westonruter, line 211:
```suggestion
		if ( count( $link_headers ) === 0 ) {
				return null;
		}

```

plugins/optimization-detective/optimization.php
```diff
@@ -164,6 +164,11 @@ function od_optimize_template_output_buffer( string $buffer ): string {
 
 	$tag_visitor_registry = new OD_Tag_Visitor_Registry();
 
+	// Send any preload links as Link headers.
+	if ( count( $preload_links ) > 0 && ! headers_sent() ) {
+		header( $preload_links->get_headers() );
+	}
+
```
westonruter: Let's move this down to combine with the existing condition:

```php
	// Inject any preload links at the end of the HEAD.
	if ( count( $preload_links ) > 0 ) {
		$walker->append_head_html( $preload_links->get_html() );
	}
```

For example:

```php
// Send any preload links in a Link response header and in a LINK tag injected at the end of the HEAD.
if ( count( $preload_links ) > 0 ) {
	header( $preload_links->get_response_header() );
	$walker->append_head_html( $preload_links->get_html() );
}
```

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -166,6 +166,37 @@ public function get_html(): string {
...
 
+	/**
+	 * Gets the HTTP Link header string.
+	 *
+	 * @return string HTTP Link header.
+	 */
+	public function get_headers(): string {
+		$link_headers = array();
+
+		foreach ( $this->get_adjacent_deduplicated_links() as $link ) {
```
westonruter: This foreach has quite a bit of duplication with `get_html()`. Can the common code be factored out into a private helper method?

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -166,6 +166,37 @@ public function get_html(): string {
...
+			}
+			if ( null !== $link['maximum_viewport_width'] && PHP_INT_MAX !== $link['maximum_viewport_width'] ) {
+				$media_features[] = sprintf( '(max-width: %dpx)', $link['maximum_viewport_width'] );
+			}
+			$link['attributes']['media'] = implode( ' and ', $media_features );
+
+			$link_header = '<' . esc_url( $link['attributes']['href'] ?? '' ) . '>; rel="preload"';
+			foreach ( $link['attributes'] as $name => $value ) {
+				if ( 'href' !== $name ) {
+					$link_header .= sprintf( '; %s="%s"', $name, esc_attr( $value ) );
```
westonruter: Use of `esc_attr()` isn't right here since this is not an HTML context.  I believe you should instead percent-encode the value here using `rawurlencode()`.

westonruter (changes requested):
Thank you! Great start.

westonruter:
Oh, we'll also need to add tests for this new method.

plugins/optimization-detective/class-od-preload-link-collection.php
AhmarZaidi: Sure makes sense, will do that

plugins/optimization-detective/optimization.php
AhmarZaidi: Hey, so I tried this but encountered errors while running tests.
```bash
➜  wordpress-performance git:(preload-links) npm run test-php
...
There were 9 errors:

1) Test_Image_Prioritizer_Helper::test_image_prioritizer_register_tag_visitors with data set "common-lcp-image-and-lazy-loaded-image-outside-viewport-with-fully-populated-sample-data" (Closure Object (...), '\n                                      <html lang="en">\n                              ...\n                           ', '\n <html lang="en">\n                               ...\n                           ')
Cannot modify header information - headers already sent by (output started at /wordpress-phpunit/includes/bootstrap.php:265)

2) Test_Image_Prioritizer_Helper::test_image_prioritizer_register_tag_visitors with data set "common-lcp-image-with-fully-incomplete-sample-data" (Closure Object (...), '\n                                    <html lang="en">\n                              ...\n                           ', '\n                                  <html lang="en">\n                              ...\n                           ')
Cannot modify header information - headers already sent by (output started at /wordpress-phpunit/includes/bootstrap.php:265)
...
``` 

This is the line in `/wordpress-phpunit/includes/bootstrap.php:265` referenced by the error where the output is started:
```
system( WP_PHP_BINARY . ' ' . escapeshellarg( __DIR__ . '/install.php' ) . ' ' . escapeshellarg( $config_file_path ) . ' ' . $ms_tests . ' ' . $core_tests, $retval );
```

I believe [this section](https://github.com/WordPress/performance/blob/trunk/plugins/optimization-detective/optimization.php#L176-L190) of the `plugins/optimization-detective/optimization.php` is producing some output but `headers()` should be called before any output because after starting to write the http content, headers cannot be sent in the http response as mentioned [here](https://stackoverflow.com/a/8028987/19726929).

Can we move the `$walker->append_head_html( $preload_links->get_html() );` to the above section?
Or is there another solution to this?

Also, should I remove the check for `headers_sent()`?

westonruter: Oh yes, keep the `headers_sent()` check. The issue is that PHPUnit itself is printing output, so it is preventing headers from being sent.

westonruter: So like:

```php
// Send any preload links in a Link response header and in a LINK tag injected at the end of the HEAD.
if ( count( $preload_links ) > 0 ) {
	if ( ! headers_sent() ) {
		header( $preload_links->get_response_header() );
	}
	$walker->append_head_html( $preload_links->get_html() );
}
```

westonruter: Oh, one more important thing: make sure to pass `false` as the second argument to the `header()` function. This will prevent it from overriding any existing `Link` headers that may have been sent.

AhmarZaidi: Thanks, will implement these changes.

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -136,6 +136,27 @@ static function ( array $carry, array $link ): array {
 		);
 	}
 
+	/**
+	 * Adds media features to the links.
+	 *
+	 * @phpstan-param Link $link
+	 * @param array $link Link array.
+	 * @phpstan-return Link
+	 * @return array Link array with media features added.
```
westonruter:
```suggestion
	 * @param Link $link Link array.
	 * @return Link Array with media features added.
```

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -145,14 +166,7 @@ public function get_html(): string {
...
 		foreach ( $this->get_adjacent_deduplicated_links() as $link ) {
-			$media_features = array( 'screen' );
-			if ( null !== $link['minimum_viewport_width'] && $link['minimum_viewport_width'] > 0 ) {
-				$media_features[] = sprintf( '(min-width: %dpx)', $link['minimum_viewport_width'] );
-			}
-			if ( null !== $link['maximum_viewport_width'] && PHP_INT_MAX !== $link['maximum_viewport_width'] ) {
-				$media_features[] = sprintf( '(max-width: %dpx)', $link['maximum_viewport_width'] );
-			}
-			$link['attributes']['media'] = implode( ' and ', $media_features );
+			$link = $this->add_media_features( (array) $link );
```
westonruter: Is the casting necessary?

```suggestion
			$link = $this->add_media_features( $link );
```

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -136,6 +136,27 @@ static function ( array $carry, array $link ): array {
...
+		if ( null !== $link['minimum_viewport_width'] && $link['minimum_viewport_width'] > 0 ) {
+			$media_features[] = sprintf( '(min-width: %dpx)', $link['minimum_viewport_width'] );
+		}
+		if ( null !== $link['maximum_viewport_width'] && PHP_INT_MAX !== $link['maximum_viewport_width'] ) {
+			$media_features[] = sprintf( '(max-width: %dpx)', $link['maximum_viewport_width'] );
+		}
+		$link['attributes']['media'] = implode( ' and ', $media_features );
+
+		return (array) $link;
+	}
```
westonruter: However, instead of amending the link, how about just returning the media features string instead?

```suggestion
	private function get_media_attribute( array $link ): string {
		$media_features = array( 'screen' );
		if ( null !== $link['minimum_viewport_width'] && $link['minimum_viewport_width'] > 0 ) {
			$media_features[] = sprintf( '(min-width: %dpx)', $link['minimum_viewport_width'] );
		}
		if ( null !== $link['maximum_viewport_width'] && PHP_INT_MAX !== $link['maximum_viewport_width'] ) {
			$media_features[] = sprintf( '(max-width: %dpx)', $link['maximum_viewport_width'] );
		}
		return implode( ' and ', $media_features );
	}
```

plugins/optimization-detective/class-od-preload-link-collection.php
westonruter: Alternatively, if the above suggestion is accepted, then:

```suggestion
			$link['media] = $this->get_media_attribute( $link );
```

plugins/optimization-detective/class-od-preload-link-collection.php
westonruter: Further still, what if this logic were rather incorporated into `get_adjacent_deduplicated_links()`? Instead of returning the result of the `array_reduce()` it could then loop over the links and construct the `media` attribute for each. This method could then be changed from `get_adjacent_deduplicated_links()` to something like `prepare_links()`.

plugins/optimization-detective/class-od-preload-link-collection.php
westonruter: Nevermind, I guess these are needed.

AhmarZaidi: Yes, I was looking for a better way but kept getting either phpstan or linter error without this.

plugins/optimization-detective/class-od-preload-link-collection.php
AhmarZaidi: This approach seems better. Will update the `get_adjacent_deduplicated_links()` function.

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -76,14 +76,14 @@ public function add_link( array $attributes, ?int $minimum_viewport_width, ?int
 	}
 
 	/**
-	 * Get adjacent-deduplicated links.
+	 * Prepare links by deduplicating adjacent links and adding media attributes.
```
westonruter:
```suggestion
	 * Prepares links by deduplicating adjacent links and adding media attributes.
```

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -76,14 +76,14 @@ public function add_link( array $attributes, ?int $minimum_viewport_width, ?int
...
 	 *
 	 * When two links are identical except for their minimum/maximum widths which are also consecutive, then merge them
-	 * together.
+	 * together. Also, add media attributes to the links.
 	 *
-	 * @return array<int, Link> Links with adjacent-duplicates merged together.
+	 * @return array<int, Link> Prepared links with adjacent-duplicates merged together and media attributes added.
 	 */
-	private function get_adjacent_deduplicated_links(): array {
+	private function prepare_links(): array {
```
westonruter: I know I suggested `prepare_links()` but could we go with `get_prepared_links()` instead? I think that would be a _tiny_ bit better.

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -100,7 +100,8 @@ static function ( array $a, array $b ): int {
 			}
 		);
 
-		return array_reduce(
+		// Deduplicating adjacent links.
```
westonruter:
```suggestion
		// Deduplicate adjacent links.
```

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -166,6 +172,31 @@ public function get_html(): string {
...
+	/**
+	 * Constructs the Link HTTP response header.
+	 *
+	 * @return string|null Link HTTP response header, or null if there are none.
+	 */
+	public function get_response_header(): ?string {
+		$link_headers = array();
+
+		foreach ( $this->prepare_links() as $link ) {
+			$link_header = '<' . esc_url_raw( $link['attributes']['href'] ?? '' ) . '>; rel="preload"';
```
westonruter: Is it valid to send a `Link` without the header? It seems that the [uri-reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link#uri-reference) is required. In reality, the only reason why `href` was made optional was for the sake of preloading responsive images since Safari at the time didn't support `imagesrcset` and `imagesizes`, so adding a preload link that omitted the `href` was a way to prevent Safari from needlessly preloading an image which might not be used. 

Nevertheless, as seen in [this web.dev article](https://web.dev/articles/preload-responsive-images#image-set:~:text=Responsive%20image%20preloading%20lets%20you%20load%20those%20images%20faster.), there is still a use case for a preload link without `href` and that is for preloading background images which have different pixel densities:

```html
<link rel="preload" as="image" imagesrcset="cat.png 1x, cat-2x.png 2x">
```

So it seems we do need `href` to remain able to be optional.

What to do in the case where there is no `href`?  I just made a [test glitch](https://imagesrcset-only-preload-link-http-header.glitch.me/) to check various scenarios. You can see that [when an empty string is used](https://imagesrcset-only-preload-link-http-header.glitch.me/?preload_url_type=empty-string), there is nothing downloaded:

image

In contrast, check what happens [when `about:blank` is used as the URL](https://imagesrcset-only-preload-link-http-header.glitch.me/?preload_url_type=about-blank):

image

It works!

So let's do this:

```suggestion
			// The about:blank is present since a Link without a reference-uri is invalid so any imagesrcset would otherwise not get downloaded.
			$link_header = '<' . ( isset( $link['attributes']['href'] ) ? esc_url_raw( $link['attributes']['href'] ) : 'about:blank' ) . '>; rel="preload"';
```

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -166,6 +172,31 @@ public function get_html(): string {
...
+			foreach ( $link['attributes'] as $name => $value ) {
+				if ( 'href' !== $name ) {
+					$link_header .= sprintf( '; %s="%s"', $name, rawurlencode( $value ) );
+				}
```
westonruter:
```suggestion
			unset( $link['attributes] );
			foreach ( $link['attributes'] as $name => $value ) {
				$link_header .= sprintf( '; %s="%s"', $name, rawurlencode( $value ) );
```

westonruter (changes requested):
Getting close!

plugins/optimization-detective/class-od-preload-link-collection.php
westonruter: I also checked and using `about:blank` here works in Firefox and Safari too. This doesn't seem to be part of the spec, so it could be an implementation accident. But the alternative is to skip sending the `Link` altogether. So I think using `about:blank` is best.

AhmarZaidi: I see, thanks this was very insightful.

AhmarZaidi:
> Oh, we'll also need to add tests for this new method.

Had some doubts regarding tests implementation:

1. I believe [`get_html()`](https://github.com/WordPress/performance/blob/trunk/plugins/optimization-detective/class-od-preload-link-collection.php#L144) is tested by using the output of [`od_optimize_template_output_buffer()`](https://github.com/WordPress/performance/blob/1ed2fbdab5fa47e2b174b813f6d5d3273b589482/plugins/optimization-detective/optimization.php#L143) since the walker adds links to the markup and compares it with the data from data provider.
Now, how would I test for `get_response_header()` since they are added to the http response and not the markup.
I explored `xdebug_get_headers()` but that doesn't seem to work. Could separately test `get_response_header()` function's output but I don't think that's the right way.

2. If we implement it like this in `optimization.php`: 
> ```php
> // Send any preload links in a Link response header and in a LINK tag injected at the end of the HEAD.
> if ( count( $preload_links ) > 0 ) {
> 	if ( ! headers_sent() ) {
> 		header( $preload_links->get_response_header() );
> 	}
> 	$walker->append_head_html( $preload_links->get_html() );
> }
> ```
Then `headers()` won't be called while testing because `headers_sent()` would be true due to [prior output](https://github.com/WordPress/performance/pull/1323#discussion_r1661437951), so how would we test it?

plugins/optimization-detective/class-od-preload-link-collection.php
AhmarZaidi: Shouldn't we add a check if `$link['attributes']['href']` is set before unsetting?
```php
if ( isset( $link['attributes']['href'] ) ) {
    unset( $link['attributes']['href'] );
}
foreach ( $link['attributes'] as $name => $value ) {
    $link_header .= sprintf( '; %s="%s"', $name, rawurlencode( $value ) );
}
```

AhmarZaidi: How about something like this:
```php
$link['attributes']['href'] = isset( $link['attributes']['href'] ) ? esc_url_raw( $link['attributes']['href'] ) : 'about:blank';
$link_header = '<' . $link['attributes']['href'] . '>; rel="preload"';
unset( $link['attributes']['href'] );
foreach ( $link['attributes'] as $name => $value ) {
    $link_header .= sprintf( '; %s="%s"', $name, rawurlencode( $value ) );
}
```

westonruter: Yes, that looks good.

westonruter:
@AhmarZaidi Yeah, testing the output of `header()` is a pain. I don't think we can practically test it being output when calling `od_optimize_template_output_buffer()`. I think it's fine to just test that `get_response_header()` method in isolation.

plugins/optimization-detective/tests/test-optimization.php
```diff
@@ -368,6 +368,44 @@ function ( OD_HTML_Tag_Walker $walker, OD_URL_Metrics_Group_Collection $url_metr
 		$this->assertEquals( $expected, $buffer );
 	}
 
+	/**
+	 * Test get_response_header().
+	 *
+	 * @covers \OD_Preload_Link_Collection::get_response_header
```
westonruter:
```suggestion
	 * @covers OD_Preload_Link_Collection::get_response_header
```

plugins/optimization-detective/tests/test-optimization.php
```diff
@@ -368,6 +368,44 @@ function ( OD_HTML_Tag_Walker $walker, OD_URL_Metrics_Group_Collection $url_metr
...
+				'fetchpriority' => 'high',
+				'imagesrcset'   => 'https://example.com/bar-480w.jpg 480w, https://example.com/bar-800w.jpg 800w',
+				'imagesizes'    => '(max-width: 600px) 480px, 800px',
+				'crossorigin'   => 'anonymous',
+			),
+			600,
+			1200
+		);
+
+		$expected_header = 'Link: <https://example.com/foo.jpg>; rel="preload"; as="image"; fetchpriority="high"; imagesrcset="https%3A%2F%2Fexample.com%2Ffoo-480w.jpg%20480w%2C%20https%3A%2F%2Fexample.com%2Ffoo-800w.jpg%20800w"; imagesizes="%28max-width%3A%20600px%29%20480px%2C%20800px"; crossorigin="anonymous"; media="screen", <https://example.com/bar.jpg>; rel="preload"; as="image"; fetchpriority="high"; imagesrcset="https%3A%2F%2Fexample.com%2Fbar-480w.jpg%20480w%2C%20https%3A%2F%2Fexample.com%2Fbar-800w.jpg%20800w"; imagesizes="%28max-width%3A%20600px%29%20480px%2C%20800px"; crossorigin="anonymous"; media="screen%20and%20%28min-width%3A%20600px%29%20and%20%28max-width%3A%201200px%29"';
```
westonruter: The `imagesrcset`, `imagesizes`, and `media` attribute values here seem wrong being URL-encoded. Do these work being encoded like this?

plugins/optimization-detective/class-od-preload-link-collection.php
```diff
@@ -166,6 +172,32 @@ public function get_html(): string {
...
+	public function get_response_header(): ?string {
+		$link_headers = array();
+
+		foreach ( $this->get_prepared_links() as $link ) {
+			// The about:blank is present since a Link without a reference-uri is invalid so any imagesrcset would otherwise not get downloaded.
+			$link['attributes']['href'] = isset( $link['attributes']['href'] ) ? esc_url_raw( $link['attributes']['href'] ) : 'about:blank';
+			$link_header                = '<' . $link['attributes']['href'] . '>; rel="preload"';
+			unset( $link['attributes']['href'] );
+			foreach ( $link['attributes'] as $name => $value ) {
+				$link_header .= sprintf( '; %s="%s"', $name, rawurlencode( $value ) );
```
westonruter: Per below, we may need to reconsider the use of `rawurlencode()` here. Maybe it's just that the generated header test "looks" bad.

plugins/optimization-detective/tests/test-optimization.php
westonruter: Indeed. I found it doesn't work, for `imagesrcset` at least. I tried sending that `Link` header on a Glitch and I saw a network request attempted for:

```
https://alert-sustaining-acrylic.glitch.me/https%3A%2F%2Fexample.com%2Ffoo-480w.jpg%20480w%2C%20https%3A%2F%2Fexample.com%2Ffoo-800w.jpg%20800w
```

westonruter: I think I got this. See d1cd8d4

westonruter (approved):

westonruter:
Great work!

Issue #1259, opened 2024-05-30
Improve image handling approach in Modern Image Formats


adamsilverstein:
## Feature Description

Switch from using `the_content` filter to using `wp_calculate_image_srcset` when replacing images.	

As @joemcgill points out in https://github.com/WordPress/performance/pull/1236#pullrequestreview-2069741635 we should probably switch the image processing approach. Currently the `webp_uploads_update_image_references` function is attached to `the_content` filter and processes all images on the page using a regex. Instead, we can leverage the filters core runs on images after it does a similar regex against the content. Namely, the `wp_calculate_image_srcset` or `wp_content_img_tag` filters are potential places to add our logic.

westonruter:
And note that `webp_uploads_wrap_image_in_picture()` is necessarily using the `wp_content_img_tag` filter.

joemcgill:
Looking at this again today, I'd prefer using the `wp_filter_content_tags` filter as it would be the most straightforward change.

b1ink0:
@joemcgill  
I believe the `wp_filter_content_tags` filter does not actually exist in the core, but there is a function named `wp_filter_content_tags()` which processes images and iframes. As described in the doc comment for `wp_filter_content_tags()`  function, it is intended to convert old HTML markup to new HTML5 markup [ref](https://github.com/WordPress/wordpress-develop/blob/2ba84334c9d412956322544564c1a8e32fb56363/src/wp-includes/media.php#L1861).

Instead, I believe the `wp_content_img_tag` filter would be a better option. This filter is defined inside the `wp_filter_content_tags()` function, which already performs regex to find the `img` tags. Using this filter would eliminate the need for additional regex that we are doing in the `webp_uploads_update_image_references` function.

PR #1325, opened 2024-07-01
Upgrade PHPStan to 1.11.6

westonruter:
This upgrades PHPStan to 1.11.6 which introduced two new issues.

```
 ------ -------------------------------------------------------------------------------------------------------------------- 
  Line   plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php                           
 ------ -------------------------------------------------------------------------------------------------------------------- 
  44     Offset 'background_image' might not exist on array{0?: string, 1?: string, background_image?: string, 2?: string}.  
 ------ -------------------------------------------------------------------------------------------------------------------- 
```

This is a case where PHPStan is not aware that at this point, the array keys are actually not nullable at this point. This is a bug in PHPStan which I've filed as https://github.com/phpstan/phpstan/issues/11262. The issue seems to be that if the return value of `preg_match()` is casted to a `bool` or `int` then the array shape doesn't come through. So it's now checking to see if the return value is exactly `1` instead. Additionally, another check for `'' !== $matches['background_image']` will be able to be removed once https://github.com/phpstan/phpstan/issues/11222 is implemented, since we'll know that the array shape here has a `non-empty-string`.

```
 ------ ----------------------------------------------------------------- 
  Line   plugins/webp-uploads/hooks.php                                   
 ------ ----------------------------------------------------------------- 
  547    Variable $class_name in empty() always exists and is not falsy.  
 ------ -----------------------------------------------------------------
```

There was addressing the second `if` statement in the following code:

https://github.com/WordPress/performance/blob/b71d0a0eab14ceaed86b45a17899284fb0a14dc3/plugins/webp-uploads/hooks.php#L542-L549

It is a useless condition because since `preg_match()` is known to have succeeded, the `$class_name` will never be empty.

This fixes the issue by removing that condition, but also makes the code more robust by leveraging the HTML Tag Processor. It also adds missing test assertions for `webp_uploads_update_image_references()`.

### DIFF
--- a/plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
@@ -39,11 +39,9 @@ final class Image_Prioritizer_Background_Image_Styled_Tag_Visitor extends Image_
 		if (
 			is_string( $style )
 			&&
-			0 < (int) preg_match( '/background(?:-image)?\s*:[^;]*?url\(\s*[\'"]?\s*(?<background_image>.+?)\s*[\'"]?\s*\)/', $style, $matches )
+			preg_match( '/background(?:-image)?\s*:[^;]*?url\(\s*[\'"]?\s*(?<background_image>.+?)\s*[\'"]?\s*\)/', $style, $matches ) // Not compatible with strict rules. See <https://github.com/phpstan/phpstan/issues/11262>.
 			&&
-			isset( $matches['background_image'] ) // PHPStan should ideally know that this is set since the above preg_match() returned successfully.
-			&&
-			'' !== $matches['background_image'] // PHPStan should ideally know that this is a non-empty string based on the `.+?` regular expression.
+			'' !== $matches['background_image'] // PHPStan should ideally know that this is a non-empty string based on the `.+?` regular expression. See <https://github.com/phpstan/phpstan/issues/11222>.
 			&&
 			! $this->is_data_url( $matches['background_image'] )
 		) {
--- a/plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
@@ -41,6 +41,8 @@ final class Image_Prioritizer_Background_Image_Styled_Tag_Visitor extends Image_
 			&&
 			0 < (int) preg_match( '/background(-image)?\s*:[^;]*?url\(\s*[\'"]?\s*(?<background_image>.+?)\s*[\'"]?\s*\)/', $style, $matches )
 			&&
+			isset( $matches['background_image'] ) // PHPStan should ideally know that this is set since the above preg_match() returned successfully.
+			&&
 			'' !== $matches['background_image'] // PHPStan should ideally know that this is a non-empty string based on the `.+?` regular expression.
 			&&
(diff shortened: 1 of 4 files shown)


plugins/webp-uploads/hooks.php
```diff
@@ -533,29 +533,28 @@ function webp_uploads_update_image_references( $content ): string {
...
+		if ( ! $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
+			// This condition won't ever be met since we're iterating over the IMG tags extracted with preg_match_all() above.
 			continue;
 		}
 
-		if ( empty( $class_name ) ) {
+		// Find the ID of each image by the class.
+		// TODO: It would be preferable to use the $processor->class_list() method but there seems to be some typing issues with PHPStan.
+		$class_name = $processor->get_attribute( 'class' );
+		if ( ! is_string( $class_name ) || ! preg_match( '/(?:^|\s)wp-image-([1-9]\d*)(?:\s|$)/i', $class_name, $matches ) ) {
```
westonruter, line 552: This is more robust than before since it constraints looking for the class name inside the `class` attribute. Previously the string could be anywhere in the `img` tag.

plugins/webp-uploads/hooks.php
```diff
@@ -533,29 +533,28 @@ function webp_uploads_update_image_references( $content ): string {
...
 			continue;
 		}
 
 		// Make sure we use the last item on the list of matches.
-		$attachment_id = (int) $class_name[1];
-
-		if ( ! $attachment_id ) {
```
westonruter, line 554: This will never be zero because the regex is now `[1-9]\d*`.

plugins/webp-uploads/hooks.php
```diff
@@ -533,29 +533,28 @@ function webp_uploads_update_image_references( $content ): string {
 	}
 
 	// This content does not have any tag on it, move forward.
+	// TODO: Eventually this should use the HTML API to parse out the image tags and then update them.
```
adamsilverstein, line 536: or, switch to using the `wp_content_img_tag` filter? https://github.com/WordPress/performance/issues/1259

adamsilverstein:
Looks good! some test failing?

westonruter:
@adamsilverstein I re-ran the failed jobs. They're passing now. Seems like a wp-env hiccup.

adamsilverstein (approved):
Spiffy!

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
```diff
@@ -39,9 +39,9 @@ public function __invoke( OD_HTML_Tag_Walker $walker ): bool {
 		if (
 			is_string( $style )
 			&&
-			0 < (int) preg_match( '/background(-image)?\s*:[^;]*?url\(\s*[\'"]?\s*(?<background_image>.+?)\s*[\'"]?\s*\)/', $style, $matches )
+			1 === preg_match( '/background(?:-image)?\s*:[^;]*?url\(\s*[\'"]?\s*(?<background_image>.+?)\s*[\'"]?\s*\)/', $style, $matches ) // Not compatible with strict rules. See <https://github.com/phpstan/phpstan/issues/11262>.
```
westonruter, line 42: Bah. I should have removed the comment as it is now irrelevant after 75cd653dd3853f9038f5e26abae3f96256ddb9a3 per https://github.com/phpstan/phpstan/issues/11262#issuecomment-2200728974. I'll do so as part of another PR.

PR #1328, opened 2024-07-02
Fix Modern Image Formats picture element support for selected image sizes

### LINKED ISSUE #1316: Picture element images: Image size issue
## Bug Description

When using the new Modern Image Formats picture element support, I noticed that the markup didn't account the image sizes that are selected from editor.

## Steps to reproduce

1. Activate the Modern Image Formats plugin and enable picture support from the Media Settings screen.
2. Add an image block, Select image size to `Thumbnail`
3. Preview the page in the front end it load `Full` size image instead `Thumbnail` also the sources is missing for thumb size only. The image width and height get selected image values accordingly.

### Editor
Screenshot 2024-06-24 at 5 02 02 PM

### Frontend:
Screenshot 2024-06-24 at 5 03 26 PM

imrraaj:
## Description
This PR addresses an issue with the Modern Image Formats plugin's picture element support, where the generated markup did not account for the image sizes selected from the editor.
Fixes #1316 

## Problem
When using the Modern Image Formats plugin with picture element support enabled, the generated markup would load the full-size image instead of the selected image size (e.g., Thumbnail) and the <source> elements were missing for the selected image size.

## Solution
To fix this issue, the PR introduces the following changes:
- When the `$sizes` is false or `$image_srcset` is empty, then `$image_srcset` should be set to the modern image format version of the `$src`.
- `$image_srcset` then correctly generates the source tags and correct image size is set based on the value selected in the editor

## Testing
All the test are passing locally.

## Outcome

Screenshot 2024-07-02 at 6 15 07 PM

### DIFF
--- a/plugins/webp-uploads/tests/test-picture-element.php
+++ b/plugins/webp-uploads/tests/test-picture-element.php
@@ -111,11 +111,10 @@ class Test_WebP_Uploads_Picture_Element extends TestCase {
 
 		// Check that the generated HTML contains the expected image URL and has the correct class.
 		$this->assertStringContainsString( $thumbnail_url, $the_image );
-		$this->assertStringContainsString( "wp-picture-{$attachment_id}", $the_image );
 
 		// Ensure that the source element exists in the picture element and has the correct mime type.
-		$webp_source = sprintf( '<source type="image/webp"', $attachment_id );
-		$jpg_source  = sprintf( '<source type="image/jpeg"', $attachment_id );
+		$webp_source = '<source type="image/webp"';
+		$jpg_source  = '<source type="image/jpeg"';
 		$this->assertStringContainsString( $webp_source, $the_image );
 		$this->assertStringContainsString( $jpg_source, $the_image );
 	}
--- a/plugins/webp-uploads/tests/test-picture-element.php
+++ b/plugins/webp-uploads/tests/test-picture-element.php
@@ -83,4 +83,40 @@ class Test_WebP_Uploads_Picture_Element extends TestCase {
 			),
 		);
 	}
+
+	/**
+	 * Test the correct image size rendered in the picture element.
+	 */
+	public function test_thumbnail_image_size_in_picture_element(): void {
+		// Enable necessary plugin settings.
+		$this->opt_in_to_jpeg_and_webp();
+		$this->opt_in_to_picture_element();
+
+		// Create an image attachment with jpeg image.
+		$attachment_id = self::factory()->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/data/images/car.jpeg' );
+
+		$image_size = 'thumbnail';
+
+		// Get the thumbnail size URL directly (simulate Gutenberg usage).
+		$thumbnail_url = wp_get_attachment_image_url( $attachment_id, $image_size );
+
+		// Check if the URL is not empty and contains the expected thumbnail size dimension.
+		$this->assertNotEmpty( $thumbnail_url );
+
+		// Fetch the HTML output of the image in Gutenberg.
+		$the_image = wp_get_attachment_image( $attachment_id, $image_size, false, array( 'class' => "wp-image-{$attachment_id}" ) );
+
+		// Apply the wp_content_img_tag filter.
+		$the_image = apply_filters( 'wp_content_img_tag', $the_image, 'the_content', $attachment_id );
+
+		// Check that the generated HTML contains the expected image URL and has the correct class.
+		$this->assertStringContainsString( $thumbnail_url, $the_image );
+		$this->assertStringContainsString( "wp-picture-{$attachment_id}", $the_image );
+
+		// Ensure that the source element exists in the picture element and has the correct mime type.
+		$webp_source = sprintf( '<source type="image/webp"', $attachment_id );
+		$jpg_source  = sprintf( '<source type="image/jpeg"', $attachment_id );
+		$this->assertStringContainsString( $webp_source, $the_image );
+		$this->assertStringContainsString( $jpg_source, $the_image );
+	}
 }
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -98,7 +98,7 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	foreach ( $mime_types as $image_mime_type ) {
 		$sizes = wp_calculate_image_sizes( $size_array, $src, $image_meta, $attachment_id );
 		if ( false === $sizes ) {
-			$image_srcset = $src;
+			continue;
 		}
 
 		// Filter core's wp_get_attachment_image_srcset to return the sources for the current mime type.
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -98,7 +98,7 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	foreach ( $mime_types as $image_mime_type ) {
 		$sizes = wp_calculate_image_sizes( $size_array, $src, $image_meta, $attachment_id );
 		if ( false === $sizes ) {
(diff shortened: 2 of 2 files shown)


westonruter:
This PR should also include some test cases that demonstrate the original bug. Perhaps start there before adding the required fix, for some good ol' test-driven development.

imrraaj:
Based on my observations, the issue only affects JPG and JPEG images. Images with different MIME types are working as expected. The issue arises when the user has a JPG or JPEG image and the image size is set to `thumbnail`. When we select the thumbnail image size, it does not output the source tags which contain information about the `srcset` attribute, ultimately rendering the entire image as is with the given height and width. All the image sizes apart from thumbnail work fine (i.e., medium, large, full size).
I have written a test case just to demonstrate this. It will fail on the current state of the main branch, requiring the fix to be applied.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -83,4 +83,39 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
+		$the_image = apply_filters( 'wp_content_img_tag', $the_image, 'the_content', $attachment_id );
+
+		// Check that the generated HTML contains the expected image URL and has the correct class.
+		$this->assertStringContainsString( $thumbnail_url, $the_image );
+
+		// Ensure that the source element exists in the picture element and has the correct mime type.
+		$webp_source = '<source type="image/webp"';
+		$jpg_source  = '<source type="image/jpeg"';
+		$this->assertStringContainsString( $webp_source, $the_image );
+		$this->assertStringContainsString( $jpg_source, $the_image );
```
westonruter, line 119: Could you rather do a single `assertSame` assertion? By checking the entire string we can be extra confident it contains everything we expect.

adamsilverstein:
I verified the change worked as expected. I'm curious why thumbnail sizes in particular are handled this way. 

I checked with default WordPress without the plugin and noticed that no `srcset` is output when I select the thumbnail size, only the image tag:
image

If I select the medium size, I get a full srcset: 
image

This explains why the bug exists; for the picture element support we want to always have these srcsets so we can provide a fallback. I'm going to dig a little more to see if I can tell why core doesn't use `srcset` for thumbnails (or @joemcgill may know?) - probably because the images are small already so adding additional sizes doesn't make sense?

westonruter:
> I'm going to dig a little more to see if I can tell why core doesn't use `srcset` for thumbnails - probably because the images are small already so adding additional sizes doesn't make sense?

In looking at `wp_calculate_image_srcset()` I see it has these comments:

```
 	 * Loop through available images. Only use images that are resized
	 * versions of the same edit.
```
```
	 // If the image dimensions are within 1px of the expected size, use it.
```

Since thumbnails are square, as I understand they would only get a populated `srcset` if the original image was also a square.

I just tried uploading a square image and then inserting it with the thumbnail resolution, and the generated markup was:

```html
<img 
  decoding="async" 
  width="150" 
  height="150" 
  src="http://localhost:8888/wp-content/uploads/2024/07/pexels-mark-a-jenkins-373676419-21771035-150x150.jpg" 
  alt="" 
  class="not-transparent wp-image-43" 
  srcset="http://localhost:8888/wp-content/uploads/2024/07/pexels-mark-a-jenkins-373676419-21771035-150x150.jpg 150w, http://localhost:8888/wp-content/uploads/2024/07/pexels-mark-a-jenkins-373676419-21771035-300x300.jpg 300w, http://localhost:8888/wp-content/uploads/2024/07/pexels-mark-a-jenkins-373676419-21771035-1024x1024.jpg 1024w, http://localhost:8888/wp-content/uploads/2024/07/pexels-mark-a-jenkins-373676419-21771035-768x768.jpg 768w, http://localhost:8888/wp-content/uploads/2024/07/pexels-mark-a-jenkins-373676419-21771035-1536x1536.jpg 1536w, http://localhost:8888/wp-content/uploads/2024/07/pexels-mark-a-jenkins-373676419-21771035-2048x2048.jpg 2048w, http://localhost:8888/wp-content/uploads/2024/07/pexels-mark-a-jenkins-373676419-21771035-1568x1568.jpg 1568w" 
  sizes="(max-width: 150px) 100vw, 150px">
```

So here, even though the thumbnail was inserted, it _did_ get a fully-populated `srcset` because the other image sizes are also square.

adamsilverstein:
> it did get a fully-populated srcset because the other image sizes are also square.

this makes me wonder if the fix proposed here is correct. If a non-square image is uploaded, i suspect the fallbacks specified in the srcset entries will be distorted when displayed. Since no srcset is provided, we could instead use the 

maybe we could construct srcsets with the original image urls in each format (untested):

so for example



```
<picture class="wp-picture-407" style="display: contents;">
<img width="150" height="150" src="https://wpdev.localhost/wp-content/uploads/2024/07/cat-2-150x150.jpg" class="attachment-150x150 size-150x150" alt="">
</picture>
```
would become
```
<picture class="wp-picture-407" style="display: contents;">
<source type="image/avif" src="https://wpdev.localhost/wp-content/uploads/2024/07/cat-2-150x150.avif">
<source type="image/jpeg" src="https://wpdev.localhost/wp-content/uploads/2024/07/cat-2-150x150.jpg">
<img width="150" height="150" src="https://wpdev.localhost/wp-content/uploads/2024/07/cat-2-150x150.jpg" class="attachment-150x150 size-150x150" alt=""></picture>

note the `-150x150` suffix, including for the `img` tag. I'm not sure if the current PR uses that image correctly.
```

adamsilverstein:
> it did get a fully-populated srcset because the other image sizes are also square.

this makes me wonder if the fix proposed here is correct. If a non-square image is uploaded, i suspect the fallbacks specified in the srcset entries will be distorted when displayed. Since no srcset is provided, we could instead use the 

maybe we could construct srcsets with the original image urls in each format (untested):

so for example



```
<picture class="wp-picture-407" style="display: contents;">
<img width="150" height="150" src="https://wpdev.localhost/wp-content/uploads/2024/07/cat-2-150x150.jpg" class="attachment-150x150 size-150x150" alt="">
</picture>
```
would become
```
<picture class="wp-picture-407" style="display: contents;">
<source type="image/avif" src="https://wpdev.localhost/wp-content/uploads/2024/07/cat-2-150x150.avif">
<source type="image/jpeg" src="https://wpdev.localhost/wp-content/uploads/2024/07/cat-2-150x150.jpg">
<img width="150" height="150" src="https://wpdev.localhost/wp-content/uploads/2024/07/cat-2-150x150.jpg" class="attachment-150x150 size-150x150" alt=""></picture>
```

note the `-150x150` suffix, including for the `img` tag. I'm not sure if the current PR uses that image correctly.

adamsilverstein (changes requested):
Needs some refinement to use only the cropped images (with matching sizes) - see comment thread

westonruter:
Punting this to the next release.

imrraaj:
So I tried uploading a non-square image to test the code in the PR. When the thumbnail size is selected, it generates following HTML. 
```html
<picture class="wp-picture-267" style="display: contents;">
  <source type="image/webp" srcset="http://contrib.local/wp-content/uploads/2024/07/lake-8847415_1280-150x150.jpg" sizes="(max-width: 150px) 100vw, 150px">
  <source type="image/jpeg" srcset="http://contrib.local/wp-content/uploads/2024/07/lake-8847415_1280-150x150.jpg" sizes="(max-width: 150px) 100vw, 150px">
  <img width="150" height="150" src="http://contrib.local/wp-content/uploads/2024/07/lake-8847415_1280.jpg" class="attachment-150x150 size-150x150" alt="">
</picture>
```
So as per @adamsilverstein's suggestion along with source urls, the `img` tag url should also have `-150x150` suffix. Am I getting this right?

mukeshpanchal27:
@westonruter @adamsilverstein This has been solved in #1354 and merged in trunk. The IMG src check is added in #1408.

Closing this PR now.

PR #1333, opened 2024-07-08
Bump minimum WP versions and WP version tested up to

westonruter:
In preparation for the WordPress 6.6 release, we need to update the `Tested up to` field to reflect 6.6. In the same way, we can now bump the `Requires at least` from 6.4 to 6.5. In addition to bumping these versions, this PR also removes obsolete code which was present exclusively for WP 6.4.

This cherry-picks some commits from https://github.com/WordPress/performance/pull/1302.

### DIFF
--- a/.github/workflows/php-test-plugins.yml
+++ b/.github/workflows/php-test-plugins.yml
@@ -42,10 +42,10 @@ jobs:
       fail-fast: false
       matrix:
         php: ['8.2', '8.1', '8.0', '7.4', '7.3', '7.2']
-        wp: [ 'latest' ]
+        wp: [ '6.6-RC1' ] # TODO: Change this back to latest.
         include:
           - php: '7.4'
-            wp: '6.4'
+            wp: '6.5'
           - php: '8.3'
             wp: 'trunk'
     env:
--- a/.github/workflows/php-test-plugins.yml
+++ b/.github/workflows/php-test-plugins.yml
@@ -42,7 +42,7 @@ jobs:
       fail-fast: false
       matrix:
         php: ['8.2', '8.1', '8.0', '7.4', '7.3', '7.2']
-        wp: [ '6.6-RC1' ] # TODO: Change this back to latest.
+        wp: [ '6.6-RC2' ] # TODO: Change this back to latest.
         include:
           - php: '7.4'
             wp: '6.5'
--- a/plugins/auto-sizes/readme.txt
+++ b/plugins/auto-sizes/readme.txt
@@ -1,8 +1,8 @@
 === Auto-sizes for Lazy-loaded Images ===
 
 Contributors:      wordpressdotorg
-Requires at least: 6.4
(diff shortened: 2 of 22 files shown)


.github/workflows/php-test-plugins.yml
```diff
@@ -42,10 +42,10 @@ jobs:
       fail-fast: false
       matrix:
         php: ['8.2', '8.1', '8.0', '7.4', '7.3', '7.2']
-        wp: [ 'latest' ]
+        wp: [ '6.6-RC2' ] # TODO: Change this back to latest.
```
westonruter, line 45: We'll switch this back to `latest` once the 6.6 release is finalized. In the meantime, `trunk` is basically the same, so we're not hurting from not testing a WP version.

plugins/embed-optimizer/hooks.php
```diff
@@ -147,29 +147,6 @@ function embed_optimizer_lazy_load_scripts(): void {
...
- * @param int    $error_level   Optional. The designated error type for this error.
- *                              Only works with E_USER family of constants. Default E_USER_NOTICE.
- */
-function embed_optimizer_trigger_error( string $function_name, string $message, int $error_level = E_USER_NOTICE ): void {
-	if ( ! function_exists( 'wp_trigger_error' ) ) {
-		return;
-	}
-	wp_trigger_error( $function_name, $message, $error_level );
-}
-
```
westonruter, line 172: This is actually WP 6.3-compat code, as `wp_trigger_error()` was introduced in 6.4.

thelovekesh (approved):

plugins/auto-sizes/readme.txt
```diff
@@ -1,8 +1,8 @@
 === Auto-sizes for Lazy-loaded Images ===
 
 Contributors:      wordpressdotorg
-Requires at least: 6.4
-Tested up to:      6.5
+Requires at least: 6.5
+Tested up to:      6.6
```
swissspidy, line 5: FWIW, "Requires at least" and "Requires PHP" headers should only be defined in the main plugin file, as that's where the information is extracted from in the plugin directory. It's redundant to have it in `readme.txt`.

westonruter: I was wondering about that, and I was going to remove `Requires at least` from the readme, but then I saw it was shown in an example in the docs: https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/#readme-header-information

But I skipped over this line:

> Since WordPress 5.8 plugin [readme files are not parsed for requirements](https://core.trac.wordpress.org/ticket/48520). This means that headers Requires PHP and Requires at least are going to be parsed from plugin’s main PHP file.

westonruter: I've opened https://github.com/WordPress/performance/pull/1334 to address this.

PR #1337, opened 2024-07-09
Facilitate embedding of Embed Optimizer

westonruter:
In line with Optimization Detective and Image Prioritizer, this updates the bootstrap logic for Embed Optimizer so that it can be embedded in a theme or plugin.

### DIFF
--- a/plugins/embed-optimizer/load.php
+++ b/plugins/embed-optimizer/load.php
@@ -20,7 +20,57 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
-define( 'EMBED_OPTIMIZER_VERSION', '0.1.2' );
+(
+	/**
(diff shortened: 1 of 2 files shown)


plugins/embed-optimizer/load.php
```diff
@@ -20,7 +20,57 @@
...
+			rtrim( WP_PLUGIN_DIR, '/' ) === dirname( __DIR__ )
+		) {
+			$GLOBALS[ $global_var_name ]['version'] = $version;
+			$GLOBALS[ $global_var_name ]['load']    = $load;
+		}
+	}
+)(
+	'embed_optimizer_pending_plugin',
+	'0.1.2',
+	static function ( string $version ): void {
```
westonruter, line 69: This block of code is a direct copy of what is in Optimization Detective's `load.php`.

swissspidy (approved):

PR #1354, opened 2024-07-17
Fix setting sizes attribute on PICTURE > SOURCE elements

### LINKED ISSUE #1349: Picture element: The accurate sizes improvement for images not working
## Bug Description

The accurate sizes improvement done in #1329 not working for Picture elements.

## Steps to reproduce

1. Activate the Modern Image Formats plugin and enable picture support from the Media Settings screen.
2. Install and activate Enhanced Responsive Images (formally known as auto-sizes)
3. Add image block with none alignment
4. The image tag get proper sizes value but source didn't.

## Screenshots

Screenshot 2024-07-16 at 5 19 19 PM

mukeshpanchal27:
## Summary

Fixes #1349

The PR changes the filter priority for the picture element filter so it will be generated before the sizes attribute updates. Add new filter for picture element in auto-sizes plugin that add accurate sizes change for `source ` tags


westonruter:
Can a test be added which demonstrates the bug if the changes aren't applied?

plugins/auto-sizes/hooks.php
```diff
@@ -216,3 +216,37 @@ function auto_sizes_improve_image_sizes_attributes( string $content ): string {
...
+	}
+
+	$has_image = $processor->next_tag( array( 'tag_name' => 'img' ) );
+	$sizes     = null;
+	// Only update the markup if an image is found.
+	if ( $has_image ) {
+		$sizes = $processor->get_attribute( 'sizes' );
+	}
+
+	if ( ! $sizes ) {
```
westonruter: In anticipation of strict rules: #1241 

```suggestion
	if ( ! is_string( $sizes ) ) {
```

plugins/auto-sizes/hooks.php
```diff
@@ -216,3 +216,37 @@ function auto_sizes_improve_image_sizes_attributes( string $content ): string {
...
+	if ( ! $processor->next_tag( array( 'tag_name' => 'picture' ) ) ) {
+		return $content;
+	}
+
+	$has_image = $processor->next_tag( array( 'tag_name' => 'img' ) );
+	$sizes     = null;
+	// Only update the markup if an image is found.
+	if ( $has_image ) {
+		$sizes = $processor->get_attribute( 'sizes' );
+	}
```
westonruter: This can be simplified:

```suggestion
	// Only update the markup if an image is found.
	if ( ! $processor->next_tag( array( 'tag_name' => 'img' ) ) ) {
		return $content;
	}
	$sizes = $processor->get_attribute( 'sizes' );
```

plugins/auto-sizes/hooks.php
```diff
@@ -216,3 +216,37 @@ function auto_sizes_improve_image_sizes_attributes( string $content ): string {
...
+
+	if ( ! $sizes ) {
+		return $content;
+	}
+
+	$processor = new WP_HTML_Tag_Processor( $content );
+	while ( $processor->next_tag( array( 'tag_name' => 'source' ) ) ) {
+		$processor->set_attribute( 'sizes', $sizes );
```
westonruter: What if there is already a `sizes` attribute on this `source` tag? Shouldn't it do nothing in that case?

```suggestion
		if ( ! is_string( $processor->get_attribute( 'sizes' ) ) ) {
			$processor->set_attribute( 'sizes', $sizes );
		}
```

plugins/auto-sizes/hooks.php
```diff
@@ -216,3 +216,37 @@ function auto_sizes_improve_image_sizes_attributes( string $content ): string {
...
+	if ( $has_image ) {
+		$sizes = $processor->get_attribute( 'sizes' );
+	}
+
+	if ( ! $sizes ) {
+		return $content;
+	}
+
+	$processor = new WP_HTML_Tag_Processor( $content );
+	while ( $processor->next_tag( array( 'tag_name' => 'source' ) ) ) {
```
westonruter: Question: Is the `srcset` getting set on each of the `source` tags, but the `sizes` attribute is not? Why?

plugins/auto-sizes/hooks.php
mukeshpanchal27: The `sizes` attribute is already set https://github.com/WordPress/performance/blob/trunk/plugins/webp-uploads/picture-element.php#L130 for `source` tags. We are updating it to be accurate based on the new API changes.

We take the correct `sizes` value from the `img` tag and update it in the `source` tags.

plugins/auto-sizes/hooks.php
mukeshpanchal27: Both `srcset` and `sizes` set through picture element code but we are updating only `sizes` to more accurate based on our new API.

mukeshpanchal27:
> Can a test be added which demonstrates the bug if the changes aren't applied?

@westonruter the bug only generated when the Enhanced Responsive Images (auto-sizes) plugin also activated.

joemcgill (changes requested):
I don't think this is the most robust approach to fixing this issue. If I am understanding the problem correctly, the issue is caused because the Enhanced Responsive Images plugin modifies the `sizes` attribute using the `wp_content_img_tag` filter, since we're not getting all of the information we need to filter the `wp_calculate_image_sizes()` value directly. 

However, when Modern Image Formats applies `webp_uploads_wrap_image_in_picture()` it adds `sizes` attributes to the `source` elements by calling `wp_calculate_image_sizes()` directly, rather than using the actual `sizes` attribute that is on the `img` element that it is wrapping. 

I think that a better solution would be to update `webp_uploads_wrap_image_in_picture()` to try to copy the `sizes` attribute from the `img` that it's wrapping, if available, rather than using `wp_calculate_image_sizes()` directly. This would ensure that any modifications that were made to the `sizes` attribute, regardless of what filter was used, will be applied to the `source` elements as well.

Alternately, we need to make sure the enhanced `sizes` attribute ends up getting applied via the `wp_calculate_image_sizes` filter so that any other code that calls `wp_calculate_image_sizes()` directly will have the optimizations applied as well.

As a side note: when encountering compatibility issues like this one, we should be skeptical about modifying filter order as a way of fixing the problem, since we can't control the filter order of all third party code that might modify the HTML markup that our plugins affect.

mukeshpanchal27:
Thanks @joemcgill for the feedback.

I have updated PR that use `sizes` value from the IMG tag and reverted the filter priority changes. I will open issue for `wp_calculate_image_sizes()`

plugins/webp-uploads/picture-element.php
```diff
@@ -95,12 +95,10 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	list( $src, $width, $height ) = $image_src;
 	$size_array                   = array( absint( $width ), absint( $height ) );
 
-	foreach ( $mime_types as $image_mime_type ) {
-		$sizes = wp_calculate_image_sizes( $size_array, $src, $image_meta, $attachment_id );
-		if ( false === $sizes ) {
-			continue;
-		}
+	// Get the sizes from the IMG tag.
+	$sizes = (string) $processor->get_attribute( 'sizes' );
```
westonruter: Note that `$processor->get_attribute( 'sizes' )` here could return `null` if no attribute is present (or `true` in the case of ``, which is unlikely). Therefore, I recommend removing the `(string)` cast and instead when using the `$sizes` attribute to only add it to the `$picture_sources` markup if `is_string( $sizes )`.

```suggestion
	$sizes = $processor->get_attribute( 'sizes' );
```

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +125,75 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
+			}
+		);
+
+		// Picture filter "webp_uploads_wrap_image_in_picture" is run prior to "wp_content_img_tag".
+		$the_image = apply_filters( 'wp_content_img_tag', $the_image, 'the_content', $attachment_id );
+
+		// 1024px sizes comes from "source" tags.
+		$this->assertStringContainsString( 'sizes="(max-width: 1024px) 100vw, 1024px"', $the_image );
+		// 333px sizes comes from "IMG" tag.
+		$this->assertStringContainsString( 'sizes="(max-width: 333px) 100vw, 333px"', $the_image );
```
westonruter: Let's have this do an `assertSame()` assertion to test the entire string. See https://github.com/WordPress/performance/pull/1328#discussion_r1667070230

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +125,75 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
+			},
+			9
+		);
+
+		$the_image = apply_filters( 'wp_content_img_tag', $the_image, 'the_content', $attachment_id );
+
+		// 1024px sizes not set in "source" tags.
+		$this->assertStringContainsString( 'sizes="(max-width: 1024px) 100vw, 1024px"', $the_image );
+		// 333px sizes set in "IMG" and "source" tag.
+		$this->assertStringContainsString( 'sizes="(max-width: 333px) 100vw, 333px"', $the_image );
```
westonruter: I don't understand. These assertions are the same as the two assertions above before adding the additional `add_filter()` call.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +125,75 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
+		add_filter(
+			'wp_content_img_tag',
+			static function (
+				string $content
+			): string {
+				$processor = new WP_HTML_Tag_Processor( $content );
+				if ( ! $processor->next_tag( array( 'tag_name' => 'img' ) ) ) {
+					return $content;
+				}
+				$processor->set_attribute( 'sizes', '(max-width: 333px) 100vw, 333px' );
```
westonruter: This closure is identical to the above one. It should be saved in a variable and then re-used in both `add_filter()` calls then.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +125,75 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
 			),
 		);
 	}
+
+	public function test_sizes_value_for_picture_element(): void {
```
westonruter: Please also add a test for when an `img` doesn't have `sizes` at all.

westonruter (changes requested):

plugins/webp-uploads/picture-element.php
westonruter:
```suggestion
	$sizes = $processor->get_attribute( 'sizes' );
```

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +125,75 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
+
+		// Picture filter "webp_uploads_wrap_image_in_picture" is run prior to "wp_content_img_tag".
+		$the_image = apply_filters( 'wp_content_img_tag', $the_image, 'the_content', $attachment_id );
+
+		// 1024px sizes comes from "source" tags.
+		$this->assertStringContainsString( 'sizes="(max-width: 1024px) 100vw, 1024px"', $the_image );
+		// 333px sizes comes from "IMG" tag.
+		$this->assertStringContainsString( 'sizes="(max-width: 333px) 100vw, 333px"', $the_image );
+
+		// Run filter prior to Picture filter "webp_uploads_wrap_image_in_picture".
```
westonruter: Should this perhaps be moved to another test? It seems like this test could be divided in two.

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: Updated.

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: Moved to second test

mukeshpanchal27:

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +125,229 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
+		if ( ! wp_image_editor_supports( array( 'mime_type' => $mime_type ) ) ) {
+			$this->markTestSkipped( "Mime type $mime_type is not supported." );
+		}
+
+		$this->opt_in_to_jpeg_and_webp();
+
+		$this->opt_in_to_picture_element();
+
+		// Create an image.
+		$attachment_id = self::factory()->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/data/images/leaves.jpg' );
```
joemcgill: We're duplicating this in several tests, which will be a bit slow. Could we move this process to a shared fixture using the `wpSetUpBeforeClass()` method and store the attachment ID as a static property? See an example [here](https://github.com/WordPress/performance/blob/fix/1349-sizes-picture-element/plugins/auto-sizes/tests/test-auto-sizes.php#L11-L23).

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +125,229 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
+			'{{img-alt}}'           => $alt,
+			'{{img-srcset}}'        => $image_srcset,
+			'{{img-sizes}}'         => '(max-width: 333px) 100vw, 333px', // For IMG tag sizes.
+			'{{source-sizes}}'      => $sizes, // For Picture sources sizes.
+			'{{jpeg-srcset}}'       => $jpeg_srcset,
+			'{{webp-srcset}}'       => $webp_srcset,
+		);
+
+		$expected_html = '<picture class="wp-picture-{{img-attachment-id}}" style="display: contents;"><source type="image/webp" srcset="{{webp-srcset}}" sizes="{{source-sizes}}"><source type="image/jpeg" srcset="{{jpeg-srcset}}" sizes="{{source-sizes}}"><img width="{{img-width}}" height="{{img-height}}" src="{{img-src}}" class="wp-image-{{img-attachment-id}}" alt="{{img-alt}}" decoding="async" loading="lazy" srcset="{{img-srcset}}" sizes="{{img-sizes}}" /></picture>';
+		$expected_html = str_replace( array_keys( $replacements ), array_values( $replacements ), $expected_html );
```
joemcgill: There are a lot of implementation details in these tests, which seems unnecessary, and makes it difficult to understand what is actually being tested.

I think all we're trying to test is whether the `source` elements generated by this plugin include the same `sizes` attribute that the original `img` element has after being filtered, it would be more straightforward to generate the image markup without picture support, then again with picture applied and assert that the `sizes` attributes of the `source` elements are the same as the `sizes` attribute of the `img` without picture support. Here's an untested example of what I would expect:

```php
$image = wp_get_attachment_image( $attachment_id );

$img_markup = apply_filters( 'the_content', $image );

// Apply picture element support.
$this->opt_in_to_picture_element();

$picture_markup = apply_filters( 'the_content', $image );

$img_processor->next_tag( array( 'tag_name' => 'IMG' );
$img_sizes = $img_processor->get_attribute( 'sizes' );

$picture_processor->next_tag( array( 'tag_name' => 'source' ) ) );
$picture_source_sizes = $picture_processor->get_attribute( 'sizes' );

$this->assertSame( $img_sizes, $picture_source_sizes );
```
Then you would need another test that changes the sizes attribute of the original markup before applying `the_content` filters to make sure the `sizes` attributes are still the same.

The same is true for the other tests in this class.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +125,229 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
+
+		$expected_html = '<picture class="wp-picture-{{img-attachment-id}}" style="display: contents;"><source type="image/webp" srcset="{{webp-srcset}}" sizes="{{img-sizes}}"><source type="image/jpeg" srcset="{{jpeg-srcset}}" sizes="{{img-sizes}}"><img width="{{img-width}}" height="{{img-height}}" src="{{img-src}}" class="wp-image-{{img-attachment-id}}" alt="{{img-alt}}" decoding="async" loading="lazy" srcset="{{img-srcset}}" sizes="{{img-sizes}}" /></picture>';
+		$expected_html = str_replace( array_keys( $replacements ), array_values( $replacements ), $expected_html );
+
+		$the_image = apply_filters( 'wp_content_img_tag', $the_image, 'the_content', $attachment_id );
+
+		$this->assertSame( $expected_html, $the_image );
+	}
+
+	public function test_that_img_sizes_is_equal_to_source_for_picture_elementsss(): void {
```
joemcgill: Not sure what this is testing, but the name could be more descriptive.

joemcgill (dismissed):
This looks like it's working, but the tests could use some improvements. I've left some suggestions but let me know if they're not clear.

plugins/webp-uploads/picture-element.php
```diff
@@ -124,10 +122,10 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 			continue;
 		}
 		$picture_sources .= sprintf(
-			'<source type="%s" srcset="%s" sizes="%s">',
+			'<source type="%s"%s%s>',
 			esc_attr( $image_mime_type ),
-			esc_attr( $image_srcset ),
-			esc_attr( $sizes )
+			! empty( $image_srcset ) ? sprintf( ' srcset="%s"', esc_attr( $image_srcset ) ) : '',
```
adamsilverstein: Great! love this approach.

plugins/webp-uploads/picture-element.php
```diff
@@ -95,12 +95,10 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	list( $src, $width, $height ) = $image_src;
 	$size_array                   = array( absint( $width ), absint( $height ) );
 
-	foreach ( $mime_types as $image_mime_type ) {
-		$sizes = wp_calculate_image_sizes( $size_array, $src, $image_meta, $attachment_id );
-		if ( false === $sizes ) {
-			continue;
-		}
+	// Get the sizes from the IMG tag.
+	$sizes = $processor->get_attribute( 'sizes' );
```
adamsilverstein, line 99: Nice!

plugins/webp-uploads/picture-element.php
westonruter: For PHPStan strict rules compliance:

```suggestion
			is_string( $image_srcset ) ? sprintf( ' srcset="%s"', esc_attr( $image_srcset ) ) : '',
```

mukeshpanchal27: Updated in https://github.com/WordPress/performance/pull/1354/commits/6954f44727b86e1a25acee32432982b75c76469c

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: Updated in new test.

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: I have used the shared fixture for the new test. However, if I use it in the old test, I will have to refactor the entire test or update it. We can address this in a follow-up PR.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +140,124 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
+
+		$img_markup = apply_filters( 'the_content', $image );
+
+		// Apply picture element support.
+		$this->opt_in_to_picture_element();
+
+		$picture_markup = apply_filters( 'the_content', $image );
+
+		$img_processor = new WP_HTML_Tag_Processor( $img_markup );
+		$img_processor->next_tag( array( 'tag_name' => 'IMG' ) );
```
westonruter:
```suggestion
		$this->assertTrue( $img_processor->next_tag( array( 'tag_name' => 'IMG' ) ) );
```

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +140,124 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
+		$img_sizes = $img_processor->get_attribute( 'sizes' );
+
+		$picture_processor = new WP_HTML_Tag_Processor( $picture_markup );
+		$picture_processor->next_tag( array( 'tag_name' => 'source' ) );
+		$picture_source_sizes = $picture_processor->get_attribute( 'sizes' );
+
+		$this->assertSame( $img_sizes, $picture_source_sizes );
```
westonruter: There will be more than one `source` tag, right? As such perhaps better to use a `while()` loop, and finally to check that the `img` at the end of the `picture` also has the same expected `sizes`:

```suggestion
		while ( $picture_processor->next_tag( array( 'tag_name' => 'source' ) ) ) {
			$this->assertSame( $img_sizes, $picture_processor->get_attribute( 'sizes' ) );
		}
		$this->assertTrue( $img_processor->next_tag( array( 'tag_name' => 'IMG' ) ) );
		$this->assertSame( $img_sizes, $picture_processor->get_attribute( 'sizes' ) 
```

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +140,124 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
```
westonruter: See suggestions for the same code below.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +140,124 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
+				string $content
+			): string {
+				$processor = new WP_HTML_Tag_Processor( $content );
+				if ( ! $processor->next_tag( array( 'tag_name' => 'img' ) ) ) {
+					return $content;
+				}
+				$processor->set_attribute( 'sizes', '(max-width: 333px) 100vw, 333px' );
+				return $processor->get_updated_html();
+			}
+		);
```
westonruter: This is the only change to this test compared with the previous test, correct? As such, maybe better to use a data provider and have this `add_filter()` call be inside of a closure that is provided as an optional callback that gets executed at this point:

```php
function data_provider_test_image_sizes_equality(): array {
	return array(
		'no_filter' => array(
			'add_filter' => null,
		),
		'with_filter' => array(
			'add_filter' => static function () {
				add_filter(
					'wp_content_img_tag',
					static function (
						string $content
					): string {
						if ( ! $processor->next_tag( array( 'tag_name' => 'img' ) ) ) {
							return $content;
						}
						$processor->set_attribute( 'sizes', '(max-width: 333px) 100vw, 333px' );
					}
				);
			},
		)
	);
}
```

So in the test you can just do:

```php
if ( $add_filter instanceof Closure ) {
	$add_filter();
}
```

westonruter (commented):
Looking good! Almost there.

mukeshpanchal27:
Thanks @westonruter @adamsilverstein (Joe OoO) for the feedback. PR ready for review.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +140,121 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
+				'add_filter' => static function (): void {
+					add_filter(
+						'wp_content_img_tag',
+						static function (
+							string $content
+						): string {
+							$processor = new WP_HTML_Tag_Processor( $content );
+							if ( ! $processor->next_tag( array( 'tag_name' => 'img' ) ) ) {
+								return $content;
+							}
```
westonruter: If there is no `IMG` in the `$content`, this would be bad, right? I don't think this should return silently.

```suggestion
						'wp_content_img_tag',
						function ( string $content ): string {
							$processor = new WP_HTML_Tag_Processor( $content );
							$this->assertTrue( $processor->next_tag( array( 'tag_name' => 'img' ) ) );
```

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -125,4 +140,121 @@ public function data_provider_it_should_maybe_wrap_images_in_picture_element():
...
+
+		while ( $picture_processor->next_tag( array( 'tag_name' => 'source' ) ) ) {
+			$this->assertSame( $img_sizes, $picture_processor->get_attribute( 'sizes' ), 'The IMG and Picture source have same sizes attributes.' );
+		}
+	}
+
+	/**
+	 * Data provider for it_should_maybe_wrap_images_in_picture_element.
+	 *
+	 * @return array<string, array<string, (Closure)|null>>
```
westonruter:
```suggestion
	 * @return array<string, array{ add_filter: Closure|null }>
```

westonruter (approved):
Just a couple nits, but pre-approving as LGTM.

adamsilverstein (approved):
Excellent!

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: It generate PHPStan error: https://github.com/WordPress/performance/pull/1354/commits/61a488f10c0b9408bdc25820a2abf2f01218ebfd

PR #1373, opened 2024-07-17
Leverage URL metrics to reserve space for embeds to reduce CLS

### LINKED ISSUE #1310: Embeds cause layout shift which can be reduced with Optimization Detective
## Feature Description

When Optimization Detective gathers the dimensions of embeds, this information can be used to reserve space for these embeds so that when they load their layout shifts can be minimized. This will improve CLS.

westonruter:
Fixes #1310 

The changes in this pull request grew larger than the strict scope of reducing CLS for embeds since the work to implement this uncovered shortcomings with the current Optimization Detective API. This PR includes the following key pieces:

1. Introduce client-side extensions in the form of script modules which are loaded when the detection logic runs. The client-side script modules are able to register functionality to run as soon as the module is loaded by implementing an exported `initialize` function. They also are able to register functionality to run just before the URL metric data is being submitted by implementing an exported `finalize` function which amends the URL metric data as needed. This builds on #1492 which enabled the URL metric schema to be extended. This PR includes Embed Optimizer extending the URL metric element schema to include a `resizedBoundingClientRect` key which is populated with the script module via a `ResizeObserver` which starts observing in the module's `initialize` function. New script modules are added simply by appending to the array passed to the `od_extension_module_urls` filter.
2. The gathered URL metric data is now sent when the page is hidden as opposed to when the page is "done" loading. This is essential in order to capture the resized heights of embeds once the user scrolls them into view and they are loaded. _This will also open up new opportunities to track INP over the life of the page to include in the submitted URL metric data._
3. New APIs are added to Optimization Detective which Embed Optimizer depends on (e.g. `OD_URL_Metric_Group_Collection::get_all_denormalized_elements()`). In order to avoid the need to do `method_exists()` checks and the like, a new `od_init` action is introduced which passes the Optimization Detective version as its only argument. Embed Optimizer and Image Prioritizer then check this passed version to see if it meets the requirements. If not, then an admin notice is displayed. If yes, then it goes ahead and loads the Optimization Detective extension code.
4. Related to the above, I found that certain plugins initialize their module code _after_ the `after_setup_theme` action. In proposing embedding Optimization Detective in Elementor (https://github.com/WordPress/performance/pull/1373), I found that [modules load](https://github.com/elementor/elementor/blob/a627ce730837b8b8d4381d8fdfab106dda35f96b/includes/plugin.php#L877) at `init` with priority 0. This makes the current bootstrap code unusable when it is loaded in such a module, necessitating an [ugly workaround](https://github.com/elementor/elementor/blob/287ddef5e78605558375d77014f38721f84671f9/modules/optimization-detective/module.php#L30-L53). Since Optimization Detective doesn't need to bootstrap before `init` anyway, the solution seems to simply be to move the bootstrapping logic to `init` as well. This is implemented in this PR by moving to `init` at priority 9 since the plugin has code that runs at `init` with the default priority of 10.

# Accounting for responsive embeds

Initially I was thinking the optimization logic should wait to set the `min-height` until there are URL metrics gathered for the smallest breakpoint group. For example, let's say the page is responsive and on desktop an embed is much taller than it is on mobile. If the desktop metrics are collected before mobile metrics are collected, then the `min-height` could be being set for desktop could be too high for mobile, resulting in a gap below the embed on mobile. For example, consider this [tweet](https://twitter.com/WordPress/status/1795196552048173268?ref_src=twsrc%5Etfw%7Ctwcamp%5Etweetembed%7Ctwterm%5E1795196552048173268%7Ctwgr%5Ebb48b5eedf52835ae714ba6980a39f012a7cbee4%7Ctwcon%5Es1_&ref_url=http%3A%2F%2Flocalhost%3A8888%2F2024%2F08%2F18%2Fmultiple-layout-shifting-embeds%2F) on mobile vs desktop:

Mobile (320w) | Desktop
--|--
Screenshot 2024-10-04 11 12 32 | Screenshot 2024-10-04 11 12 45

The tweet on mobile here is 574px high whereas on desktop it is 825px, so if desktop metrics were only collected then the result would be a worst case gap of 251px at the bottom before the next content:

> Screenshot 2024-10-04 11 18 31

A simple way to solve this would not to set the `min-height` on the `figure.wp-block-embed` element at all, but rather to create a new `style` rule with responsive styles that sets multiple `min-height` styles for the element according to each breakpoint. This is implemented in this PR. For example, the following CSS is added to the `head` for a Twitter embed:

```html
<style>
@media (max-width: 480px) { #embed-optimizer-a7659db28ecaa36ddee6ae66857dabd8 { min-height: 500px; } }
@media (min-width: 481px) and (max-width: 600px) { #embed-optimizer-a7659db28ecaa36ddee6ae66857dabd8 { min-height: 500px; } }
@media (min-width: 601px) and (max-width: 782px) { #embed-optimizer-a7659db28ecaa36ddee6ae66857dabd8 { min-height: 500px; } }
@media (min-width: 783px) { #embed-optimizer-a7659db28ecaa36ddee6ae66857dabd8 { min-height: 500px; } }
</style>
```

# For follow-up

- [ ] Move the `init` to an even later action. https://github.com/WordPress/performance/pull/1373#discussion_r1796079846
- [ ] Consider renaming `nonce` to `hmac` to disambiguate from `_wpnonce`. https://github.com/WordPress/performance/pull/1373#discussion_r1799871375

# Demo Videos

## Before

[bad-cls-without-embed-optimizer.webm](https://github.com/user-attachments/assets/df82b628-45ac-4f7d-a8ef-01731b3fc8d3)

## After

Once URL metrics have been collected from visitors by Optimization Detective:

[good-cls-with-embed-optimizer.webm](https://github.com/user-attachments/assets/924f1d5a-9459-414c-9091-20491c190a92)


plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -26,26 +26,72 @@ final class Embed_Optimizer_Tag_Visitor {
...
+			return true;
+		}
+
+		// Short-circuit if not a figure.wp-block-embed tag.
+		if ( ! $this->is_embed_figure( $processor ) ) {
 			return false;
 		}
 
-		$max_intersection_ratio = $context->url_metrics_group_collection->get_element_max_intersection_ratio( $processor->get_xpath() );
+		$embed_wrapper_xpath = $processor->get_xpath() . '/*[1][self::DIV]';
```
westonruter: It's not ideal to be constructing this XPath manually.

westonruter: I've opened https://github.com/WordPress/performance/pull/1407 to explore this, but it's not necessary for this PR to move forward.

westonruter:
I noticed Query Monitor flagging a PHP deprecation error when loading a page with Optimization Detective and Embed Optimizer active. A TikTok embed is present on the page. I'm getting:

> PHP Deprecated:  Implicit conversion from float 739.9976196289062 to int loses precision in /var/www/html/wp-content/plugins/optimization-detective/class-od-url-metrics-group-collection.php on line 486

Call stack:

image

**Update:** Addressed in https://github.com/WordPress/performance/pull/1411.

westonruter:
I also saw this for some reason:

> PHP Notice:  OD_HTML_Tag_Processor::get_updated_html(): Unable to append markup to optimization_detective_end_of_body since the bookmark no longer exists. in /var/www/html/wp-includes/functions.php on line 6085

**Update:** See fix below in https://github.com/WordPress/performance/pull/1373#issuecomment-2295405151

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -26,7 +26,7 @@
  * @phpstan-type Data        array{
  *                               url: string,
  *                               timestamp: float,
- *                               viewport: RectData,
+ *                               viewport: array{ width: int, height: int },
```
westonruter: This should be put into a reusable type.

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -26,26 +26,76 @@ final class Embed_Optimizer_Tag_Visitor {
...
+		$minimum_height      = $context->url_metrics_group_collection->get_element_minimum_height( $embed_wrapper_xpath );
+		if ( is_float( $minimum_height ) ) {
+			$style = $processor->get_attribute( 'style' );
+			if ( is_string( $style ) ) {
+				$style = rtrim( trim( $style ), ';' ) . '; ';
+			} else {
+				$style = '';
+			}
+			$style .= sprintf( 'min-height: %dpx;', $minimum_height );
```
westonruter: Given that heights are floats, should this rather be:

```suggestion
			$style .= sprintf( 'min-height: %fpx;', $minimum_height );
```

But then it could result in some very long decimals, like `min-height: 500.200000px;`

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -26,26 +26,76 @@ final class Embed_Optimizer_Tag_Visitor {
...
+			$processor->set_attribute( 'style', $style );
```
westonruter: This should be a helper method on `OD_HTML_Tag_Processor`.

westonruter: See prior art in the AMP plugin: https://github.com/ampproject/amp-toolbox-php/blob/78531851c59fa5f306315372719f0064b5542cf4/src/Dom/Element.php#L49-L69

westonruter:
Scratched code to explore using <code>onCLS</code> to detect when an embed has loaded (and causes a layout shift. Problem: In the case of the Twitter embed, the source node being reported is _REMOVED_ from the DOM, so we cannot see the ancestor which has a data-od-xpath attribute.

<details><summary>Code scratch</summary>

```js
	const { onLCP, onCLS } = await import( webVitalsLibrarySrc );

	onCLS(
		( metric ) => {
			for ( const entry of metric.entries ) {
				if (
					entry.entryType === 'layout-shift' &&
					! entry.hadRecentInput
				) {
					console.info( entry );
					for ( const source of entry.sources ) {
						console.info( source.node );
					}
				}
			}
		},
		{
			// This is necessary in order to collect all layout shifts, even those which don't cause a bad CLS score.
			reportAllChanges: true,
		}
	);
```

</details>

westonruter:
I'm seeing this again:

> OD_HTML_Tag_Processor::get_updated_html(): Unable to append markup to optimization_detective_end_of_body since the bookmark no longer exists.

```
wp-includes/functions.php:6095
wp_trigger_error()
wp-content/plugins/optimization-detective/class-od-html-tag-processor.php:612
OD_HTML_Tag_Processor->warn()
wp-content/plugins/optimization-detective/class-od-html-tag-processor.php:579
OD_HTML_Tag_Processor->get_updated_html()
wp-includes/html-api/class-wp-html-tag-processor.php:2500
WP_HTML_Tag_Processor->seek()
wp-content/plugins/optimization-detective/class-od-html-tag-processor.php:428
OD_HTML_Tag_Processor->seek()
wp-content/plugins/optimization-detective/optimization.php:229
od_optimize_template_output_buffer()
wp-content/plugins/performance-lab/includes/server-timing/load.php:138
{closure}()
wp-includes/class-wp-hook.php:324
apply_filters('od_template_output_buffer')
wp-content/plugins/optimization-detective/optimization.php:66
{closure}()
wp-content/plugins/optimization-detective/optimization.php:66
ob_end_flush()
wp-includes/functions.php:5437
wp_ob_end_flush_all()
wp-includes/class-wp-hook.php:324
do_action('shutdown')
wp-includes/load.php:1278
shutdown_action_hook()
wp-includes/load.php:1278
```

**Update:** Fixed in edc52fad0c127db5ea92e4a94dc2fdf9eaf2c062. The issue is that every time `seek()` is called, it will also call `get_updated_html()`. So I introduced a `get_final_updated_html()` which is intended to be called at the very end of iterating over the document.

plugins/optimization-detective/detect.js
```diff
@@ -239,6 +224,18 @@ export default async function detect( {
 		log( 'Proceeding with detection' );
 	}
 
+	/** @type {Extension[]} */
+	const extensions = [];
+	for ( const extensionModuleUrl of extensionModuleUrls ) {
+		const extension = await import( extensionModuleUrl );
+		extensions.push( extension );
+		// TODO: There should to be a way to pass additional args into the module. Perhaps extensionModuleUrls should be a mapping of URLs to args. It's important to pass webVitalsLibrarySrc to the extension so that onLCP, onCLS, or onINP can be obtained.
```
adamsilverstein: we also have JavaScript hooks you could leverage if the idea is to make this extensible.

swissspidy: These aren't available as script modules though, right?

plugins/optimization-detective/detect.js
```diff
@@ -361,46 +356,59 @@ export default async function detect( {
...
-		} );
+			{ once: true }
+		);
+	} );
 
-		if ( response.status === 200 ) {
-			setStorageLock( getCurrentTime() );
+	for ( const extension of extensions ) {
+		if ( extension.finalize instanceof Function ) {
+			extension.finalize( {
```
swissspidy: So extensions add their own data by directly modifying the passed `urlMetric`? Why not have them return the new data instead?

plugins/optimization-detective/detect.js
```diff
@@ -361,46 +356,59 @@ export default async function detect( {
...
-		}
+	// Even though the server may reject the REST API request, we still have to set the storage lock
+	// because we can't look at the response when sending a beacon.
+	setStorageLock( getCurrentTime() );
+
+	if ( isDebug ) {
+		log( 'Sending URL metric:', urlMetric );
 	}
 
+	const body = Object.assign( {}, urlMetric, {
```
swissspidy: Why no spread?

plugins/optimization-detective/detect.js
adamsilverstein: true, but they could be added separately to filter the an args array here for example

plugins/optimization-detective/detect.js
westonruter: As this PR is currently implemented, Embed Optimizer overrides the `boundingClientRect` with the final rect reported the ResizeObserver for embeds. But since then I've realized this isn't ideal because we can't know if the `boundingClientRect` was the initial size or the final size. So in that case I think it would be better if the schema is made extensible (cf. https://github.com/WordPress/performance/pull/1490#issuecomment-2300905345) so that Embed Optimizer can add a `resizedBoundingClientRect` which we then exclusively look for when setting the `min-height`. So in that case, the extension's `finalize` function could just return an object that gets merged on top of the `urlMetric`. But it seems there could be use cases for extensions to still be able modify the entire object?

plugins/optimization-detective/detect.js
westonruter: Good point. I forgot.

plugins/optimization-detective/detect.js
swissspidy: I think returning an object and using `structuredClone` here to avoid unintended modifications is cleaner.

westonruter: OK, sounds good. But using `structuredClone()` where? We wouldn't need to pass in the `urlMetric` anymore to the `finalize` function, so would a clone be needed?

plugins/optimization-detective/detect.js
westonruter: Fixed in def2aab

plugins/optimization-detective/detect.js
westonruter: Er wait, we do need to pass the `urlMetric` because Embed Optimizer needs to amend `Element` data in the `elements` key.

plugins/embed-optimizer/detect.js
```diff
@@ -0,0 +1,90 @@
...
+		monitorEmbedWrapperForResizes( embedWrapper );
+	}
+
+	if ( isDebug ) {
+		log( 'Loaded embed content rects:', loadedElementContentRects );
+	}
+}
+
+/**
+ * Initialize.
```
westonruter:
```suggestion
 * Finalize.
```

plugins/optimization-detective/storage/rest-api.php
```diff
@@ -131,6 +131,7 @@ function od_handle_rest_request( WP_REST_Request $request ) {
 		$data = $request->get_params();
 		// Remove params which are only used for the REST API request and which are not part of a URL Metric.
 		unset(
+			$data['_wpnonce'],
 			$data['slug'],
 			$data['nonce']
 		);
```
westonruter: This may be short-sighted. See https://github.com/WordPress/performance/issues/1567

westonruter: I originally did this in 1da219f394d546802d8b7b2c43f936152611cd59 (and 5db6f54d9d1196501008fdbcfc7206b99b75e7b0) but then reverted it in 0fa263a75ef90c0f404c71621b1d7cb9dd5482a1 should be reverted. I need to remember why I undid that.

westonruter: See https://github.com/WordPress/performance/pull/1574

plugins/optimization-detective/class-od-url-metric-group-collection.php
```diff
@@ -396,34 +397,61 @@ public function get_common_lcp_element(): ?array {
...
-	 * Gets the max intersection ratios of all elements across all groups and their captured URL metrics.
+	 * Gets all elements from all URL metrics from all groups keyed by the elements' XPaths.
 	 *
-	 * @return array<string, float> Keys are XPaths and values are the intersection ratios.
+	 * This is an O(n^3) function so its results must be cached. This being said, the number of groups should be 4 (one
+	 * more than the default number of breakpoints) and the number of URL metrics for each group should be 3
+	 * (the default sample size). Therefore, given the number (n) of visited elements on the page this will only
+	 * end up running n*4*3 times.
+	 *
+	 * @todo Should there be an OD_Element class which has a $url_metric property which then in turn has a $group property. Then this would only need to return array<string, OD_Element[]>.
```
westonruter, line 420: I want to address this in a follow-up PR since it will require more refactoring and I don't want to make this PR even larger.

plugins/optimization-detective/class-od-url-metric-group-collection.php
```diff
@@ -396,34 +397,61 @@ public function get_common_lcp_element(): ?array {
...
+
+		$result = ( function () {
+			$elements_max_intersection_ratios = array();
+			foreach ( $this->get_all_denormalized_elements() as $xpath => $denormalized_elements ) {
+				$element_intersection_ratios = array();
+				foreach ( $denormalized_elements as list( $group, $url_metric, $element ) ) {
+					$element_intersection_ratios[] = $element['intersectionRatio'];
+				}
+				$elements_max_intersection_ratios[ $xpath ] = count( $element_intersection_ratios ) > 1
+					? (float) max( ...$element_intersection_ratios )
```
westonruter: Interestingly, if there is only one element in the array than this is a fatal error: https://3v4l.org/nQNZh

Too bad it doesn't just return the initial item.

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -119,6 +158,80 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+
+		/**
+		 * Array of tuples of groups and their minimum heights keyed by the minimum viewport width.
+		 *
+		 * @var array<int, array{OD_URL_Metric_Group, int}> $group_minimum_heights
+		 */
+		$group_minimum_heights = array();
+
+		$denormalized_elements = $context->url_metric_group_collection->get_all_denormalized_elements()[ $embed_wrapper_xpath ] ?? array();
+		foreach ( $denormalized_elements as list( $group, $url_metric, $element ) ) {
```
westonruter, line 218: See note below on the `get_all_denormalized_elements()` method that I'd like to instead have `$element` be an `OD_Element` class instance which has a `$url_metric` property to access the `OD_URL_Metric` instance if is a part of. In the same way, the `OD_URL_Metric` class can have a `$group` property to indicate the `OD_URL_Metric_Group` it is a part of. This would eliminate the need to pass back an array of tuples and instead. So instead, to get the `$group` this code could do `$element->url_metric->group`.

This I want to do in a follow-up PR.

westonruter:
@felixarntz _Finally!_ Ready for review.

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -119,6 +158,80 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+	/**
+	 * Gets the XPath for the embed wrapper.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param OD_Tag_Visitor_Context $context Tag visitor context.
+	 * @return string XPath.
+	 */
+	private function get_embed_wrapper_xpath( OD_Tag_Visitor_Context $context ): string {
+		return $context->processor->get_xpath() . '/*[1][self::DIV]';
```
felixarntz: What does this additional string at the end here do?

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -119,6 +158,80 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+			}
+			$group_min_width = $group->get_minimum_viewport_width();
+			if ( ! isset( $group_minimum_heights[ $group_min_width ] ) ) {
+				$group_minimum_heights[ $group_min_width ] = array( $group, $element['resizedBoundingClientRect']['height'] );
+			} else {
+				$group_minimum_heights[ $group_min_width ][1] = min(
+					$group_minimum_heights[ $group_min_width ][1],
+					$element['resizedBoundingClientRect']['height']
+				);
+			}
```
felixarntz: Can you change this so that each entry is an associative array with e.g. `group` and `minimum_height` keys? This makes the code more intuitive to follow.

plugins/embed-optimizer/detect.js
```diff
@@ -0,0 +1,90 @@
+const consoleLogPrefix = '[Embed Optimizer]';
```
felixarntz, line 9: Could you add a doc block on top of this file to explain what its purpose is? I have a hard time deciphering how it fits into the functionality as there is only very minimal documentation.

I was expecting for JS related functionality to be already covered by Optimization Detective's own detection script, which is why I'm wondering about this.

felixarntz (commented):
@westonruter At first glance this looks good, for now mostly a few questions for clarification.

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
westonruter: Given an embed block such as the following:

```html
<figure class="wp-block-embed is-type-video is-provider-wordpress-tv wp-block-embed-wordpress-tv wp-embed-aspect-16-9 wp-has-aspect-ratio">
	<div class="wp-block-embed__wrapper">
		<iframe title="VideoPress Video Player" aria-label='VideoPress Video Player' width='750' height='422' src='https://video.wordpress.com/embed/vaWm9zO6?hd=1&amp;cover=1' frameborder='0' allowfullscreen allow='clipboard-write'></iframe>
		<script src='https://v0.wordpress.com/js/next/videopress-iframe.js?m=1674852142'></script>
	</div>
</figure>
```

If this were a direct descendant of the `BODY`, then the XPath for the `FIGURE` tag here would be:

```
/*[1][self::HTML]/*[2][self::BODY]/*[1][self::FIGURE]
```

What this method is doing is appending to this XPath the string to select the `div.wp-block-embed__wrapper` child element, which has the XPath of:

```
/*[1][self::HTML]/*[2][self::BODY]/*[1][self::FIGURE]/*[1][self::DIV]
```

I've clarified this in 455ef4fe.

I played around with drafting https://github.com/WordPress/performance/pull/1407 to add some helpers to construct XPaths which we can consider later instead of manually constructing XPath expressions.

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
westonruter: How about a390e15e919b267ee8fcc963351a2ac62dabd2b3?

plugins/embed-optimizer/detect.js
westonruter: Sure. Added in a760705fc1b908a7b5db95350ddbcd6a999f895b. 

The logic in Embed Optimizer's `detect.js` module script augments the logic in Optimization Detective's `detect.js` module script. Namely, Optimization Detective is not doing anything with `ResizeObserver` as this is exclusively an Embed Optimizer concern (as the `resizedBoundingClientRect` data is only added to the JSON schema by Embed Optimizer).

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
felixarntz: Makes sense, thanks for clarifying!

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
felixarntz: LGTM, thanks!

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
westonruter: > See note below on the `get_all_denormalized_elements()` method that I'd like to instead have `$element` be an `OD_Element` class instance which has a `$url_metric` property to access the `OD_URL_Metric` instance if is a part of. In the same way, the `OD_URL_Metric` class can have a `$group` property to indicate the `OD_URL_Metric_Group` it is a part of. This would eliminate the need to pass back an array of tuples and instead. So instead, to get the `$group` this code could do `$element->url_metric->group`.
> 
> This I want to do in a follow-up PR.

westonruter: See https://github.com/WordPress/performance/pull/1585

plugins/optimization-detective/class-od-url-metric-group-collection.php
westonruter: See https://github.com/WordPress/performance/pull/1585

plugins/image-prioritizer/load.php
```diff
@@ -44,9 +44,14 @@ static function ( string $global_var_name, string $version, Closure $load ): voi
...
-			// because it is the first action that fires once the theme is loaded.
-			add_action( 'after_setup_theme', $bootstrap, PHP_INT_MIN );
+			/*
+			 * Wait until after the plugins have loaded and the theme has loaded. The after_setup_theme action could be
+			 * used since it is the first action that fires once the theme is loaded. However, plugins may embed this
+			 * logic inside a module which initializes even later at the init action. The earliest action that this
+			 * plugin has hooks for is the init action at the default priority of 10 (which includes the rest_api_init
+			 * action), so this is why it gets initialized at priority 9.
+			 */
+			add_action( 'init', $bootstrap, 9 );
```
felixarntz, line 54: I'm not sure I understand this. Why is the timing of this (in Image Prioritizer and Embed Optimizer) relevant for other plugins? Are you thinking of extensions to _those_ plugins (rather than extensions to Optimization Detective)?

Another thought: Why can't this use the `od_init` action?

plugins/embed-optimizer/load.php
```diff
@@ -43,9 +43,14 @@ static function ( string $global_var_name, string $version, Closure $load ): voi
...
-			// because it is the first action that fires once the theme is loaded.
-			add_action( 'after_setup_theme', $bootstrap, PHP_INT_MIN );
+			/*
+			 * Wait until after the plugins have loaded and the theme has loaded. The after_setup_theme action could be
+			 * used since it is the first action that fires once the theme is loaded. However, plugins may embed this
+			 * logic inside a module which initializes even later at the init action. The earliest action that this
+			 * plugin has hooks for is the init action at the default priority of 10 (which includes the rest_api_init
+			 * action), so this is why it gets initialized at priority 9.
+			 */
+			add_action( 'init', $bootstrap, 9 );
```
felixarntz, line 53: See other comment.

plugins/optimization-detective/class-od-url-metric-group-collection.php
felixarntz: Sounds good. For the future, ideally the different plugin modifications would happen in separate PRs too. It's more convenient all in one PR, but would be better for review, plus you could still use sub-PRs like you've done before, to e.g. build something in Embed Optimizer on top of another PR change to the Optimization Detective infrastructure.

plugins/optimization-detective/detect.js
felixarntz: Looking at this old comment thread, does it still require an update? Asking since `urlMetric` is still being passed below.

plugins/image-prioritizer/load.php
westonruter: Bootstrapping at `init` is all about the embeddability of these plugins inside of other plugins, which is why this whole bootstrap logic was first devised (https://github.com/WordPress/performance/issues/1081, https://github.com/WordPress/performance/pull/1159, https://github.com/WordPress/performance/pull/1185). 

For example, in https://github.com/elementor/elementor/pull/28844 I've created a new module for Elementor that embeds the Optimization Detective plugin and its extensions. Elementor initializes its modules at `init` priority 0. Since none of the functionality in Optimization Detective needs to be initialized before `init`, there's no need for the current use of `after_setup_theme` to initialize. So this postpones initialization to latest possible point which opens the door to be embedded in more places.

plugins/embed-optimizer/load.php
westonruter: See https://github.com/WordPress/performance/pull/1373#discussion_r1794243822

plugins/optimization-detective/detect.js
westonruter: I don't think it requires an update. The `urlMetric` object needs to be passed so that the extension can amend it as needed. Sure this would mean the extension could modify "core" properties, but I don't see this as a problem (and in fact it may be desirable).

plugins/optimization-detective/class-od-url-metric-group-collection.php
westonruter: Yes, I agree. Sorry for the large PR. It was a bit of a journey.

plugins/optimization-detective/detect.js
felixarntz: I think it's a fair concern that extensions should only be able to modify what we want them to modify. I don't think we have a good use-case in mind why someone _should_ modify "core" properties, so then we shouldn't allow it, until a real reasonable need arises.

Some alternatives for how to resolve this:
* Either don't give access to `urlMetric` at all and have the functions only return data that is centrally added to `urlMetric`. Downside: May make for a worse API, since returning e.g. just an object is not exactly safe (you would need to remember property names, there's no auto-completion etc.).
* We somehow "lock" things in the passed `urlMetric` so that they can't be written by the function. Not quite sure on how to reliably do that, but that could work.
* We pass another object to the function that only contains the bits of `urlMetric` that _should_ be amendable.

I think the last point would be a cleaner solution.

plugins/image-prioritizer/load.php
felixarntz: This still feels like a rather shaky foundation to me. For example, why does it use priority `9`?

I think I'm still not following what this is meant to do: I think I understand now why certain code needs to run before this callback, but why does certain other code need to run after this callback? For instance, why couldn't we use `init` on `999999`, or even `wp_loaded` or a later hook?

westonruter: Priority 9 is one less than the default priority of 10, as noted in the comment. There are callbacks added at this default priority in the plugin. So priority 9 is the last.

If we used `init` with an even later priority, then we'd need to adjust the other `init` actions' priorities as well in this plugin, for example:

https://github.com/WordPress/performance/blob/94fc5e7215516d58dc307f973cb75ae1432ad042/plugins/optimization-detective/storage/class-od-url-metrics-post-type.php#L48

https://github.com/WordPress/performance/blob/94fc5e7215516d58dc307f973cb75ae1432ad042/plugins/embed-optimizer/hooks.php#L27

If we went with `wp_loaded`, then we couldn't use the `init` action in the plugins at all, which as I understand is too late for when post types should be registered.

plugins/optimization-detective/detect.js
westonruter: The problem is that it's not just the `urlMetric` that needs to be modified. It's also the elements located in `urlMetric.elements`. The extensions need to be able to extend the properties of those elements. I'll think about an alternative.

plugins/image-prioritizer/load.php
felixarntz: So basically this needs to be called as late as possible, but before any of Optimization Detective's own hooks would fire?

westonruter: Yes, exactly. As late as possible, and hopefully after any other possible embedding plugins' modules initialize (e.g. Elementor initializes at `init` priority `0`).

felixarntz: Makes sense. Since we have full control over the code in Optimization Detective, we may want to consider moving the plugin's `init` callbacks to a very late priority, so that we can also push this back even further, while still being on `init`. That way we can cater for other plugins even better, since I wouldn't find it unreasonable for other plugins to load modules on `init` with the default priority either.

Given the callbacks from Optimization Detective do not _have_ to run after everything else, maybe we just use a very high number (rather than `PHP_INT_MAX` which I would only use if it technically has to be the last thing).

We can open a follow up PR for this though, if you agree?

westonruter: Sounds good!

plugins/optimization-detective/detect.js
westonruter: Work in progress: 7ca1fbc

westonruter: OK, I've created a new API whereby the `finalize` function is passed an object which is defined as follows:

```ts
export type FinalizeArgs = {
	readonly getRootData: () => URLMetric,
	readonly amendRootData: ( properties: AmendedRootData ) => void,
	readonly getElementData: ( xpath: string ) => ElementData|null,
	readonly amendElementData: ( xpath: string, properties: AmendedElementData ) => void,
	readonly isDebug: boolean,
};
```

If someone tries to set a core property of the `URLMetric` (via `amendRootData()`) or on an `element` (via `amendRootData()`), then this triggers an error.

What do you think? Not entirely sold on the language of "amend". Also considered "extend".

westonruter: @swissspidy Do you like?

plugins/optimization-detective/detect.js
```diff
@@ -129,12 +112,131 @@ function getCurrentTime() {
...
+function getRootData() {
+	const immutableUrlMetric = structuredClone( urlMetric );
+	recursiveFreeze( immutableUrlMetric );
+	return immutableUrlMetric;
+}
+
+/**
+ * Amends root URL metric data.
+ *
+ * @todo Would "extend" be better than "amend"? Or something else?
```
westonruter: Or `setRootData`?

westonruter: Or not, as that implies it is overriding all of the root data which it doesn't allow.

plugins/optimization-detective/class-od-url-metric-group-collection.php
adamsilverstein: why is the spread needed here, can this just be `(float) max( $element_intersection_ratios )`?

adamsilverstein:
```suggestion
					? (float) max( $element_intersection_ratios )
```

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -26,26 +26,82 @@ final class Embed_Optimizer_Tag_Visitor {
...
+		);
+	}
+
+	/**
+	 * Determines whether the processor is currently at a div.wp-block-embed__wrapper tag (which is a child of figure.wp-block-embed).
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param OD_HTML_Tag_Processor $processor Processor.
+	 * @return bool Whether the tag should be measured and stored in URL metrics
```
adamsilverstein:
```suggestion
	 * @return bool Whether the tag should be measured and stored in URL metrics.
```

plugins/image-prioritizer/helper.php
```diff
@@ -10,6 +10,41 @@
...
+	if ( ! version_compare( (string) strtok( $optimization_detective_version, '-' ), $required_od_version, '>=' ) ) {
+		add_action(
+			'admin_notices',
+			static function (): void {
+				global $pagenow;
+				if ( ! in_array( $pagenow, array( 'index.php', 'plugins.php' ), true ) ) {
+					return;
+				}
+				wp_admin_notice(
+					esc_html__( 'The Image Prioritizer plugin requires a newer version of the Optimization Detective plugin. Please update your plugins.', 'image-prioritizer' ),
```
adamsilverstein, line 31: can we add a link to the plugin update screen here?

plugins/image-prioritizer/load.php
```diff
@@ -6,7 +6,7 @@
  * Requires at least: 6.5
  * Requires PHP: 7.2
  * Requires Plugins: optimization-detective
- * Version: 0.1.4
+ * Version: 0.1.5-alpha
```
adamsilverstein, line 9: why `-alpha`?

plugins/optimization-detective/class-od-html-tag-processor.php
```diff
@@ -559,11 +571,25 @@ public function append_body_html( string $html ): void {
...
+	 * wasteful as they are only needed once finishing iterating over the document.
+	 *
 	 * @since 0.4.0
+	 * @see WP_HTML_Tag_Processor::get_updated_html()
+	 * @see WP_HTML_Tag_Processor::seek()
 	 *
 	 * @return string The processed HTML.
 	 */
 	public function get_updated_html(): string {
+		if ( ! $this->reached_end_of_document ) {
```
adamsilverstein, line 589: Nice!

plugins/optimization-detective/detect.js
adamsilverstein: amend seems fine, although extend is more familiar WordPress parlance

plugins/optimization-detective/class-od-url-metric-group-collection.php
westonruter: Oh, good point. It seems this is an alternate signature:

image

plugins/image-prioritizer/load.php
westonruter: Because this is what we do in core as well to indicate the current version in the codebase is not stable. I also needed to bump the versions because of the version checks that are new in this PR.

plugins/image-prioritizer/helper.php
westonruter: We could, but the user may not have the capability to access the plugin update screen. Plugins may not be updatable on the environment either due to it being a read only filesystem. So these complications would seem to make adding a link maybe more trouble than its worth.

plugins/optimization-detective/class-od-url-metric-group-collection.php
westonruter: Hah, and then there's no need for the ternary since a single element array does not cause an error: https://3v4l.org/fmIl9

westonruter: Addressed in dca43e9d8ff8cf121ee5c53e1d8e1903840bedc8

plugins/optimization-detective/detect.js
```diff
@@ -371,49 +494,73 @@ export default async function detect( {
...
+	setStorageLock( getCurrentTime() );
+
+	if ( isDebug ) {
+		log( 'Sending URL metric:', urlMetric );
+	}
+
+	const url = new URL( restApiEndpoint );
+	url.searchParams.set( '_wpnonce', restApiNonce );
+	url.searchParams.set( 'slug', urlMetricSlug );
+	url.searchParams.set( 'nonce', urlMetricNonce );
```
adamsilverstein, line 554: is having two nonces excessive? or helpful?

plugins/optimization-detective/detect.js
westonruter: Renamed in 73d2252961228c07f8ec1e2af82fd1cd588bb09d

plugins/optimization-detective/detect.js
```diff
@@ -371,49 +494,73 @@ export default async function detect( {
...
+	navigator.sendBeacon(
```
adamsilverstein, line 555: 💚 💚 💚  great to see this use `navigator.sendBeacon` (I wondered about that [previously](https://github.com/WordPress/performance/pull/978#issuecomment-1936620077))

plugins/optimization-detective/detection.php
```diff
@@ -38,13 +38,23 @@ function od_get_detection_script( string $slug, OD_URL_Metric_Group_Collection $
...
 	$web_vitals_lib_src  = add_query_arg( 'ver', $web_vitals_lib_data['version'], plugin_dir_url( __FILE__ ) . 'build/web-vitals.js' );
 
+	/**
+	 * Filters the list of extension script module URLs to import when performing detection.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string[] $extension_module_urls Extension module URLs.
+	 */
+	$extension_module_urls = (array) apply_filters( 'od_extension_module_urls', array() );
```
adamsilverstein, line 48: Great!

plugins/optimization-detective/detect.js
westonruter: One nonce is just for access to the REST API. 

The other nonce is to authorize the specific the action of storing the URL metric. In reality, the nonce here is serving as an [HMAC](https://en.wikipedia.org/wiki/HMAC). It is guaranteeing that the URL being submitted and the associated query vars hash slug are valid.

adamsilverstein (commented):
Looks great, left a few questions

plugins/optimization-detective/detect.js
westonruter: Maybe this should be renamed from `nonce` to `hmac` but that should probably be done in another PR since the `nonce` param being set here is just a refactor from what is in `trunk` now. It's just that before it was setting it in the JSON body, whereas now it is being set as a GET param.

plugins/image-prioritizer/helper.php
adamsilverstein: ah, good point. I was trying to make this more helpful... maybe don't show the message to users that can't upgrade plugins?

westonruter: True, but I think this might be overkill. It would be unlikely that someone would update Image Prioritizer but not also update Optimization Detective.

adamsilverstein (approved):

PR #1374, opened 2024-07-18
Fix disabled options from reappearing in Site Health after external update

### LINKED ISSUE #1340: Autoloaded Options Health Check: Disabled options reappear in Site Health after external update
## Bug Description

In #1048, support for changing the autoload value for the largest autoloaded options in the Site Health check was added. However, there is a potential bug when the option value is updated through external code, meaning some external code autoloads the option.

## Steps to Reproduce

Add filter code to show the site health check.

If you use WP 6.6 version, use the filter:
```php
add_filter( 'site_status_autoloaded_options_size_limit', '__return_zero' );
```

If you use WP < 6.6 version, use the filter:
```php
add_filter( 'perflab_aao_autoloaded_options_limit_size_in_bytes', '__return_zero' );
```

1. Go to Admin Panel > Site Health > Autoloaded options could affect performance.
2. Click on 'Disable autoload' for any option.
3. Check that the option is moved to the new disabled option table and removed from the top table.
4. Update that option through code `update_option( 'option_name', 'value', true );`.
5. Reload the site health page.
6. See the error. The option is displayed in both tables.

## Screenshots

Screenshot 2024-07-10 at 6 20 49 PM

## Suggested Approach

We should perform an early check in the disabled table to see if any option is autoloaded again, and if so, remove that option from `perflab_aao_disabled_options` so it will not show up again. Additionally, when anyone clicks on "Revert to autoload" for a duplicate option, it shows a `Failed to disable autoload.` error.

mukeshpanchal27:
## Summary

Fixes #1340

### DIFF
--- a/plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
+++ b/plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
@@ -211,7 +211,24 @@ function perflab_aao_get_autoloaded_options_table(): string {
 function perflab_aao_get_disabled_autoloaded_options_table(): string {
 	$disabled_options = get_option( 'perflab_aao_disabled_options', array() );
 
-	if ( ! is_array( $disabled_options ) ) {
+	$autoload_option_names = array_map(
+		static function ( $option ) {
+			return $option->option_name;
+		},
+		perflab_aao_query_autoloaded_options()
+	);
+
+	// Filter out the $disabled_options that are present in $autoload_option_names.
+	$disabled_options = array_filter(
+		$disabled_options,
+		static function ( $option ) use ( $autoload_option_names ) {
+			return ! in_array( $option, $autoload_option_names, true );
+		}
+	);
+
+	update_option( 'perflab_aao_disabled_options', $disabled_options );
+
+	if ( ! $disabled_options ) {
 		return '';
 	}
 
--- a/plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
+++ b/plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
@@ -212,7 +212,7 @@ function perflab_aao_get_disabled_autoloaded_options_table(): string {
 	$disabled_options = (array) get_option( 'perflab_aao_disabled_options', array() );
 
 	$autoload_option_names = array_map(
-		static function ( $option ) {
+		static function ( object $option ): string {
 			return $option->option_name;
 		},
 		perflab_aao_query_autoloaded_options()
@@ -221,14 +221,14 @@ function perflab_aao_get_disabled_autoloaded_options_table(): string {
 	// Filter out the $disabled_options that are present in $autoload_option_names.
 	$disabled_options = array_filter(
 		$disabled_options,
-		static function ( $option ) use ( $autoload_option_names ) {
+		static function ( $option ) use ( $autoload_option_names ): bool {
 			return ! in_array( $option, $autoload_option_names, true );
 		}
 	);
 
 	update_option( 'perflab_aao_disabled_options', $disabled_options );
 
-	if ( ! $disabled_options ) {
+	if ( count( $disabled_options ) === 0 ) {
 		return '';
 	}
 
--- a/plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
+++ b/plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
@@ -209,26 +209,9 @@ function perflab_aao_get_autoloaded_options_table(): string {
  * @return string HTML formatted table.
  */
 function perflab_aao_get_disabled_autoloaded_options_table(): string {
-	$disabled_options = (array) get_option( 'perflab_aao_disabled_options', array() );
+	$disabled_options = get_option( 'perflab_aao_disabled_options', array() );
 
-	$autoload_option_names = array_map(
-		static function ( object $option ): string {
-			return $option->option_name;
-		},
-		perflab_aao_query_autoloaded_options()
-	);
-
-	// Filter out the $disabled_options that are present in $autoload_option_names.
-	$disabled_options = array_filter(
-		$disabled_options,
-		static function ( $option ) use ( $autoload_option_names ): bool {
-			return ! in_array( $option, $autoload_option_names, true );
-		}
-	);
-
-	update_option( 'perflab_aao_disabled_options', $disabled_options );
-
-	if ( count( $disabled_options ) === 0 ) {
+	if ( ! is_array( $disabled_options ) ) {
 		return '';
 	}
 
--- a/plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
+++ b/plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
@@ -209,7 +209,7 @@ function perflab_aao_get_autoloaded_options_table(): string {
  * @return string HTML formatted table.
  */
 function perflab_aao_get_disabled_autoloaded_options_table(): string {
-	$disabled_options = get_option( 'perflab_aao_disabled_options', array() );
+	$disabled_options = (array) get_option( 'perflab_aao_disabled_options', array() );
 
 	$autoload_option_names = array_map(
 		static function ( $option ) {
--- a/plugins/performance-lab/includes/site-health/audit-autoloaded-options/hooks.php
+++ b/plugins/performance-lab/includes/site-health/audit-autoloaded-options/hooks.php
@@ -129,3 +129,27 @@ function perflab_aao_extend_core_check( string $description ): string {
 	return $description . perflab_aao_get_autoloaded_options_table() . perflab_aao_get_disabled_autoloaded_options_table();
 }
 add_filter( 'site_status_autoloaded_options_limit_description', 'perflab_aao_extend_core_check' );
+
+/**
+ * Filter the list of disabled options to exclude options that are autoloaded.
+ *
+ * This filter modifies the 'option_perflab_aao_disabled_options' to ensure
+ * that autoloaded options are not included in the disabled options list.
+ *
+ * @since n.e.x.t
+ *
+ * @param array $disabled_options Array of disabled options.
+ * @return array Filtered array of disabled options excluding autoloaded options.
+ */
+add_filter(
+	'option_perflab_aao_disabled_options',
+	static function ( $disabled_options ): array {
+		$autoload_option_names = wp_list_pluck( perflab_aao_query_autoloaded_options(), 'option_name' );
+		return array_filter(
+			(array) $disabled_options,
+			static function ( $option ) use ( $autoload_option_names ): bool {
+				return ! in_array( $option, $autoload_option_names, true );
+			}
+		);
+	}
+);
--- a/plugins/performance-lab/includes/site-health/audit-autoloaded-options/hooks.php
+++ b/plugins/performance-lab/includes/site-health/audit-autoloaded-options/hooks.php
@@ -131,25 +131,23 @@ function perflab_aao_extend_core_check( string $description ): string {
 add_filter( 'site_status_autoloaded_options_limit_description', 'perflab_aao_extend_core_check' );
 
 /**
- * Filter the list of disabled options to exclude options that are autoloaded.
+ * Filters the list of disabled options to exclude options that are autoloaded.
  *
  * This filter modifies the 'option_perflab_aao_disabled_options' to ensure
  * that autoloaded options are not included in the disabled options list.
  *
  * @since n.e.x.t
  *
- * @param array $disabled_options Array of disabled options.
- * @return array Filtered array of disabled options excluding autoloaded options.
+ * @param string[]|mixed $disabled_options Array of disabled options.
+ * @return string[] Filtered array of disabled options excluding autoloaded options.
  */
-add_filter(
-	'option_perflab_aao_disabled_options',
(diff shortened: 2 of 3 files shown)


plugins/performance-lab/tests/includes/site-health/audit-autoloaded-options/test-audit-autoloaded-options.php
```diff
@@ -187,6 +187,56 @@ static function () {
...
+		$user_id = self::factory()->user->create( array( 'role' => 'administrator' ) );
+		wp_set_current_user( $user_id );
+
+		// Mock wp_redirect to avoid actual redirection.
+		add_filter(
+			'wp_redirect',
+			static function () {
+				return false;
+			}
+		);
```
swissspidy: Where is a redirect happening?

```suggestion
		add_filter( 'wp_redirect', '__return_false' );
```

plugins/performance-lab/tests/includes/site-health/audit-autoloaded-options/test-audit-autoloaded-options.php
```diff
@@ -187,6 +187,56 @@ static function () {
...
+		$table_html = perflab_aao_get_disabled_autoloaded_options_table();
+		$this->assertStringNotContainsString( self::AUTOLOADED_OPTION_KEY, $table_html );
+
+		// Remove the mock filter.
+		remove_filter(
+			'wp_redirect',
+			static function () {
+				return false;
+			}
+		);
```
swissspidy:
```suggestion
		remove_filter( 'wp_redirect', '__return_false' );
```

swissspidy: Also, usually you'd want to do this cleanup _before_ any assertions run. Otherwise, if any assertion fails, this code will not run.

westonruter: Actually, there is no need for this `remove_filter` because tests automatically reset hooks after each test.

westonruter:
```suggestion
```

plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -211,7 +211,24 @@ function perflab_aao_get_autoloaded_options_table(): string {
...
+	$disabled_options = array_filter(
+		$disabled_options,
+		static function ( $option ) use ( $autoload_option_names ) {
+			return ! in_array( $option, $autoload_option_names, true );
+		}
+	);
+
+	update_option( 'perflab_aao_disabled_options', $disabled_options );
+
+	if ( ! $disabled_options ) {
```
westonruter: Per strict rules:

```suggestion
	if ( count( $disabled_options ) === 0 ) {
```

plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -211,7 +211,24 @@ function perflab_aao_get_autoloaded_options_table(): string {
 function perflab_aao_get_disabled_autoloaded_options_table(): string {
 	$disabled_options = get_option( 'perflab_aao_disabled_options', array() );
 
-	if ( ! is_array( $disabled_options ) ) {
```
westonruter: This type check should remain somewhere because the option could be populated with anything. If removing it, then at least add an `(array)` cast to the `get_option()` call.

plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -211,7 +211,24 @@ function perflab_aao_get_autoloaded_options_table(): string {
...
+	$autoload_option_names = array_map(
+		static function ( $option ) {
```
westonruter:
```suggestion
		static function ( object $option ): string {
```

plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -211,7 +211,24 @@ function perflab_aao_get_autoloaded_options_table(): string {
...
+			return $option->option_name;
+		},
+		perflab_aao_query_autoloaded_options()
+	);
+
+	// Filter out the $disabled_options that are present in $autoload_option_names.
+	$disabled_options = array_filter(
+		$disabled_options,
+		static function ( $option ) use ( $autoload_option_names ) {
```
westonruter: At this point we aren't 100% guaranteed that `$option` is a string since `$disabled_options` is (maybe) an array (of maybe strings). But this is OK if we are doing `in_array()` strict checks. At least we can add a return type hint:

```suggestion
		static function ( $option ) use ( $autoload_option_names ): bool {
```

plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -211,7 +211,24 @@ function perflab_aao_get_autoloaded_options_table(): string {
...
+			return ! in_array( $option, $autoload_option_names, true );
+		}
+	);
+
+	update_option( 'perflab_aao_disabled_options', $disabled_options );
```
westonruter: Shouldn't this only do an update if there was a change made to the options? Granted, `update_option()` will only proceed if there is a change with the old value. But it looks a bit strange here to be calling the update in a function that is just getting the table.

plugins/performance-lab/tests/includes/site-health/audit-autoloaded-options/test-audit-autoloaded-options.php
mukeshpanchal27: In `perflab_aao_handle_update_autoload()` function.

plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
mukeshpanchal27: Added in https://github.com/WordPress/performance/pull/1374/commits/cd7458bd51a5199a763cef865a05d300987897d5

plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
mukeshpanchal27: @westonruter Yes it look strange to update value their, I have tried to update it on `init` hook but it will reflect the value after second time reload. When we click on Disabled Autoload it shows the option in both tables but if we reload it then it removed the value from disabled table. It because of `wp_load_alloptions()` cache?

```php
add_action( 'init','perflab_aao_register_init_action' );
function perflab_aao_register_init_action() {
	if ( isset( $_GET['autoload_updated'] ) && 'true' === $_GET['autoload_updated'] ) {
		$disabled_options = (array) get_option( 'perflab_aao_disabled_options', array() );
		$autoload_option_names = array_map(
			static function ( object $option ): string {
				return $option->option_name;
			},
			perflab_aao_query_autoloaded_options()
		);

		// Filter out the $disabled_options that are present in $autoload_option_names.
		$disabled_options = array_filter(
			$disabled_options,
			static function ( $option ) use ( $autoload_option_names ): bool {
				return ! in_array( $option, $autoload_option_names, true );
			}
		);

		update_option( 'perflab_aao_disabled_options', $disabled_options );
	}
}
```

cc. @joemcgill

plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
```diff
@@ -209,9 +209,26 @@ function perflab_aao_get_autoloaded_options_table(): string {
...
-	$disabled_options = get_option( 'perflab_aao_disabled_options', array() );
+	$disabled_options = (array) get_option( 'perflab_aao_disabled_options', array() );
 
-	if ( ! is_array( $disabled_options ) ) {
+	$autoload_option_names = array_map(
+		static function ( object $option ): string {
+			return $option->option_name;
+		},
+		perflab_aao_query_autoloaded_options()
+	);
```
westonruter: Another option:

```suggestion
	$autoload_option_names = wp_list_pluck( perflab_aao_query_autoloaded_options(), 'option_name' );
```

Although the PHPStan typing may not be present.

plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
westonruter: Instead of updating the `perflab_aao_disabled_options` option and had this code inside of the `perflab_aao_get_disabled_autoloaded_options_table()` function, what if you moved it to a new filter for the option value itself? Then you wouldn't even need to update the option. For example:

```php
add_filter(
	'option_perflab_aao_disabled_options',
	static function ( $disabled_options ): array {
		$autoload_option_names = wp_list_pluck( perflab_aao_query_autoloaded_options(), 'option_name' );
		return array_filter(
			(array) $disabled_options,
			static function ( $option ) use ( $autoload_option_names ): bool {
				return ! in_array( $option, $autoload_option_names, true );
			}
		);
	}
);
```

plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
swissspidy: > Although the PHPStan typing may not be present.

Might be something we could contribute to the WP stubs

westonruter: Yeah, the current stub doesn't have any typing beyond what is already in core:

```php
    /**
     * Plucks a certain field out of each object or array in an array.
     *
     * This has the same functionality and prototype of
     * array_column() (PHP 5.5) but also supports objects.
     *
     * @since 3.1.0
     * @since 4.0.0 $index_key parameter added.
     * @since 4.7.0 Uses `WP_List_Util` class.
     *
     * @param array      $input_list List of objects or arrays.
     * @param int|string $field      Field from the object to place instead of the entire object.
     * @param int|string $index_key  Optional. Field from the object to use as keys for the new array.
     *                               Default null.
     * @return array Array of found values. If `$index_key` is set, an array of found values with keys
     *               corresponding to `$index_key`. If `$index_key` is null, array keys from the original
     *               `$input_list` will be preserved in the results.
     */
    function wp_list_pluck($input_list, $field, $index_key = \null)
    {
    }
```

plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
mukeshpanchal27: It will works. Thanks!

plugins/performance-lab/includes/site-health/audit-autoloaded-options/helper.php
mukeshpanchal27: Thanks both for feedback. I have reverted the change for `perflab_aao_get_disabled_autoloaded_options_table` so we don't needs to update this. Also PHPSTAN did not show any error if we use `wp_list_pluck` in `hooks.php`. Did i miss anything?

plugins/performance-lab/includes/site-health/audit-autoloaded-options/hooks.php
```diff
@@ -129,3 +129,27 @@ function perflab_aao_extend_core_check( string $description ): string {
...
+	static function ( $disabled_options ): array {
+		$autoload_option_names = wp_list_pluck( perflab_aao_query_autoloaded_options(), 'option_name' );
+		return array_filter(
+			(array) $disabled_options,
+			static function ( $option ) use ( $autoload_option_names ): bool {
+				return ! in_array( $option, $autoload_option_names, true );
+			}
+		);
+	}
+);
```
westonruter: While in my POC code I used a closure, you should probably use a named function in the actual codebase.

```suggestion
/**
 * Filters the list of disabled options to exclude options that are autoloaded.
 *
 * This filter modifies the 'option_perflab_aao_disabled_options' to ensure
 * that autoloaded options are not included in the disabled options list.
 *
 * @since n.e.x.t
 *
 * @param string[]|mixed $disabled_options Array of disabled options.
 * @return string[] Filtered array of disabled options excluding autoloaded options.
 */
function perflab_filter_option_perflab_aao_disabled_options( $disabled_options ): array {
	$autoload_option_names = wp_list_pluck( perflab_aao_query_autoloaded_options(), 'option_name' );
	return array_filter(
		(array) $disabled_options,
		static function ( $option ) use ( $autoload_option_names ): bool {
			return ! in_array( $option, $autoload_option_names, true );
		}
	);
}
add_filter( 'option_perflab_aao_disabled_options', 'perflab_filter_option_perflab_aao_disabled_options' );
```

plugins/performance-lab/tests/includes/site-health/audit-autoloaded-options/test-audit-autoloaded-options.php
```diff
@@ -177,14 +172,43 @@ static function () {
...
+	}
 
-		// Remove the mock filter.
-		remove_filter(
-			'wp_redirect',
-			static function () {
-				return false;
-			}
-		);
+	public function test_perflab_aao_autoloaded_options_auto_enable_functionality(): void {
```
westonruter, line 182: Would be good to add the `@covers` tags for which functions this is intending to test.

mukeshpanchal27:

westonruter (approved):
LGTM

adamsilverstein (approved):
Fantastic

PR #1377, opened 2024-07-18
Prioritize loading poster image of video LCP elements

### LINKED ISSUE #1183: Prioritize loading poster image of video LCP elements
## Feature Description

When a `video` is the LCP element and it has a `poster` image, we should prioritize the loading of that poster image with a `fetchpriority=high` preload link. This would be a simple addition to the existing logic that does the same for an `img` or `background-image` which is the LCP element.

This can be done as part of the Image Prioritizer plugin. See #1088.

Aside: if there is no `poster`, should an LCP `video` element also preload the video URL as well?

thelovekesh:
## Summary

Fixes #1183

## Relevant technical choices

- Preload poster images with high-priority video LCP elements.

### DIFF
--- a/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
@@ -1,0 +1,94 @@
+<?php
+/**
+ * Tag visitor that optimizes VIDEO tags:
+ * - Adds preload links for poster images if in a breakpoint group's LCP.
+ *
+ * @package image-prioritizer
+ *
+ * @since n.e.x.t
+ */
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Image Prioritizer: Image_Prioritizer_Video_Tag_Visitor class
+ *
+ * @since n.e.x.t
+ *
+ * @access private
+ */
+final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Visitor {
+
+	/**
+	 * Video tag.
+	 *
+	 * @var string
+	 */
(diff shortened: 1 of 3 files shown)


plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -0,0 +1,94 @@
...
+					static function ( string $value ): bool {
+						return '' !== $value;
+					}
+				)
+			);
+
+			$crossorigin = $processor->get_attribute( 'crossorigin' );
+			if ( is_string( $crossorigin ) ) {
+				$link_attributes['crossorigin'] = 'use-credentials' === $crossorigin ? 'use-credentials' : 'anonymous';
+			}
```
thelovekesh, line 81: @westonruter The Poster image uses the `crossorigin` from the video tag, right?

westonruter: I presume so, yes.

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -0,0 +1,94 @@
...
+						'href' => (string) $processor->get_attribute( self::POSTER ),
+					),
+					static function ( string $value ): bool {
+						return '' !== $value;
+					}
+				)
+			);
+
+			$crossorigin = $processor->get_attribute( 'crossorigin' );
+			if ( is_string( $crossorigin ) ) {
```
westonruter, line 79: In the case of `<video src="..." crossorigin>` then it would actually be `true`. So I think what you want there is:

```suggestion
			if ( null !== $crossorigin ) {
```

thelovekesh:
Closing in favor of https://github.com/WordPress/performance/pull/1498. cc @westonruter

PR #1382, opened 2024-07-22
Update `wp_calculate_image_sizes` to Reflect Changes in sizes attribute

### LINKED ISSUE #1381: Update `wp_calculate_image_sizes` to Reflect Changes in sizes attribute
## Bug Description

As discussed in https://github.com/WordPress/performance/pull/1354#pullrequestreview-2185993737, the `wp_calculate_image_sizes()` function does not currently account for the changes made to the sizes attribute in PR #1329.

We need to update the sizes value in `wp_calculate_image_sizes()` so that any usage of this core function benefits from the improved sizes attribute. 

Alternatively, we should ensure that the enhanced sizes attribute is applied via the `wp_calculate_image_sizes` filter, so any other code that directly calls `wp_calculate_image_sizes()` will also benefit from the optimizations.

mukeshpanchal27:
## Summary

Fixes #1381

POC that fix the `wp_calculate_image_sizes()` result.


mukeshpanchal27:
@joemcgill @westonruter The POC PR is ready for internal review.

plugins/auto-sizes/tests/test-improve-sizes.php
```diff
@@ -409,4 +409,50 @@ public function test_no_image(): void {
...
+		$improve_sizes = wp_calculate_image_sizes( $image_size, $image_url, '', self::$image_id );
+
+		$this->assertStringContainsString( $improve_sizes, $result );
+		$this->assertStringNotContainsString( $sizes, $result );
+		$this->assertNotSame( $improve_sizes, $sizes );
+
+		remove_filter( 'wp_content_img_tag', 'auto_sizes_improve_image_sizes_attributes', 9 );
+
+		$new_sizes = wp_calculate_image_sizes( $image_size, $image_url, '', self::$image_id );
+		$this->assertSame( $improve_sizes, $new_sizes );
```
adamsilverstein: why would these match after removing the filter?

some inline docs to explain the tests would be helpful.

plugins/auto-sizes/hooks.php
```diff
@@ -206,6 +206,14 @@ function auto_sizes_improve_image_sizes_attributes( string $content ): string {
 
 	if ( $sizes ) {
 		$processor->set_attribute( 'sizes', $sizes );
+
+		// The sizes filter to reflate improve sizes calculations.
```
westonruter: Reflate?

plugins/auto-sizes/hooks.php
```diff
@@ -206,6 +206,14 @@ function auto_sizes_improve_image_sizes_attributes( string $content ): string {
...
+		add_filter(
+			'wp_calculate_image_sizes',
+			static function () use ( $sizes ): string {
+				return $sizes;
+			}
+		);
```
westonruter: I don't understand why this filter is being added without being removed. Every time that `wp_filter_content_tags()` is called, another filter `wp_calculate_image_sizes` will be added. And since it is a closure here, the `$sizes` value from one image could end up getting applied to another image unexpectedly?

westonruter (changes requested):

plugins/auto-sizes/hooks.php
mukeshpanchal27: Nice catch. Added remove filter code.

plugins/auto-sizes/tests/test-improve-sizes.php
mukeshpanchal27: Thanks. Inline comment added in https://github.com/WordPress/performance/pull/1382/commits/7b2dcf9d96c3fd10c3e6b72b3a59202a01419f2b

plugins/auto-sizes/hooks.php
```diff
@@ -203,7 +203,23 @@ function auto_sizes_improve_image_sizes_attributes( string $content ): string {
...
+
 		$processor->set_attribute( 'sizes', $sizes );
+
+		// Remove the filter to update the sizes calculations to "wp_calculate_image_sizes".
+		remove_filter(
+			'wp_calculate_image_sizes',
+			static function () use ( $sizes ): string {
+				return $sizes;
+			}
+		);
```
joemcgill: I generally prefer to avoid the need to add hooks from another hook's callback, and particularly as anonymous closures, see [this section](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#closures-anonymous-functions) in the Core handbook on Closures.

> Closures should not be passed as filter or action callbacks, as removing these via remove_action() / remove_filter() is complex (at this time) (see [#46635](https://core.trac.wordpress.org/ticket/46635) for a proposal to address this). 

Additionally, I don't think this will actually do anything, because the filter is added and then removed before  `wp_calculate_image_sizes()` is ever called.

plugins/auto-sizes/tests/test-improve-sizes.php
```diff
@@ -409,4 +409,49 @@ public function test_no_image(): void {
...
+		$block_content = '<!-- wp:image {"id":' . self::$image_id . ',"sizeSlug":"' . $image_size . '","linkDestination":"none"} --><figure class="wp-block-image size-' . $image_size . '"><img src="' . wp_get_attachment_image_url( self::$image_id, $image_size ) . '" alt="" class="wp-image-' . self::$image_id . '"/></figure><!-- /wp:image -->';
+
+		$result = apply_filters( 'the_content', $block_content );
+
+		$this->assertStringContainsString( $sizes, $result, 'Make sure the sizes is present in result.' );
+
+		$new_sizes = wp_calculate_image_sizes( $image_size, $image_url, '', self::$image_id );
+
+		$this->assertStringContainsString( $new_sizes, $result, 'Make sure the sizes is present in result.' );
+		$this->assertSame( $new_sizes, $sizes, 'Make sure the sizes value are the same' );
```
joemcgill: This is comparing the result of two calls to `wp_calculate_image_sizes()` with the same parameters passed, so I would expect it to pass. Did you intend to test something else?

plugins/auto-sizes/tests/test-improve-sizes.php
```diff
@@ -409,4 +409,49 @@ public function test_no_image(): void {
...
+		$block_content = '<!-- wp:image {"id":' . self::$image_id . ',"sizeSlug":"' . $image_size . '","linkDestination":"none"} --><figure class="wp-block-image size-' . $image_size . '"><img src="' . wp_get_attachment_image_url( self::$image_id, $image_size ) . '" alt="" class="wp-image-' . self::$image_id . '"/></figure><!-- /wp:image -->';
+
+		$result = apply_filters( 'the_content', $block_content );
+
+		// Retrieves the sizes from the core function. It should return same sizes as accurate ones.
+		$improve_sizes = wp_calculate_image_sizes( $image_size, $image_url, '', self::$image_id );
+
+		$this->assertStringContainsString( $improve_sizes, $result, 'Make sure the improve sizes is present in result.' );
+		$this->assertStringNotContainsString( $sizes, $result, 'Make sure the old sizes is present in result.' );
+		$this->assertNotSame( $improve_sizes, $sizes, 'Make sure the sizes are not same.' );
```
joemcgill: Similarly, this is testing the output of the same function called twice with the same input.

joemcgill (changes requested):

plugins/auto-sizes/hooks.php
westonruter: Right, this is how this should be done:

```suggestion
		// Add the filter to update the sizes calculations to "wp_calculate_image_sizes".
		$filter = static function () use ( $sizes ): string {
			return $sizes;
		};
		add_filter( 'wp_calculate_image_sizes', $filter );

		$processor->set_attribute( 'sizes', $sizes );

		// Remove the filter to update the sizes calculations to "wp_calculate_image_sizes".
		remove_filter( 'wp_calculate_image_sizes', $filter );
```

A closure is an object and it needs to be the same reference when added/removed. Two identical closures are different object references.

But yeah, I also don't see how this filter is doing anything since the only code between the `add_filter()`/`remove_filter()` calls is `$processor->set_attribute( 'sizes', $sizes );` which has no side effects.

mukeshpanchal27: Thanks for the feedback.

The `remove_filter` was added because of the feedback in https://github.com/WordPress/performance/pull/1382#discussion_r1687087689. As noted in https://github.com/WordPress/performance/issues/1381#issuecomment-2245637818, there is no way to get the required details in the `wp_calculate_image_sizes` filter, so we need to add the filter here to reflect the changes. Does that make sense?

plugins/auto-sizes/tests/test-improve-sizes.php
mukeshpanchal27: @joemcgill I want to test the `wp_calculate_image_sizes` will return the updates sizes or not.

mukeshpanchal27:
@westonruter (Joe OoO), In https://github.com/WordPress/performance/pull/1382/commits/01d1d567aaf6e048c033f04bcd6b8e97479008c9, I added the closures. After that, we didn't get the updated sizes in `wp_calculate_image_sizes()`. See the `test_wp_calculate_image_sizes_with_improve_sizes_filter` test, where I had to remove the assert that tested the new sizes. Did I miss anything?

plugins/auto-sizes/hooks.php
```diff
@@ -203,7 +203,16 @@ function auto_sizes_improve_image_sizes_attributes( string $content ): string {
...
+		// Add the filter to update the sizes calculations to "wp_calculate_image_sizes".
+		$filter = static function () use ( $sizes ): string {
+			return $sizes;
+		};
+		add_filter( 'wp_calculate_image_sizes', $filter );
+
 		$processor->set_attribute( 'sizes', $sizes );
+
+		// Remove the filter to update the sizes calculations to "wp_calculate_image_sizes".
+		remove_filter( 'wp_calculate_image_sizes', $filter );
```
westonruter: I don't see how this code will have any effect. The `wp_calculate_image_sizes` filter is applied when `wp_calculate_image_sizes()` is called. But `wp_calculate_image_sizes()` is not called here. The only code being called when the filter is applied is `$processor->set_attribute( 'sizes', $sizes )`, and this does not have any side effects. I don't see how this filter is supposed to fix things.

westonruter: In other words, with these changes removed, the tests you added still pass. So the tests don't seem to be checking the problem.

plugins/auto-sizes/hooks.php
westonruter: Sorry, not really. See https://github.com/WordPress/performance/pull/1382#discussion_r1690488581

plugins/auto-sizes/hooks.php
mukeshpanchal27: I have pushed the changes that show the failed tests: https://github.com/WordPress/performance/actions/runs/10089272264/job/27896437786?pr=1382#step:9:32.

What I want to achieve with this PR:
- When someone uses the core `wp_calculate_image_sizes()`, it should return core sizes, not the improved sizes.
- When the image is filtered through the improved sizes filter `add_filter( 'wp_content_img_tag', 'auto_sizes_improve_image_sizes_attributes', 9 );`, the core `wp_calculate_image_sizes()` should return the new sizes, not the core sizes.

This PR is a proof of concept, but I'm happy to update the approach.

westonruter: What I mean is that there should be tests that fail when this added `$filter` code is removed. As it stands, I don't see how these lines will have any effect. So I guess that is reflected in the new tests that now fail?

mukeshpanchal27:
@westonruter I have slightly updated the approach.

- Updated sizes only for the images that are processed through the improved sizes filter. Added an Image ID check in the filter so it will not affect any other images.
- Removed the `remove_filter`.
- Updated unit tests to check the new changes.

westonruter:
@mukeshpanchal27:

> * Removed the `remove_filter`.

When a filter is added, especially with a closure, and never removed, then every time here the `auto_sizes_improve_image_sizes_attributes()` function is called it will add another instance of the closure. To clean up after itself, shouldn't the filter remove itself once it is called What if someone has embedded the same image with two different sizes? Consider the case where someone embeds a thumbnail image along with a full size image that is shown when the thumbnail is clicked. Wouldn't the current filter being added potentially result in the wrong sizes being applied to the second image due to the lack of cleanup?

So to clarify my understanding, given this code:

image

Where is the filter being added here expected to apply? It couldn't be applied in the scope of that function since the remainder of the function is just updating some HTML with pre-computed values. So it's applying in some function after this function returns, correct? There is the `wp_filter_content_tags()` function in which the `wp_content_img_tag` filter applies. Then there is the `wp_image_add_srcset_and_sizes()` function which calls `wp_calculate_image_sizes()` which then applies the `wp_calculate_image_sizes` filter. The `wp_img_tag_add_srcset_and_sizes_attr()` function calls `wp_image_add_srcset_and_sizes()`, which is called by `wp_filter_content_tags()`: the same function that later applies the `wp_content_img_tag` filter. So I'm confused how adding the filter here will apply to code that has already been called?

Here is the [code in question](https://github.com/WordPress/wordpress-develop/blob/883d9b673185e9f0b40ac9929a59f4725af379d9/src/wp-includes/media.php#L1912-L1929):

> Screenshot 2024-07-29 09 43 16

So the `wp_calculate_image_sizes` filter applies first in `wp_img_tag_add_srcset_and_sizes_attr()`, and then later it applies the `wp_filter_content_tags` filters. But your code is running at the `wp_filter_content_tags` filter which adds a `wp_calculate_image_sizes` filter. But wouldn't that have already been applied at this point?

mukeshpanchal27:
Thanks, @westonruter, for the details. Your point is valid as the filter didn't change anything for `wp_calculate_image_sizes` with different image sizes.

For now, I'm going to close this PR and will open a new one once the best path is decided.

PR #1391, opened 2024-07-23
Eliminate use of empty() in Embed Optimizer and Optimization Detective

westonruter:
This is a follow-up to #1241 to edge closer to full strict rule compliance for PHPStan. See #1219

See also https://github.com/WordPress/performance/issues/775.

### DIFF
--- a/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
+++ b/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
@@ -60,44 +60,47 @@ final class Embed_Optimizer_Tag_Visitor {
 			 * for GET requests, as POST requests are not likely to be part of the critical rendering path.
 			 */
 			$preconnect_hrefs = array();
-			if ( true === $processor->has_class( 'wp-block-embed-youtube' ) ) {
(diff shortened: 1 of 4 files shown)


plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -60,44 +60,47 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
 			 * for GET requests, as POST requests are not likely to be part of the critical rendering path.
 			 */
 			$preconnect_hrefs = array();
-			if ( true === $processor->has_class( 'wp-block-embed-youtube' ) ) {
+			$has_class        = static function ( string $wanted_class ) use ( $processor ): bool {
+				return true === $processor->has_class( $wanted_class );
+			};
```
swissspidy, line 65: Is this just to reduce repetition below, or actually because of PHPStan? 🤔

swissspidy (approved):

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
westonruter: Ah yes, just to reduce repetition.

PR #1392, opened 2024-07-23
Add strict PHPStan rules for Speculative Loading

westonruter:
This is a follow-up to #1241 to edge closer to full strict rule compliance for PHPStan. See #1219.

See also https://github.com/WordPress/performance/issues/775.

This improves typing of functions in the Speculative Loading plugin, and consolidates the obtaining of the stored option value in the `plsr_get_stored_setting_value()` function.

### DIFF
--- a/plugins/speculation-rules/hooks.php
+++ b/plugins/speculation-rules/hooks.php
@@ -19,13 +19,8 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @since 1.0.0
  */
 function plsr_print_speculation_rules(): void {
-	$rules = plsr_get_speculation_rules();
-	if ( empty( $rules ) ) {
-		return;
-	}
-
 	wp_print_inline_script_tag(
-		(string) wp_json_encode( $rules ),
+		(string) wp_json_encode( plsr_get_speculation_rules() ),
 		array( 'type' => 'speculationrules' )
 	);
 }
--- a/plugins/speculation-rules/settings.php
+++ b/plugins/speculation-rules/settings.php
@@ -79,6 +79,7 @@ function plsr_get_stored_setting_value(): array {
  * Sanitizes the setting for Speculative Loading configuration.
  *
  * @since 1.0.0
+ * @todo  Consider whether the JSON schema for the setting could be reused here.
  *
  * @param mixed $input Setting to sanitize.
  * @return array{ mode: 'prefetch'|'prerender', eagerness: 'conservative'|'moderate'|'eager' } {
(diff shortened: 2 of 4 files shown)


plugins/speculation-rules/hooks.php
```diff
@@ -19,13 +19,8 @@
  * @since 1.0.0
  */
 function plsr_print_speculation_rules(): void {
-	$rules = plsr_get_speculation_rules();
-	if ( empty( $rules ) ) {
```
westonruter, line 23: This function never returns an empty array, so this condition is dead code.

plugins/speculation-rules/settings.php
```diff
@@ -197,28 +212,24 @@ static function (): void {
  * }
  */
 function plsr_render_settings_field( array $args ): void {
-	if ( empty( $args['field'] ) || empty( $args['title'] ) ) { // Invalid.
-		return;
-	}
```
westonruter, line 202: The `field` and `title` args are always provided in `plsr_add_setting_ui()`.

plugins/speculation-rules/settings.php
```diff
@@ -230,17 +241,11 @@ function plsr_render_settings_field( array $args ): void {
 					<?php echo esc_html( $label ); ?>
 				</label>
 			</p>
-			<?php
-		}
+		<?php endforeach; ?>
 
-		if ( ! empty( $args['description'] ) ) {
```
westonruter, line 236: The description is always provided in `plsr_add_setting_ui()`.

plugins/speculation-rules/settings.php
```diff
@@ -59,13 +59,29 @@ function plsr_get_setting_default(): array {
...
+ *
+ * @return array{ mode: 'prefetch'|'prerender', eagerness: 'conservative'|'moderate'|'eager' } {
+ *     Stored setting value.
+ *
+ *     @type string $mode      Mode.
+ *     @type string $eagerness Eagerness.
+ * }
+ */
+function plsr_get_stored_setting_value(): array {
+	return plsr_sanitize_setting( get_option( 'plsr_speculation_rules' ) );
```
westonruter, line 75: The `plsr_sanitize_setting()` function could perhaps be eliminated in favor of the JSON schema.

plugins/speculation-rules/settings.php
```diff
@@ -109,7 +122,7 @@ function plsr_register_setting(): void {
 		array(
 			'type'              => 'object',
 			'description'       => __( 'Configuration for the Speculation Rules API.', 'speculation-rules' ),
-			'sanitize_callback' => 'plsr_sanitize_setting',
+			'sanitize_callback' => 'plsr_sanitize_setting', // TODO: Is this even needed here due to the schema?
```
westonruter: I guess it is needed because the JSON schema is only used in a REST API request. But the `plsr_sanitize_setting()` function can reuse the schema.

swissspidy (approved):

plugins/speculation-rules/settings.php
```diff
@@ -43,9 +43,9 @@ function plsr_get_eagerness_labels(): array {
 /**
  * Returns the default setting value for Speculative Loading configuration.
  *
- * @since 1.0.0
+ * @since n.e.x.t
```
westonruter: Oops. This should not have been changed.

```suggestion
 * @since 1.0.0
```

plugins/speculation-rules/settings.php
```diff
@@ -59,13 +59,30 @@ function plsr_get_setting_default(): array {
 	);
 }
 
+/**
+ * Returns the stored setting value for Speculative Loading configuration.
+ *
+ * @since 1.0.0
```
westonruter:
```suggestion
 * @since n.e.x.t
```

Issue #1398, opened 2024-07-25
Modern Image Formats: Improve unit tests execution, to slow


mukeshpanchal27:
## Description

When we run Unit tests for Modern Image Formats (webp-uploads) it takes 3.2 minutes. We should check if anyway we can improve it. As `npm run test-php` runs all plugin test in single command so other plugins take less time then these.

## Screenshots
<img width="590" alt="Screenshot 2024-07-25 at 3 50 44 PM" src="https://github.com/user-attachments/assets/968818b2-7250-454e-94f4-91722f0ea87f">

westonruter:
As per https://github.com/WordPress/performance/issues/1397#issuecomment-2250993308:

> I've opened #1401 to help with this, but we should have commands exposed to make it easy to run the tests for individual plugins as well. This is something that @thelovekesh was exploring, but I've also opened #1402 real quick.

swissspidy:
I suppose all the image crunching takes a lot of time. Maybe we could use smaller images or so?

mukeshpanchal27:
After resetting the `wp-env` it executed in `~40 seconds`.

swissspidy:
Then I suppose we can close this one :-)

Issue #1397, opened 2024-07-25
Image Placeholder: Improve unit tests execution, to slow


mukeshpanchal27:
## Description

When we run Unit tests for Image Placeholders (dominant-color-images) it takes 4.86 minutes. We should check if anyway we can improve it. As `npm run test-php` runs all plugin test in single command so other plugins take less time then these.

## Screenshots
<img width="596" alt="Screenshot 2024-07-25 at 3 46 29 PM" src="https://github.com/user-attachments/assets/5e4f5fbd-60a8-49c4-8b74-380e8705c06f">

westonruter:
I've opened #1401 to help with this, but we should have commands exposed to make it easy to run the tests for individual plugins as well. This is something that @thelovekesh was exploring, but I've also opened #1402 real quick.

thelovekesh:
@mukeshpanchal27 On which OS/environment are you running these tests? I reviewed the CI execution time and noticed that all plugin tests typically take around 1 minute on GH runners with 16GB of RAM.

On my Linux machine with 16GB of RAM, the `dominant-color-images` test suite executed in `3.73 seconds`.

mukeshpanchal27:
After resetting the `wp-env` it executed in `~5 seconds`.

PR #1401, opened 2024-07-25
Add test-php-watch command

westonruter:
Running unit tests for all plugins can take a long time (e.g. #1397, #1398). It would be ideal to just run the tests for a single plugin during development, and for tests to run as soon as a change has been made. This PR implements a `npm run test-php-watch` command which watches for changes in the `plugins` directory and then automatically runs the unit tests for the plugin which had the changed file.

[Screen recording 2024-07-25 09.55.55.webm](https://github.com/user-attachments/assets/02f5f664-fd96-4177-a918-a9326e57ab1c)


thelovekesh:
Nice addition. There is a tool around it as well - https://github.com/spatie/phpunit-watcher.

swissspidy (approved):

PR #1403, opened 2024-07-25
Clean up object-cache drop-ins before starting tests

### LINKED ISSUE #1255: `npm run test-php` failed with error
## Bug Description

Run `npm run test-php` command to run the unit tests for all plugins. It will return error for `performance-lab` plugin.

```
> phpunit '--verbose' '--testsuite' 'performance-lab'
Installing...
Running as single site... To run multisite, use -c tests/phpunit/multisite.xml
[29-May-2024 05:17:23 UTC] PHP Fatal error:  Cannot declare class Perflab_Server_Timing_Metric, because the name is already in use in /var/www/html/wp-content/plugins/performance/plugins/performance-lab/includes/server-timing/class-perflab-server-timing-metric.php on line 14
Script phpunit handling the test event returned with error code 255
Script @test --verbose --testsuite performance-lab was called via test:performance-lab
Script @test:performance-lab was called via test:plugins
✖ Command failed with exit code 255
Command failed with exit code 255
```

## Screenshots
Screenshot 2024-05-29 at 10 51 26 AM

westonruter:
## Summary

If unit tests start while an `object-cache.php` is present in the test environment, it ends up failing with an error:

```
Running as single site... To run multisite, use -c tests/phpunit/multisite.xml
[25-Jul-2024 18:31:31 UTC] PHP Fatal error:  Cannot declare class Perflab_Server_Timing_Metric, because the name is already in use in /var/www/html/wp-content/plugins/performance/plugins/performance-lab/includes/server-timing/class-perflab-server-timing-metric.php on line 14
```

The file is apparently there due to tests not cleaning up at tear-down because of some fatal error, resulting in a non-clean initial state to bootstrap the tests.

It's not entirely clear to me why the presence of the `object-cache.php` causes this fatal error in the first place, but it certainly seems like a bug for the `object-cache.php` file to exist when the tests first bootstrap. So it should be removed.

Fixes #1255.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -110,5 +110,22 @@ if ( 'performance-lab' === $plugin_name ) {
 	);
 }
 
+// Clean up object-cache drop-ins after possible previous failed tests.
+tests_add_filter(
+	'enable_loading_object_cache_dropin',
+	static function ( bool $passthrough ): bool {
+		$cleanup_files = array(
+			WP_CONTENT_DIR . '/object-cache.php',
+			WP_CONTENT_DIR . '/object-cache-plst-orig.php',
+		);
+		foreach ( $cleanup_files as $cleanup_file ) {
+			if ( file_exists( $cleanup_file ) ) {
+				unlink( $cleanup_file );
+			}
+		}
+		return $passthrough;
+	}
+);
+
 // Start up the WP testing environment.
 require $_test_root . '/includes/bootstrap.php';


westonruter:
> It's not entirely clear to me why the presence of the `object-cache.php` causes this fatal error in the first place, but it certainly seems like a bug for the `object-cache.php` file to exist when the tests first bootstrap. So it should be removed.

I found the reason in https://github.com/WordPress/performance/issues/1255#issuecomment-2251205644

Maybe this issue could also be avoided by setting the `WP_PLUGIN_DIR` constant in the tests config so that it uses the right plugin root from the beginning?

westonruter:
> Maybe this issue could also be avoided by setting the `WP_PLUGIN_DIR` constant in the tests config so that it uses the right plugin root from the beginning?

I don't know. It's confusing how the bootstrap is loaded. I tried defining the `WP_PLUGIN_DIR` constant in our `tests/bootstrap.php` file but then when the `object-cache.php` drop-in loads, it says it isn't defined 🤷‍♂️  

All I know is that the `object-cache.php` file should not be present when bootstrapping the tests. So this PR _should_ fix that issue.

swissspidy (approved):

PR #1405, opened 2024-07-26
Move Optimization Detective test cases into separate files and add helper trait to reduce code duplication

westonruter:
The test files are too large with the cases for testing Optimization Detective behavior. They need to be split into separate files to make this easier to work with. This has been done for:

- Image Prioritizer
- Embed Optimizer
- Optimization Detective

It was overkill for Auto Sizes since the test data is smaller as it is only testing individual `img` tags not entire documents.

This PR also removes a lot of duplication across plugin test suites by moving duplicated `populate_url_metrics`, `get_validated_url_metric`, and `remove_initial_tabs` methods to a new `Optimization_Detective_Test_Helpers` helper trait.


plugins/auto-sizes/tests/test-optimization-detective.php
```diff
@@ -5,7 +5,11 @@
  * @package auto-sizes
  */
 
+require_once __DIR__ . '/../../../tests/class-optimization-detective-test-helpers.php';
```
swissspidy: We could load this just once in bootstrap, instead of in every test.

Or maybe an autoloader at some point :-)

westonruter: Yeah, adding it to the bootstrap is a good idea. I recall we did have an autoloader previously but it was removed as discussed in https://github.com/WordPress/performance/pull/1262#discussion_r1621437918

swissspidy (approved):
That's a lot of changed lines, but it makes the test cases definitely more manageable 👍

westonruter:
> That's a lot of changed lines

Yeah, sorry. But it's almost entirely just moving code from one place to another.

PR #1407, opened 2024-07-26
Add helper function to construct XPath from indexed breadcrumbs

westonruter:
See https://github.com/WordPress/performance/pull/1373#discussion_r1693643049

It would be ideal if the "indexed breadcrumbs" could be passed around in the codebase and turned into XPath strings without having to hard-code the XPath string literals.

However, passing around `array( array( 'DIV', 0 ) )` isn't necessarily feel so much better than `/*[1][self::DIV]`.

Something to let bake for awhile.


phanduynam:

PR #1408, opened 2024-07-29
Correct the fallback image in PICTURE element

mukeshpanchal27:
## Summary

This PR accounts following:
- Return `image/jpeg` as fallback image for Picture element.
- Remove `image/jpeg` mime from Picture > Source and return only additional mime.
- Use share fixture for tests.
- Add/update unit test that reflect the changes.
- Reinitialize the `WP_HTML_Tag_Processor` so it check the additional `next_tag` in while loop.


plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -176,10 +176,12 @@ public function test_img_sizes_is_equal_to_picture_source_sizes_for_picture_elem
 		$img_processor = new WP_HTML_Tag_Processor( $img_markup );
 		$this->assertTrue( $img_processor->next_tag( array( 'tag_name' => 'IMG' ) ), 'There should be an IMG tag.' );
 		$img_sizes = $img_processor->get_attribute( 'sizes' );
+		$img_src   = $img_processor->get_attribute( 'src' );
```
westonruter: Aside: It would make more sense to move these 4 lines up before `// Apply picture element support.`

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -176,10 +176,12 @@ public function test_img_sizes_is_equal_to_picture_source_sizes_for_picture_elem
...
 
 		$picture_processor = new WP_HTML_Tag_Processor( $picture_markup );
 
 		$picture_processor->next_tag( array( 'tag_name' => 'IMG' ) );
+		$this->assertSame( $img_src, $picture_processor->get_attribute( 'src' ), 'Make sure IMG src is same.' );
```
westonruter: The `IMG` tag looks like this:

```html
<img width="1024" height="683" src="http://localhost:8889/wp-content/uploads/2024/07/leaves-950-1024x683.webp" class="wp-image-9" alt="Green Leaves" decoding="async" loading="lazy" srcset="http://localhost:8889/wp-content/uploads/2024/07/leaves-950-1024x683.webp 1024w, http://localhost:8889/wp-content/uploads/2024/07/leaves-950-300x200.webp 300w, http://localhost:8889/wp-content/uploads/2024/07/leaves-950-768x512.webp 768w, http://localhost:8889/wp-content/uploads/2024/07/leaves-950-jpg.webp 1080w" sizes="(max-width: 1024px) 100vw, 1024px" />
```

And below the `PICTURE` element looks like this:

```html
<picture class="wp-picture-9" style="display: contents;">
  <source type="image/webp" srcset="http://localhost:8889/wp-content/uploads/2024/07/leaves-950-1024x683.webp 1024w, http://localhost:8889/wp-content/uploads/2024/07/leaves-950-300x200.webp 300w, http://localhost:8889/wp-content/uploads/2024/07/leaves-950-768x512.webp 768w" sizes="(max-width: 1024px) 100vw, 1024px">
  <img width="1024" height="683" src="http://localhost:8889/wp-content/uploads/2024/07/leaves-950-1024x683.webp" class="wp-image-9" alt="Green Leaves" decoding="async" loading="lazy" srcset="http://localhost:8889/wp-content/uploads/2024/07/leaves-950-1024x683.webp 1024w, http://localhost:8889/wp-content/uploads/2024/07/leaves-950-300x200.webp 300w, http://localhost:8889/wp-content/uploads/2024/07/leaves-950-768x512.webp 768w, http://localhost:8889/wp-content/uploads/2024/07/leaves-950-jpg.webp 1080w" sizes="(max-width: 1024px) 100vw, 1024px" />
</picture>
```

A couple things seem off here:

1. The `IMG` has an identical `srcset` to the `PICTURE > SOURCE`, except the `IMG` has one additional entry in the `srcset`: `http://localhost:8889/wp-content/uploads/2024/07/leaves-950-jpg.webp 1080w`. Why?

2. Apparently JPEG fallbacks are not enabled, so this is generating a `PICTURE` for no reason, since both the `PICTURE > SOURCE` and the `PICTURE > IMG` are both serving WebP. If you look at `webp_uploads_is_picture_element_enabled()` you can see that it only returns true if both the the `perflab_generate_webp_and_jpeg` and `webp_uploads_use_picture_element` options are `true`, and so in `hooks.php` it only adds the `webp_uploads_wrap_image_in_picture` filter when they are both true. So to fix this in the test, I think the `opt_in_to_picture_element` helper should be changed to also do:

```php
update_option( 'perflab_generate_webp_and_jpeg', '1' );
```

When this is done, then the effect will be a unit test failure because when the picture element is not enabled, the `IMG` should be outputting the modern image format, whereas when the picture element is enabled then the `PICTURE > IMG` should _not_ be the modern image format but rather the JPEG fallback.

westonruter (changes requested):

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: @westonruter Thanks for the feedback.

> The IMG has an identical srcset to the PICTURE > SOURCE, except the IMG has one additional entry in the srcset: http://localhost:8889/wp-content/uploads/2024/07/leaves-950-jpg.webp 1080w. Why?

Per https://github.com/WordPress/performance/blob/trunk/plugins/webp-uploads/picture-element.php#L33-L42 the scrset is account only for sub sizes for the img that's why it get missed in PICTURE > SOURCE. I will open separate tracking issue for that. 


There is some juggling in unit tests it, also found some bug in code that replace original IMG to alternate mime type through `add_filter( 'the_content', 'webp_uploads_update_image_references', 13 )` filter.

For your information, I have update PR with the following:
- Removed the filter that add through function for unit test, instead use options that add the functionality.
```
public function opt_in_to_picture_element(): void {
	add_filter( 'wp_content_img_tag', 'webp_uploads_wrap_image_in_picture', 10, 3 );
}
```
- Remove the shared fixtures as it return WEBP image instead JPEG image. It because when image get uploaded it will update with the alternate mime type it was WEBP.
- Check IMG and Picture element IMG src should end with `.jpg` (as image is `leaves.jpg`)
- Remove `add_filter( 'the_content', 'webp_uploads_update_image_references', 13 )` filter that update IMG tag with alternate mime type.
- Check Picture element requirement condition in filter and bail early if not met.


**IMG markup:**
```
<img width="1024" height="683" src="http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-1024x683.jpg" class="wp-image-74" alt="Green Leaves" decoding="async" loading="lazy" srcset="http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-1024x683.jpg 1024w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-300x200.jpg 300w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-768x512.jpg 768w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1087.jpg 1080w" sizes="(max-width: 1024px) 100vw, 1024px" />
```
**Picture markup:**
```
<picture class="wp-picture-74" style="display: contents;">
 <source type="image/webp" srcset="http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-1024x683-jpg.webp 1024w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-300x200-jpg.webp 300w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-768x512-jpg.webp 768w" sizes="(max-width: 1024px) 100vw, 1024px">
 <source type="image/jpeg" srcset="http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-1024x683.jpg 1024w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-300x200.jpg 300w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-768x512.jpg 768w" sizes="(max-width: 1024px) 100vw, 1024px">
 <img width="1024" height="683" src="http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-1024x683.jpg" class="wp-image-74" alt="Green Leaves" decoding="async" loading="lazy" srcset="http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-1024x683.jpg 1024w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-300x200.jpg 300w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1087-768x512.jpg 768w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1087.jpg 1080w" sizes="(max-width: 1024px) 100vw, 1024px" />
</picture>
```

plugins/webp-uploads/picture-element.php
```diff
@@ -22,6 +22,15 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
...
 		return $image;
 	}
+
+	// Bail early if picture element requirement is not met.
+	if ( ! webp_uploads_is_picture_element_enabled() ) {
+		return $image;
+	}
+
+	// Remove IMG replacement for picture element.
+	remove_filter( 'the_content', 'webp_uploads_update_image_references', 13 );
```
westonruter: This doesn't seem right. The filter is getting re-removed every time the `webp_uploads_wrap_image_in_picture()` function calls (which is every time `the_content` filters apply). You could modify `webp_uploads_update_image_references()` to also short-circuit if `webp_uploads_is_picture_element_enabled()` returns true. However, it would be more efficient if you only added the filters when we know they are needed. So for example, in `hooks.php`:

```php
add_action( 'init', static function () {
    if ( webp_uploads_is_picture_element_enabled() ) {
        add_filter( 'wp_content_img_tag', 'webp_uploads_wrap_image_in_picture', 10, 3 );
    } else {
        add_filter( 'the_content', 'webp_uploads_update_image_references', 13 ); // Run after wp_filter_content_tags.
    }
} );
```

(I'm using a closure as an example only. A named function would be used in commit.)

The above would eliminate this line:

https://github.com/WordPress/performance/blob/e34711dbf286a16523b3c3d35462e0e79ea4c6f9/plugins/webp-uploads/hooks.php#L579

And replace lines (modified above in this PR):

https://github.com/WordPress/performance/blob/e34711dbf286a16523b3c3d35462e0e79ea4c6f9/plugins/webp-uploads/hooks.php#L768-L770

Then there would be no need for this code here in your PR:

https://github.com/WordPress/performance/blob/37901ab13e8a8ed10c0d47647ca16a1b0da70229/plugins/webp-uploads/picture-element.php#L26-L32

westonruter (changes requested):

plugins/webp-uploads/picture-element.php
mukeshpanchal27: Thanks @westonruter If i add condition filter in `init` action the tests going failed and didn't return picture elements.

Init action:
```php
/**
 * Initializes custom functionality for handling image uploads and content filters.
 *
 * @since n.e.x.t
 */
function webp_uploads_init(): void {
	if ( webp_uploads_is_picture_element_enabled() ) {
		add_filter( 'wp_content_img_tag', 'webp_uploads_wrap_image_in_picture', 10, 3 );
	} else {
		add_filter( 'the_content', 'webp_uploads_update_image_references', 13 ); // Run after wp_filter_content_tags.
	}
}
add_action( 'init', 'webp_uploads_init' );
```

```
1) Test_WebP_Uploads_Picture_Element::test_maybe_wrap_images_in_picture_element with data set "jpeg and picture enabled" (true, true, '<picture class="wp-picture-{{...cture>')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'<picture class="wp-picture-4" style="display: contents;"><source type="image/webp" srcset="http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-1024x683-jpg.webp 1024w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-300x200-jpg.webp 300w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-768x512-jpg.webp 768w" sizes="(max-width: 1024px) 100vw, 1024px"><source type="image/jpeg" srcset="http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-1024x683.jpg 1024w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-300x200.jpg 300w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-768x512.jpg 768w" sizes="(max-width: 1024px) 100vw, 1024px"><img width="1024" height="683" src="http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-1024x683.jpg" class="wp-image-4" alt="Green Leaves" decoding="async" loading="lazy" srcset="http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-1024x683.jpg 1024w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-300x200.jpg 300w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-768x512.jpg 768w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1214.jpg 1080w" sizes="(max-width: 1024px) 100vw, 1024px" /></picture>'
+'<img width="1024" height="683" src="http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-1024x683.jpg" class="wp-image-4" alt="Green Leaves" decoding="async" loading="lazy" srcset="http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-1024x683.jpg 1024w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-300x200.jpg 300w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1214-768x512.jpg 768w, http://localhost:8889/wp-content/uploads/2024/07/leaves-1214.jpg 1080w" sizes="(max-width: 1024px) 100vw, 1024px" />'

/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-picture-element.php:92
```

Short-circuit in `webp_uploads_update_image_references` filter works fine.

westonruter: Yes, in your tests you'd need to remove the filters that `webp_uploads_init()` adds, and then you'd need to set the options to opt-in to the picture element, and then you'd re-call `webp_uploads_init()`. This logic would be put into the `opt_in_to_picture_element()` helper method.

mukeshpanchal27: Addressed in https://github.com/WordPress/performance/pull/1408/commits/f93d0693fc257bd3b16393909e7b27d47d122644

mukeshpanchal27:

plugins/webp-uploads/hooks.php
```diff
@@ -527,6 +527,11 @@ function webp_uploads_update_image_references( $content ): string {
 		$content = '';
 	}
 
+	// Bail early if picture element requirement is not met.
+	if ( webp_uploads_is_picture_element_enabled() ) {
+		return $content;
+	}
+
```
westonruter: This should no longer be necessary.

```suggestion
```

plugins/webp-uploads/tests/data/class-testcase.php
```diff
@@ -137,6 +137,6 @@ public function set_image_output_type( string $format ): void {
 	 * Opt into picture element output.
 	 */
 	public function opt_in_to_picture_element(): void {
-		add_filter( 'wp_content_img_tag', 'webp_uploads_wrap_image_in_picture', 10, 3 );
+		webp_uploads_init();
```
westonruter, line 146: Shouldn't this also make sure that it removes filters that `webp_uploads_init()` adds?

```suggestion
		remove_filter( 'wp_content_img_tag', 'webp_uploads_wrap_image_in_picture', 10, 3 );
		remove_filter( 'the_content', 'webp_uploads_update_image_references', 13 ); // Run after wp_filter_content_tags.
		webp_uploads_init();
```

mukeshpanchal27: Thanks. When it runs at first, it will add the `the_content` filter, but this is a nice addition to avoid any issues in tests.

PHPStan reports `Function remove_filter invoked with 4 parameters, 2-3 required.` so I have removed the `$accepted_args` from the `wp_content_img_tag` filter as it is not needed for `remove_filter`.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -11,51 +11,39 @@
...
 
-		if ( $jpeg_and_webp ) {
-			$this->opt_in_to_jpeg_and_webp();
+		if ( $fallback_jpeg ) {
+			update_option( 'perflab_generate_webp_and_jpeg', '1' );
 		}
 
 		if ( $picture_element ) {
-			$this->opt_in_to_picture_element();
+			update_option( 'webp_uploads_use_picture_element', '1' );
```
westonruter: Shouldn't this function call be inside of `$this->opt_in_to_picture_element()`?

mukeshpanchal27: Thanks. Updated tests in https://github.com/WordPress/performance/pull/1408/commits/8e36b2c1d4c47d92c7f151743505fdd10310c336

westonruter (approved):
LGTM but I'll defer to @adamsilverstein for final approval

westonruter:
Please update the PR title and description to reflect the bug that this fixes.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -11,40 +11,26 @@
...
-	 * Attachment ID.
-	 *
-	 * @var int
-	 */
-	public static $image_id;
-
-	/**
-	 * Setup shared fixtures.
-	 */
-	public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ): void {
```
joemcgill, line 24: Why are you removing the shared attachment fixture and instead creating new attachments in individual tests? It would be more efficient to use the same upload for each test if possible.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -152,12 +134,19 @@ public function test_img_sizes_is_equal_to_picture_source_sizes_for_picture_elem
 			$this->markTestSkipped( "Mime type $mime_type is not supported." );
 		}
 
+		// Fallback to JPEG IMG.
+		update_option( 'perflab_generate_webp_and_jpeg', '1' );
+
+		// Create an image.
+		$attachment_id = self::factory()->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/data/images/leaves.jpg' );
```
joemcgill: Seems like we could do this in `wpSetUpBeforeClass` instead to create a shareable attachment that has both WebP and JPEG sizes generated that can be used across multiple tests.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -168,18 +157,23 @@ public function test_img_sizes_is_equal_to_picture_source_sizes_for_picture_elem
...
+		// Apply picture element support.
+		$this->opt_in_to_picture_element();
+
+		$picture_markup    = apply_filters( 'the_content', $image );
 		$picture_processor = new WP_HTML_Tag_Processor( $picture_markup );
 
 		$picture_processor->next_tag( array( 'tag_name' => 'IMG' ) );
+		$picture_img_src = $picture_processor->get_attribute( 'src' );
+		$this->assertStringEndsWith( '.jpg', $picture_img_src, 'Should return .jpg img src.' );
+		$this->assertSame( $img_src, $picture_img_src, 'Make sure IMG src is same.' );
```
joemcgill: Originally, this test was written to confirm that when `picture` is enabled, that the `source` elements will have the same `sizes` attribute as the `img` element that it is wrapping, which is why the test method is named `test_img_sizes_is_equal_to_picture_source_sizes_for_picture_element()`.

Adding these assertions here expands the scope of this test so it's now testing for multiple different things. Ideally, it would be best to have a specific test that relates to the bug being fixed, https://github.com/WordPress/performance/issues/1316, rather than mixing the purpose of a test we already have. Otherwise, we should at least update the test name to be more descriptive of what's being tested.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -168,18 +157,23 @@ public function test_img_sizes_is_equal_to_picture_source_sizes_for_picture_elem
...
-		$this->opt_in_to_picture_element();
-
-		$picture_markup = apply_filters( 'the_content', $image );
-
 		$img_processor = new WP_HTML_Tag_Processor( $img_markup );
 		$this->assertTrue( $img_processor->next_tag( array( 'tag_name' => 'IMG' ) ), 'There should be an IMG tag.' );
 		$img_sizes = $img_processor->get_attribute( 'sizes' );
+		$img_src   = $img_processor->get_attribute( 'src' );
 
+		$this->assertStringEndsWith( '.jpg', $img_src, 'Should return .jpg img src.' );
```
joemcgill: Can you explain why the image `src` would be expected to be a JPEG here instead of WebP and why that is relevant to the this test?

joemcgill (commented):
I left a few comments, questions, and suggestions. Nothing blocking.

plugins/webp-uploads/tests/test-picture-element.php
westonruter: Humm. Right. This should be WebP here. The test ultimately would be wanting to ensure that the `PICTURE > IMG` is a JPEG which it is doing below, but at this point it should be WebP since the picture element isn't enabled.

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: @joemcgill Initially i implemented it but if we use the shared attachment then it will return WEBP image instead of JPEG.

```
<img width="1024" height="683" src="http://localhost:8889/wp-content/uploads/2024/08/leaves-417-1024x683.webp" class="wp-image-4" alt="Green Leaves" decoding="async" loading="lazy" srcset="http://localhost:8889/wp-content/uploads/2024/08/leaves-417-1024x683.webp 1024w, http://localhost:8889/wp-content/uploads/2024/08/leaves-417-300x200.webp 300w, http://localhost:8889/wp-content/uploads/2024/08/leaves-417-768x512.webp 768w, http://localhost:8889/wp-content/uploads/2024/08/leaves-417-jpg.webp 1080w" sizes="(max-width: 1024px) 100vw, 1024px" />
```

Per https://web.dev/learn/design/picture-element we should fallback to JPEG so i use new attachment for each image.

```
<picture>
  <source srcset="image.avif" type="image/avif">
  <source srcset="image.webp" type="image/webp">
  <img src="image.jpg" alt="A description of the image." 
    width="300" height="200" loading="lazy" decoding="async">
</picture>
```

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: Gotcha. Sorry for mixup. I make sure to check the sizes for the IMG is same. I have removed the src check from the test in https://github.com/WordPress/performance/pull/1408/commits/8b2bd7a437d50a777b9809c3581b2b3cef7dea85

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: Thanks for raising Joe. I understood that i mixup the test and it should be Webp instead of JPEG. apologize for that.

I missed because when test uploaded image in test should return webp instead of jpeg image in our test it return jpeg image 🤔 

```
// Create an image.
$attachment_id = self::factory()->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/data/images/leaves.jpg' );

// Create some content with the image.
$image = wp_get_attachment_image(
	$attachment_id,
	'large',
	false,
	array(
		'class' => "wp-image-{$attachment_id}",
		'alt'   => 'Green Leaves',
	)
);
```

mukeshpanchal27:
@westonruter @joemcgill @adamsilverstein I think we should update Picture element implementation, why?

Trunk:
```html
<picture class="wp-picture-38" style="display: contents;">
  <source type="image/webp" srcset="http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683-jpg.webp 1024w, http://localhost:8888/wp-content/uploads/2024/08/leaves-300x200-jpg.webp 300w, http://localhost:8888/wp-content/uploads/2024/08/leaves-768x512-jpg.webp 768w" sizes="(max-width: 620px) 100vw, 620px">
  <source type="image/jpeg" srcset="http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683.jpg 1024w, http://localhost:8888/wp-content/uploads/2024/08/leaves-300x200.jpg 300w, http://localhost:8888/wp-content/uploads/2024/08/leaves-768x512.jpg 768w" sizes="(max-width: 620px) 100vw, 620px">
  <img decoding="async" width="1024" height="683" src="http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683-jpg.webp" alt="" class="wp-image-38" srcset="http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683-jpg.webp 1024w, http://localhost:8888/wp-content/uploads/2024/08/leaves-300x200-jpg.webp 300w, http://localhost:8888/wp-content/uploads/2024/08/leaves-768x512-jpg.webp 768w, http://localhost:8888/wp-content/uploads/2024/08/leaves-jpg.webp 1080w" sizes="(max-width: 620px) 100vw, 620px">
</picture>
```

Current PR:
```html
<picture class="wp-picture-38" style="display: contents;">
  <source type="image/webp" srcset="http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683-jpg.webp 1024w, http://localhost:8888/wp-content/uploads/2024/08/leaves-300x200-jpg.webp 300w, http://localhost:8888/wp-content/uploads/2024/08/leaves-768x512-jpg.webp 768w" sizes="(max-width: 620px) 100vw, 620px">
  <source type="image/jpeg" srcset="http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683.jpg 1024w, http://localhost:8888/wp-content/uploads/2024/08/leaves-300x200.jpg 300w, http://localhost:8888/wp-content/uploads/2024/08/leaves-768x512.jpg 768w" sizes="(max-width: 620px) 100vw, 620px">
  <img decoding="async" width="1024" height="683" src="http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683.jpg" alt="" class="wp-image-38" srcset="http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683.jpg 1024w, http://localhost:8888/wp-content/uploads/2024/08/leaves-300x200.jpg 300w, http://localhost:8888/wp-content/uploads/2024/08/leaves-768x512.jpg 768w, http://localhost:8888/wp-content/uploads/2024/08/leaves.jpg 1080w" sizes="(max-width: 620px) 100vw, 620px">
</picture>
```

Per https://web.dev/learn/design/picture-element & https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture:
```html
<picture>
  <source srcset="image.avif" type="image/avif">
  <source srcset="image.webp" type="image/webp">
  <img src="image.jpg" alt="A description of the image." width="300" height="200" loading="lazy" decoding="async">
</picture>
```

- Trunk version return webp src in IMG tag that is wrong.
- The current PR as it return JPEG src. Per the above documents we should remove the fallback source. In our case we should remove `source > image/jpeg`.

What do you think?

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: Joe, It will return only WEBP image.

```
<picture class="wp-picture-4" style="display: contents;">
  <source type="image/webp" srcset="http://localhost:8889/wp-content/uploads/2024/08/leaves-637-1024x683.webp 1024w, http://localhost:8889/wp-content/uploads/2024/08/leaves-637-300x200.webp 300w, http://localhost:8889/wp-content/uploads/2024/08/leaves-637-768x512.webp 768w" sizes="(max-width: 1024px) 100vw, 1024px">
  <img width="1024" height="683" src="http://localhost:8889/wp-content/uploads/2024/08/leaves-637-1024x683.webp" class="wp-image-4" alt="Green Leaves" decoding="async" loading="lazy" srcset="http://localhost:8889/wp-content/uploads/2024/08/leaves-637-1024x683.webp 1024w, http://localhost:8889/wp-content/uploads/2024/08/leaves-637-300x200.webp 300w, http://localhost:8889/wp-content/uploads/2024/08/leaves-637-768x512.webp 768w, http://localhost:8889/wp-content/uploads/2024/08/leaves-637-jpg.webp 1080w" sizes="(max-width: 1024px) 100vw, 1024px" />
</picture>
```

plugins/webp-uploads/picture-element.php
```diff
@@ -22,6 +22,7 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	if ( 'the_content' !== $context ) {
 		return $image;
 	}
+
```
adamsilverstein: unrelated change

adamsilverstein (approved):
Looks good, nice fix!

plugins/webp-uploads/tests/test-picture-element.php
joemcgill: I see. I think for this test class most of the tests are based on uploading a JPEG, and generating a WebP image with JPEG fallbacks. Instead of removing the fixture, We should probably set the `perflab_generate_webp_and_jpeg` option first before creating the image and then all tests that need to test a WebP with a JPEG fallback can use the same attachment.

```
// Fallback to JPEG IMG.
update_option( 'perflab_generate_webp_and_jpeg', '1' );
```

What do you think?

plugins/webp-uploads/tests/test-picture-element.php
joemcgill: Got it. We need to set the fallback option before creating the shared attachment. See https://github.com/WordPress/performance/pull/1408#discussion_r1704383543.

joemcgill:
@mukeshpanchal27:

> I think we should update Picture element implementation, why?

Thanks for pointing this out. I agree with you that when we're using the `picture` format, we should only use `source` elements for the modern image formats (i.e., webp, avif) and always use JPEG on the fallback `img` element (like the third example you showed). The `<source type="image/jpeg" ... />` is unnecessary as long as the `img` uses JPEG.

joemcgill (changes requested):
I'm marking this as needs changes until the issue @mukeshpanchal27 identified has been addressed: https://github.com/WordPress/performance/pull/1408#issuecomment-2268302619

adamsilverstein:
> . The <source type="image/jpeg" ... /> is unnecessary as long as the img uses JPEG.

Would we still want the "image/jpeg"  srcset to serve the different JPEG images for different breakpoint. eg a mobile device that doesn't support AVIF could still use the srcset to load the mobile sized JPEG? I assume it works that way although I haven't tested, the docs didn't seem clear on this to me.

If we don't use the JPEG fallback images (sub-sizes via srcset), we shouldn't bother creating them.

westonruter:
> Current PR:
> 
> ```
> <picture class="wp-picture-38" style="display: contents;">
>   <source type="image/webp" srcset="http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683-jpg.webp 1024w, http://localhost:8888/wp-content/uploads/2024/08/leaves-300x200-jpg.webp 300w, http://localhost:8888/wp-content/uploads/2024/08/leaves-768x512-jpg.webp 768w" sizes="(max-width: 620px) 100vw, 620px">
>   <source type="image/jpeg" srcset="http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683.jpg 1024w, http://localhost:8888/wp-content/uploads/2024/08/leaves-300x200.jpg 300w, http://localhost:8888/wp-content/uploads/2024/08/leaves-768x512.jpg 768w" sizes="(max-width: 620px) 100vw, 620px">
>   
> </picture>
> ```

Yeah, the `source` for JPEG should not be needed as it is redundant. This should rather be (with line breaks added for readability):

```html
<picture class="wp-picture-38" style="display: contents;">
  <source 
    type="image/webp" 
    srcset="
      http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683-jpg.webp 1024w, 
      http://localhost:8888/wp-content/uploads/2024/08/leaves-300x200-jpg.webp 300w, 
      http://localhost:8888/wp-content/uploads/2024/08/leaves-768x512-jpg.webp 768w
    " 
    sizes="(max-width: 620px) 100vw, 620px">
  <img 
    decoding="async" 
    width="1024" 
    height="683" 
    src="http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683.jpg" 
    alt="" 
    class="wp-image-38" 
    srcset="
      http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683.jpg 1024w, 
      http://localhost:8888/wp-content/uploads/2024/08/leaves-300x200.jpg 300w, 
      http://localhost:8888/wp-content/uploads/2024/08/leaves-768x512.jpg 768w, 
      http://localhost:8888/wp-content/uploads/2024/08/leaves.jpg 1080w
    " 
    sizes="(max-width: 620px) 100vw, 620px">
</picture>
```

Note also (as is being tracked in #1417) that the `IMG` includes the original `leaves.jpg` at 1080w in its `srcset` which is absent from the `srcset` attribute on the `SOURCE` tag. An equivalent 1080w AVIF image is also missing from the initial `SOURCE` tag.

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: Thanks. It works with the option. Updated in https://github.com/WordPress/performance/pull/1408/commits/0e07201f88482e95f979cfe7eb5884245ecb69cc

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: Resolved now with https://github.com/WordPress/performance/pull/1408/commits/0e07201f88482e95f979cfe7eb5884245ecb69cc

mukeshpanchal27:
@joemcgill @westonruter @adamsilverstein I have addressed all the feedback and updated the PR description. Could you please take a look? Thanks!

adamsilverstein:
> Yeah, the source for JPEG should not be needed as it is redundant. This should rather be (with line breaks added for readability):

Oh, I see now, thanks for breaking out these lines, I was missing the srcset in the img tag. Makes sense!

adamsilverstein (approved):

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -115,32 +111,68 @@ public function test_maybe_wrap_images_in_picture_element( bool $jpeg_and_webp,
...
+		$img_srcset = $img_processor->get_attribute( 'srcset' );
+
+		// Apply picture element support.
+		$this->opt_in_to_picture_element();
+
+		$picture_markup    = apply_filters( 'the_content', $image );
+		$picture_processor = new WP_HTML_Tag_Processor( $picture_markup );
+
+		$picture_processor->next_tag( array( 'tag_name' => 'IMG' ) );
+		$this->assertSame( $img_src, $picture_processor->get_attribute( 'src' ), 'Make sure the IMG and Picture IMG have same image src.' );
```
westonruter: Is this right? The `$img_src` should be the AVIF/WebP image. The `IMG` inside the `PICTURE` should rather be the JPEG.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -115,32 +111,68 @@ public function test_maybe_wrap_images_in_picture_element( bool $jpeg_and_webp,
...
+
+		$picture_processor = new WP_HTML_Tag_Processor( $picture_markup );
+		while ( $picture_processor->next_tag( array( 'tag_name' => 'source' ) ) ) {
+			$this->assertNotSame( 'image/jpeg', $picture_processor->get_attribute( 'type' ), 'Make sure the Picture source should not return jpeg as source.' );
```
westonruter: Wouldn't this be better to test if it is the same as something we expect? And shouldn't it make sure that the WebP image is contained in the `srcset`?

```suggestion
			$this->assertSame( $mime_type, $picture_processor->get_attribute( 'type' ), 'Make sure the Picture source should not return JPEG as source.' );
			$this->assertStringContainsString( $img_src, $picture_processor->get_attribute( 'srcset' ) );
```

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -115,32 +111,68 @@ public function test_maybe_wrap_images_in_picture_element( bool $jpeg_and_webp,
...
+				'alt'   => 'Green Leaves',
+			)
+		);
+
+		$img_markup = apply_filters( 'the_content', $image );
+
+		$img_processor = new WP_HTML_Tag_Processor( $img_markup );
+		$this->assertTrue( $img_processor->next_tag( array( 'tag_name' => 'IMG' ) ), 'There should be an IMG tag.' );
+		$img_src    = $img_processor->get_attribute( 'src' );
+		$img_srcset = $img_processor->get_attribute( 'srcset' );
```
westonruter, line 176: Let's make sure these are actually WebP images here.

```suggestion
		$img_srcset = $img_processor->get_attribute( 'srcset' );
		$this->assertStringEndsWith( '.webp', $img_src );
		$this->assertStringContainsString( '.webp', $img_srcset );
```

plugins/webp-uploads/tests/test-picture-element.php
westonruter: This should add assertions that the `src` ends with `.jpg` and the `srcset` contains `.jpg`.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -115,32 +111,68 @@ public function test_maybe_wrap_images_in_picture_element( bool $jpeg_and_webp,
...
+		$picture_markup    = apply_filters( 'the_content', $image );
+		$picture_processor = new WP_HTML_Tag_Processor( $picture_markup );
+
+		$picture_processor->next_tag( array( 'tag_name' => 'IMG' ) );
+		$this->assertSame( $img_src, $picture_processor->get_attribute( 'src' ), 'Make sure the IMG and Picture IMG have same image src.' );
+
+		$picture_processor = new WP_HTML_Tag_Processor( $picture_markup );
+		while ( $picture_processor->next_tag( array( 'tag_name' => 'source' ) ) ) {
+			$this->assertNotSame( 'image/jpeg', $picture_processor->get_attribute( 'type' ), 'Make sure the Picture source should not return jpeg as source.' );
+			$this->assertNotSame( $img_srcset, $picture_processor->get_attribute( 'srcset' ), 'Make sure the IMG and Picture source should not same srcset attributes.' );
```
westonruter: This could check to see if the `srcset` contains `.webp` as expected.

```suggestion
			$this->assertStringContainsString( '.webp', $picture_processor->get_attribute( 'srcset' ) );
```

westonruter (changes requested):
Unless I'm missing something (like the issues I just commented on being addressed elsewhere) this doesn't seem quite there.

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: @westonruter The initial image is `JPEG` image so the `$img_src` return `.jpg` IMG src.

mukeshpanchal27: Why it return the `JPEG` images in unit tests?

`add_filter( 'the_content', 'webp_uploads_update_image_references', 13 );` runs that replace `JPEG` to it's additional mime.

```php
// Bail early if request is not for the frontend.
if ( ! webp_uploads_in_frontend_body() ) {
	return $content;
}
```

This condition return early for the unit tests so it will return `JPEG` instead of `WEBP` images in unit test.

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: Check my reply https://github.com/WordPress/performance/pull/1408#discussion_r1706364060

plugins/webp-uploads/tests/test-picture-element.php
westonruter: Ok, so then this should be mocked so that `webp_uploads_in_frontend_body()` returns true.

westonruter: Mocked isn't the right word. I mean, the environment should be set up so that `webp_uploads_in_frontend_body()` returns true. It seems this can be forced with code like:

```php
$GLOBALS['wp_actions']['template_redirect'] = 1;
```

mukeshpanchal27: I used `mock_frontend_body_hooks()` in https://github.com/WordPress/performance/pull/1408/commits/e0e30eff3faf1b0d7835f8b3d7be3babf4b8d3b2 it was introduce in plugin while before.

It will return the WebP now. 🎉

plugins/webp-uploads/tests/data/class-testcase.php
```diff
@@ -137,6 +137,20 @@ public function set_image_output_type( string $format ): void {
 	 * Opt into picture element output.
 	 */
 	public function opt_in_to_picture_element(): void {
-		add_filter( 'wp_content_img_tag', 'webp_uploads_wrap_image_in_picture', 10, 3 );
+		remove_filter( 'wp_content_img_tag', 'webp_uploads_wrap_image_in_picture', 10 );
+		remove_filter( 'the_content', 'webp_uploads_update_image_references', 13 ); // Run after wp_filter_content_tags.
+
+		// Apply picture element support.
+		update_option( 'webp_uploads_use_picture_element', '1' );
```
westonruter, line 144: Is enabling the `perflab_generate_webp_and_jpeg` option not needed?

```suggestion
		update_option( 'perflab_generate_webp_and_jpeg', '1' );
		update_option( 'webp_uploads_use_picture_element', '1' );
```

The `webp_uploads_is_picture_element_enabled()` function only returns true if both of these are enabled:

https://github.com/WordPress/performance/blob/e7313851f39749886a4ae0bacbcd620a6effce3d/plugins/webp-uploads/helper.php#L395-L408

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -234,19 +274,20 @@ public function test_disable_responsive_image_with_picture_element(): void {
 
 		$img_markup = apply_filters( 'the_content', $image );
```
westonruter, line 280: At this point, the IMG in `$img_markup` is a JPEG. Why isn't it a WebP? I think this method is missing `$this->mock_frontend_body_hooks();`.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -168,20 +212,20 @@ public function test_img_sizes_is_equal_to_picture_source_sizes_for_picture_elem
...
```
westonruter, line 218: This is also incorrectly returning JPEG. This method is also missing `$this->mock_frontend_body_hooks();`.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -115,43 +125,77 @@ public function test_maybe_wrap_images_in_picture_element( bool $jpeg_and_webp,
...
+			array(
+				'class' => 'wp-image-' . self::$image_id,
+				'alt'   => 'Green Leaves',
+			)
+		);
+
+		// Run critical hooks to satisfy webp_uploads_in_frontend_body() conditions.
+		$this->mock_frontend_body_hooks();
+
+		$img_markup = apply_filters( 'the_content', $image );
```
westonruter, line 170: This is also incorrectly returning JPEG. This method is also missing `$this->mock_frontend_body_hooks();`.

westonruter (changes requested):
It looks like `$this->mock_frontend_body_hooks()` needs to be added to the remaining test cases as well. Alternatively, it could be added to `set_up` instead.

plugins/webp-uploads/tests/data/class-testcase.php
mukeshpanchal27: @westonruter https://github.com/WordPress/performance/blob/fix/1316-check-img-src/plugins/webp-uploads/tests/test-picture-element.php#L43-L44 we set fallback for all the tests so it generate JPEG also for all tests. Also in `test_maybe_wrap_images_in_picture_element` test we needs to update that option to no JPEG fallback so update option to return false.

mukeshpanchal27:
@westonruter @adamsilverstein PR updated and ready for review.

plugins/webp-uploads/tests/data/class-testcase.php
westonruter: OK, got it. Still, it seems it should be added here for completeness.

Also, I believe that `wpSetUpBeforeClass` isn't the right place to set the option. It should be set in `set_up`. The reason for this is that `wpTearDownAfterClass` does not reset the database. So setting an option needs to be done in `set_up` instead, otherwise the option will leak outside the test. 

~Secondly, I believe that `wpSetUpBeforeClass` should be calling `parent::wpSetUpBeforeClass()` as otherwise there is base class functionality which won't be getting called.~

Thirdly, the current implementation of `wpSetUpBeforeClass` which is generating an image should have a corresponding `wpTearDownAfterClass` method ~(which also calls `parent::wpTearDownAfterClass()`)~ while also deleting that uploaded image. Otherwise the image will never get cleaned up, resulting in more and more wasted space with each test run.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -11,71 +11,89 @@
...
 	 */
 	public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ): void {
+		// Fallback to JPEG IMG.
+		update_option( 'perflab_generate_webp_and_jpeg', '1' );
+
 		self::$image_id = $factory->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/data/images/leaves.jpg' );
 	}
 
+	public static function wpTearDownAfterClass(): void {
+		wp_delete_attachment( self::$image_id, true );
```
westonruter, line 53: The base class method should also always be called unless it's intended to completely override it.

```suggestion
		parent::wpTearDownAfterClass();
		wp_delete_attachment( self::$image_id, true );
```

The same should be done for `Test_Perflab_Server_Timing::wpTearDownAfterClass()`

westonruter: Hummm. Actually, this doesn't seem to be valid! I'm getting:

> Call to undefined method WP_UnitTestCase::wpTearDownAfterClass()

westonruter: Oh, I see. I was looking at `WP_Canonical_UnitTestCase` in core which [defines](https://github.com/WordPress/wordpress-develop/blob/4026237dadd2f4c1f11ba4a89605e8f394c11f38/tests/phpunit/includes/testcase-canonical.php#L22-L23) this method. With a quick glance I thought it was the base class but it isn't. The `tear_down_after_class()` static method in the base class has a [`method_exists()` check](https://github.com/WordPress/wordpress-develop/blob/4026237dadd2f4c1f11ba4a89605e8f394c11f38/tests/phpunit/includes/abstract-testcase.php#L84-L89) to see if this method exists.

So yeah, since the parent class lacks the method then no need for `parent::wpTearDownAfterClass();`!

plugins/webp-uploads/tests/data/class-testcase.php
westonruter: Correction: the base class doesn't actually define `wpSetUpBeforeClass()` or  `wpTearDownAfterClass()` so no need to call them. See https://github.com/WordPress/performance/pull/1408#discussion_r1708658029

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: It will return error:

```
1) Test_WebP_Uploads_Picture_Element::tearDownAfterClass
Exception in Test_WebP_Uploads_Picture_Element::tearDownAfterClass
Call to undefined method WebP_Uploads\Tests\TestCase::wpTearDownAfterClass()

/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-picture-element.php:53
/wordpress-phpunit/includes/abstract-testcase.php:88
/var/www/html/wp-content/plugins/performance/vendor/yoast/phpunit-polyfills/src/TestCases/TestCasePHPUnitGte8.php:107
```

plugins/webp-uploads/tests/data/class-testcase.php
mukeshpanchal27: > Also, I believe that `wpSetUpBeforeClass` isn't the right place to set the option. It should be set in `set_up`. The reason for this is that `wpTearDownAfterClass` does not reset the database. So setting an option needs to be done in `set_up` instead, otherwise the option will leak outside the test.

When i move the update option in `set_up` the JPEG fallback image is not returning. 🤔 

> Thirdly, the current implementation of `wpSetUpBeforeClass` which is generating an image should have a corresponding `wpTearDownAfterClass` method ~(which also calls `parent::wpTearDownAfterClass()`)~ while also deleting that uploaded image. Otherwise the image will never get cleaned up, resulting in more and more wasted space with each test run.

Done in https://github.com/WordPress/performance/pull/1408/commits/c693e96dd73bc75cdb68ece5cea711e9d99804c5

westonruter: > When i move the update option in `set_up` the JPEG fallback image is not returning. 🤔

Oh, right, because `set_up` happens after `wpSetUpBeforeClass`. So if you intend to set the option in `wpSetUpBeforeClass()` then you should be sure to delete it in `wpTearDownAfterClass()`.

plugins/webp-uploads/tests/test-picture-element.php
westonruter, line 53:
```suggestion
		wp_delete_attachment( self::$image_id, true );
		delete_option( 'perflab_generate_webp_and_jpeg' );
```

plugins/webp-uploads/tests/data/class-testcase.php
westonruter: See https://github.com/WordPress/performance/pull/1408/files#r1708690705

westonruter (approved):

mukeshpanchal27:
@adamsilverstein @joemcgill Could you please review this when you have a moment? Thanks!

joemcgill (approved):
This is looking good.

plugins/webp-uploads/picture-element.php
```diff
@@ -44,12 +44,13 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
...
 	 * Filter the image mime types that can be used for the <picture> element.
 	 *
-	 * Default is: ['image/avif', 'image/webp', 'image/jpeg']. Returning an empty array will skip using the `picture` element.
+	 * Default is: ['image/avif', 'image/webp']. Returning an empty array will skip using the `picture` element.
 	 *
 	 * The mime types will output in the picture element in the order they are provided.
 	 * The original image will be used as the fallback image for browsers that don't support the picture element.
 	 *
 	 * @since 2.0.0
+	 *
```
adamsilverstein, line 54: maybe annotate the default change?
```suggestion
	 * @since n.e.x.t. The default value was updated, removing 'image/jpeg'.
	 *
```

adamsilverstein (approved):
Fantastic!

adamsilverstein:
Worked well in my testing, thanks for working on this @mukeshpanchal27 🎉

plugins/webp-uploads/picture-element.php
mukeshpanchal27: Added in https://github.com/WordPress/performance/pull/1408/commits/9ddf45a72d88ae466c66ac23e2e84c7d60230b91. FYI we don't need last period in `@since n.e.x.t`

PR #1411, opened 2024-07-29
Implement full support for intersectionRect/boundingClientRect, fix viewportRect typing, and harden JSON schema

westonruter:
* Add remaining properties of `intersectionRect`/`clientBoundingRect` to the JSON Schema.
* Fix type of viewport rect which has integer width/height versus DOMRect which has floats. I caught onto this via Query Monitor reporting:

```
Implicit conversion from float 739.9976196289062 to int loses precision in /var/www/html/wp-content/plugins/optimization-detective/class-od-url-metrics-group-collection.php on line 486
```

* Ensure all properties in the JSON Schema are `required` and that `additionalProperties` are not allowed.
* Refactor tests to reuse a `get_sample_url_metric()` method in the shared helper trait rather than duplicate the logic across various test classes. This is a follow-up to https://github.com/WordPress/performance/pull/1405 which refactored the tests for Optimization Detective.
* Clean up `@throws` tag usages, removing where obsolete and suppressing with `@noinspection` where irrelevant.
* Fix passing the actual function name to the `OD_HTML_Tag_Processor::warn()` method.

> [!TIP]
> Suppress whitespace when reviewing the changes in this pull request!

### DIFF
--- a/plugins/auto-sizes/tests/test-optimization-detective.php
+++ b/plugins/auto-sizes/tests/test-optimization-detective.php
@@ -102,7 +102,7 @@ class Test_Auto_Sizes_Optimization_Detective extends WP_UnitTestCase {
 	 * @covers ::auto_sizes_visit_tag
 	 *
 	 * @dataProvider data_provider_test_od_optimize_template_output_buffer
-	 * @phpstan-param array<string, mixed> $element_metrics
+	 * @phpstan-param array{ xpath: string, isLCP: bool, intersectionRatio: int } $element_metrics
 	 */
 	public function test_od_optimize_template_output_buffer( array $element_metrics, string $buffer, string $expected ): void {
 		$this->populate_url_metrics( array( $element_metrics ) );
--- a/plugins/auto-sizes/tests/test-optimization-detective.php
+++ b/plugins/auto-sizes/tests/test-optimization-detective.php
@@ -3,6 +3,8 @@
  * Tests for auto-sizes plugin's optimization-detective.php.
  *
  * @package auto-sizes
+ *
+ * @noinspection PhpUnhandledExceptionInspection
  */
 
 class Test_Auto_Sizes_Optimization_Detective extends WP_UnitTestCase {
@@ -100,7 +102,6 @@ class Test_Auto_Sizes_Optimization_Detective extends WP_UnitTestCase {
 	 * @covers ::auto_sizes_visit_tag
 	 *
 	 * @dataProvider data_provider_test_od_optimize_template_output_buffer
-	 * @throws Exception But it won't.
 	 * @phpstan-param array<string, mixed> $element_metrics
 	 */
 	public function test_od_optimize_template_output_buffer( array $element_metrics, string $buffer, string $expected ): void {
--- a/plugins/optimization-detective/class-od-url-metric.php
+++ b/plugins/optimization-detective/class-od-url-metric.php
@@ -90,46 +90,37 @@ final class OD_URL_Metric implements JsonSerializable {
 	 * @return array<string, mixed> Schema.
 	 */
 	public static function get_json_schema(): array {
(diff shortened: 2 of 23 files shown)


plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -77,39 +90,67 @@ private function validate_data( array $data ): void {
...
-					'type'    => 'number',
-					'minimum' => 0,
+					'type'     => 'number', // An 'unrestricted double'.
+					'required' => true,
+					'minimum'  => 0.0,      // The spec allows this to be negative in DOMRectReadOnly, but this doesn't make sense in the context of intersectionRect and boundingClientRect.
 				),
 				'height' => array(
-					'type'    => 'number',
-					'minimum' => 0,
+					'type'     => 'number', // An 'unrestricted double'.
```
adamsilverstein: why is this comment "an unrestricted double" and below L109 etc its "A double"? Are they different?

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -77,39 +90,67 @@ private function validate_data( array $data ): void {
...
+					'required' => true,
+					'minimum'  => 0.0,      // The spec allows this to be negative in DOMRectReadOnly, but this doesn't make sense in the context of intersectionRect and boundingClientRect.
+				),
+				'x'      => array(
+					'type'     => 'number', // A 'double'.
```
adamsilverstein: the `// A 'double'.` comments seems a bit repetitive, maybe add something at the beginning of the entire structure instead?

plugins/optimization-detective/tests/storage/test-rest-api.php
```diff
@@ -117,6 +118,33 @@ function ( $params ) {
...
 				),
+				'invalid_elements_additional_intersect_rect_property' => array(
+					'elements' => array(
+						array_merge(
+							$valid_element,
+							array(
+								'intersectionRect' => array(
+									'width'  => 640,
+									'height' => 480,
+									'wooHoo' => 'bad',
```
adamsilverstein, line 129: :)

adamsilverstein:
Overall looks good and all tests pass! I left one question.

plugins/optimization-detective/class-od-url-metric.php
westonruter: Yeah. Good idea. They're all doubles per https://developer.mozilla.org/en-US/docs/Web/API/DOMRectReadOnly/width

plugins/optimization-detective/class-od-url-metric.php
westonruter: Yeah, this is what it is referred to in the DOM spec: https://drafts.fxtf.org/geometry/#dom-domrectreadonly-width

This is probably too much information 😄

plugins/optimization-detective/class-od-url-metric.php
westonruter: Addressed in 7399131

plugins/optimization-detective/class-od-url-metric.php
westonruter: Addressed in 7399131

plugins/auto-sizes/tests/test-optimization-detective.php
```diff
@@ -3,6 +3,8 @@
  * Tests for auto-sizes plugin's optimization-detective.php.
  *
  * @package auto-sizes
+ *
+ * @noinspection PhpUnhandledExceptionInspection
```
swissspidy, line 7: That's a PHPStorm thing, right? Do we usually have IDE-specific annotations?

swissspidy (approved):

plugins/auto-sizes/tests/test-optimization-detective.php
westonruter: Yeah, elsewhere in the codebase I've been using them.

PR #1420, opened 2024-07-31
Try setting aspect-ratio on images lacking width and height

### LINKED ISSUE #10: Visual tool to scan a page and report missing width/height elements
This task was suggested by Aymen Loukil for the image focus, however it may fit better in the measurement focus.

westonruter:
## Summary

I started looking at this because I was checking [how well Image Prioritizer optimizes a Timber-based theme](https://core.trac.wordpress.org/ticket/59331#comment:5). It adds `fetchpriority=high` and lazy-loading correctly, but the Timber starter theme does not include `width` and `height` attributes on the featured image, and Image Prioritizer hasn't yet implemented this either. So I wanted to see what it would take.

This experiment tries to set an `aspect-ratio` on `img` tags that are missing `width` and `height` attributes. There is currently a big problem, however, in that it is setting the `aspect-ratio` based on the `clientBoundingRect` and not the intrinsic dimensions of the image. This means that once the `aspect-ratio` is set the next time that `detect.js` measures the image, it will use the `aspect-ratio` for subsequent measurements, meaning that the image dimensions will become locked in place. In order to get around this, I think `detect.js` may need to be extend to obtain the _intrinsic dimensions_ of elements that have them: `img`, `video`, `canvas`, `svg` (maybe), and `input[type=image]` (ewwww).

Another problem is that in order to set the `aspect-ratio`, the element has to also have `width:100%` added. This may not be what is intended!

Fixes #10

### DIFF
--- a/plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
@@ -73,6 +73,19 @@ final class Image_Prioritizer_Img_Tag_Visitor extends Image_Prioritizer_Tag_Visi
 			$processor->remove_attribute( 'fetchpriority' );
 		}
 
+		// Set the aspect-ratio of the image when it lacks a width and height.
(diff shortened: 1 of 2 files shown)


westonruter:
Closing since the `aspect-ratio` approach is not viable. Better to set the `width` and `height` attributes on the element to be the corresponding intrinsic dimensions.

PR #1429, opened 2024-08-02
Add prettier config

thelovekesh:
## Summary

We are already using prettier with ESLint in our current setup but we are missing the prettier config file which is used by the prettier extension in multiple IDEs/editors. This PR adds the missing prettier config.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,1 @@
+"@wordpress/prettier-config"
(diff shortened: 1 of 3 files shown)


swissspidy (approved):

.prettierrc
```diff
@@ -0,0 +1 @@
+"@wordpress/prettier-config"
```
mukeshpanchal27, line 1: Do we needs to add these file in [.gitignore](https://github.com/WordPress/performance/blob/trunk/.gitignore)?

swissspidy: No, `.prettierrc` must not be ignored. It must be around for everyone.

Issue #1417, opened 2024-07-30
Picture element: Full size image is missing in PICTURE > SOURCE srcset


mukeshpanchal27:
## Bug Description

As discussed https://github.com/WordPress/performance/pull/1408#discussion_r1695593652, the full size image is missing in picture element source srcset.

## Steps to reproduce

1. Activate the Modern Image Formats plugin and enable picture support from the Media Settings screen.
2. Add an image block
3. Preview the page in the front end, Check the PICTURE > SOURCE srcset.

IMG srcset:
```
http://localhost:8888/wp-content/uploads/2024/07/car-1024x576.jpeg 1024w,
http://localhost:8888/wp-content/uploads/2024/07/car-300x169.jpeg 300w,
http://localhost:8888/wp-content/uploads/2024/07/car-768x432.jpeg 768w,
http://localhost:8888/wp-content/uploads/2024/07/car.jpeg 1080w
```

Picture > source srcset:
```
http://localhost:8888/wp-content/uploads/2024/07/car-1024x576.jpeg 1024w,
http://localhost:8888/wp-content/uploads/2024/07/car-300x169.jpeg 300w,
http://localhost:8888/wp-content/uploads/2024/07/car-768x432.jpeg 768w
```

## Screenshots

<img width="1576" alt="Screenshot 2024-07-30 at 4 38 57 PM" src="https://github.com/user-attachments/assets/68cc541e-35c3-4cc1-9f2e-f6145672847d">

devansh016:
Hi @mukeshpanchal27,

In line 25, `wp_get_attachment_metadata` is not getting `sources` field for bigger sizes image ( greater than 1024 width).
Therefore, in line 34 `isset( $size['sources'] )` is false and `$mime_type_data` don't have sizes greater than 1024.

https://github.com/WordPress/performance/blob/08eba80fd8609add2df44d9960dc011868dc5d93/plugins/webp-uploads/picture-element.php#L25-L41

Sample Data for reference:
[mime_type_data](https://pastebin.com/ws5Am8AV)
[image_meta['sizes']](https://pastebin.com/wCphTD99).  1536x1536 , 2048x2048 don't have sources

mukeshpanchal27:
@devansh016 Thanks for checking these.

The `wp_get_attachment_image_srcset()` function uses `wp_calculate_image_srcset()` to get the `srcset` for the image.
- The core uses some hacky way to add the full-size image in `srcset`.
[...]

PR #1430, opened 2024-08-05
Remove picture element if sources is empty

mukeshpanchal27:
## Summary

Issue found at https://github.com/WordPress/performance/issues/1417#issuecomment-2265517786

> Image itself is 600 × 400 in jpg, in admin there are thumbnail (150x150px and full size ). No medium, no medium_large etc, because medium is 768x512px with center crop.

If someone use smaller size image then the register sizes or thumbnail size image the picture element only return PICTURE > IMG tag no `source` tags.

**Before PR:**
```
<picture class="wp-picture-44" style="display: contents;">
  <img decoding="async" width="150" height="150" src="http://localhost:8888/wp-content/uploads/2024/08/car-150x150-jpeg.avif" alt="" class="wp-image-44">
</picture>
```

**After PR:** The PR removes the picture element if the `source` is null.
```
<img decoding="async" width="150" height="150" src="http://localhost:8888/wp-content/uploads/2024/08/car-150x150-jpeg.avif" alt="" class="wp-image-44">
```

## Steps to reproduce
- Activate the Modern Image Formats plugin and enable picture support from the Media Settings screen.
- Add an image block
- Select `Thumbnail` in block settings.
- Preview the page in the front end, Check the image.

## Todos:
- [ ] Add Unit tests

### DIFF
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -129,10 +129,10 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 		);
 	}
 
-	return sprintf(
+	return ( '' === $picture_sources ? $image : sprintf(
 		'<picture class="%s" style="display: contents;">%s%s</picture>',
 		esc_attr( 'wp-picture-' . $attachment_id ),
 		$picture_sources,
 		$image
-	);
+	) );
 }


westonruter:
Why isn't there a fallback JPEG for this generated AVIF image? The size of the image being uploaded shouldn't matter. If uploading a 10x10 JPEG image, this should still result in an AVIF/WebP image being generated in addition to the JPEG fallback (if that option is enabled). The `PICTURE` should still be used even if the `IMG` lacks a `srcset` because it should provide the JPEG fallback for the AVIF.

PR #1431, opened 2024-08-05
Update readme for Performance Lab to list out feature plugins

westonruter:
I think the Performance Lab plugin's description on WordPress.org should list out the feature plugins it includes to better promote discovery before even installing the plugin. The description can be further enhanced to provide a better rationale for installing the plugin.

This also updates the order of the plugins to sort by the name rather than by the slug on the Performance screen.


plugins/performance-lab/readme.txt
```diff
@@ -11,7 +11,19 @@ Performance plugin from the WordPress Performance Team, which is a collection of
...
+
+The feature plugins which are currently featured by this plugin are:
+
+* [Embed Optimizer](https://wordpress.org/plugins/embed-optimizer/)
+* [Enhanced Responsive Images](https://wordpress.org/plugins/auto-sizes/)
+* [Image Placeholders](https://wordpress.org/plugins/dominant-color-images/)
+* [Image Prioritizer](https://wordpress.org/plugins/image-prioritizer/)
+* [Modern Image Formats](https://wordpress.org/plugins/webp-uploads/)
+* [Performant Translations](https://wordpress.org/plugins/performant-translations/)
+* [Speculative Loading](https://wordpress.org/plugins/speculation-rules/)
```
felixarntz: I think we should match the order of these with how they're displayed in the admin, and possibly include the "(experimental)" bit as well where relevant. In other words:
* Sort by slug.
* Put experimental plugins after non-experimental ones.
 
See https://github.com/WordPress/performance/blob/trunk/plugins/performance-lab/load.php#L88.

We could revisit the sorting by slug, since it makes a bit more sense to sort by the user-facing name, but IMO that's a bit of a detail that could be separately explored. My main point here is to follow a consistent approach to sorting.

adamsilverstein: +1 to sorting by user facing name vs slug

swissspidy: FWIW I fixed the sorting in #1098. It sorts by user-facing name and puts experimental ones last (as we already do today). Contains a few other code quality improvements too.  I should probably split that up into its own PR

felixarntz: Makes sense, in that case I think we should also update the PL admin screen logic to sort by user-facing name once the plugin information has been fetched from .org.

westonruter: @swissspidy I've cherry-picked your commit to sort by name.

I've put the experimental plugins at the end in the readme.

plugins/performance-lab/readme.txt
```diff
@@ -11,7 +11,19 @@ Performance plugin from the WordPress Performance Team, which is a collection of
...
+
+The feature plugins which are currently featured by this plugin are:
+
+* [Image Placeholders](https://wordpress.org/plugins/dominant-color-images/)
+* [Modern Image Formats](https://wordpress.org/plugins/webp-uploads/)
+* [Performant Translations](https://wordpress.org/plugins/performant-translations/)
+* [Speculative Loading](https://wordpress.org/plugins/speculation-rules/)
+* [Embed Optimizer](https://wordpress.org/plugins/embed-optimizer/) _(experimental)_
+* [Enhanced Responsive Images](https://wordpress.org/plugins/auto-sizes/) _(experimental)_
+* [Image Prioritizer](https://wordpress.org/plugins/image-prioritizer/) _(experimental)_
```
mukeshpanchal27, line 24: Should we add the plugin list to the GH repo [README.md](https://github.com/WordPress/performance/blob/trunk/README.md) file as well?

mukeshpanchal27 (approved):
Nice! I left a question. Nothing blocking.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -114,26 +113,25 @@ function perflab_render_plugins_ui(): void {
...
+				$api_data
+			);
 		}
 	}
 
-	if ( count( $plugins ) === 0 && count( $experimental_plugins ) === 0 ) {
+	$plugins = wp_list_sort(
+		$plugins,
+		array(
+			'experimental' => 'ASC',
```
felixarntz, line 135: This is a bit "too" smart for me, sorting by a boolean is not exactly intuitive. Not opposed to keeping it as is because it _is_ smart, but this kind of thing should have a comment explaining it I think. :)

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -114,26 +113,25 @@ function perflab_render_plugins_ui(): void {
 				),
 				array( 'type' => 'error' )
 			);
-			continue;
```
felixarntz, line 117: Why is this removed? Won't this cause errors down the line if we pass a `WP_Error` to `array_merge()`?

felixarntz (approved):
@westonruter This looks great to me, just two small comments.

Will already approve, but worth addressing them IMO.

plugins/performance-lab/includes/admin/plugins.php
swissspidy: No. There is a `if ( $api_data instanceof WP_Error ) {` safeguard, so it's always an array when merging. If not, PHPStan would complain.

For context, I refactored it this way in #1098 so that the PL modules page still works if the plugin is not yet in WordPress.org (like when developing locally). Otherwise you'd get  a`Failed to query WordPress.org Plugin Directory for plugin "web-worker-offloading". Plugin not found.` notice right now, which doesn't make sense.

felixarntz: @swissspidy Oh, what I had missed due to the diff was that now the entire code below is in an `else` for this `if`. That's why I was confused about the missing `continue`. :)

plugins/performance-lab/readme.txt
westonruter: Good idea. Done in 436d68e986a02e5b087d3de52780f7568d37b943:

image

plugins/performance-lab/includes/admin/plugins.php
westonruter: @swissspidy I'll let you amend with an explanatory comment and then you can merge 😄

swissspidy: Don‘t really think this warrants a comment, but sure, I can add one tomorrow.

PR #1442, opened 2024-08-06
Ensure only HTML documents are processed by Optimization Detective

westonruter:
_This is a sub-PR of https://github.com/WordPress/performance/pull/1317. Merge it first._

This fixes a couple cases where Optimization Detective could erroneously attempt to optimize a non-HTML response:

* A JSON response body that erroneously sends a `Content-Type: text/html` header.
* An XML response body that erroneously sends a `Content-Type: text/html` header.


felixarntz (approved):
LGTM!

PR #1443, opened 2024-08-06
Avoid sending Server-Timing header when buffer is being cleaned

westonruter:
_This is a sub-PR of #1317. Merge it first._

As @felixarntz pointed out in https://github.com/WordPress/performance/pull/1317#issuecomment-2272019860, when a template calls `ob_clean()` it can currently result in the `Server-Timing` header being sent prematurely. When the buffer is being cleaned, the output buffer callback should not send the header yet because no output is being printed, so it can continue waiting until the output is flushed.

This also fixes a bug with passing the argument names when they are missing from the `$args` passed to `\Perflab_Server_Timing::register_metric()`. The PR also improves escaping the resultant strings.


felixarntz (approved):
Thanks @westonruter!

PR #1445, opened 2024-08-07
Harden logic to add `auto` keyword to `sizes` attribute to prevent duplicate keyword

felixarntz:
Currently, the Enhanced Responsive Images functions that conditionally add the `auto` keyword to the image `sizes` attribute only check for the attribute in one specific way. This works to prevent double processing of an image that the plugin itself has added, but may lead to problems if the `auto` keyword was added through another means, where it can appear in a different position within the `sizes` attribute.

## Relevant technical choices

* Create new helper function `auto_sizes_attribute_includes_auto( string $sizes_attr ): bool` that checks a sizes attribute for presence of the `auto` keyword, so that this logic can be reused between both filter callbacks.
* Add test coverage for both the existing and new logic to prevent duplicate `auto` keywords.

### DIFF
--- a/plugins/auto-sizes/hooks.php
+++ b/plugins/auto-sizes/hooks.php
@@ -63,7 +63,7 @@ function auto_sizes_update_content_img_tag( $html ): string {
 	}
 
 	// Bail early if the image is not responsive.
-	if ( preg_match( '/sizes="([^"]+)"/', $html, $match ) === 0 || ! isset( $match[1] ) ) {
+	if ( 1 !== preg_match( '/sizes="([^"]+)"/', $html, $match ) ) {
 		return $html;
 	}
 
--- a/plugins/auto-sizes/hooks.php
+++ b/plugins/auto-sizes/hooks.php
@@ -34,7 +34,7 @@ function auto_sizes_update_image_attributes( $attr ): array {
 	}
 
 	// Don't add 'auto' to the sizes attribute if it already exists.
-	if ( str_contains( $attr['sizes'], 'auto,' ) ) {
+	if ( auto_sizes_attribute_includes_auto( $attr['sizes'] ) ) {
 		return $attr;
 	}
 
@@ -63,12 +63,12 @@ function auto_sizes_update_content_img_tag( $html ): string {
 	}
 
 	// Bail early if the image is not responsive.
-	if ( false === strpos( $html, 'sizes="' ) ) {
+	if ( preg_match( '/sizes="([^"]+)"/', $html, $match ) === 0 || ! isset( $match[1] ) ) {
 		return $html;
 	}
 
-	// Don't double process content images.
-	if ( false !== strpos( $html, 'sizes="auto,' ) ) {
+	// Don't add 'auto' to the sizes attribute if it already exists.
+	if ( auto_sizes_attribute_includes_auto( $match[1] ) ) {
 		return $html;
 	}
 
@@ -77,6 +77,26 @@ function auto_sizes_update_content_img_tag( $html ): string {
 	return $html;
 }
 add_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
+
+/**
+ * Checks whether the given 'sizes' attribute includes the 'auto' keyword.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $sizes_attr The 'sizes' attribute value.
+ * @return bool True if the 'auto' keyword is present, false otherwise.
+ */
+function auto_sizes_attribute_includes_auto( string $sizes_attr ): bool {
+	if ( str_contains( $sizes_attr, 'auto,' ) ) {
+		return true;
+	}
+
+	if ( 'auto' === $sizes_attr ) {
+		return true;
+	}
+
+	return str_ends_with( $sizes_attr, ' auto' ) || str_ends_with( $sizes_attr, ',auto' );
+}
 
 /**
  * Displays the HTML generator tag for the plugin.
--- a/plugins/auto-sizes/hooks.php
+++ b/plugins/auto-sizes/hooks.php
@@ -87,11 +87,12 @@ add_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
  * @return bool True if the 'auto' keyword is present, false otherwise.
  */
 function auto_sizes_attribute_includes_auto( string $sizes_attr ): bool {
-	return in_array( 
-		'auto', 
-		preg_split( '/\s*,\s*/', strtolower( trim( $sizes_attr ) ) ), 
-		true 
-	);
+	$parts = preg_split( '/\s*,\s*/', strtolower( trim( $sizes_attr ) ) );
+	if ( ! is_array( $parts ) ) {
+		return false;
+	}
+
+	return in_array( 'auto', $parts, true );
 }
 
 /**
--- a/plugins/auto-sizes/hooks.php
+++ b/plugins/auto-sizes/hooks.php
@@ -89,12 +89,8 @@ add_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
  * @return bool True if the 'auto' keyword is present, false otherwise.
  */
 function auto_sizes_attribute_includes_valid_auto( string $sizes_attr ): bool {
-	$parts = preg_split( '/\s*,\s*/', strtolower( trim( $sizes_attr ) ) );
-	if ( ! is_array( $parts ) || ! isset( $parts[0] ) ) {
-		return false;
-	}
-
-	return 'auto' === $parts[0];
+	$token = strtok( strtolower( $sizes_attr ), ',' );
+	return false !== $token && 'auto' === trim( $token, " \t\f\r\n" );
 }
 
 /**
--- a/plugins/auto-sizes/hooks.php
+++ b/plugins/auto-sizes/hooks.php
@@ -34,7 +34,7 @@ function auto_sizes_update_image_attributes( $attr ): array {
 	}
 
 	// Don't add 'auto' to the sizes attribute if it already exists.
-	if ( auto_sizes_attribute_includes_auto( $attr['sizes'] ) ) {
+	if ( auto_sizes_attribute_includes_valid_auto( $attr['sizes'] ) ) {
 		return $attr;
 	}
 
@@ -68,7 +68,7 @@ function auto_sizes_update_content_img_tag( $html ): string {
 	}
 
 	// Don't add 'auto' to the sizes attribute if it already exists.
-	if ( auto_sizes_attribute_includes_auto( $match[1] ) ) {
+	if ( auto_sizes_attribute_includes_valid_auto( $match[1] ) ) {
 		return $html;
 	}
 
@@ -79,20 +79,22 @@ function auto_sizes_update_content_img_tag( $html ): string {
 add_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
 
 /**
- * Checks whether the given 'sizes' attribute includes the 'auto' keyword.
+ * Checks whether the given 'sizes' attribute includes the 'auto' keyword as the first item in the list.
+ *
+ * Per the HTML spec, if present it must be the first entry.
  *
  * @since n.e.x.t
  *
  * @param string $sizes_attr The 'sizes' attribute value.
  * @return bool True if the 'auto' keyword is present, false otherwise.
  */
-function auto_sizes_attribute_includes_auto( string $sizes_attr ): bool {
+function auto_sizes_attribute_includes_valid_auto( string $sizes_attr ): bool {
 	$parts = preg_split( '/\s*,\s*/', strtolower( trim( $sizes_attr ) ) );
-	if ( ! is_array( $parts ) ) {
+	if ( ! is_array( $parts ) || ! isset( $parts[0] ) ) {
 		return false;
 	}
 
-	return in_array( 'auto', $parts, true );
+	return 'auto' === $parts[0];
 }
 
 /**
--- a/plugins/auto-sizes/hooks.php
+++ b/plugins/auto-sizes/hooks.php
@@ -87,15 +87,11 @@ add_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
  * @return bool True if the 'auto' keyword is present, false otherwise.
  */
 function auto_sizes_attribute_includes_auto( string $sizes_attr ): bool {
-	if ( str_contains( $sizes_attr, 'auto,' ) ) {
-		return true;
-	}
-
-	if ( 'auto' === $sizes_attr ) {
-		return true;
-	}
-
-	return str_ends_with( $sizes_attr, ' auto' ) || str_ends_with( $sizes_attr, ',auto' );
+	return in_array( 
+		'auto', 
+		preg_split( '/\s*,\s*/', strtolower( trim( $sizes_attr ) ) ), 
+		true 
+	);
 }
 
 /**
--- a/plugins/auto-sizes/tests/test-auto-sizes.php
+++ b/plugins/auto-sizes/tests/test-auto-sizes.php
@@ -188,11 +188,11 @@ class Test_AutoSizes extends WP_UnitTestCase {
 				'auto',
 				false,
 			),
-			'with space at beginning'                => array(
+			'with space at beginning'     => array(
 				' auto, (max-width: 1024px) 100vw, 1024px',
 				false,
 			),
-			'with uppercase'                => array(
+			'with uppercase'              => array(
 				'AUTO, (max-width: 1024px) 100vw, 1024px',
 				false,
 			),
--- a/plugins/auto-sizes/tests/test-auto-sizes.php
+++ b/plugins/auto-sizes/tests/test-auto-sizes.php
@@ -98,7 +98,7 @@ class Test_AutoSizes extends WP_UnitTestCase {
 	 * Test generated markup for an image with 'auto' keyword already present in sizes does not receive it again.
 	 *
 	 * @covers ::auto_sizes_update_image_attributes
-	 * @covers ::auto_sizes_attribute_includes_auto
+	 * @covers ::auto_sizes_attribute_includes_valid_auto
 	 * @dataProvider data_image_with_existing_auto_sizes
 	 */
 	public function test_image_with_existing_auto_sizes_is_not_processed_again( string $initial_sizes, bool $expected_processed ): void {
@@ -123,7 +123,7 @@ class Test_AutoSizes extends WP_UnitTestCase {
 	 * Test content filtered markup with 'auto' keyword already present in sizes does not receive it again.
 	 *
 	 * @covers ::auto_sizes_update_content_img_tag
-	 * @covers ::auto_sizes_attribute_includes_auto
+	 * @covers ::auto_sizes_attribute_includes_valid_auto
 	 * @dataProvider data_image_with_existing_auto_sizes
 	 */
 	public function test_content_image_with_existing_auto_sizes_is_not_processed_again( string $initial_sizes, bool $expected_processed ): void {
@@ -168,33 +168,41 @@ class Test_AutoSizes extends WP_UnitTestCase {
 				'auto, (max-width: 1024px) 100vw, 1024px',
 				false,
 			),
-			'within, without space'       => array(
-				'(max-width: 1024px) 100vw, auto,1024px',
-				false,
-			),
-			'within, with space'          => array(
-				'(max-width: 1024px) 100vw, auto, 1024px',
-				false,
-			),
-			'at the end, without space'   => array(
-				'(max-width: 1024px) 100vw,auto',
-				false,
-			),
-			'at the end, with space'      => array(
-				'(max-width: 1024px) 100vw, auto',
-				false,
-			),
 			'sole keyword'                => array(
 				'auto',
 				false,
 			),
-			'with space at beginning'     => array(
+			'with space before'           => array(
 				' auto, (max-width: 1024px) 100vw, 1024px',
 				false,
 			),
 			'with uppercase'              => array(
 				'AUTO, (max-width: 1024px) 100vw, 1024px',
 				false,
+			),
+
+			/*
+			 * The following scenarios technically include the 'auto' keyword,
+			 * but it is in the wrong place, as per the HTML spec it must be
+			 * the first entry in the list.
+			 * Therefore in these invalid cases the 'auto' keyword should still
+			 * be added to the beginning of the list.
+			 */
+			'within, without space'       => array(
+				'(max-width: 1024px) 100vw, auto,1024px',
+				true,
+			),
+			'within, with space'          => array(
+				'(max-width: 1024px) 100vw, auto, 1024px',
+				true,
+			),
+			'at the end, without space'   => array(
+				'(max-width: 1024px) 100vw,auto',
+				true,
+			),
+			'at the end, with space'      => array(
+				'(max-width: 1024px) 100vw, auto',
+				true,
 			),
 		);
 	}
--- a/plugins/auto-sizes/tests/test-auto-sizes.php
+++ b/plugins/auto-sizes/tests/test-auto-sizes.php
@@ -169,11 +169,11 @@ class Test_AutoSizes extends WP_UnitTestCase {
 				false,
 			),
 			'within, without space'       => array(
-				'(max-width: 1024px) auto,1024px',
+				'(max-width: 1024px) 100vw, auto,1024px',
 				false,
 			),
 			'within, with space'          => array(
-				'(max-width: 1024px) auto, 1024px',
+				'(max-width: 1024px) 100vw, auto, 1024px',
 				false,
 			),
 			'at the end, without space'   => array(
--- a/plugins/auto-sizes/tests/test-auto-sizes.php
+++ b/plugins/auto-sizes/tests/test-auto-sizes.php
@@ -188,6 +188,14 @@ class Test_AutoSizes extends WP_UnitTestCase {
 				'auto',
 				false,
 			),
+			'with space at beginning'                => array(
+				' auto, (max-width: 1024px) 100vw, 1024px',
+				false,
+			),
+			'with uppercase'                => array(
+				'AUTO, (max-width: 1024px) 100vw, 1024px',
+				false,
+			),
 		);
 	}
 
--- a/plugins/auto-sizes/tests/test-auto-sizes.php
+++ b/plugins/auto-sizes/tests/test-auto-sizes.php
@@ -168,6 +168,14 @@ class Test_AutoSizes extends WP_UnitTestCase {
 				'auto, (max-width: 1024px) 100vw, 1024px',
 				false,
 			),
+			'within, without space'       => array(
+				'(max-width: 1024px) auto,1024px',
+				false,
+			),
+			'within, with space'          => array(
+				'(max-width: 1024px) auto, 1024px',
+				false,
+			),
 			'at the end, without space'   => array(
 				'(max-width: 1024px) 100vw,auto',
 				false,
--- a/plugins/auto-sizes/tests/test-auto-sizes.php
+++ b/plugins/auto-sizes/tests/test-auto-sizes.php
@@ -95,6 +95,95 @@ class Test_AutoSizes extends WP_UnitTestCase {
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


plugins/auto-sizes/hooks.php
```diff
@@ -78,6 +78,26 @@ function auto_sizes_update_content_img_tag( $html ): string {
...
+function auto_sizes_attribute_includes_auto( string $sizes_attr ): bool {
+	if ( str_contains( $sizes_attr, 'auto,' ) ) {
+		return true;
+	}
+
+	if ( 'auto' === $sizes_attr ) {
+		return true;
+	}
+
+	return str_ends_with( $sizes_attr, ' auto' ) || str_ends_with( $sizes_attr, ',auto' );
```
westonruter: What about splitting the string on commas+whitespace and then seeing if `auto` is inside? 

```suggestion
	return in_array( 
		'auto', 
		preg_split( '/\s*,\s*/', strtolower( trim( $sizes_attr ) ) ), 
		true 
	);
```

Note also the inclusion of `strtolower()` since the specs ays it is case-insensitive: https://html.spec.whatwg.org/multipage/images.html#sizes-attributes

plugins/auto-sizes/hooks.php
```diff
@@ -78,6 +78,26 @@ function auto_sizes_update_content_img_tag( $html ): string {
 }
 add_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
```
westonruter, line 79: This seems like a great opportunity to use the HTML Tag Processor:

```php
function auto_sizes_update_content_img_tag( $html ): string {
	if ( ! is_string( $html ) ) {
		$html = '';
	}

	// Bail if there is no IMG tag.
		return $html;
	}

	// Bail early if the image is not lazy-loaded.
	if ( 'lazy' !== $processor->get_attribute( 'loading' ) ) {
		return $html;
	}

	// Bail early if the image is not responsive.
		return $html;
	}

	if ( auto_sizes_attribute_includes_auto( $sizes ) ) {
		return $html;
	}

}
```

plugins/auto-sizes/tests/test-auto-sizes.php
```diff
@@ -94,6 +94,103 @@ public function test_content_image_without_lazy_loading_does_not_have_auto_sizes
...
+				false,
+			),
+			'at the end, with space'      => array(
+				'(max-width: 1024px) 100vw, auto',
+				false,
+			),
+			'sole keyword'                => array(
+				'auto',
+				false,
+			),
```
westonruter, line 206:
```suggestion
			),
			'with space at beginning'                => array(
				' auto, (max-width: 1024px) 100vw, 1024px',
				false,
			),
			'with uppercase'                => array(
				'AUTO, (max-width: 1024px) 100vw, 1024px',
				false,
			),
```

plugins/auto-sizes/tests/test-auto-sizes.php
```diff
@@ -94,6 +94,103 @@ public function test_content_image_without_lazy_loading_does_not_have_auto_sizes
...
+			'within, without space'       => array(
+				'(max-width: 1024px) auto,1024px',
+				false,
+			),
+			'within, with space'          => array(
+				'(max-width: 1024px) auto, 1024px',
+				false,
+			),
```
westonruter: Are these valid `sizes` attributes? I didn't think that `auto` was supported with a media condition? See grammar in spec: https://html.spec.whatwg.org/multipage/images.html#sizes-attributes

In this case, shouldn't `auto` still be prepended in the case of invalidity?

felixarntz: Good point, I'm not sure. In terms of the tests though it doesn't really matter since they are just there to check that the logic works as expected.

If we want to handle this differently and still provide `auto` when it's used in an invalid way, I think we could explore that in a separate issue/PR.

plugins/auto-sizes/hooks.php
felixarntz: While from a DX perspective I agree, there are potential performance implications for it, given how often this is invoked. I believe some explorations in that regard were made in the past that surfaced some problems, but I don't recall the exact specifics.

Either way, I think that the right place to use `WP_HTML_Tag_Processor` for this is in WP core itself, and then introduce new filters that deal with the specific tags and attributes parsed by the processor. I don't think instantiating a new instance of the class for every single image in the content is a good approach.

plugins/auto-sizes/tests/test-auto-sizes.php
```diff
@@ -94,6 +94,111 @@ public function test_content_image_without_lazy_loading_does_not_have_auto_sizes
...
+			'in beginning, without space' => array(
+				'auto,(max-width: 1024px) 100vw, 1024px',
+				false,
+			),
+			'in beginning, with space'    => array(
+				'auto, (max-width: 1024px) 100vw, 1024px',
+				false,
+			),
+			'within, without space'       => array(
```
westonruter:
```suggestion
			// TODO: The following two cases are invalid according to the grammar in the spec: <https://html.spec.whatwg.org/multipage/images.html#sizes-attributes>. Should 'auto' still be prepended in such cases where the syntax is bad? 
			'within, without space'       => array(
```

felixarntz: I'd rather have us open an issue to discuss that :)

felixarntz: Well, turns out after your proposed change, which I committed these test scenarios fail anyway because the `auto` is not the _only_ content within. So I'll just change this to something else, which probably makes equally little sense, but at least with your change a media query with auto is already no longer supported now :)

felixarntz: See https://github.com/WordPress/performance/pull/1445/commits/471a1d0be46808a937e884b1897d0a0cd5328d47

plugins/auto-sizes/hooks.php
westonruter: It does have better DX, but it's not really about that. It's about robustness, safety, and ensuring the result is correct. Instantiating the `WP_HTML_Tag_Processor` is very lightweight. It's being used increasingly throughout core as this is its intended purpose. We're also using it in our Performance plugins including this one:

https://github.com/WordPress/performance/blob/e7313851f39749886a4ae0bacbcd620a6effce3d/plugins/auto-sizes/hooks.php#L122-L141

https://github.com/WordPress/performance/blob/e7313851f39749886a4ae0bacbcd620a6effce3d/plugins/auto-sizes/hooks.php#L153-L157

https://github.com/WordPress/performance/blob/e7313851f39749886a4ae0bacbcd620a6effce3d/plugins/webp-uploads/picture-element.php#L84-L89

https://github.com/WordPress/performance/blob/e7313851f39749886a4ae0bacbcd620a6effce3d/plugins/webp-uploads/hooks.php#L542-L562

If there are problems with HTML Tag Processor, we should surface them to fix in core. Otherwise, we're using it elsewhere for this purpose, so I think we should use it here too.

westonruter: I'm basically channeling @dmsnell right now 😂

plugins/auto-sizes/tests/test-auto-sizes.php
```diff
@@ -94,6 +94,111 @@ public function test_content_image_without_lazy_loading_does_not_have_auto_sizes
...
+				false,
+			),
+			'with space at beginning'     => array(
+				' auto, (max-width: 1024px) 100vw, 1024px',
+				false,
+			),
+			'with uppercase'              => array(
+				'AUTO, (max-width: 1024px) 100vw, 1024px',
+				false,
+			),
```
felixarntz: FYI: All of these would have failed prior to the production code changes in this PR.

plugins/auto-sizes/hooks.php
felixarntz: I'm unsure about this for the reasons previously outlined. I believe @joemcgill or @kt-12 attempted to use the tag processor for something like this before and ran into performance issues.

Either way, it's a separate change that should be separately discussed. :)

joemcgill:
> This works to prevent double processing of an image that the plugin itself has added, but may lead to problems if the auto keyword was added through another means, where it can appear in a different position within the sizes attribute.

@felixarntz can you give an example of what problem this is actually solving? Per the HTML spec, `auto` should _only_ appear as the first value in the `sizes` list. Placing it elsewhere in `sizes` would not be valid.

plugins/auto-sizes/hooks.php
```diff
@@ -63,12 +63,12 @@ function auto_sizes_update_content_img_tag( $html ): string {
 	}
 
 	// Bail early if the image is not responsive.
-	if ( false === strpos( $html, 'sizes="' ) ) {
+	if ( preg_match( '/sizes="([^"]+)"/', $html, $match ) === 0 || ! isset( $match[1] ) ) {
```
westonruter: This is simpler and with it PHPStan is smart enough to know that `$match[1]` is set:

```suggestion
	if ( 1 !== preg_match( '/sizes="([^"]+)"/', $html, $match ) ) {
```

plugins/auto-sizes/hooks.php
joemcgill: Extra performance overhead was true of the first iteration of the HTML Tag Processor, in WP 6.4 (I believe). However, a lot has been improved since then. I don't think we've conducted a thorough analysis of whether using it in cases like this still has potential negative performance implications. Given how useful this is and how quickly I've seen adoption throughout Core, I would be in support of us doing some additional analysis. In the mean time, I think we should use it where we can so we're dogfooding our own WP Core APIs.

felixarntz:
@joemcgill 

> @felixarntz can you give an example of what problem this is actually solving? Per the HTML spec, `auto` should _only_ appear as the first value in the `sizes` list. Placing it elsewhere in `sizes` would not be valid.

Is that correct? Will the value not be parsed if it's placed elsewhere?

Regardless, if that's the case, some valid usages were not covered by the implementation so far, such as `sizes="auto"` or `sizes="AUTO, ..."`.

joemcgill (dismissed):
I'm happy to see us consolidate some of this early logic for checking whether `sizes` already exists with `"auto"` applied into a helper function, but we should _only_ be looking for `"auto"` in the first position of the attribute. Otherwise, I don't believe the attribute will have the intended effect in the first place as it's expected to be the first thing in the attribute.

joemcgill:
> Regardless, if that's the case, some valid usages were not covered by the implementation so far, such as sizes="auto" or sizes="AUTO, ...".

If there are valid cases like that which we can demonstrate get parsed correctly by browsers, then I agree we should handle them. The standalone `sizes="auto"` value is a great catch, for example.

westonruter:
> can you give an example of what problem this is actually solving? Per the HTML spec, `auto` should _only_ appear as the first value in the `sizes` list. Placing it elsewhere in `sizes` would not be valid.

> I'm happy to see us consolidate some of this early logic for checking whether `sizes` already exists with `"auto"` applied into a helper function, but we should _only_ be looking for `"auto"` in the first position of the attribute. Otherwise, I don't believe the attribute will have the intended effect in the first place as it's expected to be the first thing in the attribute.

@joemcgill It seems like [the spec](https://html.spec.whatwg.org/multipage/images.html#sizes-attributes) does allow `auto` anywhere: 

```
<source-size-list> = <source-size>#? , <source-size-value>
<source-size> = <media-condition> <source-size-value> | auto
<source-size-value> = <length> | auto
```

plugins/auto-sizes/hooks.php
westonruter: > Either way, it's a separate change that should be separately discussed. :)

IMO, it's not really a separate change because it is also about hardening the logic add the `auto` keyword to the `sizes` attribute. Here's a test case where duplication occurs:

```html
<img
    data-tshirt-sizes="S M L"
    src="https://pd.w.org/2024/08/36566abfcb2952e20.99062056-300x225.jpg"
    srcset="https://pd.w.org/2024/08/36566abfcb2952e20.99062056-300x225.jpg 300w, https://pd.w.org/2024/08/36566abfcb2952e20.99062056-1024x768.jpg 1024w, https://pd.w.org/2024/08/36566abfcb2952e20.99062056-768x576.jpg 768w, https://pd.w.org/2024/08/36566abfcb2952e20.99062056-1536x1152.jpg 1536w, https://pd.w.org/2024/08/36566abfcb2952e20.99062056-2048x1536.jpg 2048w"
    sizes="auto, (max-width: 650px) 100vw, 650px"
    loading="lazy"
>
```

This results in:

```html
<img decoding="async"
    data-tshirt-sizes="auto, S M L"
    src="https://pd.w.org/2024/08/36566abfcb2952e20.99062056-300x225.jpg"
    srcset="[https://pd.w.org/2024/08/36566abfcb2952e20.99062056-300x225.jpg 300w](https://pd.w.org/2024/08/36566abfcb2952e20.99062056-300x225.jpg),[ https://pd.w.org/2024/08/36566abfcb2952e20.99062056-1024x768.jpg 1024w](https://pd.w.org/2024/08/36566abfcb2952e20.99062056-1024x768.jpg),[ https://pd.w.org/2024/08/36566abfcb2952e20.99062056-768x576.jpg 768w](https://pd.w.org/2024/08/36566abfcb2952e20.99062056-768x576.jpg),[ https://pd.w.org/2024/08/36566abfcb2952e20.99062056-1536x1152.jpg 1536w](https://pd.w.org/2024/08/36566abfcb2952e20.99062056-1536x1152.jpg),[ https://pd.w.org/2024/08/36566abfcb2952e20.99062056-2048x1536.jpg 2048w](https://pd.w.org/2024/08/36566abfcb2952e20.99062056-2048x1536.jpg)"
    sizes="auto, auto, (max-width: 650px) 100vw, 650px"
    loading="lazy"
>
```

In any case, I've opened #1446 to continue this, as well as to document other test cases which fail with the current implementation of `auto_sizes_update_content_img_tag()`.

plugins/auto-sizes/hooks.php
dmsnell: Note that `\s` may be locale-sensitive, while `" \t\f\r\n"` will always match the rules from HTML and CSS.

if performance is a concern, we could avoid the allocations that `preg_split()` introduces and check directly with a `preg_match`. according to [the spec](https://html.spec.whatwg.org/#parse-a-sizes-attribute), I believe this should indicate if `auto` is already the indicated size.

```php
/*
 * For the value to be parsed as `auto` it must either be the string `auto` or
 * start with `auto,` after removing the whitespace, followed by other content.
 *
 * @see https://html.spec.whatwg.org/#valid-source-size-list
 * @see https://html.spec.whatwg.org/#parse-a-sizes-attribute
 */
return 1 === preg_match( '/^[ \t\f\r\n]*auto[ \t\f\r\n]*(?:,|$)/i', $sizes_attr );
```

This can also be realized without the PCRE functions, which may have additional perks, including being more legible to some.

```php
function sizes_attribute_is_auto( $value ) {
	$at = strspn( $value, " \t\f\r\n" );
	if ( 0 !== substr_compare( $value, 'auto', $at, 4, true ) ) {
		return false;
	}

	$at += strspn( $value, " \t\f\r\n", $at + 4 );
	return $at === strlen( $value ) || ',' === $value[ $at ];
}
```

plugins/auto-sizes/hooks.php
```diff
@@ -78,6 +78,23 @@ function auto_sizes_update_content_img_tag( $html ): string {
 }
 add_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
 
+/**
+ * Checks whether the given 'sizes' attribute includes the 'auto' keyword.
```
dmsnell: is "includes" the right semantic? it would seem like the goal is to add `auto` if it's not _set_ and avoid double-adding it if it already is. I only skimmed the spec, but it uses strong language.

> The keyword auto is a width that is computed in [parse a sizes attribute](https://html.spec.whatwg.org/#parse-a-sizes-attribute). If present, it must be the first entry and the entire [<source-size-list>](https://html.spec.whatwg.org/#source-size-list) value must either be the string "auto" ([ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive)) or start with the string "auto," ([ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive)).

That being said, it does often state assertions that are intended for authoring HTML whereas a parser might have to have different support, for example, a parser may need to interpret `auto` anywhere in the comma-separated list of tokens to be an `auto` size.

A simple browser test should clear it up.

If it mattered, one could splice out any existing `auto` values that aren't at the front, but presumably they would be benign and ignored.

joemcgill:
The `sizes` attribute gets parsed left to right and uses the first value that is applicable. It may be technically possible to put `"auto"` somewhere other than the first value without breaking the parser, but I don't think it would work as intended. Regardless, for our purposes we're only interested in cases where people are using the intended value so that we aren't applying auto-sizes to a `sizes` attribute that is already using "auto" correctly, so checking the first position is sufficient.

felixarntz:
I'm happy to update this PR to only look for an `auto` occurrence in the beginning. The spec technically allows for it to be in any position, but at the same time it's clearly formulated that it must be at the beginning or the entire attribute.

Most of this PR is test coverage anyway, and it'll still add support for the valid case of `sizes="auto"`, which would currently lead to the plugin creating `sizes="auto, auto"`, which doesn't make sense. :)

plugins/auto-sizes/hooks.php
felixarntz: Updated in https://github.com/WordPress/performance/pull/1445/commits/f34ab233af668c6a44c9079145c566e7a20969b8

felixarntz:
@joemcgill In https://github.com/WordPress/performance/pull/1445/commits/f34ab233af668c6a44c9079145c566e7a20969b8 I've updated the PR to only look for `auto` in the beginning.

I kept all tests from before and just changed the expected value, so that it is captured that in the case of `auto` being in an invalid position, it _should_ still be added by the plugin.

plugins/auto-sizes/hooks.php
westonruter: @felixarntz Any issue with this suggestion?

felixarntz: Had missed it :)

felixarntz: Have you checked though that PHPStan is fine with it? It was requiring me to check for `isset( $match[1] )`, which I really added just to please the tool. :)

plugins/auto-sizes/hooks.php
```diff
@@ -78,6 +78,25 @@ function auto_sizes_update_content_img_tag( $html ): string {
...
+ * @param string $sizes_attr The 'sizes' attribute value.
+ * @return bool True if the 'auto' keyword is present, false otherwise.
+ */
+function auto_sizes_attribute_includes_valid_auto( string $sizes_attr ): bool {
+	$parts = preg_split( '/\s*,\s*/', strtolower( trim( $sizes_attr ) ) );
+	if ( ! is_array( $parts ) || ! isset( $parts[0] ) ) {
+		return false;
+	}
+
+	return 'auto' === $parts[0];
```
westonruter: Building on @dmsnell's suggestion in https://github.com/WordPress/performance/pull/1445#discussion_r1708845701 but now taking into account we're only looking at the first token in the list, I think this can be further simplified:

```suggestion
	$token = strtok( strtolower( $sizes_attr ), ',' );
	return false !== $token && 'auto' === trim( $token, " \t\f\r\n" );
```

Note that `trim()` without the second arg includes `\0` and `\v` which are not considered whitespace here in the HTML spec.

**Update:** Eliminated the use of type casting.

plugins/auto-sizes/hooks.php
westonruter: Yes, I tested locally.

dmsnell: couple quick thoughts:
 - there's no need to store the result or pass in `$match` since we're not using it
 - the `sizes` attribute could be in any case, but this only tests for all-lowercase matches
 - the `sizes` attribute could have single quotes, or no quotes, and have whitespace around the equals sign, e.g. `Sizes = Auto`
 - the `sizes` attribute _must_ be preceded by whitespace or a `/`, but this test also matches `data-sizes="unrelated attribute"`

westonruter: With the current code, PHPStan types `$match` as `array{0?: string, 1?: non-empty-string}`

But with the suggested change, the type becomes `array{string, non-empty-string}`.

westonruter: @dmsnell:

> * there's no need to store the result or pass in `$match` since we're not using it

Further below the `$match` is used:

```php
auto_sizes_attribute_includes_valid_auto( $match[1] )
```

> * the `sizes` attribute could be in any case, but this only tests for all-lowercase matches
> * the `sizes` attribute could have single quotes, or no quotes, and have whitespace around the equals sign, e.g. `Sizes = Auto`
> * the `sizes` attribute _must_ be preceded by whitespace or a `/`, but this test also matches `data-sizes="unrelated attribute"`

Good points. These are tracked in https://github.com/WordPress/performance/issues/1446.

dmsnell: ah. well I got lost in the regexes and the attempt to parse HTML with them 😉

`$size_attribute = $processor->get_attribute( 'sizes' );` does have a certain ring to it.

plugins/auto-sizes/hooks.php
felixarntz: @westonruter Mostly looks good.

> Note that trim() without the second arg includes \0 and \v which are not considered whitespace here in the HTML spec.

I think that's overkill to consider here. There's probably tons of other `trim()` usages in core where, strictly speaking, these characters are not relevant to consider. I think passing the 2nd parameter is unnecessary and would pose the question why it's not done like that elsewhere.

dmsnell: to be fair, those tons of other `trim()` usages are part of a body of code that misparses and leads to corruption and worse. I suspect that the reason it's often done like that has more to do with people coding to their familiarity of HTML and not having fitting tools to operate on a semantic level.

why do you think it's overkill to properly parse it when all that involves is adding the second parameter? just curious on this point, because I've come across the idea before that avoiding corruption is overkill, and I don't understand the opposition to it, particularly when it involves almost no more code.

westonruter: (Not responding to Dennis but following up with some other information I learned.)

Aside: I learned some new things about `strtok()`, I was aware that you only had to pass in the string for the first call and that it would remember it for subsequent calls. What I wasn't aware is that it collapses multiple deliminters and skip over the delimiter if it appears at the beginning: https://3v4l.org/eMMYW

This means that the following would return `true`:

```php
auto_sizes_attribute_includes_valid_auto( ',,,,,,,auto,,(max-width: 1024px) 100vw, 1024px' );
```

See [note](https://www.php.net/manual/en/function.strtok.php#:~:text=argument%20is%20found.-,Note%3A,-This%20function%20behaves) in PHP function docs.

I also [asked Gemini](https://g.co/gemini/share/e9418c9e0dbd) about the history of `strtok()` only needing the `$string` for the first call:

> The unusual behavior of PHP's `strtok()` function, where subsequent calls only require the delimiter, stems from its inheritance from C and performance optimizations. It modifies the original string in-place and uses a static pointer for efficiency. While this offers benefits for large strings or memory-constrained scenarios, it also leads to non-reentrancy, string modification, and limited flexibility. Consider alternatives like `explode()` or `preg_split()` for more flexibility or a non-destructive approach.

dmsnell: interesting. thanks @westonruter - I was unaware.

in the version I tossed out there, there are no allocations and no mutation.

felixarntz: I think this is a nit-pick discussion. Either approach works and the performance difference is likely tiny. I'll let myself be convinced by benchmarks comparing the two variants, but I don't feel strongly about updating it at this point.

If you feel strongly about it, feel free to update the PR. :)

westonruter (approved):

mukeshpanchal27 (approved):

plugins/auto-sizes/hooks.php
```diff
@@ -78,6 +78,21 @@ function auto_sizes_update_content_img_tag( $html ): string {
...
+ * Per the HTML spec, if present it must be the first entry.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $sizes_attr The 'sizes' attribute value.
+ * @return bool True if the 'auto' keyword is present, false otherwise.
+ */
+function auto_sizes_attribute_includes_valid_auto( string $sizes_attr ): bool {
+	$token = strtok( strtolower( $sizes_attr ), ',' );
+	return false !== $token && 'auto' === trim( $token, " \t\f\r\n" );
```
dmsnell, line 93: @westonruter just curious, did you confirm in the browsers that `,,,,,,,auto` is still registered as `auto` for the size?

westonruter: I did not confirm, no.

dmsnell: Since I know basically nothing about this feature, I had to look up a lot (and discovered that [only Chrome supports this feature](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/sizes)).

Also I [relied on this post](https://ericportis.com/posts/2023/auto-sizes-pretty-much-requires-width-and-height/) in addition to the HTML spec to figure out what's going on.

Ultimately I confirmed that Chrome respects the spec, and strangely enough, if the attribute value is `,,,,,auto` then _no image is lazy-loaded_ *or displayed at all*. In other words, we're still trading one broken `auto` for another. Now, I don't want to be a broken record, but the code I posted before doesn't suffer this same problem.

| Setup | Chrome | Safari |
|---|---|---|
| `sizes=auto` | ✅ | 🟡 Loads the 1200w image |
| `sizes=",auto"` | 🚫 Loads _nothing_ | 🟡 Loads the 1200w image |
| `sizes="auto,"` | ✅ | 🟡 Loads the 1200w image |


Interestingly enough this broken behavior only seems to affect _lazy-loaded_ images in Chrome. That is, _if_ the page/tab is visible while loading the page, then no image loads. However, if you can get the page to load _the first time_ when the tab is hidden, then the 1200w image will load, matching Safari's behavior. To reproduce this, open up a test page where Chrome fails to render any image with the invalid attribute value, open another tab, quit Chrome, and then re-open where it restores the session. The first `srcset` image will now be displayed. I suspect this is a bug in Chrome, but _its behavior is worse under the presence of this attribute value defect than Safari's, which lacks support for the `auto` value_.


```html
<!-- borrowed from the linked article -->
<img
	loading="lazy"
	sizes=",,,,,auto"
	srcset="
		https://o.img.rodeo/w_1200,h_1200,b_tomato/t_WxH/_.png   1200w,
		https://o.img.rodeo/w_900,h_900,b_goldenrod/t_WxH/_.png   900w,
		https://o.img.rodeo/w_600,h_600,b_forestgreen/t_WxH/_.png 400w,
		https://o.img.rodeo/w_300,h_300,b_dodgerblue/t_WxH/_.png  300w"
	src="https://o.img.rodeo/w_300,h_300,b_forestgreen/t_WxH/_.png"
	alt="An example image that reports its natural dimensions"
>
```

This also seemed to have shipped without checking if the attribute _is_ `sizes` or only _ends in `sizes`_, so anything like `data-wp-sizes="state.imageSizes"` will also be corrupted by it, and the `sizes` attribute may not be updated properly.

plugins/auto-sizes/hooks.php
westonruter: I've changed my mind on the use of `strtok()` given what @dmsnell found in https://github.com/WordPress/performance/pull/1445#discussion_r1714424178. 

Undone in https://github.com/WordPress/performance/pull/1476.

plugins/auto-sizes/hooks.php
westonruter: See https://github.com/WordPress/performance/pull/1476 which handles the case of leading `,,,` chars.

plugins/auto-sizes/hooks.php
dmsnell: > I think this is a nit-pick discussion. Either approach works 

Yes it's nit-picky in some regard, but also verifiable wrong to claim that either approach works. The code complexity for the correct and incorrect code are roughly on par, but we have clear knowable failure modes for a cascade of leading commas.

It's more like "we don't anticipate seeing the HTML cases where this algorithm leads to broken results and blank images" - that's fine too, that's your call in your own plugin. It just doesn't equivocate with "either approach works" in the same way that using a PCRE function vs. string functions might.

Issue #1447, opened 2024-08-08
Enabling picture element after having uploaded images while JPEG fallbacks were disabled results in invalid markup


westonruter:
_Originally posted in https://github.com/WordPress/performance/issues/1439#issuecomment-2271472822_

When images are uploaded while "Also output JPEG" is turned off, when that setting and the picture element are later enabled, the result is incorrectly that both the `PICTURE > SOURCE` and `PICTURE > IMG` are serving JPEG. When JPEG is not available, a `PICTURE` should not be served at all since our purpose for `PICTURE` is exclusively to serve a modern image format with a JPEG fallback.

Steps to reproduce this issue:

1. Activate the plugin
2. Leave the "Also output JPEG" option disabled
3. Upload a few JPEG images
4. Insert these images as a large size in Image blocks
5. Enable "Also output JPEG" and the "Picture element" options

Bug: Go to the frontend and see that it is generating the wrong markup with the `IMG` being the modern format, not the JPEG fallback:

```html
<picture class="not-transparent wp-picture-99" style="--dominant-color: #828ca3; display: contents;">
    <source type="image/webp" srcset="http://localhost:8888/wp-content/uploads/2024/08/Spectacled_weaver_Ploceus_ocularis_ocularis_male_Mbombela-828x1024.webp 828w, http://localhost:8888/wp-content/uploads/2024/08/Spectacled_weaver_Ploceus_ocularis_ocularis_male_Mbombela-242x300.webp 242w, http://localhost:8888/wp-content/uploads/2024/08/Spectacled_weaver_Ploceus_ocularis_ocularis_male_Mbombela-768x950.webp 768w" sizes="(max-width: 650px) 100vw, 650px">
[...]

PR #1449, opened 2024-08-08
Account for responsive images being disabled when generating a PICTURE element

### LINKED ISSUE #1448: Picture element: Return additional mime type `IMG>src` when the responsive image disabled
## Bug Description
WordPress core has a mechanism to disable responsive images through some filters.

```php
add_filter( 'wp_calculate_image_sizes', '__return_false' );
```

If we disable responsive images, the Enhance Responsive Image plugin still adds the `Picture` and `source`. The expected output should be return early only IMG tag.

## Steps to reproduce

1. Add `add_filter( 'wp_calculate_image_sizes', '__return_false' )` in your functions.php
2. Activate the Modern Image Formats plugin and enable picture support from the Media Settings screen.
3. Add an image block
4. Preview the page in the front end, See the error.

```html
<picture class="wp-picture-28" style="display: contents;">
 <source 
   type="image/webp"
   srcset="
     http://localhost:8888/wp-content/uploads/2024/08/car-1024x576-jpeg.webp 1024w,
     http://localhost:8888/wp-content/uploads/2024/08/car-300x169-jpeg.webp 300w,
     http://localhost:8888/wp-content/uploads/2024/08/car-768x432-jpeg.webp 768w
   "
 >
 <source
   type="image/jpeg"
   srcset="
     http://localhost:8888/wp-content/uploads/2024/08/car-1024x576.jpeg 1024w,
     http://localhost:8888/wp-content/uploads/2024/08/car-300x169.jpeg 300w,
     http://localhost:8888/wp-content/uploads/2024/08/car-768x432.jpeg 768w
   "
 >
 <img fetchpriority="high" decoding="async" width="1024" height="576" src="http://localhost:8888/wp-content/uploads/2024/08/car-1024x576-jpeg.webp" alt="" class="wp-image-28">
</picture>
```

## Screenshots

Screenshot 2024-08-08 at 4 53 59 PM

mukeshpanchal27:
## Summary

Fixes #1448

After the PR:
Screenshot 2024-08-12 at 12 40 00 PM

### DIFF
--- a/plugins/webp-uploads/helper.php
+++ b/plugins/webp-uploads/helper.php
@@ -438,7 +438,7 @@ function webp_uploads_get_mime_type_image( int $attachment_id, string $src, stri
 	if ( isset( $metadata['sizes'] ) && is_array( $metadata['sizes'] ) ) {
 		foreach ( $metadata['sizes'] as $size => $size_data ) {
 
-			if ( empty( $size_data['file'] ) ) {
+			if ( ! isset( $size_data['file'] ) ) {
 				continue;
 			}
 
--- a/plugins/webp-uploads/helper.php
+++ b/plugins/webp-uploads/helper.php
@@ -406,3 +406,61 @@ function webp_uploads_is_picture_element_enabled(): bool {
 function webp_uploads_is_jpeg_fallback_enabled(): bool {
 	return (bool) get_option( 'perflab_generate_webp_and_jpeg' );
 }
+
+/**
+ * Retrieves the image URL for a specified MIME type from the attachment metadata.
+ *
+ * This function attempts to locate an alternate image source URL in the
+ * attachment's metadata that matches the provided MIME type.
+ *
+ * @since n.e.x.t
+ *
+ * @param int    $attachment_id The ID of the attachment.
+ * @param string $src           The original image src url.
+ * @param string $mime          A mime type we are looking to get image url.
+ * @return string|false Returns mime type image if available, false otherwise.
+ */
+function webp_uploads_get_mime_type_image( int $attachment_id, string $src, string $mime ) {
+	$metadata     = wp_get_attachment_metadata( $attachment_id );
+	$src_basename = wp_basename( $src );
+	if ( isset( $metadata['sources'][ $mime ]['file'] ) ) {
+		$basename = wp_basename( $metadata['file'] );
+
+		if ( $src_basename === $basename ) {
+			return str_replace(
+				$basename,
+				$metadata['sources'][ $mime ]['file'],
+				$src
+			);
+		}
+	}
+
+	if ( isset( $metadata['sizes'] ) && is_array( $metadata['sizes'] ) ) {
+		foreach ( $metadata['sizes'] as $size => $size_data ) {
+
+			if ( empty( $size_data['file'] ) ) {
+				continue;
+			}
+
+			if ( ! isset( $size_data['sources'][ $mime ]['file'] ) ) {
+				continue;
+			}
+
+			if ( $size_data['file'] === $size_data['sources'][ $mime ]['file'] ) {
+				continue;
+			}
+
+			if ( $src_basename !== $size_data['file'] ) {
+				continue;
+			}
+
+			return str_replace(
+				$size_data['file'],
+				$size_data['sources'][ $mime ]['file'],
+				$src
+			);
+		}
+	}
+
+	return false;
+}
--- a/plugins/webp-uploads/helper.php
+++ b/plugins/webp-uploads/helper.php
@@ -418,9 +418,9 @@ function webp_uploads_is_jpeg_fallback_enabled(): bool {
  * @param int    $attachment_id The ID of the attachment.
  * @param string $src           The original image src url.
  * @param string $mime          A mime type we are looking to get image url.
- * @return string|false Returns mime type image if available, false otherwise.
+ * @return string|null Returns mime type image if available.
  */
-function webp_uploads_get_mime_type_image( int $attachment_id, string $src, string $mime ) {
+function webp_uploads_get_mime_type_image( int $attachment_id, string $src, string $mime ): ?string {
 	$metadata     = wp_get_attachment_metadata( $attachment_id );
 	$src_basename = wp_basename( $src );
 	if ( isset( $metadata['sources'][ $mime ]['file'] ) ) {
@@ -462,5 +462,5 @@ function webp_uploads_get_mime_type_image( int $attachment_id, string $src, stri
 		}
 	}
 
-	return false;
+	return null;
 }
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -103,7 +103,7 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	list( $src, $width, $height ) = $image_src;
 	$size_array                   = array( absint( $width ), absint( $height ) );
 
-	// Get the sizes from the IMG tag.
+	// Gets the srcset and sizes from the IMG tag.
 	$sizes  = $processor->get_attribute( 'sizes' );
 	$srcset = $processor->get_attribute( 'srcset' );
 
@@ -139,18 +139,13 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 		}
 	} else {
 		foreach ( $mime_types as $image_mime_type ) {
-			// Updates the reference of the image to the a new image format if available.
-			$img       = webp_uploads_img_tag_update_mime_type( $image, 'the_content', $attachment_id );
-			$processor = new WP_HTML_Tag_Processor( $img );
-			if ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
-				$src = $processor->get_attribute( 'src' );
-				if ( is_string( $src ) ) {
-					$picture_sources .= sprintf(
-						'<source type="%s" srcset="%s">',
-						esc_attr( $image_mime_type ),
-						esc_attr( $src )
-					);
-				}
+			$image_srcset = webp_uploads_get_mime_type_image( $attachment_id, $src, $image_mime_type );
+			if ( is_string( $image_srcset ) ) {
+				$picture_sources .= sprintf(
+					'<source type="%s" srcset="%s">',
+					esc_attr( $image_mime_type ),
+					esc_attr( $image_srcset )
+				);
 			}
 		}
 	}
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -145,12 +145,13 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 			$processor = new WP_HTML_Tag_Processor( $img );
 			if ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
 				$src = $processor->get_attribute( 'src' );
-
-				$picture_sources .= sprintf(
-					'<source type="%s"%s>',
-					esc_attr( $image_mime_type ),
-					is_string( $src ) ? sprintf( ' srcset="%s"', esc_attr( $src ) ) : ''
-				);
+				if ( null !== $src ) {
+					$picture_sources .= sprintf(
+						'<source type="%s"%s>',
+						esc_attr( $image_mime_type ),
+						is_string( $src ) ? sprintf( ' srcset="%s"', esc_attr( $src ) ) : ''
+					);
+				}
 			}
 		}
 	}
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -128,15 +128,14 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 			add_filter( 'wp_calculate_image_srcset', $filter );
 			$image_srcset = wp_get_attachment_image_srcset( $attachment_id, $size_array, $image_meta );
 			remove_filter( 'wp_calculate_image_srcset', $filter );
-			if ( false === $image_srcset ) {
-				continue;
+			if ( is_string( $image_srcset ) ) {
+				$picture_sources .= sprintf(
+					'<source type="%s"%s%s>',
+					esc_attr( $image_mime_type ),
+					sprintf( ' srcset="%s"', esc_attr( $image_srcset ) ),
+					is_string( $sizes ) ? sprintf( ' sizes="%s"', esc_attr( $sizes ) ) : ''
+				);
 			}
-			$picture_sources .= sprintf(
-				'<source type="%s"%s%s>',
-				esc_attr( $image_mime_type ),
-				is_string( $image_srcset ) ? sprintf( ' srcset="%s"', esc_attr( $image_srcset ) ) : '',
-				is_string( $sizes ) ? sprintf( ' sizes="%s"', esc_attr( $sizes ) ) : ''
-			);
 		}
 	} else {
 		foreach ( $mime_types as $image_mime_type ) {
@@ -145,11 +144,11 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 			$processor = new WP_HTML_Tag_Processor( $img );
 			if ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
 				$src = $processor->get_attribute( 'src' );
-				if ( null !== $src ) {
+				if ( is_string( $src ) ) {
 					$picture_sources .= sprintf(
 						'<source type="%s"%s>',
 						esc_attr( $image_mime_type ),
-						is_string( $src ) ? sprintf( ' srcset="%s"', esc_attr( $src ) ) : ''
+						sprintf( ' srcset="%s"', esc_attr( $src ) )
 					);
 				}
 			}
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -130,9 +130,9 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 			remove_filter( 'wp_calculate_image_srcset', $filter );
 			if ( is_string( $image_srcset ) ) {
 				$picture_sources .= sprintf(
-					'<source type="%s"%s%s>',
+					'<source type="%s" srcset="%s"%s>',
 					esc_attr( $image_mime_type ),
-					sprintf( ' srcset="%s"', esc_attr( $image_srcset ) ),
+					esc_attr( $image_srcset ),
 					is_string( $sizes ) ? sprintf( ' sizes="%s"', esc_attr( $sizes ) ) : ''
 				);
 			}
@@ -146,9 +146,9 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 				$src = $processor->get_attribute( 'src' );
 				if ( is_string( $src ) ) {
 					$picture_sources .= sprintf(
-						'<source type="%s"%s>',
+						'<source type="%s" srcset="%s">',
 						esc_attr( $image_mime_type ),
-						sprintf( ' srcset="%s"', esc_attr( $src ) )
+						esc_attr( $src )
 					);
 				}
 			}
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -98,6 +98,11 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	// Get the sizes from the IMG tag.
 	$sizes = $processor->get_attribute( 'sizes' );
 
+	// Bail early when the responsive images are disabled.
+	if ( null === $sizes ) {
+		return $image;
+	}
+
 	foreach ( $mime_types as $image_mime_type ) {
 		// Filter core's wp_get_attachment_image_srcset to return the sources for the current mime type.
 		$filter = static function ( $sources ) use ( $mime_type_data, $image_mime_type ): array {
--- a/plugins/webp-uploads/picture-element.php
+++ b/plugins/webp-uploads/picture-element.php
@@ -97,42 +97,55 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	$size_array                   = array( absint( $width ), absint( $height ) );
 
 	// Get the sizes from the IMG tag.
-	$sizes = $processor->get_attribute( 'sizes' );
+	$sizes  = $processor->get_attribute( 'sizes' );
+	$srcset = $processor->get_attribute( 'srcset' );
 
-	// Bail early when the responsive images are disabled.
-	if ( null === $sizes ) {
-		return $image;
-	}
-
-	foreach ( $mime_types as $image_mime_type ) {
-		// Filter core's wp_get_attachment_image_srcset to return the sources for the current mime type.
-		$filter = static function ( $sources ) use ( $mime_type_data, $image_mime_type ): array {
-			$filtered_sources = array();
-			foreach ( $sources as $source ) {
-				// Swap the URL for the current mime type.
-				if ( isset( $mime_type_data[ $image_mime_type ][ $source['descriptor'] ][ $source['value'] ] ) ) {
-					$filename           = $mime_type_data[ $image_mime_type ][ $source['descriptor'] ][ $source['value'] ]['file'];
-					$filtered_sources[] = array(
-						'url'        => dirname( $source['url'] ) . '/' . $filename,
-						'descriptor' => $source['descriptor'],
-						'value'      => $source['value'],
-					);
+	if ( null !== $srcset && null !== $sizes ) {
+		foreach ( $mime_types as $image_mime_type ) {
+			// Filter core's wp_get_attachment_image_srcset to return the sources for the current mime type.
+			$filter = static function ( $sources ) use ( $mime_type_data, $image_mime_type ): array {
+				$filtered_sources = array();
+				foreach ( $sources as $source ) {
+					// Swap the URL for the current mime type.
+					if ( isset( $mime_type_data[ $image_mime_type ][ $source['descriptor'] ][ $source['value'] ] ) ) {
+						$filename           = $mime_type_data[ $image_mime_type ][ $source['descriptor'] ][ $source['value'] ]['file'];
+						$filtered_sources[] = array(
+							'url'        => dirname( $source['url'] ) . '/' . $filename,
+							'descriptor' => $source['descriptor'],
+							'value'      => $source['value'],
+						);
+					}
 				}
+				return $filtered_sources;
+			};
+			add_filter( 'wp_calculate_image_srcset', $filter );
+			$image_srcset = wp_get_attachment_image_srcset( $attachment_id, $size_array, $image_meta );
+			remove_filter( 'wp_calculate_image_srcset', $filter );
+			if ( false === $image_srcset ) {
+				continue;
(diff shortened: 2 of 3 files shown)


plugins/webp-uploads/picture-element.php
```diff
@@ -98,6 +98,11 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 	// Get the sizes from the IMG tag.
 	$sizes = $processor->get_attribute( 'sizes' );
 
+	// Bail early when the responsive images are disabled.
+	if ( null === $sizes ) {
+		return $image;
+	}
```
joemcgill: While it's not recommended, it's possible to have a `srcset` attribute on an image without a `sizes` attribute, in which case the browser will use the default value of `100vw` for `sizes`. We should probably check for the existence of both a `srcset` and `sizes` attribute, which is what WordPress also checks for before modifying the img tag ([here](https://github.com/WordPress/wordpress-develop/blob/6.6/src/wp-includes/media.php#L1778)).

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -234,23 +234,24 @@ public function test_disable_responsive_image_with_picture_element(): void {
...
-		$this->opt_in_to_picture_element();
-
-		$picture_markup = apply_filters( 'the_content', $image );
-
 		$img_processor = new WP_HTML_Tag_Processor( $img_markup );
 		$img_processor->next_tag( array( 'tag_name' => 'IMG' ) );
 
-		$this->assertNull( $img_processor->get_attribute( 'sizes' ), 'Sizes attribute missing in IMG tag.' );
+		$this->assertNull( $img_processor->get_attribute( 'sizes' ), 'Make sure that there is no sizes attribute in IMG tag.' );
+		$this->assertNull( $img_processor->get_attribute( 'srcset' ), 'Make sure that there is no srcset attribute in IMG tag.' );
```
joemcgill, line 290: Nit: It's ok to have these assertions here, but really what we want the tests to focus on is that when the image doesn't have `srcset` or `sizes` then the image does not get modified by our function that wraps the image markup in  a `picture` element with `source` elements for each MIME type. 

We're not really testing that returning false from the `wp_calculate_image_sizes` filter  results in no responsive image. A simpler way of doing this is skip running the first `apply_filters()` function above (so no dynamic image modifications are made), then pass the image directly to `webp_uploads_wrap_image_in_picture()` with 'the_content' set as context and test that the image hasn't been modified.

joemcgill (dismissed):
Nice catch! Just some small nitpicks here related to [my comment](https://github.com/WordPress/performance/issues/1448#issuecomment-2275926698) on the original issue.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -234,23 +234,24 @@ public function test_disable_responsive_image_with_picture_element(): void {
...
-		$picture_processor->next_tag( array( 'tag_name' => 'source' ) );
+		$picture_processor = new WP_HTML_Tag_Processor( $picture_markup );
+		$this->assertFalse( $img_processor->next_tag( array( 'tag_name' => 'picture' ) ), 'Make sure that there is no Picture tag.' );
+		$this->assertFalse( $img_processor->next_tag( array( 'tag_name' => 'source' ) ), 'Make sure that there is no source tag.' );
+		$this->assertTrue( $picture_processor->next_tag( array( 'tag_name' => 'IMG' ) ), 'Make sure that there is a IMG tag.' );
+		$this->assertNull( $picture_processor->get_attribute( 'sizes' ), 'Make sure that there is no sizes attribute in IMG tag.' );
+		$this->assertNull( $picture_processor->get_attribute( 'srcset' ), 'Make sure that there is no srcset attribute in IMG tag.' );
 
-		$this->assertNull( $picture_processor->get_attribute( 'sizes' ), 'Sizes attribute missing in Picture source tag.' );
+		$this->assertSame( $img_markup, $picture_markup, 'Make sure both markup are the same' );
```
westonruter: I don't believe this is correct per https://github.com/WordPress/performance/issues/1448#issuecomment-2276202488.

When no `srcset` is available, then the `PICTURE > SOURCE` tag should have its `srcset` attribute populated with the `src` for the modern image format. When a JPEG fallback is available, we still need to use a `PICTURE` element so that browsers that don't support the modern image format will get served the JPEG instead. 

As seen in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture#the_type_attribute), there's no need for `srcset` to have any width or pixel descriptors:

```html
<picture>
  <source srcset="photo.avif" type="image/avif" />
  <source srcset="photo.webp" type="image/webp" />
  <img src="photo.jpg" alt="photo" />
</picture>
```

So think of `srcset` as a superset of what `src` can contain.

westonruter (changes requested):

mukeshpanchal27:
After https://github.com/WordPress/performance/pull/1449/commits/7e09a7231dc04ef3772c0be58fae185dc89a3b7b, The picture element outputs:

## Responsive image:
```html
<picture class="wp-picture-12" style="display: contents;">
  <source type="image/webp" 
    srcset="
       http://localhost:8888/wp-content/uploads/2024/08/car-300x169-jpeg.webp 300w, 
       http://localhost:8888/wp-content/uploads/2024/08/car-1024x576-jpeg.webp 1024w, 
       http://localhost:8888/wp-content/uploads/2024/08/car-768x432-jpeg.webp 768w
    " 
    sizes="(max-width: 1080px) 100vw, 1080px"
  >
  <img 
    fetchpriority="high" 
    decoding="async" 
    width="1080" 
    height="608" 
    src="http://localhost:8888/wp-content/uploads/2024/08/car.jpeg" 
    alt="" 
    class="wp-image-12" 
    srcset="
      http://localhost:8888/wp-content/uploads/2024/08/car.jpeg 1080w, 
      http://localhost:8888/wp-content/uploads/2024/08/car-300x169.jpeg 300w, 
      http://localhost:8888/wp-content/uploads/2024/08/car-1024x576.jpeg 1024w, 
      http://localhost:8888/wp-content/uploads/2024/08/car-768x432.jpeg 768w
    "
    sizes="(max-width: 1080px) 100vw, 1080px"
    >
</picture>
```

## No responsive image (use `wp_calculate_image_sizes` filter):
```html
<picture class="wp-picture-12" style="display: contents;">
  <source type="image/webp" srcset="http://localhost:8888/wp-content/uploads/2024/08/car-jpeg.webp">
  <img 
    fetchpriority="high" 
    decoding="async" 
    width="1080" 
    height="608" 
    src="http://localhost:8888/wp-content/uploads/2024/08/car.jpeg" 
    alt="" 
    class="wp-image-12"
  >
</picture>
```

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: Addressed in https://github.com/WordPress/performance/pull/1449/commits/7e09a7231dc04ef3772c0be58fae185dc89a3b7b

plugins/webp-uploads/picture-element.php
mukeshpanchal27: Thanks. Removed in https://github.com/WordPress/performance/pull/1449/commits/7e09a7231dc04ef3772c0be58fae185dc89a3b7b

plugins/webp-uploads/tests/test-picture-element.php
mukeshpanchal27: Thanks. Update the logic per https://github.com/WordPress/performance/issues/1448#issuecomment-2276202488.

plugins/webp-uploads/picture-element.php
```diff
@@ -104,37 +104,55 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
...
+			$img       = webp_uploads_img_tag_update_mime_type( $image, 'the_content', $attachment_id );
+			$processor = new WP_HTML_Tag_Processor( $img );
+			if ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
+				$src = $processor->get_attribute( 'src' );
+
+				$picture_sources .= sprintf(
+					'<source type="%s"%s>',
+					esc_attr( $image_mime_type ),
+					is_string( $src ) ? sprintf( ' srcset="%s"', esc_attr( $src ) ) : ''
+				);
```
westonruter: If there is no `$src` available, then no `SOURCE` tag should be printed at all. As it stands, it could be outputting:

```html
<source type="image/webp">
```

Which clearly won't be helpful 😄 

```suggestion
				$src = $processor->get_attribute( 'src' );
				if ( is_string( $src ) ) {
					$picture_sources .= sprintf(
						'<source type="%s" src="%s">',
						esc_attr( $image_mime_type ),
						esc_attr( $src )
					);
				}
```

This suggestion should also be applied for the previous condition when responsive images are enabled! Namely, instead of:

```php
			if ( false === $image_srcset ) {
				continue;
			}
			$picture_sources .= sprintf(
				'<source type="%s"%s%s>',
				esc_attr( $image_mime_type ),
				is_string( $image_srcset ) ? sprintf( ' srcset="%s"', esc_attr( $image_srcset ) ) : '',
				is_string( $sizes ) ? sprintf( ' sizes="%s"', esc_attr( $sizes ) ) : ''
			);
```

It should be: something like

```php
			if ( is_string( $image_srcset ) ) {
				$picture_sources .= sprintf(
					'<source type="%s" srcset="%s"%s>',
					esc_attr( $image_mime_type ),
					esc_attr( $image_srcset ),
					is_string( $sizes ) ? sprintf( ' sizes="%s"', esc_attr( $sizes ) ) : ''
				);
			}
```

plugins/webp-uploads/picture-element.php
```diff
@@ -104,37 +104,55 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
...
+				is_string( $sizes ) ? sprintf( ' sizes="%s"', esc_attr( $sizes ) ) : ''
+			);
+		}
+	} else {
+		foreach ( $mime_types as $image_mime_type ) {
+			// Updates the reference of the image to the a new image format if available.
+			$img       = webp_uploads_img_tag_update_mime_type( $image, 'the_content', $attachment_id );
+			$processor = new WP_HTML_Tag_Processor( $img );
+			if ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
+				$src = $processor->get_attribute( 'src' );
```
westonruter: Important: It can be that `webp_uploads_img_tag_update_mime_type()` returns the `$image` unchanged here, such as when the attachment isn't available. Therefore, there should be a check to see if `$src` is actually changed, right?

plugins/webp-uploads/picture-element.php
```diff
@@ -104,37 +104,55 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
...
+				'<source type="%s"%s%s>',
+				esc_attr( $image_mime_type ),
+				is_string( $image_srcset ) ? sprintf( ' srcset="%s"', esc_attr( $image_srcset ) ) : '',
+				is_string( $sizes ) ? sprintf( ' sizes="%s"', esc_attr( $sizes ) ) : ''
+			);
+		}
+	} else {
+		foreach ( $mime_types as $image_mime_type ) {
+			// Updates the reference of the image to the a new image format if available.
+			$img       = webp_uploads_img_tag_update_mime_type( $image, 'the_content', $attachment_id );
```
westonruter: While we're looping over `$mime_types`, the current `$image_mime_type` isn't applying in any way to the output of this function. If there are more than one `$mime_types` doesn't that mean that duplicate `SOURCE` tags could be output? Or worse, couldn't it be that a `SOURCE` is generated but the wrong MIME type is added as an attribute?

plugins/webp-uploads/picture-element.php
```diff
@@ -104,37 +104,55 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
...
+			$processor = new WP_HTML_Tag_Processor( $img );
+			if ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
```
westonruter: Do we need to rely on parsing the `src` out with `WP_HTML_Tag_Processor` to obtain the `src`? Wouldn't it be better to iterate over the `wp_get_attachment_metadata()` to get the specific data we're looking for explicitly?

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -260,9 +260,16 @@ function ( string $content ): string {
 		);
 	}
 
-	public function test_disable_responsive_image_with_picture_element(): void {
+	/**
+	 * @dataProvider data_provider_test_disable_responsive_image_with_picture_element
+	 *
+	 * @param Closure $add_filter The filter.
+	 */
+	public function test_disable_responsive_image_with_picture_element( ?Closure $add_filter ): void {
```
westonruter: There is never `null` passed in here, so this should be just `Closure`:

```suggestion
	public function test_disable_responsive_image_with_picture_element( Closure $add_filter ): void {
```

Also, instead of `$add_filter` I think `$set_up` is better.

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -260,9 +260,16 @@ function ( string $content ): string {
...
 		// Disable responsive images.
-		add_filter( 'wp_calculate_image_sizes', '__return_false' );
+		if ( $add_filter instanceof Closure ) {
+			$add_filter();
+		}
```
westonruter: Per above, this is never null:

```suggestion
		$add_filter();
```

plugins/webp-uploads/tests/test-picture-element.php
```diff
@@ -281,16 +288,19 @@ public function test_disable_responsive_image_with_picture_element(): void {
...
+		$this->assertNull( $img_processor->get_attribute( 'srcset' ), 'Make sure that there is no srcset attribute in IMG tag.' );
 
 		// Apply picture element support.
 		$this->opt_in_to_picture_element();
 
 		$picture_markup = apply_filters( 'the_content', $image );
 
 		$picture_processor = new WP_HTML_Tag_Processor( $picture_markup );
-		$picture_processor->next_tag( array( 'tag_name' => 'IMG' ) );
+		$this->assertTrue( $picture_processor->next_tag( array( 'tag_name' => 'picture' ) ), 'Make sure that there is a Picture tag.' );
```
westonruter: For consistency:

```suggestion
		$this->assertTrue( $picture_processor->next_tag( array( 'tag_name' => 'PICTURE' ) ), 'Make sure that there is a PICTURE tag.' );
```

westonruter (changes requested):

plugins/webp-uploads/picture-element.php
mukeshpanchal27: In https://github.com/WordPress/performance/pull/1449/commits/35a626ad56018764809608b15a8ebb181c58f48d introduce new helper function `webp_uploads_get_mime_type_image` that extract the data from `wp_get_attachment_metadata()` in which we leverage the code from [webp_uploads_img_tag_update_mime_type()](https://github.com/WordPress/performance/blob/trunk/plugins/webp-uploads/hooks.php#L591).

plugins/webp-uploads/picture-element.php
mukeshpanchal27: It will return wrong image when we have multiple mime images. Removed it and instead that use new helper function

mukeshpanchal27:
@westonruter Thanks for the feedback. After the new helper function the result of single and multi-mime type images are:

## Multi Mime
`'image/jpeg' => array( 'image/jpeg', 'image/webp', 'image/avif' )`

<details>

<summary>Medium Size</summary>

```html
<picture class="wp-picture-43" style="display: contents;">
  <source 
    type="image/avif"
    srcset="http://localhost:8888/wp-content/uploads/2024/08/car-300x169-jpeg.avif"
  >
  <source
    type="image/webp"
    srcset="http://localhost:8888/wp-content/uploads/2024/08/car-300x169-jpeg.webp"
  >
  <img
    fetchpriority="high"
    decoding="async"
    width="300"
    height="169"
    src="http://localhost:8888/wp-content/uploads/2024/08/car-300x169.jpeg"
    alt=""
    class="wp-image-43"
  >
</picture>
```
</details>
<details>
<summary>Large Size</summary>

```html
<picture class="wp-picture-43" style="display: contents;">
  <source
    type="image/avif"
    srcset="http://localhost:8888/wp-content/uploads/2024/08/car-1024x576-jpeg.avif"
  >
  <source
    type="image/webp"
    srcset="http://localhost:8888/wp-content/uploads/2024/08/car-1024x576-jpeg.webp"
  >
  <img
    decoding="async"
    width="1024" 
    height="576" 
    src="http://localhost:8888/wp-content/uploads/2024/08/car-1024x576.jpeg" 
    alt="" 
    class="wp-image-43"
  >
</picture>
```
</details>
<details>
<summary>Full Size</summary>

```html
<picture class="wp-picture-43" style="display: contents;">
  <source 
    type="image/avif" 
    srcset="http://localhost:8888/wp-content/uploads/2024/08/car-jpeg.avif"
  >
  <source
    type="image/webp"
    srcset="http://localhost:8888/wp-content/uploads/2024/08/car-jpeg.webp"
  >
  <img
    decoding="async" 
    width="1080" 
    height="608" 
    src="http://localhost:8888/wp-content/uploads/2024/08/car.jpeg" 
    alt="" 
    class="wp-image-43"
  >
</picture>
```
</details>

## Single Mime
`'image/jpeg' => array( 'image/jpeg', 'image/avif)`

<details>
<summary>Medium Size</summary>

```html
<picture class="wp-picture-46" style="display: contents;">
  <source
    type="image/avif"
    srcset="http://localhost:8888/wp-content/uploads/2024/08/leaves-300x200-jpg.avif"
  >
  <img 
    loading="lazy" 
    decoding="async"
    width="300" 
    height="200" 
    src="http://localhost:8888/wp-content/uploads/2024/08/leaves-300x200.jpg" 
    alt="" 
    class="wp-image-46"
  >
</picture>
```
</details>
<details>
<summary>Large Size</summary>

```html
<picture class="wp-picture-46" style="display: contents;">
  <source 
    type="image/avif" 
    srcset="http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683-jpg.avif"
  >
  <img 
    loading="lazy" 
    decoding="async" 
    width="1024" 
    height="683" 
    src="http://localhost:8888/wp-content/uploads/2024/08/leaves-1024x683.jpg" 
    alt="" 
    class="wp-image-46"
  >
</picture>
```
</details>
<details>
<summary>Full Size</summary>

```html
<picture class="wp-picture-46" style="display: contents;">
  <source
    type="image/avif"
    srcset="http://localhost:8888/wp-content/uploads/2024/08/leaves-jpg.avif"
  >
  <img
    loading="lazy" 
    decoding="async" 
    width="1080" 
    height="720" 
    src="http://localhost:8888/wp-content/uploads/2024/08/leaves.jpg" 
    alt="" 
    class="wp-image-46"
  >
</picture>
```
</details>

plugins/webp-uploads/helper.php
```diff
@@ -406,3 +406,61 @@ function webp_uploads_is_picture_element_enabled(): bool {
...
+ * attachment's metadata that matches the provided MIME type.
+ *
+ * @since n.e.x.t
+ *
+ * @param int    $attachment_id The ID of the attachment.
+ * @param string $src           The original image src url.
+ * @param string $mime          A mime type we are looking to get image url.
+ * @return string|false Returns mime type image if available, false otherwise.
+ */
+function webp_uploads_get_mime_type_image( int $attachment_id, string $src, string $mime ) {
```
westonruter: This seems a bit better to return null instead of false:

```suggestion
function webp_uploads_get_mime_type_image( int $attachment_id, string $src, string $mime ): ?string {
```

plugins/webp-uploads/helper.php
```diff
@@ -406,3 +406,61 @@ function webp_uploads_is_picture_element_enabled(): bool {
...
+
+			return str_replace(
+				$size_data['file'],
+				$size_data['sources'][ $mime ]['file'],
+				$src
+			);
+		}
+	}
+
+	return false;
```
```suggestion
	return null;
```

plugins/webp-uploads/helper.php
```diff
@@ -406,3 +406,61 @@ function webp_uploads_is_picture_element_enabled(): bool {
...
+ *
+ * This function attempts to locate an alternate image source URL in the
+ * attachment's metadata that matches the provided MIME type.
+ *
+ * @since n.e.x.t
+ *
+ * @param int    $attachment_id The ID of the attachment.
+ * @param string $src           The original image src url.
+ * @param string $mime          A mime type we are looking to get image url.
+ * @return string|false Returns mime type image if available, false otherwise.
```
```suggestion
 * @return string|null Returns mime type image if available.
```

plugins/webp-uploads/helper.php
```diff
@@ -406,3 +406,61 @@ function webp_uploads_is_picture_element_enabled(): bool {
...
+				$metadata['sources'][ $mime ]['file'],
+				$src
+			);
+		}
+	}
+
+	if ( isset( $metadata['sizes'] ) && is_array( $metadata['sizes'] ) ) {
+		foreach ( $metadata['sizes'] as $size => $size_data ) {
+
+			if ( empty( $size_data['file'] ) ) {
```
westonruter: Can we avoid using empty() here?

plugins/webp-uploads/helper.php
```diff
@@ -406,3 +406,61 @@ function webp_uploads_is_picture_element_enabled(): bool {
...
+ * @param string $mime          A mime type we are looking to get image url.
+ * @return string|false Returns mime type image if available, false otherwise.
+ */
+function webp_uploads_get_mime_type_image( int $attachment_id, string $src, string $mime ) {
+	$metadata     = wp_get_attachment_metadata( $attachment_id );
+	$src_basename = wp_basename( $src );
+	if ( isset( $metadata['sources'][ $mime ]['file'] ) ) {
+		$basename = wp_basename( $metadata['file'] );
+
+		if ( $src_basename === $basename ) {
```
westonruter, line 429: Will this condition ever currently be true given that we still are appending the original extension to the name? For example, `foo.jpg` gets converted to `foo-jpg.avif`. See https://github.com/WordPress/performance/issues/545.

Also, the second parameter isn't passed to `wp_basename()` so the extension isn't getting stripped.

westonruter (commented):
Curious: in https://github.com/WordPress/performance/pull/1449#issuecomment-2285810613 you show examples of how this works when different image sizes are requested, but how is the size being selected for the alternative mime types? I don't see how the size selection works.

plugins/webp-uploads/helper.php
mukeshpanchal27: @westonruter No. When we upload JPEG image the function gets the `foo.jpg` not `foo-jpg.avif`.

You can check it through:
- Add `add_filter( 'wp_calculate_image_sizes', '__return_false' );` in functions.php
- Upload JPEG image, Make sure to enabled the fallback JPEG if not then it will return IMG per #1447
- This function get the JPEG image and it return it's alternate mime type
- Use Full size image then this  condition passes. Like. `foo.jpg === foo.jpg`

<details>

<summary>Image Metadata</summary>

```php
Array
(
    [width] => 1080
    [height] => 608
    [file] => 2024/08/car.jpeg
    [filesize] => 150444
    [sizes] => Array
        (
            [medium] => Array
                (
                    [file] => car-300x169.jpeg
                    [width] => 300
                    [height] => 169
                    [mime-type] => image/jpeg
                    [filesize] => 12915
                    [sources] => Array
                        (
                            [image/jpeg] => Array
                                (
                                    [file] => car-300x169.jpeg
                                    [filesize] => 12915
                                )

                            [image/webp] => Array
                                (
                                    [file] => car-300x169-jpeg.webp
                                    [filesize] => 9580
                                )

                        )

                )

            [large] => Array
                (
                    [file] => car-1024x576.jpeg
                    [width] => 1024
                    [height] => 576
                    [mime-type] => image/jpeg
                    [filesize] => 110947
                    [sources] => Array
                        (
                            [image/jpeg] => Array
                                (
                                    [file] => car-1024x576.jpeg
                                    [filesize] => 110947
                                )

                            [image/webp] => Array
                                (
                                    [file] => car-1024x576-jpeg.webp
                                    [filesize] => 80386
                                )

                        )

                )

            [thumbnail] => Array
                (
                    [file] => car-150x150.jpeg
                    [width] => 150
                    [height] => 150
                    [mime-type] => image/jpeg
                    [filesize] => 6342
                    [sources] => Array
                        (
                            [image/jpeg] => Array
                                (
                                    [file] => car-150x150.jpeg
                                    [filesize] => 6342
                                )

                            [image/webp] => Array
                                (
                                    [file] => car-150x150-jpeg.webp
                                    [filesize] => 4834
                                )

                        )

                )

            [medium_large] => Array
                (
                    [file] => car-768x432.jpeg
                    [width] => 768
                    [height] => 432
                    [mime-type] => image/jpeg
                    [filesize] => 65538
                    [sources] => Array
                        (
                            [image/jpeg] => Array
                                (
                                    [file] => car-768x432.jpeg
                                    [filesize] => 65538
                                )

                            [image/webp] => Array
                                (
                                    [file] => car-768x432-jpeg.webp
                                    [filesize] => 48900
                                )

                        )

                )

        )

    [image_meta] => Array
        (
            [aperture] => 0
            [credit] => 
            [camera] => 
            [caption] => 
            [created_timestamp] => 0
            [copyright] => 
            [focal_length] => 0
            [iso] => 0
            [shutter_speed] => 0
            [title] => 
            [orientation] => 0
            [keywords] => Array
                (
                )

        )

    [dominant_color] => 575757
    [has_transparency] => 
    [sources] => Array
        (
            [image/jpeg] => Array
                (
                    [file] => car.jpeg
                    [filesize] => 150444
                )

            [image/webp] => Array
                (
                    [file] => car-jpeg.webp
                    [filesize] => 109568
                )

        )

)
```
</details>

> Also, the second parameter isn't passed to wp_basename() so the extension isn't getting stripped.

We can't second parameter as we don't know which image gets uploaded. In future we are adding PNG to WebP support. We leverage the code from [webp_uploads_img_tag_update_mime_type()](https://github.com/WordPress/performance/blob/trunk/plugins/webp-uploads/hooks.php#L591) in which we don't use second parameter.

mukeshpanchal27:
> Curious: in [#1449 (comment)](https://github.com/WordPress/performance/pull/1449#issuecomment-2285810613) you show examples of how this works when different image sizes are requested, but how is the size being selected for the alternative mime types? I don't see how the size selection works.

@westonruter When we disabled the responsive images then it will return only `src` for the image with size( Large, Medium , Full ) that we selected. https://github.com/WordPress/performance/pull/1449#issuecomment-2285810613 i share the result for each image size.

westonruter:
I'm not actively familiar enough with the internals of image handling in WP without taking time to dive in deeper. I'll defer to @joemcgill and @adamsilverstein who have more experience to finish the reviews.

westonruter:
Punting from the 3.4.0 release since we're out of time. If this does get reviewed by Monday morning, make sure the PR's base branch is updated from `trunk` to `release/3.4.0`.

adamsilverstein (approved):
Nice fix!

felixarntz:
IMO too big of a change to merge now 1 hour before release. I'd say let's get this into the next one.

westonruter:
Agreed. Merging into trunk and not into the release branch.

westonruter (approved):

westonruter:
@mukeshpanchal27 I updated the PR title to be more clear. Please review and amend further as needed.

PR #1451, opened 2024-08-08
Use 0.1.0 for initial version instead of n.e.x.t for Web Worker Offloading

westonruter:
When attempting to run `npm run build-plugins:zip` currently, I get an error:

```
[webpack-cli] Error: Plugin version not found for "web-worker-offloading".
```

This is because Web Worker Offloading is currently using `n.e.x.t` for all of its versions. Since it hasn't been released yet, I think it makes more sense to use `0.1.0` for the version until we get it deployed to WordPress.org.

### DIFF
--- a/plugins/web-worker-offloading/hooks.php
+++ b/plugins/web-worker-offloading/hooks.php
@@ -2,7 +2,7 @@
 /**
  * Hook callback for Web Worker Offloading.
  *
- * @since n.e.x.t
+ * @since 0.1.0
  * @package web-worker-offloading
  */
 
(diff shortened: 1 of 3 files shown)


plugins/web-worker-offloading/hooks.php
```diff
@@ -2,7 +2,7 @@
 /**
  * Hook callback for Web Worker Offloading.
  *
- * @since n.e.x.t
+ * @since 0.1.0
```
adamsilverstein, line 5: can we update this to 1.0.0 when we release? I like starting at 1.

adamsilverstein (approved):
specific

plugins/web-worker-offloading/hooks.php
westonruter: I think `0.1.0` is good specifically because this is an experimental plugin and other experimental plugins (except for `auto-sizes`) are also starting at 0.1.0.

thelovekesh (approved):

mukeshpanchal27 (approved):

Issue #1440, opened 2024-08-06
Prepare PL 3.4.0 release


sstopfer:
This issue is to track preparation of the upcoming 3.4.0 release up until publishing, which is due August 19, 2024.

- [x]  Create release/3.4.0 branch closer to the release date
- [x]  Finalize scope and punt unfinished pull requests to following release
- [x]  Smoke test (Friday, August 16, 2024)
- [x]  Prepare the release (Monday, August 19, 2024)

felixarntz:
@westonruter @thelovekesh Is the plan to release the Web Worker Offloading plugin as part of this release cycle? I noticed [it's already in `plugins.json`](https://github.com/WordPress/performance/blob/trunk/plugins.json#L10), but to my knowledge we don't have a .org repo for it yet.

If we're not ready to publish it yet, shouldn't it be removed from `plugins.json` for now, to avoid triggering a release (attempt) for it with PL 3.4.0?

westonruter:
That's right. The plan was to release it as another experimental plugin, however if we're not satisfied with the current state then we'll have to remove from `plugins.json` prior to doing a PL release.

felixarntz:
Sounds good. I'd be in favor of removing it now and only adding once we know we want to and _can_ (i.e. have a .org repo) publish it.

westonruter:
I've filed https://github.com/WordPress/performance/issues/1455 to extend Web Worker Offloading with support for automatically offloading Google Tag Manager for at least WooCommerce and Site Kit.

westonruter:
I've created the `release/3.4.0` branch and I opened #1480 with the version bumps and readme updates. See https://github.com/WordPress/performance/pull/1480#issuecomment-2294465381 for build ZIPs for smoke testing.

mukeshpanchal27:
Everything looks good from what I can see for the smoke test for version 3.4.0 through the `release/3.4.0` branch. In this release, the `picture element` received some great improvements.

Enhanced Responsive Images
- Use more robust HTML Tag Processor for auto sizes injection ✅
- Remove sizes attribute when the responsive image disabled ✅
- Harden logic to add `auto` keyword to `sizes` attribute to prevent duplicate keyword ✅

Performance Lab
- Remove Server-Timing metric for the autoloaded options query time ✅
- Fix disabled options from reappearing in Site Health after external update ✅
- Improve Performance screen when external requests to WordPress.org fail ✅

Modern Image Formats
[...]

PR #1458, opened 2024-08-12
Remove `web-worker-offloading` from plugins.json for now since it is not yet ready for release

felixarntz:
## Summary

See #1440: The next set of Performance Lab plugin releases will happen next Monday, August 19.

We don't have a .org repo for `web-worker-offloading` yet, and haven't decided whether we want to release it already or what else we want to have in the plugin for a first version.

## Relevant technical choices

* Remove `web-worker-offloading` from `plugins.json` so that we don't attempt any release processes for it. It can be re-added later once we are ready to release a first version.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -7,7 +7,6 @@
     "optimization-detective",
     "performance-lab",
     "speculation-rules",
-    "web-worker-offloading",
     "webp-uploads"
   ]
 }


westonruter (approved):

westonruter:
Reverting this change as part of https://github.com/WordPress/performance/pull/1563

PR #1469, opened 2024-08-13
Update tag visitor registration with unique ID constants

### LINKED ISSUE #1419: Dependencies for tag visitors in Optimization Detective
In at least one case, there is a need currently for a tag visitor to run after another tag visitor has run: 

https://github.com/WordPress/performance/blob/e34711dbf286a16523b3c3d35462e0e79ea4c6f9/plugins/auto-sizes/optimization-detective.php#L54-L66

The [tag visitor](https://github.com/WordPress/performance/blob/e34711dbf286a16523b3c3d35462e0e79ea4c6f9/plugins/auto-sizes/optimization-detective.php#L13-L52) in Enhanced Responsive Sizes needs to run after the [IMG tag visitor](https://github.com/WordPress/performance/blob/trunk/plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php) be

The Image Prioritizer's tag visitor is currently registered with the name ID of `img-tags` which is not ideal:

https://github.com/WordPress/performance/blob/e34711dbf286a16523b3c3d35462e0e79ea4c6f9/plugins/image-prioritizer/helper.php#L33-L35

It should rather be prefixed like `image-prioritizer-img`. The `image-prioritizer-img` string could even be defined as a class constant like `Image_Prioritizer_Img_Tag_Visitor::ID`. When Enhanced Responsive Sizes registers its tag visitor, it should be able to explicitly declare this dependency:

```php
function auto_sizes_register_tag_visitors( OD_Tag_Visitor_Registry $registry ): void { 
	$dependencies = array();
	if ( class_exists( 'Image_Prioritizer_Img_Tag_Visitor' ) ) {
		$dependencies[] = Image_Prioritizer_Img_Tag_Visitor::ID;
	}
	$registry->register( 'auto-sizes', 'auto_sizes_visit_tag', $dependencies ); 
} 
```

Then there wouldn't be a need to rely on action priorities (and the current implementation of `OD_Tag_Visitor_Registry`) to ensure that one tag visitor runs before another.

pbearne:
Replaced hardcoded ID strings with unique ID constants for tag visitors to ensure consistency and maintainability. Updated the `OD_Tag_Visitor_Registry::register` method to handle dependencies, adding them only if they are not already registered.

## Summary

Fixes #1419

### DIFF
--- a/plugins/auto-sizes/optimization-detective.php
+++ b/plugins/auto-sizes/optimization-detective.php
@@ -59,8 +59,12 @@ function auto_sizes_visit_tag( OD_Tag_Visitor_Context $context ): bool {
  * @param OD_Tag_Visitor_Registry $registry Tag visitor registry.
  */
 function auto_sizes_register_tag_visitors( OD_Tag_Visitor_Registry $registry ): void {
-	$registry->register( 'auto-sizes', 'auto_sizes_visit_tag' );
+	$registry->register(
+		'auto-sizes',
+		'auto_sizes_visit_tag',
+		defined( 'Image_Prioritizer_Img_Tag_Visitor::ID' ) ? array( Image_Prioritizer_Img_Tag_Visitor::ID => new Image_Prioritizer_Img_Tag_Visitor() ) : array()
+	);
 }
 
 // Important: The Image Prioritizer's IMG tag visitor is registered at priority 10, so priority 100 ensures that the loading attribute has been correctly set by the time the Auto Sizes visitor runs.
-add_action( 'od_register_tag_visitors', 'auto_sizes_register_tag_visitors', 100 );
+add_action( 'od_register_tag_visitors', 'auto_sizes_register_tag_visitors' );
--- a/plugins/auto-sizes/optimization-detective.php
+++ b/plugins/auto-sizes/optimization-detective.php
@@ -62,7 +62,7 @@ function auto_sizes_register_tag_visitors( OD_Tag_Visitor_Registry $registry ): 
 	$registry->register(
 		'auto-sizes',
 		'auto_sizes_visit_tag',
-		defined( 'Image_Prioritizer_Img_Tag_Visitor::ID' ) ? array( Image_Prioritizer_Img_Tag_Visitor::ID => new Image_Prioritizer_Img_Tag_Visitor() ) : array()
+		defined( 'Image_Prioritizer_Img_Tag_Visitor::ID' ) ? array( Image_Prioritizer_Img_Tag_Visitor::ID ) : array()
 	);
 }
 
--- a/plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
@@ -20,6 +20,11 @@ if ( ! defined( 'ABSPATH' ) ) {
 final class Image_Prioritizer_Img_Tag_Visitor extends Image_Prioritizer_Tag_Visitor {
 
 	/**
+	 * The ID used the register the class
+	 */
+	public const ID = 'image-prioritizer-img';
+
+	/**
 	 * Visits a tag.
 	 *
 	 * @param OD_Tag_Visitor_Context $context Tag visitor context.
--- a/plugins/image-prioritizer/helper.php
+++ b/plugins/image-prioritizer/helper.php
@@ -32,8 +32,8 @@ function image_prioritizer_render_generator_meta_tag(): void {
 function image_prioritizer_register_tag_visitors( OD_Tag_Visitor_Registry $registry ): void {
 	// Note: The class is invocable (it has an __invoke() method).
 	$img_visitor = new Image_Prioritizer_Img_Tag_Visitor();
-	$registry->register( 'img-tags', $img_visitor );
+	$registry->register( $img_visitor::ID, $img_visitor );
 
 	$bg_image_visitor = new Image_Prioritizer_Background_Image_Styled_Tag_Visitor();
-	$registry->register( 'bg-image-tags', $bg_image_visitor );
+	$registry->register( $bg_image_visitor::ID, $bg_image_visitor );
 }
--- a/plugins/image-prioritizer/helper.php
+++ b/plugins/image-prioritizer/helper.php
@@ -35,5 +35,5 @@ function image_prioritizer_register_tag_visitors( OD_Tag_Visitor_Registry $regis
 	$registry->register( Image_Prioritizer_Img_Tag_Visitor::ID, $img_visitor );
 
 	$bg_image_visitor = new Image_Prioritizer_Background_Image_Styled_Tag_Visitor();
-	$registry->register( $bg_image_visitor::ID, $bg_image_visitor );
+	$registry->register( Image_Prioritizer_Background_Image_Styled_Tag_Visitor::ID, $bg_image_visitor );
 }
--- a/plugins/image-prioritizer/helper.php
+++ b/plugins/image-prioritizer/helper.php
@@ -32,7 +32,7 @@ function image_prioritizer_render_generator_meta_tag(): void {
 function image_prioritizer_register_tag_visitors( OD_Tag_Visitor_Registry $registry ): void {
 	// Note: The class is invocable (it has an __invoke() method).
 	$img_visitor = new Image_Prioritizer_Img_Tag_Visitor();
-	$registry->register( $img_visitor::ID, $img_visitor );
+	$registry->register( Image_Prioritizer_Img_Tag_Visitor::ID, $img_visitor );
 
 	$bg_image_visitor = new Image_Prioritizer_Background_Image_Styled_Tag_Visitor();
 	$registry->register( $bg_image_visitor::ID, $bg_image_visitor );
--- a/plugins/optimization-detective/class-od-tag-visitor-registry.php
+++ b/plugins/optimization-detective/class-od-tag-visitor-registry.php
@@ -31,6 +31,13 @@ final class OD_Tag_Visitor_Registry implements Countable, IteratorAggregate {
 	private $visitors = array();
 
 	/**
+	 * sorted_Visitors.
+	 *
+	 * @var array<string, TagVisitorCallback>
+	 */
+	private $sorted_visitors = array();
+
+	/**
 	 * Registers a tag visitor.
 	 *
 	 * @phpstan-param TagVisitorCallback $tag_visitor_callback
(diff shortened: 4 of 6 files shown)


plugins/optimization-detective/class-od-tag-visitor-registry.php
```diff
@@ -37,9 +37,20 @@ final class OD_Tag_Visitor_Registry implements Countable, IteratorAggregate {
 	 *
 	 * @param string   $id                   Identifier for the tag visitor.
 	 * @param callable $tag_visitor_callback Tag visitor callback.
+	 * @param array<mixed> $dependencies array with ID as key and Callable as value
```
westonruter:
```suggestion
	 * @param string[] $dependencies Tag visitors that must run before this tag visitor.
```

plugins/optimization-detective/class-od-tag-visitor-registry.php
```diff
@@ -37,9 +37,20 @@ final class OD_Tag_Visitor_Registry implements Countable, IteratorAggregate {
...
+			foreach ($dependencies as $dependency_id => $dependency_callback) {
+				if (!self::is_registered($dependency_id)) {
+					$this->visitors[$dependency_id] = $dependency_callback;
+				}
+			}
+		}
+		if (!self::is_registered($id)) {
+			$this->visitors[$id] = $tag_visitor_callback;
+		}
 	}
```
westonruter: I was thinking rather that this would work more like how `WP_Scripts` works, where you can register a script (or tag visitor) with a given handle (or ID) and supply dependencies which are a list of strings (i.e. handles/IDs). Then later when obtaining the list of tag visitors to apply, it would recursively look up the registered tag visitors to make sure that those which have no dependencies run first, followed by those which have dependencies.

plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
```diff
@@ -19,6 +19,11 @@
  */
 final class Image_Prioritizer_Img_Tag_Visitor extends Image_Prioritizer_Tag_Visitor {
 
+	/**
+	 * The ID used the register the class
+	 */
+	public const ID = 'image-prioritizer-img';
```
westonruter, line 25: This would also make sense to add to `Embed_Optimizer_Tag_Visitor`.

westonruter (changes requested):
Thanks for getting this off the ground!

See my inline comments.

The ultimate goal would be to refactor this code:

https://github.com/WordPress/performance/blob/73f223598df758e507dcfd14e09681e156a850d2/plugins/auto-sizes/optimization-detective.php#L61-L66

So we could eliminate the need for the `100` priority.

For example:

```php
function auto_sizes_register_tag_visitors( OD_Tag_Visitor_Registry $registry ): void {
	$registry->register( 
		'auto-sizes', 
		'auto_sizes_visit_tag', 
		array( 'image-prioritizer-img' ) 
	);
}
add_action( 'od_register_tag_visitors', 'auto_sizes_register_tag_visitors' );
```

An option question is what to do if the `image-prioritizer-img` tag visitor dependency doesn't exist. Does the `auto-sizes` then never run? Or would it need to rather do something like this:

```php
function auto_sizes_register_tag_visitors( OD_Tag_Visitor_Registry $registry ): void {
	$registry->register( 
		'auto-sizes', 
		'auto_sizes_visit_tag', 
		defined( 'Image_Prioritizer_Img_Tag_Visitor::ID' ) ? array( Image_Prioritizer_Img_Tag_Visitor::ID ) : array()
	);
}
add_action( 'od_register_tag_visitors', 'auto_sizes_register_tag_visitors' );
```

plugins/optimization-detective/class-od-tag-visitor-registry.php
pbearne: I feel this should be any so we can have more than one dependency

westonruter: Yes, and that's still the case: `string[]` is an array of strings.

pbearne: changed

plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
pbearne: done

westonruter: Oh yeah, I don't see any change to the `Embed_Optimizer_Tag_Visitor` class in the Embed Optimizer plugin?

plugins/optimization-detective/class-od-tag-visitor-registry.php
westonruter: I don't see the change.

plugins/optimization-detective/class-od-tag-visitor-registry.php
pbearne: I can see a way of changing this to just passing the className and the dependencies as a series of class names
and then getting the ID from the class 

Or just building an array values and then load them in the first call to get_registered

Part of the problem is I am not sure what the code is doing

westonruter: The tag visitor registry is a collection of callables which are invoked when iterating over the tags via HTML Tag Processor. A tag visitor may be an invokable class or it may be a closure or a function or any other `callable`. See the Enhanced Responsive Images for an example tag visitor which is a regular PHP function and not a class:

https://github.com/WordPress/performance/blob/bc37748d38c6e696f085f8ffedb19389d3198151/plugins/auto-sizes/optimization-detective.php#L13-L52

So in the same way as with `WP_Scripts` where registered scripts have handles, here too tag visitors (callables) have IDs. As with script dependencies being able to register a script which depends on another script by passing an array of dependency handles, so too as seen in Enhanced Responsive Images there should be a way to register a tag visitor so that it runs after other tag visitors have done.

plugins/auto-sizes/optimization-detective.php
```diff
@@ -59,8 +59,12 @@ function auto_sizes_visit_tag( OD_Tag_Visitor_Context $context ): bool {
  * @param OD_Tag_Visitor_Registry $registry Tag visitor registry.
  */
 function auto_sizes_register_tag_visitors( OD_Tag_Visitor_Registry $registry ): void {
-	$registry->register( 'auto-sizes', 'auto_sizes_visit_tag' );
+	$registry->register(
+		'auto-sizes',
+		'auto_sizes_visit_tag',
+		defined( 'Image_Prioritizer_Img_Tag_Visitor::ID' ) ? array( Image_Prioritizer_Img_Tag_Visitor::ID => new Image_Prioritizer_Img_Tag_Visitor() ) : array()
```
westonruter: Only the ID should be passed here:

```suggestion
		defined( 'Image_Prioritizer_Img_Tag_Visitor::ID' ) ? array( Image_Prioritizer_Img_Tag_Visitor::ID ) : array()
```

Registering the `Image_Prioritizer_Img_Tag_Visitor` tag visitor is not the responsibility of the Enhanced Responsive Images plugin, but rather the Image Prioritizer plugin here:

https://github.com/WordPress/performance/blob/16b896d533fa865989f481333041669ab46b8986/plugins/image-prioritizer/helper.php#L32-L39

plugins/image-prioritizer/helper.php
```diff
@@ -32,8 +32,8 @@ function image_prioritizer_render_generator_meta_tag(): void {
 function image_prioritizer_register_tag_visitors( OD_Tag_Visitor_Registry $registry ): void {
 	// Note: The class is invocable (it has an __invoke() method).
 	$img_visitor = new Image_Prioritizer_Img_Tag_Visitor();
-	$registry->register( 'img-tags', $img_visitor );
+	$registry->register( $img_visitor::ID, $img_visitor );
```
westonruter: Any reason to reference the constant off of the instance rather than from the class?

```suggestion
	$registry->register( Image_Prioritizer_Img_Tag_Visitor::ID, $img_visitor );
```

plugins/image-prioritizer/helper.php
```diff
@@ -32,8 +32,8 @@ function image_prioritizer_render_generator_meta_tag(): void {
...
 
 	$bg_image_visitor = new Image_Prioritizer_Background_Image_Styled_Tag_Visitor();
-	$registry->register( 'bg-image-tags', $bg_image_visitor );
+	$registry->register( $bg_image_visitor::ID, $bg_image_visitor );
```
westonruter:
```suggestion
	$registry->register( Image_Prioritizer_Background_Image_Styled_Tag_Visitor::ID, $bg_image_visitor );
```

plugins/optimization-detective/class-od-tag-visitor-registry.php
```diff
@@ -35,11 +35,21 @@ final class OD_Tag_Visitor_Registry implements Countable, IteratorAggregate {
...
+		if ( ! empty( $dependencies ) ) {
+			foreach ( $dependencies as $dependency_id => $dependency_callback ) {
+				if ( ! self::is_registered( $dependency_id ) ) {
+					$this->register( $dependency_id, $dependency_callback );
+				}
+			}
+		}
+		if ( ! self::is_registered( $id ) ) {
+			$this->visitors[ $id ] = $tag_visitor_callback;
+		}
```
westonruter: I think this should go back to how it was before, but refactor the `visitors` member variable so that it is not just a mapping of ID to callable, but rather a mapping of ID to an array of information about the callable, for example:

```suggestion
		$this->visitors[ $id ] = array(
			'callback'     => $tag_visitor_callback,
			'dependencies' => $dependencies,
		);
```

Then, instead of implementing the dependency handling here, it can be implemented in the `getIterator()` method:

https://github.com/WordPress/performance/blob/16b896d533fa865989f481333041669ab46b8986/plugins/optimization-detective/class-od-tag-visitor-registry.php#L92-L99

This method would end up looking a lot like `WP_Scripts::all_deps()` in core (but much much simpler) which recursively goes through the dependencies to ensure the right order in the end.

If there is a tag visitor that lacks a dependency (recursively), then it would get omitted in the same way as scripts get omitted when one of their dependencies is not registered. It may be appropriate to include a `_doing_it_wrong()` in this case as well.

westonruter (changes requested):

plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
pbearne: @westonruter I am not sure what is going on here you might need to take this over

mukeshpanchal27:
See https://github.com/WordPress/performance/issues/1419#issuecomment-2551751841

PR #1470, opened 2024-08-14
Improve disabling checkbox for Picture Element on Media settings screen

mukeshpanchal27:
## Summary

In #1273, we improved the Settings -> Media controls for Modern Image Formats, where the `Use <picture> Element (experimental)` option is disabled if the fallback JPEG checkbox is not selected. However, the checkbox remains clickable even when disabled because we only added the disabled class without actually adding the `disabled` attribute to the checkbox.

This PR provides a quick workaround by disabling pointer events for the checkbox when the disabled class is present.

https://github.com/user-attachments/assets/cd2d4792-46a7-4cbd-be8b-8db94a12227f

### DIFF
--- a/plugins/webp-uploads/settings.php
+++ b/plugins/webp-uploads/settings.php
@@ -214,7 +214,7 @@ function webp_uploads_use_picture_element_callback(): void {
 				type="checkbox"
 				id="webp_uploads_use_picture_element"
 				aria-describedby="webp_uploads_use_picture_element_description"
-				<?php checked( get_option( 'webp_uploads_use_picture_element', false ) ); // Option intentionally used instead of webp_uploads_is_picture_element_enabled() to persist when perflab_generate_webp_and_jpeg is updated. ?>
+				<?php checked( $picture_element_option ); // Option intentionally used instead of webp_uploads_is_picture_element_enabled() to persist when perflab_generate_webp_and_jpeg is updated. ?>
 				<?php disabled( ! $jpeg_fallback_enabled ); ?>
 				onchange="document.getElementById('webp_uploads_use_picture_element_value').value = this.checked ? 1 : 0"
 			>
--- a/plugins/webp-uploads/settings.php
+++ b/plugins/webp-uploads/settings.php
@@ -185,8 +185,9 @@ function webp_uploads_generate_webp_jpeg_setting_callback(): void {
  */
 function webp_uploads_use_picture_element_callback(): void {
 	// Picture element support requires the JPEG output to be enabled.
-	$picture_element_option = 1 === (int) get_option( 'webp_uploads_use_picture_element', 0 );
-	$jpeg_fallback_enabled  = webp_uploads_is_jpeg_fallback_enabled();
+	$picture_element_option    = 1 === (int) get_option( 'webp_uploads_use_picture_element', 0 );
+	$jpeg_fallback_enabled     = webp_uploads_is_jpeg_fallback_enabled();
+	$picture_element_hidden_id = 'webp_uploads_use_picture_element_hidden';
 	?>
 	<style>
 		#webp_uploads_picture_element_fieldset.disabled label,
@@ -217,7 +218,7 @@ function webp_uploads_use_picture_element_callback(): void {
 				?>
 				<input
 					type="hidden"
-					id="webp_uploads_use_picture_element_hidden"
+					id="<?php echo esc_attr( $picture_element_hidden_id ); ?>"
 					name="webp_uploads_use_picture_element"
 					value="1"
 				>
@@ -233,6 +234,7 @@ function webp_uploads_use_picture_element_callback(): void {
 		</div>
 	</div>
 	<script>
+	( function ( pictureElementHiddenId ) {
 		document.getElementById( 'webp_uploads_use_picture_element' ).addEventListener( 'change', function () {
 			document.getElementById( 'webp_uploads_jpeg_fallback_notice' ).hidden = ! this.checked;
 		} );
@@ -249,20 +251,21 @@ function webp_uploads_use_picture_element_callback(): void {
 
 			// Remove or inject hidden input to preserve original setting value as needed.
 			if ( this.checked ) {
-				const hiddenInput = document.getElementById( 'webp_uploads_use_picture_element_hidden' );
+				const hiddenInput = document.getElementById( pictureElementHiddenId );
 				if ( hiddenInput ) {
 					hiddenInput.parentElement.removeChild( hiddenInput );
 				}
-			} else if ( checkbox.checked && ! document.getElementById( 'webp_uploads_use_picture_element_hidden' ) ) {
+			} else if ( checkbox.checked && ! document.getElementById( pictureElementHiddenId ) ) {
 				// The hidden input is only needed if the value was originally set (i.e. the checkbox enabled).
 				const hiddenInput = document.createElement( 'input' );
 				hiddenInput.setAttribute( 'type', 'hidden' );
-				hiddenInput.setAttribute( 'id', 'webp_uploads_use_picture_element_hidden' );
+				hiddenInput.setAttribute( 'id', pictureElementHiddenId );
 				hiddenInput.setAttribute( 'name', checkbox.getAttribute( 'name' ) );
 				hiddenInput.setAttribute( 'value', checkbox.getAttribute( 'value' ) );
 				checkbox.parentElement.insertBefore( hiddenInput, checkbox.nextSibling );
 			}
 		} );
+	} )( <?php echo wp_json_encode( $picture_element_hidden_id ); ?> );
 	</script>
 	<?php
 }
--- a/plugins/webp-uploads/settings.php
+++ b/plugins/webp-uploads/settings.php
@@ -244,18 +244,18 @@ function webp_uploads_use_picture_element_callback(): void {
 			document.getElementById( 'webp_uploads_picture_element_notice' ).hidden = this.checked;
 			document.getElementById( 'webp_uploads_picture_element_fieldset' ).classList.toggle( 'disabled', ! this.checked );
 
-			var checkbox = document.getElementById( 'webp_uploads_use_picture_element' );
+			const checkbox = document.getElementById( 'webp_uploads_use_picture_element' );
 			checkbox.disabled = ! this.checked;
 
 			// Remove or inject hidden input to preserve original setting value as needed.
 			if ( this.checked ) {
-				var hiddenInput = document.getElementById( 'webp_uploads_use_picture_element_hidden' );
+				const hiddenInput = document.getElementById( 'webp_uploads_use_picture_element_hidden' );
 				if ( hiddenInput ) {
 					hiddenInput.parentElement.removeChild( hiddenInput );
 				}
 			} else if ( checkbox.checked && ! document.getElementById( 'webp_uploads_use_picture_element_hidden' ) ) {
 				// The hidden input is only needed if the value was originally set (i.e. the checkbox enabled).
-				var hiddenInput = document.createElement( 'input' );
+				const hiddenInput = document.createElement( 'input' );
 				hiddenInput.setAttribute( 'type', 'hidden' );
 				hiddenInput.setAttribute( 'id', 'webp_uploads_use_picture_element_hidden' );
 				hiddenInput.setAttribute( 'name', checkbox.getAttribute( 'name' ) );
--- a/plugins/webp-uploads/settings.php
+++ b/plugins/webp-uploads/settings.php
@@ -182,8 +182,26 @@ function webp_uploads_generate_webp_jpeg_setting_callback(): void {
 			document.getElementById( 'perflab_generate_webp_and_jpeg' ).addEventListener( 'change', function () {
 				document.querySelector( '.webp-uploads-use-picture-element' ).classList.toggle( 'webp-uploads-disabled', ! this.checked );
 				document.getElementById( 'webp_uploads_picture_element_notice' ).hidden = this.checked;
-				document.getElementById( 'webp_uploads_use_picture_element' ).disabled = ! this.checked;
 				document.getElementById( 'webp_uploads_picture_element_fieldset' ).classList.toggle( 'disabled', ! this.checked );
+
+				var checkbox = document.getElementById( 'webp_uploads_use_picture_element' );
+				checkbox.disabled = ! this.checked;
+
+				// Remove or inject hidden input to preserve original setting value as needed.
+				if ( this.checked ) {
+					var hiddenInput = document.getElementById( 'webp_uploads_use_picture_element_hidden' );
+					if ( hiddenInput ) {
+						hiddenInput.parentElement.removeChild( hiddenInput );
+					}
+				} else if ( checkbox.checked && ! document.getElementById( 'webp_uploads_use_picture_element_hidden' ) ) {
+					// The hidden input is only needed if the value was originally set (i.e. the checkbox enabled).
+					var hiddenInput = document.createElement( 'input' );
+					hiddenInput.setAttribute( 'type', 'hidden' );
+					hiddenInput.setAttribute( 'id', 'webp_uploads_use_picture_element_hidden' );
+					hiddenInput.setAttribute( 'name', checkbox.getAttribute( 'name' ) );
+					hiddenInput.setAttribute( 'value', checkbox.getAttribute( 'value' ) );
+					checkbox.parentElement.insertBefore( hiddenInput, checkbox.nextSibling );
+				}
 			} );
 		</script>
 	<?php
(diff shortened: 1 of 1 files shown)


plugins/webp-uploads/settings.php
```diff
@@ -26,7 +26,7 @@ function webp_uploads_register_media_settings_field(): void {
 		array(
 			'sanitize_callback' => 'webp_uploads_sanitize_image_format',
 			'type'              => 'string',
-			'default'           => 'avif',                                                                                                                                    // AVIF is the default if the editor supports it.
+			'default'           => 'avif', // AVIF is the default if the editor supports it.
```
westonruter, line 29: I noticed this as well. Thanks for fixing this.

plugins/webp-uploads/settings.php
```diff
@@ -204,6 +204,9 @@ function webp_uploads_use_picture_element_callback(): void {
 		#webp_uploads_picture_element_fieldset.disabled p {
 			opacity: 0.7;
 		}
+		#webp_uploads_picture_element_fieldset.disabled #webp_uploads_use_picture_element_label {
+			pointer-events: none;
```
westonruter: That works! However, it will still be toggled with the keyboard, right?

Anyway, it doesn't really matter if the disabled checkbox can or can't be changed.

westonruter (approved):
LGTM

plugins/webp-uploads/settings.php
felixarntz: I don't understand this. Wouldn't toggling with the keyboard still be a problem as it would allow changing the value (since the checkbox is not technically disabled)?

I assume the reason we don't use the `disabled` attribute is that we still need to send the data with the form submission? If so, maybe we should conditionally add `readonly`, to ensure the value cannot be changed while it's "disabled"?

felixarntz (commented):
@mukeshpanchal27 One point of feedback in relation to @westonruter's comment.

plugins/webp-uploads/settings.php
westonruter: > Wouldn't toggling with the keyboard still be a problem as it would allow changing the value (since the checkbox is not technically disabled)?

Yes. IMO, this is just a UX enhancement though. Also accounting for keyboard toggling the checkbox seems overkill. It's not like it's a problem if someone successfully checks or unchecks this "disabled" checkbox, since the state is ignored. The visual `disabled` treatment is just to let them know it doesn't have an effect, which the associated admin notice also makes clear.

This being said, to improve a11y then perhaps setting the `aria-disabled=true` attribute should be done to make this crystal clear.  I've done this in 25d2c02fdc2bfbf48d7751afece024c023d7b03f.

> I assume the reason we don't use the `disabled` attribute is that we still need to send the data with the form submission? 

That is correct. I also noticed a bug where even though even though the `webp_uploads_use_picture_element` checkbox is checked, when the `perflab_generate_webp_and_jpeg` is not checked and the form is submitted, the resulting updated page will still show that the `webp_uploads_use_picture_element` checkbox is unchecked. This is in spite of the fact that the underlying DB option is still `webp_uploads_use_picture_element`. The reason for this is that the `webp_uploads_is_picture_element_enabled()` helper was being used instead of `get_option( 'webp_uploads_use_picture_element' )`, and the helper will always return false if the `perflab_generate_webp_and_jpeg` option is not `true`. I've fixed this in 8c8a294f4a8b65c5ba3b48b29abac38fe50249ea (and 97dc62779a7022bede1ef61a60e20a685c05c532).

> If so, maybe we should conditionally add `readonly`, to ensure the value cannot be changed while it's "disabled"?

Unfortunately, `readonly` doesn't prevent a checkbox from being toggled: https://codepen.io/westonruter/pen/ExBbzPx

felixarntz: > It's not like it's a problem if someone successfully checks or unchecks this "disabled" checkbox, since the state is ignored.

But I thought it isn't really ignored, since the checkbox isn't actually disabled, but only has the `disabled` class. So as far as I understand, it'll send the value to the backend, which then IMO makes it strange if you can change it (even though that probably won't have any real behavior implications).

> Unfortunately, `readonly` doesn't prevent a checkbox from being toggled: https://codepen.io/westonruter/pen/ExBbzPx

Ah, didn't know that, that's too bad. The other alternative to solve this would be to _actually_ disable, while temporarily injecting an `input[type="hidden"]` that's just there to continue sending the previous value back to the backend - since without sending it, WordPress would update the option to `null`.

westonruter: > But I thought it isn't really ignored, since the checkbox isn't actually disabled, but only has the `disabled` class.

Sorry, by ignored I mean that `webp_uploads_is_picture_element_enabled()` will always return false when the `perflab_generate_webp_and_jpeg` option is `false`, regardless of whether the `webp_uploads_use_picture_element` option is true or false. 

> So as far as I understand, it'll send the value to the backend, which then IMO makes it strange if you can change it (even though that probably won't have any real behavior implications).

So yes, technically you could change it, but it won't have any behavioral effect.

> The other alternative to solve this would be to _actually_ disable, while temporarily injecting an `input[type="hidden"]` that's just there to continue sending the previous value back to the backend - since without sending it, WordPress would update the option to `null`.

That's right. But this seems like maybe overkill.

felixarntz: > > The other alternative to solve this would be to _actually_ disable, while temporarily injecting an `input[type="hidden"]` that's just there to continue sending the previous value back to the backend - since without sending it, WordPress would update the option to `null`.
> 
> That's right. But this seems like maybe overkill.

I think we should consider doing that. Both the current approach and this one are sort of hacky. But with what I'm proposing, I think we're on the safer side when it comes to UX and accessibility. Better to actually disable a checkbox that we want disabled than not disable it, but try to make it seem disabled.

The hidden input allows us to solve the disabled checkbox UX in a clean way.

felixarntz: Actually, thinking about it further: Why don't we simply add `disabled` to the checkbox than doing all these workarounds? If the checkbox is disabled, nothing will be sent to the backend. This will set the `webp_uploads_use_picture_element` option to `null` - which is the same as an unchecked checkbox - which is actually what we want: If the checkbox is disabled, you _can't_ use `picture` element, so it makes sense for the option value to be `false`-y.

westonruter: Because let's say you had PICTURE element enabled, but then you temporarily turned off JPEG fallback. When you go back to turn on JPEG fallback, you'd then find that your PICTURE elements aren't automatically being served again. Keeping a non-disabled `input` for `webp_uploads_use_picture_element` ensures that the option isn't mutated unnecessarily. The `webp_uploads_is_picture_element_enabled()` helper function takes care of whether or not you can use PICTURE element based on the state of the two checkboxes.

felixarntz: Okay, that makes sense. In that case though, I think my previous suggestion of using the `disabled` attribute on the checkbox and temporarily injecting `input[type="hidden"]` is a cleaner solution.

We'd only need to inject that additional element if the checkbox was previously checked anyway / the option was `true`. Otherwise the missing element will lead to the same behavior as the unchecked checkbox would.

westonruter: Implemented in 42b1129b545dfce594e79516363be2ba284ffceb. I found it simpler to not have to inject/remove the hidden input, but to just leave it all the time.

[Screen recording 2024-08-14 10.44.47.webm](https://github.com/user-attachments/assets/10d1ac07-ec92-401f-8b3a-14b4f3c950bc)

mukeshpanchal27: Thanks, @westonruter, for updating the PR. The latest changes look good to me.

Clever way to exclude the checkbox in form submission.

plugins/webp-uploads/settings.php
```diff
@@ -210,7 +210,19 @@ function webp_uploads_use_picture_element_callback(): void {
...
+			<input
+				type="checkbox"
+				id="webp_uploads_use_picture_element"
+				aria-describedby="webp_uploads_use_picture_element_description"
+				<?php checked( get_option( 'webp_uploads_use_picture_element', false ) ); // Option intentionally used instead of webp_uploads_is_picture_element_enabled() to persist when perflab_generate_webp_and_jpeg is updated. ?>
+				<?php disabled( ! $jpeg_fallback_enabled ); ?>
+				onchange="document.getElementById('webp_uploads_use_picture_element_value').value = this.checked ? 1 : 0"
+			>
+			<input
+				type="hidden" name="webp_uploads_use_picture_element"
```
felixarntz:
```suggestion
				type="hidden"
				name="webp_uploads_use_picture_element"
```

plugins/webp-uploads/settings.php
```diff
@@ -210,7 +210,19 @@ function webp_uploads_use_picture_element_callback(): void {
 	</div>
 	<div id="webp_uploads_picture_element_fieldset" class="<?php echo ! $jpeg_fallback_enabled ? 'disabled' : ''; ?>">
 		<label for="webp_uploads_use_picture_element" id="webp_uploads_use_picture_element_label">
-			<input name="webp_uploads_use_picture_element" type="checkbox" id="webp_uploads_use_picture_element" aria-describedby="webp_uploads_use_picture_element_description" value="1"<?php checked( webp_uploads_is_picture_element_enabled() ); ?> class="<?php echo ! $jpeg_fallback_enabled ? 'disabled' : ''; ?>" >
+			<input
+				type="checkbox"
+				id="webp_uploads_use_picture_element"
+				aria-describedby="webp_uploads_use_picture_element_description"
+				<?php checked( get_option( 'webp_uploads_use_picture_element', false ) ); // Option intentionally used instead of webp_uploads_is_picture_element_enabled() to persist when perflab_generate_webp_and_jpeg is updated. ?>
```
felixarntz: Why not use `$picture_element_enabled` from above?
```suggestion
				<?php checked( $picture_element_enabled ); // Option intentionally used instead of webp_uploads_is_picture_element_enabled() to persist when perflab_generate_webp_and_jpeg is updated. ?>
```

Maybe rename the variable to something like `$picture_element_option` to clarify what the comment says?

plugins/webp-uploads/settings.php
```diff
@@ -210,7 +210,19 @@ function webp_uploads_use_picture_element_callback(): void {
...
+				<?php disabled( ! $jpeg_fallback_enabled ); ?>
+				onchange="document.getElementById('webp_uploads_use_picture_element_value').value = this.checked ? 1 : 0"
+			>
```
felixarntz: Does the lack of a `name` attribute on this have any unintended side effects? Makes sense we don't send it with the form submission in favor of the hidden input, but I wonder if this leads to other problems (is the `name` attribute used for anything else?).

Alternatively, we avoid tampering with this (user-facing) element completely and inject the hidden input only when needed (if the checkbox is disabled). This could then happen in the above event listener for `#perflab_generate_webp_and_jpeg`, which IMO is cleaner in terms of code organization too as it colocates all the dynamic logic in one place based on the condition (rather than the `onchange` attribute here that feels a bit out of place).

westonruter: > Does the lack of a `name` attribute on this have any unintended side effects? Makes sense we don't send it with the form submission in favor of the hidden input, but I wonder if this leads to other problems (is the `name` attribute used for anything else?).

No, AFAIK the `name` attribute is only relevant here in that it includes the value in the form submission.

> Alternatively, we avoid tampering with this (user-facing) element completely and inject the hidden input only when needed (if the checkbox is disabled). This could then happen in the above event listener for `#perflab_generate_webp_and_jpeg`, which IMO is cleaner in terms of code organization too as it colocates all the dynamic logic in one place based on the condition (rather than the `onchange` attribute here that feels a bit out of place).

IMO, adding and removing a hidden input makes the code more complicated. But if you'd like to see that, feel free to amend the PR.

felixarntz: IMO it only makes the code _more_, but not more complicated. Happy to add this.

plugins/webp-uploads/settings.php
westonruter: Yeah, the reason that `$picture_element_enabled` isn't used is due to the reason in the comment. We need the underlying DB value to be used, not the computed value as returned by `webp_uploads_is_picture_element_enabled()`. I've improved the variable naming in 7a50264e77f4f3050b3ccbed2f2149aa111a0437 and reused the variable here in 67485b434565f6d6be2e9a9b6a39bd57fa100a00.

felixarntz:
@westonruter I implemented the placing and removing of the hidden input in https://github.com/WordPress/performance/pull/1470/commits/a03979113860ac51a1117338974bb7b1246917c1. Please review that commit individually as it makes it easier.

In https://github.com/WordPress/performance/pull/1470/commits/9e5d40f4062b917d4e06552720f366bdd81c5c59 I moved the `<script>` to the other PHP callback since IMO it makes more sense there. While it's about listening to the other element, everything that is part of the logic concerns the UI in that callback, so I think it makes more sense to colocate. We could even use PHP variables for the `name` attributes etc and share them between the PHP and JS code to make it more error-proof, though I didn't do that for now. Can be added if you think it's valuable.

plugins/webp-uploads/settings.php
```diff
@@ -210,19 +199,70 @@ function webp_uploads_use_picture_element_callback(): void {
...
 		} );
+
+		// Listen for clicks on the JPEG output checkbox, enabling/disabling the
+		// picture element checkbox accordingly.
+		document.getElementById( 'perflab_generate_webp_and_jpeg' ).addEventListener( 'change', function () {
+			document.querySelector( '.webp-uploads-use-picture-element' ).classList.toggle( 'webp-uploads-disabled', ! this.checked );
+			document.getElementById( 'webp_uploads_picture_element_notice' ).hidden = this.checked;
+			document.getElementById( 'webp_uploads_picture_element_fieldset' ).classList.toggle( 'disabled', ! this.checked );
+
+			var checkbox = document.getElementById( 'webp_uploads_use_picture_element' );
```
westonruter: We can embrace `const` now. Also, PhpStorm flags the use of `var`. 

```suggestion
			const checkbox = document.getElementById( 'webp_uploads_use_picture_element' );
```

plugins/webp-uploads/settings.php
```diff
@@ -210,19 +199,70 @@ function webp_uploads_use_picture_element_callback(): void {
...
+			checkbox.disabled = ! this.checked;
+
+			// Remove or inject hidden input to preserve original setting value as needed.
+			if ( this.checked ) {
+				var hiddenInput = document.getElementById( 'webp_uploads_use_picture_element_hidden' );
```
```suggestion
				const hiddenInput = document.getElementById( 'webp_uploads_use_picture_element_hidden' );
```

plugins/webp-uploads/settings.php
```diff
@@ -210,19 +199,70 @@ function webp_uploads_use_picture_element_callback(): void {
...
+				if ( hiddenInput ) {
+					hiddenInput.parentElement.removeChild( hiddenInput );
+				}
+			} else if ( checkbox.checked && ! document.getElementById( 'webp_uploads_use_picture_element_hidden' ) ) {
+				// The hidden input is only needed if the value was originally set (i.e. the checkbox enabled).
+				var hiddenInput = document.createElement( 'input' );
```
```suggestion
				const hiddenInput = document.createElement( 'input' );
```

plugins/webp-uploads/settings.php
```diff
@@ -210,19 +199,70 @@ function webp_uploads_use_picture_element_callback(): void {
...
+				hiddenInput.setAttribute( 'type', 'hidden' );
+				hiddenInput.setAttribute( 'id', 'webp_uploads_use_picture_element_hidden' );
+				hiddenInput.setAttribute( 'name', checkbox.getAttribute( 'name' ) );
+				hiddenInput.setAttribute( 'value', checkbox.getAttribute( 'value' ) );
```
westonruter: Less verbose if properties are used instead:

```suggestion
				hiddenInput.type = 'hidden';
				hiddenInput.id = 'webp_uploads_use_picture_element_hidden';
				hiddenInput.name = checkbox.name;
				hiddenInput.value = checkbox.value;
```

plugins/webp-uploads/settings.php
```diff
@@ -210,19 +199,70 @@ function webp_uploads_use_picture_element_callback(): void {
...
+			if ( this.checked ) {
+				var hiddenInput = document.getElementById( 'webp_uploads_use_picture_element_hidden' );
+				if ( hiddenInput ) {
+					hiddenInput.parentElement.removeChild( hiddenInput );
+				}
+			} else if ( checkbox.checked && ! document.getElementById( 'webp_uploads_use_picture_element_hidden' ) ) {
+				// The hidden input is only needed if the value was originally set (i.e. the checkbox enabled).
+				var hiddenInput = document.createElement( 'input' );
+				hiddenInput.setAttribute( 'type', 'hidden' );
+				hiddenInput.setAttribute( 'id', 'webp_uploads_use_picture_element_hidden' );
```
westonruter: This string is duplicated three times. Should be added to a new constant that is reused?

westonruter (commented):
@felixarntz LGTM, just a few minor suggestions

plugins/webp-uploads/settings.php
felixarntz: Does this do the same? I believe I had run into weird issues in the past, where one is not like the other, i.e. not in sync. But maybe that was a specific situation. Are you aware of any quirks with attributes vs properties in that sense?

plugins/webp-uploads/settings.php
felixarntz: We could do that, although then it makes me think we should rather do it across PHP and JS (i.e. define variables in PHP and pass them even to JS), to share more code between server and client. Either way works.

plugins/webp-uploads/settings.php
westonruter: Yeah, although in this case they will be equivalent. We're using the properties elsewhere in this file (e.g. `hidden`), so I think also good for consistency.

plugins/webp-uploads/settings.php
westonruter: Yeah, defining it in PHP once and then reusing in PHP in JS sounds good to me.

westonruter: Done in fa311f93

plugins/webp-uploads/settings.php
westonruter: Done in ae2f0c721e8a422e2bf881f0b1d7ce96680e3705

westonruter:
@adamsilverstein This should be ready for merge. Please review and we'll get this out the door for the Monday release!

PR #1474, opened 2024-08-15
Improve Performance screen when external requests to WordPress.org fail

### LINKED ISSUE #1473: Option Page now completely broke without remote access
## Steps to reproduce

1. Upgrade plugin to 3.3.1
2. Visit Plugin option page

## Screenshots
image

## Additional Context

> Due to your site's configuration, you may not be able to activate the performance features, unless the underlying plugin is already installed. Please install the relevant plugins manually.

There is nothing wrong with our server configuration. There is something wrong with your plugin.
We don't want the website is calling home, that's why we disable any remote request. And we don't need then, because we're using a composer based setup and update/install everything in our local DEV system.

It should be still possible to change plugin options, but now i can't do anything.

 - PHP Version: 8.2
 - Plugin Version: 3.3.1

westonruter:
Fixes #1473.

With the following plugin code active:

```php
add_filter( 
	'pre_http_request',
	static function ( $pre, $args, $url ) {
		if ( strpos( $url, 'wordpress.org' ) ) {
			return true;
		} else {
			return $pre;
		}
	}, 
	10,
	3 
);
```

Before:

image

After:

image

### DIFF
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -103,12 +103,15 @@ function perflab_render_plugins_ui(): void {
 		// Skip if the plugin is not on WordPress.org or there was a network error.
 		if ( $api_data instanceof WP_Error ) {
 			wp_admin_notice(
-				esc_html(
+				wp_kses(
 					sprintf(
 						/* translators: 1: plugin slug. 2: error message. */
-						__( 'Failed to query WordPress.org Plugin Directory for plugin "%1$s". %2$s', 'performance-lab' ),
+						esc_html__( 'Failed to query WordPress.org Plugin Directory for plugin "%1$s". %2$s', 'performance-lab' ),
 						$plugin_slug,
 						$api_data->get_error_message()
+					),
+					array(
+						'a' => array( 'href' => true ),
 					)
 				),
 				array( 'type' => 'error' )
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -152,7 +152,17 @@ function perflab_render_plugins_ui(): void {
 			) .
 			$plugin_list .
 			'<p>' . esc_html( _n( 'The following error occurred:', 'The following errors occurred:', count( $error_messages ), 'performance-lab' ) ) . '</p>' .
-			'<ul><li>' . join( '</li><li>', $error_messages ) . '</li></ul>' .
+			'<ul><li>' .
+			join(
+				'</li><li>',
+				array_map(
+					static function ( string $error_message ): string {
+						return wp_kses( $error_message, array( 'a' => array( 'href' => true ) ) );
+					},
+					$error_messages
+				)
+			)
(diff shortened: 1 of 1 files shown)


plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -124,6 +116,48 @@ function perflab_render_plugins_ui(): void {
...
+
+		$error_messages = array_unique( $error_messages );
+
+		wp_admin_notice(
+			esc_html(
+				_n( 'Failed to query WordPress.org Plugin Directory for the following plugin:', 'Failed to query WordPress.org Plugin Directory for the following plugins:', count( $errors ), 'performance-lab' )
+			) .
+			$plugin_list .
+			'<p>' . esc_html( _n( 'The following error occurred:', 'The following errors occurred:', count( $error_messages ), 'performance-lab' ) ) . '</p>' .
+			'<ul><li>' . join( '</li><li>', $error_messages ) . '</li></ul>' .
```
westonruter: Missing Kses here

felixarntz (approved):
@westonruter +1 to this nice little UX improvement for when things are broken :)

This reminds me that it would be great to avoid all these individual requests. The more plugins we add, the more of a bottleneck / failure risk they'll become.

I don't think wordpress.org has an API endpoint to get the same data for multiple plugins in bulk, but exploring `Requests::request_multiple()` might be a good idea. On the other hand, this would require some hacks because the way WordPress core wraps the regular `Request::request()` method makes for different behavior. I wish WordPress core just had `wp_request_multiple()`... 😮‍💨

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -124,6 +116,61 @@ function perflab_render_plugins_ui(): void {
...
+				esc_html( $status )
+			);
+			$error_messages[] = $error->get_error_message();
+		}
+		$plugin_list .= '</ul>';
+
+		$error_messages = array_unique( $error_messages );
+
+		wp_admin_notice(
+			esc_html(
```
westonruter: @mukeshpanchal27 do we need this now?

```suggestion
			'<p>' . esc_html(
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -124,6 +116,61 @@ function perflab_render_plugins_ui(): void {
...
+				_n( 'Failed to query WordPress.org Plugin Directory for the following plugin:', 'Failed to query WordPress.org Plugin Directory for the following plugins:', count( $errors ), 'performance-lab' )
+			) .
```
```suggestion
			) . '</p>' .
```

mukeshpanchal27:
@westonruter The [wp_admin_notice](https://developer.wordpress.org/reference/functions/wp_admin_notice/) return empty `<p></p>` tag as the default is true for`paragraph_wrap` parameter. In https://github.com/WordPress/performance/pull/1474/commits/96ab82bbdd78a51d0792427593d9b693371b0845 i have set  false for that parameter and add `<p>` tag in last message.

mukeshpanchal27 (approved):

PR #1476, opened 2024-08-16
Move Auto Sizes logic from Enhanced Responsive Images to Image Prioritizer

westonruter:
Now that Auto Sizes is proposed for Core (Core-61847, https://github.com/WordPress/wordpress-develop/pull/7253), the logic in the Enhanced Responsive Images to apply `sizes=auto` in response to changes applied by Image Prioritizer doesn't make sense, as reasoned by @joemcgill in https://github.com/WordPress/performance/pull/1476#issuecomment-2297032967. So this PR eliminates the Optimization Detective logic from that plugin and instead moves it to Image Prioritizer, incorporating it as part of the `IMG` tag visitor which applies changes to the `loading` attribute.

I also realized that the `crossorigin` attribute wasn't being handled properly when constructing preload links. That is fixed in 5dc1eb5c9589c8952de482e0068af047a784ab58.

Secondly, the `IMG` tag visitor also wasn't accounting for attribute values for the `loading`, `fetchpriority`, and `crossorigin` attributes which include whitespace or non-lowercase characters. These have been normalized in ae5d3ce06489dc4481319b21d9e4b95c5daafd5d.

<details><summary>Obsolete PR Description</summary>

There are currently two different code paths in the Enhanced Responsive Images plugin for adding the `auto` value to the `sizes` attribute, the first uses the `wp_content_img_tag` filter and the second uses a tag visitor for Optimization Detective:

https://github.com/WordPress/performance/blob/c5de91c091cc47f1770cf03dd427ba622472e125/plugins/auto-sizes/hooks.php#L55-L88

https://github.com/WordPress/performance/blob/c5de91c091cc47f1770cf03dd427ba622472e125/plugins/auto-sizes/optimization-detective.php#L21-L52

As of #1471, the non-Optimization Detective code path is leveraging HTML Tag Processor in the same as the Optimization Detective code path has been using since #1322. What this means is that we can reuse the same underlying tag processor logic for in both cases. This is also the case for Embed Optimizer in #1302.

</details> 

Also, something that the Optimization Detective code path implemented was _removing_ the `auto` value from the `sizes` attribute when it is not lazy-loaded. ~This was not implemented in current logic for `auto_sizes_update_content_img_tag()`, so this PR also handles that case.~ <ins>This remains exclusively in the Optimization Detective code path, per @felixarntz in https://github.com/WordPress/wordpress-develop/pull/7253#discussion_r1747376598</ins>.

Lastly, in 7e8cebbb894b7787a09fda5d9b3fc2b710d8a280 I've gone back against myself for chosing `strtok()` which was discussed in https://github.com/WordPress/performance/pull/1445#discussion_r1710065271 with a follow-up by @dmsnell in https://github.com/WordPress/performance/pull/1445#discussion_r1714424178. The problem with `strtok()` is that if there are initial `,` characters they will be skipped over. So `,,,,,auto` is currently detected as having an initial `auto` value, whereas Chrome will not recognize this (as it aligns with the HTML spec).


plugins/auto-sizes/tests/test-optimization-detective.php
```diff
@@ -93,6 +93,16 @@ public function data_provider_test_od_optimize_template_output_buffer(): array {
...
 			),
+
+			'wrongly_auto_sized_responsive_img_with_only_auto' => array(
+				'element_metrics' => array(
+					'xpath'             => '/*[1][self::HTML]/*[2][self::BODY]/*[1][self::IMG]',
+					'isLCP'             => false,
+					'intersectionRatio' => 1,
+				),
+				'buffer'          => '<img src="https://example.com/foo.jpg" alt="Foo" width="1200" height="800" loading="lazy" srcset="https://example.com/foo-480w.jpg 480w, https://example.com/foo-800w.jpg 800w" sizes="auto">',
+				'expected'        => '<img data-od-removed-loading="lazy" data-od-replaced-sizes="auto" src="https://example.com/foo.jpg" alt="Foo" width="1200" height="800"  srcset="https://example.com/foo-480w.jpg 480w, https://example.com/foo-800w.jpg 800w" sizes="">',
```
mukeshpanchal27: Why extra space is needed before `srcset` attribute?

plugins/auto-sizes/hooks.php
```diff
@@ -64,28 +64,46 @@ function auto_sizes_update_content_img_tag( $html ): string {
 		return $html;
 	}
 
-	// Bail early if the image is not lazy-loaded.
-	$value = $processor->get_attribute( 'loading' );
-	if ( ! is_string( $value ) || 'lazy' !== strtolower( trim( $value, " \t\f\r\n" ) ) ) {
-		return $html;
+	auto_sizes_apply_tag_parser_updates( $processor );
```
dmsnell: I'll trust that you know what you're doing, but the Tag and HTML Processors are most happy when local and not passed around. There's the danger here of making code appear to change without explanation, because something the process was sent to leads to its internal state changing, and those transitions aren't evident here where the processor was created.

An alternative style would be to pass in the HTML itself to the updater function, let it create its own processor, and receive the string output from it.

plugins/auto-sizes/hooks.php
```diff
@@ -64,28 +64,46 @@ function auto_sizes_update_content_img_tag( $html ): string {
...
+	$has_auto = auto_sizes_attribute_includes_valid_auto( $sizes );
+
+	if ( $is_lazy && ! $has_auto ) {
+		$processor->set_attribute( 'sizes', "auto, $sizes" );
+	} elseif ( ! $is_lazy && $has_auto ) {
+		// Remove auto from the beginning of the list.
+		$processor->set_attribute(
+			'sizes',
+			(string) preg_replace( '/^[ \t\f\r\n]*auto[ \t\f\r\n]*(,[ \t\f\r\n]*)?/i', '', $sizes )
+		);
```
dmsnell: at this point it seems like there's some zig-zagging going on for the `auto` keyword. just going to toss this out there, but we're using one method to detect its presence, and another method to add it, but I wonder if the PCRE method doesn't already in some way capture both situations, and might not be clearer to rely on vs. splitting the detection and replacement.

```php
function toggle_auto_on_sizes_attribute( string $existing, bool $wants_auto ): string {
	// Find any existing `auto` keyword at the front of the attribute.
	preg_match( '/^([ \t\f\r\n]*auto[ \t\r\f\n]*)(?:,|$)/i', $existing, $auto_match, PREG_OFFSET_CAPTURE );
	$has_auto = isset( $auto_match[1][0] );

	if ( $wants_auto && ! $has_auto ) {
		return "auto, {$sizes}";
	} elseif ( ! $wants_auto && $has_auto ) {
		// Chop off leading `auto` keyword.
		return substr( $existing, strlen( $auto_match[1][0] ) );
	}

	return $existing;
}
```

Then the calling code uses `$is_lazy` as a proxy for `$wants_auto`.

```php
$processor->set_attribute( 'sizes', toggle_auto_on_sizes_attribute( $sizes, $is_lazy ) );
```

plugins/auto-sizes/tests/test-optimization-detective.php
westonruter: Because the `loading` attribute was previously located here and the tag processor doesn't currently strip out that extra space.

plugins/auto-sizes/hooks.php
westonruter: The reason for passing around the processor is that there are two different code paths to this logic being needed:

* When updating an IMG tag in isolation via the `wp_content_img_tag` filter.
* When updating an IMG tag in the context of an entire document that is currently being iterated over in Optimization Detective. 

In Optimization Detective the tag processor is passed to each of the registered tag visitors for each open tag while iterating over the document. See the `auto_sizes_visit_tag()` function in the diff below for how it is being used there.

> There's the danger here of making code appear to change without explanation, because something the process was sent to leads to its internal state changing, and those transitions aren't evident here where the processor was created.

I don't understand. What's the danger?

dmsnell: this is, by the way, a problem that eventually we'll want to solve in Core. I have no idea what the answer might look like today, but it's in the realm of "limited access processors" that could be handed off in a WordPress filter, for example.

here's an example of where things can go fishy.

```php
$total_images = 0;
while ( $processor->next_tag( 'IMG' ) ) {
	++$total_images;
	do_something( $processor );
}
```

At the end of this script we only see `$total_images === 1` and wonder why it's so wrong, given that we sent 1,000 images to it. Well, `do_something()` called `$processor->next_tag( 'SOURCE' )` but there weren't any, so by the time it returns control to the calling loop, the Tag Processor has reached the end of the document and skips all remaining images.

westonruter: Thanks for that. Optimization Detective is actually accounting for that by setting a bookmark on the tag prior to invoking all the tag visitors, and then after they have been invoked it checks if the cursor has moved at all, and if so, it seeks back to the bookmark before continuing to the next tag:

https://github.com/WordPress/performance/blob/4d2299803fb5ac397ca0df14567d7b4d5791caad/plugins/optimization-detective/optimization.php#L217-L232

The processor subclass introduces `$processor->get_seek_count()` and `$processor->get_next_token_count()` (the name of the latter I see is confusing) which return the number of times `seek()` and `next_token()` were called.

dmsnell: this is all baked into "I trust you" but in general, I like noting it, and don't think it's a terrible idea to caution others against following the example. it looks deceivingly simpler than it is.

joemcgill:
I was talking to @felixarntz last week and think we should reconsider whether the OD compatibility code belongs in this plugin at all. When this was added, it was to ensure that when an Image Prioritizer removed `loading="lazy"` from an image, this plugin didn't still add `sizes="auto"`. 

In principle, it is a great goal to try to avoid adding `auto-sizes` to images that have their `loading` attribute modified late in the execution cycle. However, if `auto-sizes` gets implemented  in Core, it will need to be the responsibility of any plugins that change the image markup to be responsible for removing `auto-sizes` unless we find a better way to run this post-processing very late in the template rendering process.

Even so, I'm think it would be good to think about how we write a callback that can be consumed by other plugins that need to correct the `sizes` value after modifying the `loading` value. It seems like passing in and HTML string and returning an HTML string would be the best way to handle this, but am unsure if Image Prioritizer would be able to make use of a function that doesn't except an `WP_HTML_Tag_Processor` that has already been initialized.

westonruter:
Yeah, that makes sense. The current Optimization Detective extension code should instead be put inside of [`Image_Prioritizer_Img_Tag_Visitor`](https://github.com/WordPress/performance/blob/trunk/plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php) in the Image Prioritizer plugin. I'll update this PR to make that change.

felixarntz:
> think we should reconsider whether the OD compatibility code belongs in this plugin at all. When this was added, it was to ensure that when an Image Prioritizer removed `loading="lazy"` from an image, this plugin didn't still add `sizes="auto"`.

+1 to removing the OD integration for this from `auto-sizes` and rather put parts of it into Image Prioritizer:
* Having `sizes="auto"` on a non-lazy-loaded image is not correct, so it makes sense to me that, if Image Prioritizer removes `loading="lazy"`, it should also remove the `auto` from the `sizes` attribute, if present. This is less about integrating with the `auto-sizes` plugin than it is about making sure that its own functionality doesn't lead to such a bug. FWIW, another plugin could have added the attribute similarly.
* The reverse isn't necessary though. It's not the responsibility of Image Prioritizer to add `sizes="auto"`, so it shouldn't need to do that. Sure, in principle it would be great to have `sizes="auto"` added too if Image Prioritizer adds `loading="lazy"` to an image, but I'm not sure this plugin is the right place. If we want that part, it would make sense to have it in the `auto-sizes` plugin (separately of course from what would be proposed for WP core).
* In other words, the presence of Image Prioritizer shouldn't have to make `sizes="auto"` usage better. But it shouldn't make it worse. :)

plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
```diff
@@ -90,6 +104,23 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
 		// TODO: If an image is visible in one breakpoint but not another, add loading=lazy AND add a regular-priority preload link with media queries (unless LCP in which case it should already have a fetchpriority=high link) so that the image won't be eagerly-loaded for viewports on which it is not shown.
 
+		// Ensure that sizes=auto is set properly.
+		$sizes = $processor->get_attribute( 'sizes' );
+		if ( is_string( $sizes ) ) {
+			$is_lazy  = 'lazy' === $get_attribute_value( 'loading' );
+			$has_auto = $this->sizes_attribute_includes_valid_auto( $sizes );
+
+			if ( $is_lazy && ! $has_auto ) {
+				$processor->set_attribute( 'sizes', "auto, $sizes" );
```
westonruter, line 114: @felixarntz If I understand you https://github.com/WordPress/performance/pull/1476#issuecomment-2297125561 correctly, are you thinking that this condition should be removed? IMO, including auto here would help ensure that the correct image _size_ is prioritized, so I think it makes sense in the scope of Image Prioritizer, especially if `sizes=auto` is merged into WordPress core.

felixarntz: Doesn't have to be removed, but my point was that it's not strictly in scope of the Image Prioritizer plugin. To clarify, strictly speaking IMO whether this is done should be based on whether `sizes="auto"` is _generally_ supported:
* If it's available in Core on the site or another plugin on the site adds it, it makes sense that it's supported here too. That's because it would just _maintain_ (or rather correct) Core's (or the other plugin's) behavior without making it worse (which would be _not_ having `sizes="auto"` on lazy-loaded images).
* But with current WordPress core and without another plugin doing it, it doesn't belong in this plugin. That's because it would effectively _add_ this feature without it being its responsibility.

Now more pragmatically speaking, it probably doesn't matter all that much. Since we are on the path to adding `sizes="auto"` into Core already, we might as well add this here unconditionally, since it has no negative side effects and it would probably be unrealistic to reliably check whether _something else_ in WordPress generally adds `sizes="auto"`.

Hope this all makes sense. 😆

westonruter: Yes, that makes sense. Let's leave it in then.

And while it may not be "strictly" in scope for Image Prioritizer, I think the case can be made for it to be loosely in scope, as it is prioritizing the loading of the right image size. And if it isn't the right scope still, we can just change the scope again as we've done for other plugins 😊

plugins/auto-sizes/hooks.php
```diff
@@ -98,8 +98,8 @@ function auto_sizes_update_content_img_tag( $html ): string {
  * @return bool True if the 'auto' keyword is present, false otherwise.
  */
 function auto_sizes_attribute_includes_valid_auto( string $sizes_attr ): bool {
-	$token = strtok( strtolower( $sizes_attr ), ',' );
-	return false !== $token && 'auto' === trim( $token, " \t\f\r\n" );
+	list( $first_size ) = explode( ',', $sizes_attr, 2 );
+	return 'auto' === strtolower( trim( $first_size, " \t\f\r\n" ) );
```
westonruter, line 102: This matches what is currently in https://github.com/WordPress/wordpress-develop/pull/7253

plugins/optimization-detective/class-od-link-collection.php
```diff
@@ -25,7 +25,7 @@
  *                   href?: non-empty-string,
  *                   imagesrcset?: non-empty-string,
  *                   imagesizes?: non-empty-string,
- *                   crossorigin?: ''|'anonymous'|'use-credentials',
+ *                   crossorigin?: 'anonymous'|'use-credentials',
```
westonruter, line 28: While the HTML spec allows for links the `crossorigin` attribute to be (cf. [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin)):

- `crossorigin` (boolean attribute, same as `anonymous`)
- `crossorigin=""` (empty value, same as anonymous`)
- `crossorigin=sooooo-secret` (invalid value, same as `anonymous`)
- `crossorigin=anonymous`
- `crossorigin=use-credentials`

This PR normalizes the first three cases to only use the fourth case, which aligns with the reflected DOM `crossOrigin` property. This will ensure that `OD_Link_Collection::OD_Link_Collection()` will be able to properly merge links for all `anonymous` links.

plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
```diff
@@ -90,6 +104,23 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
 		}
 		// TODO: If an image is visible in one breakpoint but not another, add loading=lazy AND add a regular-priority preload link with media queries (unless LCP in which case it should already have a fetchpriority=high link) so that the image won't be eagerly-loaded for viewports on which it is not shown.
 
+		// Ensure that sizes=auto is set properly.
+		$sizes = $processor->get_attribute( 'sizes' );
+		if ( is_string( $sizes ) ) {
+			$is_lazy  = 'lazy' === $get_attribute_value( 'loading' );
+			$has_auto = $this->sizes_attribute_includes_valid_auto( $sizes );
```
felixarntz, line 111: I think having yet another implementation of this method here is not great and overkill. Since the primary purpose of this is to support the feature in Core, I think we should look for the Core function (see https://github.com/WordPress/wordpress-develop/pull/7253) and use it if present. Otherwise, let's just go with something simple inline.
```suggestion
			if ( function_exists( 'wp_sizes_attribute_includes_valid_auto' ) ) {
				$has_auto = wp_sizes_attribute_includes_valid_auto( $sizes );
			} else {
				$has_auto = 'auto' === $sizes || str_starts_with( $sizes, 'auto,' );
			}
```

Of course the fallback is not 100% reliable, but it's also only a fallback, and it'll work perfectly well to cater for what the "Enhanced Responsive Images" plugin does. We're mostly doing this to support Core and that plugin, so I think this is a better solution in scope of this plugin.

Related: https://github.com/WordPress/performance/pull/1476/files#r1747386984

plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
```diff
@@ -126,4 +157,20 @@ static function ( string $value ): bool {
...
+	 * @since n.e.x.t
+	 * @see auto_sizes_attribute_includes_valid_auto()
+	 *
+	 * @param string $sizes_attr The 'sizes' attribute value.
+	 * @return bool True if the 'auto' keyword is present, false otherwise.
+	 */
+	private function sizes_attribute_includes_valid_auto( string $sizes_attr ): bool {
+		list( $first_size ) = explode( ',', $sizes_attr, 2 );
+		return 'auto' === strtolower( trim( $first_size, " \t\f\r\n" ) );
+	}
```
felixarntz: See above, I think this should be removed. It's too much only intended for edge cases to have this logic duplicated across three different codebases.
```suggestion
```

plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
westonruter: How about 775ab186dad9e6854d15cfc4d6f0aed295abf18c?

plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
westonruter: See 775ab186dad9e6854d15cfc4d6f0aed295abf18c

plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
felixarntz: LGTM 👍

plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
```diff
@@ -126,4 +157,25 @@ static function ( string $value ): bool {
...
 		return true;
 	}
+
+	/**
+	 * Checks whether the given 'sizes' attribute includes the 'auto' keyword as the first item in the list.
+	 *
+	 * Per the HTML spec, if present it must be the first entry.
+	 *
+	 * @since n.e.x.t
+	 * @see auto_sizes_attribute_includes_valid_auto()
```
mukeshpanchal27:
```suggestion
```

Let's remove auto-sizes function as we also use WP future function `wp_sizes_attribute_includes_valid_auto` also

mukeshpanchal27 (approved):

felixarntz (approved):

joemcgill:
Thanks for handling this. Reviewed the changes post-merge and looks good to me.

PR #1478, opened 2024-08-16
Introduce get_cursor_move_count() to use instead of get_seek_count() and get_next_token_count()

westonruter:
When explaining Optimization Detective's iteration over the tags in the document in https://github.com/WordPress/performance/pull/1476#discussion_r1720077867, I realized that the naming of the `OD_HTML_Tag_Processor::get_next_token_count()` method is poor. It's supposed to count the number of times that `next_token()` was called, but it looks like it gives you a count of the remaining tokens in the document. Also, in addition to this method there is `get_seek_count()` which returns the underlying protected `$seek_count` variable. We don't really need these two methods and can instead have a single `get_cursor_move_count()` that keeps track of the number of times that `next_token()` was called (which is what is called both by `seek()` and `next_tag()`).

### DIFF
--- a/plugins/optimization-detective/class-od-html-tag-processor.php
+++ b/plugins/optimization-detective/class-od-html-tag-processor.php
@@ -189,18 +189,6 @@ final class OD_HTML_Tag_Processor extends WP_HTML_Tag_Processor {
 	private $cursor_move_count = 0;
 
 	/**
-	 * Count for the number of times next_token() was called.
-	 *
-	 * The method that uses this is deprecated and it will be removed in a future release.
-	 *
-	 * @since 0.4.1
-	 * @var int
-	 * @see self::next_token()
-	 * @see self::get_next_token_count()
-	 */
-	private $next_token_count = 0;
-
-	/**
 	 * Finds the next tag.
 	 *
 	 * Unlike the base class, this subclass disallows querying. This is to ensure the breadcrumbs can be tracked.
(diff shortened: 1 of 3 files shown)


plugins/optimization-detective/class-od-html-tag-processor.php
```diff
@@ -338,15 +352,30 @@ public function next_token(): bool {
...
+	public function get_cursor_move_count(): int {
+		return $this->cursor_move_count;
+	}
+
 	/**
 	 * Gets the number of times next_token() was called.
 	 *
 	 * @since 0.4.1
 	 * @see self::next_token()
+	 * @deprecated Use {@see self::get_cursor_move_count()} instead.
```
adamsilverstein: Are we deprecating this so dependent plugins don't break? If we don't explicitly call from outside the plugin, couldn't we just remove this?

westonruter: Yes, we could just remove it. I'm happy to do do if you think that's best. It's extremely unlikely anyone is using these deprecated methods.

westonruter: Removed in f4d0f8c9

plugins/optimization-detective/class-od-html-tag-processor.php
```diff
@@ -179,13 +179,14 @@ final class OD_HTML_Tag_Processor extends WP_HTML_Tag_Processor {
 	private $buffered_text_replacements = array();
 
 	/**
-	 * Count for the number of times next_token() was called
+	 * Count for the number of times that the cursor was successfully moved.
```
westonruter: Actually, it may not be "successful" as it could just be an attempt to move the cursor.

```suggestion
	 * Count for the number of times that the cursor was moved.
```

felixarntz (approved):

PR #1479, opened 2024-08-16
Clarify docs around a tag visitor's boolean return value

### LINKED ISSUE #1342: Unclear purpose for tag visitors returning true or false in Optimization Detective
When registering a tag visitor in Optimization Detective, the callable must return a boolean. When the boolean returns `true`, then this is an indication that the tag was indeed "visited" or rather that the visitor is relevant to the current tag. For example:

```php
add_action( 'od_register_tag_visitors', static function ( OD_Tag_Visitor_Registry $registry ) {
	$registry->register( 'async_images', static function ( OD_Tag_Visitor_Context $context ) {
		if ( $context->processor->get_tag() !== 'IMG' ) {
			return false;
		}
		if ( $context->processor->get_attribute( 'decoding' ) !== 'async' ) {
			$context->processor->set_attribute( 'decoding', 'async' );
		}
		return true;
	} );
} );
```

Clearly this visitor should return `false` if the current tag is not an `IMG`. But should it still also return `false` if it already has a `decoding=async` attribute? 

In reality, when a visitor returns `true` this is a signal for whether or not the tag should be captured to be stored in the URL Metrics:

https://github.com/WordPress/performance/blob/4ff08f98a6e709e03068e20c5560599e98cf4afd/plugins/optimization-detective/optimization.php#L180-L192

So when a visitor returns `true` then this means that `$did_visit` will be `true` and if the URL Metrics for current URL are not complete (if `$needs_detection` is `false`), then the effect is that it will add the `data-od-xpath` attribute to the tag. What does this do? It means that when `detect.js` runs, it will include the element among all of the elements for which it captures metrics, including via IntersectionObserver and web-vitals.js:

https://github.com/WordPress/performance/blob/726f10c652628a0c0e577ffd2a39395913047c20/plugins/optimization-detective/detect.js#L242

https://github.com/WordPress/performance/blob/726f10c652628a0c0e577ffd2a39395913047c20/plugins/optimization-detective/detect.js#L285-L287

However, in the case above for a tag visitor that just makes sure that all `img` tags have `decoding=async`, there is no need for this tag to ever be captured in the URL metrics because the URL metrics do not impact how the applied optimization in any way. This is in contrast with what Image Prioritizer does with applying lazy-loading and fetchpriority=high, where it depends on the URL Metrics to know what optimization to perform. Embed Optimizer also depends on URL Metrics to know whether an embed should be lazy-loaded or instead should have preconnect links added. Conversely, the auto-sizes integration in #1322 doesn't depend on the URL Metrics and so at present its tag visitor could always return false.

In short, I think the explanation of what the boolean means being returned from a tag visitor. It's not whether the tag was visited, but rather whether or not the tag needs to be tracked in the URL Metrics.

It could also be possible to automatically detect whether or not a tag visitor depends on URL metrics by detecting whether or not it accessed the `$context->url_metrics_group_collection`. If so, then this would be a signal that indeed the tag should be tracked in the URL metrics, and the return value of the tag visitor could be eliminated entirely. However, this could be a bit too sophisticated.

westonruter:
> [!IMPORTANT]
> This is a follow-up PR to #1478. Merge that PR into `trunk` before merging this PR, since it is branched off of that PR's branch.

Fixes #1342

The terminology around a tag visitor "visiting" a tag is fuzzy. It's not clear what it means for a tag to be "visited". In reality, what this means is that the tag is tracked/recorded in URL metrics so that it appears among its `elements`, including the `intersectionRect` and `clientBoundingRect` among other aspects captured by `detect.js`. If a tag visitor does not need any of that information, then it can just always return `false`, even though it may have "visited" a tag by applying some optimization to it. For example. Enhanced Responsive Images (`auto-sizes`) has a tag visitor doesn't need that information to apply its optimizations as it just adds `auto` to the `sizes` attribute if it is lazy-loaded and doesn't have `auto` already. This is a case where the tag visitor always returns `false`.

As explained in #1342, instead of returning a boolean value for whether to track the tag in URL metrics, it could be more sophisticated by detecting whether the tag visitor tried to access information about that tag from the stored URL metrics. Or instead of returning a boolean there could be a method on the context object like `$context->track_in_url_metrics()` which would do the same. I could go either way. But these changes in the PR are a good iterative improvement at least and we can add such a method later.

### DIFF
--- a/plugins/auto-sizes/optimization-detective.php
+++ b/plugins/auto-sizes/optimization-detective.php
@@ -16,7 +16,7 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @since 1.1.0
  *
  * @param OD_Tag_Visitor_Context $context Tag visitor context.
- * @return false Whether the tag should be recorded in URL metrics.
+ * @return false Whether the tag should be tracked in URL metrics.
  */
 function auto_sizes_visit_tag( OD_Tag_Visitor_Context $context ): bool {
 	if ( 'IMG' !== $context->processor->get_tag() ) {
--- a/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
+++ b/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
@@ -32,7 +32,7 @@ final class Embed_Optimizer_Tag_Visitor {
 	 * @since 0.2.0
 	 *
 	 * @param OD_Tag_Visitor_Context $context Tag visitor context.
-	 * @return bool Whether the visit or visited the tag.
+	 * @return bool Whether the tag should be tracked in URL metrics.
 	 */
 	public function __invoke( OD_Tag_Visitor_Context $context ): bool {
 		$processor = $context->processor;
--- a/plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
@@ -23,7 +23,7 @@ final class Image_Prioritizer_Background_Image_Styled_Tag_Visitor extends Image_
 	 * Visits a tag.
 	 *
 	 * @param OD_Tag_Visitor_Context $context Tag visitor context.
-	 * @return bool Whether the visitor visited the tag.
+	 * @return bool Whether the tag should be tracked in URL metrics.
 	 */
 	public function __invoke( OD_Tag_Visitor_Context $context ): bool {
 		$processor = $context->processor;
--- a/plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php
@@ -24,7 +24,7 @@ final class Image_Prioritizer_Img_Tag_Visitor extends Image_Prioritizer_Tag_Visi
 	 *
 	 * @param OD_Tag_Visitor_Context $context Tag visitor context.
 	 *
-	 * @return bool Whether the visitor visited the tag.
+	 * @return bool Whether the tag should be tracked in URL metrics.
 	 */
 	public function __invoke( OD_Tag_Visitor_Context $context ): bool {
 		$processor = $context->processor;
--- a/plugins/image-prioritizer/class-image-prioritizer-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-tag-visitor.php
@@ -23,7 +23,7 @@ abstract class Image_Prioritizer_Tag_Visitor {
 	 * Visits a tag.
 	 *
 	 * @param OD_Tag_Visitor_Context $context Tag visitor context.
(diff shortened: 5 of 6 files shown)


felixarntz (approved):
@westonruter LGTM!

Given the clearer terminology of "track" vs "visit", should we consider renaming relevant code bits as well? Even if any of the API pieces within Optimization Detective use that, we could probably rename, assuming that so far it's only ourselves that have built integrations with it and we haven't documented it much anyway? Alternatively, we could deprecate the old ones to be super safe.

westonruter:
@felixarntz which code bits are you referring to?

plugins/auto-sizes/optimization-detective.php
```diff
@@ -16,7 +16,7 @@
  * @since 1.1.0
  *
  * @param OD_Tag_Visitor_Context $context Tag visitor context.
- * @return false Whether the tag should be recorded in URL metrics.
+ * @return false Whether the tag should be tracked in URL metrics.
  */
 function auto_sizes_visit_tag( OD_Tag_Visitor_Context $context ): bool {
```
felixarntz, line 21: @westonruter 
> which code bits are you referring to?

Haven't looked at the codebase outside this PR, but this function for example would better have a different name now. Of course this is just a plugin _consuming_ Optimization Detective, but it led me to bring this up.

But there's definitely also relevant code in Optimization Detective itself, e.g. the `od_register_tag_visitors` action could probably use a more accurate name too.

westonruter: These are still tag visitors though. That doesn't change here. What is changed is that tag visitors always visit tags (and they always have) but what is made more clear is what the return value means from a tag visitor. A true return value means that it is indicating that the tag should be tracked/registered/monitored in URL metrics. In the case of the Enhanced Responsive Images plugin, the tag visitor never needs to read the tag out of URL metrics so it always returns false.

felixarntz: But why are they called "tag visitors"? I agree with the rationale for the changes in this PR, but in the same sense I find it equally unclear what "visiting" is about when looking at the API as a whole. "Tracking" a tag makes a lot more sense for me in describing the purpose of why I should be using this API. The purpose, as far as I understand, is to track certain tags - which implies "visiting", but that's not really the purpose and less clear to me.

westonruter: Tag visitors visit every open tag in the document. 

Only a subset of tag visitors need to track the tag in URL metrics. If a tag visitor doesn't need URL metrics to perform it's optimizations (as is the case for auto-sizes) then no tracking is happening.

felixarntz: Ah, thanks for clarifying!

joemcgill (approved):

adamsilverstein (approved):

westonruter:
To merge this we first need https://github.com/WordPress/performance/pull/1478 to be reviewed and merged 🙏

PR #1487, opened 2024-08-20
[WIP] Add support for server-timing performance panel extension

swissspidy:
## Summary

Adds support for an additional `start` field in milliseconds.

Adds support for special `response-start` and `response-end` fields that should not get a `wp-` prefix.

Adds support for an optional `desc`.


plugins/performance-lab/includes/server-timing/class-perflab-server-timing.php
```diff
@@ -288,19 +288,43 @@ function ( string $output, ?int $phase ): string {
...
 
-		return sprintf( 'wp-%1$s;dur=%2$s', $name, $value );
+		if ( 'response-start' !== $name && 'response-end' !== $name ) {
+			$name = 'wp-' . $name;
+		}
+
+		$header_value = sprintf( '%s', $name );
+
+		if ( null !== $value ) {
+			$header_value .= sprintf( ';dur=%s', $value );
```
westonruter, line 317: Technically shouldn't this be formatted as a float?

```suggestion
			$header_value .= sprintf( ';dur=%f', $value );
```

Maybe not, because for some reason this seems to result in a loss of precision! https://3v4l.org/oOVJX

I consulted with Gemini: https://g.co/gemini/share/7e25d0065550

Looks like if you want to retain the precision then you have to explicitly pass the desired precision, like `%.10f` but then you can get a lot of extra zeros when you don't need them!

So nevermind :)

plugins/performance-lab/includes/server-timing/class-perflab-server-timing.php
```diff
@@ -288,19 +288,43 @@ function ( string $output, ?int $phase ): string {
...
 			$value = round( $value, 2 );
 		}
 
+		if ( is_float( $start ) ) {
+			$start = round( $start, 2 ) * 1000.00;
+		}
+
+		// If no value is set, make sure it's just passed through.
+		if ( null === $start && null === $value ) {
+			return null;
```
adamsilverstein, line 304: should this return `''` or the original passed value? Looking at the use of the return value, [this](https://github.com/WordPress/performance/blob/ec4559e69448a500b836379ddf56b269875dafa9/plugins/performance-lab/includes/server-timing/hooks.php#L35) check explicitly looks for an empty string.

swissspidy: That's for the whole header though, and this is just for 1 metric, which is a small part of the whole header.

Also note that this is just moved code. It was already returning null before.

phanduynam:

swissspidy:
@phanduynam No it's not, please stop spamming the repo with these comments.

Issue #1324, opened 2024-07-01
Plugin Proposal: Performance Dashboard


swissspidy:
## Overview

- Proposed plugin name: Performance Dashboard (or maybe _Performance Overview_ or _Performance Summary_)
- Proposed plugin slug: `performance-dashboard`
- Proposed plugin owner GitHub username(s): @swissspidy

### Purpose

Leverage Optimization Detective to collect Core Web Vitals field data inside WordPress. Enables site owners to see what real users are experiencing on their site, without CrUX.

Optimization Detective is uniquely positioned to also collect web vitals RUM data. This data can then be presented in the WordPress dashboard for site owners to keep an eye on their site's performance.

For context, this is based on a proof of concept I did in #1098, which already has a bit of a discussion there.

Note: I don't know how useful the idea is, but I wanted to create an actual issue for the POC so that it doesn't get lost.

Related:

I found an existing [Core Web Vitals Monitor](https://wordpress.org/plugins/core-web-vitals-monitor/) WordPress plugin which also does this but is somehow not maintained nor popular. We might be able to learn something from it.

See also:

* https://github.com/WordPress/gutenberg/issues/33578

### Scope

Screenshots from the POC:

Dedicated dashboard page:

<img width="1552" alt="Screenshot 2024-03-28 at 17 28 25" src="https://github.com/WordPress/performance/assets/841956/f7b1deef-c04c-4fc5-8622-c655625ddbbe">

Admin dashboard widget:

<img width="516" alt="Screenshot 2024-04-10 at 10 16 17" src="https://github.com/WordPress/performance/assets/841956/2179d925-3779-49d6-9960-1633a42de1d8">

#### Potential

In the future we could:

* Provide an easy-to-use (AI?) summary for users to better understand their site's performance.
* Get actual performance improvement suggestions based on the data, perhaps AI powered.
* Provide call-to-actions, e.g. to fix the offending LCP image in the block editor
* Identify plugins causing LCP/CLS/etc. issues and provide call-to-actions
* Combine with site activity (e.g. see that LCP dropped right after plugin X was activated)
* Combine with other (external) data sources
* Expose (aggregated) CWV data publicly for analysis purposes
	* Opt-in or opt-out
	* Alternatively send to some telemetry server
* Use [Long Animation Frames API](https://developer.chrome.com/docs/web-platform/long-animation-frames) to identify offenders
* Collect `Server-Timing` measurements per page
[...]

PR #1492, opened 2024-08-21
Allow URL metric schema to be extended

westonruter:
Discussed previously:

- https://github.com/WordPress/performance/pull/1490#issuecomment-2300905345
- https://github.com/WordPress/performance/pull/1098#discussion_r1550488934
- https://github.com/WordPress/performance/pull/1098#discussion_r1550492139

This PR introduces extensibility for the URL Metric. Two new filters are introduced:

* `od_url_metric_schema_root_additional_properties`
* `od_url_metric_schema_element_item_additional_properties`

These allow for new properties to be added to the root of the URL Metric and to the schema for an element, respectively. 

For example, to capture whether a user agent was using a touch interface:

```php
add_filter(
	'od_url_metric_schema_root_additional_properties',
	static function ( $additional_properties ) {
		$additional_properties['isTouch'] = array(
			'type' => 'boolean',
		);
		return $additional_properties;
	}
);
```

Or for example, to add a `resizedHeight` property to an `element`, you can do:

```php
add_filter(
	'od_url_metric_schema_element_item_additional_properties',
	static function ( $additional_properties ) {
		$additional_properties['resizedHeight'] = array(
			'type' => 'number',
		);
		return $additional_properties;
	}
);
```

Additional properties are not allowed to be `required` and they must not conflict with existing core properties. When a URL Metric is stored with additional properties, these properties will persist even when the plugin that added the schema for the additional property is deactivated. This is to avoid invalidating many URL metrics immediately upon deactivating an extension. Additional unrecognized properties are always disallowed when submitting new URL metrics for storage (via the new `OD_Strict_URL_Metric` class), but existing stored URL metrics allow unrecognized properties to persist.

(In order to populate the extended properties in the URL Metric, see #1373 for the client-side module extensions.)

This PR also introduces improved ergonomics for accessing properties of the `OD_URL_Metric` class, including a `get( $property )` method (which is key for accessing extended properties) as well as magic getters for core properties:

- `$url_metric->uuid`
- `$url_metric->url`
- `$url_metric->timestamp`
- `$url_metric->viewport`
- `$url_metric->elements`

Additionally, the `OD_URL_Metric` class's validation logic has now been extended to also apply sanitization to the inputs. This is to ensure that the expected PHP types are being passed around, since the string `'false'` is a valid boolean according to the WordPress REST API, but it is not a PHP boolean `false`. This could cause PHP type errors at some point. So now `rest_sanitize_value_from_schema()` is called after first doing `rest_validate_object_value_from_schema()`.

Lastly, a `od_url_metric_collected` action now fires whenever a new URL Metric is stored. This would be useful for extensions to perhaps aggregate URL Metrics data in another table such as to power a dashboard showing real user metrics, see https://github.com/WordPress/performance/issues/1324.

# To do

- [x] Add filters so that additional properties can be added to the URL metric root schema and the schema for `elements` items.
- [x] Add `get()` method to `OD_URL_Metric` to allow arbitrary parts of the URL metric data to be obtained. See https://github.com/WordPress/performance/pull/1098#discussion_r1550492139.
- [x] Make sure that filters that add additional root and element item properties don't cause existing URL metrics to all become invalid. An update to a URL metric should cause the additional properties to be removed.
- [x] Add filter to allow new server-side data to be amended to a URL metric when it is created in `od_handle_rest_request()`. (Data will be able to be amended client-side via extension module scripts in #1373.)


plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -212,6 +218,128 @@ public static function get_json_schema(): array {
...
+	 * @param string               $filter_name           Filter name used to extend.
+	 *
+	 * @return array<string, mixed> Extended schema.
+	 */
+	protected static function extend_schema_with_optional_properties( array $properties_schema, array $additional_properties, string $filter_name ): array {
+		$doing_it_wrong = static function ( string $message ) use ( $filter_name ): void {
+			_doing_it_wrong(
+				esc_html( "add_filter({$filter_name},...)" ),
+				esc_html( $message ),
+				'Optimization Detective n.e.x.t'
```
swissspidy, line 290: Just to confirm, our script will properly replace this version number?

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -212,6 +218,128 @@ public static function get_json_schema(): array {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array<string, mixed> $properties_schema     Properties schema to extend.
+	 * @param array<string, mixed> $additional_properties Additional properties.
+	 * @param string               $filter_name           Filter name used to extend.
+	 *
+	 * @return array<string, mixed> Extended schema.
+	 */
+	protected static function extend_schema_with_optional_properties( array $properties_schema, array $additional_properties, string $filter_name ): array {
```
swissspidy, line 285: The whole method is a bit complex with all the validation in it. Is there some existing WP function we could use for schema validation instead? Or does WP do this validation later on anyway?

I don't see similar validation in `register_rest_field` for instance.

plugins/optimization-detective/class-od-url-metric.php
westonruter: No, it doesn't. We'll have to manually update this.

westonruter: It will be automatic as of https://github.com/WordPress/performance/pull/1495

plugins/optimization-detective/class-od-url-metric.php
westonruter: WordPress does have a [`_doing_it_wrong()` check](https://github.com/WordPress/wordpress-develop/blob/06f7c825842f0e55fea3b82f694b43e45d1c66df/src/wp-includes/rest-api.php#L2151-L2154) which emits:

> Function rest_validate_value_from_schema was called incorrectly. The "type" schema keyword for OD_URL_Metric[timestamp] is required. (This message was added in version 5.5.0.)

But then it goes ahead and [tries to access it anyway](https://github.com/WordPress/wordpress-develop/blob/06f7c825842f0e55fea3b82f694b43e45d1c66df/src/wp-includes/rest-api.php#L2156), resulting in a warning:

image

Maybe the validation is excessive.

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -212,6 +218,128 @@ public static function get_json_schema(): array {
...
+		$doing_it_wrong = static function ( string $message ) use ( $filter_name ): void {
+			_doing_it_wrong(
+				esc_html( "add_filter({$filter_name},...)" ),
```
westonruter: I don't know if `"add_filter({$filter_name},...)"` is best here. It seems better than supplying some `OD_URL_Metric` method, however, since the user isn't calling one of those incorrectly.

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -212,6 +218,128 @@ public static function get_json_schema(): array {
...
+				$doing_it_wrong(
+					sprintf(
+						/* translators: property name */
+						__( 'Disallowed override of existing schema property "%s".', 'optimization-detective' ),
+						$property_key
+					)
+				);
+				continue;
+			}
+			if ( ! isset( $property_schema['type'] ) || ! is_string( $property_schema['type'] ) ) {
```
westonruter: Actually, `type` can also [be an array](https://github.com/WordPress/wordpress-develop/blob/06f7c825842f0e55fea3b82f694b43e45d1c66df/src/wp-includes/rest-api.php#L2156-L2169).

```suggestion
			if ( ! isset( $property_schema['type'] ) ) {
```

westonruter: Fixed in 91fcd01e

plugins/optimization-detective/class-od-url-metric.php
westonruter: Yoast SEO is doing it like this:

```php
			\_doing_it_wrong(
				'Filter: \'wpseo_breadcrumb_links\'',
				'The `wpseo_breadcrumb_links` filter should return a multi-dimensional array.',
				'YoastSEO v20.0'
			);
```

That makes sense to me.

westonruter: Done in 381c829c

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -225,12 +232,141 @@ public static function get_json_schema(): array {
...
+					sprintf(
+						/* translators: 1: property name, 2: 'type' */
+						__( 'Supplied schema property "%1$s" with missing "%2$s" key.', 'optimization-detective' ),
+						'type',
+						$property_key
+					)
+				);
+				continue;
+			}
+			// TODO: Should 'default' be required?
```
felixarntz: I'd say either yes, or we should at least automatically set it based on the `type` (e.g. `''` for type `string` etc). Since all of these parameters need to be optional, there needs to be a default I guess?

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -225,12 +232,141 @@ public static function get_json_schema(): array {
...
+						'additionalProperties' => true,
 					),
 				),
 			),
-			'additionalProperties' => false,
+			// Additional root properties may be added to the schema via the od_url_metric_schema_root_additional_properties filter.
+			// Therefore, additionalProperties is set to true so that additional properties defined in the extended schema may persist
+			// in a stored URL Metric even when the extension is deactivated. For REST API requests, the OD_Strict_URL_Metric
+			// which sets this to false so that newly-submitted URL Metrics only ever include the known properties.
+			'additionalProperties' => true,
```
felixarntz, line 240: The explanation sort of makes sense to me, but where specifically is it relevant to have `additionalProperties` set to `true` for existing URL metrics?

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -44,45 +44,51 @@
  *                                elements: ElementData[]
  *                            }
  *
+ * @property-read string        $uuid
+ * @property-read string        $url
+ * @property-read float         $timestamp
+ * @property-read ViewportRect  $viewport
+ * @property-read ElementData[] $elements
```
felixarntz: Not sure about the value of this and `__get()`. Seems unrelated to the purpose of this PR anyway, so maybe let's discuss that separately? I think the `get()` method is good enough?

plugins/optimization-detective/class-od-url-metric.php
felixarntz: I think the validation here makes sense, specifically because some of those requirements we have here don't apply to the WordPress validation (for instance, we want to prohibit `required`, but that's only relevant here, not in general WordPress validation of REST properties).

plugins/optimization-detective/storage/rest-api.php
```diff
@@ -161,6 +166,33 @@ function od_handle_rest_request( WP_REST_Request $request ) {
...
+	do_action(
+		'od_url_metric_collected',
+		array(
+			'post_id'                      => $post_id,
+			'request'                      => $request,
+			'url_metric'                   => $url_metric,
+			'url_metrics_group'            => $group,
+			'url_metrics_group_collection' => $group_collection,
+		)
+	);
```
felixarntz, line 195: What would be an example use-case for this action? Can be an unrealistic example, I'm just curious to see how you envision this to be used.

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -225,12 +232,141 @@ public static function get_json_schema(): array {
...
+		return $properties_schema;
+	}
+
+	/**
+	 * Gets property value for an arbitrary key.
+	 *
+	 * This is particularly useful in conjunction with the `od_url_metric_schema_root_additional_properties` filter.
+	 *
+	 * @since n.e.x.t
+	 * @todo Instead of returning null when the key doesn't exist, should the `default` value be returned as defined in the schema?
```
felixarntz: I think so, yes. Which if we set the default in the REST properties, it should already be set under `$this->data[ $key ]` anyways, right?

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -225,12 +232,141 @@ public static function get_json_schema(): array {
...
+	 *
+	 * @param string $key Property.
+	 * @return mixed|null
```
felixarntz:
```suggestion
	 * @return mixed|null The property value, or null if not set.
```

plugins/optimization-detective/class-od-strict-url-metric.php
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Representation of the measurements taken from a single client's visit to a specific URL without additionalProperties allowed.
+ *
+ * This is used exclusively in the REST API endpoint for capturing new URL metrics to prevent invalid additional data from being
+ * submitted in the request. For URL metrics which have been stored the looser OD_URL_Metric class is used instead.
+ *
+ * @since n.e.x.t
+ * @access private
+ */
+final class OD_Strict_URL_Metric extends OD_URL_Metric {
```
felixarntz, line 23: It's a bit confusing to have "strict" URL metric extend the "extendable" URL metric - would make more sense to be the other way around. This is also highlighted by how the strict class will still run the filters, but without that serving any purpose.

Taking this a bit further, strictly speaking neither of the two makes sense because a strict URL metric is not an extendable URL metric and vice-versa. So it would make more sense to use e.g. a decorator pattern, or make the two classes separate implementations of a `OD_URL_Metric` interface. Of course that would be a decent amount of refactoring to do - simple, but still some code changes so we could consider that separately.

So at this point I mostly think:
* Should the extension tree be the other way around? If not, why not?
* Maybe there's a more clear name we can use than just "strict"? Maybe include "REST" so that it's clear that this class is only used for REST API requests?

felixarntz (commented):
@westonruter The logic looks good mostly, I left a few comments though mostly on architecture bits.

plugins/optimization-detective/class-od-url-metric.php
westonruter: It is relevant for when URL metrics are stored, such as when they were collected at a time when the schema was extended so additional properties were allowed. If `additionalProperties` aren't allowed, then those previously-stored URL metrics would be immediately invalid when an extending plugin is deactivated.

plugins/optimization-detective/storage/rest-api.php
swissspidy: For the performance dashboard POC it's useful as I could then hook into this action, grab my relevant data, and persistently and permanently store it in a new location more suitable for that kind of analytics. Useful since OD metrics are deleted after a while.

plugins/optimization-detective/class-od-url-metric.php
westonruter: In https://github.com/WordPress/performance/pull/1098#discussion_r1550492139 you suggested `__call()` and I responded there.

plugins/optimization-detective/class-od-url-metric.php
felixarntz: I get the idea, but my question was meant to be: Where in WordPress's and this plugin's code execution does that actually become relevant? Where would `additionalProperties` be used, if not in the REST API? Like is that a real problem or just a theoretical problem?

plugins/optimization-detective/class-od-url-metric.php
felixarntz: That's fair. This works.

My only concern with such an approach is that it only works reasonably well if another plugin extending it would also have its own class that it annotated with these `@property-read` annotations. Otherwise, it becomes an incomplete documentation, and in that case I feel it might even be better to just have `get()`, so that it's more obvious to a developer they need to "manually" look at what's available - whereas these annotations may imply that you can simply rely on them, when as soon as an extension happens, this may not longer be the case.

So I think I still prefer not to have this, but not a big deal if you want to keep it.

plugins/optimization-detective/storage/rest-api.php
felixarntz: Okay, so this is mostly for more advanced/custom use-cases then? I was wondering since this PR adds it, whether it's something you _should_ be using just because you add additional metrics. But that's not the case, correct?

plugins/optimization-detective/class-od-url-metric.php
westonruter: If `additionalProperties` is `true` it's the same as if `additionalProperties` isn't set at all. Its presence here is just to have a fully documented schema. The `OD_Strict_URL_Metric` subclass extends the `get_json_schema` method to switch `additionalProperties` to `false` and this subclass is what is used with the REST API. The REST API should only accept currently-recognized properties, so that's why the "strict" version is used.

plugins/optimization-detective/storage/rest-api.php
westonruter: That's right, it's not directly related to extending the schema, but it is related to extending the plugin in general.

plugins/optimization-detective/class-od-strict-url-metric.php
westonruter: I'm not sure I understand. The filters are still used. It's just that any objects have `additionalProperties` set to `false` in the strict version.

westonruter: In other words, in the strict version, it sets `additionalProperties` to `false` which refers to the properties which aren't recognized as part of the schema. The filters are still applying, however, so if a plugin adds new properties to the schema these "additional" properties will be recognized. It's just that additional _unrecognized_ properties will be rejected in the strict verison.

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -225,12 +232,141 @@ public static function get_json_schema(): array {
...
+	 *
+	 * This is particularly useful in conjunction with the `od_url_metric_schema_root_additional_properties` filter.
+	 *
+	 * @since n.e.x.t
+	 * @todo Instead of returning null when the key doesn't exist, should the `default` value be returned as defined in the schema?
+	 *
+	 * @param string $key Property.
+	 * @return mixed|null The property value, or null if not set.
+	 */
+	public function get( string $key ) {
```
westonruter, line 357: This could make use of generics to specify the expected return type given the input `$key`.

westonruter: Maybe not, as then attempting to access an extended property will result in a PHPStan error.

plugins/optimization-detective/class-od-url-metric.php
westonruter: That's true. Plugins that extend the schema wouldn't have type definitions for these properties. However, having them for the core properties is useful so that we don't have to add `get_*()` methods for every root property. I just added 384fa53f9de60b3080b515ab95a06070b50b1a04 & ed5815158552d27359a730e1f6dfa41ccbcc61c9 to make use of these. So it's more ergonomic and feels better. (I think we can deprecate `get_uuid()`, `get_timestamp()`, `get_elements()`, etc.) 

For plugins that extend the schema, this ties back to the discussion of having default values for the properties. ~This makes me think that we should force the default to be `null` so that `isset()` checks will be `false`. It doesn't make sense to have a non-`null` default defined when the client never submitted data for that URL metric in the first place. Having a default value of `null` will avoid cases where array indices will be undefined, but at the same time `isset()` checks will be `false`.~ **Update:** See https://github.com/WordPress/performance/pull/1492#discussion_r1752697070.

As for plugins that extend the schema, I'm thinking about Embed Optimizer adding a `resizedBoundingClientRect` property for the element. This isn't actually directly connected to the `@property-read` discussion, but it is related in static analysis. The Embed Optimizer plugin could access it from an element via:

```php
foreach ( $url_metric->elements as $element ) {
	if ( null !== $element['resizedBoundingClientRect'] ) {
		// ...
	}
}
```

However, PHPStan here complains:

> phpstan: Offset 'resizedBoundingClie…' does not exist on array{isLCP: bool, isLCPCandidate: bool, xpath: string, intersectionRatio: float, intersectionRect: array{width: float, height: float, x: float, y: float, top: float, right: float, bottom: float, left: float}, boundingClientRect: array{width: float, height: float, x: float, y: float, top: float, right: float, bottom: float, left: float}}.

This makes sense because `$element` is typed as `ElementData` which does not have `resizedBoundingClientRect`. (Aside: I think `ElementData` should perhapsbe renamed to  `ElementMetrics`.) 

I tried adding the following to the `Embed_Optimizer_Tag_Visitor` class:

```
 * @phpstan-import-type DOMRect from OD_URL_Metric
 * @phpstan-import-type ElementData from OD_URL_Metric
 * @phpstan-type ExtendedElementData ElementData & array{ resizedBoundingClientRect: ?DOMRect }
```

And then explicitly overriding the type for `$element`:

```php
/**
 * Extended element.
 *
 * @var ExtendedElementData $element
 */
```

But it's not working. I'm getting either errors like:

```
PHPDoc tag @var for variable $element contains unresolvable type.
```

Or when moving `@phpstan-type ExtendedElementData` to the method, I instead get:

```
phpstan: Access to offset 'resizedBoundingClie…' on an unknown class ExtendedElementData.
phpstan: PHPDoc tag @var for variable $element contains unknown class ExtendedElementData.
phpstan: PHPDoc tag @var with type ExtendedElementData is not subtype of type array{isLCP: bool, isLCPCandidate: bool, xpath: string, intersectionRatio: float, intersectionRect: array{width: float, height: float, x: float, y: float, top: float, right: float, bottom: float, left: float}, boundingClientRect: array{width: float, height: float, x: float, y: float, top: float, right: float, bottom: float, left: float}}.
```

I'm nearing my PHPStan knowledge limits. Maybe it just can't do what I want.

plugins/optimization-detective/class-od-url-metric.php
westonruter: See https://github.com/WordPress/performance/pull/1492#discussion_r1752626491. I think we should force the default to always be `null`.

plugins/optimization-detective/class-od-url-metric.php
westonruter: Yes, per https://github.com/WordPress/performance/pull/1492#discussion_r1752626491 I think it will always return `null` anyway.

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -225,12 +232,141 @@ public static function get_json_schema(): array {
...
+	/**
+	 * Gets property value for an arbitrary key.
+	 *
+	 * This is useful with the `@property-read` annotations for the class. For accessing other data,
+	 * it's likely the `get()` method will be more useful for static analysis reasons.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $key Property.
+	 * @return mixed|null
```
westonruter:
```suggestion
	 * @return mixed|null The property value, or null if not set.
```

plugins/optimization-detective/class-od-url-metric.php
westonruter: Hummm. I had thought that `rest_sanitize_value_from_schema()` would supply the `default` value when it is absent, but it doesn't. Therefore, I think `default` should just be disallowed.

westonruter: I've just disallowed the `default` altogether in aa1e66334e52abb2e9e0acb168c080346abcc907.

plugins/optimization-detective/class-od-url-metric.php
felixarntz: I think it comes down to the aforementioned problem with magic getters and `@property-read`. You can't programmatically guarantee that any of this will work. Technically every single use of the magic getter would need to be wrapped in a magic "isset-er". Otherwise, the code is not "safe" and you may at any point unexpectedly work with `null` values.

That's related to my previous concern, and I don't think we should deprecate the getter methods in favor of magic properties. I think the win in ergonomics is tiny - only about not having to write `get_`, which IMO is not a good reason to change this, given that using the methods is a more reliable approach, which PHPStan should be fine with.

For any custom properties, we can use the `get( ... )` method, which is not ideal either, but still could facilitate a cleaner approach where an extension would implement its own class that implements additional `get_...()` methods for new properties.

plugins/optimization-detective/class-od-strict-url-metric.php
felixarntz: Ah right, I had something wrong in my logic here. Some of my other feedback on the class hierarchy still applies, but that's not a big deal in regards to this PR. I _do_ think we should separately consider using a `OD_URL_Metric` interface and use that in all (or at least most of) the type hints so that the class implementations can be independent.

plugins/optimization-detective/class-od-url-metric.php
felixarntz: Just to share this here: We clarified this outside of the PR, this is relevant because the JSON schema is also used to parse URL metrics from existing posts - so that's the use-case where `additionalProperties` needs to be `true`.

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -225,12 +232,151 @@ public static function get_json_schema(): array {
...
+	 * it's likely the `get()` method will be more useful for static analysis reasons.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $key Property.
+	 * @return mixed|null The property value, or null if not set.
+	 */
+	public function __get( string $key ) {
+		return $this->get( $key );
 	}
```
felixarntz: Per https://github.com/WordPress/performance/pull/1492/files#r1752966800, I'd prefer if we removed this and continued to rely on getter methods, plus the `get( $key )` method as a somewhat okay solution for properties added by an extension.

plugins/optimization-detective/class-od-url-metrics-group-collection.php
```diff
@@ -193,7 +193,7 @@ private function create_groups(): array {
 	 */
 	public function add_url_metric( OD_URL_Metric $new_url_metric ): void {
 		foreach ( $this->groups as $group ) {
-			if ( $group->is_viewport_width_in_range( $new_url_metric->get_viewport_width() ) ) {
+			if ( $group->is_viewport_width_in_range( $new_url_metric->viewport['width'] ) ) {
```
felixarntz: In addition to what I said in https://github.com/WordPress/performance/pull/1492/files#r1752966800, this here is an example where I would even argue it was more ergonomical before. `get_viewport_width()` is more intuitive to use than `viewport['width']` since for the latter you have to know which keys are available within the viewport array. So that means it's not even always more ergonomical to use read-only properties.

plugins/optimization-detective/class-od-url-metric.php
westonruter: Actually I think the problem here is not related to `@property-read`, since this is regarding accessing keys of the items in the `elements` array. It's the same problem if using `get_elements()` with this code:

```php
foreach ( $url_metric->get_elements() as $element ) {
	if ( null !== $element['resizedBoundingClientRect'] ) {
		// ...
	}
}
```

The problem is that `$element` is typed as `ElementData` which doesn't include the `resizedBoundingClientRect` key. But this code here is assuming the `null` was provided as a `default`. Let's consider that it may not be set at all. In that case, consider this code:

```php
foreach ( $url_metric->elements as $element ) {
	if ( isset( $element['resizedBoundingClientRect'] ) ) {
		/**
		 * Resized bounding client rect.
		 *
		 * @var DOMRect $rect
		 */
		$rect = $element['resizedBoundingClientRect'];
		PHPStan\dumpType( $rect );
	}
}
```

Here we have the `isset()` check and considering that the schema has enforced the expected type for `resizedBoundingClientRect`, PHPStan here outputs:

```
Dumped type: array{width: float, height: float, x: float, y: float, top: float, right: float, bottom: float, left: float}
```

This is what we want. However, it would be ideal if we didn't have to use `@var` and PHPStan could be told directly that `$element['resizedBoundingClientRect']` is of type `DOMRect|null`.

Otherwise, for built-in root-level keys like `elements`, `uuid`, `timestamp` they will never return `null` because the schema enforces them to have a value. For the non-builtin properties, however, they indeed may not be set. So let's say there's a new root property called `cwv_metrics` which the Dashboard (https://github.com/WordPress/performance/issues/1324) may introduce. This could be accessed in this PR in the following ways:

```php
PHPStan\dumpType( $url_metric->get( 'web_vitals' ) );
```

```
Dumped type: mixed
```

No type information as expected since `get()` returns `mixed|null`. However, this with the magic getter is worse, even though `__get()` is defined the same:

```php
PHPStan\dumpType( $url_metric->web_vitals )
```

```
Access to an undefined property OD_URL_Metric::$web_vitals.
💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property  
Dumped type: *ERROR*
```

That helpful [Learn more](https://phpstan.org/blog/solving-phpstan-access-to-undefined-property) link may have the answer and I'll investigate further tomorrow. In particular, it may come down to [writing our own PHPStan extension](https://phpstan.org/blog/solving-phpstan-access-to-undefined-property#write-your-own-extension). This could ensure that PHPStan detects the correct types for both `get()` and magic getters, as well as for extended `elements` item schemas.

plugins/optimization-detective/class-od-url-metrics-group-collection.php
westonruter: Yeah, this isn't necessarily more ergonomical. Nevertheless, the IDE should know that `$new_url_metric->viewport` is of type `ViewportRect` so it should autocomplete the keys:

image

plugins/optimization-detective/class-od-url-metric.php
felixarntz: Writing a PHPStan extension sounds like overkill to me. I think we should try to make it work against its constraints.

Regarding the `mixed` return value of  the `get()` method, I don't think that's a problem. Wouldn't an extension still be able to introduce a wrapper methods where it casts it to what's expected and comes with proper return type hint and doc? I think that's sufficient to alleviate the concerns.

And for extension that don't want to go the extra mile, they can still do it the "dirty" way of directly calling `get()` wherever they need an extra property and then checking for existence and cast the value as needed.

On the other hand, undefined properties is a real problem, as reported by PHPStan. That's why I would continue to advise against using a magic getter. I just don't see the value.

westonruter: The magic getter is only relevant to the root properties. The issue exists for extensions to `element` data. An extension plugin could indeed do its own type declarations in helper wrapper functions, for example:




```php
	/**
	 * Gets the resized bounding client rect.
	 *
	 * @phpstan-param ElementData $element
	 *
	 * @param array $element Element.
	 * @return DOMRect|null Whether visible or null if unknown.
	 */
	public static function get_resized_bounding_client_rect( array $element ): ?array {
		if ( isset( $element['resizedBoundingClientRect'] ) ) {
			return $element['resizedBoundingClientRect'];
		}
		return null;
	}

	public function do_something( OD_HTML_Tag_Processor $processor, OD_URL_Metric $url_metric ): void {
		foreach ( $url_metric->get_elements() as $element ) {
			$rect = self::get_resized_bounding_client_rect( $element );
			if ( isset( $rect ) ) {
				PHPStan\dumpType( $rect['height'] ); // => Dumped type: float
			}
		}
	}
```

It would be nicer to not have to resort to `get_resized_bounding_client_rect` and just do:

```php
	public function do_something( OD_HTML_Tag_Processor $processor, OD_URL_Metric $url_metric ): void {
		foreach ( $url_metric->get_elements() as $element ) {
			if ( isset( $element['resizedBoundingClientRect'] ) ) {
				PHPStan\dumpType( $element['resizedBoundingClientRect']['height'] );
			}
		}
	}
```

But this results in `Dumped type: mixed`.

Informing PHPStan how to resolve the extended schema would ensure the expected types are used without having to duplicate typing information which is already being enforced by validation and sanitization with the schema.

For root properties it also happens, without magic getters:

```php
	/**
	 * Gets the CWV metrics.
	 *
	 * @param OD_URL_Metric $url_metric URL Metric.
	 * @return array{ lcp: float, inp: ?float, cls: float }|null CWV data or null if not set.
	 */
	public static function get_cwv_metrics( OD_URL_Metric $url_metric ): ?array {
		$cwv_metrics = $url_metric->get( 'cwv_metrics' );
		if ( is_array( $cwv_metrics ) ) {
			/**
			 * CWV metrics.
			 *
			 * @var array{ lcp: float, inp: ?float, cls: float } $cwv_metrics
			 */
			return $cwv_metrics;
		}
		return null;
	}

	public function do_something( OD_URL_Metric $url_metric ): void {
		$cwv_metrics = self::get_cwv_metrics( $url_metric );
		if ( isset( $cwv_metrics ) ) {
			PHPStan\dumpType( $cwv_metrics['lcp'] ); // => Dumped type: float
		}
	}
```

It would be preferable to just have to do:

```php
	public function do_something( OD_URL_Metric $url_metric ): void {
		$cwv_metrics = $url_metric->get( 'cwv_metrics' );
		if ( isset( $cwv_metrics ) ) {
			PHPStan\dumpType( $cwv_metrics['lcp'] );
		}
	}
```

But again, this results in `Dumped type: mixed`.

I don't think the use of a magic getter is necessarily a problem. I believe static analysis would be able to deal with them. But it isn't necessary. I've removed it in 0db296f4. I would still like to explore the use of a PHPStan extension. It's designed to be extensible for cases like this.

plugins/optimization-detective/class-od-url-metrics-group-collection.php
westonruter: Undone in 0db296f4

felixarntz:
@westonruter I opened #1532 because of the failing tests unrelated to this PR.

westonruter:
@felixarntz I don't think what I want to [explore](https://github.com/WordPress/performance/pull/1492#discussion_r1755740489) further with a PHPStan extension to add type information from the extended schema need block this PR from being merged. So if you're happy with this PR as-is, we can merge.

felixarntz (approved):

PR #1494, opened 2024-08-22
Introduce viewport aspect ratio validation for URL Metrics

westonruter:
As mentioned in #1490, I was debugging URL Metrics on my own blog because I found that commenter avatars weren't getting `loading=lazy` even though they are far outside the initial viewport. When looking the URL Metrics for the URL, I was surprised to find a URL Metric with this viewport:

```json
"viewport": {
    "width": 1024,
    "height": 12140
}
```

This is an aspect ratio of `0.084` which seems incredibly unusual. It turns out that Googlebot actually crawls pages with a viewport set so extremely high like this (cf. [GoogleBot Crawls & Renders Tall, With 9000px High Viewport?](https://www.seroundtable.com/googlebot-9000px-high-viewport-24727.html)). The result is that when computing the `all_element_max_intersection_ratios`, since this URL Metric from Googlebot always has the entire page in the viewport, they will always be `1`. This means that `loading=lazy` would get removed from all images whenever there is a URL Metric stored from Googlebot for a page. Additionally, it could be that a large image outside of the viewport for normal visitors could end up getting identified as the LCP image due to it being in the viewport for Googlebot.

This is not just an issue related to Googlebot. It's also an issue with users visiting a site with unusual window sizes (e.g. split to the top half the monitor) which can skew the URL Metrics.

This PR introduces viewport aspect ratio validation when storing a URL Metric. The validation is also done at read time, so any existing URL Metrics with bad viewport aspect ratios will be immediately discarded.

The min/max aspect ratios I picked are to accommodate the device (Sony Xperia 1 ) with the most extreme aspect ratio I'm familiar with: 21:9. When in horizontal orientation, the aspect ratio is 2.33 and when in vertical orientation it is 0.43. So I picked 0.4 as the minimum and 2.5 as the maximum.

There are two new filters introduced to customize the aspect ratio range:

* `od_minimum_viewport_aspect_ratio`
* `od_maximum_viewport_aspect_ratio`

The `od_maximum_viewport_aspect_ratio` filter is particularly useful during development for example when you have DevTools open on the bottom of the screen.

When an error occurs when parsing URL Metrics out of the `od_url_metrics` post type, it now includes the UUID in the error message. Additionally, if the error is due to a JSON validation issue, the severity is now reduced from warning to notice, since JSON Schema changes are to be expected.

### DIFF
--- a/plugins/optimization-detective/class-od-url-metric.php
+++ b/plugins/optimization-detective/class-od-url-metric.php
@@ -86,6 +86,25 @@ final class OD_URL_Metric implements JsonSerializable {
 		if ( is_wp_error( $valid ) ) {
 			throw new OD_Data_Validation_Exception( esc_html( $valid->get_error_message() ) );
 		}
+		$aspect_ratio     = $data['viewport']['width'] / $data['viewport']['height'];
+		$min_aspect_ratio = od_get_minimum_viewport_aspect_ratio();
+		$max_aspect_ratio = od_get_maximum_viewport_aspect_ratio();
+		if (
+			$aspect_ratio < $min_aspect_ratio ||
+			$aspect_ratio > $max_aspect_ratio
+		) {
+			throw new OD_Data_Validation_Exception(
+				esc_html(
+					sprintf(
+						/* translators: 1: current aspect ratio, 2: minimum aspect ratio, 3: maximum aspect ratio */
+						__( 'Viewport aspect ratio (%1$s) is in the accepted range of %2$s to %3$s.', 'optimization-detective' ),
+						$aspect_ratio,
+						$min_aspect_ratio,
+						$max_aspect_ratio
+					)
+				)
+			);
+		}
 	}
 
 	/**
--- a/plugins/optimization-detective/class-od-url-metric.php
+++ b/plugins/optimization-detective/class-od-url-metric.php
@@ -97,7 +97,7 @@ final class OD_URL_Metric implements JsonSerializable {
 				esc_html(
 					sprintf(
 						/* translators: 1: current aspect ratio, 2: minimum aspect ratio, 3: maximum aspect ratio */
-						__( 'Viewport aspect ratio (%1$s) is in the accepted range of %2$s to %3$s.', 'optimization-detective' ),
+						__( 'Viewport aspect ratio (%1$s) is not in the accepted range of %2$s to %3$s.', 'optimization-detective' ),
 						$aspect_ratio,
 						$min_aspect_ratio,
 						$max_aspect_ratio
--- a/plugins/optimization-detective/detect.js
+++ b/plugins/optimization-detective/detect.js
@@ -202,7 +202,7 @@ export default async function detect( {
 	) {
 		if ( isDebug ) {
 			warn(
-				`Viewport aspect ratio (${ aspectRatio }) is in the accepted range of ${ minViewportAspectRatio } to ${ maxViewportAspectRatio }.`
+				`Viewport aspect ratio (${ aspectRatio }) is not in the accepted range of ${ minViewportAspectRatio } to ${ maxViewportAspectRatio }.`
 			);
 		}
 		return;
--- a/plugins/optimization-detective/detect.js
+++ b/plugins/optimization-detective/detect.js
@@ -138,6 +138,8 @@ function getCurrentTime() {
  * @param {Object}                  args                             Args.
  * @param {number}                  args.serveTime                   The serve time of the page in milliseconds from PHP via `microtime( true ) * 1000`.
  * @param {number}                  args.detectionTimeWindow         The number of milliseconds between now and when the page was first generated in which detection should proceed.
+ * @param {number}                  args.minViewportAspectRatio      Minimum aspect ratio allowed for the viewport.
+ * @param {number}                  args.maxViewportAspectRatio      Maximum aspect ratio allowed for the viewport.
  * @param {boolean}                 args.isDebug                     Whether to show debug messages.
  * @param {string}                  args.restApiEndpoint             URL for where to send the detection data.
  * @param {string}                  args.restApiNonce                Nonce for writing to the REST API.
@@ -152,6 +154,8 @@ function getCurrentTime() {
 export default async function detect( {
 	serveTime,
 	detectionTimeWindow,
+	minViewportAspectRatio,
+	maxViewportAspectRatio,
 	isDebug,
 	restApiEndpoint,
 	restApiNonce,
@@ -186,6 +190,20 @@ export default async function detect( {
 	if ( ! isViewportNeeded( win.innerWidth, urlMetricsGroupStatuses ) ) {
 		if ( isDebug ) {
 			log( 'No need for URL metrics from the current viewport.' );
+		}
+		return;
+	}
+
+	// Abort if the viewport aspect ratio is not in a common range.
+	const aspectRatio = win.innerWidth / win.innerHeight;
+	if (
+		aspectRatio < minViewportAspectRatio ||
+		aspectRatio > maxViewportAspectRatio
+	) {
+		if ( isDebug ) {
+			warn(
+				`Viewport aspect ratio (${ aspectRatio }) is in the accepted range of ${ minViewportAspectRatio } to ${ maxViewportAspectRatio }.`
+			);
 		}
 		return;
 	}
--- a/plugins/optimization-detective/readme.txt
+++ b/plugins/optimization-detective/readme.txt
@@ -96,6 +96,29 @@ add_filter( 'od_url_metric_freshness_ttl', '__return_zero' );
 
 Filters the time window between serve time and run time in which loading detection is allowed to run. This amount is the allowance between when the page was first generated (and perhaps cached) and when the detect function on the page is allowed to perform its detection logic and submit the request to store the results. This avoids situations in which there are missing URL Metrics in which case a site with page caching which also has a lot of traffic could result in a cache stampede.
 
+**Filter:** `od_minimum_viewport_aspect_ratio` (default: 0.4)
+
+Filters the minimum allowed viewport aspect ratio for URL metrics.
+
+The 0.4 value is intended to accommodate the phone with the greatest known aspect
+ratio at 21:9 when rotated 90 degrees to 9:21 (0.429).
+
+**Filter:** `od_maximum_viewport_aspect_ratio` (default: 2.0)
+
+Filters the maximum allowed viewport aspect ratio for URL metrics.
+
+The 2.5 value is intended to accommodate the phone with the greatest known aspect
+ratio at 21:9 (2.333).
+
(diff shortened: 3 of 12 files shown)


plugins/optimization-detective/storage/class-od-url-metrics-post-type.php
```diff
@@ -117,52 +117,62 @@ public static function get_post( string $slug ): ?WP_Post {
 	 * @return OD_URL_Metric[] URL metrics.
 	 */
 	public static function get_url_metrics_from_post( WP_Post $post ): array {
-		$this_function   = __FUNCTION__;
-		$trigger_warning = static function ( string $message ) use ( $this_function ): void {
-			wp_trigger_error( $this_function, esc_html( $message ), E_USER_WARNING );
+		$this_function = __METHOD__;
```
westonruter, line 120: The use of `__FUNCTION__` was wrong as it didn't include the class name.

plugins/optimization-detective/detect.js
```diff
@@ -190,6 +194,20 @@ export default async function detect( {
...
 
+	// Abort if the viewport aspect ratio is not in a common range.
+	const aspectRatio = win.innerWidth / win.innerHeight;
+	if (
+		aspectRatio < minViewportAspectRatio ||
+		aspectRatio > maxViewportAspectRatio
+	) {
+		if ( isDebug ) {
+			warn(
+				`Viewport aspect ratio (${ aspectRatio }) is in the accepted range of ${ minViewportAspectRatio } to ${ maxViewportAspectRatio }.`
```
swissspidy:
```suggestion
				`Viewport aspect ratio (${ aspectRatio }) is not in the accepted range of ${ minViewportAspectRatio } to ${ maxViewportAspectRatio }.`
```

plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -86,6 +86,25 @@ private function validate_data( array $data ): void {
...
+		$max_aspect_ratio = od_get_maximum_viewport_aspect_ratio();
+		if (
+			$aspect_ratio < $min_aspect_ratio ||
+			$aspect_ratio > $max_aspect_ratio
+		) {
+			throw new OD_Data_Validation_Exception(
+				esc_html(
+					sprintf(
+						/* translators: 1: current aspect ratio, 2: minimum aspect ratio, 3: maximum aspect ratio */
+						__( 'Viewport aspect ratio (%1$s) is in the accepted range of %2$s to %3$s.', 'optimization-detective' ),
```
swissspidy:
```suggestion
						__( 'Viewport aspect ratio (%1$s) is not in the accepted range of %2$s to %3$s.', 'optimization-detective' ),
```

swissspidy (approved):

plugins/optimization-detective/tests/test-class-od-url-metric.php
```diff
@@ -90,7 +90,33 @@ public function data_provider(): array {
...
+					'uuid'      => wp_generate_uuid4(),
+					'url'       => home_url( '/' ),
+					'viewport'  => array(
+						'width'  => 1000,
+						'height' => 10000,
+					),
+					'timestamp' => microtime( true ),
+					'elements'  => array(),
+				),
+				'error' => 'Viewport aspect ratio (0.1) is in the accepted range of 0.4 to 2.5.',
```
westonruter:
```suggestion
				'error' => 'Viewport aspect ratio (0.1) is not in the accepted range of 0.4 to 2.5.',
```

plugins/optimization-detective/readme.txt
```diff
@@ -96,6 +96,29 @@ add_filter( 'od_url_metric_freshness_ttl', '__return_zero' );
...
 Filters the time window between serve time and run time in which loading detection is allowed to run. This amount is the allowance between when the page was first generated (and perhaps cached) and when the detect function on the page is allowed to perform its detection logic and submit the request to store the results. This avoids situations in which there are missing URL Metrics in which case a site with page caching which also has a lot of traffic could result in a cache stampede.
 
+**Filter:** `od_minimum_viewport_aspect_ratio` (default: 0.4)
+
+Filters the minimum allowed viewport aspect ratio for URL metrics.
+
+The 0.4 value is intended to accommodate the phone with the greatest known aspect
+ratio at 21:9 when rotated 90 degrees to 9:21 (0.429).
+
+**Filter:** `od_maximum_viewport_aspect_ratio` (default: 2.0)
```
adamsilverstein:
```suggestion
**Filter:** `od_maximum_viewport_aspect_ratio` (default: 2.5)
```

plugins/optimization-detective/storage/data.php
```diff
@@ -179,6 +179,50 @@ function od_verify_url_metrics_storage_nonce( string $nonce, string $slug, strin
...
+ * @since n.e.x.t
+ * @access private
+ *
+ * @return float Minimum viewport aspect ratio for URL metrics.
+ */
+function od_get_minimum_viewport_aspect_ratio(): float {
+	/**
+	 * Filters the minimum allowed viewport aspect ratio for URL metrics.
+	 *
+	 * The 0.4 value is intended to accommodate the phone with the greatest known aspect
```
adamsilverstein:
```suggestion
	 * The 0.4 default value is intended to accommodate the phone with the greatest known aspect
```

plugins/optimization-detective/storage/data.php
```diff
@@ -179,6 +179,50 @@ function od_verify_url_metrics_storage_nonce( string $nonce, string $slug, strin
...
+ * @since n.e.x.t
+ * @access private
+ *
+ * @return float Maximum viewport aspect ratio for URL metrics.
+ */
+function od_get_maximum_viewport_aspect_ratio(): float {
+	/**
+	 * Filters the maximum allowed viewport aspect ratio for URL metrics.
+	 *
+	 * The 2.5 value is intended to accommodate the phone with the greatest known aspect
```
adamsilverstein:
```suggestion
	 * The 2.5 default value is intended to accommodate the phone with the greatest known aspect
```
added word "default" because if these docs get referenced elsewhere the value may not be visible so it won't be clear what this refers to

adamsilverstein (approved):
Fantastic. Added some doc block suggestions

westonruter:
I deployed this to my site and immediately after deploying I saw the URL Metrics with the extreme aspect ratio removed.

This resulted in `all_element_max_intersection_ratios` being corrected from being all `1` to being the expected values, especially the commenter avatar images which now also are getting `loading=lazy` as expected.

Before:

```json
"all_element_max_intersection_ratios": {
    "/*[0][self::HTML]/*[1][self::BODY]/*[3][self::MAIN]/*[0][self::ARTICLE]/*[1][self::FIGURE]/*[0][self::DIV]/*[0][self::IMG]": 1,
    "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[1][self::ARTICLE]/*[2][self::FIGURE]/*[1][self::DIV]/*[1][self::IMG]": 1,
    "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[1][self::ARTICLE]/*[6][self::DIV]/*[1][self::DIV]/*[2][self::DIV]/*[1][self::DIV]/*[1][self::ARTICLE]/*[1][self::FOOTER]/*[1][self::DIV]/*[1][self::A]/*[1][self::IMG]": 1,
    "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[1][self::ARTICLE]/*[6][self::DIV]/*[1][self::DIV]/*[2][self::DIV]/*[2][self::DIV]/*[1][self::ARTICLE]/*[1][self::FOOTER]/*[1][self::DIV]/*[1][self::A]/*[1][self::IMG]": 1,
    "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[1][self::ARTICLE]/*[6][self::DIV]/*[1][self::DIV]/*[2][self::DIV]/*[3][self::DIV]/*[1][self::ARTICLE]/*[1][self::FOOTER]/*[1][self::DIV]/*[1][self::A]/*[1][self::IMG]": 1
}
```

After:

```json
"all_element_max_intersection_ratios": {
    "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[1][self::ARTICLE]/*[2][self::FIGURE]/*[1][self::DIV]/*[1][self::IMG]": 0.9963193535804749,
    "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[1][self::ARTICLE]/*[6][self::DIV]/*[1][self::DIV]/*[2][self::DIV]/*[1][self::DIV]/*[1][self::ARTICLE]/*[1][self::FOOTER]/*[1][self::DIV]/*[1][self::A]/*[1][self::IMG]": 0,
    "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[1][self::ARTICLE]/*[6][self::DIV]/*[1][self::DIV]/*[2][self::DIV]/*[2][self::DIV]/*[1][self::ARTICLE]/*[1][self::FOOTER]/*[1][self::DIV]/*[1][self::A]/*[1][self::IMG]": 0,
    "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[1][self::ARTICLE]/*[6][self::DIV]/*[1][self::DIV]/*[2][self::DIV]/*[3][self::DIV]/*[1][self::ARTICLE]/*[1][self::FOOTER]/*[1][self::DIV]/*[1][self::A]/*[1][self::IMG]": 0
}
```

westonruter:
Aside: I'm not sure what this element is all about:

```
/*[0][self::HTML]/*[1][self::BODY]/*[3][self::MAIN]/*[0][self::ARTICLE]/*[1][self::FIGURE]/*[0][self::DIV]/*[0][self::IMG]
```

I'm currently fetching the page on my site every 5-10 minutes with `WP_DEBU` on and storing the results in a directory in a timestamped HTML file. I'll look at the HTML next week to see if I'm seeing any instability in the HTML structure. This is with the Twenty Twenty theme.

westonruter:
Oh, I just realized what it is. It's a very old URL Metric from before the XPath indices were fixed to be 1-based instead of 0-based! See #1191. Looking at the URL Metric data, I see the timestamp is `1712354502.752218` which is for a day back in April! This URL Metric wasn't for Googlebot as it had a viewport of 412x823. It appears that URL Metrics are not being collected frequently enough, or my site is truly devoid of traffic. I think for my site it may be in part due to page caching, where the detection time window (`od_detection_time_window`) is too short, so it is only collecting metrics for visitors who happen to visit when there is a MISS to the page cache, and if there is a stale-while-revalidate caching strategy going on with Varnish then it is even more likely for the visitor to never be within the detection time window (`currentTime - serveTime > detectionTimeWindow`):

https://github.com/WordPress/performance/blob/f8dad3497f6240e2975733c3b277e91e35b254a6/plugins/optimization-detective/detect.js#L179-L187

I think this logic needs to be revisited.

**Update:** Filed as https://github.com/WordPress/performance/issues/1496

PR #1495, opened 2024-08-22
Support replacement of n.e.x.t in `_doing_it_wrong()` version strings

westonruter:
See https://github.com/WordPress/performance/pull/1492/files#r1726503630

Given the following PHP code:

```php
		$doing_it_wrong = static function ( string $message ) use ( $filter_name ): void {
			_doing_it_wrong(
				esc_html( "add_filter({$filter_name},...)" ),
				esc_html( $message ),
				'Optimization Detective n.e.x.t'
			);
		};
```

The `n.e.x.t` should in this context be automatically replaced with the next version, in the same way that is done for `@since` tags. This is implemented in this PR.

### DIFF
--- a/bin/plugin/commands/since.js
+++ b/bin/plugin/commands/since.js
@@ -79,7 +79,7 @@ exports.handler = async ( opt ) => {
 
 		const regexps = [
 			/(@since\s+)n\.e\.x\.t/g,
-			/('[^']+?)n\.e\.x\.t(?=')/g,
+			/('[^']*?)n\.e\.x\.t(?=')/g,
 		];
 
 		let replacementCount = 0;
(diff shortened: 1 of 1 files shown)


thelovekesh (approved):

bin/plugin/commands/since.js
```diff
@@ -77,21 +77,29 @@ exports.handler = async ( opt ) => {
 			ignore,
 		} );
 
-		const regexp = /(@since\s+)n\.e\.x\.t/g;
+		const regexps = [
+			/(@since\s+)n\.e\.x\.t/g,
+			/('[^']+?)n\.e\.x\.t(?=')/g,
```
westonruter: Maybe this is better instead:

```suggestion
			/('[^']*?)n\.e\.x\.t(?=')/g,
```

This would also match strings like `'n.e.x.t'` and not just those that start with the plugin name.

felixarntz (approved):
Great catch!

Issue #1455, opened 2024-08-09
Offloading Google Analytics (gtag) to a Web Worker


westonruter:
## Feature Description

The Web Worker Offloading plugin is now merged into trunk (https://github.com/WordPress/performance/pull/1247). However, for it to be utilized a plugin author has to manually make the `web-worker-offloading` script a dependency of the script intended to be offloaded. As mentioned in https://github.com/WordPress/performance/pull/1247#issuecomment-2249065318, we should look for opportunities to offload existing scripts automatically as much as possible. From Partytown's [Common Services](https://partytown.builder.io/common-services) page, they list out 3rd party services known to be compatible. Included on that list is Google Tag Manager, which I've found to be responsible for INP issues on highly-trafficked sites. Partytown has [documentation](https://partytown.builder.io/google-tag-manager) for how to integrate with Google Tag Manager. 

The challenge for automatically implementing the Partytown integration with Google Tag Manager is that the plugins (and themes) which add gtag do so in inconsistent ways. The scripts could be registered properly via `wp_enqueue_script()` but the handles won't be consistent. Or a plugin may manually print the script at `wp_head`. 

In the case where a plugin properly enqueues the script, we can look at WPdirectory for the most common script handles used. Otherwise, for manually-printed scripts the alternative would be for Web Worker Offloading to integrate with Optimization Detective to register a tag visitor that looks for a `SCRIPT` tag with a `src` pointing to `https://www.googletagmanager.com/gtag/js`, and when present, inject the Partytown script (if not already present) and update the `SCRIPT` tag attributes as necessary.

In looking at [WPdirectory](https://wpdirectory.net/search/01J4VXSTK5GV5889HZXS4XD2CM) for plugins with at least 100K installs, the script handles used are:

Plugin | Handle | Install Count | Source
--|--|--:|--
[WooCommerce](https://wordpress.org/plugins/woocommerce/) ([Google Analytics for WooCommerce](https://wordpress.org/plugins/woocommerce-google-analytics-integration/)) | `google-tag-manager` | 5,000,000 (200,000) | [Trac](https://plugins.trac.wordpress.org/browser/woocommerce/tags/9.3.3/src/Blocks/Domain/Services/GoogleAnalytics.php#L50) ([Trac](https://plugins.trac.wordpress.org/browser/woocommerce-google-analytics-integration/tags/2.1.7/includes/class-wc-google-gtag-js.php#L66))
[...]

PR #1497, opened 2024-08-25
Update strategy for adding `script[type="text/partytown"]` to worker scripts

### LINKED ISSUE #1468: Reconsider approach of using script dependencies to opt-in a script for worker offloading
Currently the way that a script is offloaded to a web worker is to add the `web-worker-offloading` script dependency:

```php
add_action( 'wp_enqueue_scripts', function () {
    wp_enqueue_script(
  	  'non-critical-js',
  	  '/path/to/non-critical.js',
  	  array( 'web-worker-offloading' ), // Note the `web-worker-offloading` being added as a dep
  	  '1.0.0',
  	  true
    );
} );
```

To opt-in existing scripts:

```php
add_action( 'wp_print_scripts', function () {
	$scripts = wp_scripts();
	$handle  = 'foo';
	if ( array_key_exists( $handle, $scripts->registered ) ) {
		$scripts->registered[ $handle ]->deps[] = 'web-worker-offloading';
	}
} );
```

However, I just realized a significant problem: if someone deactivates the Web Worker Offloading plugin, then the `web-worker-offloading` script won't be registered and so _the script having it asa dependency won't get printed at all._ This would then necessitate having to add conditional checks like the following before doing the above code:

```php
if ( defined( 'WEB_WORKER_OFFLOADING_VERSION' ) ) {
    // Now include the web-worker-offloading dependency.
}
```

We could make this more resilient by taking a similar approach to what we did with the script loading strategies (async/defer). In fact, offloading to a web worker could itself be considered a type of loading strategy. With script loading strategies, we utilize the `strategy` key for script data. So for example an existing script can be opted-in to the `async` strategy:

```php
wp_script_add_data( 'foo', 'strategy', 'async' );
```

We could do something similar for offloading to a worker, for example:

```php
wp_script_add_data( 'foo', 'worker', true );
```

Then when scripts are being printed, we could check if any of the pending handles have a `worker` script data key with a `true` value, and if so, print the Partytown script as well as add the `type="text/partytown"` attribute to the script being printed.

thelovekesh:
## Summary

Fixes https://github.com/WordPress/performance/issues/1468

## Relevant technical choices

- Update logic to use `wp_script_add_data( $handle, 'worker', true )` to identify worker offloading scripts instead of identifying them by `web-worker-offloading` script dependency. 
- Register `web-worker-offloading` script during `wp_default_scripts` instead of `wp_enqueue_scripts` so that it can be used in the admin and to more closely align with how scripts are registered in core.
- Use the `print_scripts_array` filter to prepend the `web-worker-offloading` script to the list of scripts being printed if one is offloaded to a worker. This is more performant than adding a script dependency at the `print_scripts_array` action since it only needs to iterate over the queued script handles as opposed to iterating over all registered scripts.


plugins/web-worker-offloading/hooks.php
```diff
@@ -68,31 +68,30 @@ function wwo_init(): void {
...
-				wp_script_add_data( $handle, 'strategy', 'async' ); // The 'defer' strategy would work as well.
-				wp_script_add_data( $handle, 'wwo_strategy_added', true );
+function wwo_update_worker_scripts_deps_and_strategy(): void {
+	foreach ( wp_scripts()->registered as $dep ) {
+		if (
+			(bool) wp_scripts()->get_data( $dep->handle, 'worker' ) &&
+			! in_array( 'web-worker-offloading', wp_scripts()->registered[ $dep->handle ]->deps, true )
+		) {
+			wp_scripts()->registered[ $dep->handle ]->deps[] = 'web-worker-offloading';
+
```
westonruter: I don't think we technically have to do this because when a script is made into a `text/partytown` script, it won't by definition do anything until the Partytown script has been loaded on the page. So the Partytown script could be located anywhere (and ideally it should always be in the footer!) But I suppose the benefit of doing this way is it would get added automatically. Still I think there could be a `wp_print_footer_scripts` action that checks if any of the enqueued scripts require WWO, and if so, it could then ultimately print the Partytown script in the footer.

```suggestion
```

plugins/web-worker-offloading/hooks.php
```diff
@@ -68,31 +68,30 @@ function wwo_init(): void {
...
+			if ( false === wp_scripts()->get_data( $dep->handle, 'strategy' ) ) {
+				wp_script_add_data( $dep->handle, 'strategy', 'async' ); // The 'defer' strategy would work as well.
+				wp_script_add_data( $dep->handle, 'wwo_strategy_added', true );
 			}
```
westonruter: We may need to end up removing this logic if we are going to rely on inline after scripts also being made Partytown scripts as mentioned in https://github.com/WordPress/performance/issues/1455. But that can be done in another PR.

plugins/web-worker-offloading/hooks.php
westonruter: I've refactored this in 0c2d7b23b70b4ede5abc30ff54e922e3594ba97d to use the `print_scripts_array` filter to inject the `web-worker-offloading` script on demand. This has slight performance benefits as it doesn't require looping over all registered scripts and can instead just look at the scripts that have been enqueued for printing.

plugins/web-worker-offloading/hooks.php
westonruter: I'll address this in another PR.

westonruter (approved):

plugins/web-worker-offloading/hooks.php
```diff
@@ -63,36 +65,40 @@ function wwo_init(): void {
 		'before'
 	);
 
-	wp_add_inline_script( 'web-worker-offloading', $partytown_js );
+	$scripts->add_inline_script( 'web-worker-offloading', $partytown_js );
 }
-add_action( 'wp_enqueue_scripts', 'wwo_init' );
+add_action( 'wp_default_scripts', 'wwo_register_default_scripts' );
```
adamsilverstein, line 70: Nice!

plugins/web-worker-offloading/hooks.php
```diff
@@ -63,36 +65,40 @@ function wwo_init(): void {
...
+ * This also marks their strategy as `async`. This is needed because scripts offloaded to a worker thread can be
+ * considered async. However, they may include `before` and `after` inline scripts that need sequential execution. Once
+ * marked as async, `filter_eligible_strategies()` determines if the script is eligible for async execution. If so, it
+ * will be offloaded to the worker thread.
  *
  * @since 0.1.0
  *
- * @param string[]|mixed $script_handles Array of script handles.
- * @return string[] Array of script handles.
+ * @param string[]|mixed $to_do An array of enqueued script dependency handles.
```
adamsilverstein: why is `$script_handles` changed to `$to_do`?

adamsilverstein (commented):
Looks great, thanks for working on this! One small question.

plugins/web-worker-offloading/hooks.php
westonruter: Ah, because this is the name of the param passed to `print_scripts_array`. But I can change it back.

westonruter: Reverted in bfb9e116a6c33cffbe729d5593e07448502ec1f2

thelovekesh:
Thanks @westonruter for taking this to finish line ❤️

Issue #1035, opened 2024-03-07
Optimization Detective: Implement lazy-loading of background images


westonruter:
## Feature Description

At the WordCamp Asia Contributor Day, @shilo-ey shared a very interesting thing that Elementor is doing to optimize the loading of background images by lazy-loading them. Background images cannot be lazy-loaded using `loading=lazy` like regular images can since they are CSS and not `img` elements. So Elementor is using a clever approach to force all background images to be hidden after the first N with a style rule. Then it uses IntersectionObserver to remove the style rule once scrolling. 

https://github.com/elementor/elementor/blob/498d48e8fed6cb8db2b0b3d3835cf1e44efbcbb8/modules/lazyload/module.php#L37-L97

This could be done with Optimization Detective as well, but in a more accurate way since it wouldn't have to assume to not lazy-load the first N images and it can instead use the elements in the URL metrics to know which background images are actually in the initial viewport, and only exclude them from lazy-loading.

See also https://github.com/WordPress/gutenberg/issues/59113

When JavaScript is turned off, the lazy-loading of the background image should ideally be disabled. This can be achieved by wrapping the CSS rules to disable the background images outside the initial viewport with the [`scripting` media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/scripting):

```css
@media (scripting: enabled) {
   /*...*/
}
```

westonruter:
See also this implementation that @erikyo just shared (see [Slack thread](https://wordpress.slack.com/archives/C02KGN5K076/p1711359497883509?thread_ts=1707178551.337679&cid=C02KGN5K076)): https://gist.github.com/erikyo/8e38a9bdeddfc3fea315399c0107ff09

Nevertheless, we need to be careful not to lazy-load a background image that is in the initial viewport.

erikyo:
I'd like to provide an update to this thread since my exploration has progressed significantly since that initial gist!

My aim was to:
- Implement lazy loading for backgrounds, videos, and all other elements requiring lazy loading within the page.
- Achieve this without the need for any custom or special markup. Hence, approaches such as using [CSS backgrounds](https://web.dev/articles/lazy-loading-images) or [data-lazy attributes](https://web.dev/articles/lazy-loading-images#images-inline-intersection-observer) are not considered viable solutions.
- Detect the position of elements within the viewport and prevent lazy loading on those items.

[...]

PR #1498, opened 2024-08-25
Prioritize loading poster image of video LCP elements

### LINKED ISSUE #1183: Prioritize loading poster image of video LCP elements

thelovekesh:
## Summary

Fixes #1183

## Relevant technical choices

- Preload poster images with high-priority video LCP elements.


thelovekesh:
_TODO: Add test cases._

westonruter:
Yeah, sounds good.

plugins/image-prioritizer/helper.php
```diff
@@ -36,4 +36,7 @@ function image_prioritizer_register_tag_visitors( OD_Tag_Visitor_Registry $regis
 
 	$bg_image_visitor = new Image_Prioritizer_Background_Image_Styled_Tag_Visitor();
 	$registry->register( 'bg-image-tags', $bg_image_visitor );
+
+	$video_visitor = new Image_Prioritizer_Video_Tag_Visitor();
+	$registry->register( 'video-tags', $video_visitor );
```
westonruter: Let's call this something like `image-prioritizer-video-poster-preloading`. The names I had chosen previously are too generic.

plugins/image-prioritizer/class-image-prioritizer-tag-visitor.php
```diff
@@ -36,4 +38,24 @@ abstract public function __invoke( OD_Tag_Visitor_Context $context ): bool;
...
+	 * @since n.e.x.t
+	 * @todo Move this into the OD_HTML_Tag_Processor/OD_HTML_Processor class eventually.
+	 *
+	 * @phpstan-param NormalizedAttributeNames $attribute_name
+	 *
+	 * @param OD_HTML_Tag_Processor $processor      Processor.
+	 * @param string                $attribute_name Attribute name.
+	 * @return string|true|null Normalized attribute value.
+	 */
+	protected function get_attribute_value( OD_HTML_Tag_Processor $processor, string $attribute_name ) {
```
westonruter, line 54: I'd like to move this eventually into `OD_HTML_Tag_Processor`/`OD_HTML_Processor`. Also @dmsnell has suggested this could eventually be part of the core class as well, as most HTML attributes have their own dedicated logic to get back a normalized value.

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -0,0 +1,72 @@
...
+final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Visitor {
+
+	/**
+	 * Visits a tag.
+	 *
+	 * @param OD_Tag_Visitor_Context $context Tag visitor context.
+	 *
+	 * @return bool Whether the tag should be tracked in URL metrics.
+	 */
+	public function __invoke( OD_Tag_Visitor_Context $context ): bool {
```
swissspidy, line 32: @westonruter I suppose this is where I would add the logic for #1575?

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -0,0 +1,72 @@
...
+
+		// Skip empty poster attributes and data: URLs.
+		$poster = trim( (string) $processor->get_attribute( 'poster' ) );
+		if ( '' === $poster || $this->is_data_url( $poster ) ) {
+			return false;
+		}
+
+		$xpath = $processor->get_xpath();
+
+		// TODO: If $context->url_metric_group_collection->get_element_max_intersection_ratio( $xpath ) is 0.0, then the video is not in any initial viewport and the VIDEO tag could get the preload=none attribute added.
```
swissspidy, line 46: Thinking out loud for a future enhancement:

[`autoplay` takes precedence over `preload`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#preload). Since we here know that the video is not in the initial viewport, we could remove `autoplay` in favor of an `IntersectionObserver`-based implementation that only autoplays the video when actually visible.

Thoughts? Worth creating an issue for it?

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
westonruter: Yes exactly!

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
westonruter: Oh, interesting idea. Yes, I think that warrants an issue. 

Related to #1035 which involves `IntersectionObserver` to lazy-load background images.

swissspidy (approved):
Overall looks pretty good, seems like a great base for further enhancements like #1575

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
swissspidy: Done: #1590

PR #1509, opened 2024-08-27
Use on-device AI to predict links users are most likely to visit

adamsilverstein:
## Summary

Predict and prerender most likely links using on device AI.


SandiyosDev:
this is what google's up to? crazy

PR #1533, opened 2024-09-12
Fix unit tests for webp-uploads with WordPress 6.7-alpha after sizes=auto landed

### LINKED ISSUE #1532: Fix failing tests with WordPress `trunk`
## Bug Description

See e.g. https://github.com/WordPress/performance/actions/runs/10820446352/job/30020524054?pr=1492: The PHPUnit tests fail, but that's entirely independent of that PR.

It looks like these failures are because we're looking at the _exact_ `img` tag markup for the "Modern Image Formats" plugin, which of course was changed as of [committing `sizes="auto"` to Core](https://core.trac.wordpress.org/changeset/59008).

## Steps to reproduce

You should see those failures as soon as you open any PR, or run unit tests locally against WordPress trunk.

westonruter:
## Summary

This fixes unit tests in 6.7-alpha after sizes=auto landed in core by avoiding `sizes=auto` from being added in the first place by making all the tested images not lazy-loaded.

Fixes #1532

### DIFF
--- a/plugins/webp-uploads/tests/test-picture-element.php
+++ b/plugins/webp-uploads/tests/test-picture-element.php
@@ -77,8 +77,10 @@ class Test_WebP_Uploads_Picture_Element extends TestCase {
 			'large',
 			false,
 			array(
-				'class' => 'wp-image-' . self::$image_id,
-				'alt'   => 'Green Leaves',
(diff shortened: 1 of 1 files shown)


felixarntz (approved):
Thanks @westonruter!

PR #1546, opened 2024-09-19
Try: Use WP_HTML_Processor in Optimization Detective

westonruter:
This solves the problem with HTML Tag Processor failing to parse malformed HTML: https://wordpress.org/support/topic/optimization-detective-floods-debug-log-od_html_tag_processornext_token/

Blockers:

- [ ] #1861 
- [x] ~Core-62269 where `WP_HTML_Processor::next_token()` cannot be extended in subclasses to keep track of state.~
- [x] ~Core-62290 where calls to `WP_HTML_Processor::seek()` fail.~
- [ ] Core-62270 in which a bookmark cannot currently be set on `</body>`. But there are [workarounds](https://core.trac.wordpress.org/ticket/62270#comment:5).
- [ ] Core-63020 which may be a nice-to-have.

Related: https://github.com/WordPress/performance/pull/1919#issuecomment-2725705280. We should consider not directly exposing the `OD_HTML_Tag_Processor` to the tag visitors and template optimization start/finish action handlers since this then exposes the `append_head_html()` and `append_body_html()` methods which may not be desirable. This also exposes a lower-level API than which may be appropriate which will complicate switching between using `WP_HTML_Tag_Processor` and `WP_HTML_Processor` implementations. Namely, for `WP_HTML_Processor` there may not be a need to add a `get_xpath()` method directly to the class, and we may not need to subclass it at all. If we instead passed a wrapper class, then we could expose a `get_xpath()` on the wrapper subclass which is then able to use the appropriate implementation based on whether it is using `WP_HTML_Tag_Processor` or `WP_HTML_Processor`. In fact, this could allow Optimization Detective to use an unsubclassed original `WP_HTML_Tag_Processor` instance which is provided by core for output-buffer manipulation instead of allowing filtering the underlying string (if that is so desired, per Core-43258).


plugins/optimization-detective/class-od-html-processor.php
```diff
@@ -0,0 +1,418 @@
...
+	exit;
+}
+
+/**
+ * Extension to WP_HTML_Processor that supports injecting HTML and obtaining XPath for the current tag.
+ *
+ * @since n.e.x.t
+ * @access private
+ */
+final class OD_HTML_Processor extends WP_HTML_Processor {
```
westonruter, line 20: Uh, where is this file even being loaded?

plugins/optimization-detective/optimization.php
```diff
@@ -177,7 +177,14 @@ function od_optimize_template_output_buffer( string $buffer ): string {
 	}
 
 	// If the initial tag is not an open HTML tag, then abort since the buffer is not a complete HTML document.
-	$processor = new OD_HTML_Tag_Processor( $buffer );
+	if ( version_compare( get_bloginfo( 'version' ), strtok( '6.7', '-' ), '>=' ) ) {
```
westonruter:
```suggestion
	if ( version_compare( strtok( get_bloginfo( 'version' ), ','), '6.7' ), '>=' ) ) {
```

westonruter:
This is now unblocked for further experimentation, although one [workaround](https://core.trac.wordpress.org/ticket/62270#comment:5) is still needed.

PR #1547, opened 2024-09-19
Update auto sizes logic in Enhanced Responsive Images plugin to no longer load if already in Core

### LINKED ISSUE #1531: Update auto sizes logic in Enhanced Responsive Images plugin to no longer load if already in Core
Per the core code being merged in https://core.trac.wordpress.org/changeset/59008, we need to modify the Enhanced Responsive Images plugin to ensure the relevant logic is only loaded if not already available in Core, for compatibility with the upcoming 6.7 release.

Maybe a good way would be to check for existence of the `wp_sizes_attribute_includes_valid_auto()` function?

We still have some time to do that, but good to have an issue already. We could get started on this at any point prior to the WP 6.7 release.

mukeshpanchal27:
## Summary

Fixes #1531

### DIFF
--- a/plugins/auto-sizes/hooks.php
+++ b/plugins/auto-sizes/hooks.php
@@ -42,7 +42,6 @@ function auto_sizes_update_image_attributes( $attr ): array {
 
 	return $attr;
 }
-add_filter( 'wp_get_attachment_image_attributes', 'auto_sizes_update_image_attributes' );
 
 /**
  * Adds auto to the sizes attribute to the image, if applicable.
@@ -85,7 +84,12 @@ function auto_sizes_update_content_img_tag( $html ): string {
 	$processor->set_attribute( 'sizes', "auto, $sizes" );
 	return $processor->get_updated_html();
 }
-add_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
+
+// Skip loading plugin filters if the WordPress core is already loaded auto sizes.
+if ( ! function_exists( 'wp_sizes_attribute_includes_valid_auto' ) ) {
+	add_filter( 'wp_get_attachment_image_attributes', 'auto_sizes_update_image_attributes' );
+	add_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
+}
 
 /**
  * Checks whether the given 'sizes' attribute includes the 'auto' keyword as the first item in the list.
--- a/plugins/auto-sizes/hooks.php
+++ b/plugins/auto-sizes/hooks.php
@@ -85,7 +85,7 @@ function auto_sizes_update_content_img_tag( $html ): string {
 	return $processor->get_updated_html();
 }
 
-// Skip loading plugin filters if the WordPress core is already loaded auto sizes.
+// Skip loading plugin filters if WordPress Core already loaded the functionality.
 if ( ! function_exists( 'wp_sizes_attribute_includes_valid_auto' ) ) {
 	add_filter( 'wp_get_attachment_image_attributes', 'auto_sizes_update_image_attributes' );
 	add_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
--- a/plugins/auto-sizes/tests/test-auto-sizes.php
+++ b/plugins/auto-sizes/tests/test-auto-sizes.php
@@ -32,6 +32,12 @@ class Test_AutoSizes extends WP_UnitTestCase {
 		return get_image_tag( $attachment_id, '', '', '', 'large' );
 	}
 
+	public function test_hooks(): void {
+		$this->assertSame( function_exists( 'wp_sizes_attribute_includes_valid_auto' ) ? false : 10, has_filter( 'wp_get_attachment_image_attributes', 'auto_sizes_update_image_attributes' ) );
+		$this->assertSame( function_exists( 'wp_sizes_attribute_includes_valid_auto' ) ? false : 10, has_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' ) );
+		$this->assertSame( 10, has_action( 'wp_head', 'auto_sizes_render_generator' ) );
+	}
+
 	/**
 	 * Test generated markup for an image with lazy loading gets auto-sizes.
 	 *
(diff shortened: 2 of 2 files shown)


plugins/auto-sizes/tests/test-auto-sizes.php
```diff
@@ -33,8 +33,11 @@ public function get_image_tag( int $attachment_id ): string {
 	}
 
 	public function test_hooks(): void {
-		$this->assertSame( 10, has_filter( 'wp_get_attachment_image_attributes', 'auto_sizes_update_image_attributes' ) );
-		$this->assertSame( 10, has_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' ) );
+		// Skip loading plugin filters if the WordPress core is already loaded auto sizes.
+		if ( ! function_exists( 'wp_sizes_attribute_includes_valid_auto' ) ) {
+			$this->assertSame( 10, has_filter( 'wp_get_attachment_image_attributes', 'auto_sizes_update_image_attributes' ) );
+			$this->assertSame( 10, has_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' ) );
+		}
```
westonruter:
```suggestion
		$this->assertSame( function_exists( 'wp_sizes_attribute_includes_valid_auto' ) ? false : 10, has_filter( 'wp_get_attachment_image_attributes', 'auto_sizes_update_image_attributes' ) );
		$this->assertSame( function_exists( 'wp_sizes_attribute_includes_valid_auto' ) ? false : 10, has_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' ) );
```

westonruter (approved):
LGTM, with a minor suggestion.

plugins/auto-sizes/tests/test-auto-sizes.php
joemcgill: I think we can remove these assertions altogether. We have tests that call the functions that those filters affect, so those tests would already fail if our filter callbacks were not hooked to those functions.

We could also remove the `wp_head` hook test if we rewrote `test_auto_sizes_render_generator` to call `wp_head()` and check that our meta tags were included.

joemcgill (approved):
One suggestion, but this will get the job done.

plugins/auto-sizes/tests/test-auto-sizes.php
mukeshpanchal27: Thanks! Addressed in https://github.com/WordPress/performance/pull/1547/commits/a7410d292a974095402b1b5e160609a7ceda8f1d

plugins/auto-sizes/hooks.php
```diff
@@ -85,7 +84,12 @@ function auto_sizes_update_content_img_tag( $html ): string {
 	$processor->set_attribute( 'sizes', "auto, $sizes" );
 	return $processor->get_updated_html();
 }
-add_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
+
+// Skip loading plugin filters if the WordPress core is already loaded auto sizes.
```
felixarntz:
```suggestion
// Skip loading plugin filters if WordPress Core already loaded the functionality.
```

plugins/auto-sizes/tests/test-auto-sizes.php
```diff
@@ -32,12 +32,6 @@ public function get_image_tag( int $attachment_id ): string {
 		return get_image_tag( $attachment_id, '', '', '', 'large' );
 	}
 
-	public function test_hooks(): void {
-		$this->assertSame( 10, has_filter( 'wp_get_attachment_image_attributes', 'auto_sizes_update_image_attributes' ) );
-		$this->assertSame( 10, has_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' ) );
-		$this->assertSame( 10, has_action( 'wp_head', 'auto_sizes_render_generator' ) );
-	}
```
felixarntz: Nit-pick, but technically it would be more appropriate to keep this test but make it behave differently based on whether the Core function exists. If it does not exist, the same assertions apply like now. If it does exist, you could assert that the first two filters are _not_ added.

Note also that the generator filter is relevant either way, so that's another reason not to remove this test.

felixarntz (approved):
@mukeshpanchal27 This looks good. While the test is not super important to have (as it tests something very simple), technically it shouldn't be removed as it's still relevant. See feedback below.

plugins/auto-sizes/tests/test-auto-sizes.php
westonruter: As I suggested in https://github.com/WordPress/performance/pull/1547#discussion_r1768975834 correct?

mukeshpanchal27: Re-added in https://github.com/WordPress/performance/pull/1547/commits/de6022ca0f6d1515556f0f9b856e308ea74081bc

plugins/auto-sizes/tests/test-auto-sizes.php
```diff
@@ -206,13 +206,11 @@ public function data_image_with_existing_auto_sizes(): array {
...
-	 *
-	 * @covers ::auto_sizes_render_generator
+	 * Test auto-sizes meta tag present in wp_head hook.
 	 */
-	public function test_auto_sizes_render_generator(): void {
-		$tag = get_echo( 'auto_sizes_render_generator' );
-		$this->assertStringStartsWith( '<meta', $tag );
+	public function test_auto_sizes_meta_tag_present_in_wp_head_hook(): void {
+		$tag = get_echo( 'wp_head' );
+		$this->assertStringContainsString( '<meta', $tag );
```
westonruter: I think this change can be reverted as well.

felixarntz: Either way works for me. With `wp_head`, it's more of an integration test, with `auto_sizes_render_generator`, it's more of a unit test.

Since we already test the action hook above, I slightly lean towards reverting as Weston suggested.

felixarntz (approved):
Thanks @mukeshpanchal27!

plugins/auto-sizes/tests/test-auto-sizes.php
mukeshpanchal27: Reverted in https://github.com/WordPress/performance/pull/1547/commits/791c5a2ec412cdbb22532b9304075e494c3d1e08

westonruter (approved):

PR #1562, opened 2024-09-25
Use a single WordPress.org API request to get information for all plugins

### LINKED ISSUE #1542: Use a single WordPress.org API request to get plugin information
Currently, the Performance features screen makes individual `plugin_information` queries to the WordPress.org API, which is rather inefficient, especially the more plugins we have. We can enhance this by instead using a single request as follows:
```
https://api.wordpress.org/plugins/info/1.2/?action=query_plugins&request[author]=wordpressdotorg&request[per_page]=100
```

While the WordPress.org Plugins API does not support passing multiple specific slugs (which would be ideal), this is a decent workaround: Since all our plugins use the `wordpressdotorg` author, we can safely assume our plugins are included in that list.

This will obviously return many other plugins too, but we'd be able to filter out just the plugins we care about. While this request returns too much data, it's still far more efficient than making all those individual requests.

narenin:
## Summary

In this PR we have implemented a single `WordPress.org API` request to get plugin information because currently, the Performance features screen makes individual plugin_information queries to the WordPress.org API, which is rather inefficient, especially the more plugins we have.

`https://api.wordpress.org/plugins/info/1.2/?action=query_plugins&request[author]=wordpressdotorg&request[tag]=performance&request[per_page]=100`

Fixes #1542

### DIFF
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -52,7 +52,7 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 	}
 
 	/* Check if the response contains plugins. */
-	if ( ! is_object( $response ) ) {
+	if ( ! ( is_object( $response ) && property_exists( $response, 'plugins' ) ) ) {
 		return new WP_Error( 'no_plugins', __( 'No plugins found in the API response.', 'performance-lab' ) );
 	}
 
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -30,7 +30,7 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 		return $plugins[ $plugin_slug ]; // Return cached plugin info if found.
 	}
 
-	/* Proceed with API request since no cache hit. */
+	// Proceed with API request since no cache hit.
 	$response = plugins_api(
 		'query_plugins',
 		array(
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -33,7 +33,7 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 	);
 
 	if ( is_array( $plugins ) ) {
-		/* If the specific plugin_slug is not in the cache, return an error. */
+		// If the specific plugin_slug is not in the cache, return an error.
 		if ( ! isset( $plugins[ $plugin_slug ] ) ) {
 			return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'performance-lab' ) );
 		}
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -41,7 +41,7 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 
 	$plugins = array();
 	foreach ( $data['plugins'] as $plugin_data ) {
-		$plugins[ $plugin_data['slug'] ] = wp_array_slice_assoc(
+		$plugin_info = wp_array_slice_assoc(
 			$plugin_data,
 			array(
 				'name',
@@ -54,6 +54,17 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 				'version',
 			)
 		);
+
+		// Ensure the 'requires_plugins' is always an array.
+		if ( ! isset( $plugin_info['requires_plugins'] ) || ! is_array( $plugin_info['requires_plugins'] ) ) {
+			$plugin_info['requires_plugins'] = array();
+		}
+
+		// Ensure 'requires' and 'requires_php' are either strings or false.
+		$plugin_info['requires']     = isset( $plugin_info['requires'] ) ? $plugin_info['requires'] : false;
+		$plugin_info['requires_php'] = isset( $plugin_info['requires_php'] ) ? $plugin_info['requires_php'] : false;
+
+		$plugins[ $plugin_data['slug'] ] = $plugin_info;
 	}
 
 	set_transient( $transient_key, $plugins, HOUR_IN_SECONDS );
@@ -62,6 +73,11 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 		return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'default' ) );
 	}
 
+	/**
+	 * Validated (mostly) plugin data.
+	 *
+	 * @var array{name: string, slug: string, short_description: string, requires: string|false, requires_php: string|false, requires_plugins: array<string>, download_link: string, version: string} $plugin
+	 */
 	return $plugins[ $plugin_slug ];
 }
 
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -22,8 +22,12 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 	$transient_key = 'perflab_plugins_info';
 	$plugins       = get_transient( $transient_key );
 
-	if ( is_array( $plugins ) && isset( $plugins[ $plugin_slug ] ) ) {
-		return $plugins[ $plugin_slug ];
+	if ( is_array( $plugins ) ) {
+		/* If the specific plugin_slug is not in the cache, return an error. */
+		if ( ! isset( $plugins[ $plugin_slug ] ) ) {
+			return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'performance-lab' ) );
+		}
+		return $plugins[ $plugin_slug ]; // Return cached plugin info if found
 	}
 
 	$response = wp_remote_get( 'https://api.wordpress.org/plugins/info/1.2/?action=query_plugins&request[author]=wordpressdotorg&request[tag]=performance&request[per_page]=100' );
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -30,7 +30,15 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 		return $plugins[ $plugin_slug ]; // Return cached plugin info if found.
 	}
 
-	$response = wp_remote_get( 'https://api.wordpress.org/plugins/info/1.2/?action=query_plugins&request[author]=wordpressdotorg&request[tag]=performance&request[per_page]=100' );
+	// Proceed with API request since no cache hit.
+	$response = plugins_api(
+		'query_plugins',
+		array(
+			'author'   => 'wordpressdotorg',
+			'tag'      => 'performance',
+			'per_page' => 100,
+		)
+	);
 
 	if ( is_wp_error( $response ) ) {
 		return new WP_Error(
@@ -46,12 +54,13 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 	$body = wp_remote_retrieve_body( $response );
 	$data = json_decode( $body, true );
 
-	if ( ! isset( $data['plugins'] ) || ! is_array( $data['plugins'] ) ) {
+	// Check if the response contains plugins.
+	if ( ! isset( $response->plugins ) || ! is_array( $response->plugins ) ) {
 		return new WP_Error( 'no_plugins', __( 'No plugins found in the API response.', 'performance-lab' ) );
 	}
 
 	$plugins = array();
-	foreach ( $data['plugins'] as $plugin_data ) {
+	foreach ( $response->plugins as $plugin_data ) {
 		$plugin_info = wp_array_slice_assoc(
 			$plugin_data,
 			array(
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -77,8 +77,8 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 		}
 
 		// Ensure 'requires' and 'requires_php' are either strings or false.
-		$plugin_info['requires']     = isset( $plugin_info['requires'] ) ? $plugin_info['requires'] : false;
-		$plugin_info['requires_php'] = isset( $plugin_info['requires_php'] ) ? $plugin_info['requires_php'] : false;
+		$plugin_info['requires']     = $plugin_info['requires'] ?? false;
+		$plugin_info['requires_php'] = $plugin_info['requires_php'] ?? false;
 
 		$plugins[ $plugin_data['slug'] ] = $plugin_info;
 	}
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -69,18 +69,7 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 
 	$plugins = array();
 	foreach ( $response->plugins as $plugin_data ) {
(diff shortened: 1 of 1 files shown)


plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -19,58 +19,66 @@
...
+
+	if ( ! isset( $plugins[ $plugin_slug ] ) ) {
+		return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'default' ) );
+	}
 
 	/**
 	 * Validated (mostly) plugin data.
 	 *
-	 * @var array{name: string, slug: string, short_description: string, requires: string|false, requires_php: string|false, requires_plugins: string[], download_link: string, version: string} $plugin
+	 * @var array{name: string, slug: string, short_description: string, requires: string|false, requires_php: string|false, requires_plugins: array<string>, download_link: string, version: string} $plugin
```
westonruter: I believe this will fix the PHPStan issues:

```suggestion
	 * @var array<string, array{name: string, slug: string, short_description: string, requires: string|false, requires_php: string|false, requires_plugins: string[], download_link: string, version: string}> $plugins
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -19,58 +19,66 @@
...
-		'name',
-		'slug',
-		'short_description',
-		'requires',
-		'requires_php',
-		'requires_plugins',
-		'download_link',
-		'version', // Needed by install_plugin_install_status().
-	);
+	$request = wp_remote_get( 'https://api.wordpress.org/plugins/info/1.2/?action=query_plugins&request[author]=wordpressdotorg&request[tag]=performance&request[per_page]=100' );
```
westonruter: Why not use `plugins_api()` here?

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -19,58 +19,66 @@
...
-		array(
-			'slug'   => $plugin_slug,
-			'fields' => array_fill_keys( $fields, true ),
-		)
-	);
-
-	if ( is_wp_error( $plugin ) ) {
-		return $plugin;
+	if ( is_wp_error( $request ) ) {
+		return new WP_Error( 'api_error', __( 'Failed to retrieve plugins data from WordPress.org API.', 'default' ) );
```
westonruter: Ideally the error being returned would be passed through.

This string isn't actually part of core, right? So `default` can't be used.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -19,58 +19,66 @@
...
-		'plugin_information',
-		array(
-			'slug'   => $plugin_slug,
-			'fields' => array_fill_keys( $fields, true ),
-		)
-	);
-
-	if ( is_wp_error( $plugin ) ) {
-		return $plugin;
+	if ( is_wp_error( $request ) ) {
```
westonruter: A better variable name would be `$response`, right? But I think `plugins_api()` should be used instead in any case.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -19,58 +19,66 @@
...
+		return new WP_Error( 'api_error', __( 'Failed to retrieve plugins data from WordPress.org API.', 'default' ) );
 	}
 
-	if ( is_object( $plugin ) ) {
-		$plugin = (array) $plugin;
+	$body = wp_remote_retrieve_body( $request );
+	$data = json_decode( $body, true );
+
+	if ( ! isset( $data['plugins'] ) || ! is_array( $data['plugins'] ) ) {
+		return new WP_Error( 'no_plugins', __( 'No plugins found in the API response.', 'default' ) );
```
westonruter: Again, is this string actually found in core? If not, then `default` cannot be used.

westonruter (changes requested):

plugins/performance-lab/includes/admin/plugins.php
narenin: @westonruter Thanks for reviewing this PR. I have not used `plugins_api()` because this does not support passing multiple specific slugs (which would be ideal) .  Please let me know if I am missing anything.

plugins/performance-lab/includes/admin/plugins.php
narenin: I have corrected the same, please check.

plugins/performance-lab/includes/admin/plugins.php
narenin: @westonruter I have corrected the same please check.

plugins/performance-lab/includes/admin/plugins.php
narenin: @westonruter This has been fixed.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -19,58 +19,66 @@
...
-		array(
-			'slug'   => $plugin_slug,
-			'fields' => array_fill_keys( $fields, true ),
-		)
-	);
-
-	if ( is_wp_error( $plugin ) ) {
-		return $plugin;
+	if ( is_wp_error( $response ) ) {
+		return new WP_Error( 'api_error', __( 'Failed to retrieve plugins data from WordPress.org API: ' . $request->get_error_message(), 'performance-lab' ) );
```
westonruter: The translation string cannot have any dynamic component:

```suggestion
		return new WP_Error( 'api_error', __( 'Failed to retrieve plugins data from WordPress.org API: ', 'performance-lab' ) . $request->get_error_message() );
```

plugins/performance-lab/includes/admin/plugins.php
felixarntz: `plugins_api()` is a wrapper for any API calls to `https://api.wordpress.org/plugins/`, so you can use it here too. It's only the API itself that doesn't support passing multiple specific slugs, but we're not doing that here anyway.

You can call the `plugins_api()` function with the `query_plugins` action and provide the relevant request arguments.

plugins/performance-lab/includes/admin/plugins.php
felixarntz: Even better would be to use interpolation. Not a good practice to end translation strings in whitespace.

```suggestion
		return new WP_Error(
			'api_error',
			sprintf(
				/* translators: %s: API error message */
				__( 'Failed to retrieve plugins data from WordPress.org API: %s', 'performance-lab' ),
				$request->get_error_message()
			)
		);
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -19,58 +19,73 @@
...
-		'name',
-		'slug',
-		'short_description',
-		'requires',
-		'requires_php',
-		'requires_plugins',
-		'download_link',
-		'version', // Needed by install_plugin_install_status().
-	);
+	if ( is_array( $plugins ) && isset( $plugins[ $plugin_slug ] ) ) {
```
felixarntz: The second check here is problematic. We can't check that in the same clause, otherwise a missing plugin will lead to endless re-issuing of the API request and bypassing the cache.

If the cache data is an array, we know it's a cache hit. So then we need to check _on_ that data if the `$plugin_slug` is not set, and if not return the same error as below.

plugins/performance-lab/includes/admin/plugins.php
westonruter: This seems to work fine:

```php
$response = plugins_api(
	'query_plugins',
	array(
		'author'   => 'wordpressdotorg',
		'tag'      => 'performance',
		'per_page' => 100,
	)
);
```

I can see that `$response->plugins` is populated as expected.

plugins/performance-lab/includes/admin/plugins.php
narenin: @felixarntz I have implemented this, could you please take a look.

plugins/performance-lab/includes/admin/plugins.php
narenin: @felixarntz @westonruter Thanks for the feedback, I have implemented the same.

narenin:
@westonruter @felixarntz I have implemented the feedbacks, could you please take a look and share your valuable feedback on this.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -19,58 +19,83 @@
...
+		/* If the specific plugin_slug is not in the cache, return an error. */
+		if ( ! isset( $plugins[ $plugin_slug ] ) ) {
+			return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'performance-lab' ) );
+		}
+		return $plugins[ $plugin_slug ]; // Return cached plugin info if found.
+	}
 
-	$plugin = plugins_api(
-		'plugin_information',
+	/* Proceed with API request since no cache hit. */
```
```suggestion
	// Proceed with API request since no cache hit.
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -19,58 +19,83 @@
...
+				/* translators: %s: API error message */
+				__( 'Failed to retrieve plugins data from WordPress.org API: %s', 'performance-lab' ),
+				$response->get_error_message()
+			)
+		);
 	}
 
-	if ( is_object( $plugin ) ) {
-		$plugin = (array) $plugin;
+	/* Check if the response contains plugins. */
```
```suggestion
	// Check if the response contains plugins.
```

westonruter (commented):
Minor point in comment styles

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -19,58 +19,83 @@
...
-		'plugin_information',
+	/* Proceed with API request since no cache hit. */
+	$response = plugins_api(
+		'query_plugins',
 		array(
-			'slug'   => $plugin_slug,
-			'fields' => array_fill_keys( $fields, true ),
+			'author'   => 'wordpressdotorg',
+			'tag'      => 'performance',
+			'per_page' => 100,
```
felixarntz, line 49: Does the `query_plugins` action not support passing `fields` like `plugin_information` does? We should double-check. If passing `fields` is supported, we should still do that to reduce the footprint of the request.

felixarntz (commented):
@narenin This looks almost good to go. A few last notes.

plugins/performance-lab/includes/admin/plugins.php
narenin: @felixarntz Thanks for the feedback, `query_plugins` action also support passing fields `plugin_information` so I have implemented the same, please check.

plugins/performance-lab/includes/admin/plugins.php
narenin: @westonruter Thanks for the feedback, I have implemented this one.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -41,36 +32,69 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 		'version', // Needed by install_plugin_install_status().
 	);
 
-	$plugin = plugins_api(
-		'plugin_information',
+	if ( is_array( $plugins ) ) {
+		/* If the specific plugin_slug is not in the cache, return an error. */
```
```suggestion
		// If the specific plugin_slug is not in the cache, return an error.
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -41,36 +32,69 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
+		// Ensure the 'requires_plugins' is always an array.
+		if ( ! isset( $plugin_info['requires_plugins'] ) || ! is_array( $plugin_info['requires_plugins'] ) ) {
+			$plugin_info['requires_plugins'] = array();
+		}
 
-	// Make sure all fields default to false in case another plugin is modifying the response from WordPress.org via the plugins_api filter.
-	$plugin = array_merge( array_fill_keys( $fields, false ), $plugin );
+		// Ensure 'requires' and 'requires_php' are either strings or false.
+		$plugin_info['requires']     = isset( $plugin_info['requires'] ) ? $plugin_info['requires'] : false;
+		$plugin_info['requires_php'] = isset( $plugin_info['requires_php'] ) ? $plugin_info['requires_php'] : false;
```
westonruter: This can be simplified:

```suggestion
		$plugin_info['requires']     = $plugin_info['requires'] ?? false;
		$plugin_info['requires_php'] = $plugin_info['requires_php'] ?? false;
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -41,36 +32,69 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
-	// Only store what we need.
-	$plugin = wp_array_slice_assoc( $plugin, $fields );
+	$plugins = array();
+	foreach ( $response->plugins as $plugin_data ) {
+		$plugin_info = wp_array_slice_assoc( $plugin_data, $fields );
+
+		// Ensure the 'requires_plugins' is always an array.
+		if ( ! isset( $plugin_info['requires_plugins'] ) || ! is_array( $plugin_info['requires_plugins'] ) ) {
+			$plugin_info['requires_plugins'] = array();
+		}
```
westonruter: Why is this code needed? Right below the if `$plugin_info['requires_plugins']` is not set, then it is set to `false`, which will then never occur. Apparently this `if` statement could be removed?

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -41,36 +32,69 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
 
-	// Make sure all fields default to false in case another plugin is modifying the response from WordPress.org via the plugins_api filter.
-	$plugin = array_merge( array_fill_keys( $fields, false ), $plugin );
+		// Ensure 'requires' and 'requires_php' are either strings or false.
+		$plugin_info['requires']     = isset( $plugin_info['requires'] ) ? $plugin_info['requires'] : false;
+		$plugin_info['requires_php'] = isset( $plugin_info['requires_php'] ) ? $plugin_info['requires_php'] : false;
 
-	set_transient( 'perflab_plugin_info_' . $plugin_slug, $plugin, HOUR_IN_SECONDS );
+		$plugins[ $plugin_data['slug'] ] = $plugin_info;
+	}
```
westonruter: This would be the equivalent of the existing code:

```php
	$plugins = array();
	foreach ( $response->plugins as $plugin_data ) {
		$plugins[ $plugin_data['slug'] ] = wp_array_slice_assoc( $plugin_data, $fields );
	}
```

I'm curious why the additional `requires` and `requires_php` checks are now present?

narenin:
@westonruter thanks for the suggestions, I have implemented them, please take a look.

plugins/performance-lab/includes/admin/plugins.php
narenin: Done

plugins/performance-lab/includes/admin/plugins.php
narenin: Done

plugins/performance-lab/includes/admin/plugins.php
narendraie: @westonruter This has been implemented.

plugins/performance-lab/includes/admin/plugins.php
narendraie: @westonruter This has been implemented.

westonruter (approved):
LGTM!

felixarntz (approved):
@narenin Thank you for the PR, this looks great!

narenin:
Thanks @westonruter @felixarntz for your continuous feedback and support.

PR #1563, opened 2024-09-27
Integrate with WooCommerce to offload Google Analytics to Web Worker

### LINKED ISSUE #1455: Offloading Google Analytics (gtag) to a Web Worker

westonruter:
## Summary

This integrates WooCommerce's Google Analytics integration with Partytown to offload `gtag()` to a worker. Inline scripts now no longer block a registered script from being offloaded to a worker; any associated inline scripts are also offloaded to a worker. It also improves the way that the configuration is managed and fleshes out the README for the initial release.

See #1455 (integration with Site Kit and Rank Math can be done in separate PRs)

<details><summary>Diff of Prettier-formatted page with plugin active</summary>

```diff
--- /tmp/before.html	2024-09-28 23:41:35.338249993 -0700
+++ /tmp/after.html	2024-09-28 23:44:32.403255392 -0700
@@ -1729,7 +1729,153 @@
         visibility: visible;
       }
     </style>
-    <script id="woocommerce-google-analytics-integration-gtag-js-after">
+    <script id="web-worker-offloading-js-before">
+      window.partytown = {
+        lib: "\/wp-content\/plugins\/web-worker-offloading\/build\/",
+        forward: ["dataLayer.push"],
+        debug: true,
+        mainWindowAccessors: ["wp", "ga4w"],
+        globalFns: ["gtag"],
+      };
+    </script>
+    <script id="web-worker-offloading-js-after">
+      const t = { preserveBehavior: !1 },
+        e = (e) => {
+          if ("string" == typeof e) return [e, t];
+          const [n, r = t] = e;
+          return [n, { ...t, ...r }];
+        },
+        n = Object.freeze(
+          (() => {
+            const t = new Set();
+            let e = [];
+            do {
+              Object.getOwnPropertyNames(e).forEach((n) => {
+                "function" == typeof e[n] && t.add(n);
+              });
+            } while ((e = Object.getPrototypeOf(e)) !== Object.prototype);
+            return Array.from(t);
+          })(),
+        );
+      !(function (t, r, o, i, a, s, c, d, l, p, u = t, f) {
+        function h() {
+          f ||
+            ((f = 1),
+            "/" ==
+              (c = (s.lib || "/~partytown/") + (s.debug ? "debug/" : ""))[0] &&
+              ((l = r.querySelectorAll('script[type="text/partytown"]')),
+              i != t
+                ? i.dispatchEvent(new CustomEvent("pt1", { detail: t }))
+                : ((d = setTimeout(v, 999999999)),
+                  r.addEventListener("pt0", w),
+                  a
+                    ? y(1)
+                    : o.serviceWorker
+                    ? o.serviceWorker
+                        .register(c + (s.swPath || "partytown-sw.js"), {
+                          scope: c,
+                        })
+                        .then(function (t) {
+                          t.active
+                            ? y()
+                            : t.installing &&
+                              t.installing.addEventListener(
+                                "statechange",
+                                function (t) {
+                                  "activated" == t.target.state && y();
+                                },
+                              );
+                        }, console.error)
+                    : v())));
+        }
+        function y(e) {
+          (p = r.createElement(e ? "script" : "iframe")),
+            (t._pttab = Date.now()),
+            e ||
+              ((p.style.display = "block"),
+              (p.style.width = "0"),
+              (p.style.height = "0"),
+              (p.style.border = "0"),
+              (p.style.visibility = "hidden"),
+              p.setAttribute("aria-hidden", !0)),
+            (p.src =
+              c +
+              "partytown-" +
+              (e
+                ? "atomics.js?v=0.10.2-dev1727590485751"
+                : "sandbox-sw.html?" + t._pttab)),
+            r.querySelector(s.sandboxParent || "body").appendChild(p);
+        }
+        function v(n, o) {
+          for (
+            w(),
+              i == t &&
+                (s.forward || []).map(function (n) {
+                  const [r] = e(n);
+                  delete t[r.split(".")[0]];
+                }),
+              n = 0;
+            n < l.length;
+            n++
+          )
+            ((o = r.createElement("script")).innerHTML = l[n].innerHTML),
+              (o.nonce = s.nonce),
+              r.head.appendChild(o);
+          p && p.parentNode.removeChild(p);
+        }
+        function w() {
+          clearTimeout(d);
+        }
+        (s = t.partytown || {}),
+          i == t &&
+            (s.forward || []).map(function (r) {
+              const [o, { preserveBehavior: i }] = e(r);
+              (u = t),
+                o.split(".").map(function (e, r, o) {
+                  var a;
+                  u = u[o[r]] =
+                    r + 1 < o.length
+                      ? u[o[r]] || ((a = o[r + 1]), n.includes(a) ? [] : {})
+                      : (() => {
+                          let e = null;
+                          if (i) {
+                            const { methodOrProperty: n, thisObject: r } = ((
+                              t,
+                              e,
+                            ) => {
+                              let n = t;
+                              for (let t = 0; t < e.length - 1; t += 1)
+                                n = n[e[t]];
+                              return {
+                                thisObject: n,
+                                methodOrProperty:
+                                  e.length > 0 ? n[e[e.length - 1]] : void 0,
+                              };
+                            })(t, o);
+                            "function" == typeof n &&
+                              (e = (...t) => n.apply(r, ...t));
+                          }
+                          return function () {
+                            let n;
+                            return (
+                              e && (n = e(arguments)),
+                              (t._ptf = t._ptf || []).push(o, arguments),
+                              n
+                            );
+                          };
+                        })();
+                });
+            }),
+          "complete" == r.readyState
+            ? h()
+            : (t.addEventListener("DOMContentLoaded", h),
+              t.addEventListener("load", h));
+      })(window, document, navigator, top, window.crossOriginIsolated);
+    </script>
+    <script
+      id="woocommerce-google-analytics-integration-gtag-js-after"
+      type="text/partytown"
+    >
       /* Google Analytics for WooCommerce (gtag.js) */
       window.dataLayer = window.dataLayer || [];
       function gtag() {
@@ -3637,12 +3783,13 @@
       id="wc-product-collection-block-frontend-js"
     ></script>
     <script
-      async
+      type="text/partytown"
       src="https://www.googletagmanager.com/gtag/js?id=G-12345"
       id="google-tag-manager-js"
       data-wp-strategy="async"
     ></script>
     <script
+      type="text/partytown"
       src="http://localhost:8888/wp-content/plugins/woocommerce-google-analytics-integration/assets/js/build/main.js?ver=0fcb8d9b3b96ceffc047"
       id="woocommerce-google-analytics-integration-js"
     ></script>
```

</details> 

Looking at network logs after click an Add to Cart button...

Before:

image

After:

image

As can be seen here, with the WWO plugin active, the HTTP request to Google Analytics is happening in a worker.

## How to test

1. Install [WooCommerce](https://wordpress.org/plugins/woocommerce/) 
2. Install [Google Analytics for WooCommerce](https://wordpress.org/plugins/woocommerce-google-analytics-integration/)
3. Configure a Shop with imported sample products.
4. Make the Shop live rather than serve a Coming Soon page.
5. Navigate to `/wp-admin/admin.php?page=wc-settings&tab=integration&section=google_analytics` and supply some GA ID, for example: image
6. Try clicking an Add to Cart button on the Shop page.
7. Look at the Network log and confirm that the request to google-analytics.com is being made in the Web Worker as opposed to the main thread.

Ideally you would also look at Google Analytics itself to see that the events are being tracked as expected, but this is beyond my off-hand knowledge.

## Relevant technical choices

WooCommerce needs to access `wp.i18n` from a worker-offloaded script, but unfortunately Partytown skips serializing any object properties that begin with `_`, meaning calling `wp.i18n.__()` will throw an error. I've raised this issue as https://github.com/BuilderIO/partytown/issues/629 and implemented a workaround in my fork https://github.com/westonruter/partytown/pull/1. I've updated the `@builder.io/partytown` npm dependency to use my fork.

See PR comments for additional annotations.

### DIFF
--- a/plugins/web-worker-offloading/helper.php
+++ b/plugins/web-worker-offloading/helper.php
@@ -15,13 +15,13 @@ if ( ! defined( 'ABSPATH' ) ) {
  *
  * @since 0.1.0
  * @link https://partytown.builder.io/configuration
+ * @link https://github.com/BuilderIO/partytown/blob/b292a14047a0c12ca05ba97df1833935d42fdb66/src/lib/types.ts#L393-L548
  *
- * @return array{ debug?: bool, forward?: non-empty-string[], lib: non-empty-string, loadScriptsOnMainThread?: non-empty-string[], nonce?: non-empty-string } Configuration for Partytown.
+ * @return array<string, mixed> Configuration for Partytown.
  */
 function wwo_get_configuration(): array {
 	$config = array(
-		'lib'     => wp_parse_url( plugin_dir_url( __FILE__ ), PHP_URL_PATH ) . 'build/',
-		'forward' => array(),
+		'lib' => wp_parse_url( plugin_dir_url( __FILE__ ), PHP_URL_PATH ) . 'build/',
 	);
 
 	if ( WP_DEBUG && SCRIPT_DEBUG ) {
@@ -31,10 +31,40 @@ function wwo_get_configuration(): array {
 	/**
 	 * Add configuration for Web Worker Offloading.
 	 *
+	 * Many of the configuration options are not documented publicly, so refer to the TypeScript definitions.
+	 * Additionally, not all of the configuration options (e.g. functions) can be serialized as JSON and must instead be
+	 * defined in JavaScript instead. To do so, use the following PHP code instead of filtering `wwo_configuration`:
+	 *
+	 *     add_action(
+	 *         'wp_enqueue_scripts',
+	 *         function () {
+	 *             wp_add_inline_script(
+	 *                 'web-worker-offloading',
+	 *                 <<<JS
+	 *                 window.partytown = {
+	 *                     ...(window.partytown || {}),
+	 *                     resolveUrl: (url, location, type) => {
+	 *                         if (type === 'script') {
+	 *                             const proxyUrl = new URL('https://my-reverse-proxy.com/');
+	 *                             proxyUrl.searchParams.append('url', url.href);
+	 *                             return proxyUrl;
+	 *                         }
+	 *                         return url;
+	 *                     },
+	 *                 };
+	 *                 JS,
+	 *                 'before'
+	 *             );
+	 *         }
+	 *     );
+	 *
+	 * There are also many configuration options which are not documented, so refer to the TypeScript definitions.
+	 *
 	 * @since 0.1.0
 	 * @link https://partytown.builder.io/configuration
+	 * @link https://github.com/BuilderIO/partytown/blob/b292a14047a0c12ca05ba97df1833935d42fdb66/src/lib/types.ts#L393-L548
 	 *
-	 * @param array{ debug?: bool, forward?: non-empty-string[], lib: non-empty-string, loadScriptsOnMainThread?: non-empty-string[], nonce?: non-empty-string } $config Configuration for Partytown.
+	 * @param array<string, mixed> $config Configuration for Partytown.
 	 */
 	return (array) apply_filters( 'wwo_configuration', $config );
 }
--- a/plugins/web-worker-offloading/helper.php
+++ b/plugins/web-worker-offloading/helper.php
@@ -45,7 +45,7 @@ function wwo_get_configuration(): array {
 	 *                     ...(window.partytown || {}),
 	 *                     resolveUrl: (url, location, type) => {
 	 *                         if (type === 'script') {
-	 *                             const proxyUrl = new URL('https://my-reverse-proxy.com/');
+	 *                             const proxyUrl = new URL('https://my-reverse-proxy.example.com/');
 	 *                             proxyUrl.searchParams.append('url', url.href);
 	 *                             return proxyUrl;
 	 *                         }
--- a/plugins/web-worker-offloading/helper.php
+++ b/plugins/web-worker-offloading/helper.php
@@ -36,5 +36,5 @@ function wwo_get_configuration(): array {
 	 *
 	 * @param array{ debug?: bool, forward?: non-empty-string[], lib: non-empty-string, loadScriptsOnMainThread?: non-empty-string[], nonce?: non-empty-string } $config Configuration for Partytown.
 	 */
-	return apply_filters( 'wwo_configuration', $config );
+	return (array) apply_filters( 'wwo_configuration', $config );
 }
--- a/plugins/web-worker-offloading/helper.php
+++ b/plugins/web-worker-offloading/helper.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * Helpers for Web Worker Offloading.
+ *
+ * @since 0.1.0
+ * @package web-worker-offloading
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * Gets configuration for Web Worker Offloading.
+ *
+ * @since 0.1.0
+ * @link https://partytown.builder.io/configuration
+ *
+ * @return array{ debug?: bool, forward?: non-empty-string[], lib: non-empty-string, loadScriptsOnMainThread?: non-empty-string[], nonce?: non-empty-string } Configuration for Partytown.
+ */
+function wwo_get_configuration(): array {
+	$config = array(
+		'lib'     => wp_parse_url( plugin_dir_url( __FILE__ ), PHP_URL_PATH ) . 'build/',
+		'forward' => array(),
+	);
+
+	/**
+	 * Add configuration for Web Worker Offloading.
+	 *
+	 * @since 0.1.0
+	 * @link https://partytown.builder.io/configuration
+	 *
+	 * @param array{ debug?: bool, forward?: non-empty-string[], lib: non-empty-string, loadScriptsOnMainThread?: non-empty-string[], nonce?: non-empty-string } $config Configuration for Partytown.
+	 */
+	return apply_filters( 'wwo_configuration', $config );
+}
--- a/plugins/web-worker-offloading/helper.php
+++ b/plugins/web-worker-offloading/helper.php
@@ -24,6 +24,10 @@ function wwo_get_configuration(): array {
 		'forward' => array(),
 	);
 
+	if ( WP_DEBUG && SCRIPT_DEBUG ) {
+		$config['debug'] = true;
+	}
+
 	/**
 	 * Add configuration for Web Worker Offloading.
 	 *
--- a/plugins/web-worker-offloading/hooks.php
+++ b/plugins/web-worker-offloading/hooks.php
@@ -63,12 +63,6 @@ function wwo_filter_print_scripts_array( $script_handles ): array {
 		if ( true === (bool) $scripts->get_data( $handle, 'worker' ) ) {
 			$scripts->set_group( 'web-worker-offloading', false, 0 ); // Try to print in the head.
 			array_unshift( $script_handles, 'web-worker-offloading' );
-
-			// TODO: This should be reconsidered because scripts needing to be offloaded will often have after scripts. See <https://github.com/WordPress/performance/pull/1497/files#r1733538721>.
-			if ( false === wp_scripts()->get_data( $handle, 'strategy' ) ) {
-				wp_script_add_data( $handle, 'strategy', 'async' ); // The 'defer' strategy would work as well.
-				wp_script_add_data( $handle, 'wwo_strategy_added', true );
-			}
 		}
 	}
 	return $script_handles;
@@ -95,25 +89,7 @@ function wwo_update_script_type( $tag, string $handle ) {
 			if ( $html_processor->get_attribute( 'id' ) !== "{$handle}-js" ) {
 				continue;
 			}
-			if ( null === $html_processor->get_attribute( 'async' ) && null === $html_processor->get_attribute( 'defer' ) ) {
-				_doing_it_wrong(
-					'wwo_update_script_type',
-					esc_html(
-						sprintf(
-							/* translators: %s: script handle */
-							__( 'Unable to offload "%s" script to a worker. Script will continue to load in the main thread.', 'web-worker-offloading' ),
-							$handle
-						)
-					),
-					'Web Worker Offloading 0.1.0'
-				);
-			} else {
-				$html_processor->set_attribute( 'type', 'text/partytown' );
-			}
-			if ( true === wp_scripts()->get_data( $handle, 'wwo_strategy_added' ) ) {
-				$html_processor->remove_attribute( 'async' );
-				$html_processor->remove_attribute( 'data-wp-strategy' );
-			}
+			$html_processor->set_attribute( 'type', 'text/partytown' );
 			$tag = $html_processor->get_updated_html();
 		}
 	}
@@ -121,3 +97,24 @@ function wwo_update_script_type( $tag, string $handle ) {
 	return $tag;
 }
 add_filter( 'script_loader_tag', 'wwo_update_script_type', 10, 2 );
+
+/**
+ * Filters inline script attributes to offload to a worker if the script has been opted-in.
+ *
+ * @param array<string, mixed>|mixed $attributes Attributes.
+ * @return array<string, mixed> Attributes.
+ */
+function wwo_filter_inline_script_attributes( $attributes ): array {
+	$attributes = (array) $attributes;
+	if (
+		isset( $attributes['id'] )
+		&&
+		1 === preg_match( '/^(?P<handle>.+)-js-(?:before|after)$/', $attributes['id'], $matches )
+		&&
+		(bool) wp_scripts()->get_data( $matches['handle'], 'worker' )
+	) {
+		$attributes['type'] = 'text/partytown';
+	}
+	return $attributes;
+}
+add_filter( 'wp_inline_script_attributes', 'wwo_filter_inline_script_attributes' );
--- a/plugins/web-worker-offloading/hooks.php
+++ b/plugins/web-worker-offloading/hooks.php
@@ -11,31 +11,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 /**
- * Gets configuration for Web Worker Offloading.
- *
- * @since 0.1.0
- * @link https://partytown.builder.io/configuration
- *
- * @return array{ debug?: bool, forward?: non-empty-string[], lib: non-empty-string, loadScriptsOnMainThread?: non-empty-string[], nonce?: non-empty-string } Configuration for Partytown.
- */
-function wwo_get_configuration(): array {
-	$config = array(
-		'lib'     => wp_parse_url( plugin_dir_url( __FILE__ ), PHP_URL_PATH ) . 'build/',
-		'forward' => array(),
-	);
-
-	/**
-	 * Add configuration for Web Worker Offloading.
-	 *
-	 * @since 0.1.0
-	 * @link https://partytown.builder.io/configuration
-	 *
-	 * @param array{ debug?: bool, forward?: non-empty-string[], lib: non-empty-string, loadScriptsOnMainThread?: non-empty-string[], nonce?: non-empty-string } $config Configuration for Partytown.
-	 */
-	return apply_filters( 'wwo_configuration', $config );
-}
-
-/**
  * Registers defaults scripts for Web Worker Offloading.
  *
  * @since 0.1.0
--- a/plugins/web-worker-offloading/hooks.php
+++ b/plugins/web-worker-offloading/hooks.php
@@ -14,6 +14,7 @@ if ( ! defined( 'ABSPATH' ) ) {
  * Registers defaults scripts for Web Worker Offloading.
  *
  * @since 0.1.0
+ * @access private
  *
  * @param WP_Scripts $scripts WP_Scripts instance.
  */
@@ -48,6 +49,7 @@ add_action( 'wp_default_scripts', 'wwo_register_default_scripts' );
  * Prepends web-worker-offloading to the list of scripts to print if one of the queued scripts is offloaded to a worker.
  *
  * @since 0.1.0
+ * @access private
  *
  * @param string[]|mixed $script_handles An array of enqueued script dependency handles.
  * @return string[] Script handles.
@@ -69,6 +71,7 @@ add_filter( 'print_scripts_array', 'wwo_filter_print_scripts_array', PHP_INT_MAX
  * Updates script type for handles having `web-worker-offloading` as dependency.
  *
  * @since 0.1.0
+ * @access private
  *
  * @param string|mixed $tag    Script tag.
  * @param string       $handle Script handle.
@@ -97,6 +100,7 @@ add_filter( 'script_loader_tag', 'wwo_update_script_type', 10, 2 );
  * Filters inline script attributes to offload to a worker if the script has been opted-in.
  *
  * @since 0.1.0
+ * @access private
  *
  * @param array<string, mixed>|mixed $attributes Attributes.
  * @return array<string, mixed> Attributes.
--- a/plugins/web-worker-offloading/hooks.php
+++ b/plugins/web-worker-offloading/hooks.php
@@ -96,6 +96,8 @@ add_filter( 'script_loader_tag', 'wwo_update_script_type', 10, 2 );
 /**
  * Filters inline script attributes to offload to a worker if the script has been opted-in.
  *
+ * @since 0.1.0
+ *
  * @param array<string, mixed>|mixed $attributes Attributes.
  * @return array<string, mixed> Attributes.
  */
--- a/plugins/web-worker-offloading/hooks.php
+++ b/plugins/web-worker-offloading/hooks.php
@@ -35,7 +35,7 @@ function wwo_register_default_scripts( WP_Scripts $scripts ): void {
 	$scripts->add_inline_script(
 		'web-worker-offloading',
 		sprintf(
-			'window.partytown = %s;',
+			'window.partytown = {...(window.partytown || {}), ...%s};',
 			wp_json_encode( wwo_get_configuration() )
 		),
 		'before'
--- a/plugins/web-worker-offloading/hooks.php
+++ b/plugins/web-worker-offloading/hooks.php
@@ -47,11 +47,6 @@ add_action( 'wp_default_scripts', 'wwo_register_default_scripts' );
 /**
  * Prepends web-worker-offloading to the list of scripts to print if one of the queued scripts is offloaded to a worker.
  *
- * This also marks their strategy as `async`. This is needed because scripts offloaded to a worker thread can be
- * considered async. However, they may include `before` and `after` inline scripts that need sequential execution. Once
- * marked as async, `filter_eligible_strategies()` determines if the script is eligible for async execution. If so, it
- * will be offloaded to the worker thread.
- *
  * @since 0.1.0
  *
  * @param string[]|mixed $script_handles An array of enqueued script dependency handles.
@@ -80,20 +75,19 @@ add_filter( 'print_scripts_array', 'wwo_filter_print_scripts_array' );
  */
 function wwo_update_script_type( $tag, string $handle ) {
 	if (
-		is_string( $tag ) &&
+		is_string( $tag )
+		&&
 		(bool) wp_scripts()->get_data( $handle, 'worker' )
 	) {
 		$html_processor = new WP_HTML_Tag_Processor( $tag );
-
 		while ( $html_processor->next_tag( array( 'tag_name' => 'SCRIPT' ) ) ) {
-			if ( $html_processor->get_attribute( 'id' ) !== "{$handle}-js" ) {
-				continue;
+			if ( $html_processor->get_attribute( 'id' ) === "{$handle}-js" ) {
+				$html_processor->set_attribute( 'type', 'text/partytown' );
+				$tag = $html_processor->get_updated_html();
+				break;
 			}
-			$html_processor->set_attribute( 'type', 'text/partytown' );
-			$tag = $html_processor->get_updated_html();
 		}
 	}
-
 	return $tag;
 }
 add_filter( 'script_loader_tag', 'wwo_update_script_type', 10, 2 );
--- a/plugins/web-worker-offloading/hooks.php
+++ b/plugins/web-worker-offloading/hooks.php
@@ -58,11 +58,12 @@ function wwo_filter_print_scripts_array( $script_handles ): array {
 		if ( true === (bool) $scripts->get_data( $handle, 'worker' ) ) {
 			$scripts->set_group( 'web-worker-offloading', false, 0 ); // Try to print in the head.
 			array_unshift( $script_handles, 'web-worker-offloading' );
+			break;
 		}
 	}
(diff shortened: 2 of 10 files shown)


westonruter:
@felixarntz @adamsilverstein This isn't ready for a full review, but I wanted to get a pre-review from you guys as the overall foundation is in place here.

<details><summary>Current diff with WooCommerce after formatting with Prettier</summary>

```diff
--- /tmp/before.html	2024-09-27 15:07:35.737764370 -0700
+++ /tmp/after.html	2024-09-27 15:09:10.461767260 -0700
@@ -2196,7 +2196,10 @@
         visibility: visible;
       }
     </style>
-    <script id="woocommerce-google-analytics-integration-gtag-js-after">
+    <script
+      id="woocommerce-google-analytics-integration-gtag-js-after"
+      type="text/partytown"
+    >
       /* Google Analytics for WooCommerce (gtag.js) */
       window.dataLayer = window.dataLayer || [];
       function gtag() {
@@ -3185,6 +3188,144 @@
         document.body.className = c;
       })();
     </script>
+    <script id="web-worker-offloading-js-before">
+      window.partytown = {
+        lib: "\/wp-content\/plugins\/web-worker-offloading\/build\/",
+        forward: ["dataLayer.push"],
+      };
+    </script>
+    <script id="web-worker-offloading-js-after">
+      const t = { preserveBehavior: !1 },
+        e = (e) => {
+          if ("string" == typeof e) return [e, t];
+          const [n, r = t] = e;
+          return [n, { ...t, ...r }];
+        },
+        n = Object.freeze(
+          ((t) => {
+            const e = new Set();
+            let n = [];
+            do {
+              Object.getOwnPropertyNames(n).forEach((t) => {
+                "function" == typeof n[t] && e.add(t);
+              });
+            } while ((n = Object.getPrototypeOf(n)) !== Object.prototype);
+            return Array.from(e);
+          })(),
+        );
+      !(function (t, r, o, i, a, s, c, d, l, p, u = t, f) {
+        function h() {
+          f ||
+            ((f = 1),
+            "/" ==
+              (c = (s.lib || "/~partytown/") + (s.debug ? "debug/" : ""))[0] &&
+              ((l = r.querySelectorAll('script[type="text/partytown"]')),
+              i != t
+                ? i.dispatchEvent(new CustomEvent("pt1", { detail: t }))
+                : ((d = setTimeout(v, 1e4)),
+                  r.addEventListener("pt0", w),
+                  a
+                    ? y(1)
+                    : o.serviceWorker
+                    ? o.serviceWorker
+                        .register(c + (s.swPath || "partytown-sw.js"), {
+                          scope: c,
+                        })
+                        .then(function (t) {
+                          t.active
+                            ? y()
+                            : t.installing &&
+                              t.installing.addEventListener(
+                                "statechange",
+                                function (t) {
+                                  "activated" == t.target.state && y();
+                                },
+                              );
+                        }, console.error)
+                    : v())));
+        }
+        function y(e) {
+          (p = r.createElement(e ? "script" : "iframe")),
+            (t._pttab = Date.now()),
+            e ||
+              ((p.style.display = "block"),
+              (p.style.width = "0"),
+              (p.style.height = "0"),
+              (p.style.border = "0"),
+              (p.style.visibility = "hidden"),
+              p.setAttribute("aria-hidden", !0)),
+            (p.src =
+              c +
+              "partytown-" +
+              (e ? "atomics.js?v=0.10.2" : "sandbox-sw.html?" + t._pttab)),
+            r.querySelector(s.sandboxParent || "body").appendChild(p);
+        }
+        function v(n, o) {
+          for (
+            w(),
+              i == t &&
+                (s.forward || []).map(function (n) {
+                  const [r] = e(n);
+                  delete t[r.split(".")[0]];
+                }),
+              n = 0;
+            n < l.length;
+            n++
+          )
+            ((o = r.createElement("script")).innerHTML = l[n].innerHTML),
+              (o.nonce = s.nonce),
+              r.head.appendChild(o);
+          p && p.parentNode.removeChild(p);
+        }
+        function w() {
+          clearTimeout(d);
+        }
+        (s = t.partytown || {}),
+          i == t &&
+            (s.forward || []).map(function (r) {
+              const [o, { preserveBehavior: i }] = e(r);
+              (u = t),
+                o.split(".").map(function (e, r, o) {
+                  var a;
+                  u = u[o[r]] =
+                    r + 1 < o.length
+                      ? u[o[r]] || ((a = o[r + 1]), n.includes(a) ? [] : {})
+                      : (() => {
+                          let e = null;
+                          if (i) {
+                            const { methodOrProperty: n, thisObject: r } = ((
+                              t,
+                              e,
+                            ) => {
+                              let n = t;
+                              for (let t = 0; t < e.length - 1; t += 1)
+                                n = n[e[t]];
+                              return {
+                                thisObject: n,
+                                methodOrProperty:
+                                  e.length > 0 ? n[e[e.length - 1]] : void 0,
+                              };
+                            })(t, o);
+                            "function" == typeof n &&
+                              (e = (...t) => n.apply(r, ...t));
+                          }
+                          return function () {
+                            let n;
+                            return (
+                              e && (n = e(arguments)),
+                              (t._ptf = t._ptf || []).push(o, arguments),
+                              n
+                            );
+                          };
+                        })();
+                });
+            }),
+          "complete" == r.readyState
+            ? h()
+            : (t.addEventListener("DOMContentLoaded", h),
+              t.addEventListener("load", h));
+      })(window, document, navigator, top, window.crossOriginIsolated);
+    </script>
     <script
       src="http://localhost:8888/wp-includes/js/dist/vendor/wp-polyfill.js?ver=3.15.0"
       id="wp-polyfill-js"
@@ -3260,7 +3401,7 @@
       id="wc-mini-cart-block-frontend-js"
     ></script>
     <script
-      async
+      type="text/partytown"
       src="https://www.googletagmanager.com/gtag/js?id=G-12345"
       id="google-tag-manager-js"
       data-wp-strategy="async"
```

</details> 

I'm not entirely sure yet why `async` is getting removed from the `google-tag-manager` script here, but this won't matter from a functional perspective.

westonruter:
> I'm not entirely sure yet why `async` is getting removed from the `google-tag-manager` script here, but this won't matter from a functional perspective.

Ah, the reason is that WooCommerce is [doing this](https://plugins.trac.wordpress.org/browser/woocommerce/tags/9.3.3/src/Blocks/Domain/Services/GoogleAnalytics.php#L120):

```php
	/**
	 * Add async to script tags with defined handles.
	 *
	 * @param string $tag HTML for the script tag.
	 * @param string $handle Handle of script.
	 * @param string $src Src of script.
	 * @return string
	 */
	public function async_script_loader_tags( $tag, $handle, $src ) {
		if ( ! in_array( $handle, array( 'google-tag-manager' ), true ) ) {
			return $tag;
		}
		// If script was output manually in wp_head, abort.
		if ( did_action( 'woocommerce_gtag_snippet' ) ) {
			return '';
		}
		return str_replace( '<script src', '<script async src', $tag );
	}
```

This is forcing the injection of the `async` attribute even though WooCommerce is _also_ [registering](https://plugins.trac.wordpress.org/browser/woocommerce-google-analytics-integration/tags/2.1.7/includes/class-wc-google-gtag-js.php#L73) the script with `'strategy' => 'async'`. Core removes `async` from this `SCRIPT` tag because the script has an [inline after script](https://plugins.trac.wordpress.org/browser/woocommerce-google-analytics-integration/tags/2.1.7/includes/class-wc-google-gtag-js.php#L87). So this is a case where an async script  with an after script should be forced a la Core-58632 which we closed as `maybelater` since it wasn't a common enough occurrence. (In reality, WooCommerce could change this to a inline _before_ script instead and the behavior would be unchanged while the `async` attribute would be preserved without having to filter `script_loader_tag`.) In any case, there is also a `wc-blocks-google-analytics` script which is a dependent of this script but it is [registered](https://plugins.trac.wordpress.org/browser/woocommerce-google-analytics-integration/tags/2.1.7/includes/class-wc-google-gtag-js.php#L121) with `$in_footer = true`, without the necessary `array( 'strategy' => 'async' )` to allow the `google-tag-manager` dependency to be served as `async`.

The reason why WooCommerce's `script_loader_tag` isn't able to inject `async` anymore into the script tag is because WWO is setting the `type` attribute to `text/partytown` and this is added to the beginning of the tag like `<script type="text/partytown"` so WooCommerce's replacement of `<script src` no longer matches.

In the end, this doesn't matter at all because a Partytown-run script is effectively also async.

plugins/web-worker-offloading/helper.php
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Gets configuration for Web Worker Offloading.
+ *
+ * @since 0.1.0
+ * @link https://partytown.builder.io/configuration
+ *
+ * @return array{ debug?: bool, forward?: non-empty-string[], lib: non-empty-string, loadScriptsOnMainThread?: non-empty-string[], nonce?: non-empty-string } Configuration for Partytown.
+ */
+function wwo_get_configuration(): array {
```
westonruter, line 22: I just moved this function from `hooks.php` to `helpers.php`

plugins/web-worker-offloading/hooks.php
```diff
@@ -88,17 +58,12 @@ function wwo_filter_print_scripts_array( $script_handles ): array {
 		if ( true === (bool) $scripts->get_data( $handle, 'worker' ) ) {
 			$scripts->set_group( 'web-worker-offloading', false, 0 ); // Try to print in the head.
 			array_unshift( $script_handles, 'web-worker-offloading' );
-
-			// TODO: This should be reconsidered because scripts needing to be offloaded will often have after scripts. See <https://github.com/WordPress/performance/pull/1497/files#r1733538721>.
-			if ( false === wp_scripts()->get_data( $handle, 'strategy' ) ) {
-				wp_script_add_data( $handle, 'strategy', 'async' ); // The 'defer' strategy would work as well.
-				wp_script_add_data( $handle, 'wwo_strategy_added', true );
-			}
```
westonruter, line 96: Inline after scripts are often used with analytics, and they need to be offloaded to a worker (maybe always?).

plugins/web-worker-offloading/hooks.php
```diff
@@ -88,17 +58,12 @@ function wwo_filter_print_scripts_array( $script_handles ): array {
...
+			break;
```
westonruter, line 63: We `break` because we only need to prepend the WWO script when encountering the first script that is to be offloaded to a worker.

plugins/web-worker-offloading/hooks.php
```diff
@@ -88,17 +58,12 @@ function wwo_filter_print_scripts_array( $script_handles ): array {
...
 		}
 	}
 	return $script_handles;
 }
-add_filter( 'print_scripts_array', 'wwo_filter_print_scripts_array' );
+add_filter( 'print_scripts_array', 'wwo_filter_print_scripts_array', PHP_INT_MAX );
```
westonruter, line 68: The priority is increased to `PHP_INT_MAX` so that functions like `wwo_mark_scripts_for_offloading()` can add `print_scripts_array` filters to mark scripts for worker offloading easily at earlier priorities.

plugins/web-worker-offloading/hooks.php
```diff
@@ -111,38 +76,40 @@ function wwo_filter_print_scripts_array( $script_handles ): array {
...
-							$handle
-						)
-					),
-					'Web Worker Offloading 0.1.0'
-				);
-			} else {
+			if ( $html_processor->get_attribute( 'id' ) === "{$handle}-js" ) {
 				$html_processor->set_attribute( 'type', 'text/partytown' );
+				$tag = $html_processor->get_updated_html();
+				break;
```
westonruter, line 91: Only one tag should have a given ID, so once we found it, we can bail.

plugins/web-worker-offloading/readme.txt
```diff
@@ -11,15 +11,23 @@ Offload JavaScript execution to a Web Worker.
...
 
 `
 wp_script_add_data( 'foo', 'worker', true );
 `
 
+Otherwise, the plugin currently ships with built-in integrations to offload Google Analytics to a web worker for the following plugins:
+
+* [Rank Math SEO](https://wordpress.org/plugins/seo-by-rank-math/)
+* [Site Kit by Google](https://wordpress.org/plugins/google-site-kit/)
+* [WooCommerce](https://wordpress.org/plugins/woocommerce/)
```
westonruter: Perhaps this PR should just land WooCommerce support, with Site Kit and Rank Math in subsequent PRs.

```suggestion
* [WooCommerce](https://wordpress.org/plugins/woocommerce/)
```

westonruter:
If this is approved, I'm thinking to submit it to the plugin directory for a Monday release.

plugins/web-worker-offloading/third-party.php
```diff
@@ -0,0 +1,65 @@
...
+		// TODO: google-site-kit.
+		// TODO: seo-by-rank-math.
+		'woocommerce',
+	);
+
+	// Load corresponding file for each string in $plugins if the WordPress plugin is installed and active.
+	$active_plugin_slugs = array_filter(
+		array_map(
+			static function ( $plugin_file ) {
+				if ( is_string( $plugin_file ) && str_contains( $plugin_file, '/' ) ) {
```
swissspidy: `str_starts_with` would be more robust. Otherwise this results in false positives for things like `not-woocommerce`.

Even more robust would be to check for a specific constant or function defined by the plugin, as the plugin might not necessarily be in that exact folder.

For example on my local dev site, Site Kit is installed in `plugins/site-kit-wp`, not `plugins/google-site-kit`

westonruter: > `str_starts_with` would be more robust. Otherwise this results in false positives for things like `not-woocommerce`.

I don't think this is right because the `str_contains()` is just checking whether it is a directory-based plugin and not a file-based plugin. The exact match of `woocommerce` is handled below with the `array_intersect()`.

> Even more robust would be to check for a specific constant or function defined by the plugin, as the plugin might not necessarily be in that exact folder.

That's true. I was wanting to avoid manually keeping track of those things, but I suppose it's not so bad.

westonruter: Done in 74b93c79

swissspidy (approved):

plugins/web-worker-offloading/helper.php
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @since 0.1.0
+ * @link https://partytown.builder.io/configuration
+ * @link https://github.com/BuilderIO/partytown/blob/b292a14047a0c12ca05ba97df1833935d42fdb66/src/lib/types.ts#L393-L548
+ *
+ * @return array<string, mixed> Configuration for Partytown.
+ */
+function wwo_get_configuration(): array {
+	$config = array(
+		'lib' => wp_parse_url( plugin_dir_url( __FILE__ ), PHP_URL_PATH ) . 'build/',
```
felixarntz, line 24: There used to be `forward` in here, can you clarify why you removed it?

plugins/web-worker-offloading/third-party/woocommerce.php
```diff
@@ -0,0 +1,40 @@
...
+ * @param array<string, mixed>|mixed $configuration Configuration.
+ * @return array<string, mixed> Configuration.
+ */
+function wwo_woocommerce_configure( $configuration ): array {
+	$configuration = (array) $configuration;
+
+	$configuration['mainWindowAccessors'][] = 'wp';   // Because woocommerce-google-analytics-integration needs to access wp.i18n.
+	$configuration['mainWindowAccessors'][] = 'ga4w'; // Because woocommerce-google-analytics-integration needs to access window.ga4w.
+	$configuration['globalFns'][]           = 'gtag'; // Because gtag() is defined in one script and called in another.
+	$configuration['forward'][]             = 'dataLayer.push'; // Because the Partytown integration has this in its example config.
```
felixarntz, line 29: Don't these cause PHP notices because the keys they add to are not initialized as arrays?

plugins/web-worker-offloading/helper.php
```diff
@@ -0,0 +1,70 @@
...
+	 *
+	 * There are also many configuration options which are not documented, so refer to the TypeScript definitions.
+	 *
+	 * @since 0.1.0
+	 * @link https://partytown.builder.io/configuration
+	 * @link https://github.com/BuilderIO/partytown/blob/b292a14047a0c12ca05ba97df1833935d42fdb66/src/lib/types.ts#L393-L548
+	 *
+	 * @param array<string, mixed> $config Configuration for Partytown.
+	 */
+	return (array) apply_filters( 'wwo_configuration', $config );
```
felixarntz, line 69: I think it would be good to add some sanitization here.

For example, since several supported properties/keys are arrays and several plugins/sources may add similar entries to them, we could sanitize these arrays using `array_unique()`. Just thinking about what this PR implements for WooCommerce for example allowlisting `wp`, I'm sure lots of scripts would need that.

plugins/web-worker-offloading/helper.php
```diff
@@ -0,0 +1,70 @@
...
+ * @return array<string, mixed> Configuration for Partytown.
+ */
+function wwo_get_configuration(): array {
+	$config = array(
+		'lib' => wp_parse_url( plugin_dir_url( __FILE__ ), PHP_URL_PATH ) . 'build/',
+	);
+
+	if ( WP_DEBUG && SCRIPT_DEBUG ) {
+		$config['debug'] = true;
+	}
```
felixarntz, line 29: How about always having this key set, and making the conditional only around whether it's `true` or `false`?

felixarntz (commented):
@westonruter This mostly looks good, just a few questions / suggestions.

plugins/web-worker-offloading/helper.php
westonruter: Well, because it defaults to `false` anyway, and by not setting it we avoid sending (a tiny amount) of bytes down the wire.

felixarntz: That's probably negligible though for performance. I'd say initializing properties here makes it easier to understand what's available, also since this PHP layer is entirely separate from the actual JS API where this is consumed.

westonruter: I'm not sure. If someone wants to know what is available they should look at the docs which are linked in the source code:

https://github.com/WordPress/performance/blob/db9d10e1cacd8ffea0cc8c7dea4bdc3d00e42111/plugins/web-worker-offloading/helper.php#L17-L18

westonruter: We don't want to have to supply the default values for very single configuration option (many of which are under-documented in Partytown).

felixarntz: That's fair. Let's leave this as is then.

plugins/web-worker-offloading/third-party/woocommerce.php
westonruter: No it doesn't: https://3v4l.org/YPITA

westonruter: In fact, this also doesn't cause any notice:

```php
$foo['bar']['baz'][] = 'quux';
```

It results in `$foo` being defined as:

```
array(1) {
  ["bar"]=>
  array(1) {
    ["baz"]=>
    array(1) {
      [0]=>
      string(4) "quux"
    }
  }
}
```

https://3v4l.org/rMVi7

plugins/web-worker-offloading/helper.php
westonruter: Yes, because it was empty so it wasn't having any purpose.

plugins/web-worker-offloading/helper.php
westonruter: I would agree with you, but I think sanitizing for this initial version could go down a rabbit hole. 

The number of possible configuration options is large and the value types can be complex: https://github.com/BuilderIO/partytown/blob/main/src/lib/types.ts

We could implement a JSON Schema to mirror the TypeScript types and use this to provide usage warnings. However, I think this validation would be better done inside Partytown itself where the types are being defined.

In the case of `mainWindowAccessors`, there's no need for `array_unique()` because Partytown is using the value like [this](https://github.com/BuilderIO/partytown/blob/b292a14047a0c12ca05ba97df1833935d42fdb66/src/lib/web-worker/worker-window.ts#L370-L371):

```js
              } else if (webWorkerCtx.$config$.mainWindowAccessors?.includes(propName)) {
                return getter(this, [propName]);
```

So I think we should consider what validation and sanitization we do as part of this plugin for a subsequent release.

plugins/web-worker-offloading/third-party/woocommerce.php
felixarntz: Hmm, I learned something today. :)

plugins/web-worker-offloading/helper.php
felixarntz: I think it depends also on how _we_ want to use their configuration options. We don't need to holistically capture everything Partytown supports, that would indeed be a rabbit hole. But for instance, it seems that `mainWindowAccessors` will be relevant in how we would use Partytown for quite a few integrations, so we could do some sanitization there where it makes sense. In other words, I think we should do this based on demand.

As long as all of these 4 arguments only support an array of strings, and as long as duplicate values are not a problem, it's probably okay not to sanitize anything. So we can move ahead with this as is, but I think we should look out for when things are more complicated to use, or could be of more than one type, and especially then we should do some sanitization for how _we_ want those arguments to be used.

felixarntz (approved):

PR #1569, opened 2024-09-30
Web Worker Offloading: Address feedback from Plugin Review team

westonruter:
Addresses feedback:

* No publicly documented resource for your compressed content (the Partytown scripts in the build directory)
* Generic function/class/define/namespace/option names (that is, not unique enough)


swissspidy (approved):

felixarntz (approved):

westonruter:
I'll leave this open in case there is additional feedback to address from the plugin review team.

PR #1573, opened 2024-10-01
Only store info for relevant standalone plugins in the transient cache

mukeshpanchal27:
Follow-up on #1562

The current API query `https://api.wordpress.org/plugins/info/1.2/?action=query_plugins&request[author]=wordpressdotorg&request[tag]=performance&request[per_page]=100` is returning some unrelated plugins like `Plugin Check (PCP)` and `SQLite Database Integration`. In the future, if anyone else uses `Performance` as a tag, it will also be added to our `transient`. However, we only need the standalone plugins defined in `perflab_get_standalone_plugin_data()`.

I’ve also moved the array definition for `$fields` after the cache check, as it's redundant in the cached version.

### DIFF
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -21,7 +21,16 @@ if ( ! defined( 'ABSPATH' ) ) {
 function perflab_query_plugin_info( string $plugin_slug ) {
 	$transient_key = 'perflab_plugins_info';
 	$plugins       = get_transient( $transient_key );
-	$fields        = array(
+
+	if ( is_array( $plugins ) ) {
+		// If the specific plugin_slug is not in the cache, return an error.
+		if ( ! isset( $plugins[ $plugin_slug ] ) ) {
+			return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'performance-lab' ) );
+		}
+		return $plugins[ $plugin_slug ]; // Return cached plugin info if found.
+	}
+
+	$fields = array(
 		'name',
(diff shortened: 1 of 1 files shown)


swissspidy (approved):

adamsilverstein (approved):
Nice!

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -69,6 +70,9 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 
 	$plugins = array();
 	foreach ( $response->plugins as $plugin_data ) {
+		if ( ! in_array( $plugin_data['slug'], perflab_get_standalone_plugins(), true ) ) {
```
felixarntz: More efficient would be to move the call outside the loop and store the plugins in a variable.

If you do that, you could even add another little performance enhancement: You could use `array_flip()` on the plugin list, and then use `isset()` instead of `in_array()` (as `isset()` is faster).

felixarntz (approved):
@mukeshpanchal27 Great catch!

PR looks good, though one recommendation to improve performance.

plugins/performance-lab/includes/admin/plugins.php
mukeshpanchal27: Done in https://github.com/WordPress/performance/pull/1573/commits/dc9f0e349a430b029141895784ad9e16222b9b2a

westonruter:
FYI: We forgot to account for plugin dependencies here. Follow-up PR: https://github.com/WordPress/performance/pull/1613

PR #1574, opened 2024-10-02
Fix storage of URL Metric when plain non-pretty permalinks are enabled

### LINKED ISSUE #1567: REST API calls to store URL metrics fail when pretty permalinks aren't enabled
When pretty permalinks aren't enabled, Optimization Detective sends the URL metrics data to the `/index.php?rest_route=%2Foptimization-detective%2Fv1%2Furl-metrics%3Astore&_wpnonce=d6ef56f704`. This ends up causing the request to fail with:

```json
{
    "code": "rest_invalid_param",
    "message": "Failed to validate URL metric: rest_route is not a valid property of Object.",
    "data": {
        "status": 400
    }
}
```

This is because `rest_route` is returned, unexpectedly, as one of the params when calling `$request->get_params()`:

https://github.com/WordPress/performance/blob/ddf297dca97641cfb21bef75fc4fd45ff49ea95b/plugins/optimization-detective/storage/rest-api.php#L131-L136

The quick fix would be to `unset( $data['rest_route'] )`, but it would be better if the params we obtain are exclusively from the JSON body. 

This code is also being touched in https://github.com/WordPress/performance/pull/1373 where the issue is also occurring.

Granted, likely only a miniscule number of sites don't have pretty permalinks enabled, so this isn't a critical issue.

westonruter:
Fixes #1567

The REST API merges params from GET, POST, and JSON into one params array. This causes problems when pretty permalinks, as the fetched URL is:

```
http://localhost:8888/index.php?rest_route=/optimization-detective/v1/url-metrics:store
```

Where the `rest_route` is erroneously picked up as one of the parameters.

To fix this, this PR separates meta parameters from the core URL metric parameters. The meta parameters (`nonce` and `slug`) are added as GET parameters whereas the parameters that comprise the URL metric data are all exclusively in the content body as JSON. This feels cleaner from a REST API perspective since the entity then becomes truly URL metric data, rather than URL metric with additional meta parameters added to it.

This also tidies up some symbol names, with "URL Metrics" (plural) being changed to "URL Metric" (singular) where appropriate.

Aside: Perhaps `OD_URL_Metrics_Group_Collection` should be renamed to `OD_URL_Metric_Group_Collection` as well. The "metrics" plural here is redundant with the group in the name.

### DIFF
--- a/plugins/optimization-detective/class-od-tag-visitor-context.php
+++ b/plugins/optimization-detective/class-od-tag-visitor-context.php
@@ -17,7 +17,7 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @since 0.4.0
  * @access private
  *
- * @property-read OD_URL_Metric_Group_Collection $url_metrics_group_collection
+ * @property-read OD_URL_Metric_Group_Collection $url_metrics_group_collection Deprecated property accessed via magic getter. Use the url_metric_group_collection property instead.
  */
 final class OD_Tag_Visitor_Context {
 
@@ -59,7 +59,10 @@ final class OD_Tag_Visitor_Context {
 	}
 
 	/**
-	 * Gets magic properties.
+	 * Gets deprecated property.
+	 *
+	 * @since n.e.x.t
+	 * @todo Remove this when no plugins are possibly referring to the url_metrics_group_collection property anymore.
 	 *
 	 * @param string $name Property name.
 	 * @return OD_URL_Metric_Group_Collection URL metric group collection.
--- a/plugins/optimization-detective/class-od-tag-visitor-context.php
+++ b/plugins/optimization-detective/class-od-tag-visitor-context.php
@@ -16,6 +16,8 @@ if ( ! defined( 'ABSPATH' ) ) {
  *
  * @since 0.4.0
  * @access private
+ *
+ * @property-read OD_URL_Metric_Group_Collection $url_metrics_group_collection
  */
 final class OD_Tag_Visitor_Context {
 
@@ -28,12 +30,12 @@ final class OD_Tag_Visitor_Context {
 	public $processor;
 
 	/**
-	 * URL metrics group collection.
+	 * URL metric group collection.
 	 *
-	 * @var OD_URL_Metrics_Group_Collection
+	 * @var OD_URL_Metric_Group_Collection
 	 * @readonly
 	 */
-	public $url_metrics_group_collection;
+	public $url_metric_group_collection;
 
 	/**
 	 * Link collection.
@@ -46,13 +48,47 @@ final class OD_Tag_Visitor_Context {
 	/**
 	 * Constructor.
 	 *
-	 * @param OD_HTML_Tag_Processor           $processor                    HTML tag processor.
-	 * @param OD_URL_Metrics_Group_Collection $url_metrics_group_collection URL metrics group collection.
-	 * @param OD_Link_Collection              $link_collection              Link collection.
+	 * @param OD_HTML_Tag_Processor          $processor                   HTML tag processor.
+	 * @param OD_URL_Metric_Group_Collection $url_metric_group_collection URL metric group collection.
+	 * @param OD_Link_Collection             $link_collection             Link collection.
 	 */
-	public function __construct( OD_HTML_Tag_Processor $processor, OD_URL_Metrics_Group_Collection $url_metrics_group_collection, OD_Link_Collection $link_collection ) {
-		$this->processor                    = $processor;
-		$this->url_metrics_group_collection = $url_metrics_group_collection;
-		$this->link_collection              = $link_collection;
+	public function __construct( OD_HTML_Tag_Processor $processor, OD_URL_Metric_Group_Collection $url_metric_group_collection, OD_Link_Collection $link_collection ) {
+		$this->processor                   = $processor;
+		$this->url_metric_group_collection = $url_metric_group_collection;
+		$this->link_collection             = $link_collection;
+	}
+
+	/**
+	 * Gets magic properties.
+	 *
+	 * @param string $name Property name.
+	 * @return OD_URL_Metric_Group_Collection URL metric group collection.
+	 *
+	 * @throws Error When property is unknown.
+	 */
+	public function __get( string $name ): OD_URL_Metric_Group_Collection {
+		if ( 'url_metrics_group_collection' === $name ) {
+			_doing_it_wrong(
+				__CLASS__ . '::$url_metrics_group_collection',
+				esc_html(
+					sprintf(
+						/* translators: %s is class member variable name */
+						__( 'Use %s instead.', 'optimization-detective' ),
+						__CLASS__ . '::$url_metric_group_collection'
+					)
+				),
+				'optimization-detective n.e.x.t'
+			);
+			return $this->url_metric_group_collection;
+		}
+		throw new Error(
+			esc_html(
+				sprintf(
+					/* translators: %s is class member variable name */
+					__( 'Unknown property %s.', 'optimization-detective' ),
+					__CLASS__ . '::$' . $name
+				)
+			)
+		);
 	}
 }
--- a/plugins/optimization-detective/readme.txt
+++ b/plugins/optimization-detective/readme.txt
@@ -205,7 +205,7 @@ The [plugin source code](https://github.com/WordPress/performance/tree/trunk/plu
 
 **Enhancements**
 
-* Log URL metrics group collection to console when debugging is enabled (`WP_DEBUG` is true). ([1295](https://github.com/WordPress/performance/pull/1295))
+* Log URL metric group collection to console when debugging is enabled (`WP_DEBUG` is true). ([1295](https://github.com/WordPress/performance/pull/1295))
 
 **Bug Fixes**
 
--- a/plugins/optimization-detective/readme.txt
+++ b/plugins/optimization-detective/readme.txt
@@ -205,7 +205,7 @@ The [plugin source code](https://github.com/WordPress/performance/tree/trunk/plu
 
 **Enhancements**
 
-* Log URL metric group collection to console when debugging is enabled (`WP_DEBUG` is true). ([1295](https://github.com/WordPress/performance/pull/1295))
+* Log URL metrics group collection to console when debugging is enabled (`WP_DEBUG` is true). ([1295](https://github.com/WordPress/performance/pull/1295))
 
 **Bug Fixes**
 
--- a/plugins/optimization-detective/storage/class-od-url-metric-stored-context.php
+++ b/plugins/optimization-detective/storage/class-od-url-metric-stored-context.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * Optimization Detective: OD_URL_Metric_Stored_Context class
+ *
+ * @package optimization-detective
+ * @since n.e.x.t
+ */
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Context for when a URL metric is successfully stored via the REST API.
+ *
+ * @since n.e.x.t
+ * @access private
+ */
+final class OD_URL_Metric_Stored_Context {
+
+	/**
+	 * Request.
+	 *
+	 * @var WP_REST_Request<array<string, mixed>>
+	 * @readonly
+	 */
+	public $request;
+
+	/**
+	 * ID for the URL metric post.
+	 *
+	 * @var int
+	 * @readonly
+	 */
+	public $post_id;
+
+	/**
+	 * URL metric group collection.
+	 *
+	 * @var OD_URL_Metric_Group_Collection
+	 * @readonly
+	 */
+	public $url_metric_group_collection;
+
+	/**
+	 * URL metric group.
+	 *
+	 * @var OD_URL_Metric_Group
+	 * @readonly
+	 */
+	public $url_metric_group;
+
+	/**
+	 * URL metric.
+	 *
+	 * @var OD_URL_Metric
+	 * @readonly
+	 */
+	public $url_metric;
+
+	/**
+	 * Constructor.
+	 *
+	 * @phpstan-param WP_REST_Request<array<string, mixed>> $request
+	 *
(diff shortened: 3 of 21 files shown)


felixarntz (commented):
@westonruter This makes sense to me and looks good, however I'm curious: Does this actually fix a bug, or is the change only about improving the code?

You were previously removing `slug` and `nonce` from the REST `get_params()` anyway, which seems it would be the same as now using `get_json_params()`. You mention the `rest_route` query parameter, but I thought this would not get included anyway since it's not a supported parameter of the endpoint?

> Aside: Perhaps `OD_URL_Metrics_Group_Collection` should be renamed to `OD_URL_Metric_Group_Collection` as well. The "metrics" plural here is redundant with the group in the name.

+1 for that. Since it seems nobody other than us is adopting the plugin yet, this is probably fine to just rename. Once we document this further and encourage external usage, we'll need to be more careful about this kind of change.

westonruter:
> Does this actually fix a bug, or is the change only about improving the code?

Yes, it fixes the bug, as well as improving the code. In reality, the most minimal patch to only fix the code would be to unset `rest_route` from the `$params` as well.  The bug is that WordPress erroneously includes `rest_route` in the return value of `get_params()` even though it is not registered as one of the endpoint args.

So for non-pretty permalinks, using `get_json_params()` would be functionally the same as `get_params()` with the addition of the existing `unset()` of `$data['nonce']` and `$data['slug']`. 

Getting the URL Metric data from the JSON exclusively is safer as it's less likely that arbitrary query parameters will be added to the REST API URL to cause a data validation problem in the future. For example, if I had modified the plugin to add some cachebuster query parameter:

```diff
--- a/plugins/optimization-detective/detect.js
+++ b/plugins/optimization-detective/detect.js
@@ -392,7 +392,9 @@ export default async function detect( {
 	} );
 
 	try {
-		const response = await fetch( restApiEndpoint, {
+		const restUrl = new URL( restApiEndpoint );
+		restUrl.searchParams.append( 'cachebust', '123' );
+		const response = await fetch( restUrl, {
 			method: 'POST',
 			headers: {
 				'Content-Type': 'application/json',
```

(Such a parameter could be added by a proxy layer even.)

Then I get the same error even when pretty permalinks are enabled:

> Failed to validate URL metric: cachebust is not a valid property of Object.

So again, it seems like a core bug that `get_params()` returns all of the GET and POST params, even those which are not registered.

westonruter:
> +1 for that. Since it seems nobody other than us is adopting the plugin yet, this is probably fine to just rename. Once we document this further and encourage external usage, we'll need to be more careful about this kind of change.

@felixarntz Done in 942b397, although it touched quite a bit of code.

plugins/optimization-detective/class-od-tag-visitor-context.php
```diff
@@ -46,13 +48,47 @@ final class OD_Tag_Visitor_Context {
...
+	 * @param string $name Property name.
+	 * @return OD_URL_Metric_Group_Collection URL metric group collection.
+	 *
+	 * @throws Error When property is unknown.
 	 */
-	public function __construct( OD_HTML_Tag_Processor $processor, OD_URL_Metrics_Group_Collection $url_metrics_group_collection, OD_Link_Collection $link_collection ) {
-		$this->processor                    = $processor;
-		$this->url_metrics_group_collection = $url_metrics_group_collection;
-		$this->link_collection              = $link_collection;
+	public function __get( string $name ): OD_URL_Metric_Group_Collection {
```
westonruter, line 72: This is needed to account for someone updating Optimization Detective but not right away updating Image Prioritizer or Embed Optimizer.  They are using the `url_metrics_group_collection` symbol, so without this back-compat code, a fatal error would occur. This is similar to following code in Image Prioritizer:

https://github.com/WordPress/performance/blob/5952d556d9632f93c2fbbda76c3516576ebdb1a5/plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php#L69-L88

The `is_any_group_populated()` method was added to Optimization Detective in a recent release, and to avoid a fatal error if someone updates Optimization Detective without updating Image Prioritizer as well, this prevents the code from causing a fatal error.

All of this code could be removed later once there is no longer this risk.

plugins/optimization-detective/storage/rest-api.php
```diff
@@ -173,24 +176,24 @@ function od_handle_rest_request( WP_REST_Request $request ) {
...
-			'request'                      => $request,
-			'url_metric'                   => $url_metric,
-			'url_metrics_group'            => $group,
-			'url_metrics_group_collection' => $group_collection,
+			'post_id'                     => $post_id,
+			'request'                     => $request,
+			'url_metric'                  => $url_metric,
+			'url_metric_group'            => $url_metric_group,
+			'url_metric_group_collection' => $url_metric_group_collection,
 		)
```
westonruter: Instead of using an array here, it would be better to implement something like `OD_Tag_Visitor_Context` which has properties instead of array keys. This would allow for autocompletion as well as it would support possible future renaming of the members.

westonruter: Like this: 1cf24f17f5f06706d3cc4db7586de1a4adb3b071

plugins/optimization-detective/storage/class-od-url-metric-stored-context.php
```diff
@@ -0,0 +1,80 @@
...
+	 *
+	 * @phpstan-param WP_REST_Request<array<string, mixed>> $request
+	 *
+	 * @param WP_REST_Request                $request                     REST API request.
+	 * @param int                            $post_id                     ID for the URL metric post.
+	 * @param OD_URL_Metric_Group_Collection $url_metric_group_collection URL metric group collection.
+	 * @param OD_URL_Metric_Group            $url_metric_group            URL metric group.
+	 * @param OD_URL_Metric                  $url_metric                  URL metric.
+	 */
+	public function __construct( WP_REST_Request $request, int $post_id, OD_URL_Metric_Group_Collection $url_metric_group_collection, OD_URL_Metric_Group $url_metric_group, OD_URL_Metric $url_metric ) {
```
westonruter: This class has a terrible amount of redundancy. However, in newer versions of PHP this can be made much simpler, thanks to [Constructor Promotion](https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion).

So this could be replaced with:

```php
final class OD_URL_Metric_Stored_Context {
	public function __construct(
		public WP_REST_Request $request,
		public int $post_id,
		public OD_URL_Metric_Group_Collection $url_metric_group_collection,
		public OD_URL_Metric_Group $url_metric_group,
		public OD_URL_Metric $url_metric
	) {}
}
```

This works in PHP 8.0+: https://3v4l.org/OqvTC

So we can't take advantage of it. But it will greatly simplify this class and `OD_Tag_Visitor_Context`.

plugins/optimization-detective/class-od-tag-visitor-context.php
```diff
@@ -16,6 +16,8 @@
  *
  * @since 0.4.0
  * @access private
+ *
+ * @property-read OD_URL_Metric_Group_Collection $url_metrics_group_collection
```
felixarntz: I don't think it's worth adding this, given it's only for backward compatibility.

plugins/optimization-detective/class-od-tag-visitor-context.php
felixarntz: Let's add a `TODO` to remove this later.

plugins/optimization-detective/readme.txt
```diff
@@ -205,7 +205,7 @@ The [plugin source code](https://github.com/WordPress/performance/tree/trunk/plu
 
 **Enhancements**
 
-* Log URL metrics group collection to console when debugging is enabled (`WP_DEBUG` is true). ([1295](https://github.com/WordPress/performance/pull/1295))
+* Log URL metric group collection to console when debugging is enabled (`WP_DEBUG` is true). ([1295](https://github.com/WordPress/performance/pull/1295))
```
felixarntz: Nit-pick, but I don't think we should update older changelogs as they are "history artifacts" (e.g. we probably wouldn't update this in an old GitHub release either). Let's revert this.

felixarntz:
> > Does this actually fix a bug, or is the change only about improving the code?
> 
> Yes, it fixes the bug, as well as improving the code. In reality, the most minimal patch to only fix the code would be to unset `rest_route` from the `$params` as well. The bug is that WordPress erroneously includes `rest_route` in the return value of `get_params()` even though it is not registered as one of the endpoint args.
> 
> [...]
> 
> So again, it seems like a core bug that `get_params()` returns all of the GET and POST params, even those which are not registered.

Is there a Trac ticket for this? If not, could you please open one? :)

plugins/optimization-detective/storage/class-od-url-metric-stored-context.php
felixarntz: How do you mean redundancy? Just because the properties are declared in several places?

plugins/optimization-detective/storage/rest-api.php
```diff
@@ -169,28 +172,20 @@ function od_handle_rest_request( WP_REST_Request $request ) {
...
-			'url_metrics_group'            => $group,
-			'url_metrics_group_collection' => $group_collection,
+		'od_url_metric_stored',
+		new OD_URL_Metric_Stored_Context(
+			$request,
+			$post_id,
+			$url_metric_group_collection,
+			$url_metric_group,
+			$url_metric
 		)
```
felixarntz: Should the old filter still be applied as deprecated rather than removed?

More importantly though, why wrap the data in a class? Not saying we shouldn't do it, but I'm not sure what's the value, even more so why this needs to happen _now_ (as part of this PR).

plugins/optimization-detective/class-od-tag-visitor-context.php
westonruter: Added in 5960ff3c

plugins/optimization-detective/class-od-tag-visitor-context.php
westonruter: It's needed for static analysis. In 5960ff3c I've added a note that it is deprecated.

plugins/optimization-detective/readme.txt
westonruter: Reverted in ef9f3ae7

plugins/optimization-detective/storage/class-od-url-metric-stored-context.php
westonruter: Yes. There are 5 instances of the `url_metric_group_collection` string in this class file when there could just be 1.

plugins/optimization-detective/storage/rest-api.php
westonruter: The old action was just introduced in the most recent version and no plugins are using it yet. So I don't think there is any need for deprecation here. This is in contrast with the other deprecation logic in this PR, as Image Prioritizer and Embed Optimizer are using the changed API surface.

Wrapping the data in a class gives more flexibility to change the property names later if need be, using a similar deprecation strategy with the magic getter in this PR. It also allows for autocompletion for the properties. 

I figured to include it here because I was already changing the `url_metrics_group` and `url_metrics_group_collection` array keys, so it seemed like a logical opportunity to change the names but also switch from an array to an object.

westonruter:
> Is there a Trac ticket for this? If not, could you please open one? :)

I've just opened Core-62163

plugins/optimization-detective/storage/class-od-url-metric-stored-context.php
felixarntz: But that's just about how PHP (and many other languages work) right? I wouldn't necessarily call that redundancy, it's just more code to write. :)

plugins/optimization-detective/storage/rest-api.php
```diff
@@ -169,28 +172,20 @@ function od_handle_rest_request( WP_REST_Request $request ) {
...
-	 * @param array $context {
-	 *     Context about the successful URL Metric collection.
-	 *
-	 *     @var int                                   $post_id
-	 *     @var WP_REST_Request<array<string, mixed>> $request
-	 *     @var OD_Strict_URL_Metric                  $url_metric
-	 *     @var OD_URL_Metrics_Group                  $group
-	 *     @var OD_URL_Metrics_Group_Collection       $group_collection
-	 * }
+	 * @param OD_URL_Metric_Stored_Context $context Context.
```
felixarntz: Maybe keep the more precise description from before?
```suggestion
	 * @param OD_URL_Metric_Stored_Context $context Context about the successful URL Metric collection.
```

plugins/optimization-detective/storage/rest-api.php
```diff
@@ -169,28 +172,20 @@ function od_handle_rest_request( WP_REST_Request $request ) {
...
 	do_action(
-		'od_url_metric_collected',
-		array(
-			'post_id'                      => $post_id,
-			'request'                      => $request,
-			'url_metric'                   => $url_metric,
-			'url_metrics_group'            => $group,
-			'url_metrics_group_collection' => $group_collection,
+		'od_url_metric_stored',
+		new OD_URL_Metric_Stored_Context(
```
felixarntz: Since the only reason for this class to exist is the filter, I think we should clarify that in the name, which could be confusing if you look at the class alone.

Either we go with `OD_URL_Metric_Context`, which sounds like a good name, but could have the risk of sounding more generic than it actually is. Or we go with `OD_URL_Metric_Stored_Filter_Context`, which keeps the scope narrow as it is now, but clarifies that it's for a filter. I think the latter was unclear to me when I first saw the class, and I was wondering what "stored context" is.

plugins/optimization-detective/storage/class-od-url-metric-stored-context.php
westonruter: Correct. The verbosity can be reduced in the future.

plugins/optimization-detective/storage/rest-api.php
westonruter: How about eaed94f82f0933ac974f50acbeaa93b92abe9eab?

felixarntz: WFM 👍

felixarntz (approved):

PR #1577, opened 2024-10-03
Add Web Worker Offloading to list of Performance features

westonruter:
The Web Worker Offloading plugin is now live! https://wordpress.org/plugins/web-worker-offloading/

This adds the plugin to the list of experimental performance features to activate:

image

Note the modification of the `per_page` field is simply to bust the WordPress.org API response cache, as otherwise the plugin is not currently included in the results and the screen is:

Screenshot 2024-10-03 12 02 06

I'm not sure how long the responses are cached. Perhaps in another hour it will be flushed.

### DIFF
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -47,7 +47,7 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 		array(
 			'author'   => 'wordpressdotorg',
 			'tag'      => 'performance',
-			'per_page' => 99, // Decremented from 100 to bust WordPress.org response cache.
+			'per_page' => 100,
 			'fields'   => array_fill_keys( $fields, true ),
 		)
 	);
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -47,7 +47,7 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 		array(
 			'author'   => 'wordpressdotorg',
 			'tag'      => 'performance',
-			'per_page' => 100,
+			'per_page' => 99, // Decremented from 100 to bust WordPress.org response cache.
 			'fields'   => array_fill_keys( $fields, true ),
 		)
 	);
--- a/plugins/performance-lab/load.php
+++ b/plugins/performance-lab/load.php
@@ -114,6 +114,10 @@ function perflab_get_standalone_plugin_data(): array {
 		'speculation-rules'       => array(
 			'constant' => 'SPECULATION_RULES_VERSION',
 		),
+		'web-worker-offloading'   => array(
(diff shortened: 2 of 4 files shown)


plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -47,7 +47,7 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 		array(
 			'author'   => 'wordpressdotorg',
 			'tag'      => 'performance',
-			'per_page' => 100,
+			'per_page' => 99, // Decremented from 100 to bust WordPress.org response cache.
```
felixarntz: This is a bit of a strange workaround. 🙃 

I think we can omit this, the cache should refresh soon enough.

felixarntz (commented):
@westonruter Mostly looks good, except the hack, see below.

Also worth noting that the description of the plugin doesn't look right. At a minimum, it should also start with a third-person singular verb like all the other descriptions. Additionally, we should consider making the description more meaningful to an end user - for instance it could lead to the question "why would it help to offload scripts to a webworker?"

plugins/performance-lab/includes/admin/plugins.php
westonruter: Should hopefully. Still cached as of now. We can leave this open until the cache expires I guess.

westonruter:
> Also worth noting that the description of the plugin doesn't look right. At a minimum, it should also start with a third-person singular verb like all the other descriptions. Additionally, we should consider making the description more meaningful to an end user - for instance it could lead to the question "why would it help to offload scripts to a webworker?"

How about ~28b0628f~?

**Update:** 4fc8751

plugins/performance-lab/includes/admin/plugins.php
felixarntz: I'm sure this won't be relevant anymore by the time we release this 😆

felixarntz:
> > Also worth noting that the description of the plugin doesn't look right. At a minimum, it should also start with a third-person singular verb like all the other descriptions. Additionally, we should consider making the description more meaningful to an end user - for instance it could lead to the question "why would it help to offload scripts to a webworker?"
> 
> How about [28b0628](https://github.com/WordPress/performance/commit/28b0628fb67c942d460843baca609b0b542f197f)?

That description for WWO looks good to me! Note that it looks like you accidentally updated the Image Prioritizer description too?

plugins/performance-lab/includes/admin/plugins.php
westonruter: Cache still not expired! Wow. I wonder if it only is invalidated once one of the plugins in the response is touched.

westonruter: Maybe I should directly commit the description update to the 0.1.0 tag to see if that invalidates it.

mukeshpanchal27 (approved):

PR #1585, opened 2024-10-09
Introduce OD_Element class and improve PHP API

westonruter:
See https://github.com/WordPress/performance/pull/1373/files#r1792652840:

> See note below on the `get_all_denormalized_elements()` method that I'd like to instead have `$element` be an `OD_Element` class instance which has a `$url_metric` property to access the `OD_URL_Metric` instance if is a part of. In the same way, the `OD_URL_Metric` class can have a `$group` property to indicate the `OD_URL_Metric_Group` it is a part of. This would eliminate the need to pass back an array of tuples and instead. So instead, to get the `$group` this code could do `$element->url_metric->group`.

The `OD_Element` class implements `ArrayAccess` so it is fully backwards-compatible with using values as arrays. So where existing plugins may do `$element['isLCP']` this will still continue to work, although `$element->is_lcp()` would now be preferable. There is also a `$element->get()` method which allows accessing extended element data. This is closely related to #1492 which did the same for `OD_URL_Metric`.

### DIFF
--- a/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
+++ b/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
@@ -14,6 +14,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 /**
  * Tag visitor that optimizes embeds.
  *
+ * @phpstan-import-type DOMRect from OD_URL_Metric
+ *
  * @since 0.2.0
  * @access private
  */
@@ -216,6 +218,11 @@ final class Embed_Optimizer_Tag_Visitor {
 
 		$elements = $context->url_metric_group_collection->get_all_elements()[ $embed_wrapper_xpath ] ?? array();
 		foreach ( $elements as $element ) {
+			/**
+			 * Resized bounding client rect.
+			 *
+			 * @var DOMRect|null $resized_bounding_client_rect
+			 */
 			$resized_bounding_client_rect = $element->get( 'resizedBoundingClientRect' );
 			if ( ! is_array( $resized_bounding_client_rect ) ) {
 				continue;
--- a/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
+++ b/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
@@ -216,7 +216,8 @@ final class Embed_Optimizer_Tag_Visitor {
 
 		$elements = $context->url_metric_group_collection->get_all_elements()[ $embed_wrapper_xpath ] ?? array();
 		foreach ( $elements as $element ) {
-			if ( ! isset( $element['resizedBoundingClientRect'] ) ) {
+			$resized_bounding_client_rect = $element->get( 'resizedBoundingClientRect' );
+			if ( ! is_array( $resized_bounding_client_rect ) ) {
 				continue;
 			}
 			$group           = $element->url_metric->group;
@@ -224,12 +225,12 @@ final class Embed_Optimizer_Tag_Visitor {
 			if ( ! isset( $minimums[ $group_min_width ] ) ) {
 				$minimums[ $group_min_width ] = array(
 					'group'  => $group,
-					'height' => $element['resizedBoundingClientRect']['height'],
+					'height' => $resized_bounding_client_rect['height'],
 				);
 			} else {
 				$minimums[ $group_min_width ]['height'] = min(
 					$minimums[ $group_min_width ]['height'],
-					$element['resizedBoundingClientRect']['height']
+					$resized_bounding_client_rect['height']
 				);
 			}
 		}
--- a/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
+++ b/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
@@ -214,11 +214,12 @@ final class Embed_Optimizer_Tag_Visitor {
 		 */
 		$minimums = array();
 
-		$denormalized_elements = $context->url_metric_group_collection->get_all_denormalized_elements()[ $embed_wrapper_xpath ] ?? array();
-		foreach ( $denormalized_elements as list( $group, $url_metric, $element ) ) {
+		$elements = $context->url_metric_group_collection->get_all_elements()[ $embed_wrapper_xpath ] ?? array();
+		foreach ( $elements as $element ) {
 			if ( ! isset( $element['resizedBoundingClientRect'] ) ) {
 				continue;
 			}
+			$group           = $element->url_metric->group;
 			$group_min_width = $group->get_minimum_viewport_width();
 			if ( ! isset( $minimums[ $group_min_width ] ) ) {
 				$minimums[ $group_min_width ] = array(
--- a/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
+++ b/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
@@ -216,7 +216,7 @@ final class Embed_Optimizer_Tag_Visitor {
 		 */
 		$minimums = array();
 
-		$elements = $context->url_metric_group_collection->get_all_elements()[ $embed_wrapper_xpath ] ?? array();
+		$elements = $context->url_metric_group_collection->get_xpath_elements_map()[ $embed_wrapper_xpath ] ?? array();
 		foreach ( $elements as $element ) {
 			/**
 			 * Resized bounding client rect.
--- a/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
+++ b/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
@@ -227,7 +227,10 @@ final class Embed_Optimizer_Tag_Visitor {
 			if ( ! is_array( $resized_bounding_client_rect ) ) {
 				continue;
 			}
-			$group           = $element->url_metric->group;
+			$group = $element->get_url_metric_group();
+			if ( null === $group ) {
+				continue; // Technically could be null but in practice it never will be.
+			}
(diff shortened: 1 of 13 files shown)


plugins/optimization-detective/class-od-url-metric.php
```diff
@@ -406,10 +414,15 @@ public function get_timestamp(): float {
 	/**
 	 * Gets elements.
 	 *
-	 * @return ElementData[] Elements.
+	 * @return OD_Element[] Elements.
 	 */
 	public function get_elements(): array {
```
westonruter, line 457: I don't like that `$url_metric->get( 'elements' )` returns an array of arrays (the underlying values), whereas `$url_metric->get_elements()` returns the data wrapped in `OD_Element` objects.

plugins/optimization-detective/class-od-url-metric-group-collection.php
```diff
@@ -417,26 +417,25 @@ public function get_common_lcp_element(): ?array {
...
 	 * end up running n*4*3 times.
 	 *
-	 * @todo Should there be an OD_Element class which has a $url_metric property which then in turn has a $group property. Then this would only need to return array<string, OD_Element[]>.
 	 * @since n.e.x.t
 	 *
-	 * @return array<string, non-empty-array<int, array{OD_URL_Metric_Group, OD_URL_Metric, ElementData}>> Keys are XPaths and values are arrays of tuples consisting of the group, URL metric, and element data.
+	 * @return array<string, non-empty-array<int, OD_Element>> Keys are XPaths and values are the element instances.
 	 */
-	public function get_all_denormalized_elements(): array {
+	public function get_all_elements(): array {
```
westonruter: I don't really like the `get_all_elements` name. It's more like `get_elements_grouped_by_xpath`.

plugins/optimization-detective/class-od-element.php
```diff
@@ -0,0 +1,115 @@
...
+
+	/**
+	 * Retrieves an offset.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param mixed $offset Key.
+	 * @return mixed Can return all value types.
+	 */
+	public function offsetGet( $offset ) { // phpcs:ignore SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint
```
westonruter: I implemented `ArrayAccess` here for back-compat with existing code accessing the offsets. But for the way we access data, perhaps we should have accessor `get` methods similar to what is done for `OD_URL_Metric`. So this would have:

* `is_lcp()`
* `is_lcp_candidate()`
* `get_xpath()`
* `get_intersection_ratio()`
* `get_intersection_rect()`
* `get_bounding_client_rect()`

As well as a `get()` method which would return arbitrary values, especially those which are from an extension-extended schema, like `$element->get( 'resizedBoundingClientRect )` in Embed Optimizer.

westonruter: Done in 57964fb8a6d8a6a1a4326be66e6041d30583c8fc

plugins/optimization-detective/class-od-url-metric.php
westonruter: Addressed in 5dfcb07e

westonruter:
@felixarntz Thoughts on this?

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -214,21 +214,23 @@ private function reduce_layout_shifts( OD_Tag_Visitor_Context $context ): void {
...
 				$minimums[ $group_min_width ] = array(
 					'group'  => $group,
-					'height' => $element['resizedBoundingClientRect']['height'],
+					'height' => $resized_bounding_client_rect['height'],
 				);
 			} else {
 				$minimums[ $group_min_width ]['height'] = min(
 					$minimums[ $group_min_width ]['height'],
-					$element['resizedBoundingClientRect']['height']
+					$resized_bounding_client_rect['height']
```
felixarntz: Just curious: Wouldn't all this marked code also have worked in the way it was before this change, given `OD_Element` implements `ArrayAccess`?

plugins/optimization-detective/class-od-element.php
```diff
@@ -0,0 +1,209 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param mixed $offset Key.
+	 * @param mixed $value  Value.
+	 *
+	 * @throws Exception When attempting to set a property.
+	 */
+	public function offsetSet( $offset, $value ): void {
+		throw new Exception( 'Offsets may not be set.' );
```
felixarntz: Maybe more clear as something like this?
```suggestion
		throw new Exception( 'Element data may not be set.' );
```

plugins/optimization-detective/class-od-element.php
```diff
@@ -0,0 +1,209 @@
...
+	 * This is disallowed. Attempting to unset a property will throw an error.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param mixed $offset Offset.
+	 *
+	 * @throws Exception When attempting to unset a property.
+	 */
+	public function offsetUnset( $offset ): void {
+		throw new Exception( 'Offsets may not be unset.' );
```
felixarntz: See above:
```suggestion
		throw new Exception( 'Element data may not be unset.' );
```

plugins/optimization-detective/class-od-url-metric-group-collection.php
felixarntz: How about `get_all_elements_by_xpath()`?

plugins/optimization-detective/class-od-element.php
```diff
@@ -0,0 +1,209 @@
...
+	protected $data;
+
+	/**
+	 * URL metric that this element belongs to.
+	 *
+	 * @since n.e.x.t
+	 * @var OD_URL_Metric
+	 * @readonly
+	 */
+	public $url_metric;
```
felixarntz: Let's not use a public property, as we wouldn't want this to be writable (the annotation is not enough). Can you instead add e.g. a `get_url_metric()` method? For convenience, you may even want to add a `get_url_metric_group()` or `get_group()` method as a shorthand, given that `$group` was previously easily accessible in the return value of the `get_all_denormalized_elements()` method.

felixarntz (commented):
@westonruter This looks great so far. Only a few points of feedback.

plugins/optimization-detective/class-od-url-metric-group-collection.php
westonruter: That works, but then it would seem like you could pass an `$xpath` to get just the elements with the given XPath.

westonruter: Like `document.getElementById()`

felixarntz: That's fair, though I'm not sure how including the word `grouped` would clarify that. FWIW to me that `all` helps to clarify this will return "all" elements, not just the elements filtered by something.

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
westonruter: Yes, it would have worked the same. I made this change because of a PHPStan issue where attempting to access `$element['resizedBoundingClientRect']['height']` results in a PHPStan error:

> phpstan: Cannot access offset 'height' on array{width: float, height: float, x: float, y: float, top: float, right: float, bottom: float, left: float}|bool|float|non-empty-string.

PHPStan isn't constraining the type based on the type of `$element['resizedBoundingClientRect']` based on the key `resizedBoundingClientRect` so the possible types are a union of all values in `ElementData`, which also doesn't necessarily include this extended `resizedBoundingClientRect` key's type (although here it is a DOMRect).

This goes back to wanting to be able to provide richer typing in PHPStan for getters like this, but in the meantime resorting to using `OD_Element::get()` which returns `mixed` and then checking the return value as being an array avoids the PHPStan error. 

Although I'm a bit surprised it's not complaining that the `$resized_bounding_client_rect['height']` key may not be defined since if I do `PHPStan\dumpType($resized_bounding_client_rect)` it is typed as just `array`. But this appears to be as expected: <https://phpstan.org/r/bc084e48-691a-4232-be15-7b806fda8b3e>. I recall there was some PHPStan extension to make checks even more strict than strict rules mandate all array keys be checked as well, but I can't find how to do that off hand.

plugins/optimization-detective/class-od-element.php
westonruter: What about the public `$group` property on `OD_URL_Metric`? It has to be mutable since `\OD_URL_Metric_Group::add_url_metric()` needs to be able to assign it to the group when it is added.

felixarntz: I must have missed that one before. Better to use a setter then for `OD_URL_Metric::$group`.

For this one though, would we even need to set it?

westonruter: No, I don't think there is a need for a `set_url_metric()`. I'll add a getter for both and a setter for one.

westonruter: Done in 344d6e101844eaa844e33ce7866baa2c832841ff

plugins/optimization-detective/class-od-url-metric-group-collection.php
westonruter: How about `get_xpath_elements_map()`?

westonruter: Done in 98d3fd8a

felixarntz (approved):
@westonruter Thanks, LGTM!

Issue #1284, opened 2024-06-06
Unit tests lack code coverage reporting and @covers annotations


westonruter:
## Feature Description

In looking at [`plugins/webp-uploads/tests/test-load.php`](https://github.com/WordPress/performance/blob/c6fe7bd38cc4f6ea08c8b501895cf72f320b9344/plugins/webp-uploads/tests/test-load.php), I noticed most of the tests lack `@covers` annotations. We should add those throughout our unit tests to better keep track of the code being tested. We should also add code coverage reporting for the repo (with Codecov). The `phpunit` calls in `composer.json` should also add `--strict-coverage` to report warnings when `@covers` tags are wrong.

AhmarZaidi:
Hey, I'd like to work on this issue.

westonruter:
@sarthak-19 @thelovekesh Since the Performance repo is a monorepo of plugins, and since PHPUnit is going to be run separately for each plugin and thus there will be multiple coverage files, does this mean there could be multiple coverage reports? I think that would be somewhat helpful if each plugin's coverage was reported separately, although this for me is only a nice-to-have.

thelovekesh:
Yes @westonruter, that should be achievable.

sarthak-19:
### Problem
I managed to run the test locally and created a coverage-report.xml, previously report was not getting generated that's why one test was failing.

But for testing purpose when I pushed the code to my [draft PR](https://github.com/WordPress/performance/pull/1653), all the test is failing.

### Reason 
For PHP 8.1^ it's failing because PHPUnit 10 is getting used instead of 9, and for below 8.1 PHP I'm not able to understand why it's failing.

- When PHP Env is ^8.0, required PHPUnit should be 9.0^ for generating coverage report (PHPUnit ^10.0 throwing errors)
Previously I was getting 
> This version of PHPUnit does not support code coverage on PHP 8
PHPUnit Version - 8.5.39
PHP Version -8.2.24 

- When PHP Env. is ^7.0, then PHPUnit ^9.0 will work or not, not sure.

### Questions 

1. When different test case runs for different PHP versions in GH actions, how can I set PHPUnit versions for each test?
2. For generating coverage report, xdebug must be set else we'll get `No coverage driver available`. 
So for testing purpose I've modified `package.json` as --> `"wp-env": "wp-env start --xdebug=develop,coverage",`, But I assume this is not the correct way to do it, so how can I enable xdebug only for env where code coverage is needed.
[...]

PR #1586, opened 2024-10-10
Add @cover annotation in test-load file and add support for codecov

sarthak-19:
## Summary
This PR adds `@cover` annotation for test cases and add .yml file that can be used to configure codecov for automated test coverage in github actions.

Addresses https://github.com/WordPress/performance/issues/1284

## Relevant technical choices

I've addressed the following file `plugins/webp-uploads/tests/test-load.php`  for adding `@cover` annotation, if the approach is approved will extend this to other files.


composer.json
```diff
@@ -31,8 +31,10 @@
     "wp-coding-standards/wpcs": "^3.1",
     "wp-phpunit/wp-phpunit": "^6.5",
     "yoast/phpunit-polyfills": "^2.0",
-    "phpstan/php-8-stubs": "^0.3.84",
-    "phpstan/phpstan-strict-rules": "^1.6"
+    "phpstan/php-8-stubs": "^0.4.0",
+    "phpstan/phpstan-strict-rules": "^1.6",
+    "phpunit/phpunit": "^9.5",
```
swissspidy: This isn't needed, PHPUnit is installed via `yoast/phpunit-polyfills` above

composer.json
```diff
@@ -31,8 +31,10 @@
...
+    "codecov/codecov": "^1.0"
```
swissspidy: This isn't needed either, nor does it actually exist

.github/workflows/codecov-coverage.yml
```diff
@@ -0,0 +1,39 @@
+name: PHP Test and Coverage
```
swissspidy: Can we run coverage in the existing `php-test-plugins.yml` workflow instead?

swissspidy:
@sarthak-19 Welcome, and thanks for opening this! I just stumbled upon the PR and left a couple of comments for things that caught my eye.

sarthak-19:
> @sarthak-19 Welcome, and thanks for opening this! I just stumbled upon the PR and left a couple of comments for things that caught my eye.

@swissspidy Thank you so much for the feedback. I've addressed those and looking forward to keep improving 😄

.github/workflows/php-test-plugins.yml
```diff
@@ -68,6 +68,20 @@ jobs:
       - name: Composer Install
         run: npm run wp-env run tests-cli -- --env-cwd="wp-content/plugins/$(basename $(pwd))" composer install --no-interaction --no-progress
       - name: Running single site unit tests
-        run: npm run test-php
+        run: npm run test-php -- --coverage-clover=coverage.xml
```
westonruter: Since collecting code coverage takes much longer, as I recall, should the coverage only happen for one of the jobs instead of for all of them?

sarthak-19: Yes you're right, collecting code coverage takes time, but I'm not quite sure that for which job it'll be ideal to run.
Should I run for `PHP v8.0` and if this is a viable option, then will the below implementation work since we are using the latest WP version anyways?
```yml

      - name: Running single site unit tests with coverage
        if: matrix.php == '8.0'
        run: npm run test-php -- --coverage-clover=coverage.xml
      - name: Running multisite unit tests with coverage
        if: matrix.php == '8.0'
        run: npm run test-php-multisite -- --coverage-clover=coverage-multisite.xml

      - name: Upload single site coverage reports to Codecov
        if: matrix.php == '8.0'
        uses: codecov/codecov-action@v4
        with:
          token: ${{ secrets.CODECOV_TOKEN }}
          files: coverage.xml
          flags: unittests
          name: ${{ matrix.php }}-single-site-coverage
      - name: Upload multisite coverage reports to Codecov
        if: matrix.php == '8.0'
        uses: codecov/codecov-action@v2
        with:
          token: ${{ secrets.CODECOV_TOKEN }}
          files: coverage-multisite.xml
          flags: multisite
          name: ${{ matrix.php }}-multisite-coverage
```

cc : @westonruter

pbearne:
I am really glad to see I have been try to get into core ten+ years

.github/workflows/php-test-plugins.yml
westonruter: You may want to refer to what was done for the AMP plugin: https://github.com/ampproject/amp-wp/blob/60a655cdad9e3431b47b2dcc3b2e5ddb86ac5c5e/.github/workflows/build-test-measure.yml#L261-L337

Note the new `coverage` matrix variable which is only set to `true` for one configuration. This would be better than having to test against a specific PHP version everywhere. Yes, I think running coverage on the most recent PHP version would be best as it should complete the fastest. So maybe do it in PHP 8.2?

cc @thelovekesh

sarthak-19:
> Note the new coverage matrix variable which is only set to true for one configuration. This would be better than having to test against a specific PHP version everywhere. Yes, I think running coverage on the most recent PHP version would be best as it should complete the fastest. So maybe do it in PHP 8.2?

Apologies for the delayed response, made the suggested changes.
Does this works or any other changes are to be made, highly appreciate your insights here.

cc : @westonruter

westonruter:
@sarthak-19 That looks right!

.github/workflows/php-test-plugins.yml
```diff
@@ -68,6 +72,32 @@ jobs:
...
+          files: coverage.xml
+          flags: unittests
+          name: ${{ matrix.php }}-single-site-coverage
+      - name: Upload multisite coverage reports to Codecov
+        if: ${{ matrix.coverage == true }}
+        uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: coverage-multisite.xml
+          flags: multisite
```
westonruter, line 103: Is `multisite` a valid flag?  Shouldn't this be `unittests`?

```suggestion
          flags: unittests
```

.github/workflows/php-test-plugins.yml
```diff
@@ -43,11 +43,15 @@ jobs:
...
         wp: [ 'latest' ]
+        coverage: [false]
         include:
           - php: '7.4'
             wp: '6.5'
           - php: '8.3'
             wp: 'trunk'
+          - php: '8.2'
+            wp: 'latest'
+            coverage: true
```
felixarntz, line 54: I wonder, given we include this as an extra scenario, should we maybe remove `8.2` from the regular `php` list above? Otherwise we'd be running the exact same test (WP latest, PHP 8.2) twice, only with and without code coverage, which shouldn't change the outcome of the tests.

If there's any benefit to keeping both (which would only be because of the implications that running code coverage could have), then I think the better choice would be to implement a separate workflow entirely.

.github/workflows/php-test-plugins.yml
```diff
@@ -68,6 +72,32 @@ jobs:
       - name: Composer Install
         run: npm run wp-env run tests-cli -- --env-cwd="wp-content/plugins/$(basename $(pwd))" composer install --no-interaction --no-progress
       - name: Running single site unit tests
-        run: npm run test-php
+        run: |
+          if [ "${{ matrix.coverage }}" == "true" ]; then
+            npm run test-php -- --coverage-clover=coverage.xml
```
felixarntz: Maybe append commit SHA to avoid chance of conflicts?
```suggestion
            npm run test-php -- --coverage-clover=coverage-${{ github.sha }}.xml
```

.github/workflows/php-test-plugins.yml
```diff
@@ -68,6 +72,32 @@ jobs:
...
+          else
+            npm run test-php
+          fi
       - name: Running multisite unit tests
-        run: npm run test-php-multisite
+        run: |
+          if [ "${{ matrix.coverage }}" == "true" ]; then
+            npm run test-php-multisite -- --coverage-clover=coverage-multisite.xml
```
felixarntz: See above:
```suggestion
            npm run test-php-multisite -- --coverage-clover=coverage-multisite-${{ github.sha }}.xml
```

.github/workflows/php-test-plugins.yml
```diff
@@ -68,6 +72,32 @@ jobs:
...
+          else
+            npm run test-php-multisite
+          fi
+      - name: Upload single site coverage reports to Codecov
+        if: ${{ matrix.coverage == true }}
+        uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: coverage.xml
```
felixarntz: See above:
```suggestion
          files: coverage-${{ github.sha }}.xml
```

.github/workflows/php-test-plugins.yml
```diff
@@ -68,6 +72,32 @@ jobs:
...
+          flags: unittests
+          name: ${{ matrix.php }}-single-site-coverage
+      - name: Upload multisite coverage reports to Codecov
+        if: ${{ matrix.coverage == true }}
+        uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: coverage-multisite.xml
```
felixarntz: See above:
```suggestion
          files: coverage-multisite-${{ github.sha }}.xml
```

.github/workflows/php-test-plugins.yml
```diff
@@ -68,6 +72,32 @@ jobs:
...
+          else
+            npm run test-php-multisite
+          fi
+      - name: Upload single site coverage reports to Codecov
+        if: ${{ matrix.coverage == true }}
+        uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: coverage.xml
+          flags: unittests
```
felixarntz: How about we use `single` instead of `unittests`, since the differentiator between the two datasets is "single site" vs "multisite", not "unit tests" vs "multisite".
```suggestion
          flags: single
```

.github/workflows/php-test-plugins.yml
felixarntz: These are arbitrary values to separate test scenarios, so I think passing `multisite` makes sense here (separating from "single site").

.github/workflows/php-test-plugins.yml
```diff
@@ -68,6 +72,32 @@ jobs:
...
+          name: ${{ matrix.php }}-single-site-coverage
```
felixarntz, line 95: Let's make sure we're alerted of failures for coverage reporting:
```suggestion
          name: ${{ matrix.php }}-single-site-coverage
          fail_ci_if_error: true
```

.github/workflows/php-test-plugins.yml
```diff
@@ -68,6 +72,32 @@ jobs:
...
+      - name: Upload multisite coverage reports to Codecov
+        if: ${{ matrix.coverage == true }}
+        uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: coverage-multisite.xml
+          flags: multisite
+          name: ${{ matrix.php }}-multisite-coverage
```
felixarntz, line 104: See above:
```suggestion
          name: ${{ matrix.php }}-multisite-coverage
          fail_ci_if_error: true
```

felixarntz (commented):
@sarthak-19 Thank you for the PR! This looks quite solid, a few suggestions and questions.

.github/workflows/php-test-plugins.yml
westonruter: How would there be a conflict? Each GHA job should run from a pristine state, no? For the `coverage.xml` file to persist across runs this would require some explicit caching.

felixarntz: Probably. I find this safer, but either works for me.

.github/workflows/php-test-plugins.yml
sarthak-19: Yes make sense, removing the 8.2 version from PHP list.

.github/workflows/php-test-plugins.yml
sarthak-19: > I find this safer.

I think better safe than sorry.
So implementing this as well. 😀

westonruter:
FYI: I've added the `CODECOV_TOKEN` to the repository secrets.

westonruter (approved):
I suppose the one job will continue to fail since this PR was opened from a fork. I'll leave it to Felix to approve and merge.

felixarntz (approved):
Thanks for the PR @sarthak-19, LGTM!

Before we merge though: Doesn't one of us has to configure / set up this project on https://about.codecov.io/ first? 🤔

westonruter:
@felixarntz I _believe_ it's all already configured according to the steps listed at https://app.codecov.io/gh/WordPress/performance

The only thing outstanding is the third step, which is to merge into the preferred branch:

> image

I'll do so now!

westonruter:
Alas, no: https://github.com/WordPress/performance/actions/runs/11823769472/job/32943781732

westonruter:
I've been chatting with @thelovekesh about what's going on. 

For one, we're missing some additional `--` arg separators since we're passing from `npm` to `composer` and then to `phpunit`. So here:

https://github.com/WordPress/performance/blob/745c145709e8d6eec655c109f472ca77b151178d/.github/workflows/php-test-plugins.yml#L77

https://github.com/WordPress/performance/blob/745c145709e8d6eec655c109f472ca77b151178d/.github/workflows/php-test-plugins.yml#L84

These need to instead be, respectively:

```bash
npm run test-php -- -- --coverage-clover=coverage-${{ github.sha }}.xml 
```

and 

```bash
npm run test-php-multisite -- -- --coverage-clover=coverage-multisite-${{ github.sha }}.xml
```

Additionally, the coverage report for each plugin is getting overwritten to the same XML file. There needs to be separate XML files generated for each plugin.

Lovekesh has some additional thoughts.

westonruter:
I've opened https://github.com/WordPress/performance/pull/1652 to disable Codecov for the time being.

PR #1589, opened 2024-10-15
Add TypeScript checking

westonruter:
Disclaimer: This is my first time setting up TypeScript for a project. So I probably missed some things!

* Add TypeScript as dependency.
* Add `tsc` to the npm scripts.
* Include both `js` and `ts` files in `lint-staged`.
* Run `npm run tsc` in `lint-staged`. Problem: Passing the files does not work, as TypeScript forces them to be considered TS files and not JS files. See comment.
* Rename misnamed `types.d.ts` to `types.ts` and apply Prettier formatting.


lint-staged.config.js
```diff
@@ -25,7 +25,12 @@ const joinFiles = ( files ) => {
 const PLUGIN_BASE_NAME = path.basename( __dirname );
 
 module.exports = {
-	'**/*.js': ( files ) => `npm run lint-js -- ${ joinFiles( files ) }`,
+	'**/*.{js,ts}': ( files ) => {
+		return [
+			`npm run lint-js -- ${ joinFiles( files ) }`,
+			`npm run tsc`, // TODO: How to pass joinFiles( files ) here?
```
westonruter: I wanted this to rather be:

```suggestion
			`npm run tsc -- ${ joinFiles( files ) }`,
```

But when I do this, I get an error:

```
✖ npm run tsc -- '/home/westonruter/repos/performance/lint-staged.config.js':

> tsc
> tsc /home/westonruter/repos/performance/lint-staged.config.js

error TS6504: File '/home/westonruter/repos/performance/lint-staged.config.js' is a JavaScript file. Did you mean to enable the 'allowJs' option?
  The file is in the program because:
    Root file specified for compilation
husky - pre-commit script failed (code 1)
```

Maybe what I want to do is not supported in TypeScript in the same way that PHPStan needs to run on the entire project and not just the changed files?

westonruter: If this is the case, then:

```suggestion
			`npm run tsc`,
```

thelovekesh:
> Rename misnamed types.d.ts to types.ts.

If the `type.ts` is purely for type checking then I think we can keep it as `.d.ts` .

lint-staged.config.js
swissspidy: > But when I do this, I get an error:

`[..] is a JavaScript file. Did you mean to enable the 'allowJs' option?` is pretty clear`.

If you don't want to run tsc against JS files, then don't set up lint-staged + tsc for JS files.

----

Never used tsc with lint-staged, but https://github.com/lint-staged/lint-staged/issues/825 might be helpful. Some people use https://www.npmjs.com/package/tsc-files.

westonruter: I did enable `allowJs` in the tsconfig already though.

westonruter:
> If the `type.ts` is purely for type checking then I think we can keep it as `.d.ts` .

Unfortunately then TypeScript won't check it at all when `skipLibCheck` is enabled, and that is required to prevent TypeScript from checking code in `node_modules`. So I renamed it to `.ts` so that TypeScript will check it.

lint-staged.config.js
swissspidy: Then I think the issue is that the tsconfig file is not read when you pass it a list of files, according to https://github.com/lint-staged/lint-staged/issues/825. That's why some use https://www.npmjs.com/package/tsc-files for that.

felixarntz (commented):
@westonruter Looks reasonable to me. But I'm not the best person to review this, would be great to get @swissspidy's approval on this.

lint-staged.config.js
swissspidy: We could also consider removing tsc from lint-staged for now, probably not worth the hassle. Or just run it for the whole project because it's a tiny project anyway.

westonruter: OK, Let's just run it on the entire project then, same as we do for PHPStan.

westonruter:
Also, related to `.ts` vs `.d.ts`, it turns out that `wp-scripts` only lints `.js`, `.jsx`, `.ts`, and `.tsx` files. As soon as I renamed `types.d.ts` to `types.ts` then I got linting errors which could then be automatically fixed with `wp-scripts format`.

swissspidy (approved):

Issue #1592, opened 2024-10-15
Appropriate preload value should be selected for video based on whether it is in the viewport


westonruter:
I noticed that the default preload value for videos in the Video block is set to "Metadata":

![Image](https://github.com/user-attachments/assets/1126e0d5-6a4c-4712-801b-c5ffee140aaa)

Nevertheless, this does not result in `preload=metadata` being added to the `VIDEO` tag in Gutenberg but rather no `preload` attribute at all. This leaves it up to the browser to decide what the default behavior is. In the worst case, the browser could decide the default is `preload=auto` which [according](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#preload) to MDN, this "Indicates that the whole video file can be downloaded, even if the user is not expected to use it." This is not good as it means bytes are needlessly being downloaded. Nevertheless, Chrome's default is [`preload=metadata`](https://googlechrome.github.io/samples/media/preload-metadata) which is what the spec encourages browsers to use. When using `preload=auto`

On an example page that has a Video with a `poster` outside the viewport, the following is the impact on the total number of bytes downloaded for a video that is 59 MB in size:

`preload=auto` (4.5 MB) | `preload=metadata` (1.5 MB) | `preload=none` (0.9 MB)
--|--|--
![Image](https://github.com/user-attachments/assets/1e758349-88dd-43a6-872c-c516b787c32d) | ![Image](https://github.com/user-attachments/assets/4e9888b7-3ca3-4515-99f7-ed0bbd00fba4) | ![Image](https://github.com/user-attachments/assets/fe9251fd-6098-4ac3-9df2-321e3490175c)

Note: The values of `preload=auto` and `preload=metadata` actually often seem to result in the same bytes being downloaded by Chrome. What is consistent, however, is that `preload=none` saves on a lot of bytes. Therefore, it may not make much of a difference setting `preload=auto` versus `preload=metadata` for videos in the initial viewport, but what surely will make a difference is to set `preload=none` for a video that doesn't appear in any initial viewport breakpoint.

The `preload=none` value should be automatically changed to `preload=metadata` (or `preload=auto`) when a video enters the viewport (via `IntersectionObserver`). This will ensure that the video is ready to play when the user sees it (similarly to https://github.com/WordPress/performance/issues/1590 which would start autoplaying when the video nears the viewport):

`preload=none` | `preload=metadata`
--|--
[...]

PR #1593, opened 2024-10-16
Try: Add fetchpriority=high for LCP video elements that autoplay

westonruter:
This PR attempts to implement this aspect of #1592: 

> When a video is in all initial viewports, it should probably get `preload=auto` especially when it is the LCP element. It may even make sense to add `fetchpriority=high` preload links when it is the LCP element, particularly when the video has `autoplay`.

However, when I implemented this I discovered that it is not supported in Chrome!

> `<link rel=preload>` uses an unsupported `as` value

Apparently this used to be part of the Preload spec but was removed. See [StackOverflow answer](https://stackoverflow.com/a/68368601/93579):

> As of July 13, 2021, https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preload has a note about 1/3rd of the way down that says "Note: video preloading is included in the Preload spec, but is not currently implemented by browsers." Perhaps that is why you see this warning (I see it too)? If so, then no, it is not possible to use preload this.

There is currently no mention of `video` on that MDN page.

### DIFF
--- a/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
@@ -35,36 +35,101 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 			return false;
 		}
 
-		// Skip empty poster attributes and data: URLs.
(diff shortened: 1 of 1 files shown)


westonruter:
Maybe later.

PR #1404, opened 2024-07-26
Fix handling of image prioritization when only some viewport groups are populated

westonruter:
I discovered an issue on my blog in which the most recent blog post lacks a featured image but the second-most recent post does have a featured image. The featured image on the second post is getting `fetchpriority=high` added by server-side heuristics but Image Prioritizer was not removing this even though the image was not visible on the mobile viewport, although it is correctly the LCP element on desktop:

Mobile | Desktop
--|--
image | image

My URL metrics were fully populated on the mobile and desktop viewport groups, but they are _empty_ for phablet and tablet viewport groups. I don't get a lot of traffic on my site, and apparently _nobody_ with a phablet or tablet visits my site. So I have no metrics for them:

<details><summary>URL Metrics Dump</summary>

```json
{
    "breakpoints": [
        480,
        600,
        782
    ],
    "freshness_ttl": 86400,
    "sample_size": 3,
    "all_element_max_intersection_ratios": {
        "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[3][self::ARTICLE]/*[2][self::FIGURE]/*[1][self::DIV]/*[1][self::IMG]": 0.74560546875,
        "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[5][self::ARTICLE]/*[2][self::FIGURE]/*[1][self::DIV]/*[1][self::IMG]": 0,
        "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[7][self::ARTICLE]/*[2][self::FIGURE]/*[1][self::DIV]/*[1][self::IMG]": 0,
        "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[9][self::ARTICLE]/*[2][self::FIGURE]/*[1][self::DIV]/*[1][self::IMG]": 0,
        "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[13][self::ARTICLE]/*[2][self::FIGURE]/*[1][self::DIV]/*[1][self::IMG]": 0,
        "/*[1][self::HTML]/*[2][self::BODY]/*[4][self::MAIN]/*[15][self::ARTICLE]/*[2][self::FIGURE]/*[1][self::DIV]/*[1][self::IMG]": 0,
[...]

PR #1595, opened 2024-10-16
Choose smaller poster image size based on actual dimensions

### LINKED ISSUE #1575: Improve video poster image URL based on actual dimensions
## Feature Description

On [my recent blog post](https://pascalbirchler.com/web-ai-wordpress/) just noticed that my videos are as wide as the content width, but the poster image is still the full size URL. This could be improved by using a more appropriate image size. The `video[poster]` attribute doesn't support `srcset` or anything, so this needs to be done differently.

We could:

1. Detect width of the video
2. Check the poster URL and get the attachment ID from that URL
3. Get a smaller sized version of that image and use that as the poster instead.

@westonruter Something for Optimization Detective?

@joemcgill @mukeshpanchal27 Something that could perhaps also be done with the thing you're looking into where we take a block's layout into account for image sizes?

swissspidy:
## Summary

Fixes #1575

On my test post with 7 large videos, the videos were displayed ~610px wide on desktop. With this PR, the full image URLs (which were 1920px wide) are replaced with 767px wide versions. 

Image file sizes before this PR: 738 KB

After: 175 KB (!)

### DIFF
--- a/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
@@ -45,6 +45,8 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 	/**
 	 * Gets the poster from the current VIDEO element.
 	 *
+	 * Skips empty poster attributes and data: URLs.
+	 *
 	 * @since n.e.x.t
 	 *
 	 * @param OD_Tag_Visitor_Context $context Tag visitor context.
@@ -69,12 +71,6 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 	private function reduce_poster_image_size( OD_Tag_Visitor_Context $context ): bool {
 		$processor = $context->processor;
 
-		// Skip empty poster attributes and data: URLs.
-		$poster = trim( (string) $processor->get_attribute( 'poster' ) );
-		if ( '' === $poster || $this->is_data_url( $poster ) ) {
-			return false;
-		}
-
 		$xpath = $processor->get_xpath();
 
 		$max_element_width = 0;
@@ -83,6 +79,12 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 
 		foreach ( $denormalized_elements as list( , , $element ) ) {
 			$max_element_width = max( $max_element_width, $element['boundingClientRect']['width'] ?? 0 );
+		}
+
+		$poster = $this->get_poster( $context );
+
+		if ( null === $poster ) {
+			return false;
 		}
 
 		$poster_id = attachment_url_to_postid( $poster );
@@ -106,9 +108,9 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 	private function preload_poster_image( OD_Tag_Visitor_Context $context ): bool {
 		$processor = $context->processor;
 
-		// Skip empty poster attributes and data: URLs.
-		$poster = trim( (string) $processor->get_attribute( 'poster' ) );
-		if ( '' === $poster || $this->is_data_url( $poster ) ) {
+		$poster = $this->get_poster( $context );
+
+		if ( null === $poster ) {
 			return false;
 		}
 
--- a/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
@@ -42,13 +42,28 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 
 		return $reduced_poster_size || $preload_poster;
 	}
-
 	/**
-	 * Reduce poster image size by choosing one that fits the maximum video size more closely.
+	 * Gets the poster from the current VIDEO element.
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @param OD_Tag_Visitor_Context $context Tag visitor context, with the cursor currently at an embed block.
+	 * @param OD_Tag_Visitor_Context $context Tag visitor context.
+	 * @return non-empty-string|null Poster or null if not defined or is a data: URL.
+	 */
+	private function get_poster( OD_Tag_Visitor_Context $context ): ?string {
+		$poster = trim( (string) $context->processor->get_attribute( 'poster' ) );
+		if ( '' === $poster || $this->is_data_url( $poster ) ) {
+			return null;
+		}
+		return $poster;
+	}
+
+	/**
+	 * Reduces poster image size by choosing one that fits the maximum video size more closely.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param OD_Tag_Visitor_Context $context Tag visitor context, with the cursor currently at a VIDEO tag.
 	 * @return bool Whether the tag should be tracked in URL metrics.
 	 */
 	private function reduce_poster_image_size( OD_Tag_Visitor_Context $context ): bool {
@@ -81,11 +96,11 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 	}
 
 	/**
-	 * Preload poster image for the LCP <video> element.
+	 * Preloads poster image for the LCP <video> element.
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @param OD_Tag_Visitor_Context $context Tag visitor context, with the cursor currently at an embed block.
+	 * @param OD_Tag_Visitor_Context $context Tag visitor context, with the cursor currently at a VIDEO tag.
 	 * @return bool Whether the tag should be tracked in URL metrics.
 	 */
 	private function preload_poster_image( OD_Tag_Visitor_Context $context ): bool {
--- a/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
@@ -51,7 +51,7 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 	 * @param OD_Tag_Visitor_Context $context Tag visitor context, with the cursor currently at an embed block.
 	 * @return bool Whether the tag should be tracked in URL metrics.
 	 */
-	private function reduce_poster_image_size( OD_Tag_Visitor_Context $context ) {
+	private function reduce_poster_image_size( OD_Tag_Visitor_Context $context ): bool {
 		$processor = $context->processor;
 
 		// Skip empty poster attributes and data: URLs.
@@ -88,7 +88,7 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 	 * @param OD_Tag_Visitor_Context $context Tag visitor context, with the cursor currently at an embed block.
 	 * @return bool Whether the tag should be tracked in URL metrics.
 	 */
-	private function preload_poster_image( OD_Tag_Visitor_Context $context ) {
+	private function preload_poster_image( OD_Tag_Visitor_Context $context ): bool {
 		$processor = $context->processor;
 
 		// Skip empty poster attributes and data: URLs.
--- a/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
@@ -37,10 +37,16 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 
 		// TODO: If $context->url_metric_group_collection->get_element_max_intersection_ratio( $xpath ) is 0.0, then the video is not in any initial viewport and the VIDEO tag could get the preload=none attribute added.
 
-		$reduced_poster_size = $this->reduce_poster_image_size( $context );
-		$preload_poster      = $this->preload_poster_image( $context );
+		$poster = $this->get_poster( $context );
 
-		return $reduced_poster_size || $preload_poster;
+		if ( null !== $poster ) {
+			$reduced_poster_size = $this->reduce_poster_image_size( $poster, $context );
+			$preload_poster      = $this->preload_poster_image( $poster, $context );
+
+			return true;
+		}
+
+		return false;
 	}
 	/**
 	 * Gets the poster from the current VIDEO element.
@@ -65,10 +71,11 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 	 *
 	 * @since n.e.x.t
 	 *
+	 * @param non-empty-string       $poster  Poster image URL.
 	 * @param OD_Tag_Visitor_Context $context Tag visitor context, with the cursor currently at a VIDEO tag.
 	 * @return bool Whether the tag should be tracked in URL metrics.
 	 */
-	private function reduce_poster_image_size( OD_Tag_Visitor_Context $context ): bool {
+	private function reduce_poster_image_size( string $poster, OD_Tag_Visitor_Context $context ): bool {
 		$processor = $context->processor;
 
 		$xpath = $processor->get_xpath();
@@ -79,12 +86,6 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 
 		foreach ( $denormalized_elements as list( , , $element ) ) {
 			$max_element_width = max( $max_element_width, $element['boundingClientRect']['width'] ?? 0 );
-		}
-
-		$poster = $this->get_poster( $context );
-
-		if ( null === $poster ) {
-			return false;
 		}
 
 		$poster_id = attachment_url_to_postid( $poster );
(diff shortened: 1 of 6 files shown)


plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -35,6 +35,25 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+		// TODO: If $context->url_metric_group_collection->get_element_max_intersection_ratio( $xpath ) is 0.0, then the video is not in any initial viewport and the VIDEO tag could get the preload=none attribute added.
+
+		$reduced_poster_size = $this->reduce_poster_image_size( $context );
+		$preload_poster      = $this->preload_poster_image( $context );
+
+		return $reduced_poster_size || $preload_poster;
+	}
+
+	/**
+	 * Reduce poster image size by choosing one that fits the maximum video size more closely.
```
westonruter:
```suggestion
	 * Reduces poster image size by choosing one that fits the maximum video size more closely.
```

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -43,7 +62,42 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+		if ( $poster_id > 0 && $max_element_width > 0 ) {
+			$smaller_image_url = wp_get_attachment_image_url( $poster_id, array( (int) $max_element_width, 0 ) );
+			$processor->set_attribute( 'poster', $smaller_image_url );
+		}
+
+		return true;
+	}
+
+	/**
+	 * Preload poster image for the LCP <video> element.
```
westonruter:
```suggestion
	 * Preloads poster image for the LCP <video> element.
```

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -43,7 +62,42 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
 
 		$xpath = $processor->get_xpath();
 
-		// TODO: If $context->url_metric_group_collection->get_element_max_intersection_ratio( $xpath ) is 0.0, then the video is not in any initial viewport and the VIDEO tag could get the preload=none attribute added.
+		$max_element_width = 0;
+
+		$denormalized_elements = $context->url_metric_group_collection->get_all_denormalized_elements()[ $xpath ] ?? array();
+
+		foreach ( $denormalized_elements as list( , , $element ) ) {
```
westonruter: This will be made more ergonomic with #1585

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -43,7 +62,42 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+			$max_element_width = max( $max_element_width, $element['boundingClientRect']['width'] ?? 0 );
+		}
+
+		$poster_id = attachment_url_to_postid( $poster );
+
+		if ( $poster_id > 0 && $max_element_width > 0 ) {
+			$smaller_image_url = wp_get_attachment_image_url( $poster_id, array( (int) $max_element_width, 0 ) );
+			$processor->set_attribute( 'poster', $smaller_image_url );
+		}
```
westonruter: One possible consideration to make here is that if there are no URL Metrics gathered for the largest breakpoint group that in this case the poster should not be reduced in size. The reason being is that if there are only mobile visitors from whom URL Metrics have been gathered, as soon as a desktop visitor visits they could see a poster that is much lower-resolution than it should be.

Maybe this is an edge case that we shouldn't worry about right now, but it's something to keep in mind.

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -35,6 +35,25 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+
+		return $reduced_poster_size || $preload_poster;
+	}
+
+	/**
+	 * Reduce poster image size by choosing one that fits the maximum video size more closely.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param OD_Tag_Visitor_Context $context Tag visitor context, with the cursor currently at an embed block.
```
westonruter:
```suggestion
	 * @param OD_Tag_Visitor_Context $context Tag visitor context, with the cursor currently at a VIDEO tag.
```

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -43,7 +62,42 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+
+		return true;
+	}
+
+	/**
+	 * Preload poster image for the LCP <video> element.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param OD_Tag_Visitor_Context $context Tag visitor context, with the cursor currently at an embed block.
```
westonruter:
```suggestion
	 * @param OD_Tag_Visitor_Context $context Tag visitor context, with the cursor currently at a VIDEO tag.
```

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -35,6 +35,25 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
 		}
 
+		// TODO: If $context->url_metric_group_collection->get_element_max_intersection_ratio( $xpath ) is 0.0, then the video is not in any initial viewport and the VIDEO tag could get the preload=none attribute added.
+
+		$reduced_poster_size = $this->reduce_poster_image_size( $context );
+		$preload_poster      = $this->preload_poster_image( $context );
+
+		return $reduced_poster_size || $preload_poster;
+	}
+
```
westonruter, line 51: We could reduce some duplication by adding this helper method:

```suggestion
	/**
	 * Gets the poster from the current VIDEO element.
	 *
	 * @since n.e.x.t
	 *
	 * @param OD_Tag_Visitor_Context $context Tag visitor context.
	 * @return non-empty-string|null Poster or null if not defined or is a data: URL.
	 */
	private function get_poster( OD_Tag_Visitor_Context $context ): ?string {
		$poster = trim( (string) $context->processor->get_attribute( 'poster' ) );
		if ( '' === $poster || $this->is_data_url( $poster ) ) {
			return null;
		}
		return $poster;
	}

```

This would simplify the code a bit as follows:

```diff
diff --git a/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php b/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
index 772c290a..9bfe8a89 100644
--- a/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
@@ -37,10 +37,14 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 
 		// TODO: If $context->url_metric_group_collection->get_element_max_intersection_ratio( $xpath ) is 0.0, then the video is not in any initial viewport and the VIDEO tag could get the preload=none attribute added.
 
-		$reduced_poster_size = $this->reduce_poster_image_size( $context );
-		$preload_poster      = $this->preload_poster_image( $context );
+		$poster = $this->get_poster( $context );
+		if ( null !== $poster ) {
+			$this->reduce_poster_image_size( $poster, $context );
+			$this->preload_poster_image( $poster, $context );
+			return true;
+		}
 
-		return $reduced_poster_size || $preload_poster;
+		return false;
 	}
 
 	/**
@@ -48,19 +52,12 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 	 *
 	 * @since n.e.x.t
 	 *
+	 * @param non-empty-string       $poster  Poster image URL.
 	 * @param OD_Tag_Visitor_Context $context Tag visitor context, with the cursor currently at an embed block.
-	 * @return bool Whether the tag should be tracked in URL metrics.
 	 */
-	private function reduce_poster_image_size( OD_Tag_Visitor_Context $context ): bool {
+	private function reduce_poster_image_size( string $poster, OD_Tag_Visitor_Context $context ): void {
 		$processor = $context->processor;
-
-		// Skip empty poster attributes and data: URLs.
-		$poster = trim( (string) $processor->get_attribute( 'poster' ) );
-		if ( '' === $poster || $this->is_data_url( $poster ) ) {
-			return false;
-		}
-
-		$xpath = $processor->get_xpath();
+		$xpath     = $processor->get_xpath();
 
 		$max_element_width = 0;
 
@@ -76,8 +73,6 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 			$smaller_image_url = wp_get_attachment_image_url( $poster_id, array( (int) $max_element_width, 0 ) );
 			$processor->set_attribute( 'poster', $smaller_image_url );
 		}
-
-		return true;
 	}
 
 	/**
@@ -85,19 +80,12 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 	 *
 	 * @since n.e.x.t
 	 *
+	 * @param non-empty-string       $poster  Poster image URL.
 	 * @param OD_Tag_Visitor_Context $context Tag visitor context, with the cursor currently at an embed block.
-	 * @return bool Whether the tag should be tracked in URL metrics.
 	 */
-	private function preload_poster_image( OD_Tag_Visitor_Context $context ): bool {
+	private function preload_poster_image( string $poster, OD_Tag_Visitor_Context $context ): void {
 		$processor = $context->processor;
-
-		// Skip empty poster attributes and data: URLs.
-		$poster = trim( (string) $processor->get_attribute( 'poster' ) );
-		if ( '' === $poster || $this->is_data_url( $poster ) ) {
-			return false;
-		}
-
-		$xpath = $processor->get_xpath();
+		$xpath     = $processor->get_xpath();
 
 		// If this element is the LCP (for a breakpoint group), add a preload link for it.
 		foreach ( $context->url_metric_group_collection->get_groups_by_lcp_element( $xpath ) as $group ) {
@@ -120,7 +108,5 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 				$group->get_maximum_viewport_width()
 			);
 		}
-
-		return true;
 	}
 }
```

westonruter (changes requested):
Love this!! A few suggestions.

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
felixarntz: Great point, I think this is relevant to consider. I'd say we should be cautious in such a case and prioritize a "not broken" experience over better performance. Basically, unless we have data for all three viewports, we cannot reasonably make a decision to optimize.

felixarntz (commented):
@swissspidy Looks great to me, other than what @westonruter already outlined.

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
westonruter: Related to this is https://github.com/WordPress/performance/pull/1596/files#r1803569058 where I realize that lazy-loading may currently be incorrectly applied to an element on desktop if it is not visible on mobile and only URL Metrics for mobile have been collected. But if a site only ever gets mobile traffic, should the lack of desktop data be a blocker?

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -35,15 +35,86 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
-		// TODO: If $context->url_metric_group_collection->get_element_max_intersection_ratio( $xpath ) is 0.0, then the video is not in any initial viewport and the VIDEO tag could get the preload=none attribute added.
+		$max_element_width = 0;
+
+		$denormalized_elements = $context->url_metric_group_collection->get_all_denormalized_elements()[ $xpath ] ?? array();
+
+		foreach ( $denormalized_elements as list( , , $element ) ) {
+			$max_element_width = max( $max_element_width, $element['boundingClientRect']['width'] ?? 0 );
+		}
+
+		$poster = $this->get_poster( $context );
```
westonruter: What about passing the `$poster` in as an argument to `reduce_poster_image_size()` and `preload_poster_image()` as shown in https://github.com/WordPress/performance/pull/1595#discussion_r1803433939?

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
swissspidy: I think a low-res poster image on desktop if there's a lack of desktop traffic is acceptable.

But we can be more strict for now and then see how well it works. So how can I easily check whether values exist for every viewport? :-)

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
swissspidy: Ah, sorry, missed that second diff.

Doesn't make a big difference to me, I can add it.

But the early return in `__invoke` will need to change a little bit in #1596.

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
westonruter: The `OD_URL_Metric_Group_Collection::is_every_group_populated()` method will indicate whether all groups have at least one URL Metric. But for some sites all viewports may never get populated, such as if no tablet visitors show  up. See https://github.com/WordPress/performance/pull/1404. 

So should this only check if the largest group (desktop) is populated?

westonruter: I believe this would do the trick:

```php
$widest_group = array_reduce(
	iterator_to_array( $context->url_metric_group_collection ),
	static function ( $carry, OD_URL_Metric_Group $group ) {
		return ( null === $carry || $group->get_minimum_viewport_width() > $carry->get_minimum_viewport_width() ) ? $group : $carry;
	}
);
foreach ( $widest_group as $url_metric ) {
	foreach ( $url_metric->get_elements() as $element ) {
		if ( $element['xpath'] === $xpath ) {
			$max_element_width = max( $max_element_width, $element['boundingClientRect']['width'] ?? 0 );
			break; // Move on to the next URL Metric.
		}
	}
}
```

westonruter: In reality, the last item in `iterator_to_array( $context->url_metric_group_collection )` will have the widest viewport since `OD_URL_Metric_Group_Collection` sorts the breakpoints before constructing the groups.

westonruter: It would probably make sense to add a helper to `OD_URL_Metric_Group_Collection` which returns the widest group (i.e. desktop). But we can do that later.

swissspidy: OK I can give that a try tomorrow.

westonruter: I didn't want to lead you astray, so I went ahead and tested it out in 3521f9ec2312477e6813b0c4975c5e3f6b40819c. It seems to work! There are two test cases, one for when desktop URL metrics are collected and another when they are absent.

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -35,15 +35,96 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+		// If the element wasn't present in any URL Metrics gathered for desktop, then abort downsizing the poster.
+		if ( 0 === $max_element_width ) {
+			return;
+		}
+
+		$poster_id = attachment_url_to_postid( $poster );
+
+		if ( $poster_id > 0 && $max_element_width > 0 ) {
+			$smaller_image_url = wp_get_attachment_image_url( $poster_id, array( (int) $max_element_width, 0 ) );
+			$processor->set_attribute( 'poster', $smaller_image_url );
```
westonruter: While unlikely, should this check to make sure that the `$smaller_image_url` is actually a string since `wp_get_attachment_image_url()` can technically return `false` if no image size is available (e.g. `wp_get_attachment_image_src` could even be filtered to return `false`)? 

```suggestion
			if ( is_string( $smaller_image_url ) ) {
				$processor->set_attribute( 'poster', $smaller_image_url );
			}
```

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -35,15 +35,96 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+		}
+
+		// If the element wasn't present in any URL Metrics gathered for desktop, then abort downsizing the poster.
+		if ( 0 === $max_element_width ) {
+			return;
+		}
+
+		$poster_id = attachment_url_to_postid( $poster );
+
+		if ( $poster_id > 0 && $max_element_width > 0 ) {
```
westonruter: Oh, with my `if` statement added above this is now obsolete.

```suggestion
		if ( $poster_id > 0 ) {
```

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -35,15 +35,96 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+
+		if ( null !== $poster ) {
+			$this->reduce_poster_image_size( $poster, $context );
+			$this->preload_poster_image( $poster, $context );
+
+			return true;
+		}
+
+		return false;
+	}
```
westonruter, line 50:
```suggestion
	}

```

westonruter (approved):
Just a couple minor suggestions left. So I'm pre-approving!

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
swissspidy: Amazing, thanks a lot!

PR #1596, opened 2024-10-16
Lazy load videos and video posters

### LINKED ISSUE #1590: Improve autoplay for videos outside initial viewport
This is a follow-up to #1498 /  #1183

[`autoplay` takes precedence over `preload`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#preload). When we know that a video is not in the initial viewport, we could remove `autoplay` in favor of an `IntersectionObserver`-based implementation that only autoplays the video when actually visible.

Related to #1035 which involves `IntersectionObserver` to lazy-load background images.

### LINKED ISSUE #1591: Poster images for videos outside viewport should be lazy-loaded
When a `VIDEO` with a `poster` is outside the viewport, the `poster` image will be downloaded with a medium priority even though the element is not visible:

> Image

This being the case, the `poster` should ideally be lazy-loaded when the video nears the viewport using `IntersectionObserver`.

Related:

* https://github.com/WordPress/performance/issues/1590
* https://github.com/WordPress/performance/issues/1035

### LINKED ISSUE #1592: Appropriate preload value should be selected for video based on whether it is in the viewport

swissspidy:
## Summary

This builds upon #1595 / #1575 which chooses a more suitable poster image size depending on the viewport dimension.

What this does:

Disable autoplay for videos outside the initial viewport and trigger autoplay using `IntersectionObserver`

Fixes #1590

Remove `poster` attribute for videos outside the initial viewport, adding it back using `IntersectionObserver`.

Fixes #1591

Set `preload=none` for videos outside the initial viewport, adding back the original value using `IntersectionObserver`.

Fixes #1592

In my test post with 7 videos, this caused only 1 video & poster image to fully load. This massively reduces the amount of bytes loaded.

To-do:

* [ ] Cross-browser testing
* [x] Unit tests
* [x] When a video is in all initial viewports, it should probably get preload=auto especially when it is the LCP element. 

## Relevant technical choices

I looked at how Embed Optimizer does the lazy loading. It works both with and without Optimization Detective. For Image Prioritizer I am not sure it's worth adding a separate implementation that also works without OD. So I left this out for now.


plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -123,4 +128,48 @@ private function preload_poster_image( OD_Tag_Visitor_Context $context ): bool {
 
 		return true;
 	}
+
+	/**
+	 * Adjust `autoplay` and `preload` values for videos outside initial viewport.
```
westonruter:
```suggestion
	 * Adjusts `autoplay` and `preload` values for videos outside initial viewport.
```

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -123,4 +128,48 @@ private function preload_poster_image( OD_Tag_Visitor_Context $context ): bool {
...
+			return false;
+		}
+
+		$preload = $processor->get_attribute( 'preload' );
+		if ( 'none' !== $preload ) {
+			$processor->set_attribute( 'data-original-preload', null !== $preload ? $preload : 'default' );
+			$processor->set_attribute( 'preload', 'none' );
+		}
+
+		if ( true === $processor->get_attribute( 'autoplay' ) ) {
```
westonruter: In HTML `autoplay="true"` also will enable autoplay, same as `autoplay="false"`, since this is a boolean attribute. For best compatibility, this should rather be:

```suggestion
		if ( null !== $processor->get_attribute( 'autoplay' ) ) {
```

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -123,4 +128,48 @@ private function preload_poster_image( OD_Tag_Visitor_Context $context ): bool {
...
+			$processor->set_attribute( 'data-original-autoplay', true );
+			$processor->remove_attribute( 'autoplay' );
```
westonruter, line 234:
```suggestion
			$processor->remove_attribute( 'autoplay' );
			$processor->add_class( 'wp-lazy-video' );
```

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -123,4 +128,48 @@ private function preload_poster_image( OD_Tag_Visitor_Context $context ): bool {
...
+		}
+
+		$poster = trim( (string) $processor->get_attribute( 'poster' ) );
+		if ( '' !== $poster && ! $this->is_data_url( $poster ) ) {
+			$processor->set_attribute( 'data-original-poster', $poster );
+			$processor->remove_attribute( 'poster' );
```
westonruter, line 240:
```suggestion
			$processor->remove_attribute( 'poster' );
			$processor->add_class( 'wp-lazy-video' );
```

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -123,4 +128,48 @@ private function preload_poster_image( OD_Tag_Visitor_Context $context ): bool {
...
+		$intersection_ratio = $context->url_metric_group_collection->get_element_max_intersection_ratio( $xpath );
+
+		if ( $intersection_ratio > 0 ) {
+			return false;
+		}
+
+		$preload = $processor->get_attribute( 'preload' );
+		if ( 'none' !== $preload ) {
+			$processor->set_attribute( 'data-original-preload', null !== $preload ? $preload : 'default' );
+			$processor->set_attribute( 'preload', 'none' );
```
westonruter, line 228:
```suggestion
			$processor->set_attribute( 'preload', 'none' );
			$processor->add_class( 'wp-lazy-video' );
```

plugins/image-prioritizer/hooks.php
```diff
@@ -11,3 +11,55 @@
...
+					}
+				}
+			},
+			{
+				rootMargin: '100% 0% 100% 0%',
+				threshold: 0
+			}
+		);
+
+		const videos = document.querySelectorAll( 'video' );
```
westonruter: This can be made more specific to avoid querying irrelevant videos, such as when the class is added:

```suggestion
		const videos = document.querySelectorAll( 'video.wp-lazy-video' );
```

plugins/image-prioritizer/hooks.php
```diff
@@ -11,3 +11,55 @@
...
+ * Gets the script to lazy-load videos.
+ *
+ * Load a video and its poster image when it approaches the viewport using an IntersectionObserver.
+ *
+ * Handles 'autoplay' and 'preload' attributes accordingly.
+ *
+ * @since n.e.x.t
+ */
+function image_prioritizer_get_lazy_load_script(): string {
+	return <<<JS
```
westonruter: Eventually should this script and the script in `embed_optimizer_get_lazy_load_script()` be moved to external files that get inlined? This would allow them to be checked with TypeScript as well as allow them to be minified.

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
westonruter: I'm not sold on the `wp-lazy-video` class name.

westonruter (changes requested):
❤️  A few suggestions.

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -123,4 +128,48 @@ private function preload_poster_image( OD_Tag_Visitor_Context $context ): bool {
...
+	 */
+	private function lazy_load_videos( OD_Tag_Visitor_Context $context ): bool {
+		$processor = $context->processor;
+
+		$xpath = $processor->get_xpath();
+
+		$intersection_ratio = $context->url_metric_group_collection->get_element_max_intersection_ratio( $xpath );
+
+		if ( $intersection_ratio > 0 ) {
+			return false;
```
westonruter: This should not return `false` because the element still needs to be tracked in URL metrics. If not tracked, then `$context->url_metric_group_collection->get_element_max_intersection_ratio( $xpath )` would always return `null` since the element wouldn't ever be in URL Metrics.

I think this method can actually return `void` instead per https://github.com/WordPress/performance/pull/1595#discussion_r1803433939

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -123,4 +128,48 @@ private function preload_poster_image( OD_Tag_Visitor_Context $context ): bool {
...
+	 * @return bool Whether the tag should be tracked in URL metrics.
+	 */
+	private function lazy_load_videos( OD_Tag_Visitor_Context $context ): bool {
+		$processor = $context->processor;
+
+		$xpath = $processor->get_xpath();
+
+		$intersection_ratio = $context->url_metric_group_collection->get_element_max_intersection_ratio( $xpath );
+
+		if ( $intersection_ratio > 0 ) {
```
westonruter: I wonder if `$intersection_ratio > 0` is the right condition here because it could very well be that the URL Metrics are not fully populated. For example, it could be that if URL Metrics were gathered for mobile visitors in which the video is not in the viewport. For desktop visitors, the video could be in the viewport but no URL Metrics have been collected yet for it. Here the `$intersection_ratio` would be `0.0` since mobile the video is not visible on any mobile visitors but this would result in the video being lazy-loaded which would end up hurting desktop visitors for whom the video is visible in the viewport.

I realize this may be a problem for lazy-loaded images and embeds as well.

To deal with this, the lazy-loading optimization should perhaps depend on the smallest viewport group (mobile) being populated as well as the largest group (desktop). But if a site only ever gets mobile or desktop visitors, then this could indefinitely delay the optimization from being applied, which isn't good. Since this is a wholistic problem/concern/question for videos, images, and embeds, it probably isn't something needing to be addressed in this PR and we can open an issue to consider further.

Related: https://github.com/WordPress/performance/pull/1595#discussion_r1803421782

For prioritizing the loading of LCP images the situation is a bit different since we aren't dependent on the one `fetchpriority=high` attribute being on the `IMG` tag. Instead, we're able to add preload links with `fetchpriority=high` that also have media queries to ensure that the populated URL Metrics will ensure the image gets loaded properly.

Also related https://github.com/WordPress/performance/pull/1404

plugins/image-prioritizer/hooks.php
swissspidy: Yeah a follow-up PR for that would make sense

westonruter: Filed an issue to track this: https://github.com/WordPress/performance/issues/1600

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -151,4 +155,56 @@ private function preload_poster_image( string $poster, OD_Tag_Visitor_Context $c
...
+		$xpath = $processor->get_xpath();
+
+		$intersection_ratio = $context->url_metric_group_collection->get_element_max_intersection_ratio( $xpath );
+
+		// Set preload="auto" if the video is the LCP element among all viewports.
+		$common_lcp_element = $context->url_metric_group_collection->get_common_lcp_element();
+		if ( null !== $common_lcp_element && $xpath === $common_lcp_element['xpath'] ) {
+			$processor->set_attribute( 'preload', 'auto' );
+			return;
+		}
```
swissspidy: Todo: this should happen even if it's not the LCP element but still in the viewport. But then only if that's the case across all viewports.

Something like this perhaps:

```suggestion
		// Set preload="auto" if the video is in the initial viewport among all breakpoints.
		if ( $context->url_metric_group_collection->is_every_group_populated() && $intersection_ratio > 0 ) {
			$processor->set_attribute( 'preload', 'auto' );
			return;
		}
```

westonruter: The problem with `$context->url_metric_group_collection->is_every_group_populated()` is often tablets or phablets don't end up getting URL Metrics gathered for them. So I worry this would never end up returning true. Similar to https://github.com/WordPress/performance/pull/1404.

So maybe this should use `$context->url_metrics_group_collection->is_any_group_populated()` as well?

Maybe it should be _either_ of those two conditions. Either:

* it is the common LCP element among gathered URL metrics, or 
* there are URL Metrics gathered for any group and the max intersection ratio is greater than 0.

adamsilverstein (commented):
Looks great so far!

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
felixarntz: @westonruter Given that this problem is now becoming relevant in a few places and the relatively low usage of tablet devices appears to be the root of the issue, maybe we should provide an additional helper method in OD? Maybe something like `is_every_critical_group_populated()` (where only certain breakpoints would be considered critical, could be filtered)? Or, we could consider giving the breakpoints more human-friendly names like `phone`, `tablet`, `desktop` for instance, and then have a method like `is_every_group_populated( array $groups )` where you could pass those you care about as a string?

`is_every_group_populated()` has the problem we're seeing here, but `is_any_group_populated()` doesn't sound like an adequate solution either because why would we arbitrarily consider it fine to have either desktop or mobile data fully populated?

westonruter: I like that idea. In https://github.com/WordPress/performance/pull/1602 I've proposed adding a `get_first_group()` and `get_last_group()` to `OD_URL_Metric_Group_Collection` which would return the mobile and desktop URL metric groups respectively. Based on the low usage of tablets relative to desktop and mobile, I think in reality the critical groups will always be the first and last, representing the outer bounds of the viewport widths.

So with that PR, the result would be to determine whether the mobile and desktop groups are populated, this logic could be employed:

```php
(
    $context->url_metric_group_collection->get_first_group()->count() > 0 
    && 
    $context->url_metric_group_collection->get_last_group()->count() > 0
)
```

I think this could be used here instead of `$context->url_metric_group_collection->is_every_group_populated()` and `$context->url_metric_group_collection->is_any_group_populated()`.

Caveat: It could be that for a particular site the traffic is almost exclusively coming by tablet visitors. If this is the case, a site owner could indicate this by making it so that there is only one group:

```php
add_filter( 'od_breakpoint_max_widths', '__return_empty_array' );
```

In this case the first and last group are the same group.

plugins/image-prioritizer/hooks.php
westonruter: Since a contributor already landed this improvement for Embed Optimizer, I've amended this PR with the same change for Image Prioritizer in bde289fc947c99e219a3331edc1f37e66ca974e2

westonruter (approved):
@swissspidy This is looking great!

On a page which has a featured image and a video with poster outside the viewport:

## Before

image

## After

Screenshot 2024-10-17 16 24 38

plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -145,4 +160,90 @@ private function preload_poster_image( string $poster, OD_Tag_Visitor_Context $c
...
+	private function lazy_load_videos( ?string $poster, OD_Tag_Visitor_Context $context ): void {
+		$processor = $context->processor;
+
+		/*
+		 * Do not do any lazy-loading if the mobile and desktop viewport groups lack URL metrics. This is important
+		 * because if there is a VIDEO in the initial viewport on desktop but not mobile, if then there are only URL
+		 * metrics collected for mobile then the VIDEO will get lazy-loaded which is good for mobile but for desktop
+		 * it will hurt performance. So this is why it is important to have URL metrics collected for both desktop and
+		 * mobile to verify whether maximum intersectionRatio is accounting for both screen sizes.
+		 * TODO: Add this same condition to IMG lazy-loading and Embed lazy-loading.
```
westonruter, line 181: This has been implemented in the follow-up PR: https://github.com/WordPress/performance/pull/1604

swissspidy:
Amazing, thank you @westonruter! 🎉

PR #1601, opened 2024-10-17
Move embed-optimizer-lazy-load script to JS file

### LINKED ISSUE #1600: JavaScript module code inlined in PHP should be loaded from external JS files
Initially discussed in https://github.com/WordPress/performance/pull/1596#discussion_r1803449251.

There are a couple places where JS code is being inlined in PHP:

https://github.com/WordPress/performance/blob/842915b6bc738122b937da608dbe15edfdcdd016/plugins/image-prioritizer/hooks.php#L24-L65

https://github.com/WordPress/performance/blob/865fc2191610a66426558640e79521c1101c8711/plugins/embed-optimizer/hooks.php#L327-L366

This is not ideal because the JS code does not get the benefit of linting or TypeScript checks (#1589). Instead, the JS should be moved to external JS files and then loaded via `file_get_contents()`. This is what core does in `_print_emoji_detection_script()` and `print_embed_scripts()`.

This will also allow for the scripts to be minified once #1493 lands.

devansh016:
## Summary

Fixes #1600


plugins/embed-optimizer/hooks.php
```diff
@@ -325,44 +325,13 @@ function embed_optimizer_lazy_load_scripts(): void {
  * @since 0.2.0
  */
 function embed_optimizer_get_lazy_load_script(): string {
-	return <<<JS
-		const lazyEmbedsScripts = document.querySelectorAll( 'script[type="application/vnd.embed-optimizer.javascript"]' );
-		const lazyEmbedScriptsByParents = new Map();
+	$script = file_get_contents( __DIR__ . '/embed-optimizer-lazy-load.js.js' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents -- It's a local filesystem path not a remote request.
```
westonruter: I don't think this file needs to be prefixed with `embed-optimizer-`. Just `lazy-load.js` is sufficient. Also, I see you've duplicated the `.js` at the end.

```suggestion
	$script = file_get_contents( __DIR__ . '/lazy-load.js' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents -- It's a local filesystem path not a remote request.
```

plugins/embed-optimizer/embed-optimizer-lazy-load.js
westonruter, line 1: Per below, let's rename this to just `lazy-load.js`.

westonruter (changes requested):

plugins/embed-optimizer/embed-optimizer-lazy-load.js
devansh016: Thanks for the suggestion. I have made the requested changes.

adamsilverstein (commented):
Looks good, thanks!

westonruter (approved):

plugins/embed-optimizer/lazy-load.js
```diff
@@ -0,0 +1,55 @@
...
+	( entries ) => {
+		for ( const entry of entries ) {
+			if ( entry.isIntersecting ) {
+				const lazyEmbedParent = entry.target;
+				const lazyEmbedScript =
+					/** @type {HTMLScriptElement} */ lazyEmbedScriptsByParents.get(
+						lazyEmbedParent
+					);
+				const embedScript =
+					/** @type {HTMLScriptElement} */ document.createElement(
```
westonruter, line 22: I added this type comment because PhpStorm was alerting me of this:

Argument type Element is not assignable to parameter type Node | string

I'm not sure why PhpStorm thinks that `document.createElement()` can return a `string`.

PR #1602, opened 2024-10-17
Add group collection helper methods to get the first/last groups

westonruter:
In #1596 we need to exclusively look at the desktop group of URL metrics to determine the widest possible width for a given element, excluding widths from narrower viewport groups (e.g. mobile) since they could result in an image size being chosen which would look bad on desktop. This adds a `OD_URL_Metric_Group_Collection::get_last_group()` helper method to facilitate this. It also adds a `OD_URL_Metric_Group_Collection::get_first_group()` for parity.

### DIFF
--- a/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
+++ b/plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
@@ -86,13 +86,7 @@ final class Image_Prioritizer_Video_Tag_Visitor extends Image_Prioritizer_Tag_Vi
 		 * mobile viewports that an excessively-small poster would end up getting served to the first desktop visitor.
 		 */
 		$max_element_width = 0;
-		$widest_group      = array_reduce(
-			iterator_to_array( $context->url_metric_group_collection ),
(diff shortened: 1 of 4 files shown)


plugins/image-prioritizer/class-image-prioritizer-video-tag-visitor.php
```diff
@@ -86,13 +86,7 @@ private function reduce_poster_image_size( string $poster, OD_Tag_Visitor_Contex
 		 * mobile viewports that an excessively-small poster would end up getting served to the first desktop visitor.
 		 */
 		$max_element_width = 0;
-		$widest_group      = array_reduce(
-			iterator_to_array( $context->url_metric_group_collection ),
-			static function ( $carry, OD_URL_Metric_Group $group ) {
-				return ( null === $carry || $group->get_minimum_viewport_width() > $carry->get_minimum_viewport_width() ) ? $group : $carry;
-			}
-		);
```
westonruter, line 94: This iteration is now eliminated for every `video[poster]` in the document.

felixarntz (approved):
Thanks @westonruter, LGTM!

PR #1604, opened 2024-10-18
Avoid lazy-loading images and embeds unless there are URL Metrics for both mobile and desktop

westonruter:
See discussion in https://github.com/WordPress/performance/pull/1596#discussion_r1804594668

We can only safely lazy-load anything once we have collected URL metrics for both mobile and desktop. If we've only collected URL metrics for mobile and there is a lazy-loadable element which is not visible, it would be an error to lazy-load it if that element does get displayed on a larger viewport (e.g. desktop).

### DIFF
--- a/plugins/image-prioritizer/tests/test-cases/common-lcp-image-with-fully-incomplete-sample-data.php
+++ b/plugins/image-prioritizer/tests/test-cases/common-lcp-image-with-fully-incomplete-sample-data.php
@@ -4,7 +4,7 @@ return array(
 		$slug = od_get_url_metrics_slug( od_get_normalized_query_vars() );
 		$sample_size = od_get_url_metrics_breakpoint_sample_size();
 
-		// Only populate the largest viewport group.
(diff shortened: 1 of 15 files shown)


plugins/image-prioritizer/tests/test-cases/common-lcp-image-with-fully-incomplete-sample-data.php
```diff
@@ -26,6 +26,7 @@
 			);
 		}
 	},
+	// Note: loading=lazy is not removed from these images URL Metrics are only gathered for desktop so far. Both mobile and desktop URL Metrics are required to proceed with lazy-loading.
```
swissspidy:
```suggestion
	// Note: loading=lazy is not removed from these images because URL Metrics are only gathered for desktop so far. Both mobile and desktop URL Metrics are required to proceed with lazy-loading.
```

swissspidy (approved):

felixarntz (approved):

PR #1605, opened 2024-10-18
Improve formatting of `.json` and `.yml` files

### LINKED ISSUE #1603: Running npm run format-js formats more files than expected
When I run `npm run format-js` it ends up formatting more files than I expected:

```
$ npm run format-js

> format-js
> wp-scripts format

.eslintrc.js 29ms
.github/dependabot.yml 20ms
.github/workflows/codeql-analysis.yml 6ms
.github/workflows/deploy-plugins.yml 15ms
.github/workflows/js-lint.yml 4ms
.github/workflows/php-lint.yml 5ms
.github/workflows/php-test-plugins.yml 8ms
.github/workflows/pr-validation.yml 3ms
.github/workflows/props-bot.yml 5ms
.github/workflows/spell-check.yml 2ms
.wp-env.json 2ms
.wp-env.override-disabled.json 1ms
.wp-env.override-old.json 1ms
.wp-env.override.json 1ms
bin/plugin/cli.js 13ms
bin/plugin/commands/changelog.js 22ms
bin/plugin/commands/readme.js 13ms
bin/plugin/commands/since.js 6ms
bin/plugin/commands/versions.js 6ms
bin/plugin/config.js 1ms
bin/plugin/lib/logger.js 2ms
bin/plugin/lib/milestone.js 5ms
composer.json 2ms
lint-staged.config.js 3ms
package-lock.json 130ms
package.json 1ms
plugins.json 0ms
plugins/embed-optimizer/detect.js 5ms
plugins/embed-optimizer/lazy-load.js 2ms
plugins/optimization-detective/detect.js 12ms
plugins/optimization-detective/types.ts 107ms
plugins/performance-lab/.wordpress-org/blueprints/blueprint.json 1ms
scratches/after.json 0ms
scratches/after.pretty.json 1ms
scratches/query-after.js 28ms
scratches/query-before.js 26ms
test-plugins/try-jquery-callback/script.js 2ms
tools/webpack/utils.js 5ms
tsconfig.json 1ms
webpack.config.js 5ms
```

The `scratches` and `test-plugins` directories I have on my machine which are not checked-in to Git.

The specific checked-in files that were modified:

```
M	.github/dependabot.yml
M	.github/workflows/codeql-analysis.yml
M	.github/workflows/deploy-plugins.yml
M	.github/workflows/js-lint.yml
M	.github/workflows/php-lint.yml
M	.github/workflows/php-test-plugins.yml
M	.github/workflows/pr-validation.yml
M	.github/workflows/props-bot.yml
M	.github/workflows/spell-check.yml
M	.wp-env.json
M	composer.json
M	package-lock.json
M	package.json
M	plugins.json
M	plugins/performance-lab/.wordpress-org/blueprints/blueprint.json
M	tsconfig.json
```

Either we should commit these changes to YML and JSON files, or we should ignore these files from being formatted.

devansh016:
## Summary

Fixes https://github.com/WordPress/performance/issues/1603

## Relevant technical choices

The following files have been formatted by using `npm run format-js`
Note: .yml and .json files were not ignored to ensure they are properly formatted as well.

```
.eslintrc.js 22ms
.github/dependabot.yml 12ms
.github/workflows/bump-wordpress-tested-up-to.yml 7ms
.github/workflows/codeql-analysis.yml 3ms
.github/workflows/deploy-plugins.yml 8ms
.github/workflows/e2e-test.yml 3ms
.github/workflows/js-lint.yml 2ms
.github/workflows/php-lint.yml 3ms
.github/workflows/php-test-plugins.yml 5ms
.github/workflows/plugin-check.yml 4ms
.github/workflows/pr-validation.yml 1ms
.github/workflows/props-bot.yml 2ms
.github/workflows/spell-check.yml 1ms
.prettierrc.js 2ms
.wp-env.json 2ms
.wp-env.override.json 2ms
bin/plugin/cli.js 7ms
bin/plugin/commands/changelog.js 15ms
bin/plugin/commands/readme.js 7ms
bin/plugin/commands/since.js 4ms
bin/plugin/commands/versions.js 4ms
bin/plugin/config.js 1ms
bin/plugin/lib/logger.js 1ms
bin/plugin/lib/milestone.js 2ms
bin/validate-json-schema.js 3ms
codecov.yml 1ms
composer.json 1ms
lint-staged.config.js 3ms
package.json 1ms
plugins.json 0ms
plugins/auto-sizes/tests/e2e/specs/improve-calculate-sizes.spec.js 3ms
plugins/embed-optimizer/detect.js 2ms
plugins/embed-optimizer/lazy-load.js 2ms
plugins/image-prioritizer/detect.js 3ms
plugins/image-prioritizer/lazy-load-bg-image.js 2ms
plugins/image-prioritizer/lazy-load-video.js 2ms
plugins/image-prioritizer/tests/test-cases/preload-links-with-one-half-stale-group/url-metrics.json 8ms
plugins/optimization-detective/detect-loader.js 1ms
plugins/optimization-detective/detect.js 14ms
plugins/optimization-detective/tests/data/url-metrics/tablet-viewport-half-stale.json 5ms
plugins/optimization-detective/types.ts 64ms
plugins/performance-lab/.wordpress-org/blueprints/blueprint.json 1ms
plugins/performance-lab/includes/admin/plugin-activate-ajax.js 2ms
plugins/view-transitions/js/types.ts 2ms
plugins/view-transitions/js/view-transitions.js 4ms
tools/e2e/playwright.config.ts 2ms
tools/webpack/utils.js 3ms
tsconfig.json 1ms
webpack.config.js 8ms
```


westonruter:
Let's also do linting of JSON and YML files in the lint-staged config so this doesn't happen again. Also, why isn't the formatting not causing a failure on the GHA checks?

westonruter:
Will Dependabot updates undo the changes to `composer.json`, `package-lock.json`, and `package.json`?

.wp-env.json
```diff
@@ -1,24 +1,24 @@
...
-    "tests": {
-      "config": {
-        "FS_METHOD": "direct"
-      },
-      "mappings": {
-        "/wp-content/plugins/performance": "."
-      }
-    }
-  }
+	"core": null,
```
mukeshpanchal27, line 3: Per https://github.com/WordPress/performance/blob/trunk/.editorconfig#L23-L25 the json config it should use two space not tab 🤔

mukeshpanchal27 (changes requested):

.wp-env.json
devansh016: It seems that when running `npm run format-js`, the `.editorconfig` settings are not being applied, which is causing JSON files to use tabs instead of 2 spaces as specified in the `.editorconfig`. It appears the command is currently using the .prettierrc configuration instead.

thelovekesh: When both `.editorconfig` and `.prettierrc` are present, Prettier gives precedence to `.prettierrc`. You can override config for yml and json to change this behavior from the Prettier config file - https://prettier.io/docs/en/configuration.html#configuration-overrides

devansh016:
> Also, why isn't the formatting not causing a failure on the GHA checks?

`npm run lint-js` doesn't check for .yml and .json file and hence not causing a failure on the GHA checks.

thelovekesh:
I think we squash and merge this one and include an ignore commit for the merge commit to hide it in git blame. Ref: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view.

.wp-env.json
swissspidy: Seems strange that we have to apply a manual override for the default Prettier configuration used by wp-scripts, as that is the de facto standard.

Our `.editorconfig` file looks quite different from what's in [core](https://github.com/WordPress/wordpress-develop/blob/6bb92c17e1acceec7a04c57f3c5d10cb1677b2e8/.editorconfig) or [Gutenberg](https://github.com/WordPress/gutenberg/blob/3e2e434c55347584b25273006a03ce82677c3ee3/.editorconfig). They both use tabs for JSON files for example.

So I think we should rather update our `.editorconfig` rather than the Prettier config.

swissspidy:
We should exclude generated lock files from Prettier, just to avoid conflicts

thelovekesh:
> We should exclude generated lock files from Prettier, just to avoid conflicts

Yes, we should ignore any auto-generated files. @devansh016 can you please add a https://prettier.io/docs/en/ignore.html with such files?

westonruter:
See also https://github.com/WordPress/gutenberg/pull/30714 in which Gutenberg started applying formatting to JSON files.

.wp-env.json
felixarntz: Yeah, I noticed that too the other week. FWIW the `.editorconfig` we use is old, and a long time ago I believe this was common in WP land to use two spaces for those kinds of files. But since Gutenberg provides centralized tooling that dictates otherwise, I feel like it would make sense to align with that.

devansh016: I have updated the `.editorconfig` file to align with the settings used in Core and Gutenberg, where JSON files use tabs instead of 2 spaces. This should prevent conflicts with the existing Prettier configuration and ensure consistency across the project.

.editorconfig
```diff
@@ -1,39 +1,24 @@
+# This file is for unifying the coding style for different editors and IDEs
+# editorconfig.org
+
 # WordPress Coding Standards
-# https://make.wordpress.org/core/handbook/best-practices/coding-standards/
+# https://make.wordpress.org/core/handbook/coding-standards/
```
mukeshpanchal27:
```suggestion
# https://developer.wordpress.org/coding-standards/wordpress-coding-standards/
```

Instead of lets use redirected URL.

.editorconfig
```diff
@@ -1,39 +1,24 @@
...
-[*.yml]
-insert_final_newline = false
-quote_type = single
-indent_style = space
-indent_size = 2
-
-[.github/CODEOWNERS]
-indent_style = space
+[*.txt]
+end_of_line = crlf
```
mukeshpanchal27:
```suggestion
end_of_line = crlf

```

End line missing

westonruter:
> Will Dependabot updates undo the changes to `composer.json`, `package-lock.json`, and `package.json`?

I think we still need to confirm Dependabot's behavior. We wouldn't want it constantly reverting tabs back to spaces.

devansh016:
> > Will Dependabot updates undo the changes to `composer.json`, `package-lock.json`, and `package.json`?
> 
> I think we still need to confirm Dependabot's behavior. We wouldn't want it constantly reverting tabs back to spaces.

Dependabot generally respects the indentation style in `package.json` when updating the lockfile so it should not revert tabs back to spaces. [Ref](https://github.com/dependabot/dependabot-core/pull/4763)

.prettierrc.js
```diff
@@ -0,0 +1,16 @@
...
+	overrides: [
+		{
+			files: '*.yml',
+			options: {
+				useTabs: false,
+				tabWidth: 2,
+			},
+		},
+	],
+};
```
swissspidy: Why do we need this override?

.wp-env.json
swissspidy: Thanks. Can we then remove this `.prettierrc.js` override for yml files? We should be able to use the default config for those too, no?

.prettierrc.js
devansh016: @swissspidy When both `.editorconfig` and `.prettierrc` are present, Prettier gives precedence to` .prettierrc`. Further the recommended indentation for YAML files is two spaces per level, so we override it here.

swissspidy: But the default Prettier config should be correct for our needs, and the editorconfig should match the Prettier config 🤔

devansh016: @swissspidy The default Prettier configuration should suffice for our needs, and ensuring that the `.editorconfig `aligns with it eliminates the need for overrides. I have remove this overrides and now the .yml files formating looks similar to one in gutenberg project [ref](https://github.com/WordPress/gutenberg/blob/trunk/.github/dependabot.yml).

phanduynam:

.editorconfig
```diff
@@ -1,39 +1,24 @@
...
 
 [*]
 charset = utf-8
 end_of_line = lf
-indent_size = 4
-tab_width = 4
-indent_style = tab
 insert_final_newline = true
 trim_trailing_whitespace = true
+indent_style = tab
```
felixarntz, line 18: Why this change? The tab width for actual code (PHP and JS) should be 4, so I don't see why this needs to be removed here.

It doesn't apply to `.json` and `.yml`, but then we should have specific rules for these file types.

swissspidy:
Note that there are now some merge conflicts too

.editorconfig
devansh016: @felixarntz 
When both .editorconfig and .prettierrc are present, Prettier gives precedence to .prettierrc, which ensures PHP and JS files retain their intended 4-space/tab indentation.

I removed the global indent_size = 4 from .editorconfig since it was unintentionally being inherited by .json and .yml files. These files conventionally follow 2-space indentation, so I added specific overrides for them instead.

These updates follow the coding standards defined in the [WordPress Core](https://github.com/WordPress/wordpress-develop/blob/trunk/.editorconfig) and [Gutenberg](https://github.com/WordPress/gutenberg/blob/trunk/.editorconfig) repositories to stay aligned with WordPress development practices.

Please let me know if you'd like me to make any adjustments or if there's anything I might have missed. Thanks!

.prettierrc.js
westonruter, line 1: Is this any different than the `.prettierrc`?

westonruter:
The following files are now being formatted when running `npm run format-js`:

```
.eslintrc.js 19ms
.github/dependabot.yml 11ms
.github/workflows/bump-wordpress-tested-up-to.yml 3ms
.github/workflows/codeql-analysis.yml 2ms
.github/workflows/deploy-plugins.yml 5ms
.github/workflows/e2e-test.yml 1ms
.github/workflows/js-lint.yml 1ms
.github/workflows/php-lint.yml 2ms
.github/workflows/php-test-plugins.yml 4ms
.github/workflows/pr-validation.yml 1ms
.github/workflows/props-bot.yml 2ms
.github/workflows/spell-check.yml 0ms
.prettierrc.js 2ms
.wp-env.json 1ms
.wp-env.override.json 1ms
bin/plugin/cli.js 6ms
bin/plugin/commands/changelog.js 11ms
bin/plugin/commands/readme.js 4ms
bin/plugin/commands/since.js 4ms
bin/plugin/commands/versions.js 2ms
bin/plugin/config.js 0ms
bin/plugin/lib/logger.js 1ms
bin/plugin/lib/milestone.js 2ms
codecov.yml 0ms
composer.json 1ms
lint-staged.config.js 1ms
package.json 0ms
plugins.json 0ms
plugins/auto-sizes/tests/e2e/specs/improve-calculate-sizes.spec.js 2ms
plugins/embed-optimizer/detect.js 1ms
plugins/embed-optimizer/lazy-load.js 1ms
plugins/image-prioritizer/detect.js 2ms
plugins/image-prioritizer/lazy-load-bg-image.js 1ms
plugins/image-prioritizer/lazy-load-video.js 1ms
plugins/image-prioritizer/tests/test-cases/preload-links-with-one-half-stale-group/url-metrics.json 6ms
plugins/optimization-detective/detect.js 11ms
plugins/optimization-detective/tests/data/url-metrics/tablet-viewport-half-stale.json 5ms
plugins/optimization-detective/types.ts 58ms
plugins/performance-lab/.wordpress-org/blueprints/blueprint.json 0ms
plugins/performance-lab/includes/admin/plugin-activate-ajax.js 2ms
plugins/view-transitions/js/types.ts 2ms
plugins/view-transitions/js/view-transitions.js 3ms
tools/e2e/playwright.config.ts 2ms
tools/webpack/utils.js 2ms
tsconfig.json 0ms
webpack.config.js 4ms
```

westonruter (approved):
This LGTM. @swissspidy could you second?

.editorconfig
felixarntz: Following up on this, I am still not sure about it.

You mention it was removed because it made `.json` and `.yml` files use the indent of 4, but that's exactly what is currently happening: All `.json` and `.yml` files use the indent of 4. So from that perspective, I don't see why having this here would be problematic.

.editorconfig
```diff
@@ -1,39 +1,24 @@
...
 
-[*.md]
-trim_trailing_whitespace = false
+[*.{yml,yaml}]
 indent_style = space
 indent_size = 2
```
felixarntz: This is confusing to me. As far as I can tell, all the `.yml` files were now updated to use indent size 4, which looks right. But then why does this say indent size 2 here? 🤔

felixarntz (commented):
@devansh016 Thank you for the iterations.

I'm still slightly confused by some of the changes in this PR, which seem misaligned / inconsistent in itself.

.editorconfig
devansh016: You're absolutely right about the inconsistency. Looking at the actual formatting changes in this PR, all `.yml` files are indeed using an indent size of 4, which aligns with the existing codebase standards.

The `indent_size = 2 `in the `.editorconfig` is misleading since Prettier takes precedence and formats these files with 4-space indentation. I initially copied this configuration from the Gutenberg and Core repositories, but it seems now `.editorconfig` file may not be updated correctly in them.

For consistency and clarity, we should update this to indent_size = 4 to accurately reflect how the files are actually being formatted. This way, the .editorconfig and the actual file formatting are aligned, even though Prettier is the one enforcing the formatting rules.

Would you prefer I update it to indent_size = 4, or we could remove the YAML-specific configuration entirely since Prettier is handling the formatting?

westonruter: Probably add `indent_size = 4` since then the editor will automatically format it properly without requiring Prettier to run, potentially.

.prettierrc.js
devansh016: No, functionally they're identical in this case. Both .prettierrc.js and .prettierrc contain the same configuration options and produce the same formatting results.
We can remove `.prettierrc.js` here

westonruter: Alternatively, it could be added to `package.json`:

```json
"prettier": "@wordpress/prettier-config",
```

.editorconfig
devansh016: Looking back at this more carefully, I realize I may have overcomplicated the situation. The global indent_size = 4 shouldn't actually cause issues with .json and .yml files if we have proper Prettier configuration in place, since Prettier takes precedence for formatting.

But still we should keep the .editorconfig aligned with  [WordPress Core](https://github.com/WordPress/wordpress-develop/blob/trunk/.editorconfig) and [Gutenberg](https://github.com/WordPress/gutenberg/blob/trunk/.editorconfig) repositories for consistency.

felixarntz: > But still we should keep the .editorconfig aligned with [WordPress Core](https://github.com/WordPress/wordpress-develop/blob/trunk/.editorconfig) and [Gutenberg](https://github.com/WordPress/gutenberg/blob/trunk/.editorconfig) repositories for consistency.

I'm not sure about aligning `.editorconfig` just for the sake of aligning - aren't the configurations for WordPress Core and Gutenberg somewhat incompldete?
d
The WordPress Coding Standards clearly state that indent size of a tab should be 4. If this is not expressed in `.editorconfig`, doesn't it leave things up to chance? That wouldn't be right.

> Looking back at this more carefully, I realize I may have overcomplicated the situation. The global indent_size = 4 shouldn't actually cause issues with .json and .yml files if we have proper Prettier configuration in place, since Prettier takes precedence for formatting.

I don't think we should accept that just because Prettier handles it. The Coding Standards setup of the repository should be consistently configured between different tooling, it shouldn't conflict.

In other words, I think we have two options:
* Either declare `indent_size = 4` globally and override it for any file types where it isn't 4.
* Or leave this out of the global space and instead declare it for all code files (e.g. `.php`, `.js`, `.css`, `.html`, `.ts`, `.tsx`).

felixarntz (commented):
I'm still not sold on the changes in `.editorconfig` - making things more vague just to not conflict doesn't seem right to me. For any problems that the current `.editorconfig` is causing, we should fix them, not avoid them. Related: https://github.com/WordPress/performance/pull/1605#discussion_r2190399584

In general, I'd also love to get @swissspidy to sign off on the changes in this PR.

.editorconfig
```diff
@@ -1,39 +1,24 @@
...
-[*.yml]
-insert_final_newline = false
-quote_type = single
-indent_style = space
-indent_size = 2
-
-[.github/CODEOWNERS]
-indent_style = space
+[*.txt]
+end_of_line = crlf
```
swissspidy: I don't see a reason for this. WP core has historically done this for its readme.txt for Windows, but we shouldn't do this here. Let's remove it.

swissspidy:
Same here.

Happy to take another look once those changes are (partially) reverted and the merge conflicts resolved.

westonruter:
Once https://github.com/WordPress/performance/pull/2325 is merged, let's pick this up again. It's annoying to me how every time I run `npm run format-js` it modifies many files.

.editorconfig
westonruter: Removed in 981334955e3d24e896996ac130ec8421d15f3f79

westonruter:
Merge conflicts have been resolved.

.editorconfig
```diff
@@ -1,39 +1,21 @@
...
-[.*rc]
-insert_final_newline = false
-indent_style = space
-indent_size = 2
-
-[*.yml]
-insert_final_newline = false
-quote_type = single
-indent_style = space
-indent_size = 2
```
westonruter, line 36: Humm. Gutenberg's [`.editorconfig`](https://github.com/WordPress/gutenberg/blob/trunk/.editorconfig) includes:

```
[*.{yml,yaml}]
indent_style = space
indent_size = 2
```

westonruter: Same as in wordpress-develop's [`.editorconfig`](https://github.com/WordPress/wordpress-develop/blob/trunk/.editorconfig)

.editorconfig
```diff
@@ -1,39 +1,21 @@
...
 insert_final_newline = true
 trim_trailing_whitespace = true
+indent_style = tab
 
-[*.md]
-trim_trailing_whitespace = false
+[*.{yml,yaml}]
 indent_style = space
-indent_size = 2
+indent_size = 4
```
westonruter: Humm. Gutenberg's [`.editorconfig`](https://github.com/WordPress/gutenberg/blob/trunk/.editorconfig) includes:

```
[*.{yml,yaml}]
indent_style = space
indent_size = 2
```

Same as in wordpress-develop's [`.editorconfig`](https://github.com/WordPress/wordpress-develop/blob/trunk/.editorconfig)

And yet, running `npm run format-js` automatically adds 4 spaces to YAML files. 😕

devansh016: I just tested by making changes to  [.editorconfig](https://github.com/WordPress/wordpress-develop/blob/trunk/.editorconfig) in wordpress-develop's and running the format command. It is not following the changes.

devansh016: For reference  adding `.prettierrc.js` is making the format follow these rules. Not sure why `.editorconfig` rules are been followed.

```.js
/**
 * Prettier configuration.
 *
 * @see https://prettier.io/docs/en/configuration.html
 */

const wordpressConfig = require( '@wordpress/prettier-config' );

/** @type {import("prettier").Config} */
const config = {
	...wordpressConfig,
	overrides: [
		...( wordpressConfig.overrides || [] ),
		{
			files: '*.{yml,yaml}',
			options: {
				useTabs: false,
				tabWidth: 2,
			},
		},
	],
};

module.exports = config;
```

westonruter: Should we override the prettier config to force it to match `.editorconfig`?

westonruter: Gutenberg's YAML files are formatted with 4 spaces.

thelovekesh: yes you'll need to override settings in a prettier config because at the moment prettier is loading its options from `(package.json).prettier`  
```sh
$ npx prettier --find-config-path webpack.config.js              
package.json
```

that loads `@wordpress/prettier-config` which have `tabWidth: 4` for every file.
```js
const config = {
	useTabs: true,
	tabWidth: 4,
	printWidth: 80,
	singleQuote: true,
	trailingComma: 'es5',
	bracketSameLine: false,
	bracketSpacing: true,
	semi: true,
	arrowParens: 'always',
	...customOptions,
	overrides: [
		{
			files: '*.{css,sass,scss}',
			options: {
				singleQuote: false,
				...customStyleOptions,
			},
		},
	],
};
```

---
`.editorconfig` is just here to provide styles to editors and it don't have any effect on formatting, but when prettier config doesn't exists, it tries to resolve defaults from `.editorconfig`, which have no effect here since prettier config is already being resolved from `@wordpress/prettier-config`.

thelovekesh: also, i think the override changes should be proposed to `@wordpress/prettier-config` since gutenberg and core `.editorconfig` says so, which is correct.

westonruter: My head is kinda spinning here. 

Gutenberg is using 4 spaces in its YAML files, for example: https://github.com/WordPress/gutenberg/blob/trunk/.github/dependabot.yml

wordpress-develop is using 2 spaces ints in YAML files, for example: https://github.com/WordPress/wordpress-develop/blob/trunk/.github/workflows/php-compatibility.yml

Nevertheless, both Gutenberg and wordpress-develop have `.editorconfig` files with:

```
[*.{yml,yaml}]
indent_style = space
indent_size = 2
```

So Gutenberg's YAML files are _incorrectly_ formatted according to the `.editorconfig` and this appears to be due to `format-js` which is configured to use 4 spaces instead of 2.

So we need to override the Prettier config to force it to use 2 spaces for YAML files. I've committed this in 15f2f0396b1928dad5056f37ce153a91f5a57710.

.editorconfig
```diff
@@ -1,39 +1,25 @@
...
-insert_final_newline = false
-indent_style = space
-indent_size = 2
+indent_style = tab
 
 [*.yml]
-insert_final_newline = false
-quote_type = single
 indent_style = space
 indent_size = 2
```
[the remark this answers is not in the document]
westonruter: @copilot I'm not following you. The `@wordpress/prettier-config` package uses tabs for JSON and tabs are being used for the JSON files now in this repo. So why are you suggesting this?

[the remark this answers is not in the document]
westonruter: In other words, this is the intended behavior. YAML files use 2 spaces for indentation, but JSON uses tabs.

.editorconfig
thelovekesh: yes that's correct

PR #1606, opened 2024-10-18
Add option to delete the original file on upload

### LINKED ISSUE #1564: Modern Image Formats: Option to delete the original file on upload
Hi team:

To save disk space, add an option to the settings to delete the original JPEG / PNG.

The file could also be named filename.avif instead of, for example, filename-jpg.avif.

The downsize that I image could be that for images that need to be .jpg or .png for applications like og:image, you'd have to temporarily turn off the option.

AhmarZaidi:
## Summary

Fixes #1564 

This pull request introduces a new option in the Modern Image Format plugin settings to delete the original JPEG/PNG files after conversion to modern image formats like AVIF or WebP. This will help save disk space by removing the original files. Also, removed the old extension suffix (eg: `-jpg`, `-png`) from the filename.

## Relevant technical choices

- Added a new checkbox option in the settings page to enable or disable the deletion of original JPEG/PNG files after conversion. 

- Updated the `original_image` metadata when the delete option is enabled so that the original image link in _Attachment details_ popup doesn't give 404.

- Added checks to ensure that the original file is only deleted if new image files have been generated. Also taken care of the case where the original image itself is in `avif` or `webp` format.

- Update the existing test cases by removing `-jpg`/`-jpeg`/`-png` from filenames checks.

## Screenshot:
Screenshot 2024-10-18 at 6 32 53 PM

### DIFF
--- a/plugins/webp-uploads/helper.php
+++ b/plugins/webp-uploads/helper.php
@@ -411,6 +411,17 @@ function webp_uploads_is_fallback_enabled(): bool {
 }
 
 /**
+ * Checks if the `webp_uploads_delete_original` option is enabled.
+ *
+ * @since 2.2.0
+ *
+ * @return bool True if the option is enabled, false otherwise.
+ */
+function webp_uploads_is_delete_original_enabled(): bool {
+	return (bool) get_option( 'webp_uploads_delete_original' );
+}
+
+/**
  * Retrieves the image URL for a specified MIME type from the attachment metadata.
  *
  * This function attempts to locate an alternate image source URL in the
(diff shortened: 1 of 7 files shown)


adamsilverstein:
Hey @AhmarZaidi - I appreciate your effort here contributing to the plugin. However, I'm not in favor of including this feature, see my comment on the issue and feel free to disagree if you feel I missed some important aspect. I know you put work into this, so thanks for that!

banago:
@adamsilverstein the plugin creates a replica of the original image in AVIF/WebP format which renders all the savings insignificant in terms of overall storage sizes.

While it makes sense to not delete the original uploaded image, it does make sense to add an option **skip replicating the original image** in the new format and only produce the thumbnails.

PR #1612, opened 2024-10-21
Introduce AVIF header health check

### LINKED ISSUE #1263: Site Health: detect web server AVIF mime type support
## Feature Description

When the server supports AVIF processing, add detection in site health to check if AVIF images are being served with the avif type or with octet-stream. Warn users when avif isn't being used. Consider attempting to apply fix via .htaccess?

## Details

Since launching AVIF support in WordPress 6.5, some users are discovering that their AVIF images are being served with the network type "octet-stream".  The impact is relatively low since browsers know to serve these `.avif` files as AVIF images, however it is worth trying to detect and fix if possible.

## Examples

The website https://ground-cd.com/blog/avif/ uses AVIF images, checking the network panel the type shows as "octet-stream" instead of "avif".
image

On the other hand, this site serves up AVIF images correctly, with the avif type: https://aomedia.org/blog%20posts/avif-meet-the-next-level-image-file-format/
image

## Underlying issue

This issue can happen if the web server does not include the `image/avif` MIME types in its configuration. For Apache, the host may have a custom config, and configurations could (possibly) be added via `.htaccess`

Nginx added AVIF to their base config in 2021 ([PR](https://github.com/nginx/nginx/commit/67d2a9541826ecd5db97d604f23460210fd3e517)).

Apache config would probably be host or distribution dependent, so these might need updating. For Apache, I opened [this PR](https://github.com/apache/httpd/pull/451) to add it to their config. Users can ask their hosts to add the configuration and this is a very low risk change for hosts.

## Suggested fix
* Add detection for the file type served when an AVIF image is loaded and warn users when the type is not correct
* Offer/try to fix the issue automatically (?)
* Maybe no need to run test if server doesn't support AVIF
* Suggest users ask their hosts add support and link to docs

adamsilverstein:
## Summary

Add a health check to verify that AVIF images are served correctly.

Fixes https://github.com/WordPress/performance/issues/1263

## Relevant technical choices

Currently this offers miinimal suggestions for how to fix the issue.

* Consider including instructions on how to enable AVIF, see https://shortpixel.com/blog/avif-mime-type-delivery-apache-nginx/.
* Possibly update this article and link there instead: https://make.wordpress.org/hosting/2024/06/19/the-image-revolution-avif-and-webp/


swissspidy:
Great idea! Definitely re-using this for https://github.com/swissspidy/media-experiments to check the content type of WASM files.

plugins/performance-lab/includes/site-health/avif-headers/helper.php
```diff
@@ -0,0 +1,82 @@
...
+
+	$avif_headers_enabled = avif_headers_is_enabled();
+
+	if ( ! $avif_headers_enabled ) {
+		$result['status']  = 'recommended';
+		$result['label']   = __( 'Your site does not send AVIF image headers correctly', 'performance-lab' );
+		$result['actions'] = sprintf(
+			'<p>%s</p>',
+			/* translators: Accessibility text. */
+			__( 'AVIF headers can be enabled with a small configuration change by your hosting provider.', 'performance-lab' )
```
westonruter:
```suggestion
			esc_html__( 'AVIF headers can be enabled with a small configuration change by your hosting provider.', 'performance-lab' )
```

plugins/performance-lab/includes/site-health/avif-headers/helper.php
```diff
@@ -0,0 +1,82 @@
...
+	$result = array(
+		'label'       => __( 'Your site sends AVIF image headers', 'performance-lab' ),
+		'status'      => 'good',
+		'badge'       => array(
+			'label' => __( 'Performance', 'performance-lab' ),
+			'color' => 'blue',
+		),
+		'description' => sprintf(
+			'<p>%s</p>',
+			__( 'AVIF images served without the `image/avif` `content-type` header may not render properly in the browser.', 'performance-lab' )
```
westonruter:
```suggestion
			esc_html__( 'AVIF images served without the `image/avif` `content-type` header may not render properly in the browser.', 'performance-lab' )
```

plugins/performance-lab/includes/site-health/avif-headers/helper.php
```diff
@@ -0,0 +1,82 @@
...
+		return false;
+	}
+
+	$content_type = $headers['content-type'];
+
+	if ( 'image/avif' === $content_type ) {
+		return true;
+	}
+
+	return false;
```
westonruter:
```suggestion
	return ( 'image/avif' === $content_type );
```

plugins/performance-lab/includes/site-health/avif-headers/helper.php
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Callback for avif_headers test.
+ *
+ * @since n.e.x.t
+ *
+ * @return array{label: string, status: string, badge: array{label: string, color: string}, description: string, actions: string, test: string} Result.
+ */
+function avif_headers_check_avif_headers_test(): array {
+	$result = array(
+		'label'       => __( 'Your site sends AVIF image headers', 'performance-lab' ),
```
westonruter, line 22: I can't remember the specifics of escaping fields like this.

plugins/performance-lab/includes/site-health/avif-headers/helper.php
mukeshpanchal27:
```suggestion
	return ( 'image/avif' === $headers['content-type']);
```

plugins/performance-lab/includes/site-health/avif-headers/helper.php
```diff
@@ -0,0 +1,82 @@
...
+	);
+
+	$avif_headers_enabled = avif_headers_is_enabled();
+
+	if ( ! $avif_headers_enabled ) {
+		$result['status']  = 'recommended';
+		$result['label']   = __( 'Your site does not send AVIF image headers correctly', 'performance-lab' );
+		$result['actions'] = sprintf(
+			'<p>%s</p>',
+			/* translators: Accessibility text. */
```
mukeshpanchal27:
```suggestion
			/* translators: Accessibility text. */
			'<p>%s</p>',
```

mukeshpanchal27 (commented):
Thanks for the PR. Left some feedbacks.

Does it possible that Site health shows `Your site does not support AVIF` as failed but `Your site sends AVIF image headers` return headers 🤔

plugins/performance-lab/includes/site-health/avif-headers/helper.php
adamsilverstein: I will match other instances

adamsilverstein:
> Does it possible that Site health shows Your site does not support AVIF as failed but Your site sends AVIF image headers return headers 🤔

@mukeshpanchal27 yes, it is possible - AVIF server support and AVIF file header support are two very separate things.

swissspidy:
> > Does it possible that Site health shows Your site does not support AVIF as failed but Your site sends AVIF image headers return headers 🤔
> 
> @mukeshpanchal27[](https://profiles.wordpress.org/mukesh27) yes, it is possible - AVIF server support and AVIF file header support are two very separate things.

Begs the question: is it worth adding this health check if the server doesn't support AVIF files anyway?

mukeshpanchal27:
> Begs the question: is it worth adding this health check if the server doesn't support AVIF files anyway?

That's my next question. Thanks @swissspidy for asking.

adamsilverstein:
> Begs the question: is it worth adding this health check if the server doesn't support AVIF files anyway?

I thought about that and I'd say yes - one reason is you could potentially migrate a site with AVIF images to a new server where they still need to get served. Also, AVIF files could be included in the theme even if not supported by the server for uploads, eg as part of a template.

plugins/performance-lab/includes/site-health/avif-headers/helper.php
```diff
@@ -0,0 +1,78 @@
...
+		'test'        => 'is_avif_headers_enabled',
+	);
+
+	$avif_headers_enabled = avif_headers_is_enabled();
+
+	if ( ! $avif_headers_enabled ) {
+		$result['status']  = 'recommended';
+		$result['label']   = __( 'Your site does not send AVIF image headers correctly', 'performance-lab' );
+		$result['actions'] = sprintf(
+			/* translators: Accessibility text. */
```
westonruter: No need for translator comment since the string is not translated.

```suggestion
```

plugins/performance-lab/includes/site-health/avif-headers/helper.php
```diff
@@ -0,0 +1,78 @@
...
+	$result = array(
+		'label'       => __( 'Your site sends AVIF image headers', 'performance-lab' ),
+		'status'      => 'good',
+		'badge'       => array(
+			'label' => __( 'Performance', 'performance-lab' ),
+			'color' => 'blue',
+		),
+		'description' => sprintf(
+			'<p>%s</p>',
+			esc_html__( 'AVIF images served without the `image/avif` `content-type` header may not render properly in the browser.', 'performance-lab' )
```
westonruter: Since backticks aren't turned in to `code`, maybe they can just be removed?

```suggestion
			esc_html__( 'AVIF images served without the Content-Type of image/avif may not render properly in the browser.', 'performance-lab' )
```

swissspidy: This should use `sprintf` for values such as `image/avif` and `Content-Type` (if referring to the header specifically), and have a translator comment.

adamsilverstein: > This should use `sprintf` for values such as `image/avif` and `Content-Type` (if referring to the header specifically), and have a translator comment.

Ah, good point, I will rework it that way.

adamsilverstein: Took a pass at improving this in [`84292b0` (#1612)](https://github.com/WordPress/performance/pull/1612/commits/84292b0a3b2073ed7d686ab978f5d2836c7432ee)

plugins/performance-lab/includes/site-health/avif-headers/helper.php
adamsilverstein: Escaping for labels is only included in `plugins/performance-lab/tests/data/class-site-health-mock-responses.php`, maybe we should remove it there.

plugins/performance-lab/includes/site-health/avif-headers/helper.php
```diff
@@ -0,0 +1,85 @@
...
+			'<p>%s</p>',
+			esc_html__(
+				/* translators: 1: image/avif, 2: content-type */
+				sprintf(
+					'Images with the mime type %1$s served without the correct %2$s header may not render properly.',
+					'image/avif',
+					'content-type'
+				),
+				'performance-lab'
+			)
```
westonruter: I believe this has to be rearranged a bit:

```suggestion
			esc_html(
				sprintf(
					/* translators: 1: image/avif, 2: content-type */
					__( 'Images with the mime type %1$s served without the correct %2$s header may not render properly.', 'performance-lab' ),
					'image/avif',
					'content-type'
				)
			)
```

westonruter (approved):

mukeshpanchal27 (approved):

Issue #1511, opened 2024-08-29
Incorporate layout constraints from ancestor blocks into the sizes calculations


joemcgill:
As outlined in #760, the initial work for accurately setting the sizes attribute for image and cover blocks has been released through the [Enhanced Responsive Images plugin](https://wordpress.org/plugins/auto-sizes/). While the current implementation effectively handles individual image blocks, it does not account for images constrained by other ancestor blocks. This issue aims to address this issue by focusing on multiple block types, including column, group, row, and stack blocks (alignments only). Enhancements for grid blocks are also being considered (though they are of lower priority due to their complexity and recent introduction).

The primary goal is to improve the sizes attribute across these block types to ensure images are displayed with the correct sizes for optimal rendering across different screen sizes.

joemcgill:
@mukeshpanchal27 and I have been exploring this issue over the last several weeks and I wanted to give a quick progress update.


## Technical discovery

In order to incorporate layout information from column, group, row, and stack blocks that are ancestors of images, we considered several different approaches.

### Use `wp_calculate_image_sizes` directly

Modifying the `wp_calculate_image_sizes` filter seems like the most straightforward was to improve the calculation of the sizes attribute. However, one of the key challenges is that this filter does not receive any layout information from the block context, which is crucial for accurately determining the appropriate sizes value for images. Instead, that function is only aware of the `img` markup itself, not even the `figure` that wraps the `img` and provides its own layout alignment.

We created [an experimental plugin](https://gist.github.com/joemcgill/77a37e862d49bf47db9c06c8265338e0) to demonstrate a proof of concept for calculating the sizes value during block rendering, when more information is available. However, this approach is a bit fragile and will result in a DB query for each image.

### Hooking into the block rendering process

In this approach, layout constraints would be passed from ancestor blocks to images using block rendering filter `render_block_data`. Inspired by how the [core gallery block](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/blocks/gallery.php#L21-L35) handles this process.

While this approach could work, we quickly found that passing block layout information down to `innerBlocks` was not performant.

### Passing layout information as block context

In this approach, layout constraints would be passed from ancestor blocks to images using [block context](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-context/).

[...]

PR #1625, opened 2024-10-23
Accurate sizes: Calculate image sizes during block rendering

mukeshpanchal27:
## Summary

These PR updates the existing code to use the new block context approach that we discussed in https://github.com/WordPress/performance/issues/1511#issuecomment-2405794554


plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,147 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
+		$context = current_filter();
+	}
 
-	// Only update the markup if an image is found.
-	if ( $has_image ) {
-		$processor->set_attribute( 'data-needs-sizes-update', true );
-		if ( isset( $parsed_block['attrs']['align'] ) ) {
-			$processor->set_attribute( 'data-align', $parsed_block['attrs']['align'] );
-		}
+	if ( false === preg_match_all( '/<img\s[^>]+>/', $content, $matches, PREG_SET_ORDER ) ) {
```
westonruter: Wouldn't this be better using HTML Tag Processor?

mukeshpanchal27: Done in https://github.com/WordPress/performance/pull/1625/commits/b87adcfa0481c964ea61ab0d5053a2ada6957369

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,142 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
-	// Only update the markup if an image is found.
-	if ( $has_image ) {
-		$processor->set_attribute( 'data-needs-sizes-update', true );
-		if ( isset( $parsed_block['attrs']['align'] ) ) {
-			$processor->set_attribute( 'data-align', $parsed_block['attrs']['align'] );
+	$images = array();
+	while ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
+		$class = $processor->get_attribute( 'class' );
+
+		// Only apply the dominant color to images that have a src attribute.
```
westonruter: Looks like this comment needs to be updated to reference `class`. Maybe it could be removed entirely since the purpose is clear.

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,142 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
+ * @param WP_Block                                                 $block        Block instance.
  * @return string The updated block content.
  */
-function auto_sizes_improve_image_sizes_attributes( string $content ): string {
+function auto_sizes_filter_image_tag( string $content, array $parsed_block, WP_Block $block ): string {
 	$processor = new WP_HTML_Tag_Processor( $content );
-	if ( ! $processor->next_tag( array( 'tag_name' => 'img' ) ) ) {
-		return $content;
-	}
+	$has_image = $processor->next_tag( array( 'tag_name' => 'img' ) );
```
westonruter: For consistency with above.

```suggestion
	$has_image = $processor->next_tag( array( 'tag_name' => 'IMG' ) );
```

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,142 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
-	};
+	// Only update the markup if an image is found.
+	if ( $has_image ) {
+
+		/**
+		 * Callback for calculating image sizes attribute value for an image block.
+		 *
+		 * This is a workaround to use block context data when calculating the img sizes attribute.
+		 *
+		 * @since n.e.x.t
```
westonruter: Probably doesn't need a `@since` tag since this is a closure and isn't part of any public interface.

westonruter: Probably doesn't need a `@since` tag since this is a closure and isn't part of any public interface.

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,139 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
- * @param array{ attrs?: array{ align?: string, width?: string } } $parsed_block The parsed block.
- * @return string The updated block content.
+ * @param string $content The HTML content.
+ * @param string $context Optional. Additional context to pass to the filters.
+ *                        Defaults to `current_filter()` when not set.
+ * @return string The HTML content.
  */
-function auto_sizes_filter_image_tag( string $content, array $parsed_block ): string {
+function auto_sizes_prime_attachment_caches( string $content, string $context = null ): string {
+	if ( null === $context ) {
```
westonruter: To guard against an error in case someone filters `the_content` to return a non-string value.

```suggestion
function auto_sizes_prime_attachment_caches( $content, string $context = null ): string {
	if ( ! is_string( $content ) ) {
		return '';
	}
	if ( null === $context ) {
```

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,139 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
  *
- * @since 1.1.0
+ * @since n.e.x.t
  *
- * @param string                                                   $content      The block content about to be rendered.
- * @param array{ attrs?: array{ align?: string, width?: string } } $parsed_block The parsed block.
- * @return string The updated block content.
+ * @param string $content The HTML content.
+ * @param string $context Optional. Additional context to pass to the filters.
+ *                        Defaults to `current_filter()` when not set.
```
westonruter:
```suggestion
 * @param string|mixed $content The HTML content.
 * @param string       $context Optional. Additional context to pass to the filters.
 *                              Defaults to `current_filter()` when not set.
```

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,139 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
  * @since 1.1.0
  *
- * @param string $content The block content about to be rendered.
+ * @param string                                                   $content      The block content about to be rendered.
+ * @param array{ attrs?: array{ align?: string, width?: string } } $parsed_block The parsed block.
+ * @param WP_Block                                                 $block        Block instance.
  * @return string The updated block content.
  */
-function auto_sizes_improve_image_sizes_attributes( string $content ): string {
+function auto_sizes_filter_image_tag( string $content, array $parsed_block, WP_Block $block ): string {
```
westonruter:
```suggestion
function auto_sizes_filter_image_tag( $content, array $parsed_block, WP_Block $block ): string {
	if ( ! is_string( $content ) ) {
		return '';
	}
```

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,139 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
-add_filter( 'render_block_core/cover', 'auto_sizes_filter_image_tag', 10, 2 );
+add_filter( 'the_content', 'auto_sizes_prime_attachment_caches', 6 );
 
 /**
  * Filter the sizes attribute for images to improve the default calculation.
  *
  * @since 1.1.0
  *
- * @param string $content The block content about to be rendered.
+ * @param string                                                   $content      The block content about to be rendered.
```
westonruter:
```suggestion
 * @param string|mixed                                             $content      The block content about to be rendered.
```

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,139 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
+		 * @param string $sizes The image sizes attribute value.
+		 * @param string $size  The image size data.
+		 */
+		$filter = static function ( $sizes, $size ) use ( $block ) {
+			$id        = $block->attributes['id'] ?? 0;
+			$alignment = $block->attributes['align'] ?? '';
+			$width     = $block->attributes['width'] ?? '';
+
+			// Hypothetical function to calculate better sizes.
+			$sizes = auto_sizes_calculate_better_sizes( $id, $size, $alignment, $width );
```
westonruter: Perhaps overkill, but we don't know for sure what types that `$block->attributes` contains.

```suggestion
			$sizes = auto_sizes_calculate_better_sizes( (int) $id, (string) $size, (string), $alignment, (string) $width );
```

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,145 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
+		 * @param string $sizes The image sizes attribute value.
+		 * @param string $size  The image size data.
+		 */
+		$filter = static function ( $sizes, $size ) use ( $block ) {
+			$id        = $block->attributes['id'] ?? 0;
+			$alignment = $block->attributes['align'] ?? '';
+			$width     = $block->attributes['width'] ?? '';
+
+			// Hypothetical function to calculate better sizes.
+			$sizes = auto_sizes_calculate_better_sizes( (int) $id, (string) $size, (string), $alignment, (string) $width );
```
mukeshpanchal27:
```suggestion
			$sizes = auto_sizes_calculate_better_sizes( (int) $id, (string) $size, (string) $alignment, (string) $width );
```

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,145 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
+ * Primes attachment into the cache with a single database query.
  *
- * @since 1.1.0
+ * @since n.e.x.t
  *
- * @param string                                                   $content      The block content about to be rendered.
- * @param array{ attrs?: array{ align?: string, width?: string } } $parsed_block The parsed block.
- * @return string The updated block content.
+ * @param string|mixed $content The HTML content.
+ * @param string       $context Optional. Additional context to pass to the filters.
```
westonruter:
```suggestion
 * @param string|null  $context Optional. Additional context to pass to the filters.
```

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,145 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
+ *                              Defaults to `current_filter()` when not set.
+ * @return string The HTML content.
  */
-function auto_sizes_filter_image_tag( string $content, array $parsed_block ): string {
+function auto_sizes_prime_attachment_caches( $content, string $context = null ): string {
```
westonruter:
```suggestion
function auto_sizes_prime_attachment_caches( $content, ?string $context = null ): string {
```

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,145 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
+	if ( ! is_string( $content ) ) {
+		return '';
+	}
+	if ( null === $context ) {
+		$context = current_filter();
+	}
```
westonruter: Perhaps my above comments should be adjusted to delete the `$context` param altogether because actually the variable isn't being used at all. So should it just be removed?

```suggestion
	if ( null === $context ) {
		$context = current_filter();
	}
```

plugins/auto-sizes/hooks.php
westonruter: Or per below:

```suggestion
function auto_sizes_prime_attachment_caches( $content ): string {
```

plugins/auto-sizes/hooks.php
westonruter: Or per below:

```suggestion
```

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,145 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
+		$class = $processor->get_attribute( 'class' );
+
+		if ( ! is_string( $class ) ) {
+			continue;
 		}
 
-		// Resize image width.
-		if ( isset( $parsed_block['attrs']['width'] ) ) {
-			$processor->set_attribute( 'data-resize-width', $parsed_block['attrs']['width'] );
+		if ( preg_match( '/wp-image-([0-9]+)/i', $class, $class_id ) === 1 ) {
```
westonruter: This should ensure that integers starting at 1 are only considered and that the class `wp-image-123` class is only considered, not `my-wp-image-123`. Granted, this is very unlikely. Also, there doesn't seem to be a need to do a case-insensitive check.

```suggestion
		if ( preg_match( '/(?:^|\s)wp-image-([1-9][0-9]*)(?:\s|$)/', $class, $class_id ) === 1 ) {
```

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,145 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
+			$attachment_id = absint( $class_id[1] );
```
westonruter: Since a negative will never be encountered anyway, seems we can just cast this here:

```suggestion
			$attachment_id = (int) $class_id[1];
```

plugins/auto-sizes/hooks.php
westonruter: Also, as it stands right now, the 2nd parameter is never passed in anyway:

```php
add_filter( 'the_content', 'auto_sizes_prime_attachment_caches', 6 );
```

mukeshpanchal27:
We are tracking the unit test issue here #1634

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,140 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
+		 * images to avoid making individual database calls.
+		 */
+		_prime_post_caches( $attachment_ids, false, true );
 	}
+
 	return $content;
 }
-add_filter( 'render_block_core/image', 'auto_sizes_filter_image_tag', 10, 2 );
-add_filter( 'render_block_core/cover', 'auto_sizes_filter_image_tag', 10, 2 );
+add_filter( 'the_content', 'auto_sizes_prime_attachment_caches', 6 );
```
westonruter: Curious as to why priority 6 is used?

westonruter (approved):
LGTM, although @joemcgill would know better what to look for in what this is seeking to achieve.

plugins/auto-sizes/hooks.php
mukeshpanchal27: Priority 6 is used to ensure the new filter runs right after the `wp-image-$attachment_id` class is added to the `img` tag at priority 5, allowing modifications that rely on this class being in place.

westonruter: Cool. Probably good to add a comment to explain why that number was chosen.

mukeshpanchal27:
Once #1663 get merged the unit tests issue will be resolved.

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,145 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
 	return $content;
 }
-add_filter( 'render_block_core/image', 'auto_sizes_filter_image_tag', 10, 2 );
-add_filter( 'render_block_core/cover', 'auto_sizes_filter_image_tag', 10, 2 );
+
+/*
+ * The priority 6 is used to ensure the new filter runs right after the "wp-image-$attachment_id" class
+ * is added to the img tag at priority 5, allowing modifications that rely on this class being in place.
+ */
+add_filter( 'the_content', 'auto_sizes_prime_attachment_caches', 6 );
```
joemcgill: I don't believe the added comment is accurate. The `wp-image-$attachment_id` classname is not added on a filter, but is instead added to stored `img` markup saved to the database when an image block is authored.

Currently, WP adds `srcset` and `sizes` on `the_content` filter after the content is rendered. As part of that process, it primes the attachment data for all images in the content being filtered before looping through all of the images on the page and adding calculating `srcset` and `sizes` for each image. All of this logic happens in [wp_filter_content_tags()](https://github.com/WordPress/wordpress-develop/blob/61b9859728b31130274b31bde9c6d73ee915fa03/src/wp-includes/media.php#L1861). The cache priming is necessary to avoid making individual DB calls for each image being modified.

The approach explored in this PR moves the logic for adding the `sizes` value from post processing on `the_content`, to the block rendering process (i.e., by modifying the `sizes` value on the `render_block_core/image` filter). The reason the priming logic needs to run on an earlier priority is because it must run prior to the `do_blocks` process that [runs on 'the_content' filter at priority 9](https://github.com/WordPress/wordpress-develop/blob/61b9859728b31130274b31bde9c6d73ee915fa03/src/wp-includes/default-filters.php#L195).

Registering this hook on priority 8 still works as expected.

```suggestion
// This must run before 'do_blocks', which runs at priority 9.
add_filter( 'the_content', 'auto_sizes_prime_attachment_caches', 8 );
```

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,145 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
 	$processor = new WP_HTML_Tag_Processor( $content );
-	$has_image = $processor->next_tag( array( 'tag_name' => 'img' ) );
 
-	// Only update the markup if an image is found.
-	if ( $has_image ) {
-		$processor->set_attribute( 'data-needs-sizes-update', true );
-		if ( isset( $parsed_block['attrs']['align'] ) ) {
-			$processor->set_attribute( 'data-align', $parsed_block['attrs']['align'] );
+	$images = array();
+	while ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
```
joemcgill, line 157: This seems like a good use of the `WP_HTML_Tag_Processor`, and is fine for a first pass at this functionality, but it would be good to do some profiling of this method against the `preg_match_all` approach that [is currently in use in Core](https://github.com/WordPress/wordpress-develop/blob/61b9859728b31130274b31bde9c6d73ee915fa03/src/wp-includes/media.php#L1868-L1870).

If we eventually want to land this code in WP, we'll want to move this priming logic from `wp_filter_content_tags` rather than duplicate it in two places, at which time we should be mindful of any performance differences between the current logic and this slightly updated version that makes use of the `WP_HTML_Tag_Processor`.

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,145 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
+		 *
+		 * @param string $sizes The image sizes attribute value.
+		 * @param string $size  The image size data.
+		 */
+		$filter = static function ( $sizes, $size ) use ( $block ) {
+			$id        = $block->attributes['id'] ?? 0;
+			$alignment = $block->attributes['align'] ?? '';
+			$width     = $block->attributes['width'] ?? '';
+
+			// Hypothetical function to calculate better sizes.
```
joemcgill: This function is not hypothetical, you've defined it below 😄 

```suggestion
```

plugins/auto-sizes/hooks.php
```diff
@@ -139,79 +139,145 @@ function auto_sizes_get_width( string $layout_width, int $image_width ): string
...
-		return $content;
-	}
+	return $content;
+}
+add_filter( 'render_block_core/image', 'auto_sizes_filter_image_tag', 10, 3 );
+add_filter( 'render_block_core/cover', 'auto_sizes_filter_image_tag', 10, 3 );
 
-	$align = $processor->get_attribute( 'data-align' );
+/**
+ * Hypothetical function to calculate better sizes.
```
joemcgill:
```suggestion
 * Modifies the sizes attribute of an image based on layout context.
```

plugins/auto-sizes/tests/test-improve-sizes.php
```diff
@@ -31,12 +31,13 @@ public function set_up(): void {
 
 		// Disable auto sizes.
 		remove_filter( 'wp_content_img_tag', 'auto_sizes_update_content_img_tag' );
+
+		// Disable lazy loading attribute.
+		add_filter( 'wp_img_tag_add_loading_attr', '__return_false' );
```
joemcgill, line 36: Why does this filter need to be disabled for these tests?

joemcgill (changes requested):
After chatting with @mukeshpanchal27 about this PR, I now better understand that the purpose of this change is to replace the current implementation for more accurate `sizes` from the initial approach, which filters core's automatic sizes calculation in `wp_calculate_image_sizes`, to a new process that adds `sizes` attributes to blocks during the block rendering process, so that we can make use of block information that is available during the block rendering process that is not currently available to the `wp_calculate_image_sizes` filter as it is currently implemented in Core.

A small nitpick for a future changelog — the title of this PR would more accurately be: `Accurate sizes: calculate image sizes during block rendering`

I've left some additional feedback inline. Once this is updated, I think we can merge this and use this as a starting point for incorporating other block information (e.g., block context) into the calculation logic.

plugins/auto-sizes/hooks.php
westonruter: > This seems like a good use of the `WP_HTML_Tag_Processor`, and is fine for a first pass at this functionality, but it would be good to do some profiling of this method against the `preg_match_all` approach that [is currently in use in Core](https://github.com/WordPress/wordpress-develop/blob/61b9859728b31130274b31bde9c6d73ee915fa03/src/wp-includes/media.php#L1868-L1870).

Haven't we already debated this at length, with the conclusion that using `WP_HTML_Tag_Processor` is preferred for correctness and that the practical performance impact is inconsequential?

mukeshpanchal27: In previous discussion in #1471 we preferred to use `WP_HTML_Tag_Processor`.

plugins/auto-sizes/tests/test-improve-sizes.php
mukeshpanchal27: Due to the `auto` value being added to the `sizes` attribute through WordPress Core, if we cannot disable the filter, I will need to update the majority of the tests.  

For example, the current assertion:  
```php
$this->assertStringContainsString( 'sizes="100vw" ', $result );
```  
would need to be replaced with:  
```php
$this->assertStringContainsString( 'sizes="auto, 100vw" ', $result );
```

plugins/auto-sizes/hooks.php
joemcgill: > Haven't we already debated this at length, with the conclusion that using WP_HTML_Tag_Processor is preferred for correctness and that the practical performance impact is inconsequential?

I keep seeing this same question asked in almost every PR where `WP_HTML_Tag_Processor` is used in place of other approaches, so if a consensus conclusion was reached, I was not aware of it. I'm more than happy for this to be the decision though.

joemcgill (approved):
Thanks for the updates. Looks good to me.

PR #1635, opened 2024-11-06
Fix bug that would prevent uploaded images from being converted to the intended output format when having fallback formats enabled

### LINKED ISSUE #1634: Unit tests for Modern Image Formats are failing on WP trunk
There are currently 10 unit tests for Modern Image Formats which are failing on `trunk`:

<details><summary>There were 10 failures:</summary>

```
1) Test_WebP_Uploads_Image_Edit::test_it_should_restore_the_sources_array_from_the_backup_when_an_image_is_edited
Failed asserting that an image doesn't have the source with the "image/jpeg" mime type.

/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/data/class-testcase.php:48
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-image-edit.php:80

2) Test_WebP_Uploads_Load::test_it_should_not_create_the_original_mime_type_for_jpeg_images with data set "webp" ('webp')
Failed asserting that an image doesn't have the source with the "image/jpeg" mime type.

/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/data/class-testcase.php:48
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:58

3) Test_WebP_Uploads_Load::test_it_should_not_create_the_original_mime_type_for_jpeg_images with data set "avif" ('avif')
Failed asserting that an image doesn't have the source with the "image/jpeg" mime type.

/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/data/class-testcase.php:48
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:58

4) Test_WebP_Uploads_Load::test_it_should_create_a_webp_version_with_all_the_required_properties
Failed asserting that an image doesn't have the source with the "image/jpeg" mime type.

/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/data/class-testcase.php:48
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:266

5) Test_WebP_Uploads_Load::test_it_should_create_the_full_size_images_when_no_size_is_available
Failed asserting that an image doesn't have the source with the "image/jpeg" mime type.

/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/data/class-testcase.php:48
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:288

6) Test_WebP_Uploads_Load::test_it_should_remove_the_generated_webp_images_when_the_attachment_is_deleted
Failed asserting that file "/var/www/html/wp-content/uploads/2024/11/leaves-61-1080x720-webp.webp" does not exist.

/var/www/html/wp-content/plugins/performance/vendor/yoast/phpunit-polyfills/src/Polyfills/AssertionRenames.php:113
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:347

7) Test_WebP_Uploads_Load::test_it_should_remove_the_attached_webp_version_if_the_attachment_is_force_deleted
Failed asserting that file "/var/www/html/wp-content/uploads/2024/11/leaves-61-1080x720-webp.webp" does not exist.

/var/www/html/wp-content/plugins/performance/vendor/yoast/phpunit-polyfills/src/Polyfills/AssertionRenames.php:113
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:372

8) Test_WebP_Uploads_Load::test_it_should_remove_full_size_images_when_no_size_image_exists
Failed asserting that file "/var/www/html/wp-content/uploads/2024/11/leaves-61-1080x720-webp.webp" does not exist.

/var/www/html/wp-content/plugins/performance/vendor/yoast/phpunit-polyfills/src/Polyfills/AssertionRenames.php:113
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:395

9) Test_WebP_Uploads_Load::test_it_should_allow_the_upload_of_a_webp_image_if_at_least_one_editor_supports_the_format with data set "webp" ('webp')
Failed asserting that an image doesn't have the source with the "image/jpeg" mime type.

/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/data/class-testcase.php:48
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:709

10) Test_WebP_Uploads_Load::test_it_should_allow_the_upload_of_a_webp_image_if_at_least_one_editor_supports_the_format with data set "avif" ('avif')
Failed asserting that an image doesn't have the source with the "image/jpeg" mime type.

/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/data/class-testcase.php:48
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:709
```

</details> 

These tests aren't failing on `6.7-branch`, so something has been committed to `trunk` (6.8-alpha) that is causing this failure.

mukeshpanchal27:
## Summary

Fixes #1634

### DIFF
--- a/plugins/webp-uploads/hooks.php
+++ b/plugins/webp-uploads/hooks.php
@@ -52,18 +52,27 @@ if ( ! defined( 'ABSPATH' ) ) {
  * } An array with the updated structure for the metadata before is stored in the database.
  */
 function webp_uploads_create_sources_property( array $metadata, int $attachment_id ): array {
-	// This should take place only on the JPEG image.
-	$valid_mime_transforms = webp_uploads_get_upload_image_mime_transforms();
-
-	// Not a supported mime type to create the sources property.
-	$mime_type = get_post_mime_type( $attachment_id );
-	if ( ! is_string( $mime_type ) || ! isset( $valid_mime_transforms[ $mime_type ] ) ) {
-		return $metadata;
-	}
-
 	$file = get_attached_file( $attachment_id, true );
 	// File does not exist.
 	if ( false === $file || ! file_exists( $file ) ) {
+		return $metadata;
+	}
+
+	/*
+	 * We need to get the MIME type ideally from the file, as WordPress Core may have already altered it.
+	 * The post MIME type is typically not updated during that process.
+	 */
+	$filetype = wp_check_filetype( $file );
+	if ( isset( $filetype['type'] ) ) {
+		$mime_type = $filetype['type'];
+	} else {
+		$mime_type = get_post_mime_type( $attachment_id );
+	}
+
+	$valid_mime_transforms = webp_uploads_get_upload_image_mime_transforms();
+
+	// Not a supported mime type to create the sources property.
+	if ( ! is_string( $mime_type ) || ! isset( $valid_mime_transforms[ $mime_type ] ) ) {
 		return $metadata;
 	}
 
--- a/plugins/webp-uploads/hooks.php
+++ b/plugins/webp-uploads/hooks.php
@@ -62,17 +62,16 @@ function webp_uploads_create_sources_property( array $metadata, int $attachment_
 	 * We need to get the MIME type ideally from the file, as WordPress Core may have already altered it.
 	 * The post MIME type is typically not updated during that process.
 	 */
-	$filetype = wp_check_filetype( $file );
-	if ( isset( $filetype['type'] ) ) {
-		$mime_type = $filetype['type'];
-	} else {
-		$mime_type = get_post_mime_type( $attachment_id );
+	$filetype  = wp_check_filetype( $file );
+	$mime_type = $filetype['type'] ?? get_post_mime_type( $attachment_id );
+	if ( ! is_string( $mime_type ) ) {
+		return $metadata;
 	}
 
 	$valid_mime_transforms = webp_uploads_get_upload_image_mime_transforms();
 
 	// Not a supported mime type to create the sources property.
-	if ( ! is_string( $mime_type ) || ! isset( $valid_mime_transforms[ $mime_type ] ) ) {
+	if ( ! isset( $valid_mime_transforms[ $mime_type ] ) ) {
 		return $metadata;
 	}
 
--- a/plugins/webp-uploads/tests/test-load.php
+++ b/plugins/webp-uploads/tests/test-load.php
@@ -117,42 +117,6 @@ class Test_WebP_Uploads_Load extends TestCase {
 	}
 
 	/**
-	 * Create JPEG and output type for JPEG images, if opted in.
-	 *
-	 * @dataProvider data_provider_supported_image_types
-	 */
-	public function test_it_should_create_jpeg_and_webp_for_jpeg_images_if_opted_in( string $image_type ): void {
-		$mime_type = 'image/' . $image_type;
-		if ( ! webp_uploads_mime_type_supported( $mime_type ) ) {
-			$this->markTestSkipped( "Mime type $mime_type is not supported." );
-		}
-		$this->set_image_output_type( $image_type );
-
-		update_option( 'perflab_generate_webp_and_jpeg', true );
-		$attachment_id = self::factory()->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/data/images/leaves.jpg' );
-
-		// There should be JPEG and mime_type sources for the full image.
-		$this->assertImageHasSource( $attachment_id, 'image/jpeg' );
-		$this->assertImageHasSource( $attachment_id, $mime_type );
-
-		$metadata = wp_get_attachment_metadata( $attachment_id );
-
-		// The full image should be a JPEG.
-		$this->assertArrayHasKey( 'file', $metadata );
-		$this->assertStringEndsWith( $metadata['sources']['image/jpeg']['file'], $metadata['file'] );
-		$this->assertStringEndsWith( $metadata['sources']['image/jpeg']['file'], get_attached_file( $attachment_id ) );
-
-		// The post MIME type should be JPEG.
-		$this->assertSame( 'image/jpeg', get_post_mime_type( $attachment_id ) );
-
-		// There should be JPEG and WebP sources for all sizes.
-		foreach ( array_keys( $metadata['sizes'] ) as $size_name ) {
-			$this->assertImageHasSizeSource( $attachment_id, $size_name, 'image/jpeg' );
-			$this->assertImageHasSizeSource( $attachment_id, $size_name, $mime_type );
-		}
-	}
-
-	/**
 	 * Create JPEG and output format for JPEG images, if perflab_generate_webp_and_jpeg option set.
 	 *
 	 * @dataProvider data_provider_supported_image_types
--- a/plugins/webp-uploads/tests/test-load.php
+++ b/plugins/webp-uploads/tests/test-load.php
@@ -45,7 +45,7 @@ class Test_WebP_Uploads_Load extends TestCase {
 	 *
 	 * @dataProvider data_provider_supported_image_types_with_threshold
 	 */
-	public function test_it_should_not_create_the_original_mime_type_for_jpeg_images( string $image_type, bool $apply_threshold = false ): void {
+	public function test_it_should_not_create_the_original_mime_type_for_jpeg_images( string $image_type, bool $above_big_image_size = false ): void {
 		$mime_type = 'image/' . $image_type;
 		$this->set_image_output_type( $image_type );
 		if ( ! webp_uploads_mime_type_supported( $mime_type ) ) {
@@ -53,7 +53,7 @@ class Test_WebP_Uploads_Load extends TestCase {
 		}
 		$attachment_id = self::factory()->attachment->create_upload_object( TESTS_PLUGIN_DIR . '/tests/data/images/leaves.jpg' );
 
-		if ( $apply_threshold ) {
+		if ( $above_big_image_size ) {
 			// Add threshold to create a `-scaled` output image for testing.
 			add_filter(
 				'big_image_size_threshold',
--- a/plugins/webp-uploads/tests/test-load.php
+++ b/plugins/webp-uploads/tests/test-load.php
@@ -55,6 +55,7 @@ class Test_WebP_Uploads_Load extends TestCase {
 
 		// There should be an image_type source, but no JPEG source for the full image.
 		$this->assertImageHasSource( $attachment_id, $mime_type );
+		$this->assertImageNotHasSource( $attachment_id, 'image/jpeg' );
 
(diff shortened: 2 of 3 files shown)


mukeshpanchal27:
There are currently 5 unit tests which are failing on trunk:

<details><summary>There were 5 failures:</summary>

```
1) Test_WebP_Uploads_Load::test_it_should_not_create_the_original_mime_type_for_jpeg_images with data set "webp" ('webp')

Failed asserting that '2024/11/leaves-17-1080x720.webp' ends with "leaves-17-1080x720-webp.webp".
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:63

2) Test_WebP_Uploads_Load::test_it_should_not_create_the_original_mime_type_for_jpeg_images with data set "avif" ('avif')

Failed asserting that '2024/11/leaves-18-1080x720.avif' ends with "leaves-18-1080x720-avif.avif".
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:63

3) Test_WebP_Uploads_Load::test_it_should_remove_the_generated_webp_images_when_the_attachment_is_deleted

Failed asserting that file "/var/www/html/wp-content/uploads/2024/11/leaves-28-1080x720-webp.webp" does not exist.
/var/www/html/wp-content/plugins/performance/vendor/yoast/phpunit-polyfills/src/Polyfills/AssertionRenames.php:113
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:343

4) Test_WebP_Uploads_Load::test_it_should_remove_the_attached_webp_version_if_the_attachment_is_force_deleted

Failed asserting that file "/var/www/html/wp-content/uploads/2024/11/leaves-28-1080x720-webp.webp" does not exist.
/var/www/html/wp-content/plugins/performance/vendor/yoast/phpunit-polyfills/src/Polyfills/AssertionRenames.php:113
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:368

5) Test_WebP_Uploads_Load::test_it_should_remove_full_size_images_when_no_size_image_exists

Failed asserting that file "/var/www/html/wp-content/uploads/2024/11/leaves-28-1080x720-webp.webp" does not exist.
/var/www/html/wp-content/plugins/performance/vendor/yoast/phpunit-polyfills/src/Polyfills/AssertionRenames.php:113
/var/www/html/wp-content/plugins/performance/plugins/webp-uploads/tests/test-load.php:391
```

</details>

The issue is in the core, it change the metadata for `file` key and our plugin use that key. Check https://github.com/WordPress/wordpress-develop/pull/7733#issuecomment-2458936274 for more details.

adamsilverstein (approved):
Thanks!

mukeshpanchal27:
@adamsilverstein Have you had chance to review the failed unit tests for trunk? Is it expected behaviour for trunk? Have you review https://github.com/WordPress/wordpress-develop/pull/7733?

felixarntz (changes requested):
@mukeshpanchal27 I'm a bit concerned about simply removing the assertions. Why is that the right choice?

To me, these assertions failing means that something may be broken by the Core changes, that we either have to fix in the plugin or in Core. Or can you clarify why the expected behavior should have changed now?

felixarntz:
On another note: It looks like the following tests are doing exactly the same:
* `test_it_should_create_jpeg_and_webp_for_jpeg_images_if_opted_in`
* `test_it_should_create_jpeg_and_webp_for_jpeg_images_if_generate_webp_and_jpeg_set`

One of them should probably be removed?

felixarntz:
@mukeshpanchal27 @adamsilverstein I found the actual problem here: It's a bug that's been in the plugin for a long time, but so far would only have affected images _above_ the "big image" threshold. We just don't have any tests with a large image, so we should definitely add one with an image above that threshold to have that covered in the future.

Here's the problem:
* The plugin relies on the `post_mime_type` to know what format the image has. This leads to the `sources` property having entries like `'image/jpeg' => array( /* data about an AVIF or WebP image */ )` which makes no sense.
* Using `post_mime_type` is unreliable, as WordPress Core does not update `post_mime_type` when the image format is changed.
    * While it may be worth thinking about changing that, this is how WordPress Core has always worked, so it's an existing problem we just never spotted.
    * FWIW our plugin doesn't update the `post_mime_type` either.
* We need to instead get the MIME type from the actual file, if possible. See the following code adjustment for the beginning of the `webp_uploads_create_sources_property()` function:

```php
function webp_uploads_create_sources_property( array $metadata, int $attachment_id ): array {
	$file = get_attached_file( $attachment_id, true );
	// File does not exist.
	if ( false === $file || ! file_exists( $file ) ) {
		return $metadata;
	}

	/*
	 * We need to get the MIME type ideally from the file, as WordPress Core may have already altered it.
	 * The post MIME type is typically not updated during that process.
	 */
	$filetype = wp_check_filetype( $file );
	if ( $filetype['type'] ) {
		$mime_type = $filetype['type'];
	} else {
	}

	$valid_mime_transforms = webp_uploads_get_upload_image_mime_transforms();

	if ( ! is_string( $mime_type ) || ! isset( $valid_mime_transforms[ $mime_type ] ) ) {
		return $metadata;
	}

	// Make sure the top level `sources` key is a valid array.
	if ( ! isset( $metadata['sources'] ) || ! is_array( $metadata['sources'] ) ) {
		$metadata['sources'] = array();
	}

	// Remaining code can probably stay as is.
}
```

Additionally to fixing that, there's a few more places in the plugin where we call `get_post_mime_type()`, so we should validate this doesn't cause other bugs in those places or fix them accordingly.

Last but not least, as mentioned let's add test coverage using an image that is above the "big image" threshold. The easiest way is probably to keep using the existing tests, but expand the `dataProvider` approach to not only specify the MIME type to use but also the source file. This way we ensure the expected outcome is the same regardless of image size, which should ideally be the case.

swissspidy:
> Last but not least, as mentioned let's add test coverage using an image that is above the "big image" threshold. The easiest way is probably to keep using the existing tests, but expand the `dataProvider` approach to not only specify the MIME type to use but also the source file. This way we ensure the expected outcome is the same regardless of image size, which should ideally be the case.

Or just use the `big_image_size_threshold` filter to reduce the number a little bit? Otherwise tests will take a longer time with larger images.

adamsilverstein:
> Using post_mime_type is unreliable, as WordPress Core does not update post_mime_type when the image format is changed.

`get_post_mime_type` should always return the mime type of the uploaded image (input type).. It looks like here we want the output mime type, is that right?

felixarntz:
> > Using post_mime_type is unreliable, as WordPress Core does not update post_mime_type when the image format is changed.
> 
> `get_post_mime_type` should always return the mime type of the uploaded image (input type).. It looks like here we want the output mime type, is that right?

It always returns the MIME type of the uploaded image, but here we need the MIME type that the image is currently in, which may or may not be the output MIME type. The issue is that Core may already have processed the image before this function in our plugin is called, and at that point `get_post_mime_type` would still return the originally uploaded image MIME type, not the MIME type the attachment file is actually in.

azaozz:
> It always returns the MIME type of the uploaded image, but here we need the MIME type that the image is currently in

Yea, this is a limitation of the image metadata: it doesn't store the actual mime type of the uploaded image if it was converted. One workaround to speed this up (that I think is still used in core) is to look at the first item from the `sizes` array where the actual mime-type is set. Of course it would be better to get the mime-type from the file if speed is not a big concern.

plugins/webp-uploads/tests/test-load.php
```diff
@@ -43,16 +43,26 @@ static function ( string $filename ) {
 	/**
 	 * Don't create the original mime type for JPEG images.
 	 *
-	 * @dataProvider data_provider_supported_image_types
+	 * @dataProvider data_provider_supported_image_types_with_threshold
 	 */
-	public function test_it_should_not_create_the_original_mime_type_for_jpeg_images( string $image_type ): void {
+	public function test_it_should_not_create_the_original_mime_type_for_jpeg_images( string $image_type, bool $apply_threshold = false ): void {
```
felixarntz: Maybe call this something like `$above_big_image_size` for more clarity?

plugins/webp-uploads/hooks.php
```diff
@@ -52,21 +52,30 @@
...
+		$mime_type = $filetype['type'];
+	} else {
+		$mime_type = get_post_mime_type( $attachment_id );
+	}
+
 	$valid_mime_transforms = webp_uploads_get_upload_image_mime_transforms();
 
 	// Not a supported mime type to create the sources property.
-	$mime_type = get_post_mime_type( $attachment_id );
 	if ( ! is_string( $mime_type ) || ! isset( $valid_mime_transforms[ $mime_type ] ) ) {
```
felixarntz: Nit-pick: The `! is_string( $mime_type )` early return check could happen before the call to `webp_uploads_get_upload_image_mime_transforms()` to avoid that function call when not relevant.

felixarntz (approved):
@mukeshpanchal27 Thanks for the update, this fixes the test failures. Just two non-blocking notes.

Additionally, please consider my note about other usages of `get_post_mime_type()`. I just checked, and the one in `webp_uploads_wrap_image_in_picture()` looks incorrect to me as well, for the same reason here. Maybe we could take the logic you have here in lines 65-70 and move them to their own helper function? That way we could use that in both places. This could be done in a separate PR though, as it's not causing test failures today.

plugins/webp-uploads/hooks.php
```diff
@@ -52,21 +52,30 @@
...
+	/*
+	 * We need to get the MIME type ideally from the file, as WordPress Core may have already altered it.
+	 * The post MIME type is typically not updated during that process.
+	 */
+	$filetype = wp_check_filetype( $file );
+	if ( isset( $filetype['type'] ) ) {
+		$mime_type = $filetype['type'];
+	} else {
+		$mime_type = get_post_mime_type( $attachment_id );
+	}
```
westonruter: This can be simplified:

```suggestion
	$filetype  = wp_check_filetype( $file );
	$mime_type = $filetype['type'] ?? get_post_mime_type( $attachment_id );
	if ( ! is_string( $mime_type ) ) {
		return $metadata;
	}
```

Also applying the suggestion from @felixarntz in https://github.com/WordPress/performance/pull/1635/files#r1832951586

plugins/webp-uploads/hooks.php
westonruter: Suggested change above.

```suggestion
	if ( ! isset( $valid_mime_transforms[ $mime_type ] ) ) {
```

mukeshpanchal27:
> This could be done in a separate PR though, as it's not causing test failures today.

That's the plan once these PR merged.

PR #1636, opened 2024-11-07
Update readmes for Image Prioritizer, Embed Optimizer, and Optimization Detective

### LINKED ISSUE #1624: Update readmes for Image Prioritizer and Embed Optimizer with the new features that landed with PL 3.5.0
The [v3.5.0](https://github.com/WordPress/performance/releases/tag/3.5.0) of Performance Lab included key features for Image Prioritizer and Embed Optimizer. These were noted in the respective plugin changelogs, but they did not get written up in the readme copy. This should be done to ensure users know what optimizations to expect.

westonruter:
Fixes #1624 

For edits, please suggest changes on the [Google Doc](https://docs.google.com/document/d/1mkCfKBdTcTjaZzHf3G8dcJpA8gfNDlsgFgzDk0H7GCI/edit?usp=sharing).

Additionally, the term "URL Metric" has been normalized in copy to have a capital M since it is a thing. So "URL Metric" instead of "URL metric".


plugins/image-prioritizer/readme.txt
```diff
@@ -11,22 +11,27 @@ Optimizes LCP image loading with `fetchpriority=high` and applies image lazy-loa
 
 == Description ==
 
-This plugin optimizes the loading of images which are the LCP (Largest Contentful Paint) element, including both `img` elements and elements with CSS background images (where there is a `style` attribute with an `background-image` property). Different breakpoints in a theme's responsive design may result in differing elements being the LCP element. Therefore, the LCP element for each breakpoint is captured so that high-fetchpriority preload links with media queries are added which prioritize loading the LCP image specific to the viewport of the visitor.
+This plugin optimizes the loading of images (and videos) with prioritization, lazy loading, and even some size reduction:
```
felixarntz: Two things regarding this:
* "even some size reduction" can be misunderstood that the plugin may reduce file size of images, which of course it does not. I would rephrase this to be about "size selection" rather than "size reduction".
* I think this sentence should end in a period, as sort of a short description, and then you can use "The current optimizations include:", just how it is for Embed Optimizer.
```suggestion
This plugin optimizes the loading of images (and videos) with prioritization, lazy loading, and more accurate image size selection.

The current optimizations include:
```

plugins/embed-optimizer/readme.txt
```diff
@@ -11,9 +11,23 @@ Optimizes the performance of embeds by lazy-loading iframes and scripts.
 
 == Description ==
 
-This plugin's purpose is to optimize the performance of [embeds in WordPress](https://wordpress.org/documentation/article/embeds/), such as YouTube videos, TikToks, and so on. Initially this is achieved by lazy-loading them only when they come into view. This improves performance because embeds are generally very resource-intensive and so lazy-loading them ensures that they don't compete with resources when the page is loading. [Other optimizations](https://github.com/WordPress/performance/issues?q=is%3Aissue+is%3Aopen+label%3A%22%5BPlugin%5D+Embed+Optimizer%22) are planned for the future.
+This plugin's purpose is to optimize the performance of [embeds in WordPress](https://wordpress.org/documentation/article/embeds/), such as Tweets, YouTube videos, TikToks, and so on. The current optimizations include:
```
felixarntz: Maybe have the first sentence on its own, as a short description at a glance.

Also, to me "and so on" sounds too colloquial, maybe better "and others"?

```suggestion
This plugin's purpose is to optimize the performance of [embeds in WordPress](https://wordpress.org/documentation/article/embeds/), such as Tweets, YouTube videos, TikToks, and others.

The current optimizations include:
```

plugins/image-prioritizer/readme.txt
```diff
@@ -11,22 +11,27 @@ Optimizes LCP image loading with `fetchpriority=high` and applies image lazy-loa
...
+5. Reducing the size of the `poster` image of a `VIDEO` from full size to the size appropriate for the maximum width of the video (on desktop).
+6. Lazy-loading `VIDEO` tags by setting the appropriate attributes based on whether they appear in the initial viewport. If a `VIDEO` is the LCP element, it gets `preload=auto`; if it is in an initial viewport, the `preload=metadata` default is left; if it is not in an initial viewport, it gets `preload=none`. Lazy-loaded videos also get initial `preload`, `autoplay`, and `poster` attributes restored when the `VIDEO` is going to enter the viewport.
 
-1. 0-320 (small smartphone)
-2. 321-480 (normal smartphone)
-3. 481-576 (phablets)
-4. >576 (desktop)
+This plugin requires the [Optimization Detective](https://wordpress.org/plugins/optimization-detective/) as a dependency. Please refer to that plugin for additional background on how this plugin works as well as additional developer options. Information about images and videos displayed on a given visit are stored in URL Metrics grouped by the following viewport widths according to WordPress’s common responsive breakpoints:
```
felixarntz: Missing "plugin" word, also I'd suggest to make this bold like you did for Embed Optimizer.

For content, I think this has too many details about Optimization Detective, e.g. the breakpoints are defined by Optimization Detective, so such information should be in the readme of Optimization Detective, which is already referred to here for more information. Otherwise, such information would be just as appropriate to have in the Embed Optimizer readme, at which point we are duplicating information.
```suggestion
**This plugin requires the [Optimization Detective](https://wordpress.org/plugins/optimization-detective/) plugin as a dependency.** Please refer to that plugin for additional background on how this plugin works as well as additional developer options.
```

plugins/image-prioritizer/readme.txt
```diff
@@ -11,22 +11,27 @@ Optimizes LCP image loading with `fetchpriority=high` and applies image lazy-loa
...
 
-If an image does not appear in the initial viewport for any of these viewport groups, then `loading=lazy` is added to the `img` element. 
+Mobile: 0-480px  
+Phablet: 481-600px  
+Tablet: 601-782px  
+Desktop: \>782px
```
felixarntz: See above:
```suggestion
```

plugins/image-prioritizer/readme.txt
```diff
@@ -11,22 +11,27 @@ Optimizes LCP image loading with `fetchpriority=high` and applies image lazy-loa
...
 
-👉 **Note:** This plugin optimizes pages for actual visitors, and it depends on visitors to optimize pages (since URL metrics need to be collected). As such, you won't see optimizations applied immediately after activating the plugin. And since administrator users are not normal visitors typically, optimizations are not applied for admins by default.
+This grouping is essential for proper optimization since different breakpoints in a theme's responsive design may result in differing elements being the LCP element or whether an element is visible at all.
```
felixarntz: See above:
```suggestion
```

plugins/image-prioritizer/readme.txt
```diff
@@ -11,22 +11,27 @@ Optimizes LCP image loading with `fetchpriority=high` and applies image lazy-loa
...
 
-There are currently **no settings** and no user interface for this plugin since it is designed to work without any configuration.
+👉 **Note:** This plugin optimizes pages for actual visitors, and it depends on visitors to optimize pages. As such, you won't see optimizations applied immediately after activating the plugin. Please wait for URL Metrics to be gathered for both mobile and desktop visits. And since administrator users are not normal visitors typically, optimizations are not applied for admins by default.
```
felixarntz, line 27: Can you add a similar note to the Optimization Detective specific section in the Embed Optimizer readme?

plugins/optimization-detective/readme.txt
```diff
@@ -11,7 +11,7 @@ Provides an API for leveraging real user metrics to detect optimizations to appl
...
+This plugin captures real user metrics about what elements are displayed on the page across a variety of device form factors (e.g. desktop, tablet, and phone) in order to apply loading optimizations which are not possible with WordPress’s current server-side heuristics. This plugin is a dependency which does not provide end-user functionality on its own. For that, please install the dependent plugin [Image Prioritizer](https://wordpress.org/plugins/image-prioritizer/) or [Embed Optimizer](https://wordpress.org/plugins/embed-optimizer/) (among [others](https://github.com/WordPress/performance/labels/%5BPlugin%5D%20Optimization%20Detective) to come from the WordPress Core Performance team).
```
felixarntz: Similar to the other readmes, I'd suggest to have the first sentence as a short description on its own:
```suggestion
This plugin captures real user metrics about what elements are displayed on the page across a variety of device form factors (e.g. desktop, tablet, and phone) in order to apply loading optimizations which are not possible with WordPress’s current server-side heuristics.

This plugin is a dependency which does not provide end-user functionality on its own. For that, please install the dependent plugin [Image Prioritizer](https://wordpress.org/plugins/image-prioritizer/) or [Embed Optimizer](https://wordpress.org/plugins/embed-optimizer/) (among [others](https://github.com/WordPress/performance/labels/%5BPlugin%5D%20Optimization%20Detective) to come from the WordPress Core Performance team).
```

plugins/optimization-detective/readme.txt
```diff
@@ -25,7 +25,7 @@ At the core of Optimization Detective is the “URL Metric”, information about
```
felixarntz, line 34: Can you either include the breakpoints here, or at least somehow reference the filter doc below where you have them outlined?

felixarntz (commented):
@westonruter These updated readmes look great, thanks for all the effort you put into this.

I left some suggestions here, but mostly minor iterations.

(FWIW I personally find it easier to review this here than in a Google doc, as it also shows the diff from the current readmes.)

plugins/image-prioritizer/readme.txt
westonruter: I think there is already such a note? See:

https://github.com/WordPress/performance/blob/9d9136645cd94734e296dadf336473333f7a4f3f/plugins/embed-optimizer/readme.txt#L28

plugins/optimization-detective/readme.txt
westonruter: How about 4e2d2fb7?

plugins/image-prioritizer/readme.txt
felixarntz: Ah missed that, thanks!

plugins/optimization-detective/readme.txt
felixarntz: Mostly looks good, one follow up separately.

plugins/optimization-detective/readme.txt
```diff
@@ -21,11 +23,18 @@ In order to increase the accuracy of identifying the LCP element, including acro
...
+At the core of Optimization Detective is the “URL Metric”, information about a page according to how it was loaded by a client with a specific viewport width. This includes which elements were visible in the initial viewport and which one was the LCP element. The URL Metric data is also extensible. Each URL on a site can have an associated set of these URL Metrics (stored in a custom post type) which are gathered from the visits of real users. It gathers samples of URL Metrics which are grouped according to WordPress's default responsive breakpoints:
+
+1. Mobile: 0-480px
+2. Phablet: 481-600px
```
felixarntz, line 29: Never heard that word 😆 

Wondering, do we really need that breakpoint? Phablet and tablet are only ~120px and ~180px ranges respectively, which I find very little. Why not combine them like you have in the filter docs below? (Just a thought, that would need to be another PR though.)

plugins/optimization-detective/readme.txt
```diff
@@ -37,16 +46,47 @@ When the `WP_DEBUG` constant is enabled, additional logging for Optimization Det
...
+Refer to [Image Prioritizer](https://github.com/WordPress/performance/tree/trunk/plugins/image-prioritizer) and [Embed Optimizer](https://github.com/WordPress/performance/tree/trunk/plugins/embed-optimizer) for real world examples of how tag visitors are used. Registered tag visitors need only be callables, so in addition to providing a closure you may provide a `callable-string` or even a class which has an `__invoke()` method.
 
-The default breakpoints are reused from Gutenberg which appear to be used the most in media queries that affect frontend styles.
+**Filter:** `od_breakpoint_max_widths` (default: `array(480, 600, 782)`)
+
+Filters the breakpoint max widths to group URL Metrics for various viewports. Each number represents the maximum width (inclusive) for a given breakpoint. So if there is one number, 480, then this means there will be two viewport groupings, one for 0\<=480, and another \>480. If instead there are the two breakpoints defined, 480 and 782, then this means there will be three viewport groups of URL Metrics:
+
+1. Mobile: 0-480px
+2. Tablet: 481-782px
+3. Desktop: \>782px
```
felixarntz: I think you should avoid using a list for this and put it inline in the text, just like you did for the other example of two widths.

The list makes it seem somehow "official", confusing with the list of actual default breakpoints the plugin uses. For instance, the example doesn't mention giving the breakpoints a name, but the list has names, which I think makes it even more confusing. So I think it would be best to only refer to:
```suggestion
Filters the breakpoint max widths to group URL Metrics for various viewports. Each number represents the maximum width (inclusive) for a given breakpoint. So if there is one number, 480, then this means there will be two viewport groupings, one for 0\<=480, and another \>480. If instead there are the two breakpoints defined, 480 and 782, then this means there will be three viewport groups of URL metrics, one for 0\<=480, another 481\<=782, and another \>782.
```

felixarntz (approved):
@westonruter Looks great, thanks! Just two minor notes, potential for follow up work.

plugins/optimization-detective/readme.txt
westonruter: > Never heard that word 😆

It's actually a thing! Per [Wikipedia](https://en.wikipedia.org/wiki/Phablet) and in Google Search:

image

> Wondering, do we really need that breakpoint? Phablet and tablet are only ~120px and ~180px ranges respectively, which I find very little. Why not combine them like you have in the filter docs below? (Just a thought, that would need to be another PR though.)

Good question. These numbers come from Gutenberg's breakpoints:

https://github.com/WordPress/gutenberg/blob/ab6e729daf775ee6065a48c210826af8281a0842/packages/base-styles/_breakpoints.scss#L11-L13

```
$break-medium: 782px;	// adminbar goes big
$break-small: 600px;
$break-mobile: 480px;
```

It's [used](https://github.com/WordPress/gutenberg/blob/ab6e729daf775ee6065a48c210826af8281a0842/packages/block-library/src/media-text/style.scss#L126-L145) in the Media & Text block, for example:

[Screen recording 2024-11-07 15.56.13.webm](https://github.com/user-attachments/assets/4d8e65a4-f403-4280-8fcf-e36fac835e11)

So this can be a critical breakpoint which affects the the LCP element.

Note also that this is the breakpoint at which the navigation menu transitions from mobile to non-mobile.

There are quite a few other uses of this `break-small` width: https://github.com/search?q=repo%3AWordPress%2Fgutenberg+%2F%5C%24break-small%7Cbreak-small%5C%28%5C%29%2F&type=code

I recall looking through the various breakpoints defined in Gutenberg and finding that these are the ones most commonly used in responsive styles. So that's why I picked these specifically.

plugins/optimization-detective/readme.txt
westonruter: I made a small addition on top of this suggestion: e9827afc

westonruter:
READMEs have been deployed:

- https://wordpress.org/plugins/image-prioritizer/
- https://wordpress.org/plugins/embed-optimizer/
- https://wordpress.org/plugins/optimization-detective/

Issue #1496, opened 2024-08-22
Detection time window may not allow enough URL Metrics to be collected


westonruter:
Originally in https://github.com/WordPress/performance/issues/1494#issuecomment-2305825246.

I discovered on my blog that I had a very old URL Metric from before the XPath indices were fixed to be 1-based instead of 0-based (#1191). Looking at the URL Metric data, I see the timestamp is `1712354502.752218` which is for a day back in April. It appears that URL Metrics are not being collected frequently enough, or my site is truly devoid of traffic. I think for my site it may be in part due to page caching, where the detection time window (`od_detection_time_window`) is too short, so it is only collecting metrics for visitors who happen to visit when there is a MISS to the page cache, and if there is a stale-while-revalidate caching strategy going on with Varnish then it is even more likely for the visitor to never be within the detection time window (`currentTime - serveTime > detectionTimeWindow`):

https://github.com/WordPress/performance/blob/f8dad3497f6240e2975733c3b277e91e35b254a6/plugins/optimization-detective/detect.js#L179-L187

I think this logic needs to be revisited.

felixarntz:
Related is a conversation @westonruter and I had at WordCamp US: To reliably handle dynamic behavior on the client-side, the behavior needs to be controlled by the client-side, to account for full page caches which will lead to most requests not to reach the PHP logic. This basically means two things:
* We need to inject the script, or at least a very minimal version of the script, unconditionally, because we can't know from the server-side whether the current request needs the detection to run (e.g. if it never hits the PHP process).
* We need to retrieve data that is dynamic in another way, it is not reliable to inline it with the script. Some data that is specific to the current URL (e.g. `urlMetricsSlug`) or a general configuration setting (e.g. `storageLockTTL`) is probably okay to inline, but most other data (e.g. `urlMetricsGroupStatuses`) will get out of date when a full page cache is used. Such data would probably need to come from a REST endpoint instead.

Defining the right solution for this probably goes beyond this issue, maybe we should open a new one, that is more broadly about how to support full page caches in a way that the Optimization Detective functionality is still reliable? I think prioritizing this will be crucial to get broader adoption, as lots of sites use a full page cache which will affect the behavior negatively - sometimes minor, sometimes more severely, depending on the cache configuration.

westonruter:
Here is a table for what I've found the default max-age to be for caching plugins:

Plugin Name | Default Max-Age
--|--:
[...]

PR #1637, opened 2024-11-08
Revise the use of nonces in requests to store a URL Metric and block cross-origin requests

westonruter:
This is the first PR to address #1496.

1. Eliminate the use of the `wp_rest` nonce when making POST requests to store a URL Metric. Since the endpoint does not require authentication, the nonce adds no value.
2. Add enforcement of the `Origin` request header so that it must match `home_url()` in order for the URL Metric storage POST request to be accepted. The WordPress REST API by default allows for cross-origin requests, but this makes no sense specifically for storing a URL Metric since the client must by definition be on the frontend of the site.
3. Eliminate the use of `wp_create_nonce()` to compute the HMAC to validate that the provided query-vars `slug` actually corresponds to the provided `url` in the request (that is, the URL of the current page). There is no need to use `wp_create_nonce()` here, since there does not need to be an expiration and it does not need to be tied to the current logged-in user.

By eliminating the use of nonces, clients will not have issues when submitting a URL Metric for storage when a page cache is holding onto the rendered page for longer than a day.

### DIFF
--- a/plugins/optimization-detective/storage/data.php
+++ b/plugins/optimization-detective/storage/data.php
@@ -141,42 +141,45 @@ function od_get_url_metrics_slug( array $query_vars ): string {
 }
 
 /**
- * Computes nonce for storing URL Metrics for a specific slug.
+ * Computes HMAC for storing URL Metrics for a specific slug.
  *
  * This is used in the REST API to authenticate the storage of new URL metrics from a given URL.
  *
  * @since 0.1.0
+ * @since n.e.x.t Renamed from od_get_url_metrics_storage_nonce().
  * @access private
  *
- * @see wp_create_nonce()
- * @see od_verify_url_metrics_storage_nonce()
+ * @see od_verify_url_metrics_storage_hmac()
  * @see od_get_url_metrics_slug()
  *
  * @param string $slug Slug (hash of normalized query vars).
  * @param string $url  URL.
- * @return string Nonce.
+ *
+ * @return string HMAC.
  */
-function od_get_url_metrics_storage_nonce( string $slug, string $url ): string {
-	return wp_create_nonce( "store_url_metrics:$slug:$url" );
+function od_get_url_metrics_storage_hmac( string $slug, string $url ): string {
+	$action = "store_url_metric:$slug:$url";
+	return wp_hash( $action, 'nonce' );
 }
 
 /**
- * Verifies nonce for storing URL metrics for a specific slug.
+ * Verifies HMAC for storing URL metrics for a specific slug.
  *
  * @since 0.1.0
+ * @since n.e.x.t Renamed from od_verify_url_metrics_storage_nonce().
  * @access private
  *
- * @see wp_verify_nonce()
- * @see od_get_url_metrics_storage_nonce()
+ * @see od_get_url_metrics_storage_hmac()
  * @see od_get_url_metrics_slug()
  *
- * @param string $nonce Nonce.
- * @param string $slug  Slug (hash of normalized query vars).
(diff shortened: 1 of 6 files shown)


plugins/optimization-detective/storage/data.php
```diff
@@ -141,42 +141,45 @@ function od_get_url_metrics_slug( array $query_vars ): string {
 }
 
 /**
- * Computes nonce for storing URL Metrics for a specific slug.
+ * Computes HMAC for storing URL Metrics for a specific slug.
  *
  * This is used in the REST API to authenticate the storage of new URL Metrics from a given URL.
  *
  * @since 0.1.0
+ * @since n.e.x.t Renamed from od_get_url_metrics_storage_nonce().
```
felixarntz: Better to consider this an entirely new function. I would argue it wasn't "just" renamed, the purpose was also changed as it now returns a different kind of value.

plugins/optimization-detective/storage/data.php
```diff
@@ -141,42 +141,45 @@ function od_get_url_metrics_slug( array $query_vars ): string {
...
+	$action = "store_url_metric:$slug:$url";
+	return wp_hash( $action, 'nonce' );
 }
 
 /**
- * Verifies nonce for storing URL Metrics for a specific slug.
+ * Verifies HMAC for storing URL Metrics for a specific slug.
  *
  * @since 0.1.0
+ * @since n.e.x.t Renamed from od_verify_url_metrics_storage_nonce().
```
felixarntz: See above, I'd prefer we consider this a new function.

plugins/optimization-detective/storage/rest-api.php
```diff
@@ -88,6 +88,42 @@ function od_register_endpoint(): void {
...
+ * @since n.e.x.t
+ * @access private
+ *
+ * @see get_allowed_http_origins()
+ * @see is_allowed_http_origin()
+ *
+ * @param string $origin Origin to check.
+ * @return bool Whether the origin is allowed.
+ */
+function od_is_allowed_http_origin( string $origin ): bool {
```
felixarntz, line 106: Why can't we use `is_allowed_http_origin()` directly instead of this function?

If it's just about the port support, I'd say that's an edge case so then I think we should at least use the original function anytime there's no port, and the custom logic only if there is one.

I'd also say we could simplify it: If there is a port in the passed `$origin`, simply remove it for the comparison and call `is_allowed_http_origin()` with that. It's not really that important to validate that the port also matches IMO, especially because of how much of an edge case that is.

felixarntz (approved):
@westonruter This looks great! Just some minor feedback, which would be good to address, but no blockers.

plugins/optimization-detective/storage/data.php
westonruter: Sure. Done in 9d6e80784

plugins/optimization-detective/storage/data.php
westonruter: Sure. Done in 9d6e80784

plugins/optimization-detective/storage/rest-api.php
westonruter: Well, unit tests run the home being `localhost:8889` so it's not an edge case from that perspective. 

I've gone ahead and just stripped out the port number in 545f0c3e6.

felixarntz: Great, I think that simplifies the implementation, and validating the same port number is probably not super important (for `localhost` sites there's probably not too much of a security implication anyway).

Issue #1466, opened 2024-08-13
Internal state of the page (e.g. queried posts or queried object) should factor into whether a stored URL metric is fresh


westonruter:
This was originally brought up in https://github.com/WordPress/performance/pull/878#discussion_r1393132794.

There is a TODO currently in `od_get_normalized_query_vars()`:

https://github.com/WordPress/performance/blob/73f223598df758e507dcfd14e09681e156a850d2/plugins/optimization-detective/storage/data.php#L59

Consider a homepage that shows 10 posts. None of the posts have featured images. URL metrics have been fully populated for the page in this state. Then someone publishes a new post that includes a featured image. All of a sudden there is a new image on the page which may be LCP, and yet the URL metrics have no record of this element. Nevertheless, by default the URL metrics are considered fresh by default for [1 day](https://github.com/WordPress/performance/blob/73f223598df758e507dcfd14e09681e156a850d2/plugins/optimization-detective/storage/data.php#L34), which means it could take a day for that new image to start getting optimized.

One way to deal with this would be to include the IDs of the posts in the query among the normalized query vars. This could similarly handle the case where the static front page is changed. For example, in [`od_get_normalized_query_vars()`](https://github.com/WordPress/performance/blob/73f223598df758e507dcfd14e09681e156a850d2/plugins/optimization-detective/storage/data.php#L66C10-L86):

```php
if ( is_singular() ) {
    $normalized_query_vars['queried_object_id'] = get_queried_object_id();
} else {
    $normalized_query_vars['queried_post_ids'] = join( ',', wp_list_pick( $GLOBALS['wp_query']->posts, 'ID' ) );
}
```

When the query vars change, this results in an entirely separate `od_url_metrics` post being used, as the normalized query vars are passed into [`od_get_url_metrics_slug()`](https://github.com/WordPress/performance/blob/73f223598df758e507dcfd14e09681e156a850d2/plugins/optimization-detective/storage/data.php#L126-L141) to obtain a MD5 hash which is used as the `post_name`.

Nevertheless, this seems somewhat of an abuse of the normalized query vars here, since these new "query vars" aren't actually part of the query at all. In reality, a new URL metrics post shouldn't be created, as it should instead update the existing post. If it creates a new post, then the old one will stick around until it is [garbage-collected after a month](https://github.com/WordPress/performance/blob/73f223598df758e507dcfd14e09681e156a850d2/plugins/optimization-detective/storage/class-od-url-metrics-post-type.php#L276-L300).

[...]

PR #1641, opened 2024-11-10
Send post ID of queried object or first post in loop in URL Metric storage request to schedule page cache validation

### LINKED ISSUE #1496: Detection time window may not allow enough URL Metrics to be collected

westonruter:
Fixes #1496.

This seeks to proactively address the problem of a page caching plugin holding onto an unoptimized cached page even after URL Metrics have been submitted by visitors.

Page caches generally will flush caches for a queried object's URL when the relevant actions fire (e.g. `clean_post_cache`, `edit_post`, `transition_post_status`). By sending the queried object information with the URL Metric storage request we can trigger these actions so that page caching plugins will flush the page caches in order so that optimizations leveraging the newly-stored URL Metric will promptly be reflected on the frontend.

These actions are not triggered at the same moment that the new URL Metric is stored, as this could cause a cache stampede. Instead, a scheduled event is created 10 minutes into the future to trigger the cache invalidation for the post ID. Since the post ID is passed as an argument to the scheduled event, WP Cron will automatically prevent duplicates.

### DIFF
--- a/plugins/optimization-detective/detect.js
+++ b/plugins/optimization-detective/detect.js
@@ -2,7 +2,6 @@
  * @typedef {import("web-vitals").LCPMetric} LCPMetric
  * @typedef {import("./types.ts").ElementData} ElementData
  * @typedef {import("./types.ts").URLMetric} URLMetric
- * @typedef {import("./types.ts").QueriedObject} QueriedObject
  * @typedef {import("./types.ts").URLMetricGroupStatus} URLMetricGroupStatus
  * @typedef {import("./types.ts").Extension} Extension
  * @typedef {import("./types.ts").ExtendedRootData} ExtendedRootData
@@ -240,7 +239,7 @@ function extendElementData( xpath, properties ) {
  * @param {string}                 args.restApiEndpoint            URL for where to send the detection data.
  * @param {string}                 args.currentUrl                 Current URL.
  * @param {string}                 args.urlMetricSlug              Slug for URL Metric.
- * @param {QueriedObject}          [args.queriedObject]            Queried object.
+ * @param {number|null}            args.cachePurgePostId           Cache purge post ID.
  * @param {string}                 args.urlMetricHMAC              HMAC for URL Metric storage.
  * @param {URLMetricGroupStatus[]} args.urlMetricGroupStatuses     URL Metric group statuses.
  * @param {number}                 args.storageLockTTL             The TTL (in seconds) for the URL Metric storage lock.
@@ -255,7 +254,7 @@ export default async function detect( {
 	restApiEndpoint,
 	currentUrl,
 	urlMetricSlug,
-	queriedObject,
+	cachePurgePostId,
 	urlMetricHMAC,
 	urlMetricGroupStatuses,
 	storageLockTTL,
@@ -449,10 +448,6 @@ export default async function detect( {
 		elements: [],
 	};
 
-	if ( queriedObject ) {
-		urlMetric.queriedObject = queriedObject;
-	}
-
 	const lcpMetric = lcpMetricCandidates.at( -1 );
 
 	for ( const elementIntersection of elementIntersections ) {
@@ -539,6 +534,12 @@ export default async function detect( {
 
 	const url = new URL( restApiEndpoint );
 	url.searchParams.set( 'slug', urlMetricSlug );
+	if ( typeof cachePurgePostId === 'number' ) {
+		url.searchParams.set(
+			'cache_purge_post_id',
+			cachePurgePostId.toString()
+		);
+	}
 	url.searchParams.set( 'hmac', urlMetricHMAC );
 	navigator.sendBeacon(
 		url,
--- a/plugins/optimization-detective/detect.js
+++ b/plugins/optimization-detective/detect.js
@@ -459,16 +459,21 @@ export default async function detect( {
 			continue;
 		}
 
-		const isLCP =
-			elementIntersection.target === lcpMetric?.entries[ 0 ]?.element;
+		const element = /** @type {Element|null} */ (
+			lcpMetric?.entries[ 0 ]?.element
+		);
+		const isLCP = elementIntersection.target === element;
 
 		/** @type {ElementData} */
 		const elementData = {
 			isLCP,
 			isLCPCandidate: !! lcpMetricCandidates.find(
-				( lcpMetricCandidate ) =>
-					lcpMetricCandidate.entries[ 0 ]?.element ===
-					elementIntersection.target
+				( lcpMetricCandidate ) => {
+					const candidateElement = /** @type {Element|null} */ (
+						lcpMetricCandidate.entries[ 0 ]?.element
+					);
+					return candidateElement === elementIntersection.target;
+				}
 			),
 			xpath,
 			intersectionRatio: elementIntersection.intersectionRatio,
--- a/plugins/optimization-detective/detect.js
+++ b/plugins/optimization-detective/detect.js
@@ -2,6 +2,7 @@
  * @typedef {import("web-vitals").LCPMetric} LCPMetric
  * @typedef {import("./types.ts").ElementData} ElementData
  * @typedef {import("./types.ts").URLMetric} URLMetric
+ * @typedef {import("./types.ts").QueriedObject} QueriedObject
  * @typedef {import("./types.ts").URLMetricGroupStatus} URLMetricGroupStatus
  * @typedef {import("./types.ts").Extension} Extension
  * @typedef {import("./types.ts").ExtendedRootData} ExtendedRootData
@@ -239,6 +240,7 @@ function extendElementData( xpath, properties ) {
  * @param {string}                 args.restApiEndpoint            URL for where to send the detection data.
  * @param {string}                 args.currentUrl                 Current URL.
  * @param {string}                 args.urlMetricSlug              Slug for URL Metric.
+ * @param {QueriedObject}          [args.queriedObject]            Queried object.
  * @param {string}                 args.urlMetricHMAC              HMAC for URL Metric storage.
  * @param {URLMetricGroupStatus[]} args.urlMetricGroupStatuses     URL Metric group statuses.
  * @param {number}                 args.storageLockTTL             The TTL (in seconds) for the URL Metric storage lock.
@@ -253,6 +255,7 @@ export default async function detect( {
 	restApiEndpoint,
 	currentUrl,
 	urlMetricSlug,
+	queriedObject,
 	urlMetricHMAC,
 	urlMetricGroupStatuses,
 	storageLockTTL,
@@ -445,6 +448,10 @@ export default async function detect( {
 		},
 		elements: [],
 	};
+
+	if ( queriedObject ) {
+		urlMetric.queriedObject = queriedObject;
+	}
 
 	const lcpMetric = lcpMetricCandidates.at( -1 );
 
--- a/plugins/optimization-detective/detection.php
+++ b/plugins/optimization-detective/detection.php
@@ -32,6 +32,20 @@ function od_get_detection_script( string $slug, OD_URL_Metric_Group_Collection $
 	 */
 	$extension_module_urls = (array) apply_filters( 'od_extension_module_urls', array() );
 
+	// Obtain the queried object so when a URL Metric is stored the endpoint will know which object's cache to clean.
+	// Note that WP_Post_Type is intentionally excluded here since there is no equivalent to clean_post_cache(), clean_term_cache(), and clean_user_cache().
+	$queried_object = get_queried_object();
+	if ( $queried_object instanceof WP_Post ) {
+		$queried_object_type = 'post';
+	} elseif ( $queried_object instanceof WP_Term ) {
+		$queried_object_type = 'term';
+	} elseif ( $queried_object instanceof WP_User ) {
+		$queried_object_type = 'user';
+	} else {
+		$queried_object_type = null;
+	}
+	$queried_object_id = null === $queried_object_type ? null : (int) get_queried_object_id();
+
 	$current_url = od_get_current_url();
 	$detect_args = array(
 		'minViewportAspectRatio' => od_get_minimum_viewport_aspect_ratio(),
@@ -41,7 +55,11 @@ function od_get_detection_script( string $slug, OD_URL_Metric_Group_Collection $
 		'restApiEndpoint'        => rest_url( OD_REST_API_NAMESPACE . OD_URL_METRICS_ROUTE ),
 		'currentUrl'             => $current_url,
 		'urlMetricSlug'          => $slug,
-		'urlMetricHMAC'          => od_get_url_metrics_storage_hmac( $slug, $current_url ),
+		'queriedObject'          => null === $queried_object_type ? null : array(
+			'type' => $queried_object_type,
+			'id'   => $queried_object_id,
+		),
+		'urlMetricHMAC'          => od_get_url_metrics_storage_hmac( $slug, $current_url, $queried_object_type, $queried_object_id ),
 		'urlMetricGroupStatuses' => array_map(
 			static function ( OD_URL_Metric_Group $group ): array {
 				return array(
--- a/plugins/optimization-detective/detection.php
+++ b/plugins/optimization-detective/detection.php
@@ -11,6 +11,54 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 /**
+ * Obtains the ID for a post related to this response so that page caches can be told to invalidate their cache.
+ *
+ * If the queried object for the response is a post, then that post's ID is used. Otherwise, it uses the ID of the first
+ * post in The Loop.
+ *
+ * When the queried object is a post (e.g. is_singular, is_posts_page, is_front_page w/ show_on_front=page), then this
+ * is the perfect match. A page caching plugin will be able to most reliably invalidate the cache for a URL via
+ * this ID if the relevant actions are triggered for the post (e.g. clean_post_cache, save_post, transition_post_status).
+ *
+ * Otherwise, if the response is an archive page or the front page where show_on_front=posts (i.e. is_home), then
+ * there is no singular post object that represents the URL. In this case, we obtain the first post in the main
+ * loop. By triggering the relevant actions for this post ID, page caches will have their best shot at invalidating
+ * the related URLs. Page caching plugins which leverage surrogate keys will be the most reliable here. Otherwise,
+ * caching plugins may just resort to automatically purging the cache for the homepage whenever any post is edited,
+ * which is better than nothing.
+ *
+ * There should not be any situation by default in which a page optimized with Optimization Detective does not have such
+ * a post available for cache purging. As seen in {@see od_can_optimize_response()}, when such a post ID is not
+ * available for cache purging then it returns false, as it also does in another case like if is_404().
+ *
+ * @since n.e.x.t
+ * @access private
+ *
+ * @return int|null Post ID or null if none found.
+ */
+function od_get_cache_purge_post_id(): ?int {
+	$queried_object = get_queried_object();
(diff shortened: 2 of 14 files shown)


westonruter:
This PR is not currently accounting for prompting page caches to flush the cache for archive pages in which there is no queried object, that is, the home page (which isn't a static front page). Author and term archive pages should be accounted for with Pantheon Advanced Cache since there is a queried object and Pantheon Advanced Cache uses `clean_term_cache` and `clean_user_cache` actions to flush caches by the [relevant surrogate keys](https://github.com/pantheon-systems/pantheon-advanced-page-cache/blob/e3b5552b0cb9268d9b696cb200af56cc044920d9/README.md?plain=1#L314-L356). But it seems other caching plugins like WP Super Cache, WP Total Cache, and WP Rocket do not flush page caches for such queried objects.

Nevertheless, sometimes page cache plugins will automatically flush the homepage cache when a post is updated:

* [Batcache](https://github.com/Automattic/batcache/blob/ed0e6b2d9bcbab3924c49a6c3247646fb87a0957/batcache.php#L36-L39)
* [WP Rocket](https://github.com/wp-media/wp-rocket/blob/e5bca6673a3669827f3998edebc0c785210fe561/inc/common/purge.php#L252)
* [WP Super Cache](https://github.com/Automattic/wp-super-cache/blob/73b428d2fce397fd874b3056ad3120c343bc1a0c/wp-cache-phase2.php#L3321-L3339)
* [Pantheon Advanced Cache](https://github.com/pantheon-systems/pantheon-advanced-page-cache/blob/e3b5552b0cb9268d9b696cb200af56cc044920d9/README.md?plain=1#L323-L326) (if the post appeared on the homepage)
* [W3 Total Cache](https://www.boldgrid.com/support/w3-total-cache/page-caching/#:~:text=The%20default%20settings%2C%20including%20options%20like%20the%20front%20page%2C%20posts%20page%2C%20blog%20feed%2C%20and%20various%20archive%20pages%2C%20are%20recommended%20to%20maintain%20optimal%20server%20performance.)

What this means is that even if the loaded URL doesn't have a singular "queried object", if we instead opt to flush the cache _for the first post that appears in the loop_, then this should generally result in the page cache(s) being flushed for the desired archive pages, or at very least the non-static homepage.

plugins/optimization-detective/storage/data.php
```diff
@@ -151,14 +151,16 @@ function od_get_url_metrics_slug( array $query_vars ): string {
...
- *
+ * @param string      $slug                Slug (hash of normalized query vars).
+ * @param string      $url                 URL.
+ * @param string|null $queried_object_type Queried object type.
+ * @param int|null    $queried_object_id   Queried object ID.
  * @return string HMAC.
  */
-function od_get_url_metrics_storage_hmac( string $slug, string $url ): string {
-	$action = "store_url_metric:$slug:$url";
+function od_get_url_metrics_storage_hmac( string $slug, string $url, ?string $queried_object_type = null, ?int $queried_object_id = null ): string {
```
felixarntz: Based on what we previously discussed. I think it would be great to abstract the URL representation into a custom type. For example a class like `URL_Entity` or `URL_Content` or so, which would be able to support different ways to identify content (e.g. not everything has a "queried object" or a numeric ID, some URLs need other kinds of identifiers).

Maybe have that class implement `__toString()`, which would be a perfect fit for the usage here.

westonruter: Yeah, potentially. However, that may make more sense for https://github.com/WordPress/performance/issues/1466. In particular, I'm wary to not expose too much of the internals of how the page is constructed, as these arguments have to be passed to the frontend. The exclusive purpose of this parameter here is to communicate the endpoint which object should have its cache cleaned. Since it won't always be the "queried object" in WP parlance in the case of many archives but rather could be simply the first post in the loop. In that way, there should _always_ be some object to pass here with an integer ID. Otherwise, if there is no object, then this would normally be `is_404()` case which is not optimized anyway.

felixarntz: > I'm wary to not expose too much of the internals of how the page is constructed, as these arguments have to be passed to the frontend.

I don't think any of this would be sensitive - most of it is already surfacing publicly anyway, via `body` classes that WordPress Core adds.

> Since it won't always be the "queried object" in WP parlance in the case of many archives but rather could be simply the first post in the loop.

For that though, you would need to know _which_ loop. For example, if there was a custom post type archive "type", it would still be missing an identifier for which type, so that you could fetch the first post for that post type archive, not _any_ post type archive.

All of this IMO would make more sense to be abstracted away in a class. The class could also have it built-in to return the relevant object ID in case of an archive. It would make passing around this data more intuitive too (e.g. in JS it would just be an object with the necessary data), since all you need to know would be to pass around one object (e.g. `urlEntity` or whatever we'd call it).

westonruter: I don't think subqueries are relevant here, are they? Why would we need to know which loop the post was in?

Most caching plugins would have no idea which URL to flush when provided the post in a subquery since there is nothing connecting the top URL query vars to the arbitrary query vars passed into the subquery. More sophisticated caching plugins _would_ be able to connect these dots if they're using surrogate keys and if they do output buffering of the page and accumulate a new key for each `the_post` action. 

Pantheon Advanced Page Cache only computes surrogate keys for the posts in the main loop: https://github.com/pantheon-systems/pantheon-advanced-page-cache/blob/e3b5552b0cb9268d9b696cb200af56cc044920d9/inc/class-emitter.php#L211-L322

LightSpeed Cache's surrogate key computation is quite similar: https://github.com/litespeedtech/lscache_wp/blob/7c707469b3c88b4f45d9955593b92f9aeaed54c3/src/tag.cls.php#L239-L313

By default it is only computing surrogate keys for posts in the main loop. It does expose [actions](https://github.com/litespeedtech/lscache_wp/blob/7c707469b3c88b4f45d9955593b92f9aeaed54c3/src/api.cls.php#L83-L91) which plugins can use to add surrogate keys for subqueries, which seems to only be used in a WooCommerce integration for its related products:https://wpdirectory.net/search/01JCKMEYKEX35RM0JVJ4TE8RV3

For our purposes here, we just need to communicate to the endpoint which object should be cleaned from the cache. Thinking about this some more, there's no need for the `queriedObject` top-level key in the URL Metric schema. We just need the endpoint to accept two parameters alongside the `url` and the `slug` (in addition to the `hmac` which validates the parameters):

* The type of object which should be cleaned from the cache.
* The ID of the object which should be cleaned from the cache.

Pantheon Advanced Page Cache does the right thing to flush term and author archives when the term or user is updated. But it seems unique in this regard. I don't see that LightSpeed Cache has [any hooks](https://github.com/litespeedtech/lscache_wp/blob/7c707469b3c88b4f45d9955593b92f9aeaed54c3/src/purge.cls.php#L51-L75) to flush caches in response to anything bot a post edit (or a comment change). This being the case, perhaps using the queried object as the basis for flushing the cache isn't ideal anyway and that we should only ever pass along the ID for the post that should have its cache cleaned. This would handle the cases for:

1. Singular post/page in which the queried object is a `post`.
1. Non-static homepage in which case there is no queried object but there is an initial `post` in the loop.
2. Author/term/date archives in which there is also an initial post in the loop.

So yeah, I think this can be further simplified so that a `cache_purge_post_id` parameter is added to the endpoint, and when the endpoint receives it, it schedules an event in WP Cron to then fire the hooks to simulate an edit to that post in order for page caches to invalidate their cache for the URL(s). In reality, this will result in more than the intended URL being purged from the cache (e.g. the post permalink but also the homepage and maybe some archive as well), but there's no way to avoid that given the lack of any WordPress APIs for page caching. It would be really nice if WordPress had a standard API for page caching plugins to be told to purge the cache for a given URL!

plugins/optimization-detective/detect.js
```diff
@@ -457,16 +459,21 @@ export default async function detect( {
...
 			isLCPCandidate: !! lcpMetricCandidates.find(
-				( lcpMetricCandidate ) =>
-					lcpMetricCandidate.entries[ 0 ]?.element ===
-					elementIntersection.target
+				( lcpMetricCandidate ) => {
+					const candidateElement = /** @type {Element|null} */ (
+						lcpMetricCandidate.entries[ 0 ]?.element
+					);
+					return candidateElement === elementIntersection.target;
+				}
```
westonruter, line 476: This is just to work around a static analysis issue I get complained about in my IDE. It's because web-vitals defines the `Element` interface but it is not the same `Element` defined in the IDE's DOM types.

plugins/optimization-detective/storage/rest-api.php
```diff
@@ -226,3 +233,62 @@ function od_handle_rest_request( WP_REST_Request $request ) {
...
+ *
+ * This assumes that a page caching plugin flushes the page cache for a queried object via `clean_post_cache`,
+ * `clean_term_cache`, and `clean_user_cache` actions. Other actions may make sense to trigger as well as can be seen in
+ * {@link https://github.com/pantheon-systems/pantheon-advanced-page-cache/blob/e3b5552/README.md?plain=1#L314-L356}.
+ *
+ * @since n.e.x.t
+ *
+ * @param OD_URL_Metric_Store_Request_Context $context Context.
+ */
+function od_clean_queried_object_cache_for_stored_url_metric( OD_URL_Metric_Store_Request_Context $context ): void {
```
westonruter: TODO: Schedule an action to do this instead of doing it immediately.

plugins/optimization-detective/storage/data.php
felixarntz: I'm not entirely sure what you're referring to with "subqueries", so I only understand parts of your response.

I wouldn't tie our approach too much to what cache plugins are actually doing (particularly when there is more than they could do). That said, the approach of always clearing a specific post makes sense to me, since every URL contains at least one post in some capacity (except 404s).

What I was referring to before was the idea to pass the context around (e.g. type etc.) so that then the receiving endpoint would figure out which post to clear the cache for. I think what you're saying now is that instead OD would already determine that for the URL initially and then pass around the resulting post ID. So I think we were more or less talking about the same end goal, but with different approaches. Passing around the post ID to clear the cache for makes sense to me.

westonruter: @felixarntz Sorry, I misinterpreted what you meant by "_which_ loop". My mind went to adding additional `WP_Query` loops inside of the main loop.

I think your idea makes sense as a future enhancement to give caching plugins more say over how cache invalidation is handled. But for this first iteration, I think sending along more context would be useful eventually.

Aside: What if the query vars were sent as-is without being pre-hashed into a `slug`? Having access to these query vars would be useful for caching plugins to determine what cache should be cleared as well. They would also be useful for debugging a URL Metric. I was hesitant to expose the query vars, thinking that perhaps there could be some private data located in there, but maybe not since these query vars would all be exposed if pretty permalinks aren't enabled? Maybe I'm being overly cautious. Probably something for another PR.

felixarntz: I think keeping the query vars hashed for now is best. I'm cautious of exposing more if we don't have to, so let's discuss that separately.

> I think your idea makes sense as a future enhancement to give caching plugins more say over how cache invalidation is handled. But for this first iteration, I think sending along more context would be useful eventually.

Which idea are you referring to? Based on my understanding, it was _your_ idea from the earlier comment to pass around only a post ID that needs to have its cache invalidated when metrics for the URL change, instead of passing around queried object type and queried object ID. And that makes sense to me, based on the idea that the logic would determine that post ID...
* either based on the queried object ID
* or if there's no queried object ID, the first post in the current loop

westonruter: Yes, that is my idea, as that is what I have found to be most reliable to invalidate caches for caching plugins. However some caching plugins, like Advanced Pantheon Page Cache, _would_ be able to handle a more targeted cache invalidation when the queried object is a term or author. But they'll also work when invalidating the cache for a post that appears in the loop on the archive page, so I think for now what is currently implementing in this PR is a good first step.

plugins/optimization-detective/detection.php
```diff
@@ -10,6 +10,54 @@
...
+ * which is better than nothing.
+ *
+ * There should not be any situation by default in which a page optimized with Optimization Detective does not have such
+ * a post available for cache purging. As seen in {@see od_can_optimize_response()}, when such a post ID is not
+ * available for cache purging then it returns false, as it also does in another case like if is_404().
+ *
+ * @since n.e.x.t
+ * @access private
+ *
+ * @return int|null Post ID or null if none found.
```
felixarntz, line 37: Nit-pick, but why not stick with `int` and return `0` if none found? `0` can never be an ID anyway and is false-y too.

plugins/optimization-detective/optimization.php
```diff
@@ -120,7 +120,10 @@ function od_can_optimize_response(): bool {
 		// users, additional elements will be present like the script from wp_customize_support_script() which will
 		// interfere with the XPath indices. Note that od_get_normalized_query_vars() is varied by is_user_logged_in()
 		// so membership sites and e-commerce sites will still be able to be optimized for their normal visitors.
-		current_user_can( 'customize' )
+		current_user_can( 'customize' ) ||
+		// Page caching plugins can only reliably be told to invalidate a cached page when a post is available to trigger
+		// the relevant actions on.
+		null !== od_get_cache_purge_post_id()
```
felixarntz, line 126: See above.

plugins/optimization-detective/storage/data.php
```diff
@@ -150,14 +150,15 @@ function od_get_url_metrics_slug( array $query_vars ): string {
...
- * @param string $url  URL.
- *
+ * @param string   $slug                Slug (hash of normalized query vars).
+ * @param string   $url                 URL.
+ * @param int|null $cache_purge_post_id Cache purge post ID.
  * @return string HMAC.
  */
-function od_get_url_metrics_storage_hmac( string $slug, string $url ): string {
-	$action = "store_url_metric:$slug:$url";
+function od_get_url_metrics_storage_hmac( string $slug, string $url, ?int $cache_purge_post_id = null ): string {
```
felixarntz, line 160: See above.

Also, should this maybe be a _required_ parameter? That would emphasize that it's needed to optimize responses anyway, as you already check above.

plugins/optimization-detective/storage/data.php
```diff
@@ -170,14 +171,14 @@ function od_get_url_metrics_storage_hmac( string $slug, string $url ): string {
...
- *
+ * @param string   $hmac                HMAC.
+ * @param string   $slug                Slug (hash of normalized query vars).
+ * @param String   $url                 URL.
+ * @param int|null $cache_purge_post_id Cache purge post ID.
  * @return bool Whether the HMAC is valid.
  */
-function od_verify_url_metrics_storage_hmac( string $hmac, string $slug, string $url ): bool {
-	return hash_equals( od_get_url_metrics_storage_hmac( $slug, $url ), $hmac );
+function od_verify_url_metrics_storage_hmac( string $hmac, string $slug, string $url, ?int $cache_purge_post_id = null ): bool {
```
felixarntz, line 180: See above. Maybe make this required?

plugins/optimization-detective/storage/rest-api.php
```diff
@@ -226,3 +242,49 @@ function od_handle_rest_request( WP_REST_Request $request ) {
...
+	do_action( 'clean_post_cache', $post->ID, $post );
+
+	/*
+	 * The transition_post_status action is used to flush page caches by:
+	 * - Jetpack Boost <https://github.com/Automattic/jetpack-boost-production/blob/4090a3f9414c2171cd52d8a397f00b0d1151475f/app/modules/optimizations/page-cache/pre-wordpress/Boost_Cache.php#L76>
+	 * - WP Super Cache <https://github.com/Automattic/wp-super-cache/blob/73b428d2fce397fd874b3056ad3120c343bc1a0c/wp-cache-phase2.php#L1616>
+	 * - LightSpeed Cache <https://github.com/litespeedtech/lscache_wp/blob/7c707469b3c88b4f45d9955593b92f9aeaed54c3/src/purge.cls.php#L68>
+	 */
+	/** This action is documented in wp-includes/post.php. */
+	do_action( 'transition_post_status', $post->post_status, $post->post_status, $post );
```
felixarntz, line 281: I wonder if this works well. I could see how caching plugins compare the before and after value, and this here would be a weird scenario that Core I believe never triggers (the same value in both).

Probably still better to keep this than not have it, but this may not work reliably.

felixarntz (approved):
@westonruter This looks great. Just left some minor feedback for your consideration.

plugins/optimization-detective/detection.php
westonruter: True. But `null` seems to be more universally semantic to indicate nothing was found. If `null` then you can check that case with either `is_null()` or `!is_int()`. And `null` is also false-y.

felixarntz: My counter-argument is that `int` is a single type, and the check can simply be made with `! $post_id` either way. Not a big deal either way.

plugins/optimization-detective/storage/rest-api.php
westonruter: Core does actually trigger when it is the same value for both. If you have published a post and then you update the post again, then this same `transition_post_status` action is triggered with the same `publish` value for both of those arguments. I just tested to confirm this.

And we need to keep this anyway because some caching plugins _only_ listen for this action to decide whether to invalidate their caches.

plugins/optimization-detective/detection.php
westonruter: Except not with our strict PHPStan rules in place 😄

plugins/optimization-detective/storage/data.php
westonruter: > Also, should this maybe be a _required_ parameter? That would emphasize that it's needed to optimize responses anyway, as you already check above.

The check I added means that _by default_ it will not offer to optimize responses when `od_get_cache_purge_post_id()` returns `null`. But this can be overridden with the `od_can_optimize_response` filter. So we're not guaranteed that it will always be not-`null`.

felixarntz: What would be a use-case for that? I'm thinking it might be better to _not_ make that possible. For example, we could make it so that the filter only is applied when there is a post ID. That feels more robust to me.

westonruter: Sometimes I've made sites where I use author archive pages to create team bio pages, for example. In such a case, there often isn't any blog post actually authored by that user. Nevertheless, if I go to their author page it doesn't actually return a 404. It serves a page with the author's name in the document title and if the theme has an `author.php` template which is configured to display the user bio even when there are no posts, then it will work as intended. For example, the Twenty Ten theme shows the user bio on `author.php` even when there are no posts by that author:

image

plugins/optimization-detective/optimization.php
westonruter, line 126: This was wrong!!! It should have been:

```php
null === od_get_cache_purge_post_id()
```

See https://github.com/WordPress/performance/pull/1659

PR #1642, opened 2024-11-11
Introduce `webp_uploads_get_file_mime_type` helper function

mukeshpanchal27:
## Summary

Follow-up work per https://github.com/WordPress/performance/pull/1635#pullrequestreview-2421429667

### DIFF
--- a/plugins/webp-uploads/helper.php
+++ b/plugins/webp-uploads/helper.php
@@ -483,7 +483,7 @@ function webp_uploads_get_mime_type_image( int $attachment_id, string $src, stri
  * @param int    $attachment_id The attachment ID.
  * @return string The MIME type of the file, or false if not found.
  */
-function get_file_mime_type( string $file, int $attachment_id ): string {
+function webp_uploads_get_file_mime_type( string $file, int $attachment_id ): string {
 	/*
 	 * We need to get the MIME type ideally from the file, as WordPress Core may have already altered it.
 	 * The post MIME type is typically not updated during that process.
--- a/plugins/webp-uploads/helper.php
+++ b/plugins/webp-uploads/helper.php
@@ -468,3 +468,27 @@ function webp_uploads_get_mime_type_image( int $attachment_id, string $src, stri
 
 	return null;
 }
+
+/**
+ * Retrieves the MIME type of a file, checking the file directly if possible,
+ * and falling back to the attachment's MIME type if needed.
+ *
+ * The function attempts to determine the MIME type directly from the file.
+ * If that information is unavailable, it uses the MIME type from the attachment metadata.
+ * If neither is available, it defaults to an empty string.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $file          The path to the file.
+ * @param int    $attachment_id The attachment ID.
+ * @return string The MIME type of the file, or false if not found.
+ */
+function get_file_mime_type( string $file, int $attachment_id ): string {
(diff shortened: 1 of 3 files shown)


plugins/webp-uploads/helper.php
```diff
@@ -468,3 +468,27 @@ function webp_uploads_get_mime_type_image( int $attachment_id, string $src, stri
...
+ * If that information is unavailable, it uses the MIME type from the attachment metadata.
+ * If neither is available, it defaults to an empty string.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $file          The path to the file.
+ * @param int    $attachment_id The attachment ID.
+ * @return string The MIME type of the file, or false if not found.
+ */
+function get_file_mime_type( string $file, int $attachment_id ): string {
```
swissspidy: The function needs to be prefixed with `webp_uploads_`

mukeshpanchal27: Completely missed 😄

plugins/webp-uploads/helper.php
```diff
@@ -468,3 +468,27 @@ function webp_uploads_get_mime_type_image( int $attachment_id, string $src, stri
...
+ *
+ * The function attempts to determine the MIME type directly from the file.
+ * If that information is unavailable, it uses the MIME type from the attachment metadata.
+ * If neither is available, it defaults to an empty string.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $file          The path to the file.
+ * @param int    $attachment_id The attachment ID.
+ * @return string The MIME type of the file, or false if not found.
```
```suggestion
 * @return string The MIME type of the file, or an empty string if not found.
```

westonruter (approved):
LGTM, aside from a tiny docs fix.

adamsilverstein (approved):
Looks good!

plugins/webp-uploads/helper.php
```diff
@@ -468,3 +468,27 @@ function webp_uploads_get_mime_type_image( int $attachment_id, string $src, stri
...
+ * If that information is unavailable, it uses the MIME type from the attachment metadata.
+ * If neither is available, it defaults to an empty string.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $file          The path to the file.
+ * @param int    $attachment_id The attachment ID.
+ * @return string The MIME type of the file, or an empty string if not found.
+ */
+function webp_uploads_get_file_mime_type( string $file, int $attachment_id ): string {
```
felixarntz, line 486: Maybe rename this to `webp_uploads_get_attachment_file_mime_type()`? This clarifies the function is intended for attachments, not _any_ files.

Another idea: Maybe make `$attachment_id` the first parameter, and then `$file` as second parameter could be optional? If it is not passed, we could get the file ourselves by calling `get_attached_file()` in the function, so that would be a good default.

It would allow us to simplify code in places where we need to get the file only for this check, while still being efficient for other places where we need to use the file for other purposes as well.

For example, in `webp_uploads_wrap_image_in_picture()` we get the file name only for this purpose, which wouldn't require an extra call anymore with the suggested change.

felixarntz: @mukeshpanchal27 I guess I was a bit too late with this, but LMK your thoughts.

mukeshpanchal27: Thanks for the feedback. Make sense. Open #1662 for review and merge as it's quick one.

PR #1643, opened 2024-11-11
Add unminified source JS files to builds

### LINKED ISSUE #1493: Unminified source JS files and source maps are not distributed with builds
I just realized that when doing a production build of a plugin, the JavaScript is getting minified (which is good) but there is no unminified version which is also being distributed. When `SCRIPT_DEBUG` is enabled, the unminified dev version of scripts should be used instead. Additionally, when using the minified version it would be great if a source map could connect it to the unminified version.

As it stands, the only JS distributed with Optimization Detective is the minified versions: 

- https://plugins.trac.wordpress.org/browser/optimization-detective/tags/0.5.0/detect.js
- https://plugins.trac.wordpress.org/browser/optimization-detective/tags/0.5.0/build/web-vitals.js

ShyamGadde:
## Summary

Fixes #1493 

## Relevant technical choices

- The unminified version of `web-vitals.js` is not included because the file is copied directly from the pre-minified version in `node_modules`.
- Source maps are excluded, as the unminified development version of the script will be loaded when `SCRIPT_DEBUG` is enabled.

### DIFF
--- a/plugins/optimization-detective/detection.php
+++ b/plugins/optimization-detective/detection.php
@@ -76,15 +76,10 @@ function od_get_detection_script( string $slug, OD_URL_Metric_Group_Collection $
 		$detect_args['urlMetricGroupCollection'] = $group_collection;
 	}
 
-	$detection_script_file = 'detect.min.js';
-	if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) {
-		$detection_script_file = 'detect.js';
-	}
-
 	return wp_get_inline_script_tag(
 		sprintf(
 			'import detect from %s; detect( %s );',
-			wp_json_encode( add_query_arg( 'ver', OPTIMIZATION_DETECTIVE_VERSION, plugin_dir_url( __FILE__ ) . $detection_script_file ) ),
+			wp_json_encode( add_query_arg( 'ver', OPTIMIZATION_DETECTIVE_VERSION, plugin_dir_url( __FILE__ ) . sprintf( 'detect%s.js', wp_scripts_get_suffix() ) ) ),
 			wp_json_encode( $detect_args )
 		),
 		array( 'type' => 'module' )
--- a/plugins/optimization-detective/detection.php
+++ b/plugins/optimization-detective/detection.php
@@ -76,10 +76,15 @@ function od_get_detection_script( string $slug, OD_URL_Metric_Group_Collection $
 		$detect_args['urlMetricGroupCollection'] = $group_collection;
 	}
 
+	$detection_script_file = 'detect.min.js';
+	if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) {
+		$detection_script_file = 'detect.js';
+	}
+
 	return wp_get_inline_script_tag(
 		sprintf(
 			'import detect from %s; detect( %s );',
-			wp_json_encode( add_query_arg( 'ver', OPTIMIZATION_DETECTIVE_VERSION, plugin_dir_url( __FILE__ ) . 'detect.js' ) ),
+			wp_json_encode( add_query_arg( 'ver', OPTIMIZATION_DETECTIVE_VERSION, plugin_dir_url( __FILE__ ) . $detection_script_file ) ),
 			wp_json_encode( $detect_args )
 		),
 		array( 'type' => 'module' )
--- a/plugins/web-worker-offloading/hooks.php
+++ b/plugins/web-worker-offloading/hooks.php
@@ -21,7 +21,12 @@ if ( ! defined( 'ABSPATH' ) ) {
 function plwwo_register_default_scripts( WP_Scripts $scripts ): void {
 	// The source code for partytown.js is built from <https://github.com/BuilderIO/partytown/blob/b292a14047a0c12ca05ba97df1833935d42fdb66/src/lib/main/snippet.ts>.
 	// See webpack config in the WordPress/performance repo: <https://github.com/WordPress/performance/blob/282a068f3eb2575d37aeb9034e894e7140fcddca/webpack.config.js#L84-L130>.
-	$partytown_js = file_get_contents( __DIR__ . '/build/partytown.js' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents -- It's a local filesystem path not a remote request.
+	$partytown_js_path = '/build/partytown.js';
+	if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) {
+		$partytown_js_path = '/build/debug/partytown.js';
+	}
+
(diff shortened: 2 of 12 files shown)


plugins/optimization-detective/detection.php
```diff
@@ -76,10 +76,15 @@ static function ( OD_URL_Metric_Group $group ): array {
 		$detect_args['urlMetricGroupCollection'] = $group_collection;
 	}
 
+	$detection_script_file = 'detect.min.js';
+	if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) {
+		$detection_script_file = 'detect.js';
+	}
```
westonruter: We should be able to reuse `wp_scripts_get_suffix()` here as seen  in `wp_default_scripts()` in core:

```suggestion
	$detection_script_file = sprintf( 'detect%s.js', wp_scripts_get_suffix() );
```

westonruter (changes requested):

westonruter:
> * The unminified version of `web-vitals.js` is not included because the file is copied directly from the pre-minified version in `node_modules`.

I was thinking alternatively that we could load the `web-vitals` file from `src` instead of `dist` when `SCRIPT_DEBUG` is enabled _and_ there is a local `node_modules` directory. But this isn't feasible either because the source files are in TypeScript, so this would require transpilation. Probably doesn't make sense.

westonruter:
Oh, there are also now 3 other JS files that need to have their minified version included conditionally:

1. `plugins/embed-optimizer/detect.js`
2. `plugins/embed-optimizer/lazy-load.js`
3. `plugins/image-prioritizer/lazy-load.js`

webpack.config.js
```diff
@@ -164,6 +168,7 @@ const buildPlugin = ( env ) => {
 					{
 						from,
 						to,
+						info: { minimized: true },
```
westonruter, line 248: This won't attempt to create a minified version of `web-vitals.js`, will it?

ShyamGadde: No, actually it's the opposite - setting `info: { minimized: true }` tells Webpack to skip running these files through the minimizer. It's a way to tell Webpack to copy the files 'as is' without trying to minimize them. The documentation specifically mentions this is 'Useful if you need to simply copy *.js files to destination "as is" without evaluating and minimizing them using Terser.'

ref: https://webpack.js.org/plugins/copy-webpack-plugin/#skip-running-javascript-files-through-a-minimizer

webpack.config.js
```diff
@@ -61,16 +61,20 @@ const optimizationDetective = ( env ) => {
 				patterns: [
 					{
 						from: `${ source }/dist/web-vitals.js`,
-						to: `${ destination }/web-vitals.js`,
+						to: `${ destination }/build/web-vitals.js`,
```
westonruter, line 139: Is this missing the flag to say it is minified?

```suggestion
						to: `${ destination }/build/web-vitals.js`,
						info: { minimized: true }`,
```

webpack.config.js
westonruter: OK, thank you. I'm a little confused. We want to create minified versions of all the scripts, except for those which are copied from `node_modules`, right? So we'd want to copy from `node_modules/web-vitals/dist` and `node_modules/@builder.io/partytown` as pre-minified, but everything else we'd want to create minified versions for. I don't see how that is all configured here. (I'm a webpack novice!)

webpack.config.js
ShyamGadde: Yes! I just realized this. I should probably make this change in the config for Web Worker Offloading as well, which is copying the partytown files.

webpack.config.js
ShyamGadde: Let me clarify the current state and what needs to be fixed:

1. For files from `node_modules` (`web-vitals` and `partytown`):
   - These come pre-minified, but when copying them to the `build` directory for their respective plugins, we're not telling Webpack that. Currently, I have configured minimization skipping only for the production build.
   - Need to add `info: { minimized: true }` to prevent re-minification attempts

2. For plugin JS files that need minified versions:
   - Minified version for `optimization-detective/detect.js` is being created (`detect.min.js`)
   - Need to add minified versions for (I missed these as they were not initially part of the issue description):
     - `embed-optimizer/detect.js`
     - `embed-optimizer/lazy-load.js`
     - `image-prioritizer/lazy-load.js`

I'll update this PR to handle both these cases. Would that address your concerns?

westonruter: Yes, I believe so!

ShyamGadde:
I wanted to bring up something I noticed regarding the partytown package used in the web worker offloading plugin:

* The partytown package does provide unminified files under `node_modules/@builder.io/partytown/lib/debug`
* In the initial webpack config, these were being copied but also getting minified
* Now that we're skipping/disabling minification, we could potentially use these unminified script files when `SCRIPT_DEBUG` is enabled

Since the partytown code is inlined as a script tag, should I modify this plugin to use the unminified partytown JavaScript when `SCRIPT_DEBUG` is enabled? Or would it be better to stick with the current approach of using the pre-minified files?

westonruter:
@ShyamGadde definitely! The non-minified debug version should be used here when `SCRIPT_DEBUG` is enabled:

https://github.com/WordPress/performance/blob/b20b43ffee6412d6f3f6ec2bc824eccfea6d5c58/plugins/web-worker-offloading/hooks.php#L24

westonruter:
And good catch on the config erroneously minifying the debug scripts:

https://plugins.trac.wordpress.org/browser/web-worker-offloading/tags/0.1.1/build/partytown-atomics.js
https://plugins.trac.wordpress.org/browser/web-worker-offloading/tags/0.1.1/build/debug/partytown-atomics.js

The second one shouldn't have been minified!

plugins/web-worker-offloading/hooks.php
```diff
@@ -21,7 +21,12 @@
 function plwwo_register_default_scripts( WP_Scripts $scripts ): void {
 	// The source code for partytown.js is built from <https://github.com/BuilderIO/partytown/blob/b292a14047a0c12ca05ba97df1833935d42fdb66/src/lib/main/snippet.ts>.
 	// See webpack config in the WordPress/performance repo: <https://github.com/WordPress/performance/blob/282a068f3eb2575d37aeb9034e894e7140fcddca/webpack.config.js#L84-L130>.
-	$partytown_js = file_get_contents( __DIR__ . '/build/partytown.js' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents -- It's a local filesystem path not a remote request.
+	$partytown_js_path = '/build/partytown.js';
+	if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) {
+		$partytown_js_path = '/build/debug/partytown.js';
+	}
```
westonruter: Nit:

```suggestion
	if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) {
		$partytown_js_path = '/build/debug/partytown.js';
	} else {
		$partytown_js_path = '/build/partytown.js';
	}
```

westonruter (approved):
Great work! It's everything I ever dreamed for!

PR #1644, opened 2024-11-11
Install and activate Optimization Detective when the Embed Optimizer feature is activated from the Performance screen

### LINKED ISSUE #1615: Optimization Detective should be automatically installed and activated when activating Embed Optimizer
The Embed Optimizer plugin works best Optimization Detective is also running. In fact, there are notices on the plugin list table to encourage users to also install Optimization Detective:

Image

However, when activating the Embed Optimizer plugin there is no advice or ability to install Optimization Detective:

Image

This being the case, when activating the Embed Optimizer plugin, we should automatically install Optimization Detective as a soft dependency. 

When Optimization Detective is active alongside Embed Optimizer, the relationship between the two is already indicated in the plugin metadata for Optimization Detective:

Image

So users should be able to see why it was installed.

westonruter:
Fixes #1615

Most of the optimizations provided by Embed Optimizer require the Optimization Detective plugin. So this PR will automatically install and activate Optimization Detective when the Embed Optimizer feature is activated from the Performance screen.

Screen recording:

[Screen recording 2024-11-11 09.03.26.webm](https://github.com/user-attachments/assets/4a3b40f5-6fb9-457c-8e1b-533539a9d947)

Build for testing: [performance-lab.zip](https://github.com/user-attachments/files/17706319/performance-lab.zip)

### DIFF
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -325,12 +325,8 @@ function perflab_install_and_activate_plugin( string $plugin_slug, array &$proce
 	}
 
 	// Add recommended plugins (soft dependencies) to the list of plugins installed and activated.
-	$standalone_plugin_data = perflab_get_standalone_plugin_data();
-	if ( isset( $standalone_plugin_data[ $plugin_slug ]['suggest'] ) ) {
-		$plugin_data['requires_plugins'] = array_merge(
-			$plugin_data['requires_plugins'],
-			$standalone_plugin_data[ $plugin_slug ]['suggest']
-		);
+	if ( 'embed-optimizer' === $plugin_slug ) {
+		$plugin_data['requires_plugins'][] = 'optimization-detective';
 	}
 
 	// Install and activate plugin dependencies first.
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -325,8 +325,12 @@ function perflab_install_and_activate_plugin( string $plugin_slug, array &$proce
 	}
 
 	// Add recommended plugins (soft dependencies) to the list of plugins installed and activated.
-	if ( 'embed-optimizer' === $plugin_slug ) {
-		$plugin_data['requires_plugins'][] = 'optimization-detective';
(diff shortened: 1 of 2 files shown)


plugins/performance-lab/load.php
```diff
@@ -83,14 +83,15 @@ function perflab_render_generator(): void {
  *
  * @since 3.0.0
  *
- * @return array<string, array{'constant': string, 'experimental'?: bool}> Associative array of $plugin_slug => $plugin_data pairs.
+ * @return array<string, array{'constant': string, 'experimental'?: bool, 'suggest'?: string[]}> Associative array of $plugin_slug => $plugin_data pairs.
```
felixarntz: I'm not sure we want to establish this as something we may want to do more, given that this is not a thing that WordPress Core supports.

Therefore instead of introducing a "schema" change here, I think we should for now hard-code this in `perflab_install_and_activate_plugin()`. Feels more hacky, but is more future-proof as it doesn't change any portion of the API. Given that this PR simply takes the "suggested" feature as if it was "required", that part would need more thought if we truly wanted to establish a way to _suggest_ a feature.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -324,6 +324,15 @@ function perflab_install_and_activate_plugin( string $plugin_slug, array &$proce
...
 	}
 
+	// Add recommended plugins (soft dependencies) to the list of plugins installed and activated.
+	$standalone_plugin_data = perflab_get_standalone_plugin_data();
+	if ( isset( $standalone_plugin_data[ $plugin_slug ]['suggest'] ) ) {
+		$plugin_data['requires_plugins'] = array_merge(
+			$plugin_data['requires_plugins'],
+			$standalone_plugin_data[ $plugin_slug ]['suggest']
+		);
+	}
```
felixarntz: See my other comment: For those reasons, I think we should simply hard-code here: If `embed-optimizer`, add `optimization-detective` to `requires_plugins`.

felixarntz (changes requested):
@westonruter While functionally this is a good call, I think we should not introduce a "half-baked" API for something like plugin suggestions which would need more thought.

felixarntz (approved):
Thanks @westonruter!

PR #1646, opened 2024-11-11
Use AJAX for activating features / plugins in Performance Lab

### LINKED ISSUE #1583: Use AJAX for activating features / plugins in Performance Lab
## Feature Description

This is part of the improving the onboarding experience for Performance Lab (#1032).

A problem today is that activating a feature causes a full page reload. This makes it tedious to quickly activate multiple features, since you can't activate multiple at a time. It can even lead to errors, if users manage to click another feature activation button before the page reloads due to already having clicked another one.

We should implement the feature activation logic in a WP AJAX handler callback and adjust the JavaScript logic to use that.

## Technical notes

* The AJAX callback should use the `perflab_install_and_activate_plugin()` function, which already has this logic abstracted away from the concrete technical use-case.
* The existing `admin_action_*` callback should remain in place, and the buttons should still link to the relevant URL, as it allows the functionality to still work, should the administrator have JavaScript disabled or in case there is a bug in the JavaScript code.
    * While this is not very crucial to support in today's time, there's very little overhead in maintaining the extra code, particularly because the bulk of the logic is already abstracted away to be reusable via `perflab_install_and_activate_plugin()`.
* In other words: The AJAX based activation should be implemented as a progressive enhancement.

b1ink0:
## Summary

Fixes [#1583](https://github.com/WordPress/performance/issues/1583)

## Relevant technical choices

1. Frontend / UX
    - Features / plugins in Performance Lab are now activated using AJAX improving user experience.
    - No page refresh when activating features / plugins.
    - Multiple plugin can be activated as each plugin activation sends different AJAX request.
    - The click event is added to the anchor link as if JavaScript is enabled then the default behaviour is prevented so it functions as a button else It behaves as a link.

2. Backend
    - Removed `perflab_print_plugin_progress_indicator_script` function as it was previously used for adding a progress indicator to the activate button on click. This functionality is now moved to the separate JavaScript file which handles the activation of plugin.
    - Added `perflab-plugin-activate-ajax` (script handle) script which handles activation of features / plugins using AJAX.
    - Updated `perflab_render_plugin_card` function to add data attribute `data-plugin-slug` in plugin activate link.
    - Added new REST API endpoints
        - `performance-lab/v1/features/(?P<slug>[a-z0-9_-]+):activate`
        - `performance-lab/v1/features/(?P<slug>[a-z0-9_-]+)`
  
3. Unable to Implement fully
   - Regarding the setting URL which is displayed in the admin notices and the plugin card.
   - I have implemented the settings URL part but it only works for the `Modern Image Formats` plugin.
   - Context: Currently only two plugins provide the settings URL  `Modern Image Formats` and `Speculative Loading`.
   - It does not work with the `Speculative Loading` because of how this plugin is loaded.
       - This plugin gets loaded ( loaded in this context is when are its helper files are loaded like `settings.php` ) on `after_setup_theme` hook.
       - But when the AJAX callback is triggered this hook has already been executed. As the plugin is activated later and which requries the `after_setup_theme` hook to fully load the plugin which includes the settings URL.
       - The function which is executed in the `after_setup_theme` calls a closure function which is stored in the global variable ( `plsr_pending_plugin_info` for the `Speculative Loading` plugin ) this closure function fully loads the plugin.
   - I tried and tested this could be fixed by calling the closure function in AJAX callback but we will be hardcoding the global variable name and will have to update the callback function if other plugins also add settings URL.
 
 4. Demo

https://github.com/user-attachments/assets/da3858aa-5359-4077-8015-b59b73c922a4

### DIFF
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -228,7 +228,7 @@ function perflab_enqueue_features_page_scripts(): void {
 	// Enqueue plugin activate AJAX script and localize script data.
 	wp_enqueue_script(
 		'perflab-plugin-activate-ajax',
-		plugins_url( 'includes/admin/plugin-activate-ajax.js', PERFLAB_MAIN_FILE ),
+		plugin_dir_url( PERFLAB_MAIN_FILE ) . 'includes/admin/plugin-activate-ajax.js',
 		array( 'wp-i18n', 'wp-a11y', 'wp-api-fetch' ),
 		PERFLAB_VERSION,
 		true
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -225,9 +225,6 @@ function perflab_enqueue_features_page_scripts(): void {
 	wp_enqueue_style( 'thickbox' );
 	wp_enqueue_script( 'plugin-install' );
 
-	// Enqueue the a11y script.
-	wp_enqueue_script( 'wp-a11y' );
-
 	// Enqueue plugin activate AJAX script and localize script data.
 	wp_enqueue_script(
 		'perflab-plugin-activate-ajax',
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -232,7 +232,7 @@ function perflab_enqueue_features_page_scripts(): void {
 	wp_enqueue_script(
 		'perflab-plugin-activate-ajax',
 		plugins_url( 'includes/admin/plugin-activate-ajax.js', PERFLAB_MAIN_FILE ),
-		array( 'jquery', 'wp-i18n', 'wp-a11y' ),
+		array( 'wp-i18n', 'wp-a11y' ),
 		PERFLAB_VERSION,
 		true
 	);
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -444,17 +444,18 @@ function perflab_print_plugin_progress_indicator_script(): void {
 	$nonce = wp_create_nonce( 'perflab_install_activate_plugin_ajax' );
 
 	$script_data = array(
-		'ajaxUrl'        => admin_url( 'admin-ajax.php' ),
-		'nonce'          => $nonce,
-		'activatingText' => __( 'Activating...', 'performance-lab' ),
-		'activateText'   => __( 'Activate', 'performance-lab' ),
-		'activatedText'  => __( 'Active', 'performance-lab' ),
-		'dismissText'    => __( 'Dismiss this notice.', 'performance-lab' ),
-		'successText'    => __( 'Feature activated.', 'performance-lab' ),
-		'reviewText'     => __( ' Review ', 'performance-lab' ),
-		'settingsText'   => __( 'settings', 'performance-lab' ),
-		'endText'        => _x( '.', 'Punctuation mark', 'performance-lab' ),
-		'errorText'      => __( 'There was an error activating the plugin. Please try again.', 'performance-lab' ),
+		'ajaxUrl'               => admin_url( 'admin-ajax.php' ),
+		'nonce'                 => $nonce,
+		'activatingText'        => __( 'Activating...', 'performance-lab' ),
+		'activateText'          => __( 'Activate', 'performance-lab' ),
+		'activatedText'         => __( 'Active', 'performance-lab' ),
+		'dismissText'           => __( 'Dismiss this notice.', 'performance-lab' ),
+		'successText'           => __( 'Feature activated.', 'performance-lab' ),
+		'reviewText'            => __( ' Review ', 'performance-lab' ),
+		'settingsPrimaryText'   => __( 'settings', 'performance-lab' ),
+		'settingsSecondaryText' => __( 'Settings', 'performance-lab' ),
+		'endText'               => _x( '.', 'Punctuation mark', 'performance-lab' ),
+		'errorText'             => __( 'There was an error activating the plugin. Please try again.', 'performance-lab' ),
 	);
 
 	$js_function = <<<JS
@@ -475,7 +476,7 @@ function perflab_print_plugin_progress_indicator_script(): void {
 
 						wp.a11y.speak( data.activatingText );
 
-						const pluginSlug = target.getAttribute( 'data-plugin-slug' );
+						const pluginSlug = target.getAttribute( 'data-plugin-slug' ).trim();
 
 						try {
 							const response = await fetch( data.ajaxUrl, {
@@ -510,8 +511,23 @@ function perflab_print_plugin_progress_indicator_script(): void {
 							newButton.textContent = data.activatedText;
 
 							target.parentNode.replaceChild( newButton, target );
-							console.log(responseData)
-							showAdminNotice( data.successText, 'success', responseData?.data?.pluginSettingsURL );
+
+							const pluginSettingsURL = responseData?.data?.pluginSettingsURL;
+
+							const actionButtonList = document.querySelector( '.plugin-card-' + pluginSlug + ' .plugin-action-buttons' )
+
+							if ( pluginSettingsURL && actionButtonList ) {
+								const listItem = document.createElement( 'li' );
+								const anchor = document.createElement( 'a' );
+
+								anchor.setAttribute( 'href', pluginSettingsURL );
+								anchor.textContent = data.settingsSecondaryText;
+
+								listItem.appendChild( anchor );
+								actionButtonList.appendChild( listItem );
+							}
+							
+							showAdminNotice( data.successText, 'success', pluginSettingsURL );
 						} catch (error) {
 							showAdminNotice( data.errorText );
 
@@ -535,7 +551,7 @@ function perflab_print_plugin_progress_indicator_script(): void {
 					
 					const anchor = document.createElement( 'a' );
 					anchor.setAttribute( 'href', pluginSettingsURL );
-					anchor.textContent = data.settingsText;
+					anchor.textContent = data.settingsPrimaryText;
 					
 					para.appendChild( anchor );
 					para.appendChild( document.createTextNode( data.endText ) );
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -310,6 +310,8 @@ add_action( 'admin_action_perflab_install_activate_plugin', 'perflab_install_act
 
 /**
  * Callback for handling installation/activation of plugin using ajax.
+ *
+ * @since n.e.x.t
  */
 function perflab_install_activate_plugin_ajax_callback(): void {
 	check_ajax_referer( 'perflab_install_activate_plugin_ajax', '_ajax_nonce' );
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -231,9 +231,9 @@ function perflab_enqueue_features_page_scripts(): void {
 	// Enqueue plugin activate AJAX script and localize script data.
 	wp_enqueue_script(
 		'perflab-plugin-activate-ajax',
-		PERFLAB_PLUGIN_DIR_URL . 'includes/admin/plugin-activate-ajax.js',
+		plugins_url( 'includes/admin/plugin-activate-ajax.js', PERFLAB_MAIN_FILE ),
 		array( 'jquery', 'wp-i18n', 'wp-a11y' ),
-		(string) filemtime( PERFLAB_PLUGIN_DIR_PATH . 'includes/admin/plugin-activate-ajax.js' ),
+		PERFLAB_VERSION,
 		true
 	);
 
@@ -337,7 +337,7 @@ function perflab_install_activate_plugin_ajax_callback(): void {
 	if ( null !== $plugin_settings_url ) {
 		wp_send_json_success(
 			array(
-				'pluginSettingsURL' => esc_url( $plugin_settings_url ),
+				'pluginSettingsURL' => esc_url_raw( $plugin_settings_url ),
 			)
 		);
 	}
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -229,18 +229,9 @@ function perflab_enqueue_features_page_scripts(): void {
 	wp_enqueue_script(
 		'perflab-plugin-activate-ajax',
 		plugins_url( 'includes/admin/plugin-activate-ajax.js', PERFLAB_MAIN_FILE ),
-		array( 'wp-i18n', 'wp-a11y' ),
+		array( 'wp-i18n', 'wp-a11y', 'wp-api-fetch' ),
 		PERFLAB_VERSION,
 		true
-	);
-
-	wp_localize_script(
-		'perflab-plugin-activate-ajax',
-		'perflabPluginActivateAjaxData',
-		array(
-			'ajaxUrl' => admin_url( 'admin-ajax.php' ),
-			'nonce'   => wp_create_nonce( 'perflab_install_activate_plugin_ajax' ),
-		)
 	);
 }
 
@@ -301,47 +292,6 @@ function perflab_install_activate_plugin_callback(): void {
 	}
 }
 add_action( 'admin_action_perflab_install_activate_plugin', 'perflab_install_activate_plugin_callback' );
-
-/**
- * Callback for handling installation/activation of plugin using ajax.
- *
- * @since n.e.x.t
- */
-function perflab_install_activate_plugin_ajax_callback(): void {
-	check_ajax_referer( 'perflab_install_activate_plugin_ajax', '_ajax_nonce' );
-
-	require_once ABSPATH . 'wp-admin/includes/plugin.php';
-	require_once ABSPATH . 'wp-admin/includes/plugin-install.php';
-	require_once ABSPATH . 'wp-admin/includes/class-wp-upgrader.php';
-	require_once ABSPATH . 'wp-admin/includes/class-wp-ajax-upgrader-skin.php';
-
-	if ( ! isset( $_POST['slug'] ) ) {
-		wp_send_json_error( __( 'Missing required parameter.', 'performance-lab' ), 400 );
-	}
-
-	$plugin_slug = perflab_sanitize_plugin_slug( wp_unslash( $_POST['slug'] ) );
-	if ( null === $plugin_slug ) {
-		wp_send_json_error( __( 'Invalid plugin.', 'performance-lab' ), 400 );
-	}
-
-	// Install and activate the plugin and its dependencies.
-	$result = perflab_install_and_activate_plugin( $plugin_slug );
-	if ( $result instanceof WP_Error ) {
-		wp_send_json_error( $result->get_error_message() );
-	}
-
-	$plugin_settings_url = perflab_get_plugin_settings_url( $plugin_slug );
-	if ( null !== $plugin_settings_url ) {
-		wp_send_json_success(
-			array(
-				'pluginSettingsURL' => esc_url_raw( $plugin_settings_url ),
-			)
-		);
-	}
-
-	wp_send_json_success();
-}
-add_action( 'wp_ajax_perflab_install_activate_plugin', 'perflab_install_activate_plugin_ajax_callback' );
 
 /**
  * Callback function to handle admin inline style.
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -39,6 +39,8 @@ add_action( 'admin_menu', 'perflab_add_features_page' );
  * @since 1.0.0
  * @since 3.0.0 Renamed to perflab_load_features_page(), and the
  *              $module and $hook_suffix parameters were removed.
+ * @since n.e.x.t Removed added action for perflab_print_plugin_progress_indicator_script
+ *                as this function is removed.
  */
 function perflab_load_features_page(): void {
 	// Handle script enqueuing for settings page.
@@ -49,9 +51,6 @@ function perflab_load_features_page(): void {
 
 	// Handle style for settings page.
 	add_action( 'admin_head', 'perflab_print_features_page_style' );
-
-	// Handle script for settings page.
-	add_action( 'admin_footer', 'perflab_print_plugin_progress_indicator_script' );
 }
 
 /**
@@ -221,6 +220,7 @@ add_action( 'wp_ajax_dismiss-wp-pointer', 'perflab_dismiss_wp_pointer_wrapper', 
  *
  * @since 2.8.0
  * @since 3.0.0 Renamed to perflab_enqueue_features_page_scripts().
+ * @since n.e.x.t Added enqueue plugin activate AJAX script.
  */
 function perflab_enqueue_features_page_scripts(): void {
 	// These assets are needed for the "Learn more" popover.
@@ -230,6 +230,24 @@ function perflab_enqueue_features_page_scripts(): void {
 
 	// Enqueue the a11y script.
 	wp_enqueue_script( 'wp-a11y' );
+
+	// Enqueue plugin activate AJAX script and localize script data.
+	wp_enqueue_script(
+		'perflab-plugin-activate-ajax',
+		PERFLAB_PLUGIN_DIR_URL . 'includes/admin/plugin-activate-ajax.js',
+		array( 'jquery', 'wp-i18n', 'wp-a11y' ),
+		(string) filemtime( PERFLAB_PLUGIN_DIR_PATH . 'includes/admin/plugin-activate-ajax.js' ),
+		true
+	);
+
+	wp_localize_script(
+		'perflab-plugin-activate-ajax',
+		'perflabPluginActivateAjaxData',
+		array(
+			'ajaxUrl' => admin_url( 'admin-ajax.php' ),
+			'nonce'   => wp_create_nonce( 'perflab_install_activate_plugin_ajax' ),
+		)
+	);
 }
 
 /**
@@ -433,181 +451,6 @@ function perflab_plugin_admin_notices(): void {
 			)
 		);
 	}
-}
-
-/**
- * Callback function that print plugin progress indicator script.
- *
- * @since 3.1.0
- */
-function perflab_print_plugin_progress_indicator_script(): void {
-	$nonce = wp_create_nonce( 'perflab_install_activate_plugin_ajax' );
-
-	$script_data = array(
-		'ajaxUrl'               => admin_url( 'admin-ajax.php' ),
-		'nonce'                 => $nonce,
-		'activatingText'        => __( 'Activating...', 'performance-lab' ),
-		'activateText'          => __( 'Activate', 'performance-lab' ),
-		'activatedText'         => __( 'Active', 'performance-lab' ),
-		'dismissText'           => __( 'Dismiss this notice.', 'performance-lab' ),
-		'successText'           => __( 'Feature activated.', 'performance-lab' ),
-		'reviewText'            => __( ' Review ', 'performance-lab' ),
-		'settingsPrimaryText'   => __( 'settings', 'performance-lab' ),
-		'settingsSecondaryText' => __( 'Settings', 'performance-lab' ),
-		'endText'               => _x( '.', 'Punctuation mark', 'performance-lab' ),
-		'errorText'             => __( 'There was an error activating the plugin. Please try again.', 'performance-lab' ),
-	);
-
-	$js_function = <<<JS
-		function addPluginProgressIndicator( data ) {
-			document.addEventListener( 'DOMContentLoaded', function () {
-				document.addEventListener( 'click', async function ( event ) {
-					if (
-						event.target.classList.contains(
-							'perflab-install-active-plugin'
-						)
-					) {
-						// Prevent the default link behavior.
-						event.preventDefault();
-
-						const target = event.target;
-						target.classList.add( 'updating-message' );
-						target.textContent = data.activatingText;
-
-						wp.a11y.speak( data.activatingText );
-
-						const pluginSlug = target.getAttribute( 'data-plugin-slug' ).trim();
-
-						try {
-							const response = await fetch( data.ajaxUrl, {
-								method: 'POST',
-								credentials: 'same-origin',
-								headers: {
-									'Content-Type': 'application/x-www-form-urlencoded',
-								},
-								body: new URLSearchParams( {
-									action: 'perflab_install_activate_plugin',
-									slug: pluginSlug,
-									_ajax_nonce: data.nonce,
-								} ),
-							} );
-
-							const responseData = await response.json();
-
-							if ( !responseData.success ) {
-								showAdminNotice( data.errorText );
-
-								target.classList.remove('updating-message');
-								target.textContent = data.activateText;
-
-								return;
-							}
-
-							const newButton = document.createElement( 'button' );
-
-							newButton.type = 'button';
-							newButton.className = 'button button-disabled';
-							newButton.disabled = true;
-							newButton.textContent = data.activatedText;
-
-							target.parentNode.replaceChild( newButton, target );
-
-							const pluginSettingsURL = responseData?.data?.pluginSettingsURL;
-
-							const actionButtonList = document.querySelector( '.plugin-card-' + pluginSlug + ' .plugin-action-buttons' )
-
-							if ( pluginSettingsURL && actionButtonList ) {
-								const listItem = document.createElement( 'li' );
-								const anchor = document.createElement( 'a' );
-
-								anchor.setAttribute( 'href', pluginSettingsURL );
-								anchor.textContent = data.settingsSecondaryText;
-
-								listItem.appendChild( anchor );
-								actionButtonList.appendChild( listItem );
-							}
-							
-							showAdminNotice( data.successText, 'success', pluginSettingsURL );
-						} catch (error) {
-							showAdminNotice( data.errorText );
-
-							target.classList.remove( 'updating-message' );
-							target.textContent = data.activateText;
-						}
-					}
-				} );
-			} );
-
-			function showAdminNotice( message, type = 'error', pluginSettingsURL = undefined ) {
-				// Create the notice container elements.
-				const notice = document.createElement( 'div' );
-				const para = document.createElement( 'p' );
-				
-				notice.className = 'notice is-dismissible notice-' + type;
-				para.textContent = message;
-				
-				if ( pluginSettingsURL ) {
-					para.textContent = para.textContent + data.reviewText;
-					
-					const anchor = document.createElement( 'a' );
-					anchor.setAttribute( 'href', pluginSettingsURL );
-					anchor.textContent = data.settingsPrimaryText;
-					
-					para.appendChild( anchor );
-					para.appendChild( document.createTextNode( data.endText ) );
-				}
-				
-				notice.appendChild( para );
-				
-				const dismissButton = document.createElement( 'button' );
-				const dismissButtonTextWrap = document.createElement( 'span' );
-
-				dismissButton.type = 'button';
-				dismissButton.className = 'notice-dismiss';
-	
-				dismissButtonTextWrap.className = 'screen-reader-text';
-				dismissButtonTextWrap.textContent = data.dismissText;	
-
-				dismissButton.appendChild( dismissButtonTextWrap );
-
-				// Add event listener to remove the notice when dismissed.
-				dismissButton.addEventListener( 'click', () => {
-					notice.remove();
-				} );
-
-				notice.appendChild( dismissButton );
-	
-				// Insert the notice at the top of the admin notices area.
-				const noticeContainer = document.querySelector( '.wrap.plugin-install-php' ) || document.body;
-	
-				if ( !noticeContainer ) {
-					// Fallback append to body if no suitable container is found.
-					document.body.prepend( notice );
-
-					return;
-				} 
-
-				if ( noticeContainer.children.length >= 1 ) {
-					// Insert as the second child.
-					noticeContainer.insertBefore( notice, noticeContainer.children[1] );
-
-					return;
-				}
-
-				// If there's only one child or none, append as the last child.
-				noticeContainer.appendChild( notice );
-			}
-		}
-JS;
-
-	wp_print_inline_script_tag(
-		sprintf(
-			'( %s )( %s );',
-			$js_function,
-			wp_json_encode( $script_data )
-		),
-		array( 'type' => 'module' )
-	);
 }
 
 /**
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -291,6 +291,36 @@ function perflab_install_activate_plugin_callback(): void {
 add_action( 'admin_action_perflab_install_activate_plugin', 'perflab_install_activate_plugin_callback' );
 
 /**
+ * Callback for handling installation/activation of plugin using ajax.
+ */
+function perflab_install_activate_plugin_ajax_callback(): void {
+	check_ajax_referer( 'perflab_install_activate_plugin_ajax', '_ajax_nonce' );
+
+	require_once ABSPATH . 'wp-admin/includes/plugin.php';
+	require_once ABSPATH . 'wp-admin/includes/plugin-install.php';
+	require_once ABSPATH . 'wp-admin/includes/class-wp-upgrader.php';
+	require_once ABSPATH . 'wp-admin/includes/class-wp-ajax-upgrader-skin.php';
+
+	if ( ! isset( $_POST['slug'] ) ) {
+		wp_send_json_error( __( 'Missing required parameter.', 'performance-lab' ), 400 );
+	}
+
+	$plugin_slug = perflab_sanitize_plugin_slug( wp_unslash( $_POST['slug'] ) );
+	if ( null === $plugin_slug ) {
+		wp_send_json_error( __( 'Invalid plugin.', 'performance-lab' ), 400 );
+	}
+
+	// Install and activate the plugin and its dependencies.
+	$result = perflab_install_and_activate_plugin( $plugin_slug );
+	if ( $result instanceof WP_Error ) {
+		wp_send_json_error( $result->get_error_message() );
+	}
+
+	wp_send_json_success();
+}
+add_action( 'wp_ajax_perflab_install_activate_plugin', 'perflab_install_activate_plugin_ajax_callback' );
+
+/**
  * Callback function to handle admin inline style.
  *
  * @since 3.0.0
@@ -402,8 +432,18 @@ function perflab_plugin_admin_notices(): void {
  * @since 3.1.0
  */
 function perflab_print_plugin_progress_indicator_script(): void {
+	$nonce = wp_create_nonce( 'perflab_install_activate_plugin_ajax' );
+
+	$script_data = array(
+		'ajaxUrl'        => admin_url( 'admin-ajax.php' ),
+		'nonce'          => $nonce,
+		'activatingText' => __( 'Activating...', 'performance-lab' ),
+		'activateText'   => __( 'Activate', 'performance-lab' ),
+		'activatedText'  => __( 'Active', 'performance-lab' ),
+	);
+
 	$js_function = <<<JS
-		function addPluginProgressIndicator( message ) {
+		function addPluginProgressIndicator( data ) {
 			document.addEventListener( 'DOMContentLoaded', function () {
 				document.addEventListener( 'click', function ( event ) {
 					if (
@@ -411,11 +451,51 @@ function perflab_print_plugin_progress_indicator_script(): void {
 							'perflab-install-active-plugin'
 						)
 					) {
+						// Prevent the default link behavior.
+						event.preventDefault();
+
 						const target = event.target;
 						target.classList.add( 'updating-message' );
-						target.textContent = message;
+						target.textContent = data.activatingText;
 
-						wp.a11y.speak( message );
+						wp.a11y.speak( data.activatingText );
+
+						const pluginSlug = target.getAttribute('data-plugin-slug');
+
+						fetch(data.ajaxUrl, {
+							method: 'POST',
+							credentials: 'same-origin',
+							headers: {
+								'Content-Type': 'application/x-www-form-urlencoded',
+							},
+							body: new URLSearchParams({
+								action: 'perflab_install_activate_plugin',
+								slug: pluginSlug,
+								_ajax_nonce: data.nonce,
+							}),
+						})
+						.then(response => response.json())
+						.then(response => {
+							if (response.success) {
+								const newButton = document.createElement('button');
+								newButton.type = 'button';
+								newButton.className = 'button button-disabled';
+								newButton.disabled = true;
+								newButton.textContent = data.activatedText;
+								target.parentNode.replaceChild(newButton, target);
+							} else {
+								console.error('Error:', response);
+								// TODO: Show error using admin notice.
+								target.classList.remove('updating-message');
+								target.textContent = data.activateText;
+							}
+                    	})
+						.catch(error => {
+							console.error('Error:', error);
+							// TODO: Show error using admin notice.
+							target.classList.remove('updating-message');
+							target.textContent = data.activateText;
+						});
 					}
 				} );
 			} );
@@ -426,7 +506,7 @@ JS;
 		sprintf(
 			'( %s )( %s );',
 			$js_function,
-			wp_json_encode( __( 'Activating...', 'default' ) )
+			wp_json_encode( $script_data )
 		),
 		array( 'type' => 'module' )
 	);
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -316,6 +316,15 @@ function perflab_install_activate_plugin_ajax_callback(): void {
 		wp_send_json_error( $result->get_error_message() );
 	}
 
+	$plugin_settings_url = perflab_get_plugin_settings_url( $plugin_slug );
+	if ( null !== $plugin_settings_url ) {
+		wp_send_json_success(
+			array(
+				'pluginSettingsURL' => esc_url( $plugin_settings_url ),
+			)
+		);
+	}
+
 	wp_send_json_success();
 }
 add_action( 'wp_ajax_perflab_install_activate_plugin', 'perflab_install_activate_plugin_ajax_callback' );
@@ -440,12 +449,18 @@ function perflab_print_plugin_progress_indicator_script(): void {
 		'activatingText' => __( 'Activating...', 'performance-lab' ),
 		'activateText'   => __( 'Activate', 'performance-lab' ),
 		'activatedText'  => __( 'Active', 'performance-lab' ),
+		'dismissText'    => __( 'Dismiss this notice.', 'performance-lab' ),
+		'successText'    => __( 'Feature activated.', 'performance-lab' ),
+		'reviewText'     => __( ' Review ', 'performance-lab' ),
+		'settingsText'   => __( 'settings', 'performance-lab' ),
+		'endText'        => _x( '.', 'Punctuation mark', 'performance-lab' ),
+		'errorText'      => __( 'There was an error activating the plugin. Please try again.', 'performance-lab' ),
 	);
 
 	$js_function = <<<JS
 		function addPluginProgressIndicator( data ) {
 			document.addEventListener( 'DOMContentLoaded', function () {
-				document.addEventListener( 'click', function ( event ) {
+				document.addEventListener( 'click', async function ( event ) {
 					if (
 						event.target.classList.contains(
 							'perflab-install-active-plugin'
@@ -460,45 +475,112 @@ function perflab_print_plugin_progress_indicator_script(): void {
 
 						wp.a11y.speak( data.activatingText );
 
-						const pluginSlug = target.getAttribute('data-plugin-slug');
+						const pluginSlug = target.getAttribute( 'data-plugin-slug' );
 
-						fetch(data.ajaxUrl, {
-							method: 'POST',
-							credentials: 'same-origin',
-							headers: {
-								'Content-Type': 'application/x-www-form-urlencoded',
-							},
-							body: new URLSearchParams({
-								action: 'perflab_install_activate_plugin',
-								slug: pluginSlug,
-								_ajax_nonce: data.nonce,
-							}),
-						})
-						.then(response => response.json())
-						.then(response => {
-							if (response.success) {
-								const newButton = document.createElement('button');
-								newButton.type = 'button';
-								newButton.className = 'button button-disabled';
-								newButton.disabled = true;
-								newButton.textContent = data.activatedText;
-								target.parentNode.replaceChild(newButton, target);
-							} else {
-								console.error('Error:', response);
-								// TODO: Show error using admin notice.
+						try {
+							const response = await fetch( data.ajaxUrl, {
+								method: 'POST',
+								credentials: 'same-origin',
+								headers: {
+									'Content-Type': 'application/x-www-form-urlencoded',
+								},
+								body: new URLSearchParams( {
+									action: 'perflab_install_activate_plugin',
+									slug: pluginSlug,
+									_ajax_nonce: data.nonce,
+								} ),
+							} );
+
+							const responseData = await response.json();
+
+							if ( !responseData.success ) {
+								showAdminNotice( data.errorText );
+
 								target.classList.remove('updating-message');
 								target.textContent = data.activateText;
+
+								return;
 							}
-                    	})
-						.catch(error => {
-							console.error('Error:', error);
-							// TODO: Show error using admin notice.
-							target.classList.remove('updating-message');
+
+							const newButton = document.createElement( 'button' );
+
+							newButton.type = 'button';
+							newButton.className = 'button button-disabled';
+							newButton.disabled = true;
+							newButton.textContent = data.activatedText;
+
+							target.parentNode.replaceChild( newButton, target );
+							console.log(responseData)
+							showAdminNotice( data.successText, 'success', responseData?.data?.pluginSettingsURL );
+						} catch (error) {
+							showAdminNotice( data.errorText );
+
+							target.classList.remove( 'updating-message' );
 							target.textContent = data.activateText;
-						});
+						}
 					}
 				} );
 			} );
+
+			function showAdminNotice( message, type = 'error', pluginSettingsURL = undefined ) {
+				// Create the notice container elements.
+				const notice = document.createElement( 'div' );
+				const para = document.createElement( 'p' );
+				
+				notice.className = 'notice is-dismissible notice-' + type;
+				para.textContent = message;
+				
+				if ( pluginSettingsURL ) {
+					para.textContent = para.textContent + data.reviewText;
+					
+					const anchor = document.createElement( 'a' );
+					anchor.setAttribute( 'href', pluginSettingsURL );
+					anchor.textContent = data.settingsText;
+					
+					para.appendChild( anchor );
+					para.appendChild( document.createTextNode( data.endText ) );
+				}
+				
+				notice.appendChild( para );
+				
+				const dismissButton = document.createElement( 'button' );
+				const dismissButtonTextWrap = document.createElement( 'span' );
+
+				dismissButton.type = 'button';
+				dismissButton.className = 'notice-dismiss';
+	
+				dismissButtonTextWrap.className = 'screen-reader-text';
+				dismissButtonTextWrap.textContent = data.dismissText;	
+
+				dismissButton.appendChild( dismissButtonTextWrap );
+
+				// Add event listener to remove the notice when dismissed.
+				dismissButton.addEventListener( 'click', () => {
+					notice.remove();
+				} );
+
+				notice.appendChild( dismissButton );
+	
+				// Insert the notice at the top of the admin notices area.
+				const noticeContainer = document.querySelector( '.wrap.plugin-install-php' ) || document.body;
+	
+				if ( !noticeContainer ) {
+					// Fallback append to body if no suitable container is found.
+					document.body.prepend( notice );
+
+					return;
+				} 
+
+				if ( noticeContainer.children.length >= 1 ) {
+					// Insert as the second child.
+					noticeContainer.insertBefore( notice, noticeContainer.children[1] );
+
+					return;
+				}
+
+				// If there's only one child or none, append as the last child.
+				noticeContainer.appendChild( notice );
+			}
 		}
 JS;
 
--- a/plugins/performance-lab/includes/admin/load.php
+++ b/plugins/performance-lab/includes/admin/load.php
@@ -39,8 +39,6 @@ add_action( 'admin_menu', 'perflab_add_features_page' );
  * @since 1.0.0
  * @since 3.0.0 Renamed to perflab_load_features_page(), and the
  *              $module and $hook_suffix parameters were removed.
- * @since n.e.x.t Removed added action for perflab_print_plugin_progress_indicator_script
- *                as this function is removed.
  */
 function perflab_load_features_page(): void {
 	// Handle script enqueuing for settings page.
@@ -220,7 +218,6 @@ add_action( 'wp_ajax_dismiss-wp-pointer', 'perflab_dismiss_wp_pointer_wrapper', 
  *
  * @since 2.8.0
  * @since 3.0.0 Renamed to perflab_enqueue_features_page_scripts().
- * @since n.e.x.t Added enqueue plugin activate AJAX script.
  */
 function perflab_enqueue_features_page_scripts(): void {
 	// These assets are needed for the "Learn more" popover.
--- a/plugins/performance-lab/includes/admin/plugin-activate-ajax.js
+++ b/plugins/performance-lab/includes/admin/plugin-activate-ajax.js
@@ -3,96 +3,98 @@
  */
 
 /* global perflabPluginActivateAjaxData */
-( function ( $ ) {
+( function () {
 	// @ts-ignore
 	const { i18n, a11y } = wp;
 	const { __ } = i18n;
 
-	$( document ).on(
-		'click',
-		'.perflab-install-active-plugin',
-		/**
-		 * Adds a click event listener to the plugin activate buttons.
-		 *
-		 * @param {MouseEvent} event - The click event object that is triggered when the user clicks on the document.
-		 *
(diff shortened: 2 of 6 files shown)


plugins/performance-lab/includes/admin/load.php
```diff
@@ -230,6 +227,24 @@ function perflab_enqueue_features_page_scripts(): void {
 
 	// Enqueue the a11y script.
 	wp_enqueue_script( 'wp-a11y' );
+
+	// Enqueue plugin activate AJAX script and localize script data.
+	wp_enqueue_script(
+		'perflab-plugin-activate-ajax',
+		PERFLAB_PLUGIN_DIR_URL . 'includes/admin/plugin-activate-ajax.js',
```
westonruter: This `PERFLAB_PLUGIN_DIR_URL` is `plugin_dir_url( PERFLAB_MAIN_FILE )`, but I believe here it would be better to use `plugins_url()`:

```suggestion
		plugins_url( 'includes/admin/plugin-activate-ajax.js', PERFLAB_MAIN_FILE ),
```

The reason is that this function allows for filtering as needed.

plugins/performance-lab/includes/admin/load.php
```diff
@@ -290,6 +305,47 @@ function perflab_install_activate_plugin_callback(): void {
...
+	$result = perflab_install_and_activate_plugin( $plugin_slug );
+	if ( $result instanceof WP_Error ) {
+		wp_send_json_error( $result->get_error_message() );
+	}
+
+	$plugin_settings_url = perflab_get_plugin_settings_url( $plugin_slug );
+	if ( null !== $plugin_settings_url ) {
+		wp_send_json_success(
+			array(
+				'pluginSettingsURL' => esc_url( $plugin_settings_url ),
```
westonruter: The `esc_url()` function outputs HTML entities which we don't want in a JSON context here.

```suggestion
				'pluginSettingsURL' => esc_url_raw( $plugin_settings_url ),
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,175 @@
...
+	const { i18n, a11y } = wp;
+	const { __, _x } = i18n;
+
+	/**
+	 * Adds a click event listener to the plugin activate buttons.
+	 *
+	 * @param {MouseEvent} event - The click event object that is triggered when the user clicks on the document.
+	 *
+	 * @return {Promise<void>} - The asynchronous function returns a promise.
+	 */
```
westonruter: Shouldn't this jsdoc block be moved down to the line right before `async function ( event ) {`?

plugins/performance-lab/load.php
```diff
@@ -22,6 +22,7 @@
 define( 'PERFLAB_VERSION', '3.5.1' );
 define( 'PERFLAB_MAIN_FILE', __FILE__ );
 define( 'PERFLAB_PLUGIN_DIR_PATH', plugin_dir_path( PERFLAB_MAIN_FILE ) );
+define( 'PERFLAB_PLUGIN_DIR_URL', plugin_dir_url( PERFLAB_MAIN_FILE ) );
```
westonruter: If my suggestion above is accepted, this can then be removed.

```suggestion
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,175 @@
...
+	$( document ).on(
+		'click',
+		'.perflab-install-active-plugin',
```
westonruter: What if someone clicks this button while another request is currently in flight? Should it short-circuit so as to avoid duplicating requests?

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,175 @@
...
+			const target = $( this );
+
+			target
+				.addClass( 'updating-message' )
+				.text( __( 'Activating…', 'performance-lab' ) );
+
+			a11y.speak( __( 'Activating…', 'performance-lab' ) );
+
+			// Retrieve the plugin slug from the data attribute.
+			const pluginSlug = $.trim( target.attr( 'data-plugin-slug' ) );
```
westonruter: I think we should avoid using jQuery unless we have to.

```suggestion
			const pluginSlug = target.attr( 'data-plugin-slug' ).trim();
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,175 @@
...
+		if ( noticeContainer.length ) {
+			// If the container exists, insert the notice after the first child.
+			noticeContainer.children().eq( 0 ).after( notice );
+		} else {
+			$( 'body' ).prepend( notice );
+		}
+	}
+
+	// @ts-ignore
+} )( window.jQuery );
```
westonruter: Does it make sense to make jQuery a dependency? Would plain JS not be feasible?

westonruter (changes requested):

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,175 @@
...
+								} )
+							)
+						);
+					}
+
+					showAdminNotice(
+						__( 'Feature activated.', 'performance-lab' ),
+						'success',
+						pluginSettingsURL
+					);
```
westonruter: In the demo video you showed that this notice is causing layout shifts, resulting in you accidentally clicking the title of a plugin instead of the Activate button. I don't think an admin notice is even needed here, because the button transitions from Activate to Activating... to Active.

westonruter:
> It does not work with the `Speculative Loading` because of how this plugin is loaded.
> * This plugin gets loaded ( loaded in this context is when are its helper files are loaded like `settings.php` ) on `after_setup_theme` hook.
> * But when the AJAX callback is triggered this hook has already been executed. As the plugin is activated later and which requries the `after_setup_theme` hook to fully load the plugin which includes the settings URL.
> * The function which is executed in the `after_setup_theme` calls a closure function which is stored in the global variable ( `plsr_pending_plugin_info` for the `Speculative Loading` plugin ) this closure function fully loads the plugin.

So the issue is that when a plugin is activated, its source files are included, but they are included after the `after_setup_theme` action has fired, so then it is too late for the plugin to initialize in order to add the `plugin_action_links_*` filter. Interesting. 

Given that plugin activation inherently happens partyway through another WP execution lifecycle after many of the actions potentially used by the activated plugin have already fired, I think it makes sense to introduce a secondary endpoint exclusively for obtaining the settings link. So you'd have one endpoint for activating the plugin and another endpoint for getting the settings link. This should be more reliable across how plugins bootstrap themselves.

plugins/performance-lab/includes/admin/load.php
felixarntz: @westonruter I don't think we should randomly change it here though. If not using `plugins_url()` leads to a problem, let's discuss that in a separate issue and implement properly if needed.

plugins/performance-lab/load.php
felixarntz: See above, let's discuss separately, it's not really relevant for this enhancement.

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
felixarntz: +1, it would be great to use vanilla JavaScript rather than relying on jQuery, to set a good example for performance.

plugins/performance-lab/includes/admin/load.php
```diff
@@ -230,6 +227,24 @@ function perflab_enqueue_features_page_scripts(): void {
 
 	// Enqueue the a11y script.
 	wp_enqueue_script( 'wp-a11y' );
+
+	// Enqueue plugin activate AJAX script and localize script data.
+	wp_enqueue_script(
+		'perflab-plugin-activate-ajax',
+		PERFLAB_PLUGIN_DIR_URL . 'includes/admin/plugin-activate-ajax.js',
+		array( 'jquery', 'wp-i18n', 'wp-a11y' ),
```
felixarntz: Related to @westonruter's previous feedback, it would be great not to use jQuery for the functionality, but rather vanilla JavaScript (e.g. `fetch()` function).

plugins/performance-lab/load.php
westonruter: At least let's avoid introducing a new constant in this PR.

felixarntz (commented):
Thanks for the PR @b1ink0!

Not much to add from myself to the feedback @westonruter already provided. It would be great to not have to use jQuery for this.

plugins/performance-lab/load.php
felixarntz: That's fair, we can simply use `plugin_dir_url( PERFLAB_MAIN_FILE )` in the `wp_enqueue_script()` call directly. I'm personally not a fan of having constants for all this either, particularly as the values are filterable, like you're saying.

westonruter: I've filed https://github.com/WordPress/performance/issues/1647 for the filtering.

plugins/performance-lab/includes/admin/load.php
westonruter: I've filed https://github.com/WordPress/performance/issues/1647 for this.

westonruter:
> Given that plugin activation inherently happens partyway through another WP execution lifecycle after many of the actions potentially used by the activated plugin have already fired, I think it makes sense to introduce a secondary endpoint exclusively for obtaining the settings link. So you'd have one endpoint for activating the plugin and another endpoint for getting the settings link. This should be more reliable across how plugins bootstrap themselves.

This said, it seems like we can avoid having to introduce a separate endpoint by tweaking how Speculative Loading bootstraps. Instead of:

https://github.com/WordPress/performance/blob/5fbd82eb64514df39ac7ccef5a445b08fb117879/plugins/speculation-rules/load.php#L31-L65

We could do this:

```php
	static function ( string $global_var_name, string $version, Closure $load ): void {
		$needs_bootstrap = ! isset( $GLOBALS[ $global_var_name ] );

		// Register this copy of the plugin.
		if (
			// Register this copy if none has been registered yet.
			! isset( $GLOBALS[ $global_var_name ]['version'] )
			||
			// Or register this copy if the version greater than what is currently registered.
			version_compare( $version, $GLOBALS[ $global_var_name ]['version'], '>' )
			||
			// Otherwise, register this copy if it is actually the one installed in the directory for plugins.
			rtrim( WP_PLUGIN_DIR, '/' ) === dirname( __DIR__ )
		) {
			$GLOBALS[ $global_var_name ]['version'] = $version;
			$GLOBALS[ $global_var_name ]['load']    = $load;
		}

		if ( $needs_bootstrap ) {
			$bootstrap = static function () use ( $global_var_name ): void {
				if (
					isset( $GLOBALS[ $global_var_name ]['load'], $GLOBALS[ $global_var_name ]['version'] )
					&&
					$GLOBALS[ $global_var_name ]['load'] instanceof Closure
					&&
					is_string( $GLOBALS[ $global_var_name ]['version'] )
				) {
					call_user_func( $GLOBALS[ $global_var_name ]['load'], $GLOBALS[ $global_var_name ]['version'] );
					unset( $GLOBALS[ $global_var_name ] );
				}
			};

			// Wait until after the plugins have loaded and the theme has loaded. The after_setup_theme action is used
			// because it is the first action that fires once the theme is loaded.
			if ( did_action( 'after_setup_theme' ) ) {
				$bootstrap();
			} else {
				add_action( 'after_setup_theme', $bootstrap, PHP_INT_MIN );
			}
		}
	}
```

Where if `after_setup_theme` already fired, we then invoke the bootstrap logic immediately. This would fix the issue with Speculative Loading, but this would essentially be a one-off fix that wouldn't necessarily work for other plugins. For example, Optimization Detective initializes at the `init` action, and at that point it triggers its own `od_init` action for other plugins to register their own dependent logic. Image Prioritizer, for example, waits to load any of its logic until the `od_init` action fires. If the Optimization Detective plugin was already installed and active, and then someone uses Ajax to activate the Image Prioritizer plugin, then the `od_init` action will never fire either, resulting in this exact same problem. So I think it is better to let the PHP process to activate the plugin fully terminate and then to start a new PHP process in a new endpoint to obtain the settings URL. This will avoid all issues with hook ordering across all plugins.

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
b1ink0: I think as the request could have been half finished we should not short-circuit the request. I think it will be better to just disable it until the request is finished. But as the click event is on the anchor tag we will have to add `pointer-events: none` CSS to the anchor link using JavaScript to disable it.

plugins/performance-lab/includes/admin/load.php
```diff
@@ -230,6 +227,24 @@ function perflab_enqueue_features_page_scripts(): void {
 
 	// Enqueue the a11y script.
 	wp_enqueue_script( 'wp-a11y' );
```
mukeshpanchal27: Do we need these enqueue now? as it already added through new script dependancy.

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
westonruter: The `pointer-events` style wouldn't address the issue of the button being activated by keyboard.

plugins/performance-lab/includes/admin/load.php
b1ink0: We could remove it as it is not being used anywhere other than the AJAX script as it already added in its dependancy. Should I remove it then?

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
b1ink0: As the click event is on the anchor tag and not button we will have to add some other work around for disabling on keyboard one is setting its `tabindex` attribute to `-1`.

b1ink0: I tested this but if user is already focused then `tabindex` will not work. One other way is to add a attribute like `aria-disabled` and then check on `click` event if it has that attribute then return from the execution. Should I add `tabindex` and `aria-disabled` both?

plugins/speculation-rules/load.php
westonruter, line 1: Actually, in https://github.com/WordPress/performance/pull/1646#issuecomment-2468871567 I just posed this change as a possibility but then recommended against it since it would only account for Speculative Loading. It wouldn't account for other plugins that have delayed bootstrapping. So to address the problem of accessing the settings link after a plugin is activated, I think the best solution is to make another request specifically to an endpoint that exposes the settings link URLs.

Aside: I think the REST API would be better for this instead of admin-ajax.

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,107 @@
...
+
+		a11y.speak( __( 'Activating…', 'performance-lab' ) );
+
+		const pluginSlug = target.getAttribute( 'data-plugin-slug' ).trim();
+
+		// Send an AJAX POST request to activate the plugin.
+		try {
+			const response = await fetch(
+				// @ts-ignore
+				perflabPluginActivateAjaxData.ajaxUrl,
```
westonruter: Why not use the REST API? You could just use `wp.apiFetch` then without having to export `perflabPluginActivateAjaxData` to the frontend.

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,107 @@
...
+		event.preventDefault();
+
+		target.classList.add( 'updating-message' );
+		target.textContent = __( 'Activating…', 'performance-lab' );
+		target.style.pointerEvents = 'none';
+		target.parentElement.style.cursor = 'not-allowed';
+
+		a11y.speak( __( 'Activating…', 'performance-lab' ) );
+
+		const pluginSlug = target.getAttribute( 'data-plugin-slug' ).trim();
```
westonruter:
```suggestion
		const pluginSlug = target.dataset.pluginSlug.trim();
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
westonruter: You could add the `disabled` attribute, but that wouldn't be good because it would lose the tab focus for users navigating with the keyboard. Instead of this, WordPress has style rules for when you add the `disabled` class to a button so that it will look disabled. Adding `aria-disabled` here too would be good. But then for how to detect whether request is currently in progress, you could just short-circuit the click handler if the button has the `updating-message` class, or if it has the `disabled` class, or if it has the `aria-disabled` attribute.

plugins/performance-lab/includes/admin/load.php
```diff
@@ -230,6 +227,24 @@ function perflab_enqueue_features_page_scripts(): void {
...
+
+	// Enqueue plugin activate AJAX script and localize script data.
+	wp_enqueue_script(
+		'perflab-plugin-activate-ajax',
+		plugins_url( 'includes/admin/plugin-activate-ajax.js', PERFLAB_MAIN_FILE ),
+		array( 'wp-i18n', 'wp-a11y' ),
```
westonruter: As noted in a comment below, this can use `wp-api-fetch` instead if the REST API is used.

```suggestion
		array( 'wp-i18n', 'wp-a11y', 'wp-api-fetch' ),
```

plugins/performance-lab/includes/admin/load.php
```diff
@@ -230,6 +227,24 @@ function perflab_enqueue_features_page_scripts(): void {
...
+	);
+
+	wp_localize_script(
+		'perflab-plugin-activate-ajax',
+		'perflabPluginActivateAjaxData',
+		array(
+			'ajaxUrl' => admin_url( 'admin-ajax.php' ),
+			'nonce'   => wp_create_nonce( 'perflab_install_activate_plugin_ajax' ),
+		)
+	);
```
westonruter: When using `wp.apiFetch`, this is not needed.

```suggestion
```

plugins/performance-lab/includes/admin/load.php
```diff
@@ -290,6 +305,47 @@ function perflab_install_activate_plugin_callback(): void {
...
+		wp_send_json_success(
+			array(
+				'pluginSettingsURL' => esc_url_raw( $plugin_settings_url ),
+			)
+		);
+	}
+
+	wp_send_json_success();
+}
+add_action( 'wp_ajax_perflab_install_activate_plugin', 'perflab_install_activate_plugin_ajax_callback' );
```
westonruter: I think this would be better as a REST API endpoint.

And then a separate one for obtaining the settings URLs.

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,107 @@
+/**
+ * Handles activation of Performance Features (Plugins) using AJAX.
+ */
+
+/* global perflabPluginActivateAjaxData */
+( function () {
+	// @ts-ignore
+	const { i18n, a11y } = wp;
```
westonruter: Per above:

```suggestion
	const { i18n, a11y, apiFetch } = wp;
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,107 @@
...
+	async function handlePluginActivationClick( event ) {
+		const target = /** @type {HTMLElement} */ ( event.target );
+
+		if ( ! target.classList.contains( 'perflab-install-active-plugin' ) ) {
+			return;
+		}
+
+		// Prevent the default link behavior.
+		event.preventDefault();
+
```
westonruter, line 25:
```suggestion
		if ( target.classList.contains( 'updating-message' ) ) {
			return;
		}
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,107 @@
...
+		target.classList.add( 'updating-message' );
+		target.textContent = __( 'Activating…', 'performance-lab' );
+		target.style.pointerEvents = 'none';
+		target.parentElement.style.cursor = 'not-allowed';
```
westonruter:
```suggestion
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,107 @@
...
+				}
+			);
+
+			const responseData = await response.json();
+
+			if ( ! responseData.success ) {
+				target.classList.remove( 'updating-message' );
+				target.textContent = __( 'Activate', 'performance-lab' );
+				target.style.pointerEvents = '';
+				target.parentElement.style.cursor = '';
```
westonruter:
```suggestion
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,107 @@
...
+
+			const newButton = document.createElement( 'button' );
+
+			newButton.type = 'button';
+			newButton.className = 'button button-disabled';
+			newButton.disabled = true;
+			newButton.textContent = __( 'Active', 'performance-lab' );
+			target.parentElement.style.cursor = '';
+
+			target.parentNode.replaceChild( newButton, target );
```
westonruter: Maybe just update the existing button instead of replace the button? By keeping the same element in place, users with a keyboard won't have the focus removed from the document.

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,107 @@
...
+
+			const actionButtonList = document.querySelector(
+				`.plugin-card-${ pluginSlug } .plugin-action-buttons`
+			);
+
+			if ( pluginSettingsURL && actionButtonList ) {
+				const listItem = document.createElement( 'li' );
+				const anchor = document.createElement( 'a' );
+
+				anchor.setAttribute( 'href', pluginSettingsURL );
```
westonruter:
```suggestion
				anchor.href = pluginSettingsURL;
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,107 @@
...
+				anchor.textContent = __( 'Settings', 'performance-lab' );
+
+				listItem.appendChild( anchor );
+				actionButtonList.appendChild( listItem );
+			}
+		} catch ( error ) {
+			target.classList.remove( 'updating-message' );
+			target.textContent = __( 'Activate', 'performance-lab' );
+			target.style.pointerEvents = '';
+			target.parentElement.style.cursor = '';
```
westonruter:
```suggestion
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,107 @@
...
+						slug: pluginSlug,
+						// @ts-ignore
+						_ajax_nonce: perflabPluginActivateAjaxData.nonce,
+					} ),
+				}
+			);
+
+			const responseData = await response.json();
+
+			if ( ! responseData.success ) {
```
westonruter: Maybe this should throw an error because the same logic here is located in the `catch` block.

westonruter (changes requested):

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
westonruter: Aside: The trim isn't really needed, is it? No whitespace is added when the attribute is generated in the HTML.

```suggestion
		const pluginSlug = target.dataset.pluginSlug;
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
b1ink0: The target is actually `<a>` tag ( because it should be able to work even if JavaScript is disabled ) and to match the mark up of loaded page which is disabled button because of that I am replacing the target `<a>` tag with button.

westonruter: I think the `disabled` class works with a link as well. I think you can keep it as a link instead of replacing it with a button.

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,89 @@
...
+
+	/**
+	 * Handles click events on elements with the class 'perflab-install-active-plugin'.
+	 *
+	 * This asynchronous function listens for click events on the document and executes
+	 * the provided callback function if triggered.
+	 *
+	 * @param {MouseEvent} event - The click event object that is triggered when the user clicks on the document.
+	 *
+	 * @return {Promise<void>} - The asynchronous function returns a promise that resolves to void.
```
westonruter: Nit: the hyphen can be removed since it is [only used with `@param`](https://jsdoc.app/tags-param#:~:text=You%20can%20add%20a%20hyphen%20before%20the%20description%20to%20make%20it%20more%20readable) to make it more readable. It's not seen in the [docs for `@return`](https://jsdoc.app/tags-returns).

```suggestion
	 * @return {Promise<void>} The asynchronous function returns a promise that resolves to void.
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+ * @var string
+ */
+const PERFLAB_REST_API_NAMESPACE = 'performance-lab/v1';
+
+/**
+ * Route for activating plugin.
+ *
+ * @var string
+ */
+const PERFLAB_ACTIVATE_PLUGIN_ROUTE = '/activate-plugin';
```
westonruter: Since this endpoint is not a noun, it may make more sense as `/plugin:activate`. See for example in Optimization Detective:

https://github.com/WordPress/performance/blob/2bec0e8edbfbc3d8b2fb1216b40ec11988e4a61c/plugins/optimization-detective/storage/rest-api.php#L20-L30

```suggestion
/**
 * Route for activating plugin.
 *
 * Note the `:activate` art of the endpoint follows Google's guidance in AIP-136 for the use of the POST method in a way 
 * that does not strictly follow the standard usage.
 * 
 * @link https://google.aip.dev/136 
 * @var string
 */
const PERFLAB_ACTIVATE_PLUGIN_ROUTE = '/plugin:activate;
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	$plugin_slug = perflab_sanitize_plugin_slug( wp_unslash( $params['slug'] ) );
+	if ( null === $plugin_slug ) {
+		return new WP_Error(
+			'invalid_plugin',
+			__( 'Invalid plugin slug provided.', 'performance-lab' ),
+			array( 'status' => 400 )
+		);
+	}
```
westonruter: This can be eliminated in favor of adding the validation to the REST API arg.

```suggestion
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+
+	$plugin_settings_url = perflab_get_plugin_settings_url( $plugin_slug );
```
westonruter:
```suggestion
	$plugin_settings_url = perflab_get_plugin_settings_url( $request['slug'] );
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	$plugin_slug = perflab_sanitize_plugin_slug( wp_unslash( $params['slug'] ) );
+	if ( null === $plugin_slug ) {
+		return new WP_Error(
+			'invalid_plugin',
+			__( 'Invalid plugin slug provided.', 'performance-lab' ),
+			array( 'status' => 400 )
+		);
+	}
```
westonruter: This code isn't needed because the validation is handed by the REST API, per above.

```suggestion
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+			array(
+				'success'           => true,
+				'pluginSettingsURL' => false,
+			)
+		);
+	}
+
+	return new WP_REST_Response(
+		array(
+			'success'           => true,
```
westonruter:
```suggestion
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+			'pluginSettingsURL' => esc_url_raw( $plugin_settings_url ),
```
westonruter: I don't think sanitization of the URL is really necessary since it's not user data.

```suggestion
			'pluginSettingsURL' => $plugin_settings_url,
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	$plugin_settings_url = perflab_get_plugin_settings_url( $plugin_slug );
+	if ( null === $plugin_settings_url ) {
+		return new WP_REST_Response(
+			array(
+				'success'           => true,
+				'pluginSettingsURL' => false,
+			)
+		);
```
westonruter: I think it would make more sense to return a 404 status code here:

```suggestion
		return new WP_Error( 
			'no_settings_url',
			__( 'No settings URL', 'performance-lab' ),
			array( 'status' => 404 )
		);
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+	if ( null === $plugin_slug ) {
+		return new WP_Error(
+			'invalid_plugin',
+			__( 'Invalid plugin slug provided.', 'performance-lab' ),
+			array( 'status' => 400 )
+		);
+	}
+
+	// Install and activate the plugin and its dependencies.
+	$result = perflab_install_and_activate_plugin( $plugin_slug );
```
westonruter:
```suggestion
	$result = perflab_install_and_activate_plugin( $request['slug'] );
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,89 @@
...
+			await apiFetch( {
+				path: '/performance-lab/v1/activate-plugin',
+				method: 'POST',
+				data: { slug: pluginSlug },
+			} );
+
+			// Fetch the plugin settings URL via the REST API.
+			const settingsResponse = await apiFetch( {
+				path: '/performance-lab/v1/plugin-settings-url',
+				method: 'POST',
```
westonruter:
```suggestion
				method: 'GET',
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+				return new WP_Error( 'cannot_install_plugin', __( 'Sorry, you are not allowed to install plugins on this site.', 'performance-lab' ) );
+			},
+		)
+	);
+
+	register_rest_route(
+		PERFLAB_REST_API_NAMESPACE,
+		PERFLAB_PLUGIN_SETTINGS_URL_ROUTE,
+		array(
+			'methods'             => 'POST',
```
westonruter: This endpoint is "idempotent" so it should use the GET method

```suggestion
			'methods'             => 'GET',
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+ */
+const PERFLAB_ACTIVATE_PLUGIN_ROUTE = '/activate-plugin';
+
+/**
+ * Route for fetching plugin settings URL.
+ *
+ * @var string
+ */
+const PERFLAB_PLUGIN_SETTINGS_URL_ROUTE = '/plugin-settings-url';
+
```
westonruter, line 40:
```suggestion

/**
 * Validates whether the provided plugin slug is a valid Performance Lab plugin.
 *
 * Note that an enum is not being used because additional PHP files have to be required to access the necessary functions,
 * and this would not be ideal to do at rest_api_init.
 * 
 * @since n.e.x.t
 * @access private
 *
 * @param string $slug Plugin slug.
 * @return bool Whether valid.
 */
function perflab_validate_slug_endpoint_arg( string $slug ): bool {
	require_once ABSPATH . 'wp-admin/includes/plugin.php';
	require_once PERFLAB_PLUGIN_DIR_PATH . 'includes/admin/load.php';
	require_once PERFLAB_PLUGIN_DIR_PATH . 'includes/admin/plugins.php';
	return in_array( $slug, perflab_get_standalone_plugins(), true );
}

```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+	register_rest_route(
+		PERFLAB_REST_API_NAMESPACE,
+		PERFLAB_ACTIVATE_PLUGIN_ROUTE,
+		array(
+			'methods'             => 'POST',
+			'args'                => array(
+				'slug' => array(
+					'type'        => 'string',
+					'description' => __( 'Plugin slug of plugin that needs to be activated.', 'performance-lab' ),
+					'required'    => true,
```
westonruter:
```suggestion

					'type'              => 'string',
					'description'       => __( 'Plugin slug of plugin that needs to be activated.', 'performance-lab' ),
					'required'          => true,
					'validate_callback' => 'perflab_validate_slug_endpoint_arg',
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+	register_rest_route(
+		PERFLAB_REST_API_NAMESPACE,
+		PERFLAB_PLUGIN_SETTINGS_URL_ROUTE,
+		array(
+			'methods'             => 'POST',
+			'args'                => array(
+				'slug' => array(
+					'type'        => 'string',
+					'description' => __( 'Plugin slug of plugin whose settings URL is needed.', 'performance-lab' ),
+					'required'    => true,
```
westonruter:
```suggestion
					'type'              => 'string',
					'description'       => __( 'Plugin slug of plugin whose settings URL is needed.', 'performance-lab' ),
					'required'          => true,
					'validate_callback' => 'perflab_validate_slug_endpoint_arg',
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+	$params = $request->get_json_params();
+
+	// Ensure the 'slug' parameter is present.
+	if ( ! isset( $params['slug'] ) ) {
+		return new WP_Error(
+			'missing_parameter',
+			__( 'Missing required parameter "slug".', 'performance-lab' ),
+			array( 'status' => 400 )
+		);
+	}
```
westonruter: This is not needed because the REST API schema enforces that the `slug` parameter will be present.

```suggestion
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,89 @@
...
+			if ( settingsResponse?.pluginSettingsURL && actionButtonList ) {
+				const listItem = document.createElement( 'li' );
+				const anchor = document.createElement( 'a' );
+
+				anchor.href = settingsResponse?.pluginSettingsURL;
+				anchor.textContent = __( 'Settings', 'performance-lab' );
+
+				listItem.appendChild( anchor );
+				actionButtonList.appendChild( listItem );
+			}
```
westonruter: All of the logic specifically for getting the settings link could be moved to a separate try/catch block. If the settings URL endpoint returns a 404, then an error will be thrown and then there is no need to check if `settingsResponse?.pluginSettingsURL` exists.

```suggestion
			a11y.speak( __( 'Plugin activated.', 'performance-lab' ) );

			target.textContent = __( 'Active', 'performance-lab' );
			target.classList.remove( 'updating-message' );
			target.classList.add( 'disabled' );
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,89 @@
...
+		target.textContent = __( 'Activating…', 'performance-lab' );
+
+		a11y.speak( __( 'Activating…', 'performance-lab' ) );
+
+		const pluginSlug = target.dataset.pluginSlug;
+
+		try {
+			// Activate the plugin via the REST API.
+			await apiFetch( {
+				path: '/performance-lab/v1/activate-plugin',
```
westonruter:
```suggestion
				path: '/performance-lab/v1/plugin:activate',
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+ * @var string
+ */
+const PERFLAB_ACTIVATE_PLUGIN_ROUTE = '/activate-plugin';
+
+/**
+ * Route for fetching plugin settings URL.
+ *
+ * @var string
+ */
+const PERFLAB_PLUGIN_SETTINGS_URL_ROUTE = '/plugin-settings-url';
```
westonruter:
```suggestion
const PERFLAB_PLUGIN_SETTINGS_URL_ROUTE = '/plugin-settings-url/(?P<slug>[a-z0-9_-]+)';
```

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,89 @@
...
+				actionButtonList.appendChild( listItem );
+			}
+		} catch ( error ) {
+			a11y.speak( __( 'Plugin failed to activate.', 'performance-lab' ) );
+
+			target.classList.remove( 'updating-message' );
+			target.textContent = __( 'Activate', 'performance-lab' );
+		}
+	}
+
```
westonruter, line 81: Per above:

```suggestion

		try {
			// Fetch the plugin settings URL via the REST API.
			const settingsResponse = await apiFetch( {
				path: `/performance-lab/v1/plugin-settings-url/${ pluginSlug }`,
				method: 'GET',
			} );

			const actionButtonList = document.querySelector(
				`.plugin-card-${ pluginSlug } .plugin-action-buttons`
			);

			const listItem = document.createElement( 'li' );
			const anchor = document.createElement( 'a' );

			anchor.href = settingsResponse.pluginSettingsURL;
			anchor.textContent = __( 'Settings', 'performance-lab' );

			listItem.appendChild( anchor );
			actionButtonList.appendChild( listItem );
		} catch ( error ) {}
```

westonruter (changes requested):
Getting closer!

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,89 @@
...
+	// Attach the event listener.
+	document.addEventListener( 'click', handlePluginActivationClick );
```
felixarntz: Why does this add a listener for the entire `document` instead of adding listeners to the buttons specifically?

I feel like that would be a more reasonable choice, as it avoids having the listener fire for more or less every click on the page. Since the buttons are present in the HTML response from the beginning, I don't see a need for listening on `document`.

plugins/performance-lab/includes/admin/rest-api.php
felixarntz: I disagree here. The caller provides the plugin slug, not an identifier for the settings URL. I'd argue a 404 would only be appropriate if the plugin slug was for a plugin that's not active or doesn't exist. But if the plugin is active and just doesn't provide a settings URL, then it shouldn't be a 404, since the endpoint's sole purpose is to retrieve the settings URL for an active plugin.

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * @since n.e.x.t
+ * @access private
+ *
+ * @phpstan-param WP_REST_Request<array<string, mixed>> $request
+ *
+ * @param WP_REST_Request $request Request.
+ * @return WP_REST_Response|WP_Error Response.
+ */
+function perflab_handle_activate_plugin( WP_REST_Request $request ) {
```
felixarntz: Now that we expose this as a REST endpoint, I wonder whether we should restrict it to plugins that are part of the Performance Lab program. Otherwise we may need to deal with other implications related to arbitrary other plugins. I feel like limiting to our own would be a good idea, for both of these endpoints.

For example, we may make certain assumptions about our plugins, that aren't reasonable to make for _any_ plugin.

plugins/performance-lab/includes/admin/rest-api.php
felixarntz: Indeed this is not a great resource URI for a REST URL, as it's not truly RESTful. REST endpoints should be centered around a resource, in this case a plugin.

But even better would be something like: `/plugins/<slug>:activate`. This would also be more intuitive in that the required parameter `slug` would be directly in the route path.
```suggestion
/**
 * Route for activating plugin.
 *
 * @var string
 */
const PERFLAB_ACTIVATE_PLUGIN_ROUTE = '/plugins/(?P<slug>[a-z0-9_-]+):activate';
```

plugins/performance-lab/includes/admin/rest-api.php
felixarntz: See above, better would be:
```suggestion
const PERFLAB_PLUGIN_SETTINGS_URL_ROUTE = '/plugins/(?P<slug>[a-z0-9_-]+):settings-url';
```

This way the two routes are also consistently centered around a specific plugin.

felixarntz (changes requested):
@b1ink0 This looks like it's on the right track.

I think going with a REST API is great, though it would have been okay to go with Admin AJAX for this kind of endpoint as well, that would have been a bit simpler. If we use the REST API, I think we should ensure we use it correctly. Most of my feedback is centered around that.

But this is coming along nicely!

plugins/performance-lab/includes/admin/rest-api.php
westonruter: This is already restricted to the PL plugins, is it not? It's using `perflab_sanitize_plugin_slug()`  which returns `null` if it is not a valid PL plugin slug. But I also suggested an alternative in https://github.com/WordPress/performance/pull/1646/files#r1840894846

So I think this is fine.

plugins/performance-lab/includes/admin/rest-api.php
westonruter: But in this case it's just a regular REST API endpoint. No need for any workaround for a non-RESTy `POST` request, right?

plugins/performance-lab/includes/admin/rest-api.php
westonruter: The endpoint is for a settings URL entity though. You provide the slug of the plugin to get the settings URL. This suggestion is intended to be paired with https://github.com/WordPress/performance/pull/1646/files#r1840919242 so that if you `GET /plugin-settings-url/foo` and there is no settings URL for that plugin Foo, then it returns a 404.

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,201 @@
...
+			)
+		);
+	}
+
+	return new WP_REST_Response(
+		array(
+			'success'           => true,
+			'pluginSettingsURL' => esc_url_raw( $plugin_settings_url ),
+		)
+	);
```
westonruter: Alternatively, this can just return the JSON string:

```suggestion
	return new WP_REST_Response( $plugin_settings_url );
```

plugins/performance-lab/includes/admin/rest-api.php
felixarntz: How do you mean? How is this a workaround?

westonruter: I mean, as I understand, a workaround like this is needed when the REST API endpoint is not a noun and/or the existing HTTP methods don't make semantic sense. According to [AIP-136](https://google.aip.dev/136) on custom methods:

> Resource-oriented design ([AIP-121](https://google.aip.dev/121)) uses custom methods to provide a means to express arbitrary actions that are difficult to model using only the standard methods.

What we discussed above in https://github.com/WordPress/performance/pull/1646/files#r1841160336 is to add a verb to the endpoint for use with `POST`. But here for the settings URL, it's a classic REST API endpoint, like most other REST API endpoints in WordPress, where we're just `GET` a noun. So so the normal REST design makes more sense to me.

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
b1ink0: Hi, There will be UX issue if settings URL try catch is moved after the changing DOM for the anchor tag / button. As it will become disabled after first request is finished signaling the user that the action is done but then the settings URL fetch request after some time if the plugin has settings URL then the DOM will be modified after some delay.

Also I think we should only send the settings URL request when the activate request is successful. 

Should I just move the settings URL try catch inside the activate plugin try catch and handle DOM of anchor tag / button after the settings URL try catch block?

https://github.com/user-attachments/assets/85a2850a-f396-409e-994e-73f5d1df05ae

plugins/performance-lab/includes/admin/rest-api.php
b1ink0: If we implement the `404` approach browser will log error in console for all the plugins which does not provide settings URL. As the browser logs error for `404` status code returned request even if we wrap it in try catch. How should I approach this issue?

plugins/performance-lab/includes/admin/rest-api.php
felixarntz: That's a good point. Though the current naming still seems incorrect in that it's the settings URL of a specific plugin. So a more appropriate name would be:
```suggestion
const PERFLAB_PLUGIN_SETTINGS_URL_ROUTE = '/plugins/(?P<slug>[a-z0-9_-]+)/settings-url';
```

This way it is a resource, but under the plugin it belongs to. Arguably with that approach, the 404 error if there's no settings URL makes more sense too.

plugins/performance-lab/includes/admin/rest-api.php
westonruter: Yeah, maybe a 404 is too noisy. The endpoint could just return `null` in that case as opposed to the URL string.

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
westonruter: Yeah, that works.

westonruter: Also, if the endpoint returns `null` instead of a 404, then no separate try/catch would be needed.

plugins/performance-lab/includes/admin/rest-api.php
westonruter:
```suggestion
		return new WP_REST_Response( null );
```

felixarntz: With `null` I have the concern that this may also be the result of a erroring JSON-decode operation.

Maybe we return an object with just `pluginSettingsURL` key, which is either the URL or `false`? That feels reasonable to me.

westonruter: JSON decode where? If in JavaScript then this is `JSON.parse()` throws an error when a parse error occurs. 

I just tried patching the comments endpoint to return bad JSON intentionally:

```diff
--- a/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
+++ b/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
@@ -422,6 +422,8 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 	 * @return WP_REST_Response|WP_Error Response object on success, or error object on failure.
 	 */
 	public function get_item( $request ) {
+		die( '{"BADJSON":' );
+
 		$comment = $this->get_comment( $request['id'] );
 		if ( is_wp_error( $comment ) ) {
 			return $comment;
```

Then if I do:

```js
await wp.apiFetch({ path: '/wp/v2/comments/1', method: 'GET' })
```

I get an JS error:

image

I don't feel strongly against wrapping the response in an object, however.

plugins/performance-lab/includes/admin/rest-api.php
b1ink0: @felixarntz I was just curious what should be the criteria for deciding whether to add the slug to the URL itself or include it in the body.

felixarntz: @b1ink0 It's one of the best practices for designing RESTful APIs, they should be centered around resources. See for example https://google.aip.dev/121 and https://google.aip.dev/122. (While these links are for Google's best practices, the best practices are not really specific to Google.)

By centering the API endpoints and their names around resources and their hierarchies, you future-proof the API to remain consistent and intuitive even when you add more and more features to it.

Here, the resource is a plugin. That's why using something like `/plugins/<slug>` as foundation for the routes is a good starting point. Such an endpoint could be used to get data for a specific plugin, or update data for a plugin for example.

Activating a plugin is not one of the CRUD operations though, so that's where a custom method is needed. See https://google.aip.dev/136 in that regard: That's how `/plugins/<slug>:activate` makes sense.

For example, if we later needed a method to deactivate a plugin, that would be straightforward via `/plugins/<slug>:deactivate`. If we used `/activate-plugin`, this would need to be `/deactivate-plugin`. That's still somewhat consistent but not centered around the resource and less intuitive. For example, nothing in the name tells you that you have to provide a slug. Additionally, even if you guessed that some identifier for the plugin would be needed, you wouldn't know whether it needs to be a parameter called `slug` or `id` or `identifier` for example. The resource identifier being part of the endpoint path makes that easier to work with.

felixarntz: @westonruter Following up on the other thread (since it probably makes sense to center that conversation about the endpoint naming): Per https://google.aip.dev/122, I think what would make most sense for the settings URL would be to simply be a value on the object returned by `GET /plugins/<slug>`.

A settings URL is not part of a collection, because a plugin does not have multiple settings URL. So `GET /plugins/<slug>/settings-url` would not be right. Because it's not an action (as you previously stated), `GET /plugins/<slug>:settings-url` wouldn't be right either.

The settings URL is one scalar piece of data for a plugin, so it makes most sense to be exposed on a plugin object. Obviously for this PR we don't care about creating a fully fledged `GET /plugins/<slug>` endpoint, but that's fine. We can simply for now return an object that e.g. only has `slug` and `settingsUrl` properties. The schema for such an object could be extended in the future in case we ever found a need for it, without backward compatibility breaks.

So I think going with `POST /plugins/<slug>:activate` and `GET /plugins/<slug>` (the latter of which return an object that includes the settings URL) is best in line with the API design guidelines.

b1ink0: Okey thank you!

plugins/performance-lab/includes/admin/rest-api.php
felixarntz: See https://github.com/WordPress/performance/pull/1646#discussion_r1842717861, discard my previous comment.

Let's continue that discussion on the other thread, to center the conversation about endpoint naming in one place.

plugins/performance-lab/includes/admin/rest-api.php
westonruter: @felixarntz there is already a [`plugins` endpoint](https://developer.wordpress.org/rest-api/reference/plugins/) in core, though. Perhaps we should add a new `perflab_settings_url` field to the plugin entity returned there?

westonruter: @felixarntz Additionally, there is also already a way to activate a plugin using the REST API by setting the `status` field  on the entity to `active`, without resorting to a special endpoint naming scheme: https://developer.wordpress.org/rest-api/reference/plugins/#update-a-plugin

However, reusing this endpoint is more complicated since we also automatically install the plugin prior to activation, as well as installing and activating any dependencies.

So maybe we should keep our own endpoint for installation and activation but instead of calling it `plugin:activate` we call it `feature:activate` to differentiate it from general plugin activation, as these plugins represent performance features. 

In the end, this API is almost guaranteed to be used 100% internally and so we don't really need to stress about it being getting it right forever.

felixarntz: I don't think the existing plugin endpoints in Core are a problem. We use a different namespace specific to Performance Lab, so I wouldn't worry about that. It's IMO not unreasonable to provide custom endpoints for similar use-cases under a different namespace when the Core endpoints do not satisfy the requirements we have.

Regarding updating a `status` field to `active`, that's how they chose to build it, but doesn't seem that great to me per https://google.aip.dev/216.

That said, I'd be okay using the terminology "features" instead of "plugins" for our endpoints, since that's also what we use in the UI, and I agree it can help differentiate.

felixarntz: So let's go with `POST /features/<slug>:activate` and `GET /features/<slug>` then?

westonruter: Where `GET /features/<slug>` returns an object with one key, `settingsUrl` which may be either a `string` or `null`, right? Sounds good.

felixarntz: Exactly.

If we like, we could also include e.g. `slug` in the response object as that's simple enough and clarifies the intent of the endpoint even more (that it's supposed to return plugin information, so not necessarily _just_ the settings URL).

plugins/performance-lab/includes/admin/plugin-activate-ajax.js
```diff
@@ -0,0 +1,87 @@
...
+				method: 'POST',
+			} );
+
+			// Fetch the plugin/feature information via the REST API.
+			const featureInfo = await apiFetch( {
+				path: `/performance-lab/v1/features/${ pluginSlug }`,
+				method: 'GET',
+			} );
+
+			if ( featureInfo?.settingsUrl ) {
```
westonruter: Nit: no need for the `?`:

```suggestion
			if ( featureInfo.settingsUrl ) {
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,167 @@
...
+ * @since n.e.x.t
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * Namespace for performance-lab REST API.
+ *
```
westonruter, line 15:
```suggestion
 *
 * @since n.e.x.t
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,167 @@
...
+ * @var string
+ */
+const PERFLAB_REST_API_NAMESPACE = 'performance-lab/v1';
+
+/**
+ * Route for activating plugin/feature.
+ *
+ * Note the `:activate` art of the endpoint follows Google's guidance in AIP-136 for the use of the POST method in a way
+ * that does not strictly follow the standard usage.
+ *
```
westonruter, line 26:
```suggestion
 *
 * @since n.e.x.t
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,167 @@
...
+ * @link https://google.aip.dev/136
+ * @var string
+ */
+const PERFLAB_FEATURES_ACTIVATE_ROUTE = '/features/(?P<slug>[a-z0-9_-]+):activate';
+
+/**
+ * Route for fetching plugin/feature information.
+ *
```
westonruter, line 35:
```suggestion
 *
 * @since n.e.x.t
```

westonruter (approved):
This looks great! Just a couple trivial suggestions, but I'm pre-approving.

plugins/performance-lab/includes/admin/load.php
```diff
@@ -228,8 +225,14 @@ function perflab_enqueue_features_page_scripts(): void {
 	wp_enqueue_style( 'thickbox' );
 	wp_enqueue_script( 'plugin-install' );
 
-	// Enqueue the a11y script.
-	wp_enqueue_script( 'wp-a11y' );
+	// Enqueue plugin activate AJAX script and localize script data.
+	wp_enqueue_script(
+		'perflab-plugin-activate-ajax',
+		plugins_url( 'includes/admin/plugin-activate-ajax.js', PERFLAB_MAIN_FILE ),
```
felixarntz: I think we should use `plugin_dir_url()`, as that's what we use elsewhere for this purpose. I think `plugins_url()` is a lower-level function and is in fact used in `plugin_dir_url()`.

If we change to use `plugins_url()`, this should be first discussed in #1647, instead of arbitrarily changing here.
```suggestion
		plugin_dir_url( PERFLAB_MAIN_FILE ) . 'includes/admin/plugin-activate-ajax.js',
```

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,170 @@
...
+		PERFLAB_REST_API_NAMESPACE,
+		PERFLAB_FEATURES_ACTIVATE_ROUTE,
+		array(
+			'methods'             => 'POST',
+			'args'                => array(
+				'slug' => array(
+					'type'              => 'string',
+					'description'       => __( 'Plugin slug of plugin/feature that needs to be activated.', 'performance-lab' ),
+					'required'          => true,
+					'validate_callback' => 'perflab_validate_slug_endpoint_arg',
```
felixarntz, line 58: Since this validates against the list of PL features/plugins, should we include an `enum` entry here with the possible values?

If that feels like too much, I think we should at least mention it in the description, e.g. "Must be one of the Performance Lab feature slugs."

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,170 @@
...
+		PERFLAB_REST_API_NAMESPACE,
+		PERFLAB_FEATURES_INFORMATION_ROUTE,
+		array(
+			'methods'             => 'GET',
+			'args'                => array(
+				'slug' => array(
+					'type'              => 'string',
+					'description'       => __( 'Plugin slug of plugin/feature whose information is needed.', 'performance-lab' ),
+					'required'          => true,
+					'validate_callback' => 'perflab_validate_slug_endpoint_arg',
```
felixarntz, line 83: See above.

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,170 @@
...
+				),
+			),
+			'callback'            => 'perflab_handle_feature_activation',
+			'permission_callback' => static function () {
+				if ( current_user_can( 'install_plugins' ) ) {
+					return true;
+				}
+
+				return new WP_Error( 'cannot_install_plugin', __( 'Sorry, you are not allowed to install plugins on this site.', 'performance-lab' ) );
```
felixarntz: Checking this is not enough. For example, if the plugin you're sending a request for is already activated, we'd need to check for `activate_plugin` instead. At the bare minimum, we would need to check for both, but that would still lead to problems, e.g. for sites that disable writing to the filesystem (which means `install_plugins` will be `false` for everyone).

Therefore we need to implement this check more granularly, taking into account the given `slug`:
* If the plugin is not installed, check for `install_plugins` and `activate_plugins`.
    * Ideally the latter should instead use `activate_plugin` with the plugin's basename, but we don't know that unless the plugin is installed.
* If the plugin is installed, check for `activate_plugin` with the plugin's `plugin_basename( $file )` value.

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,170 @@
...
+ * @param WP_REST_Request $request Request.
+ * @return WP_REST_Response|WP_Error Response.
+ */
+function perflab_handle_feature_activation( WP_REST_Request $request ) {
+	require_once ABSPATH . 'wp-admin/includes/plugin-install.php';
+	require_once ABSPATH . 'wp-admin/includes/class-wp-upgrader.php';
+	require_once ABSPATH . 'wp-admin/includes/class-wp-ajax-upgrader-skin.php';
+
+	// Install and activate the plugin/feature and its dependencies.
+	$result = perflab_install_and_activate_plugin( $request['slug'] );
```
felixarntz, line 136: Before calling this, I think we should check here whether the plugin is already active. If yes, we should return an error (probably with response code `400` since it's a bad request).

Alternatively (maybe even better), we handle that directly in the `perflab_install_and_activate_plugin()` function because it's more centralized to implement it there: Return an error if the plugin is already active.

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,170 @@
...
+	if ( is_wp_error( $result ) ) {
+		return new WP_Error(
+			'plugin_activation_failed',
+			$result->get_error_message(),
+			array( 'status' => 500 )
+		);
```
felixarntz: It's not quite right to use `500` here, since several of the errors returned from `perflab_install_and_activate_plugin()` are not 500 errors:
* `cannot_install_plugin` and `cannot_activate_plugin` are `rest_authorization_required_code()` kinds of errors.
* `plugin_not_found` is a 404 kind of error.

I would recommend that we include those error codes directly in the `WP_Error` objects returned from `perflab_install_and_activate_plugin()` and then pass them through as is here. There's no value in using a custom error code and status here, it only obfuscates the real error.

westonruter (approved):

felixarntz (commented):
@b1ink0 A few more points of follow up feedback. Some nit-picks, but most importantly we have to fix the capability checks.

plugins/performance-lab/includes/admin/rest-api.php
westonruter: I initially proposed using an `enum` here, but it won't work unless we then also always `require` the additional admin-only PHP files used in PerfLab. So what is implemented with this `validate_callback` is a sort of just-in-time enum. This is already explained in the phpdoc for `perflab_validate_slug_endpoint_arg()`.

plugins/performance-lab/includes/admin/rest-api.php
westonruter: Note that `perflab_install_and_activate_plugin()` already has capability checks for `install_plugins` and `activate_plugin`, so adding the these granular permission callbacks here would be redundant, wouldn't they? 

At most I think this `permission_callback` should add a `current_user_can( 'activate_plugins' )` check:

```suggestion
				// Important: The endpoint calls perflab_install_and_activate_plugin() which does more granular capability checks.
				if ( current_user_can( 'install_plugins' ) && current_user_can( 'activate_plugins' ) ) {
					return true;
				}

				return new WP_Error( 'cannot_activate', __( 'Sorry, you are not allowed to activate this feature.', 'performance-lab' ) );
```

plugins/performance-lab/includes/admin/rest-api.php
felixarntz: SG. In that case let's just mention it in the description.

plugins/performance-lab/includes/admin/rest-api.php
felixarntz: To some degree it would be redundant. But if we don't want to have any redundancy, we need to _only_ check for `activate_plugins` here, not `install_plugins` - since `activate_plugins` is the only capability that you need to perform this action in any case. `install_plugins` is only relevant if the plugin is not installed yet, and we wouldn't want to prevent people from activating an already installed plugin only because they cannot install.

plugins/performance-lab/includes/admin/rest-api.php
```diff
@@ -0,0 +1,170 @@
...
+function perflab_register_endpoint(): void {
+	register_rest_route(
+		PERFLAB_REST_API_NAMESPACE,
+		PERFLAB_FEATURES_ACTIVATE_ROUTE,
+		array(
+			'methods'             => 'POST',
+			'args'                => array(
+				'slug' => array(
+					'type'              => 'string',
+					'description'       => __( 'Plugin slug of plugin/feature that needs to be activated.', 'performance-lab' ),
```
westonruter:
```suggestion
					'description'       => __( 'Plugin slug of the Performance Lab feature to be activated.', 'performance-lab' ),
```

plugins/performance-lab/includes/admin/rest-api.php
westonruter: Like so? https://github.com/WordPress/performance/pull/1646/files#r1844278835

plugins/performance-lab/includes/admin/rest-api.php
westonruter: Sounds good, so:

```suggestion
				// Important: The endpoint calls perflab_install_and_activate_plugin() which does more granular capability checks.
				if ( current_user_can( 'activate_plugins' ) ) {
					return true;
				}

				return new WP_Error( 'cannot_activate', __( 'Sorry, you are not allowed to activate this feature.', 'performance-lab' ) );
```

plugins/performance-lab/includes/admin/rest-api.php
westonruter: What about just short-circuiting with a `200` response? It would be a bad user experience if someone has two instances of the Performance screen open and they activate features in one tab, and then later try to activate features in the other tab. When they try to do so on the other tab, they'd find they can't activate the features even though they are already active.

In fact, `perflab_install_and_activate_plugin()` will already essentially skip doing anything if the plugin is already installed and active, so I don't think anything needs to be done here.

plugins/performance-lab/includes/admin/rest-api.php
westonruter: Like so?

```suggestion
		switch ( $result->get_error_code() ) {
			case 'cannot_install_plugin':
			case 'cannot_activate_plugin':
				$response_code = rest_authorization_required_code();
				break;
			case 'plugin_not_found':
				$response_code = 404;
				break;
			default:
				$response_code = 500;
		}
		return new WP_Error(
			$result->get_error_code(),
			$result->get_error_message(),
			array( 'status' => $response_code )
		);
```

plugins/performance-lab/includes/admin/rest-api.php
felixarntz: Ok, that works too.

plugins/performance-lab/includes/admin/rest-api.php
felixarntz: That works, though we could alternatively adjust the called function itself to include response codes. But for the sake of time this is good too.

felixarntz (approved):
Thanks for all the work on this @b1ink0, and thanks @westonruter for the final iterations!

felixarntz:
@b1ink0 @westonruter I just tested this locally. While it works great for the most part, I get a 500 error when trying to install Performant Translations (probably because it's not installed): 

```
Uncaught Error: Call to undefined function request_filesystem_credentials() in \/var\/www\/html\/wp-admin\/includes\/class-wp-upgrader-skin.php:139
```

We may have missed to test this with plugins that _aren't_ yet installed. Looks like there are more files to require.

**Update:** Fixed in https://github.com/WordPress/performance/pull/1646/commits/e0691bf214cd2208574be8a76580155c14339b76.

PR #1650, opened 2024-11-12
Update plugin descriptions for Embed Optimizer and Image Prioritizer

westonruter:
I realized after deploying the README updates that the descriptions were stale for Image Prioritizer and Embed Optimizer.


felixarntz (approved):
Good catch!

PR #1613, opened 2024-10-21
Account for plugin dependencies when storing relevant plugin info

westonruter:
This is a follow-up hotfix to #1573 which broke the ability to install the Image Prioritizer plugin from the Performance features screen since the fetched plugin info did not include plugin dependencies, specifically Optimization Detective.

This also bumps the plugin version to 3.5.1.

PR #1651, opened 2024-11-13
Fix uses of 'Plugin not found' string

westonruter:
I [replied](https://wordpress.org/support/topic/plugin-not-found-9/#post-18130638) to a [comment](https://wordpress.org/support/topic/plugin-not-found-9/#post-18130597) from @gordonrjones (maybe the same GitHub username?) on the WordPress support forum:

> There are three possible locations where this error could be coming from:
> https://github.com/WordPress/performance/blob/7c6aa818b5c9707d7619a91614548ef51efba033/plugins/performance-lab/includes/admin/plugins.php#L28
> https://github.com/WordPress/performance/blob/7c6aa818b5c9707d7619a91614548ef51efba033/plugins/performance-lab/includes/admin/plugins.php#L86
> https://github.com/WordPress/performance/blob/7c6aa818b5c9707d7619a91614548ef51efba033/plugins/performance-lab/includes/admin/plugins.php#L363
> 
> All three are in that same file: includes/admin/plugins.php. Unfortunately we’re reusing the same string for each error. Are you comfortable editing the strings in the plugin file editor in WordPress to, for example, add `__LINE__` after the strings so we can differentiate between them? This will help us debug where the error is coming from.
> 
> Aside: I also see that we’re using the string with the default text domain once, but with the performance-lab text domain twice. That’s not good either. I’ll open a PR to fix this for the next release, as well as to make the strings unique.

This PR seeks to improve the usage of the strings, including their disambiguation.

The error message now includes an "error code" which is the line number from where the error originated:

image

### DIFF
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -359,7 +359,7 @@ function perflab_install_and_activate_plugin( string $plugin_slug, array &$proce
 		}
 
 		$plugins = get_plugins( '/' . $plugin_slug );
-		if ( empty( $plugins ) ) {
+		if ( count( $plugins ) === 0 ) {
 			return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'default' ) );
 		}
 
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -25,7 +25,7 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 	if ( is_array( $plugins ) ) {
 		// If the specific plugin_slug is not in the cache, return an error.
 		if ( ! isset( $plugins[ $plugin_slug ] ) ) {
-			return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'performance-lab' ) );
+			return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'default' ) );
 		}
 		return $plugins[ $plugin_slug ]; // Return cached plugin info if found.
 	}
@@ -83,7 +83,7 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 	set_transient( $transient_key, $plugins, HOUR_IN_SECONDS );
 
 	if ( ! isset( $plugins[ $plugin_slug ] ) ) {
-		return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'performance-lab' ) );
+		return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'default' ) );
 	}
 
 	/**
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -29,7 +29,7 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 				'plugin_not_found',
 				__( 'Plugin not found.', 'default' ) . ' ' .
 				/* translators: %s is error code */
-				sprintf( '(Error code: %d)', __LINE__ )
+				sprintf( '(Error line: %d)', __LINE__ )
 			);
 		}
 		return $plugins[ $plugin_slug ]; // Return cached plugin info if found.
@@ -92,7 +92,7 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 			'plugin_not_found',
 			__( 'Plugin not found.', 'default' ) . ' ' .
 			/* translators: %s is error code */
-			sprintf( '(Error code: %d)', __LINE__ )
+			sprintf( '(Error line: %d)', __LINE__ )
 		);
 	}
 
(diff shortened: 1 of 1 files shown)


plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -19,13 +19,18 @@
  * @return array{name: string, slug: string, short_description: string, requires: string|false, requires_php: string|false, requires_plugins: string[], download_link: string, version: string}|WP_Error Array of plugin data or WP_Error if failed.
  */
 function perflab_query_plugin_info( string $plugin_slug ) {
-	$transient_key = 'perflab_plugins_info-v2';
+	$transient_key = 'perflab_plugins_info';
```
westonruter, line 22: This was changed in https://github.com/WordPress/performance/pull/1613 to account for the format change. Since the transient only lasts for 1 day anyway, we can now restore the original transient key.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -359,8 +369,13 @@ function perflab_install_and_activate_plugin( string $plugin_slug, array &$proce
 		}
 
 		$plugins = get_plugins( '/' . $plugin_slug );
-		if ( empty( $plugins ) ) {
-			return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'default' ) );
+		if ( count( $plugins ) === 0 ) {
```
westonruter, line 368: The replacement of `empty()` with `count()` is part of #1219

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -19,13 +19,18 @@
...
 
 	if ( is_array( $plugins ) ) {
 		// If the specific plugin_slug is not in the cache, return an error.
 		if ( ! isset( $plugins[ $plugin_slug ] ) ) {
-			return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'performance-lab' ) );
+			return new WP_Error(
+				'plugin_not_found',
+				__( 'Plugin not found.', 'default' ) . ' ' .
+				/* translators: %s is error code */
+				sprintf( '(Error code: %d)', __LINE__ )
```
felixarntz: I find the "error code" terminology confusing. It makes me think this is either an HTTP response code or an actual error code I can look up in some docs. Why not say "line"?

westonruter: Sure. Done in 49cb889ca

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -19,13 +19,18 @@
...
 
 	if ( is_array( $plugins ) ) {
 		// If the specific plugin_slug is not in the cache, return an error.
 		if ( ! isset( $plugins[ $plugin_slug ] ) ) {
-			return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'performance-lab' ) );
+			return new WP_Error(
+				'plugin_not_found',
+				__( 'Plugin not found.', 'default' ) . ' ' .
+				/* translators: %s is error code */
+				sprintf( '(Error line: %d)', __LINE__ )
```
felixarntz: Better IMO:
```suggestion
				__( 'Plugin not found in cached API response.', 'performance-lab' )
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -83,7 +88,12 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 	set_transient( $transient_key, $plugins, HOUR_IN_SECONDS );
 
 	if ( ! isset( $plugins[ $plugin_slug ] ) ) {
-		return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'performance-lab' ) );
+		return new WP_Error(
+			'plugin_not_found',
+			__( 'Plugin not found.', 'default' ) . ' ' .
+			/* translators: %s is error code */
+			sprintf( '(Error line: %d)', __LINE__ )
```
felixarntz:
```suggestion
			__( 'Plugin not found in API response.', 'performance-lab' )
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -359,8 +369,13 @@ function perflab_install_and_activate_plugin( string $plugin_slug, array &$proce
...
 
 		$plugins = get_plugins( '/' . $plugin_slug );
-		if ( empty( $plugins ) ) {
-			return new WP_Error( 'plugin_not_found', __( 'Plugin not found.', 'default' ) );
+		if ( count( $plugins ) === 0 ) {
+			return new WP_Error(
+				'plugin_not_found',
+				__( 'Plugin not found.', 'default' ) . ' ' .
+				/* translators: %s is error code */
+				sprintf( '(Error line: %d)', __LINE__ )
```
felixarntz:
```suggestion
				__( 'Plugin not found among installed plugins.', 'performance-lab' )
```

felixarntz (commented):
@westonruter On second thought, I think it would be better to provide contextual error messages. Providing which line an error was triggered seems a bit too low-level and points to the messages themselves not being helpful enough. Error lines should come from stack traces etc., it doesn't feel right to include them in the messages arbitrarily.

This means we can't use a Core translation string anymore, but that's not necessarily a bad thing. @swissspidy had mentioned somewhere else before that that's a bit fragile to do anyway (e.g. because Core strings can change, outside of our control).

I left suggestions for ways to make the messages more helpful below.

felixarntz (approved):

westonruter:
Build for testing (3.6.0-alpha): [performance-lab.zip](https://github.com/user-attachments/files/17740381/performance-lab.zip)

PR #1653, opened 2024-11-13
Fix code coverage collection & reporting

### LINKED ISSUE #1284: Unit tests lack code coverage reporting and @covers annotations

sarthak-19:
## Summary

This PR fixes the currently broken automated codecov coverage in github workflow.

Fixes #1284

Problem Reference - https://github.com/WordPress/performance/pull/1652#issue-2656506027
## Additional Technical Details Required

As mentioned here - https://github.com/WordPress/performance/pull/1586#issuecomment-2474498387

> Additionally, the coverage report for each plugin is getting overwritten to the same XML file. There needs to be separate XML files generated for each plugin.
Lovekesh has some additional thoughts.

Currently the xml file name is being generated like `coverage-multisite-${{ github.sha }}.xml`, so if it's getting overwritten for each plugin, what different identifier can we use here so that it's unique for each plugin ?

cc : @thelovekesh @westonruter


sarthak-19:
The [test](https://github.com/WordPress/performance/actions/runs/11825633049/job/32949852740?pr=1653) is still failing, even after adding an additional arg seperator `--`, I ran a `ls -la` command to debug, the coverage file is not getting stored at root of the project.
Is that the cause or any other.

I'm not able to debug and stuck here, any help or insights would be instrumental. 

cc : @thelovekesh @westonruter

westonruter:
I see this https://github.com/WordPress/performance/actions/runs/11825633049/job/32949852740?pr=1653#step:12:49 which seems to be the problem:

> warning - 2024-11-13 20:58:58,206 -- Some files were not found --- {"not_found_files": ["coverage-6c3bb1148143cc38a09cb04f2cf7400b806d21c2.xml"]}

sarthak-19:
> I see this https://github.com/WordPress/performance/actions/runs/11825633049/job/32949852740?pr=1653#step:12:49 which seems to be the problem:
> 
> > warning - 2024-11-13 20:58:58,206 -- Some files were not found --- {"not_found_files": ["coverage-6c3bb1148143cc38a09cb04f2cf7400b806d21c2.xml"]}

Yes, there's no coverage file to upload and hence the test is failing, however the WorkFlow before uploading is running (where xml file is generated ) and the hash is also same I verified.
So is there a way to determine that if the file is properly getting generated or not or is this some directory path issue.
Maybe file is getting stored somewhere else, and I'm trying to upload the file from root of the project.

cc : @westonruter

westonruter:
I believe @thelovekesh has in mind what is wrong there.

.github/workflows/php-test-plugins.yml
```diff
@@ -71,17 +73,38 @@ jobs:
...
-            npm run test-php -- --coverage-clover=coverage-${{ github.sha }}.xml
+            npm run test-php:performance-lab -- -- -- --coverage-clover=coverage-performance-lab-${{ github.sha }}.xml
+            npm run test-php:auto-sizes -- -- -- --coverage-clover=coverage-auto-sizes-${{ github.sha }}.xml
+            npm run test-php:dominant-color-images -- -- -- --coverage-clover=coverage-dominant-color-images-${{ github.sha }}.xml
+            npm run test-php:embed-optimizer -- -- -- --coverage-clover=coverage-embed-optimizer-${{ github.sha }}.xml
+            npm run test-php:image-prioritizer -- -- -- --coverage-clover=coverage-image-prioritizer-${{ github.sha }}.xml
+            npm run test-php:optimization-detective -- -- -- --coverage-clover=coverage-optimization-detective-${{ github.sha }}.xml
+            npm run test-php:speculation-rules -- -- -- --coverage-clover=coverage-speculation-rules-${{ github.sha }}.xml
+            npm run test-php:web-worker-offloading -- -- -- --coverage-clover=coverage-web-worker-offloading-${{ github.sha }}.xml
+            npm run test-php:webp-uploads -- -- -- --coverage-clover=coverage-webp-uploads-${{ github.sha }}.xml
```
thelovekesh: @sarthak-19 What's the use case behind including SHA here?

phpunit.xml.dist
```diff
@@ -41,4 +41,9 @@
 			<group>ms-required</group>
 		</exclude>
 	</groups>
+	<coverage>
+    <include>
+        <directory suffix=".php">plugins</directory>
+    </include>
+</coverage>
```
thelovekesh: Seems like indentation is off here?

Also @sarthak-19, have you tried generating HTML coverage reports with this config? My speculation is, this config will generate coverage reports in the respect of whole codebase in the plugins directory instead in the respect of a single plugin.

cc @westonruter

package.json
```diff
@@ -63,7 +63,16 @@
...
+    "test-php-multisite:embed-optimizer": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test-multisite:embed-optimizer",
+    "test-php-multisite:image-prioritizer": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test-multisite:image-prioritizer",
+    "test-php-multisite:optimization-detective": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test-multisite:optimization-detective",
+    "test-php-multisite:speculation-rules": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test-multisite:speculation-rules",
+    "test-php-multisite:web-worker-offloading": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test-multisite:web-worker-offloading",
+    "test-php-multisite:webp-uploads": "wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/performance composer test-multisite:webp-uploads",
+    "wp-env": "wp-env start --xdebug=develop,coverage",
```
thelovekesh: Do we really need these commands?

.github/workflows/php-test-plugins.yml
sarthak-19: I added that as a [suggestion](https://github.com/WordPress/performance/pull/1586#discussion_r1834718375 ) by [felixarntz](https://github.com/felixarntz), but removed since it was causing error in file upload to codecov.

phpunit.xml.dist
sarthak-19: > Seems like indentation is off here?
I'll fix this.

> have you tried generating HTML coverage reports with this config? My speculation is, this config will generate coverage reports in the respect of whole codebase in the plugins directory instead in the respect of a single plugin.

No I have just tried html generation for single plugin, I'll try once what you are suggesting and will update here whatever seems a viable approach.

package.json
sarthak-19: As suggested by [westonruter](https://github.com/westonruter), we'll have to run each test command seperately for each plugin that was already added in the `composer.json`, so I thought the same will applyt for multisite as well.
Am I missing something here?

sarthak-19:
### Issue with Codecov upload report job : 

Currently Creating report in codecov is failing : 
```
==> Running create-report
      ./codecov  create-report --fail-on-error --git-service github --sha c56fc50c9c306134344d9b64e5d00d789b90dc64
info - 2024-12-03 17:56:44,090 -- ci service found: github-actions
info - 2024-12-03 17:56:44,464 -- Process Report creating complete
error - 2024-12-03 17:56:44,464 -- Report creating failed: {"message":"Token required because branch is protected"}
==> Failed to create-report
```
Latest Commit : https://github.com/WordPress/performance/actions/runs/12145298761/job/33866607688?pr=1653


However it was running fine in previous commits : 
- https://github.com/WordPress/performance/actions/runs/12145028699/job/33865747938
- https://github.com/WordPress/performance/actions/runs/12144752300/job/33864838801

```
==> Running create-report
      ./codecov  create-report --fail-on-error --git-service github --sha 72183d4acc81ffb64510050ff793f4bbfa568427
info - 2024-12-03 17:40:20,844 -- ci service found: github-actions
info - 2024-12-03 17:40:21,379 -- Process Report creating complete
info - 2024-12-03 17:40:21,379 -- Finished creating report successfully --- {"response": "{\"status\":\"queued\"}\n"}
```

Error says token is required, but previously it was working, any idea why it might be the case here ?

cc : @thelovekesh @westonruter

westonruter:
Seems like it is working now? Perhaps it had something to do with the PR coming from a fork and so the token isn't available.

composer.json
```diff
@@ -40,7 +40,7 @@
       "phpstan/extension-installer": true
     },
     "platform": {
-      "php": "7.2"
+      "php": "7.3"
```
westonruter: Why is this necessary? The plugin requires PHP 7.2, which is still being tested. Doesn't this cause problems when trying to do `composer install`?

.github/workflows/php-test-plugins.yml
```diff
@@ -68,20 +73,51 @@ jobs:
...
+            npm run wp-env start
+          fi
       - name: Composer Install
         run: npm run wp-env run tests-cli -- --env-cwd="wp-content/plugins/$(basename $(pwd))" composer install --no-interaction --no-progress
+      - name: Update Composer Dependencies
+        run: composer update --with-all-dependencies --no-interaction --no-progress
+      - name: Install PHPUnit
+        run: |
+          if [ "${{ matrix.php }}" == "7.2" ]; then
+            composer require --dev phpunit/phpunit:${{ matrix.phpunit }} --with-all-dependencies
```
westonruter: This is needed because `composer.lock` may have dependencies which require PHP 7.3?

.github/workflows/php-test-plugins.yml
```diff
@@ -68,20 +73,51 @@ jobs:
...
+          else
+            composer require --dev phpunit/phpunit:${{ matrix.phpunit }}
+          fi
+        if: matrix.phpunit != ''
```
westonruter, line 89: Nice, so if the matrix doesn't specify the specific version of PHPUnit then it defaults to whatever is in `composer.json`.

westonruter:
The props bot action is failing due to `codecov` commenting on the issue. Can we exclude it? Or does it need to be added to the list of [`skippedUsers`](https://github.com/WordPress/props-bot-action/blob/46bf335d082b0c77c91118d8f3e704d4304830fd/src/contribution-collector.js#L246-L250) in the action itself?

westonruter:
> Or does it need to be added to the list of [`skippedUsers`](https://github.com/WordPress/props-bot-action/blob/46bf335d082b0c77c91118d8f3e704d4304830fd/src/contribution-collector.js#L246-L250) in the action itself?

I've opened https://github.com/WordPress/props-bot-action/pull/141 to implement this.

composer.json
sarthak-19: When installing PHPUnit, I was getting the following errors : 
```
Problem 1
    - Root composer.json requires php ^7.3 || ^8.0 but your php version (7.2; overridden via config.platform, actual: 8.3.2) does not satisfy that requirement.
  Problem 2
    - phpunit/phpunit[9.0.0, ..., 9.2.6] require php ^7.3 -> your php version (7.2; overridden via config.platform, actual: 8.3.2) does not satisfy that requirement.
    - phpunit/phpunit[9.3.0, ..., 9.3.8] require php ^7.3 || ^8.0 -> your php version (7.2; overridden via config.platform, actual: 8.3.2) does not satisfy that requirement.
    - phpunit/phpunit[9.3.9, ..., 9.6.21] require php >=7.3 -> your php version (7.2; overridden via config.platform, actual: 8.3.2) does not satisfy that requirement.
```

So after changing the plafrom version to `7.3` the issue got fixed.

cc : @westonruter

.github/workflows/php-test-plugins.yml
sarthak-19: Was getting the following errors.

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires phpunit/phpunit 7.5 -> satisfiable by phpunit/phpunit[7.5.0].
    - phpunit/phpunit 7.5.0 requires phar-io/manifest ^1.0.2 -> found phar-io/manifest[1.0.2, 1.0.3] but the package is fixed to 2.0.4 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
  Problem 2
    - yoast/phpunit-polyfills is locked to version 2.0.2 and an update of this package was not requested.
    - phpunit/phpunit 7.5.0 requires phar-io/manifest ^1.0.2 -> found phar-io/manifest[1.0.2, 1.0.3] but the package is fixed to 2.0.4 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
    - yoast/phpunit-polyfills 2.0.2 requires phpunit/phpunit ^5.7.21 || ^6.0 || ^7.0 || ^8.0 || ^9.0 || ^10.0 -> satisfiable by phpunit/phpunit[7.5.0].

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.
```

Since the above needed platform ^7.3 and here when PHP 7.2 is being used in testing I thought maybe that's the issue so I moved forward with ` --with-all-dependencies` this flag as a solution.

.github/workflows/php-test-plugins.yml
sarthak-19: Yes, this was the case previously when default PHPUnit version 10 was getting used in all the environment. 😅

composer.json
swissspidy: We should be able to keep this at 7.2, `yoast/phpunit-polyfills` can then install whatever PHPUnit version that's supported on the current PHP version. See https://github.com/Yoast/PHPUnit-Polyfills

phpunit.xml.dist
```diff
@@ -41,4 +41,9 @@
 			<group>ms-required</group>
 		</exclude>
 	</groups>
+	<coverage>
+		<include>
+        	<directory suffix=".php">plugins</directory>
+    	</include>
```
swissspidy: Note the mixed use of tabs and spaces

.github/workflows/php-test-plugins.yml
swissspidy: With `yoast/phpunit-polyfills` we can keep platform at 7.2 and then should be able to use `composer update --dev yoast/phpunit-polyfills --with-dependencies` here instead

westonruter:
I re-ran the Props Bot job that had failed due to `codecov` not being recognized where my [PR](https://github.com/WordPress/props-bot-action/pull/141) to fix this has been merged. The job is now passing. ✅

sarthak-19:
I'll fix the platform ^7.2 issue and indentation issue soon, then we it'll be completed. 😄

.github/workflows/php-test-plugins.yml
```diff
@@ -90,7 +124,7 @@ jobs:
         uses: codecov/codecov-action@v5
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          files: coverage-${{ github.sha }}.xml
+          file: ./single-site-reports
```
westonruter: Should this rather be `files` instead of `file`? I see this [mentioned](https://github.com/codecov/codecov-action?tab=readme-ov-file#migration-guide):

image


```suggestion
          files: ./single-site-reports
```

.github/workflows/php-test-plugins.yml
```diff
@@ -99,7 +133,7 @@ jobs:
         uses: codecov/codecov-action@v5
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          files: coverage-multisite-${{ github.sha }}.xml
+          file: ./multisite-reports
```
westonruter:
```suggestion
          directory: ./multisite-reports
```

.github/workflows/php-test-plugins.yml
westonruter: Oh, but actually should this not rather be `directory`? I see that `files` is [defined](https://github.com/codecov/codecov-action?tab=readme-ov-file#migration-guide:~:text=Comma%2Dseparated%20explicit%20list%20of%20files%20to%20upload.%20These%20will%20be%20added%20to%20the%20coverage%20files%20found%20for%20upload.%20If%20you%20wish%20to%20only%20upload%20the%20specified%20files%2C%20please%20consider%20using%20%22disable%2Dsearch%22%20to%20disable%20uploading%20other%20files.) as:

> Comma-separated explicit list of files to upload. These will be added to the coverage files found for upload. If you wish to only upload the specified files, please consider using "disable-search" to disable uploading other files.	

Whereas `directory` is defined as:

> Folder to search for coverage files. Default to the current working directory	

This seems to be what is intended here.

```suggestion
          directory: ./single-site-reports
```

.github/workflows/php-test-plugins.yml
```diff
@@ -45,13 +45,16 @@ jobs:
         wp: [ 'latest' ]
         coverage: [false]
         include:
+          - php: '7.2'
+            wp: 'latest'
```
westonruter: Is this needed since the `matrix` already has `7.2` and `wp` as `latest`?

```suggestion
```

.github/workflows/php-test-plugins.yml
```diff
@@ -68,20 +71,51 @@ jobs:
       - name: Build assets
         run: npm run build
       - name: Install WordPress
-        run: npm run wp-env start
+        run: |
+          if [ "${{ matrix.coverage }}" == "true" ]; then
+            npm run wp-env start -- --xdebug=develop,coverage
```
westonruter: Is `develop` a value for `--xdebug`?  I only see `profile`, `trace`, and `debug` listed: https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env/#using-xdebug

westonruter (changes requested):

.github/workflows/php-test-plugins.yml
sarthak-19: No, now it's not needed, previously was setting Phpunit version also that's why it was needed.
Removing it.

.github/workflows/php-test-plugins.yml
sarthak-19: Yes `develop` is also a value for `--xdebug` along with `coverage` as mentioned here : https://xdebug.org/docs/code_coverage#mode 

But we no longer need `develop` so removing it.

cc : @westonruter

westonruter (approved):
Looks great! I'll defer to @felixarntz or @swissspidy to give final approval and merge.

swissspidy (approved):

westonruter:
When running tests locally in wp-env I'm now seeing this warning:

```
  Warning - The configuration file did not pass validation!
  The following problems have been detected:

  Line 44:
  - Element 'coverage': This element is not expected.

  Test results may not be as expected.
```

I guess this is because it is running PHPUnit 8.5.39 which is a version prior to this element being introduced.

westonruter:
Note: I noticed that code coverage was being reported for test files, so I've excluded them in ca0ca13.

Yedfree (approved):

PR #1654, opened 2024-11-15
Incorporate media queries into preconnect links to account for whether embeds are in viewport

### LINKED ISSUE #1341: Incorporate breakpoints into preconnect links added by Embed Optimizer
> It would be slightly better if the viewport widths were also taken into account, so that an embed only visible on desktop would not get preconnected on mobile. This can be addressed in a follow-up PR.

_Originally posted by @westonruter in https://github.com/WordPress/performance/pull/1302#discussion_r1670865833_

Code in question:

https://github.com/WordPress/performance/blob/84bf4ee2e1033751fdc16beaad1faaeb57e1da5c/plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php#L89-L96

ShyamGadde:
## Summary

Fixes #1341 

For example, if an embed is not visible on mobile:
image

## Relevant technical choices

- **Sorting and Merging of Links**: The `href` attributes of preconnect links are sorted to group identical URLs together. If the URLs are identical, they are further sorted by `minimum_viewport_width`. This strategy enables merging of consecutive links, reducing redundancy. Refer to the discussion in https://github.com/WordPress/performance/issues/1341#issuecomment-2477466495.
- Introduced `OD_URL_Metric_Group::get_xpath_elements_map()`,  `OD_URL_Metric_Group::get_all_element_max_intersection_ratios()` and `OD_URL_Metric_Group::get_element_max_intersection_ratio()` methods to help determine if a link should get preconnected on that particular viewport group.
- But the lazy load script is injected only when an embed element is not visible in any viewport group.


plugins/optimization-detective/helper.php
```diff
@@ -37,7 +37,7 @@ function od_initialize_extensions(): void {
  */
 function od_generate_media_query( ?int $minimum_viewport_width, ?int $maximum_viewport_width ): ?string {
 	if ( is_int( $minimum_viewport_width ) && is_int( $maximum_viewport_width ) && $minimum_viewport_width > $maximum_viewport_width ) {
-		_doing_it_wrong( __FUNCTION__, esc_html__( 'The minimum width must be greater than the maximum width.', 'optimization-detective' ), 'Optimization Detective 0.7.0' );
+		_doing_it_wrong( __FUNCTION__, esc_html__( 'The minimum width cannot be greater than the maximum width.', 'optimization-detective' ), 'Optimization Detective 0.7.0' );
```
westonruter, line 40: Right 😅

plugins/optimization-detective/class-od-link-collection.php
```diff
@@ -130,18 +130,29 @@ function ( array $links ): array {
...
 			 * @param Link $b Second link.
 			 * @return int Comparison result.
 			 */
 			static function ( array $a, array $b ): int {
-				return $a['minimum_viewport_width'] <=> $b['minimum_viewport_width'];
+				// Get href values, defaulting to empty string if not present.
+				$href_a = $a['attributes']['href'] ?? '';
+				$href_b = $b['attributes']['href'] ?? '';
+
+				$href_comparison = strcmp( $href_a, $href_b );
```
westonruter, line 150: This would be the same as:

```php
$href_comparison = ( $href_a <=> $href_b );
```

Right?

ShyamGadde: I believe in this case it would be the same.

westonruter:
FYI: The screenshot is broken.

plugins/optimization-detective/class-od-url-metric-group.php
```diff
@@ -320,6 +320,62 @@ public function get_lcp_element(): ?OD_Element {
...
 
+	/**
+	 * Gets all elements from all URL Metrics in the viewport group keyed by the elements' XPaths.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array<string, non-empty-array<int, OD_Element>> Keys are XPaths and values are the element instances.
+	 */
+	public function get_xpath_elements_map(): array {
+		$result = ( function () {
```
westonruter, line 337: The purpose of the closure is to facilitate caching the value. So there should be new keys added to:

https://github.com/WordPress/performance/blob/85d4fd4aa9729543f7fb4a3fc412404a4a85e065/plugins/optimization-detective/class-od-url-metric-group.php#L70-L78

for `get_xpath_elements_map` and `get_all_element_max_intersection_ratios`.

And then these methods can start with:

https://github.com/WordPress/performance/blob/85d4fd4aa9729543f7fb4a3fc412404a4a85e065/plugins/optimization-detective/class-od-url-metric-group.php#L224-L226

And end with:

https://github.com/WordPress/performance/blob/85d4fd4aa9729543f7fb4a3fc412404a4a85e065/plugins/optimization-detective/class-od-url-metric-group.php#L242-L243

plugins/optimization-detective/class-od-url-metric-group.php
```diff
@@ -320,6 +320,62 @@ public function get_lcp_element(): ?OD_Element {
...
+	}
+
+	/**
+	 * Gets the max intersection ratios of all elements in the viewport group and its captured URL Metrics.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array<string, float> Keys are XPaths and values are the intersection ratios.
+	 */
+	public function get_all_element_max_intersection_ratios(): array {
```
westonruter, line 358: There is an existing `get_all_element_max_intersection_ratios` on the `OD_URL_Metric_Group_Collection` class. It could be refactored to instead loop over all of `groups` to then call this method and merge the results.

westonruter (changes requested):
Looking good!

Added some feedback.

Let's also add a test case specifically for when there is an embed visible on a desktop viewport but hidden on a mobile viewport.

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -171,12 +173,20 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+
+						$context->link_collection->add_link(
+							array(
+								'rel'  => 'preconnect',
+								'href' => $preconnect_href,
+							),
+							$group->get_minimum_viewport_width(),
+							$group->get_maximum_viewport_width()
+						);
+					}
```
westonruter: The analogous logic that adds preload links for the LCP image:

https://github.com/WordPress/performance/blob/df29967bfc5eed55a35606595df6344bb4ca9f8c/plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php#L146-L180

For that we use `$context->url_metric_group_collection->get_groups_by_lcp_element()` to get the groups, whereas here it is looping over each group in the collection and then checking for each one whether the current embed is visible in it. So this makes sense to me.

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
```diff
@@ -171,12 +173,20 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
 
 				foreach ( $preconnect_hrefs as $preconnect_href ) {
-					$context->link_collection->add_link(
-						array(
-							'rel'  => 'preconnect',
-							'href' => $preconnect_href,
-						)
-					);
+					foreach ( $context->url_metric_group_collection as $group ) {
+						if ( $group->get_element_max_intersection_ratio( $embed_wrapper_xpath ) <= 0 ) {
```
westonruter: Can the intersection ratio ever be less than 0? I don't think so, and [Gemini confirms](https://g.co/gemini/share/d7c2bfb37ad1). Since we're comparing floats, it's probably not safe to check if the return value is exactly equal to 0. How about:

```suggestion
						if ( ! ( $group->get_element_max_intersection_ratio( $embed_wrapper_xpath ) > 0.0 ) ) {
```

plugins/optimization-detective/class-od-url-metric-group.php
ShyamGadde: Done in c22701f

plugins/optimization-detective/class-od-url-metric-group.php
ShyamGadde: Made the changes in aa06707. Also refactored the `get_xpath_elements_map` method in a similar fashion.

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
ShyamGadde: You are right. Changed it in 5ecfcf2.

plugins/embed-optimizer/class-embed-optimizer-tag-visitor.php
ShyamGadde: I had a thought: would it make sense to refactor this into a method like `get_groups_by_visible_element`? This could encapsulate the logic for finding groups where the element is visible, potentially making the code more modular and readable. However, it does introduce an extra function call. Would you recommend this, or is the current approach preferable?

ShyamGadde: I noticed that the analogous code includes `$link_attributes['media'] = 'screen';`. However, based on a [previous comment](https://github.com/WordPress/performance/pull/1302#discussion_r1670902233) you made on another PR, I assume that specifying 'screen' as the media type might not be appropriate in this context. Am I correct in thinking that?

westonruter (approved):
Great work!

westonruter:
In testing this I found a small issue preventing `preconnect` links from being added unless URL Metrics are collected for _both_ desktop and mobile: https://github.com/WordPress/performance/issues/1757

PR #1656, opened 2024-11-15
Improve debugging stored URL Metrics in Optimization Detective

westonruter:
While all of the URL Metrics are dumped out as part of the collection, it makes it difficult to drill down to see what the most recently-stored URL Metrics are. This is improved here, including a `creationDate` property which includes a `Date` instance of the `timestamp`:

image


westonruter:
@felixarntz Does this look good to you?

adamsilverstein (approved):
Nifty!

PR #1657, opened 2024-11-16
Prepare 3.6.0 release

westonruter:
See #1645
Previously #1609

- **Bump versions for 3.6.0 release**
- **Run npm run since**
- **Run npm run readme**

The following plugins are included in this release:

1. performance-lab 3.6.0
2. optimization-detective v0.8.0
3. webp-uploads v2.3.0

PR #1662, opened 2024-11-18
Rename `webp_uploads_get_file_mime_type` to `webp_uploads_get_attachment_file_mime_type` to clarify scope

mukeshpanchal27:
## Summary

These PR address the feedback for https://github.com/WordPress/performance/pull/1642#discussion_r1836955274.

### DIFF
--- a/plugins/webp-uploads/helper.php
+++ b/plugins/webp-uploads/helper.php
@@ -488,7 +488,7 @@ function webp_uploads_get_attachment_file_mime_type( int $attachment_id, string 
 	if ( '' === $file ) {
 		$file = get_attached_file( $attachment_id, true );
 		// File does not exist.
-		if ( false === $file || ! file_exists( $file ) ) {
+		if ( false === $file ) {
 			return '';
 		}
 	}
--- a/plugins/webp-uploads/helper.php
+++ b/plugins/webp-uploads/helper.php
@@ -479,11 +479,19 @@ function webp_uploads_get_mime_type_image( int $attachment_id, string $src, stri
  *
  * @since n.e.x.t
  *
- * @param string $file          The path to the file.
  * @param int    $attachment_id The attachment ID.
+ * @param string $file          Optional. The path to the file.
  * @return string The MIME type of the file, or an empty string if not found.
  */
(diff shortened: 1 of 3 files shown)


mukeshpanchal27:
It would be good to include these changes in the 2.3.0 release #1657 of the Modern Image Formats plugin.

plugins/webp-uploads/helper.php
```diff
@@ -479,11 +479,19 @@ function webp_uploads_get_mime_type_image( int $attachment_id, string $src, stri
...
  * @param int    $attachment_id The attachment ID.
+ * @param string $file          Optional. The path to the file.
  * @return string The MIME type of the file, or an empty string if not found.
  */
-function webp_uploads_get_file_mime_type( string $file, int $attachment_id ): string {
+function webp_uploads_get_attachment_file_mime_type( int $attachment_id, string $file = '' ): string {
+	if ( '' === $file ) {
+		$file = get_attached_file( $attachment_id, true );
+		// File does not exist.
+		if ( false === $file || ! file_exists( $file ) ) {
```
felixarntz: This function doesn't do anything with the actual file contents, so the second check here is irrelevant. It would also be inconsistent to only do it when no `$file` was provided. We can remove it.
```suggestion
		if ( false === $file ) {
```

plugins/webp-uploads/picture-element.php
```diff
@@ -23,13 +23,7 @@ function webp_uploads_wrap_image_in_picture( string $image, string $context, int
 		return $image;
 	}
 
-	$file = get_attached_file( $attachment_id, true );
-	// File does not exist.
-	if ( false === $file || ! file_exists( $file ) ) {
```
felixarntz, line 28: Related to the above, the `file_exists()` check was never needed here anyway, so this is okay to be removed.

felixarntz (commented):
Thanks @mukeshpanchal27, this looks almost good to go.

I agree this should go into today's release, as that's otherwise introducing the function under the wrong name.

felixarntz (approved):

joemcgill (approved):

PR #1666, opened 2024-11-18
Only check plugin version integrity in deployment workflow if the plugin actually requires to be deployed

felixarntz:
## Summary

See #1664: This PR attempts to properly fix the underlying issue: We shouldn't check the version integrity of plugins that do not need deployment during the deployment workflow.

Therefore, this PR moves that check after the conditional for whether a deployment is needed. This will still catch problems for any plugins that need to be deployed, but ignore problems from other plugins (which may temporarily exist, and that's okay).

The PR also effectively reverts the short-term fix from #1664, re-including those future changelog entries, as they should no longer be a problem after this.


westonruter (approved):

.github/workflows/deploy-plugins.yml
```diff
@@ -101,6 +98,10 @@ jobs:
 
           echo "deploy=true" >> $GITHUB_OUTPUT
 
+      - name: Check plugin version integrity
+        if: steps.check-deployment.outputs.deploy == 'true'
```
mukeshpanchal27, line 102: Not blocker: Instead of adding `if: steps.check-deployment.outputs.deploy == 'true'` condition in every steps can we add new step if output is not true the exit and remove condition from all other steps below?

mukeshpanchal27 (approved):
Thanks for the PR. LGTM. Left one non-blocking feedback.

.github/workflows/deploy-plugins.yml
felixarntz: Not sure how that would work in GitHub Actions. Maybe it's possible, but I'm not sure there's a way to properly exit out of a GitHub workflow. Do you have an idea about it?

mukeshpanchal27: We can add the below new step after the `Check if deployment is needed` step and we can remove the `if: steps.check-deployment.outputs.deploy == 'true'` condition from all other steps after adding these step.

```php
- name: Fail early if deployment check is false
  if: steps.check-deployment.outputs.deploy != 'true'
  run: |
    echo "Deployment condition not met. Exiting."
    exit 1
```

I did test in my folk repo - https://github.com/mukeshpanchal27/performance/actions/runs/11926864996/job/33241344801

felixarntz: @swissspidy Curious what's your take on the idea of early returns/exits in GitHub workflows.

mukeshpanchal27:
@felixarntz Let's merge these PRs. We can always open a follow-up if needed, as https://github.com/WordPress/performance/pull/1666#discussion_r1848084465 is not a blocker.

.github/workflows/deploy-plugins.yml
felixarntz: Ping @swissspidy :)

felixarntz:
@mukeshpanchal27 Agreed. Happy for this to be merged, given it has two approvals. I pinged Pascal separately, still curious on his thoughts given his CI experience.

.github/workflows/deploy-plugins.yml
swissspidy: Thanks for the ping. I was on vacation when you sent the first message.

It depends on what you want to achieve.

If you `exit 1` early that will mark the whole job as failed. So the question is whether that is desirable.

Also depends on if there are any other steps after the exit 1 that still should be executed. Then obviously you can't exit early.

IMO the behavior at the moment is quite nice. The jobs are marked as passed, but you still get the notice saying no deployment was needed:

Screenshot 2024-12-18 at 09 57 41

Whereas with the suggested early exit you'll get a failure status and my first instinct is "oh oh, something is broken" and not "oh, this is just a skipped plugin"

felixarntz: That's a great point, and I agree. We shouldn't exit with a failure code, as skipping the deployment because it's not needed is indeed not a failure.

mukeshpanchal27 (approved):

PR #1675, opened 2024-11-19
Fix race condition bug where activating multiple features sequentially could fail to activate some features

### LINKED ISSUE #1660: Activating multiple features / plugins sequentially fails to activate some features / plugins
## Bug Description

Following the implementation of #1583, Performance features are now installed and activated via AJAX. However, during smoke testing, I observed issues when activating multiple features / plugins sequentially.

### Steps to Reproduce

1. Activate `Modern Image Formats`, `Embed Optimizer`, and `Web Worker Offloading` one by one.
2. Check the status on the Installed Plugins page and the Performance Features page.

### Observed Behaviour

- On the Installed Plugins page:
  - `Embed Optimizer` and its dependent `Optimization Detective` plugins are not marked as activated.
- On the Performance Features page:
  - Initially, `Embed Optimizer` appears as activated. However, after reloading the page, its state changes to Activate.

### Additional Scenario
- When activating all performance features / plugins one by one, some features remain only installed and not activated.

### Screen Recording & Screenshots:

https://github.com/user-attachments/assets/83f359ed-345d-479c-9abf-e5f779139662

Image
Image

b1ink0:
## Summary

Fixes #1660 

## Relevant technical choices

Implemented a queue based approach for activation of features / plugins using the AJAX in JavaScript.

Now only one features / plugin sends a request at a time other request are processed after previous request is finished.

## Demo

https://github.com/user-attachments/assets/6d6c3b9d-6f33-4b4b-a882-aa3d6c1f7fe1

### DIFF
--- a/plugins/performance-lab/includes/admin/plugin-activate-ajax.js
+++ b/plugins/performance-lab/includes/admin/plugin-activate-ajax.js
@@ -7,21 +7,20 @@
 	const { i18n, a11y, apiFetch } = wp;
 	const { __ } = i18n;
 
-	/**
-	 * Handles click events on elements with the class 'perflab-install-active-plugin'.
-	 *
-	 * This asynchronous function listens for click events on the document and executes
-	 * the provided callback function if triggered.
-	 *
-	 * @param {MouseEvent} event - The click event object that is triggered when the user clicks on the document.
-	 *
-	 * @return {Promise<void>} The asynchronous function returns a promise that resolves to void.
-	 */
-	async function handlePluginActivationClick( event ) {
-		const target = /** @type {HTMLElement} */ ( event.target );
+	// Queue to hold pending activation requests.
+	const activationQueue = [];
+	let isProcessingActivation = false;
 
+	/**
+	 * Enqueues plugin activation requests and starts processing if not already in progress.
+	 *
+	 * @param {MouseEvent} event - The click event object.
+	 */
+	function enqueuePluginActivation( event ) {
 		// Prevent the default link behavior.
 		event.preventDefault();
+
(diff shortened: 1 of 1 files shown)


mukeshpanchal27 (dismissed):
Thanks @b1ink0 for the PR.

When plugins are activated simultaneously, while some plugins in the `waiting` state reloading the settings page. After the reload, some plugins display the `Activate` button again, even though they are already installed and activated, while others remain unactivated.  

This is not a blocker but something I observed during testing.


https://github.com/user-attachments/assets/e8c85ccd-63a6-4ef8-83ca-4523b1021649

b1ink0:
> 
> When plugins are activated simultaneously, while some plugins in the `waiting` state reloading the settings page. After the reload, some plugins display the `Activate` button again, even though they are already installed and activated, while others remain unactivated.
> 
@mukeshpanchal27 
I believe this is happening because when the plugin is in the `Activating...` state, the request to activate the plugin is sent. If you reload quickly, it will show `Activate` because the request to activate the plugin is not finished. This means the plugin is not activated yet, which correctly reflects on the Performance Features page. Then, when you later go to the plugins page, it shows the plugins as activated because the request to activate the plugin has now finished.




https://github.com/user-attachments/assets/07810f92-7833-44f4-a2d2-c8520000bdbc

westonruter (approved):
LGTM

felixarntz:
FYI I've changed the base branch to a new `release/3.6.1`, since this should go into a hot fix release for purely this fix.

felixarntz (approved):
@b1ink0 This looks great, thank you for the fix and fast turnaround time!

PR #1679, opened 2024-11-20
Automatically opt into 1536x1536 and 2048x2048 sizes when generating fallback images

### LINKED ISSUE #1512: Automatically opt into 1536x1536 and 2048x2048 sizes when generating fallback images
## Feature Description

As discussed in https://github.com/WordPress/performance/pull/1421 when "generate fallback images" is enabled, we should generate all of the sizes that core generates by default in both mime types (the uploaded type which could be JPEG or PNG and the output type which could be AVIF or WebP). 

This includes two "extra" sizes that were added to the image generation routines later than the original sizes:  1536x1536 and 2048x2048 (typically these are used for high DPI screens).

b1ink0:
## Summary

Fixes #1512 

## Relevant technical choices

This PR automatically opt into 1536x1536 and 2048x2048 sizes when generating fallback images. This only happens when the "generate fallback images" is enabled.

The opt in is done by modifying the `$_wp_additional_image_sizes` global which is the array containing extra image sizes. The `provide_additional_mime_types` property is set to true for the 1536x1536 and 2048x2048 image sizes so that the selected modern image format of these sizes is also generated.

### Before 
For the demo img these images were generated ( fallback images is enabled )
Screenshot 2024-11-20 at 1 58 21 PM

### After
For the demo img these images are now generated ( fallback images is enabled )
Screenshot 2024-11-20 at 1 58 55 PM

### DIFF
--- a/plugins/webp-uploads/hooks.php
+++ b/plugins/webp-uploads/hooks.php
@@ -780,3 +780,31 @@ function webp_uploads_init(): void {
 	}
 }
 add_action( 'init', 'webp_uploads_init' );
+
+/**
+ * Automatically opt into extra image sizes when generating fallback images.
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_opt_in_extra_image_sizes(): void {
+	if ( ! webp_uploads_is_fallback_enabled() ) {
+		return;
+	}
+
+	global $_wp_additional_image_sizes;
+
+	if ( ! is_array( $_wp_additional_image_sizes ) ) {
+		return;
+	}
+
+	// NOTE: Modifying global to mimic the "hypothetical" WP core API behavior.
+
+	if ( isset( $_wp_additional_image_sizes['1536x1536'] ) && ! isset( $_wp_additional_image_sizes['1536x1536']['provide_additional_mime_types'] ) ) {
+		$_wp_additional_image_sizes['1536x1536']['provide_additional_mime_types'] = true; // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+	}
+
+	if ( isset( $_wp_additional_image_sizes['2048x2048'] ) && ! isset( $_wp_additional_image_sizes['2048x2048']['provide_additional_mime_types'] ) ) {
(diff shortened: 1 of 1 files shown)


plugins/webp-uploads/hooks.php
```diff
@@ -780,3 +780,31 @@ function webp_uploads_init(): void {
 	}
 }
 add_action( 'init', 'webp_uploads_init' );
+
+/**
+ * Automatically opt into extra image sizes when generating fallback images.
+ *
+ * @since n.e.x.t
```
mukeshpanchal27, line 787:
```suggestion
 * @since n.e.x.t
 *
 * @global array $_wp_additional_image_sizes Associative array of additional image sizes.
```

Add missing global doc.

plugins/webp-uploads/hooks.php
```diff
@@ -780,3 +780,31 @@ function webp_uploads_init(): void {
...
+function webp_uploads_opt_in_extra_image_sizes(): void {
+	if ( ! webp_uploads_is_fallback_enabled() ) {
+		return;
+	}
+
+	global $_wp_additional_image_sizes;
+
+	if ( ! is_array( $_wp_additional_image_sizes ) ) {
+		return;
+	}
```
mukeshpanchal27:
```suggestion
```

Don't think we needs to add these check. We already added `isset` check so it will not return any error.

plugins/webp-uploads/hooks.php
```diff
@@ -780,3 +780,31 @@ function webp_uploads_init(): void {
...
+
+	// NOTE: Modifying global to mimic the "hypothetical" WP core API behavior.
+
```
felixarntz:
```suggestion
	// Modify global to mimic the "hypothetical" WP core API behavior via an additional `add_image_size()` parameter.
```

felixarntz (approved):
@b1ink0 This basically looks good to me, only two outstanding comments to address.

mukeshpanchal27 (approved):

PR #1680, opened 2024-11-20
Automatically discover plugin dependencies when obtaining Performance feature plugins from WordPress.org

### LINKED ISSUE #1616: Automatically discover plugin dependencies when obtaining Performance feature plugins from WordPress.org
This is a follow-up to https://github.com/WordPress/performance/pull/1613 in which `optimization-detective` was hard-coded as a plugin that needs to be included among the list of plugins to obtain info for. Instead of hard-coding it, we should try to automatically discover the dependencies to be included.

b1ink0:
## Summary

Fixes #1616 

## Relevant technical choices

This PR add a way to programmatically discover plugin dependancies whose information needs to be obtained. This solves the need of hardcoding the dependency plugins.

This PR solves this issue by using the `requires_plugins` value obtained from the standalone plugins information.


plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -71,16 +71,29 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
-	foreach ( $response->plugins as $plugin_data ) {
-		if ( ! isset( $standalone_plugins[ $plugin_data['slug'] ] ) ) {
+	$plugins      = array();
+	$plugin_queue = perflab_get_standalone_plugins();
+
+	// Index the plugins from the API response by their slug for efficient lookup.
+	$plugins_by_slug = array_column( $response->plugins, null, 'slug' );
+
+	// Start processing the plugins using a queue-based approach.
+	while ( ! empty( $plugin_queue ) ) {
```
felixarntz: Why use a `while` loop here? I think it would be more intuitive to use a `foreach` loop like before, that iterates through the `perflab_get_standalone_plugins()` items. What we need to iterate over here is known, so I don't see a need for using `while`.

Then in the loop you could populate `$plugins` with the data, and if there are dependencies (`requires_plugins`), add the data for them to the array as well (unless it was already added).

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -71,16 +71,29 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
 
-	$plugins            = array();
-	$standalone_plugins = array_merge(
-		array_flip( perflab_get_standalone_plugins() ),
-		array( 'optimization-detective' => array() ) // TODO: Programmatically discover the plugin dependencies and add them here. See <https://github.com/WordPress/performance/issues/1616>.
-	);
-	foreach ( $response->plugins as $plugin_data ) {
-		if ( ! isset( $standalone_plugins[ $plugin_data['slug'] ] ) ) {
+	$plugins      = array();
+	$plugin_queue = perflab_get_standalone_plugins();
```
felixarntz, line 75: I think the previous name of this variable, `$standalone_plugins`, makes more sense. It doesn't really need to be a queue, particularly related to my other comment that using a `while` loop seems unnecessarily complicated.

felixarntz (commented):
Thank you for the PR @b1ink0! A bit of feedback below.

plugins/performance-lab/includes/admin/plugins.php
b1ink0: The reason was if `foreach` is used then there is need of the nested `foreach` for handling the dependencies which will have the same logic of outer `foreach` to add it into the `$plugins`. So by using the `while` we can eliminate the duplicate logic. 

Also I tried using `foreach` with appending elements to the array but in PHP `foreach` works by creating an internal copy of the array's state at the beginning of the loop. Which means if we add other elements to that array it will not iterate over them. 
This could also can be solved by using the simple `for` loop but `while` looked more clean.

plugins/performance-lab/includes/admin/plugins.php
b1ink0: I renamed it because in the `while` logic the dependancies are getting merged with the `$plugin_queue` so the `$standalone_plugins` variable name seems misleading as the dependancies are not standalone plugins.

plugins/performance-lab/includes/admin/plugins.php
felixarntz: Ah I see, because there could in theory be dependencies that have dependencies themselves. While we don't currently have that, that's a good point, so let's keep it as is.

plugins/performance-lab/includes/admin/plugins.php
felixarntz: Makes sense.

felixarntz (approved):

plugins/performance-lab/includes/admin/plugins.php
westonruter: Per #1219:

```suggestion
	while ( count( $plugin_queue ) > 0 ) {
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -71,16 +71,29 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
+
+		if ( isset( $plugins[ $current_plugin_slug ] ) || ! isset( $plugins_by_slug[ $current_plugin_slug ] ) ) {
 			continue;
 		}
-		$plugins[ $plugin_data['slug'] ] = wp_array_slice_assoc( $plugin_data, $fields );
+
+		$plugin_data                     = $plugins_by_slug[ $current_plugin_slug ];
+		$plugins[ $current_plugin_slug ] = wp_array_slice_assoc( $plugin_data, $fields );
+
+		if ( empty( $plugin_data['requires_plugins'] ) || ! is_array( $plugin_data['requires_plugins'] ) ) {
```
westonruter: Also per #1219:

```suggestion
		if ( ! isset( $plugin_data['requires_plugins'] ) || ! is_array( $plugin_data['requires_plugins'] ) ) {
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -71,16 +71,29 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
-		array_flip( perflab_get_standalone_plugins() ),
-		array( 'optimization-detective' => array() ) // TODO: Programmatically discover the plugin dependencies and add them here. See <https://github.com/WordPress/performance/issues/1616>.
-	);
-	foreach ( $response->plugins as $plugin_data ) {
-		if ( ! isset( $standalone_plugins[ $plugin_data['slug'] ] ) ) {
+	$plugins      = array();
+	$plugin_queue = perflab_get_standalone_plugins();
+
+	// Index the plugins from the API response by their slug for efficient lookup.
+	$plugins_by_slug = array_column( $response->plugins, null, 'slug' );
```
westonruter: Maybe rename `$plugins_by_slug` to `$all_performance_plugins`. It's confusing to me how it is named now because `$plugins` is also structured by slug.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -71,16 +71,29 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
+
+	// Start processing the plugins using a queue-based approach.
+	while ( ! empty( $plugin_queue ) ) {
+		$current_plugin_slug = array_shift( $plugin_queue );
+
+		if ( isset( $plugins[ $current_plugin_slug ] ) || ! isset( $plugins_by_slug[ $current_plugin_slug ] ) ) {
```
westonruter: In reality, the `! isset( $plugins_by_slug[ $current_plugin_slug ] )` here would technically be an error case, right? It essentially means either that one of our Performance plugins got removed from the repo or that one of our plugins has a `requires_plugins` field that references a non-existent plugin, correct?

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -71,16 +71,29 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
+
+		$plugin_data                     = $plugins_by_slug[ $current_plugin_slug ];
+		$plugins[ $current_plugin_slug ] = wp_array_slice_assoc( $plugin_data, $fields );
+
+		if ( empty( $plugin_data['requires_plugins'] ) || ! is_array( $plugin_data['requires_plugins'] ) ) {
+			continue;
+		}
+
+		// Enqueue the required plugins slug by adding it to the queue.
+		$plugin_queue = array_merge( $plugin_queue, $plugin_data['requires_plugins'] );
```
westonruter: Might as well refactor this a bit:

```suggestion
		// Enqueue the required plugins slug by adding it to the queue.
		if ( isset( $plugin_data['requires_plugins'] ) && is_array( $plugin_data['requires_plugins'] ) ) {
			$plugin_queue = array_merge( $plugin_queue, $plugin_data['requires_plugins'] );
		}
```

plugins/performance-lab/includes/admin/plugins.php
westonruter: But see below.

westonruter (changes requested):

plugins/performance-lab/includes/admin/plugins.php
b1ink0: I also think error would be better in this case.

westonruter (approved):

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -71,16 +71,34 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
+
+		if ( isset( $plugins[ $current_plugin_slug ] ) ) {
 			continue;
 		}
-		$plugins[ $plugin_data['slug'] ] = wp_array_slice_assoc( $plugin_data, $fields );
+
+		if ( ! isset( $all_performance_plugins[ $current_plugin_slug ] ) ) {
+			return new WP_Error(
+				'plugin_not_found',
+				__( 'Plugin not found in Performance Lab\'s features / plugins.', 'performance-lab' )
```
felixarntz: WordPress Core discourages unnecessary escaping in strings, so this would be better as:
```suggestion
				__( "Plugin not found in Performance Lab's features / plugins.", 'performance-lab' )
```

Also I think we shouldn't use this "features / plugins" syntax, it reads weird. Maybe better as follows? It also avoids the apostrophe.
```suggestion
				__( 'Plugin not found in WordPress.org API response.', 'performance-lab' )
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -71,16 +71,34 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
+	while ( count( $plugin_queue ) > 0 ) { // phpcs:ignore Squiz.PHP.DisallowSizeFunctionsInLoops.Found
+		$current_plugin_slug = array_shift( $plugin_queue );
+
+		if ( isset( $plugins[ $current_plugin_slug ] ) ) {
 			continue;
 		}
-		$plugins[ $plugin_data['slug'] ] = wp_array_slice_assoc( $plugin_data, $fields );
+
+		if ( ! isset( $all_performance_plugins[ $current_plugin_slug ] ) ) {
+			return new WP_Error(
```
felixarntz, line 89: We shouldn't directly return `WP_Error` objects from the logic as that can cause API requests to be made on every page load. Initially we had that fixed, but I see some of the recent changes above already broke that contract, so we should fix it.

I'm fine leaving this here for now as the problem is not strictly introduced by this PR alone, but we should fix it in a follow up PR: Basically, if there's an error, the error should still be cached - either the error directly, or at least an empty array so that the initial error from line 27 is used. It doesn't matter too much that the cached results return the exact same error, but it's crucial that errors don't lead to the request being fired over and over again. What we could do potentially is use a shorter transient expiration if we cache an error response.

felixarntz (commented):
A few new points of feedback after the latest changes.

plugins/performance-lab/includes/admin/plugins.php
westonruter: Filed this as https://github.com/WordPress/performance/issues/1691

PR #1689, opened 2024-11-21
Enable end user opt-in to generate all sizes in fallback format 

### LINKED ISSUE #1513: Enable end user opt-in to generate all sizes in fallback format
## Feature Description

As discussed in https://github.com/WordPress/performance/pull/1421 when "generate fallback images" is enabled, we only generate the core default sizes in the fallback format. The reasoning here is that users can have many sizes and generating all of the sub-sizes can require an unexpectedly large amount of storage for users. I am unsure about this so let's give users the _option to generate all sizes_.

This proposal is to add a checkbox users can check right below the "generate fallback images" checkbox (and it would only be enabled when that checkbox is checked).  Similar the the generate fallback checkbox, the note would indicate this requires even more disk space.

Something like:

```
 - [ ] **Generate fallback images in all sizes including custom sizes** 
 * Note: uses even more storage space.
```

Ideally when users enable this feature we should notate that in the Generator tag so we can later discover how many sites opt into generating all image sizes.

b1ink0:
## Summary

Fixes #1513 

## Relevant technical choices

This PR adds a checkbox users can check right below the "generate fallback images" checkbox (and it would only be enabled when "generate fallback images" checkbox is checked). 

### Settings Page Screenshots
Screenshot 2024-11-21 at 6 14 24 PM
Screenshot 2024-11-21 at 6 14 53 PM

### Sample outputs for the below custom sizes
```php
add_image_size( 'custom_size_1', 400, 400, true );
add_image_size( 'custom_size_2', 800, 800, true );
add_image_size( 'custom_size_3', 1200, 1200, true );
```

- When the new option is off
Screenshot 2024-11-21 at 6 19 03 PM

- When the new option is on

Screenshot 2024-11-21 at 6 20 51 PM


plugins/webp-uploads/helper.php
```diff
@@ -411,6 +411,17 @@ function webp_uploads_is_fallback_enabled(): bool {
...
 }
 
+/**
+ * Checks if the `perflab_generate_fallback_all_sizes` option is enabled.
+ *
+ * @since n.e.x.t
+ *
+ * @return bool True if the option is enabled, false otherwise.
+ */
+function webp_uploads_is_fallback_all_sizes_enabled(): bool {
```
felixarntz: Maybe a better name would be something like `webp_uploads_should_generate_all_fallback_sizes()`?

plugins/webp-uploads/hooks.php
```diff
@@ -719,7 +719,13 @@ function webp_uploads_get_image_sizes_additional_mime_type_support(): array {
...
 
 	foreach ( $additional_sizes as $size => $size_details ) {
-		$allowed_sizes[ $size ] = ! empty( $size_details['provide_additional_mime_types'] );
+		if ( isset( $size_details['provide_additional_mime_types'] ) ) {
+			// Give priority to the 'provide_additional_mime_types' property.
+			$allowed_sizes[ $size ] = (bool) $size_details['provide_additional_mime_types'];
+		} else {
+			// If 'provide_additional_mime_types' is not set, use the fallback all sizes setting.
+			$allowed_sizes[ $size ] = webp_uploads_is_fallback_all_sizes_enabled();
+		}
```
felixarntz: Rather than doing this here, I think a more elegant solution would be to decouple the UI control and option from the main logic, by using the `webp_uploads_image_sizes_with_additional_mime_type_support` filter below. So I think this would be better separately implemented.

It would also allow the code to be a bit more efficient: We should only need to call `webp_uploads_is_fallback_all_sizes_enabled()` once, not for every size available. If `false` we can immediately return without changing anything. If `true`, we go through the array of sizes and set them to `true`.

I think it's okay not to check the `provide_additional_mime_types` value directly to see if it's `false`. It's primary purpose it to _enable_ (not disable) this behavior anyway, and since the setting explicitly says "all", it's fine to do it for all sizes, regardless of what other configuration may be present.

plugins/webp-uploads/hooks.php
```diff
@@ -764,6 +770,10 @@ function webp_uploads_modify_webp_quality( int $quality, string $mime_type ): in
 function webp_uploads_render_generator(): void {
 	// Use the plugin slug as it is immutable.
 	echo '<meta name="generator" content="webp-uploads ' . esc_attr( WEBP_UPLOADS_VERSION ) . '">' . "\n";
+
+	if ( webp_uploads_is_fallback_all_sizes_enabled() ) {
+		echo '<meta name="generator" content="webp-uploads-fallback-all-sizes ' . esc_attr( WEBP_UPLOADS_VERSION ) . '">' . "\n";
+	}
```
felixarntz: Why is that being added? We shouldn't have an extra generator tag for this feature specifically. If we need feature-specific indicators in the frontend, let's discuss that in a separate issue.

plugins/webp-uploads/settings.php
```diff
@@ -40,6 +40,18 @@ function webp_uploads_register_media_settings_field(): void {
 			'show_in_rest' => false,
 		)
 	);
+
+	// Add a setting to generate fallback images in all sizes including custom sizes.
+	register_setting(
+		'media',
+		'perflab_generate_fallback_all_sizes',
```
felixarntz: Maybe rename this to `perflab_generate_all_fallback_sizes`, that reads a bit better.

plugins/webp-uploads/settings.php
```diff
@@ -178,6 +200,94 @@ function webp_uploads_generate_webp_jpeg_setting_callback(): void {
 	<?php
 }
 
+
+/**
+ * Renders the settings field for generating fallback images in all sizes.
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_generate_fallback_all_sizes_callback(): void {
```
felixarntz: See above in terms of naming.

plugins/webp-uploads/settings.php
```diff
@@ -96,6 +108,16 @@ function webp_uploads_add_media_settings_fields(): void {
 		array( 'class' => 'perflab-generate-webp-and-jpeg' )
 	);
 
+	// Add setting field for generating fallback images in all sizes including custom sizes.
+	add_settings_field(
+		'perflab_generate_fallback_all_sizes',
+		__( 'Output all fallback images', 'webp-uploads' ),
```
felixarntz:
```suggestion
		__( 'Generate all fallback image sizes', 'webp-uploads' ),
```

felixarntz (changes requested):
@b1ink0 A bit of feedback here, but overall this looks great!

plugins/webp-uploads/hooks.php
b1ink0: In the #1513 there is mention of the adding the generator tag thats why I added it.
> Ideally when users enable this feature we should notate that in the Generator tag so we can later discover how many sites opt into generating all image sizes.

Should I remove the generator tag code from this PR?

felixarntz: Yeah, I think so, let's not do it in this PR, as it needs further discussion.

Maybe worth opening an issue to discuss whether/how we want to make more specific features discoverable in the frontend? Such an issue could link to that comment as one example reference.

b1ink0: Removed the generator tag code from this PR.

felixarntz (approved):
Thanks @b1ink0, LGTM!

Paging @adamsilverstein for additional review.

plugins/webp-uploads/settings.php
```diff
@@ -178,6 +200,94 @@ function webp_uploads_generate_webp_jpeg_setting_callback(): void {
...
+	$all_fallback_sizes_hidden_id = 'perflab_generate_all_fallback_sizes_hidden';
+
+	?>
+	<style>
+		#perflab_generate_all_fallback_sizes_fieldset.disabled label,
+		#perflab_generate_all_fallback_sizes_fieldset.disabled p {
+			opacity: 0.7;
+		}
+	</style>
+	<div id="perflab_generate_all_fallback_sizes_notice" class="notice notice-info inline" <?php echo $fallback_enabled ? 'hidden' : ''; ?>>
```
mukeshpanchal27, line 221: `hidden` should be a class isn't it?

plugins/webp-uploads/helper.php
```diff
@@ -411,6 +411,17 @@ function webp_uploads_is_fallback_enabled(): bool {
...
 
+/**
+ * Checks if the `perflab_generate_all_fallback_sizes` option is enabled.
+ *
+ * @since n.e.x.t
+ *
+ * @return bool True if the option is enabled, false otherwise.
+ */
+function webp_uploads_should_generate_all_fallback_sizes(): bool {
+	return (bool) get_option( 'perflab_generate_all_fallback_sizes' );
```
mukeshpanchal27: Should add default value here as well.

```suggestion
	return (bool) get_option( 'perflab_generate_all_fallback_sizes', 0 );
```

plugins/webp-uploads/hooks.php
```diff
@@ -806,3 +806,25 @@ function webp_uploads_opt_in_extra_image_sizes(): void {
...
 add_action( 'plugins_loaded', 'webp_uploads_opt_in_extra_image_sizes' );
+
+/**
+ * Enables additional MIME type support for all image sizes based on the generate all fallback sizes settings.
+ *
+ * @since n.e.x.t
+ *
+ * @param array<string, bool> $allowed_sizes A map of image size names and whether they are allowed to have additional MIME types.
+ *
+ * @return array<string, bool> Modified map of image sizes with additional MIME type support.
```
mukeshpanchal27:
```suggestion
 * @param array<string, bool> $allowed_sizes A map of image size names and whether they are allowed to have additional MIME types.
 * @return array<string, bool> Modified map of image sizes with additional MIME type support.
```

plugins/webp-uploads/settings.php
```diff
@@ -178,6 +200,94 @@ function webp_uploads_generate_webp_jpeg_setting_callback(): void {
...
 }
 
+
+/**
+ * Renders the settings field for generating all fallback image sizes.
+ *
+ * @since n.e.x.t
+ */
+function webp_uploads_generate_all_fallback_sizes_callback(): void {
+	$all_fallback_sizes_enabled   = 1 === (int) get_option( 'perflab_generate_all_fallback_sizes', 0 );
```
mukeshpanchal27:
```suggestion
	$all_fallback_sizes_enabled   = 1 === webp_uploads_should_generate_all_fallback_sizes();
```

Can we re-use `webp_uploads_should_generate_all_fallback_sizes()` here as well?

plugins/webp-uploads/settings.php
```diff
@@ -178,6 +200,94 @@ function webp_uploads_generate_webp_jpeg_setting_callback(): void {
...
+	$fallback_enabled             = webp_uploads_is_fallback_enabled();
+	$all_fallback_sizes_hidden_id = 'perflab_generate_all_fallback_sizes_hidden';
+
+	?>
+	<style>
```
mukeshpanchal27, line 215: Future request for `<style>` and `<script>` tags for individual functions:

Currently, for the `<picture>` element, we have added additional CSS and script tags. Can we centralize these in a common location ( Single tag for `<style>` and `<script>` )? This would ensure that in the future, we don’t need to re-add them in specific callback function.

mukeshpanchal27 (commented):
Overall looks good to me. Left some feedbacks.

plugins/webp-uploads/settings.php
b1ink0: No actually we are using the HTML attribute [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) for hiding the notice.

plugins/webp-uploads/settings.php
b1ink0: Yes we can done in https://github.com/WordPress/performance/pull/1689/commits/3a647137c35afd2d2fe57d7af7a80c653d3defa4

plugins/webp-uploads/settings.php
b1ink0: Would moving `<style>` and `<script>` to files and enqueuing them better option for centralization then?

mukeshpanchal27: Something we can do in follow-up PR once these one is committed.

plugins/webp-uploads/helper.php
```diff
@@ -411,6 +411,17 @@ function webp_uploads_is_fallback_enabled(): bool {
 	return (bool) get_option( 'perflab_generate_webp_and_jpeg' );
 }
 
+/**
+ * Checks if the `perflab_generate_all_fallback_sizes` option is enabled.
+ *
+ * @since n.e.x.t
+ *
+ * @return bool True if the option is enabled, false otherwise.
```
adamsilverstein:
```suggestion
 * @return bool Whether the option is enabled. Default is false.
```

adamsilverstein (commented):
Overall looks great. One small doc change suggestion. Unit tests to validate the feature works as expected would also be nice to have.

adamsilverstein (approved):
Nice work!

plugins/webp-uploads/tests/test-load.php
```diff
@@ -1045,4 +1045,44 @@ public function temp_filename(): string {
...
+
+		// Clean up custom sizes.
+		remove_image_size( 'custom_size_1' );
+		remove_image_size( 'custom_size_2' );
+
+		// Verify that fallback images are generated for custom sizes.
+		foreach ( array( 'custom_size_1', 'custom_size_2' ) as $size_name ) {
+			$this->assertImageHasSizeSource( $attachment_id, $size_name, 'image/jpeg' );
+			$this->assertImageHasSizeSource( $attachment_id, $size_name, $mime_type );
+		}
```
mukeshpanchal27, line 1086:
```suggestion
		}

		wp_delete_attachment( $attachment_id );
```

mukeshpanchal27 (approved):

PR #1693, opened 2024-11-22
Fix latest plugin version not being downloaded consistently

### LINKED ISSUE #1690: Old Plugin Version Downloaded Due to Cached API Response in Transient
## Bug Description

When activating a plugin from the Performance Lab settings page shortly after a new version is released, there is a possibility that an older version will be downloaded. This is because the response from the plugins API query is cached in a WordPress transient (`_transient_perflab_plugins_info`), which has a default lifetime of one hour. During this time, the cached plugin data, including the version and download link, may still point to the older version.

Code responsible for the plugin installation:
https://github.com/WordPress/performance/blob/20a038f720a9b0b54d9d9680bc99df473d7f9f80/plugins/performance-lab/includes/admin/plugins.php#L354-L357

To illustrate, the unserialized values of the `_transient_perflab_plugins_info` transient look like this:

```log
Array
(
    ...

    [optimization-detective] => Array
        (
            [name] => Optimization Detective
            [slug] => optimization-detective
            [short_description] => Provides an API for leveraging real user metrics to detect optimizations to apply on the frontend to improve page performance.
            [requires] => 6.5
            [requires_php] => 7.2
            [requires_plugins] => Array
                (
                )

            [download_link] => https://downloads.wordpress.org/plugin/optimization-detective.0.7.0.zip
            [version] => 0.7.0
        )

        ...
)
```

### Scope of the Issue

This issue is relatively rare and occurs only under specific conditions:

1. A new version of a plugin in Performance Lab has just been released.
2. The transient data is already present and has not expired.
3. The plugin activation attempt happens within the transient's one-hour lifetime.

While this is not a critical issue (users can always go to the main Plugins page to update manually), it may cause confusion in situations where the latest version is expected to be installed immediately from the settings page.

## Steps to reproduce

Reproducing this behavior required manually altering the transient data (`_transient_perflab_plugins_info`) in the WordPress database. I did this by modifying the `download_link` value for the Optimization Detective plugin in the `wp_options` table. Specifically, I changed the link to `https://downloads.wordpress.org/plugin/optimization-detective.0.7.0.zip` to simulate the presence of an older version.

This manual adjustment was necessary because there isn't an easy way to trigger this scenario without an actual new plugin version release.

ShyamGadde:
## Summary

Fixes #1690

### DIFF
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -372,7 +372,9 @@ function perflab_install_and_activate_plugin( string $plugin_slug, array &$proce
 		// Replace new Plugin_Installer_Skin with new Quiet_Upgrader_Skin when output needs to be suppressed.
 		$skin     = new WP_Ajax_Upgrader_Skin( array( 'api' => $plugin_data ) );
 		$upgrader = new Plugin_Upgrader( $skin );
-		$result   = $upgrader->install( $plugin_data['download_link'] );
+		// Remove the version number from the link to download the latest plugin version.
+		$download_link = (string) preg_replace( '#(\/plugin\/[^\/]+)\.\d+\.\d+\.\d+\.zip#', '$1.zip', $plugin_data['download_link'] );
+		$result        = $upgrader->install( $download_link );
 
 		if ( is_wp_error( $result ) ) {
 			return $result;
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -343,9 +343,22 @@ function perflab_install_and_activate_plugin( string $plugin_slug, array &$proce
 	}
 	$processed_plugins[] = $plugin_slug;
 
-	$plugin_data = perflab_query_plugin_info( $plugin_slug );
+	$plugin_data = plugins_api(
+		'plugin_information',
+		array(
+			'slug'   => $plugin_slug,
+			'fields' => array(
+				'sections' => false,
+			),
+		)
+	);
+
 	if ( $plugin_data instanceof WP_Error ) {
 		return $plugin_data;
+	}
+
+	if ( is_object( $plugin_data ) ) {
+		$plugin_data = (array) $plugin_data;
 	}
 
 	// Add recommended plugins (soft dependencies) to the list of plugins installed and activated.
@@ -372,9 +385,7 @@ function perflab_install_and_activate_plugin( string $plugin_slug, array &$proce
 		// Replace new Plugin_Installer_Skin with new Quiet_Upgrader_Skin when output needs to be suppressed.
 		$skin     = new WP_Ajax_Upgrader_Skin( array( 'api' => $plugin_data ) );
 		$upgrader = new Plugin_Upgrader( $skin );
-		// Remove the version number from the link to download the latest plugin version.
-		$download_link = (string) preg_replace( '#(\/plugin\/[^\/]+)\.\d+\.\d+\.\d+(?:-\w+)?\.zip#', '$1.zip', $plugin_data['download_link'] );
-		$result        = $upgrader->install( $download_link );
+		$result   = $upgrader->install( $plugin_data['download_link'] );
 
 		if ( is_wp_error( $result ) ) {
 			return $result;
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -373,7 +373,7 @@ function perflab_install_and_activate_plugin( string $plugin_slug, array &$proce
 		$skin     = new WP_Ajax_Upgrader_Skin( array( 'api' => $plugin_data ) );
 		$upgrader = new Plugin_Upgrader( $skin );
 		// Remove the version number from the link to download the latest plugin version.
-		$download_link = (string) preg_replace( '#(\/plugin\/[^\/]+)\.\d+\.\d+\.\d+\.zip#', '$1.zip', $plugin_data['download_link'] );
+		$download_link = (string) preg_replace( '#(\/plugin\/[^\/]+)\.\d+\.\d+\.\d+(?:-\w+)?\.zip#', '$1.zip', $plugin_data['download_link'] );
 		$result        = $upgrader->install( $download_link );
 
 		if ( is_wp_error( $result ) ) {
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -347,7 +347,9 @@ function perflab_install_and_activate_plugin( string $plugin_slug, array &$proce
 		array(
 			'slug'   => $plugin_slug,
 			'fields' => array(
-				'sections' => false,
+				'download_link'    => true,
+				'requires_plugins' => true,
+				'sections'         => false, // Omit the bulk of the response which we don't need.
 			),
 		)
 	);
--- a/plugins/performance-lab/includes/admin/plugins.php
+++ b/plugins/performance-lab/includes/admin/plugins.php
@@ -16,7 +16,7 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @since 2.8.0
  *
  * @param string $plugin_slug The string identifier for the plugin in questions slug.
- * @return array{name: string, slug: string, short_description: string, requires: string|false, requires_php: string|false, requires_plugins: string[], download_link: string, version: string}|WP_Error Array of plugin data or WP_Error if failed.
+ * @return array{name: string, slug: string, short_description: string, requires: string|false, requires_php: string|false, requires_plugins: string[], version: string}|WP_Error Array of plugin data or WP_Error if failed.
  */
 function perflab_query_plugin_info( string $plugin_slug ) {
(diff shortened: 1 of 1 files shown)


plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -372,7 +372,9 @@ function perflab_install_and_activate_plugin( string $plugin_slug, array &$proce
 		// Replace new Plugin_Installer_Skin with new Quiet_Upgrader_Skin when output needs to be suppressed.
 		$skin     = new WP_Ajax_Upgrader_Skin( array( 'api' => $plugin_data ) );
 		$upgrader = new Plugin_Upgrader( $skin );
-		$result   = $upgrader->install( $plugin_data['download_link'] );
+		// Remove the version number from the link to download the latest plugin version.
+		$download_link = (string) preg_replace( '#(\/plugin\/[^\/]+)\.\d+\.\d+\.\d+\.zip#', '$1.zip', $plugin_data['download_link'] );
```
westonruter: It might be safer to do:

```suggestion
		$download_link = (string) preg_replace( '#(\/plugin\/[^\/]+)\.\d.*.zip#', '$1.zip', $plugin_data['download_link'] );
```

The reason being that it's possible for plugin versions to have `-rc` or `-beta` tagged on. 

For example: 
https://wordpress.org/plugins/rest-api/
=>
https://downloads.wordpress.org/plugin/rest-api.2.0-beta15.zip

Granted, it's highly unlikely that we'll ever release one of our plugins with such a Stable Tag, and it's not super common across the repo: https://wpdirectory.net/search/01JDAFQMKVDE6ME17SH47V33D9

westonruter: So I don't feel strongly about this.

westonruter: Worst case in such a scenario is that the original URL ends up getting used without the version removed, which will likely still work (even if it has the even slighter possibility of downloading a stale version).

westonruter (approved):

plugins/performance-lab/includes/admin/plugins.php
ShyamGadde: Thank you for pointing this out! I hadn’t considered the possibility of version tags like `-rc` or `-beta`. That’s an oversight on my part.

However, I believe the regex `'#(\/plugin\/[^\/]+)\.\d.*.zip#'` might not work as intended because it could capture the plugin name greedily. For instance, in the example provided, the download link might end up being `https://downloads.wordpress.org/plugin/rest-api.2.zip` instead of `https://downloads.wordpress.org/plugin/rest-api.zip`.

To address this, one solution could be to simply modify the regex to capture the plugin name non-greedily:

```php
'#(\/plugin\/[^\/]+?)\.\d.*.zip#'
```

Alternatively, another approach would be to use the following regex:

```php
'#(\/plugin\/[^\/]+)\.\d+\.\d+\.\d+(?:-\w+)?\.zip#'
```

This pattern accounts for optional tags like `-rc` or `-beta` and aligns with the regex used in the `versions` command, so it seems like a good choice to me.

westonruter: Note in the example https://downloads.wordpress.org/plugin/rest-api.2.0-beta15.zip there is no patch version either, just a major and minor. So that would need to be optional as well.

This should address the greediness issue, by using `\.\d.*?.zip` instead of `\.\d.*.zip`:

```php
$download_link = (string) preg_replace( '#(\/plugin\/[^\/]+)\.\d.*?.zip#', '$1.zip', $plugin_data['download_link'] );
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -372,7 +372,9 @@ function perflab_install_and_activate_plugin( string $plugin_slug, array &$proce
 		// Replace new Plugin_Installer_Skin with new Quiet_Upgrader_Skin when output needs to be suppressed.
 		$skin     = new WP_Ajax_Upgrader_Skin( array( 'api' => $plugin_data ) );
 		$upgrader = new Plugin_Upgrader( $skin );
-		$result   = $upgrader->install( $plugin_data['download_link'] );
+		// Remove the version number from the link to download the latest plugin version.
+		$download_link = (string) preg_replace( '#(\/plugin\/[^\/]+)\.\d+\.\d+\.\d+(?:-\w+)?\.zip#', '$1.zip', $plugin_data['download_link'] );
```
```suggestion
		$download_link = (string) preg_replace( '#(\/plugin\/[^\/]+)\.\d.*?.zip#', '$1.zip', $plugin_data['download_link'] );
```

westonruter:
Lastly, note that many plugins also use `Stable tag: trunk`: https://wpdirectory.net/search/01JDB0D3QBVXR8KWR1X9BRYYPB

I haven't seen `trunk` show up in the download URL, but in theory it might?

Here's another search for plugins using non-standard versions: https://wpdirectory.net/search/01JDB0R6MBNNEMH0A1X8AHKQYC

The WP Umbrella plugin version doesn't begin with a digit:

https://downloads.wordpress.org/plugin/wp-health.v2.16.4.zip

That being said, perhaps we should just strip off everything up to the first `.` since I don't believe any plugin is allowed to have that as a character in the slug.

plugins/performance-lab/includes/admin/plugins.php
westonruter: Or rather, per https://github.com/WordPress/performance/pull/1693#issuecomment-2495000449

```suggestion
		$download_link = (string) preg_replace( '#(\/plugin\/[^\/.]+)\..+?\.zip#', '$1.zip', $plugin_data['download_link'] );
```

westonruter:
Again, this is all likely splitting hairs since none of the plugins we include here should ever use a non-standard version!

plugins/performance-lab/includes/admin/plugins.php
felixarntz: I'm not a fan of regexing this from the `download_link`. Is there no externally controlled place (API responses or elsewhere) that exposes the format we want here?

ShyamGadde: Stripping off everything up to the first `.` makes sense. However, I also noticed that some plugins have their download link as `https://downloads.wordpress.org/plugin/<plugin-slug>.zip`, particularly those with tags like `trunk`, such as https://downloads.wordpress.org/plugin/better-search-replace.zip.

Given this, I think we might need a regex like this:

```php
$download_link = (string) preg_replace( '#(\/plugin\/[^\.]+).*\.zip#', '$1.zip', $plugin_data['download_link'] );
```

This should account for both cases where the version exists (like `rest-api.2.0-beta15.zip`) and where it doesn't (like `better-search-replace.zip`).

ShyamGadde: On second thought, can’t we just construct the download link directly from the slug in the `$plugin_data` as:

```
https://downloads.wordpress.org/plugin/<plugin-slug>.zip
```

This would simplify and remove the need for regex.

felixarntz: Seeing comments like https://github.com/WordPress/performance/pull/1693#discussion_r1856636899 confirms my above concern: Using a regex for this is error-prone. Is there no way to get the actual URL from somewhere?

ShyamGadde: You’re right; relying on regex for this isn’t ideal. The `plugins_api()` function, when called with the `plugin_information` action, provides a `versions` field in the response that seems to include the URLs in the required format. Here's an example of the response:

```json
{
    ...
    "versions": {
        "0.1.0": "https://downloads.wordpress.org/plugin/optimization-detective.0.1.0.zip",
        "0.1.1": "https://downloads.wordpress.org/plugin/optimization-detective.0.1.1.zip",
        ...
        "trunk": "https://downloads.wordpress.org/plugin/optimization-detective.zip"
    },
    ...
}
```

We could use the `trunk` key to get the required URL in most cases. However, there are scenarios where the `trunk` key doesn’t exist in the `versions` field. In those cases, the `download_link` itself is typically already in the desired format without a version or tag.

That said, implementing this approach would require making multiple API calls—one for each plugin—which again might not be ideal.

felixarntz: Don't we only need the plugin download URL if we actually want to download the plugin? I think making an extra request for that purpose makes sense. We shouldn't make all these requests for all the plugins when we cache the information. But for the moment when someone clicks "Activate" and we have to install that plugin, I think it's reasonable to make the request. I think that's even what WordPress Core does to install a plugin.

And even if the plugin also requires dependencies, making individual requests for these should not be an issue since realistically that shouldn't be more than 2-3 plugins.

This solution would be more reliable than us assuming things about the download URL and using a regex.

westonruter: Hah. Good point.

westonruter: That makes sense to me.

westonruter (approved):
Since we're not using the `download_link` anymore in the response to `perflab_query_plugin_info()`, I removed it from being explicitly returned. I also explicitly requested `download_link` and `requires_plugins` when installing, although it seems these are all returned by default anyway.

felixarntz (approved):
Thanks @ShyamGadde @westonruter, LGTM!

PR #1694, opened 2024-11-22
Disregard transient cache in `perflab_query_plugin_info()` when a plugin is absent

### LINKED ISSUE #1617: Disregard transient cache in perflab_query_plugin_info() when a plugin is absent
This is a follow-up to https://github.com/WordPress/performance/pull/1613 specifically to @joemcgill's [suggestion](https://github.com/WordPress/performance/pull/1613#discussion_r1809266308):

> Oh, additionally, we should probably update the logic below so that if a slug is not found in the cache, it tries to request it from the API instead of returning a `WP_Error` right away.

When a new feature plugin is published, it may not yet exist in the `perflab_plugins_info-v2` transient. Therefore, whenever one of the expected `$standalone_plugins` is absent, we should act as if the transient was not set in the first place so that we can obtain the latest plugin info right away.

Related: #1616

### LINKED ISSUE #1691: Calls to `perflab_query_plugin_info()` should have `WP_Error` return cached
> We shouldn't directly return `WP_Error` objects from the logic as that can cause API requests to be made on every page load. Initially we had that fixed, but I see some of the recent changes above already broke that contract, so we should fix it.
>
> I'm fine leaving this here for now as the problem is not strictly introduced by this PR alone, but we should fix it in a follow up PR: Basically, if there's an error, the error should still be cached - either the error directly, or at least an empty array so that the initial error from line 27 is used. It doesn't matter too much that the cached results return the exact same error, but it's crucial that errors don't lead to the request being fired over and over again. What we could do potentially is use a shorter transient expiration if we cache an error response.

_Originally posted by @felixarntz in https://github.com/WordPress/performance/pull/1680#discussion_r1852637291_

b1ink0:
## Summary

Fixes #1617

Fixes #1691 

## Relevant technical choices

When a new feature plugin is published, it may not yet exist in the transient. Therefore, whenever one of the expected feature plugin is absent, we should act as if the transient was not set in the first place so that we can obtain the latest plugin info right away.

Logic for this to work:
- If the plugin slug is not found in the transient then instead of returning a error we proceeded to fetch information from API.
- If the plugin slug is also not found in the API it is also stored in transient but with key as plugin slug and value as `error`.
- If the plugin slug is found in the transient but if `error` is set then raise the error.

This logic will make sure that the multiple request are not made for the plugin which is absent.


plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -86,10 +88,9 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 		}
 
 		if ( ! isset( $all_performance_plugins[ $current_plugin_slug ] ) ) {
-			return new WP_Error(
-				'plugin_not_found',
-				__( 'Plugin not found in WordPress.org API response.', 'performance-lab' )
-			);
+			// Cache the fact that the plugin was not found.
+			$plugins[ $current_plugin_slug ] = false;
```
westonruter: Instead of storing `false` here, what about storing the `WP_Error` instance? As it stands right now, there are two conditions resulting in `false` being logged, which conflates these two error scenarios which we tried to get rid of in #1651 to assist with debugging.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -101,9 +102,14 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 		}
 	}
 
+	if ( ! isset( $plugins[ $plugin_slug ] ) ) {
+		// Cache the fact that the plugin was not found.
+		$plugins[ $plugin_slug ] = false;
```
westonruter: See above about storing a `WP_Error` instance instead.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -23,14 +23,16 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
-				__( 'Plugin not found in cached API response.', 'performance-lab' )
-			);
+		if ( isset( $plugins[ $plugin_slug ] ) ) {
+			if ( false === $plugins[ $plugin_slug ] ) {
+				// Plugin was requested before and not found.
+				return new WP_Error(
+					'plugin_not_found',
+					__( 'Plugin not found in cached API response.', 'performance-lab' )
+				);
+			}
```
westonruter: If `$plugins[ $plugin_slug ]` is either an `array` or a `WP_Error`, then this can just return that.

```suggestion
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -101,9 +102,14 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 		}
 	}
 
+	if ( ! isset( $plugins[ $plugin_slug ] ) ) {
+		// Cache the fact that the plugin was not found.
+		$plugins[ $plugin_slug ] = false;
+	}
+
 	set_transient( $transient_key, $plugins, HOUR_IN_SECONDS );
```
westonruter: Here I think it would be good to check if there was an error condition, and if so, set the expiration to 1 minute instead of `HOUR_IN_SECONDS`.

westonruter (changes requested):

plugins/performance-lab/includes/admin/plugins.php
felixarntz: Overall this makes sense to me. We probably shouldn't store the `WP_Error` _instance_ though, but just its data, since storing PHP class instances is a bit unpredictable with what would happen in different object cache implementations.

westonruter:
There are still these error scenarios that isn't getting cached:

https://github.com/WordPress/performance/blob/2a344c9a287401bf4a0cf10bef747038f0be0087/plugins/performance-lab/includes/admin/plugins.php#L58-L72

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -101,12 +105,24 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
+			'error' => array(
+				'code'    => 'plugin_not_found',
+				'message' => __( 'Plugin not found in API response.', 'performance-lab' ),
+			),
+		);
+
+		set_transient( $transient_key, $plugins, MINUTE_IN_SECONDS );
+	} else {
+		set_transient( $transient_key, $plugins, HOUR_IN_SECONDS );
+	}
```
felixarntz: This differentiation does not happen in the correct clause. We should use a shorter cache duration in case of any error, not just in the case of this specific error.

I think it would be most straightforward to implement this via a boolean flag like `$has_errors` that is set if anywhere in the (uncached) logic an error is set. Then the code here could be changed like:
```suggestion
	}

	if ( $has_errors ) {
		set_transient( $transient_key, $plugins, MINUTE_IN_SECONDS );
	} else {
		set_transient( $transient_key, $plugins, HOUR_IN_SECONDS );
	}
```

felixarntz (changes requested):
Thanks @b1ink0! This looks good, just one thing that needs to be fixed I think.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -54,61 +54,94 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
-	if ( ! ( is_object( $response ) && property_exists( $response, 'plugins' ) ) ) {
-		return new WP_Error( 'no_plugins', __( 'No plugins found in the API response.', 'performance-lab' ) );
+	if ( ! $has_errors && ! ( is_object( $response ) && property_exists( $response, 'plugins' ) ) ) {
+		$plugins[ $plugin_slug ] = array(
+			'error' => array(
+				'code'    => 'no_plugins',
+				'message' => __( 'No plugins found in the API response.', 'performance-lab' ),
+			),
+		);
+
```
felixarntz: I think for both of these scenarios we need to set the error for every single plugin of ours. In other words, iterate through the list from `perflab_get_standalone_plugins()` and set the same error for every plugin's slug.

Otherwise, this can cause a problem where the API failing would cause the request to be issued for every single plugin, which would defeat the purpose of the transient cache and the combination of those plugin requests into a single API request (see #1542). This would effectively fix #1691, and since this PR is touching the exact same logic, I wonder whether we should go straight to a solution that addresses both issues.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -54,61 +54,94 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
+					'code'    => 'plugin_not_found',
+					'message' => __( 'Plugin not found in API response.', 'performance-lab' ),
+				),
+			);
 
-		// Enqueue the required plugins slug by adding it to the queue.
-		if ( isset( $plugin_data['requires_plugins'] ) && is_array( $plugin_data['requires_plugins'] ) ) {
-			$plugin_queue = array_merge( $plugin_queue, $plugin_data['requires_plugins'] );
+			$has_errors = true;
 		}
```
felixarntz: Now that I look at this again, what is this particular check needed for? Wouldn't this already be catered for by the check above in line 101?

felixarntz (commented):
@b1ink0 There are two points of follow-up feedback, though some of them have overlap with #1691. But since this is broken too and affected by the same logic, maybe we should work towards a solution that addresses both?

cc @westonruter

plugins/performance-lab/includes/admin/plugins.php
b1ink0: To make things clear as its getting confusing please answer these below question or my understanding of the solution needed.

- If error is encounterd at line `57` and `76` we should add the error to the all `perflab_get_standalone_plugins` plugins so that if any one of the plugin is requested it should return error?

plugins/performance-lab/includes/admin/plugins.php
b1ink0: This check is needed because if the `$plugin_slug` is not in the `perflab_get_standalone_plugins()` then this `$plugin_slug` will not be processed inside the loop.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -54,61 +54,94 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
-			);
+			if ( ! isset( $all_performance_plugins[ $current_plugin_slug ] ) ) {
+				// Cache the fact that the plugin was not found.
+				$plugins[ $current_plugin_slug ] = array(
+					'error' => array(
+						'code'    => 'plugin_not_found',
+						'message' => __( 'Plugin not found in API response.', 'performance-lab' ),
+					),
+				);
+				continue;
```
b1ink0: Due to the recent commit [ff1b440](https://github.com/b1ink0/performance/tree/ff1b4409a9e3faa819ab3b5a2796d972643f3080) this could cause incorrect behavour if `$plugin_slug` is not found in the `all_performance_plugins` it will add error for that `$plugin_slug` as we are now using the `$has_errors` flag which will not be set to `true` in this case. The below logic will fail. 
Solution could be to check it using the `isset( $plugins[ $plugin_slug ]['error'] )` condition but it will defect the usage of the `$has_errors` flag.
```php
	if ( $has_errors ) {
		set_transient( $transient_key, $plugins, MINUTE_IN_SECONDS );

		return new WP_Error(
			$plugins[ $plugin_slug ]['error']['code'],
			$plugins[ $plugin_slug ]['error']['message']
		);
	}
```

Reason for the `$has_errors` flag not setting: As the errors could occur for other plugin ( not `$plugin_slug` ) currently code will cache the error but will return the requested `$plugin_slug` information.

Question: Should we return error even if the error is encountered for the other plugin ( not `$plugin_slug` ) ?
cc: @felixarntz

plugins/performance-lab/includes/admin/plugins.php
felixarntz: Correct. That's because a failing API response is by definition an error that affects all plugins.

plugins/performance-lab/includes/admin/plugins.php
felixarntz: In that case, I think we should make this check at the very beginning of the function, even before looking at the transient cache. This function should never be used for a plugin that is not among our plugins I think, so we could already check that before all the main logic in this function.

We wouldn't want that someone calling the function with e.g. `jetpack` would trigger this API request and logic to run for every single page load just because `jetpack` would (rightfully) never be included in the response.

plugins/performance-lab/includes/admin/plugins.php
felixarntz: See my reply in https://github.com/WordPress/performance/pull/1694#discussion_r1860904156 - I think this problem would become irrelevant if we checked that the plugin slug is among our performance plugins in the beginning of the function?

plugins/performance-lab/includes/admin/plugins.php
b1ink0: If we only check the `$plugin_slug` inside `perflab_get_standalone_plugins()` it will also cause error for the dependency plugins. And if we want to dynamically check the dependency then we need to request the API.

felixarntz: Why does the function need to be called for dependency plugins?

b1ink0: The `perflab_render_plugin_card()` function is calling the `perflab_get_plugin_availability()`  function
and inside the `perflab_get_plugin_availability()` on [here](https://github.com/WordPress/performance/blob/69eca9a7936701475f84a0cabbfec36e1cb3fe27/plugins/performance-lab/includes/admin/plugins.php#L310C2-L320C3) it is calling the `perflab_query_plugin_info()` for the dependency plugin.

felixarntz: Ah, that's a good point. But in that case, I think at least this error should use a different message? I would try to clarify it as: If the plugin is not found at this point, we know it's not one of our plugins, since it's not in the API response, and the API response should always included all our plugins _plus_ its dependencies.

plugins/performance-lab/includes/admin/plugins.php
felixarntz: Revisiting this: Why should we _not_ set the `$has_errors` flag to `true` here? I think we should, given this is an error too.

b1ink0: > Reason for the $has_errors flag not setting: As the errors could occur for other plugin ( not $plugin_slug ) currently code will cache the error but will return the requested $plugin_slug information.

> Question: Should we return error even if the error is encountered for the other plugin ( not $plugin_slug ) ?

This above is reason. And doubt related to your question.

Also if we decide to set the `$has_errors` then should we add the same error for all the standalone plugin?

felixarntz: I asked because I couldn't follow your initial explanation. Can you clarify what you mean?

This error is only set for the specific plugin slug that is not found. So it would not affect another plugin's return value. I think what you're referring to is that the function right now always returns a `WP_Error` when there are any errors, but that's not right, so it needs to be fixed. The `$has_errors` flag should only be used to change the transient duration.

After the transient is set, we need to check the individual `$plugins[ $plugin_slug ]` entry and only return a `WP_Error` if _that_ plugin has an error in it.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -54,61 +54,101 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
-	if ( ! isset( $plugins[ $plugin_slug ] ) ) {
 		return new WP_Error(
-			'plugin_not_found',
-			__( 'Plugin not found in API response.', 'performance-lab' )
+			$plugins[ $plugin_slug ]['error']['code'],
+			$plugins[ $plugin_slug ]['error']['message']
 		);
 	}
 
+	set_transient( $transient_key, $plugins, HOUR_IN_SECONDS );
```
felixarntz: See my response in https://github.com/WordPress/performance/pull/1694#discussion_r1866171486, this logic is problematic.

We should only use `$has_errors` to change the transient duration.

The actual return value needs to depend on the specific `$plugins[ $plugin_slug ]` value.

plugins/performance-lab/includes/admin/plugins.php
b1ink0: Ohh thanks for clarification I misunderstood what was suggested before. We should set `$has_errors` if any errors is encountered and use it only for changing the transient duration and only return error if requested `$plugins[ $plugin_slug ]` has a error.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -54,58 +54,102 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
 	}
 
-	$plugins      = array();
-	$plugin_queue = perflab_get_standalone_plugins();
+	// Cache error for all standalone plugins.
+	if ( $has_errors ) {
+		foreach ( perflab_get_standalone_plugins() as $standalone_plugin ) {
+			$plugins[ $standalone_plugin ] = $plugins[ $plugin_slug ];
+		}
+	}
```
felixarntz: This works as is, but is hard to follow in terms of code path, because it relies on `$has_errors`, which is intended for _any_ errors but here it's used to check for whether there are _general_ errors that affect all plugins. It only works correctly because of the position in the code, which is a bit fragile.

I think this would be more intuitive and easier to understand for people looking at the code in the future if we moved this `foreach` loop into both of the individual conditions where general errors are found (i.e. the conditions in line 57 and 76).

This also avoids an extra `if` check here that technically is unnecessary.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -54,58 +54,102 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
 
-	// Index the plugins from the API response by their slug for efficient lookup.
-	$all_performance_plugins = array_column( $response->plugins, null, 'slug' );
+	if ( ! $has_errors && is_object( $response ) && property_exists( $response, 'plugins' ) ) {
```
felixarntz: Instead of reiterating an almost similar condition as above, can we move this to an `else` or `elseif` please?

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -54,58 +54,102 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
+			),
 		);
+
+		$has_errors = true;
 	}
 
 	// Check if the response contains plugins.
-	if ( ! ( is_object( $response ) && property_exists( $response, 'plugins' ) ) ) {
-		return new WP_Error( 'no_plugins', __( 'No plugins found in the API response.', 'performance-lab' ) );
+	if ( ! $has_errors && ! ( is_object( $response ) && property_exists( $response, 'plugins' ) ) ) {
```
felixarntz: I think it would make more sense to have this as an `elseif`. Easier to understand, and then there's no need to check for `! $has_errors`.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -22,12 +22,12 @@ function perflab_query_plugin_info( string $plugin_slug ) {
 	$transient_key = 'perflab_plugins_info';
 	$plugins       = get_transient( $transient_key );
 
-	if ( is_array( $plugins ) ) {
-		// If the specific plugin_slug is not in the cache, return an error.
-		if ( ! isset( $plugins[ $plugin_slug ] ) ) {
+	if ( is_array( $plugins ) && isset( $plugins[ $plugin_slug ] ) ) {
+		if ( isset( $plugins[ $plugin_slug ]['error'] ) ) {
+			// Plugin was requested before and not found.
```
felixarntz:
```suggestion
			// Plugin was requested before but an error occurred for it.
```

felixarntz (commented):
@b1ink0 This technically looks like it's working well, though I think there's a few places where the logic can be simplified / consolidated.

This kind of thing is usually not a big deal, but since the conditions here are a bit complex, I think it would be great to make the code more understandable for people looking at it in the future without all the context from what we've been discussing.

b1ink0:
Some checks are failing because of [disruption](https://www.githubstatus.com/incidents/4349zxvb8stp) to the GItHub services [current status](https://www.githubstatus.com/).
cc: @felixarntz

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -54,58 +54,98 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
-		return new WP_Error( 'no_plugins', __( 'No plugins found in the API response.', 'performance-lab' ) );
-	}
+		foreach ( perflab_get_standalone_plugins() as $standalone_plugin ) {
+			$plugins[ $standalone_plugin ] = $plugins[ $plugin_slug ];
+		}
 
-	$plugins      = array();
-	$plugin_queue = perflab_get_standalone_plugins();
+		$has_errors = true;
+	} elseif ( ! ( is_object( $response ) && property_exists( $response, 'plugins' ) ) ) {
```
felixarntz: Cognitively easier to remove the negation around the whole clause.
```suggestion
	} elseif ( ! is_object( $response ) || ! property_exists( $response, 'plugins' ) ) {
```

felixarntz (approved):
@b1ink0 Thank you for the continuous iterations, this looks great!

One minor nit-pick, but not a blocker.

felixarntz:
@b1ink0 @westonruter FYI I've updated the PR description to indicate this addresses both #1617 and #1691.

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -54,58 +54,98 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
+				continue;
+			}
+
+			$plugin_data                     = $all_performance_plugins[ $current_plugin_slug ];
+			$plugins[ $current_plugin_slug ] = wp_array_slice_assoc( $plugin_data, $fields );
+
+			// Enqueue the required plugins slug by adding it to the queue.
+			if ( isset( $plugin_data['requires_plugins'] ) && is_array( $plugin_data['requires_plugins'] ) ) {
+				$plugin_queue = array_merge( $plugin_queue, $plugin_data['requires_plugins'] );
+			}
```
westonruter: Trivial point, but I think an `else` here makes sense:

```suggestion
			} else {
				$plugin_data                     = $all_performance_plugins[ $current_plugin_slug ];
				$plugins[ $current_plugin_slug ] = wp_array_slice_assoc( $plugin_data, $fields );
	
				// Enqueue the required plugins slug by adding it to the queue.
				if ( isset( $plugin_data['requires_plugins'] ) && is_array( $plugin_data['requires_plugins'] ) ) {
					$plugin_queue = array_merge( $plugin_queue, $plugin_data['requires_plugins'] );
				}
			}
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -54,58 +54,98 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
-		if ( ! isset( $all_performance_plugins[ $current_plugin_slug ] ) ) {
-			return new WP_Error(
-				'plugin_not_found',
-				__( 'Plugin not found in WordPress.org API response.', 'performance-lab' )
-			);
+		// Start processing the plugins using a queue-based approach.
+		while ( count( $plugin_queue ) > 0 ) { // phpcs:ignore Squiz.PHP.DisallowSizeFunctionsInLoops.Found
+			$current_plugin_slug = array_shift( $plugin_queue );
+
+			if ( isset( $plugins[ $current_plugin_slug ] ) ) {
```
westonruter, line 101:
```suggestion
			// Skip already-processed plugins.
			if ( isset( $plugins[ $current_plugin_slug ] ) ) {
```

plugins/performance-lab/includes/admin/plugins.php
```diff
@@ -54,58 +54,98 @@ function perflab_query_plugin_info( string $plugin_slug ) {
...
+			$has_errors = true;
 		}
 	}
 
-	set_transient( $transient_key, $plugins, HOUR_IN_SECONDS );
+	if ( $has_errors ) {
+		set_transient( $transient_key, $plugins, MINUTE_IN_SECONDS );
+	} else {
+		set_transient( $transient_key, $plugins, HOUR_IN_SECONDS );
+	}
```
westonruter: To reduce code duplication:

```suggestion
	set_transient( $transient_key, $plugins, $has_errors ? MINUTE_IN_SECONDS : HOUR_IN_SECONDS );
```

westonruter (approved):

mukeshpanchal27 (approved):

PR #1697, opened 2024-11-22
Preload image URLs for LCP elements with external background images

### LINKED ISSUE #1584: Image Prioritizer should be able to store the LCP (background) image URL for prioritization
## Feature Description

Optimization Detective detects the LCP element and stores its elementa metadata in URL Metrics. It stores aspects including XPath of the element, the `boundingClientRect`, and the `intersectionRatio`. It does not, however, store the actual image URL associated with the LCP element. Instead, at runtime it determines the image URL by inspecting the element at the given XPath at runtime. For `IMG` elements, this means extracting the `src` and `srcset` attributes to populate in the preload link. For elements with CSS background images (e.g. the Cover block or Group block), it parses the image URL out of the inline `style` attribute. However, this only works when the background image is defined inline. It is not able to discover background images defined in external CSS stylesheets or applied via inline stylesheets in `style` elements. For example, Twenty Thirteen's LCP element is the `header#masthead` element:

Image

The background image is applied via the inline style rule:

```html
<style type="text/css" id="twentythirteen-header-css">
.site-header {
	background: url(http://localhost:8888/wp-content/themes/twentythirteen/images/headers/circle.png) no-repeat scroll top;
	background-size: 1600px auto;
}
@media (max-width: 767px) {
	.site-header {
		background-size: 768px auto;
	}
}
@media (max-width: 359px) {
	.site-header {
		background-size: 360px auto;
	}
}
</style>
```

This means that Image Prioritizer can't add a preload link for this background image. I've seen this issue occur more frequently. To work around it, specific integrations would have to be worked out that look up whatever image URL is being emitted into the stylesheet via various PHP APIs for not only WordPress core but also various page builder plugins that also add background images like Divi and Elementor (for the latter, see [POC plugin](https://github.com/westonruter/elementor-lcp-background-image-preloader)). This is not scalable, although **adding a specific integration with the header image would make sense**.

Nevertheless, the LCP metric data in the browser _does_ include the URL of the image:

> Image

But I didn't want to store this and use it for preloading for two reasons:

1. A site builder could select a background image and then change it later, resulting in a period of time in which collected URL metrics could preload the stale background image instead of the new background image. 
2. Someone could attempt to maliciously submit a URL metric including an LCP image URL which is extremely large and isn't actually the correct image, which would hurt performance.

Given the common occurrence for non-inline LCP CSS background images, it seems there should at least be a way to enable this, even if it isn't on by default. Worst case or the first point is that a stale URL would be preloaded while waiting for new URL metrics to be collected. For the second point, some safeguards should be put in place such as to only allow preloading image URLs on the same origin (or whatever CDN the site is using).

# Aside on Randomness

To add an aside to the above first reason about a LCP background-image stored in URL Metrics becoming out-of-date, this possibility also exists in core when someone chooses the "Randomize Suggested Headers" option in the Customizer.

In the Gallery block, a user can select the "random order" setting, but since the `IMG` elements then we can still preload the LCP element's URL dynamically based on whichever URL is currently assigned to that `IMG` tag, although it's true that this `IMG` tag may no longer be the LCP element.

Nevertheless, I do not believe the random header (or random gallery order) to be a popular options. But it is something to keep in mind.

westonruter:
Fixes #1584

This captures the URL for the background image of the LCP element which is defined in a CSS stylesheet and not in an inline `style` attribute. A new client-side extension module from Image Prioritizer is introduced to implement this. A new root property is added to the URL Schema called `lcpElementExternalBackgroundImage` which includes not only the `url` of the background image but also the `tag` name, `id`, `class` for the LCP element that has the background image. When the `Image_Prioritizer_Background_Image_Styled_Tag_Visitor` iterates over tags in the document, it checks to see if there is a matching tag (and thus the original element with the background image is still present). If so, and if all URL Metrics in the group are in agreement on that being the LCP element with that same background image, then the URL for the background image is added as a `fetchpriority=high` preload link for that viewport group. The requirement that all gathered URL Metrics be in agreement helps ensure that if a randomized background image is used (e.g. as can be done in core's header image), then no preloading will be done (as it is likely the wrong image will be preloaded).

# Twenty Thirteen Example

Given the Twenty Thirteen theme which has a CSS background image in the header and some images in the post content, on desktop the header is the LCP element whereas on mobile the first image is the LCP element:

Desktop | Mobile
--|--
2013-desktop | 2013-mobile

With the changes in this PR, the CSS background image gets a `fetchpriority=high` preload link for desktop, whereas the first image in the content gets a preload link for mobile:

```html
<link
  data-od-added-tag
  rel="preload"
  fetchpriority="high"
  as="image"
  href="http://localhost:8888/wp-content/themes/twentythirteen/images/headers/circle.png"
  media="screen and (min-width: 783px)"
>
<link
  data-od-added-tag
  rel="preload"
  fetchpriority="high"
  as="image"
  href="http://localhost:8888/wp-content/uploads/2024/06/bison2-1024x673-jpg.webp"
  imagesrcset="http://localhost:8888/wp-content/uploads/2024/06/bison2-1024x673-jpg.webp 1024w, http://localhost:8888/wp-content/uploads/2024/06/bison2-300x197-jpg.webp 300w, http://localhost:8888/wp-content/uploads/2024/06/bison2-768x505-jpg.webp 768w, http://localhost:8888/wp-content/uploads/2024/06/bison2-1536x1010.jpg 1536w, http://localhost:8888/wp-content/uploads/2024/06/bison2-2048x1347.jpg 2048w, http://localhost:8888/wp-content/uploads/2024/06/bison2-1568x1031.jpg 1568w"
  imagesizes="(max-width: 1024px) 100vw, 1024px"
  media="screen and (max-width: 480px)"
>
```

The impact of preloading the background image on mobile is reflected in the network log, where without the optimization the `circle.png` image is loaded with an initial low priority long after the other assets on the page have started loading:

image

In contrast to when the image is preloaded, it is the initial resource loaded and has an initial high priority:

image

Here are the before/after metrics testing with [`benchmark-web-vitals`](https://github.com/GoogleChromeLabs/wpp-research/tree/main/cli#benchmark-web-vitals) (with https://github.com/GoogleChromeLabs/wpp-research/pull/164) without any page caching:

  | Desktop Before | Desktop After | Mobile Before | Mobile After
-- | --: | --: | --: | --:
FCP (median) | 121.55 | 143.9 | 113.25 | 113.6
LCP (median) | 151.95 | 143.9 | 113.25 | 113.6
TTFB (median) | 30.1 | 32.55 | 32 | 33.75
LCP-TTFB (median) | 122.1 | 111.35 | 80.6 | 79.9

<details><summary>Test setup</summary>

I created a site in Local and set the theme to Twenty Thirteen. I then created a post which had a 3-columns block with the center column being wider. I put an image in each column, so the middle image was larger. I visited the URL in a desktop viewport and mobile viewport multiple times to ensure the URL Metrics were fully populated. I then created a text file called `preload-bgimage-urls.txt` which contained:

```
http://localhost:10053/bison-columns/?optimization_detective_disabled
http://localhost:10053/bison-columns/
```

I then ran 100 iterations before/after on desktop and mobile:

```bash
npm run research benchmark-web-vitals -- --file=preload-bgimage-urls.txt -n 100 -w "1920x1080" -o csv
npm run research benchmark-web-vitals -- --file=preload-bgimage-urls.txt -n 100 -w "360x800" -o csv
```

</details> 

So on desktop, the LCP-TTFB is improved by 8.8%. On mobile, the LCP-TTFB is improved by a lesser degree by 0.87% because WordPress was already adding `fetchpriority=high` to the first image, so the marginal improvement is gained by the preload link.

# Elementor Example

As another test, I created a site with Elementor and the Hello Elementor theme. I used the [Ceramic Studio](https://library.elementor.com/ceramic-workshops-flexbox/) "website kit" to create a page, and I added a mobile-specific image to the hero's second container: 

Desktop | Mobile
--|--
elementor-desktop | elementor-mobile

Elementor implements the images here as background images pulled from an external CSS file (`http://localhost:10053/wp-content/uploads/elementor/css/post-67.css?ver=1732401999`):

```css
.elementor-67 .elementor-element.elementor-element-35a28f8:not(.elementor-motion-effects-element-type-background), .elementor-67 .elementor-element.elementor-element-35a28f8 > .elementor-motion-effects-container > .elementor-motion-effects-layer {
    background-image: url("http://localhost:10053/wp-content/uploads/2024/11/HomePage-Hero.jpg");
    background-position: center center;
    background-repeat: no-repeat;
    background-size: cover;
}
/*...*/
@media(max-width: 767px) {
    /*...*/
    .elementor-67 .elementor-element.elementor-element-35a28f8:not(.elementor-motion-effects-element-type-background), .elementor-67 .elementor-element.elementor-element-35a28f8 > .elementor-motion-effects-container > .elementor-motion-effects-layer {
        background-image: url("http://localhost:10053/wp-content/uploads/2024/11/WorkshopPage-WorkGallery-img_3.jpg");
    }
    /*...*/
}
```

The element being targeted is:

```html
<div
  class="elementor-element elementor-element-35a28f8 e-con-full e-flex e-con e-child"
  data-id="35a28f8"
  data-element_type="container"
  data-settings='{"background_background":"classic"}'
></div>
```

Running the same tests as before with 100 iterations on both desktop and mobile, before and after the optimizations, yields the following results (again where mobile is 360x800 and desktop is 1920x1080):

  | Desktop Before | Desktop After | Mobile Before | Mobile After
-- | --: | --: | --: | --:
FCP (median) | 120.45 | 122.4 | 121.3 | 112.8
LCP (median) | 158.85 | 122.75 | 145.65 | 113.25
TTFB (median) | 54.5 | 39.1 | 55 | 39.4
LCP-TTFB (median) | 103.35 | 84.55 | 91.55 | 71.8

**The LCP-TTFB on desktop is improved by 18.19% and on mobile it is improved by 21.57%!** 🎉 (What is surprising to me as well is that TTFB is reduced when the optimizations are applied, which doesn't make any sense since the HTML Tag Processor spends cycles doing work.)

It's important to note that this page doesn't just have CSS background images. Further down the page outside the viewport of both desktop and mobile, there are three `IMG` tags in another section:

> image

Elementor is adding `fetchpriority=high` to this `IMG` even though it is not even displayed in any initial viewport:

```html
<figure class="elementor-image-box-img">
  <img
    fetchpriority="high"
    decoding="async"
    width="327"
    height="293"
    src="http://localhost:10053/wp-content/uploads/2024/11/HomePage-Offers_1.jpg"
    class="elementor-animation-sink attachment-full size-full wp-image-70"
    alt=""
    srcset="
      http://localhost:10053/wp-content/uploads/2024/11/HomePage-Offers_1.jpg         327w,
      http://localhost:10053/wp-content/uploads/2024/11/HomePage-Offers_1-300x269.jpg 300w
    "
    sizes="(max-width: 327px) 100vw, 327px"
  >
</figure>
```

The Elementor code responsible is the [`maybe_add_fetchpriority_high_attr()` method](https://github.com/elementor/elementor/blob/562b0d21e2a2cef8923d540c0d830bea9b68c2b2/modules/image-loading-optimization/module.php#L315-L347) which appears to be heavily inspired by what WordPress core does.

Here's a diff of the page (with Prettier formatting). Note how Image Prioritizer is adding responsive preload links for the two different background images while at the same time removing `fetchpriority=high` from the `IMG` that Elementor added it to. In addition, Image Prioritizer is adding `loading=lazy` and `sizes=auto` to all of these images since none of them appear in the initial viewport on desktop or mobile:

```diff
--- /tmp/disabled.html	2024-11-23 21:33:35.088604290 -0800
+++ /tmp/enabled.html	2024-11-23 21:33:42.844604516 -0800
@@ -712,6 +712,22 @@
       }
     </style>
     <meta name="generator" content="image-prioritizer 0.2.0" />
+    <link
+      data-od-added-tag
+      rel="preload"
+      fetchpriority="high"
+      as="image"
+      href="http://localhost:10053/wp-content/uploads/2024/11/HomePage-Hero.jpg"
+      media="screen and (min-width: 783px)"
+    />
+    <link
+      data-od-added-tag
+      rel="preload"
+      fetchpriority="high"
+      as="image"
+      href="http://localhost:10053/wp-content/uploads/2024/11/WorkshopPage-WorkGallery-img_3.jpg"
+      media="screen and (max-width: 480px)"
+    />
   </head>
   <body
     class="home page-template page-template-elementor_header_footer page page-id-67 elementor-default elementor-template-full-width elementor-kit-23 elementor-page elementor-page-67"
@@ -970,7 +986,11 @@
                   <div class="elementor-image-box-wrapper">
                     <figure class="elementor-image-box-img">
                       <img
-                        fetchpriority="high"
+                        data-od-added-loading
+                        data-od-removed-fetchpriority="high"
+                        data-od-replaced-sizes="(max-width: 327px) 100vw, 327px"
+                        data-od-xpath="/*[1][self::HTML]/*[2][self::BODY]/*[3][self::DIV]/*[3][self::DIV]/*[2][self::DIV]/*[2][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[1][self::FIGURE]/*[1][self::IMG]"
+                        loading="lazy"
                         decoding="async"
                         width="327"
                         height="293"
@@ -981,7 +1001,7 @@
                           http://localhost:10053/wp-content/uploads/2024/11/HomePage-Offers_1.jpg         327w,
                           http://localhost:10053/wp-content/uploads/2024/11/HomePage-Offers_1-300x269.jpg 300w
                         "
-                        sizes="(max-width: 327px) 100vw, 327px"
+                        sizes="auto, (max-width: 327px) 100vw, 327px"
                       />
                     </figure>
                     <div class="elementor-image-box-content">
@@ -1032,6 +1052,10 @@
                   <div class="elementor-image-box-wrapper">
                     <figure class="elementor-image-box-img">
                       <img
+                        data-od-added-loading
+                        data-od-replaced-sizes="(max-width: 327px) 100vw, 327px"
+                        data-od-xpath="/*[1][self::HTML]/*[2][self::BODY]/*[3][self::DIV]/*[3][self::DIV]/*[2][self::DIV]/*[2][self::DIV]/*[2][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[1][self::FIGURE]/*[1][self::IMG]"
+                        loading="lazy"
                         decoding="async"
                         width="327"
                         height="293"
@@ -1042,7 +1066,7 @@
                           http://localhost:10053/wp-content/uploads/2024/11/HomePage-Offers_2.jpg         327w,
                           http://localhost:10053/wp-content/uploads/2024/11/HomePage-Offers_2-300x269.jpg 300w
                         "
-                        sizes="(max-width: 327px) 100vw, 327px"
+                        sizes="auto, (max-width: 327px) 100vw, 327px"
                       />
                     </figure>
                     <div class="elementor-image-box-content">
@@ -1093,6 +1117,10 @@
                   <div class="elementor-image-box-wrapper">
                     <figure class="elementor-image-box-img">
                       <img
+                        data-od-added-loading
+                        data-od-replaced-sizes="(max-width: 327px) 100vw, 327px"
+                        data-od-xpath="/*[1][self::HTML]/*[2][self::BODY]/*[3][self::DIV]/*[3][self::DIV]/*[2][self::DIV]/*[2][self::DIV]/*[3][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[1][self::FIGURE]/*[1][self::IMG]"
+                        loading="lazy"
                         decoding="async"
                         width="327"
                         height="293"
@@ -1103,7 +1131,7 @@
                           http://localhost:10053/wp-content/uploads/2024/11/HomePage-Offers_3.jpg         327w,
                           http://localhost:10053/wp-content/uploads/2024/11/HomePage-Offers_3-300x269.jpg 300w
                         "
-                        sizes="(max-width: 327px) 100vw, 327px"
+                        sizes="auto, (max-width: 327px) 100vw, 327px"
                       />
                     </figure>
                     <div class="elementor-image-box-content">
@@ -1388,6 +1416,9 @@
                 >
                   <figure class="swiper-slide-inner">
                     <img
+                      data-od-added-loading
+                      data-od-xpath="/*[1][self::HTML]/*[2][self::BODY]/*[3][self::DIV]/*[5][self::DIV]/*[2][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[1][self::FIGURE]/*[1][self::IMG]"
+                      loading="lazy"
                       decoding="async"
                       class="swiper-slide-image"
                       src="http://localhost:10053/wp-content/uploads/2024/11/HomePage-WorkGallery-img_1.jpg"
@@ -1403,6 +1434,9 @@
                 >
                   <figure class="swiper-slide-inner">
                     <img
+                      data-od-added-loading
+                      data-od-xpath="/*[1][self::HTML]/*[2][self::BODY]/*[3][self::DIV]/*[5][self::DIV]/*[2][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[2][self::DIV]/*[1][self::FIGURE]/*[1][self::IMG]"
+                      loading="lazy"
                       decoding="async"
                       class="swiper-slide-image"
                       src="http://localhost:10053/wp-content/uploads/2024/11/HomePage-WorkGallery-img_2.jpg"
@@ -1418,6 +1452,9 @@
                 >
                   <figure class="swiper-slide-inner">
                     <img
+                      data-od-added-loading
+                      data-od-xpath="/*[1][self::HTML]/*[2][self::BODY]/*[3][self::DIV]/*[5][self::DIV]/*[2][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[3][self::DIV]/*[1][self::FIGURE]/*[1][self::IMG]"
+                      loading="lazy"
                       decoding="async"
                       class="swiper-slide-image"
                       src="http://localhost:10053/wp-content/uploads/2024/11/HomePage-WorkGallery-img_3.jpg"
@@ -1433,6 +1470,9 @@
                 >
                   <figure class="swiper-slide-inner">
                     <img
+                      data-od-added-loading
+                      data-od-xpath="/*[1][self::HTML]/*[2][self::BODY]/*[3][self::DIV]/*[5][self::DIV]/*[2][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[1][self::DIV]/*[4][self::DIV]/*[1][self::FIGURE]/*[1][self::IMG]"
+                      loading="lazy"
                       decoding="async"
                       class="swiper-slide-image"
                       src="http://localhost:10053/wp-content/uploads/2024/11/HomePage-WorkGallery-img_4.jpg"
@@ -1722,5 +1762,32 @@
       src="http://localhost:10053/wp-content/plugins/elementor/assets/js/frontend.min.js?ver=3.25.9"
       id="elementor-frontend-js"
     ></script>
+
+    <script type="module">
+      import detect from "http:\/\/localhost:10053\/wp-content\/plugins\/optimization-detective\/detect.min.js?ver=0.9.0-alpha";
+      detect({
+        minViewportAspectRatio: 0.40000000000000002,
+        maxViewportAspectRatio: 2.5,
+        isDebug: false,
+        extensionModuleUrls: [
+          "http:\/\/localhost:10053\/wp-content\/plugins\/image-prioritizer\/detect.min.js?ver=0.2.0",
+        ],
+        restApiEndpoint:
+          "http:\/\/localhost:10053\/wp-json\/optimization-detective\/v1\/url-metrics:store",
+        currentUrl: "http:\/\/localhost:10053\/",
+        urlMetricSlug: "d751713988987e9331980363e24189ce",
+        cachePurgePostId: 67,
+        urlMetricHMAC: "35086facb74f518514588b7e83751043",
+        urlMetricGroupStatuses: [
+          { minimumViewportWidth: 0, complete: true },
+          { minimumViewportWidth: 481, complete: false },
+          { minimumViewportWidth: 601, complete: false },
+          { minimumViewportWidth: 783, complete: true },
+        ],
+        storageLockTTL: 60,
+        webVitalsLibrarySrc:
+          "http:\/\/localhost:10053\/wp-content\/plugins\/optimization-detective\/build\/web-vitals.js?ver=4.2.4",
+      });
+    </script>
   </body>
 </html>
```

Network log without the optimizations up until the LCP element's background image is loaded:

image

Compared with after the optimizations applied:

image

Note how the LCP element's background image is now loaded as early as possible with initial `high` priority, whereas without the optimizations the image is loaded very late and has an initial priority of `low`.

# Takeaway

This represents an critical performance advancement for optimizing LCP in WordPress because on the web a `DIV` is the second most common LCP element after `IMG`. Since images account for the LCP type 82% desktop and and 72% on mobile, it's likely that most of the `DIV` LCP elements represent background images. Additionally, page builders like Elementor and Divi leverage external background images extensively, including separate background images for desktop and mobile.

[Top LCP element types](https://almanac.httparchive.org/en/2022/performance#fig-7) | [Top LCP content types](https://almanac.httparchive.org/en/2022/performance#fig-8)
--|--
[Top LCP element types](https://almanac.httparchive.org/en/2022/performance#lcp-content-types) | Top LCP content types 

# Todo

- [x] Add additional validation constraints on the new root property (e.g. require same-origin).
- [x] Add tests.


westonruter:
@felixarntz This is getting close! I'm excited about this one since it unlocks a very common use case which we've seen especially among page builders. This unlocks potential optimization of the `DIV` (assuming it has a background image) which is the second most common LCP element.

westonruter:
I've just updated the description with findings on how this [Ceramic Studio](https://library.elementor.com/ceramic-workshops-flexbox/). In short, I'm seeing a ~20% improvement in LCP on both desktop and mobile!

westonruter:
I've just updated the description with findings on how this [impacts Elementor](https://github.com/WordPress/performance/pull/1697#:~:text=the%20preload%20link.-,Elementor%20Example,-As%20another%20test). In short, I'm seeing a ~20% improvement in LCP on both desktop and mobile!

plugins/optimization-detective/detect.js
```diff
@@ -127,6 +127,12 @@ function recursiveFreeze( obj ) {
 	Object.freeze( obj );
 }
 
+// This needs to be captured early in case the user later resizes the window.
+const viewport = {
+	width: win.innerWidth,
+	height: win.innerHeight,
+};
+
```
westonruter: I added this here to capture the initial width/height when the page loads. Later when about to submit the metrics for storage, it checks to see if the viewport width or height has changed, and then it aborts. This is important because if the viewport window size changes during the life of the page, all bets are off as to what is the expected LCP element.

plugins/image-prioritizer/hooks.php
```diff
@@ -11,22 +11,6 @@
...
- */
-function image_prioritizer_get_lazy_load_script(): string {
-	$script = file_get_contents( __DIR__ . sprintf( '/lazy-load%s.js', wp_scripts_get_suffix() ) ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents -- It's a local filesystem path not a remote request.
-
-	if ( false === $script ) {
-		return '';
-	}
-
-	return $script;
-}
```
westonruter: This function was just moved to `helper.php`.

plugins/optimization-detective/types.ts
```diff
@@ -30,9 +30,10 @@ export interface URLMetricGroupStatus {
 
 export type InitializeArgs = {
 	readonly isDebug: boolean;
+	readonly webVitalsLibrarySrc: string;
 };
 
-export type InitializeCallback = ( args: InitializeArgs ) => void;
+export type InitializeCallback = ( args: InitializeArgs ) => Promise< void >;
```
westonruter, line 48: This is now consistent with the `FinalizeCallback`.

plugins/image-prioritizer/helper.php
```diff
@@ -76,3 +76,137 @@ function image_prioritizer_register_tag_visitors( OD_Tag_Visitor_Registry $regis
...
+ */
+function image_prioritizer_get_lazy_load_script(): string {
+	$script = file_get_contents( __DIR__ . sprintf( '/lazy-load%s.js', wp_scripts_get_suffix() ) ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents -- It's a local filesystem path not a remote request.
+
+	if ( false === $script ) {
+		return '';
+	}
+
+	return $script;
+}
```
westonruter: This function was just moved from `hooks.php`.

plugins/optimization-detective/detect.js
```diff
@@ -340,23 +346,52 @@ export default async function detect( {
...
-			// TODO: There should to be a way to pass additional args into the module. Perhaps extensionModuleUrls should be a mapping of URLs to args. It's important to pass webVitalsLibrarySrc to the extension so that onLCP, onCLS, or onINP can be obtained.
+			// TODO: There should to be a way to pass additional args into the module. Perhaps extensionModuleUrls should be a mapping of URLs to args.
 			if ( extension.initialize instanceof Function ) {
-				extension.initialize( { isDebug } );
+				extensionInitializePromises.push(
+					extension.initialize( {
+						isDebug,
+						webVitalsLibrarySrc,
+					} )
+				);
```
westonruter: This needs to account for an extension (erroneously) not implementing `initialize` as an async function, in which case it will not return a promise. It should check for a `Promise` return value.

plugins/optimization-detective/detect.js
```diff
@@ -505,28 +537,67 @@ export default async function detect( {
...
-						extendRootData,
-					} );
+					extensionFinalizePromises.push(
+						extension.finalize( {
+							isDebug,
+							getRootData,
+							getElementData,
+							extendElementData,
+							extendRootData,
+						} )
```
westonruter: This needs to account for an extension (erroneously) not implementing `finalize` as an async function, in which case it will not return a promise. It should check for a `Promise` return value.

westonruter:
> **The LCP-TTFB on desktop is improved by 18.19% and on mobile it is improved by 21.57%!** 🎉 (What is surprising to me as well is that TTFB is reduced when the optimizations are applied, which doesn't make any sense since the HTML Tag Processor spends cycles doing work.)

There seems to be a problem with the TTFB reporting in `benchmark-web-vitals`. I ran `benchmark-server-timing` (with https://github.com/GoogleChromeLabs/wpp-research/pull/165):

```bash
npm run research -- benchmark-server-timing --file=preload-bgimage-urls.txt --number 100 -o csv
```

And I got:

Metric | OD Disabled | OD Enabled
-- | --: | --:
Response Time (median) | 29.13 | 31.53
wp-before-template (median) | 12.31 | 12
wp-template (median) | 15.22 | 18.27
wp-total (median) | 27.99 | 30.41
wp-optimization-detective (median) |   | 3.15

This is what I was expecting for the impact of Optimization Detective to have on TTFB: ~3 milliseconds.

I just re-ran `benchmark-web-vitals`:

```bash
npm run research benchmark-web-vitals -- --file=preload-bgimage-urls.txt -n 100 -o csv
```

And now I'm getting a more expected result where TTFB is slower when OD is enabled (without page caching):

Metric | OD Disabled | OD Enabled
-- | --: | --:
FCP (median) | 110 | 124.95
LCP (median) | 141.3 | 124.95
TTFB (median) | 36.1 | 42.05
LCP-TTFB (median) | 103.2 | 82.05

The LCP-TTFB remains at a ~20% improvement, and here the absolute LCP including TTFB is a 11.57% improvement.

Note that in these commands, `preload-bgimage-urls.txt` contains:

```
http://localhost:10053/?optimization_detective_disabled=1
http://localhost:10053/
```

plugins/image-prioritizer/helper.php
```diff
@@ -77,6 +77,142 @@ function image_prioritizer_register_tag_visitors( OD_Tag_Visitor_Registry $regis
...
+		$validity = (bool) $validity;
+	}
+
+	$data = $url_metric->get( 'lcpElementExternalBackgroundImage' );
+	if ( ! is_array( $data ) ) {
+		return $validity;
+	}
+
+	$r = wp_safe_remote_head(
+		$data['url'],
```
westonruter: The `wp_safe_remote_head()` function passes the supplied URL through `wp_http_validate_url()`, but that doesn't necessarily mean the request is safe. I'm thinking there should perhaps be an allowlist of hosts that this is allowed to connect to. Maybe we query for the first image attachment and pass it into `wp_get_attachment_image_url()` to then extract the hostname from the image. This would be better than having to try to come up with a list of all image CDNs that are used by plugins.

The list of allowed hosts could also be sent to the frontend so as to avoid sending the invalid URL in the first place. See related todo:

https://github.com/WordPress/performance/blob/bc81588943e835707d70948bdd6a3c817c948a25/plugins/optimization-detective/detect.js#L361-L366

Then again, would be a shape to continuously reject all of the URL Metrics in the case that an unexpected host is being used for a given background image. In that case, perhaps we should allow `OD_URL_Metric` to be mutable, so in the case where an invalid host is provided this callback could instead do:

```php
$url_metric->set( 'lcpElementExternalBackgroundImage', null );
```

or rather

```php
$url_metric->unset( 'lcpElementExternalBackgroundImage' );
```

When mutating an instance of `OD_URL_Metric` it should re-trigger the validation so that the mutation would cause the JSON Schema to be made invalid that it would throw an exception.

Also, whenever mutation happens then it should call a new `$this->group->clear_cache()` method which empties out `$group->result_cache` and which would in turn call `$collection->clear_cache()`.

westonruter: So this `od_store_url_metric_validity` filter could be used in two ways:

1. A filter callback could return `false` or a non-empty `WP_Error` instance to block the URL Metric from being stored at all.
2. A filter callback could mutate the supplied `OD_URL_Metric` object to apply custom sanitizations or validation constraints which then would end up getting persisted in storage.

plugins/optimization-detective/storage/rest-api.php
```diff
@@ -209,6 +210,35 @@ function od_handle_rest_request( WP_REST_Request $request ) {
...
+	 * This filter only applies when storing a URL Metric via the REST API. It does not run when a stored URL Metric
+	 * loaded from the od_url_metric post type. This means that validation logic enforced via this filter can be more
+	 * expensive, such as doing filesystem checks or HTTP requests.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param bool|WP_Error        $validity   Validity. Valid if true or a WP_Error without any errors, or invalid otherwise.
+	 * @param OD_Strict_URL_Metric $url_metric URL Metric, already validated against the JSON Schema.
+	 */
+	$validity = apply_filters( 'od_store_url_metric_validity', true, $url_metric );
```
westonruter: To compliment https://github.com/WordPress/performance/pull/1697/files#r1859381128, this would allow a filter callback to always have access to whatever the original data was, regardless of any changes which have been made by other callbacks to the `$url_metric` object. This would be analogous to the second `$postarr` argument to the `wp_insert_post_data` filter, after the filtered `$data` arg.

```suggestion
	 * @param bool|WP_Error        $validity        Validity. Valid if true or a WP_Error without any errors, or invalid otherwise.
	 * @param OD_Strict_URL_Metric $url_metric      URL Metric, already validated against the JSON Schema.
	 * @param array<string, mixed> $url_metric_data Original URL Metric data before any mutations.
	 */
	$validity = apply_filters( 'od_store_url_metric_validity', true, $url_metric, $url_metric->jsonSerialize() );
```

plugins/image-prioritizer/detect.js
```diff
@@ -0,0 +1,227 @@
...
+	const { onLCP } = await import( webVitalsLibrarySrc );
+	onLCP(
+		( /** @type {LCPMetric} */ metric ) => {
+			handleLCPMetric( metric, isDebug );
+		},
+		{
+			// This avoids needing to click to finalize LCP candidate. While this is helpful for testing, it also
+			// ensures that we always get an LCP candidate reported. Otherwise, the callback may never fire if the
+			// user never does a click or keydown, per <https://github.com/GoogleChrome/web-vitals/blob/07f6f96/src/onLCP.ts#L99-L107>.
+			reportAllChanges: true,
```
adamsilverstein, line 75: nice

plugins/image-prioritizer/detect.js
```diff
@@ -0,0 +1,227 @@
...
+		// Skip URLs that are excessively long. This is the maxLength defined in image_prioritizer_add_element_item_schema_properties().
+		if ( entry.url.length > 500 ) {
+			if ( isDebug ) {
+				log( `Skipping very long URL: ${ entry.url }` );
+			}
+			return;
+		}
+
+		// Also skip Custom Elements which have excessively long tag names.
+		if ( entry.element.tagName.length > 25 ) {
```
adamsilverstein: curious where these limits come from/how you decided on them.

plugins/image-prioritizer/helper.php
```diff
@@ -77,6 +77,142 @@ function image_prioritizer_register_tag_visitors( OD_Tag_Visitor_Registry $regis
...
 }
 
+/**
+ * Gets the script to lazy-load videos.
+ *
+ * Load a video and its poster image when it approaches the viewport using an IntersectionObserver.
+ *
+ * Handles 'autoplay' and 'preload' attributes accordingly.
+ *
+ * @since 0.2.0
```
adamsilverstein: is this since annotation correct? not 'n.e.x.t'?

westonruter: I had forgotten to add this in a previous release, so this is fixing up some missing `@since` tags.

plugins/image-prioritizer/helper.php
```diff
@@ -77,6 +77,142 @@ function image_prioritizer_register_tag_visitors( OD_Tag_Visitor_Registry $regis
...
+ */
+function image_prioritizer_get_lazy_load_script(): string {
+	$path   = image_prioritizer_get_asset_path( 'lazy-load.js' );
+	$script = file_get_contents( __DIR__ . '/' . $path ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents -- It's a local filesystem path not a remote request.
```
adamsilverstein: Any advantage or disadvantage to using `WP_Filesystem_Direct` for local file access?

plugins/image-prioritizer/helper.php
```diff
@@ -77,6 +77,142 @@ function image_prioritizer_register_tag_visitors( OD_Tag_Visitor_Registry $regis
...
+/**
+ * Validates that the provided background image URL is valid.
+ *
+ * @since n.e.x.t
+ *
+ * @param bool|WP_Error|mixed  $validity   Validity. Valid if true or a WP_Error without any errors, or invalid otherwise.
+ * @param OD_Strict_URL_Metric $url_metric URL Metric, already validated against the JSON Schema.
+ * @return bool|WP_Error Validity. Valid if true or a WP_Error without any errors, or invalid otherwise.
+ */
+function image_prioritizer_filter_store_url_metric_validity( $validity, OD_Strict_URL_Metric $url_metric ) {
```
adamsilverstein: is part of the goal here to validate that the stored path actually returns an image? Is this something that could be checked on the client side?

plugins/image-prioritizer/tests/test-helper.php
```diff
@@ -33,7 +33,7 @@ public function data_provider_to_test_image_prioritizer_init(): array {
 				'expected' => false,
 			),
 			'with_new_version' => array(
-				'version'  => '0.7.0',
+				'version'  => '99.0.0',
```
adamsilverstein, line 36: why the giant version bump?

plugins/optimization-detective/class-od-url-metric-group.php
```diff
@@ -24,13 +24,17 @@ final class OD_URL_Metric_Group implements IteratorAggregate, Countable, JsonSer
 	/**
 	 * URL Metrics.
 	 *
+	 * @since 0.1.0
```
adamsilverstein, line 27: not 'n.e.x.t'?

plugins/image-prioritizer/detect.js
westonruter: The longest tags in HTML are 10 characters: `FIGCAPTION` and `BLOCKQUOTE`. Oh, but in SVG the longest is `feComponentTransfer`. For Custom Element names, they could be much longer, so I recall I just asked Gemini what the longest known Custom Elements were known to be and gave this gut guess. 

However, I just checked [Element Popularity data](https://docs.google.com/spreadsheets/d/1grkd2_1xSV3jvNK6ucRQ0OL1HmGTsScHuwA8GZuRLHU/edit?gid=2057119066#gid=2057119066) [from Web Almanac 2022](https://almanac.httparchive.org/en/2022/markup#fig-11):

There are 9 Custom Elements that have a length longer than 25:

Count | Custom Element
-- | --
26 | i-amphtml-scroll-container
26 | feefowidget-container-main
26 | mat-expansion-panel-header
26 | broadstreet-zone-container
27 | s_if_var_sidebarcategoryuse
27 | popup_notifications_wrapper
28 | mfa-finish-enrolling-message
28 | enroll-extra-factors-message
32 | more-verification-options-button

So probably should make this limit 50 and then add a note explaining where thus number comes from.

westonruter: Oh, here's [updated data](https://docs.google.com/spreadsheets/d/1TtOMr_w58HvqNBv4RIWX021Lxm6m5ajYOcRykrPdAJc/edit?gid=1606033584#gid=1606033584) [from 2024](https://almanac.httparchive.org/en/2024/markup#fig-18):

Client | Length | Custom Element | Pages
-- | -- | -- | --:
desktop | 32 | header-sidebar-collapsible-panel | 12922
mobile | 32 | header-sidebar-collapsible-panel | 14726
desktop | 29 | featured-collections-carousel | 5844
mobile | 29 | featured-collections-carousel | 6538
desktop | 28 | discourse-assets-stylesheets | 6834
mobile | 28 | discourse-assets-stylesheets | 6288
mobile | 27 | wc-order-attribution-inputs | 67920
desktop | 27 | wc-order-attribution-inputs | 50385
desktop | 26 | oem-container-bootstrapper | 5477
mobile | 26 | announcement-bar-component | 6972
desktop | 26 | announcement-bar-component | 5517

plugins/image-prioritizer/helper.php
westonruter: The goal is to validate that the URL is valid. Since a malicious client could submit a URL to point to anything (e.g. even some 3P tracker URL), it's important to validate that (1) it's actually an image, (2) it's on a safe host.

plugins/optimization-detective/class-od-url-metric-group.php
westonruter: No because this was introduced in the first release. I had forgotten to add this.

plugins/image-prioritizer/helper.php
westonruter: Actually, `WP_Filesystem_Direct` uses the same:

https://github.com/WordPress/wordpress-develop/blob/daad8631f707f730530711ed874efe94c7fec968/src/wp-admin/includes/class-wp-filesystem-direct.php#L30-L40

And WordPress core uses `file_get_contents()` itself in similar situations:

* https://github.com/WordPress/wordpress-develop/blob/daad8631f707f730530711ed874efe94c7fec968/src/wp-includes/class-wp-editor.php#L527
* https://github.com/WordPress/wordpress-develop/blob/daad8631f707f730530711ed874efe94c7fec968/src/wp-includes/embed.php#L520-L522
* https://github.com/WordPress/wordpress-develop/blob/daad8631f707f730530711ed874efe94c7fec968/src/wp-includes/embed.php#L1088
* https://github.com/WordPress/wordpress-develop/blob/daad8631f707f730530711ed874efe94c7fec968/src/wp-includes/formatting.php#L5947

plugins/image-prioritizer/tests/test-helper.php
westonruter: I had done this before for Embed Optimizer in 2c799add22ee5cdda8f795b6aa555a56ae0c3b03 but neglected to do the same here. This test is to ensure that the init logic for Image Prioritizer does if the installed version of Optimization Detective required is greater than or equal to the required version. I didn't have much foresight when I picked the original `0.7.0` version since that version soon became the actual version. So this picks a version way off into the future (which will never be reached) to test that the logic always runs as expected.

plugins/image-prioritizer/detect.js
westonruter: Length increased to 100 in d2e90d36e.

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
```diff
@@ -65,33 +79,126 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+		foreach ( $this->group_common_lcp_element_external_background_images as $i => list( $group, $common ) ) {
+			if (
+				// Note that the browser may send a lower-case tag name in the case of XHTML or embedded SVG/MathML, but
+				// the HTML Tag Processor is currently normalizing to all upper-case. The HTML Processor on the other
+				// hand may return the expected case.
+				strtoupper( $common['tag'] ) === $tag_name
+				&&
+				$processor->get_attribute( 'id' ) === $common['id'] // May be checking equality with null.
+				&&
+				$processor->get_attribute( 'class' ) === $common['class'] // May be checking equality with null.
```
felixarntz, line 172: I don't understand why these checks are needed, can you explain please?

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
```diff
@@ -65,33 +79,126 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+			) {
+				$this->add_preload_link( $context->link_collection, $group, $common['url'] );
+
+				// Now that the preload link has been added, eliminate the entry to stop looking for it while iterating over the rest of the document.
+				unset( $this->group_common_lcp_element_external_background_images[ $i ] );
```
felixarntz, line 177: Can this not impact the `foreach` loop in a problematic way, given this same array we're iterating through? 🤔

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
```diff
@@ -65,33 +79,126 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+		// Gather the tuples of URL Metric group and the common LCP element external background image.
+		if ( ! is_array( $this->group_common_lcp_element_external_background_images ) ) {
+			$this->group_common_lcp_element_external_background_images = array();
+			foreach ( $context->url_metric_group_collection as $group ) {
+				$common = $this->get_common_lcp_element_external_background_image( $group );
+				if ( is_array( $common ) ) {
+					$this->group_common_lcp_element_external_background_images[] = array( $group, $common );
+				}
+			}
+		}
```
felixarntz, line 153: Why is this being written to a class property, rather than just a local variable? Is it for memoization? If so, why would this method be called multiple times for the same context?

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
```diff
@@ -65,33 +79,126 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+	/**
+	 * Adds an image preload link for the group.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param OD_Link_Collection  $link_collection Link collection.
+	 * @param OD_URL_Metric_Group $group           URL Metric group.
+	 * @param non-empty-string    $url             Image URL.
+	 */
+	private function add_preload_link( OD_Link_Collection $link_collection, OD_URL_Metric_Group $group, string $url ): void {
```
felixarntz: Consider adding this as a public method on `OD_Link_Collection` directly (then without the first parameter of course), wrapping its `add_link()` method.

Regardless of whether it remains here or there, I think it would be good to rename it to `add_image_preload_link()`, as it's only for images but that's not obvious from the method name.

plugins/image-prioritizer/helper.php
felixarntz: Where does this function come from? It seems unrelated and the PR diff makes it look like it's new? 🤔

plugins/image-prioritizer/helper.php
felixarntz: I think this would be good to open a separate issue for. This PR is already quite big as is, and this could be an iterative enhancement.

plugins/optimization-detective/detect.js
felixarntz: This seems to be a separate fix, that relates to the problem this PR is solving, but is relevant on its own too? Ideally this kind of thing would be its own PR (and then you could use sub-PRs), as it's really difficult to review large PRs like this with multiple changes :)

plugins/optimization-detective/detect.js
```diff
@@ -507,28 +540,70 @@ export default async function detect( {
...
+		window.innerWidth !== urlMetric.viewport.width ||
+		window.innerHeight !== urlMetric.viewport.height
+	) {
+		if ( isDebug ) {
+			log(
+				'Aborting URL Metric collection due to viewport size change.'
+			);
+		}
+		return;
+	}
```
felixarntz: Couldn't this be moved much further up in the logic, given we'll know whether the viewport has changed far earlier in this function? This could avoid doing quite a bit of "useless" processing as far as I can tell.

plugins/optimization-detective/storage/rest-api.php
```diff
@@ -209,6 +210,35 @@ function od_handle_rest_request( WP_REST_Request $request ) {
...
+	if ( false === $validity || ( $validity instanceof WP_Error && $validity->has_errors() ) ) {
+		if ( false === $validity ) {
+			$validity = new WP_Error( 'invalid_url_metric', __( 'Validity of URL Metric was rejected by filter.', 'optimization-detective' ) );
+		}
+		if ( ! isset( $validity->error_data['code'] ) ) {
+			$validity->error_data['code'] = 400;
+		}
```
felixarntz: Doesn't this need to use `status` instead of `code`?

felixarntz (commented):
@westonruter This looks solid as far as I can tell, but I have a few questions as it's a large and complex PR to review.

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
westonruter: These checks are looking to see if the HTML Processor is currently at a tag which matches what a client had reported being the LCP element with a background image. The client amends the URL Metric with the `lcpElementExternalBackgroundImage` data which includes the tag name, ID, and class name of the element. As a safeguard to ensure we only preload a background image if it is still relevant, these checks ensure that the tag is still present in the document.

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
westonruter: I had wondered that as well, but I found that `foreach()` operates on a copy of the array, so it is safe to unset keys in the loop.

Examples: 
https://3v4l.org/OOlpa
https://3v4l.org/oQF2h

But this could be made less concerning upon inspection if this iterates over `array_keys()` instead.

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
felixarntz: That makes sense, thanks. That said, I wonder is relying on the web vitals library and its LCP detection the best approach for this functionality? I'm asking mostly since I find it a bit fragile to rely on the LCP element detection and then compare some "key" attributes of the element to assume it's the same. Is there no way to determine the CSS background image for an element by using e.g. `computedStyle` in JS?

westonruter: Hummm. We'd still have to rely on Web Vitals to detect the LCP element regardless. We could use `getComputedStyle` on the LCP element to obtain the CSS background image, but there is a cost to getting the computed style which we can avoid by just using the `url` provided by the `LCPMetric`.

plugins/image-prioritizer/helper.php
westonruter: Sorry. This was a merge conflict resolution mistake on my part. This function was renamed from `image_prioritizer_get_lazy_load_script()` to `image_prioritizer_get_video_lazy_load_script()` in #1708 since now there is also `image_prioritizer_get_lazy_load_bg_image_script()`. Removed in f54673128.

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
felixarntz: That's fair.

It feels a bit fragile to me though because we'd be relying on Web Vitals for something we shouldn't need it for. For the LCP detection, it makes sense to use the library, and I get that it gives us the background image "for free". But what about background images for elements that are not the LCP (for instance if we wanted to lazy-load them)? Thinking this further than only the use-case this PR is addressing, I think it would make sense to solve this in a way that is decoupled from the LCP detection.

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
westonruter: Yes, it is for memoization. This class is a tag visitor which is invoked for every tag in the document. So above in the `__invoke()` method there is this logic:

```php
		if ( is_null( $background_image_url ) ) {
			$this->maybe_preload_external_lcp_background_image( $context );
			return false;
		}
```

So the first time that the tag visitor is invoked, it populates this class variable with the data from the collection. If there is are LCP background image elements for the viewport groups, then immediately below it short-circuits:

```php
		// There are no common LCP background images, so abort.
		if ( count( $this->group_common_lcp_element_external_background_images ) === 0 ) {
			return;
		}
```

Otherwise it then proceeds to check if the current tag we're visiting is the same tag that was reported as the LCP element which had a background image, and if so, it adds the preload link (further below).

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
westonruter: I'm not sure what's fragile about it? The `LargestContentfulPaint` object has a [`url` property](https://developer.mozilla.org/en-US/docs/Web/API/LargestContentfulPaint/url) which we're reusing.

For background images which are not LCP, lazy-loading them is now being handled as of #1708. However, only background images which are defined inline via `style` attributes are currently lazy-loaded. This is because the tag visitor is able to see that a given element has a background image. For lazy-loading background images which aren't defined with inline `style` attributes, we'd have to compute the style for every single element on the page which would be quite expensive.

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
westonruter: Addressed in ec59d85fa

plugins/optimization-detective/detect.js
westonruter: You're right, it was a separate fix. But it was something that tripped me up when I was working on this PR. I've just cherry-picked it into a new PR: https://github.com/WordPress/performance/pull/1712

Once you review and merge that, I'll merge `trunk` back into this branch to remove it from the diff.

plugins/optimization-detective/detect.js
westonruter: Resolving in favor of https://github.com/WordPress/performance/pull/1712/files#r1866526591

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
felixarntz: Ah okay, this would make sense from a performance perspective.

So you're saying we're _not_ going to add lazy-loading for background images that aren't inline? If so, then this makes the current approach more reasonable because all that's in scope is to do it for the LCP element anyway. So then using an approach that works only for the LCP element makes more sense.

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
felixarntz: Makes sense.

What throws me off a bit is that the memoization depends on a value that comes from the `$context` but it is memoized independently of that `$context`. I assume the `$group`s would always be the same across all contexts for a page? If so, this makes sense as is. My hesitation comes from the technical dependency on `$context` but not "key-ing" the memoized value with it. So I think we should at least add a comment on why this is not needed here.

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
```diff
@@ -65,33 +79,127 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
+
+	/**
+	 * Maybe preloads external background image.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param OD_Tag_Visitor_Context $context Context.
+	 */
+	private function maybe_preload_external_lcp_background_image( OD_Tag_Visitor_Context $context ): void {
+		// Gather the tuples of URL Metric group and the common LCP element external background image.
```
westonruter, line 143:
```suggestion
		// Gather the tuples of URL Metric group and the common LCP element external background image.
		// Note the groups of URL Metrics do not change across invocations, we just need to compute this once for all.
```

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
westonruter: Correct, the groups of URL Metrics do not change while visiting the tags in the document.

How about this: https://github.com/WordPress/performance/pull/1697/files#r1866624177

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
westonruter: Correct, I don't envision adding lazy-loading for non-inline background images.

plugins/image-prioritizer/helper.php
westonruter: Conversation moved to https://github.com/WordPress/performance/pull/1713/files#r1866642173

plugins/image-prioritizer/helper.php
westonruter: I went ahead and reverted the new filter from this PR and moved it into a follow-up one: https://github.com/WordPress/performance/pull/1713

plugins/optimization-detective/storage/rest-api.php
westonruter: Comment moved to https://github.com/WordPress/performance/pull/1713/files#r1866646772

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
westonruter: Good idea. This is closely related to the direction that https://github.com/WordPress/performance/pull/1707 is going as well.

westonruter: I've just renamed it for now in fbb607664d7198fe70b8f535be42efa4820b0a59.

It's more complicated to make a general method for image preloading because a link may not have an `href` at all in the case of responsive preloads in which `imagesrcset` and `imagesizes` would be used instead (or in addition). It's simpler here for the background image case because it's just a single possible URL value as the `href`.

I'll keep this in mind for a future enhancement for facilitating image preloads, perhaps as part of https://github.com/WordPress/performance/pull/1707 assuming this PR is merged first, or else a separate PR.

plugins/optimization-detective/storage/rest-api.php
westonruter: Good catch. I've applied that in 57a35e4 as part of #1713

plugins/optimization-detective/detect.js
westonruter: Now eliminated from the diff.

plugins/image-prioritizer/helper.php
```diff
@@ -77,6 +77,65 @@ function image_prioritizer_register_tag_visitors( OD_Tag_Visitor_Registry $regis
...
+/**
+ * Filters additional properties for the element item schema for Optimization Detective.
+ *
+ * @since n.e.x.t
+ *
+ * @param array<string, array{type: string}> $additional_properties Additional properties.
+ * @return array<string, array{type: string}> Additional properties.
+ */
+function image_prioritizer_add_element_item_schema_properties( array $additional_properties ): array {
+	$additional_properties['lcpElementExternalBackgroundImage'] = array(
```
felixarntz, line 105: Maybe I should have asked this before... But why are we introducing this in Image Prioritizer, when Optimization Detective already has awareness of the LCP element?

It feels like we're duplicating a bit of what Optimization Detective already handles (e.g. in the `detect.js` script the `onLCP` etc.). Why not enhance this via Optimization Detective? Or provide a more targeted extension point for the LCP element specifically that Image Prioritizer could leverage?

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
```diff
@@ -65,33 +80,128 @@ public function __invoke( OD_Tag_Visitor_Context $context ): bool {
...
 
 		if ( is_null( $background_image_url ) ) {
+			$this->maybe_preload_external_lcp_background_image( $context );
 			return false;
 		}
 
 		$xpath = $processor->get_xpath();
 
 		// If this element is the LCP (for a breakpoint group), add a preload link for it.
 		foreach ( $context->url_metric_group_collection->get_groups_by_lcp_element( $xpath ) as $group ) {
```
felixarntz, line 90: This is missing the fix we're including via #1723. I think we should merge that PR since it's more of a fix while this is an enhancement, and then update this PR to follow suit. Otherwise we're introducing the same problem for LCP background images that #1723 is fixing for the LCP generally.

This re-emphasizes to me what I mentioned in my previous point: If we were using Optimization Detective's own LCP handling, we would be able to get this kind of correct behavior "for free", rather than having to re-implement it here too. So I'm curious why we're not doing that.

At a minimum, this needs to be fixed here to be aligned with what #1723 does.

felixarntz (commented):
@westonruter I may be realizing this a bit late, but now I'm curious why this is not more closely tying into what Optimization Detective already provides.

plugins/image-prioritizer/helper.php
westonruter: Humm. Well, the normal case for how data is collected in Optimization Detective is for a tag visitor to mark tags of interest (which result in `data-od-xpath` attributes being emitted on them on the frontend), and then for OD's `detect.js` to collect data about those specific elements to then send back for storage with the URL Metric.

What is needed here is different because we don't know what the element of interest will be, since any element could be LCP with a background image applied with an external stylesheet.

So that's why this new `lcpElementExternalBackgroundImage` root property is being introduced in the URL Metric schema. There is no XPath because the element is not explicitly visited by tag visitors, so it doesn't necessarily appear in `elements`. Now, if it just so happens that there is a tag which shows up in `elements` (which was processed by a tag visitor) which is also the LCP element... it would make sense to store the LCP metric's `url` there as well. However, this would be coincidental and not something we can rely on specifically for optimizing non-inline background images in LCP elements.

In regards to `onLCP`, we could potentially expose that `onLCP` callback (in addition to `onCLS`, `onFCP`, `onINP`, and `onTTFB`) to the `initialize` function for each extension. This would eliminate the need to pass the `webVitalsLibrarySrc`.

felixarntz: I may be missing some understanding here, so apologies in advance. But don't we only care about the background image that's relevant to the LCP element?

So from that perspective, couldn't we have an extension point that allows extensions to expand what data is collected for the LCP element? And then Image Prioritizer could add the background image data as needed?

westonruter: The problem is that the tag visitors don't know what the LCP element with an external background image could possibly be, so they don't know to `return true` to flag the tag for capturing in URL Metrics. I mean, there could be a tag visitor that just returns `true` for every visited tag, which would result in `data-od-xpath` attributes being added to every single tag and then for every single tag to be stored in the `elements` of URL Metrics. But this would be wasteful.

In the case of an `IMG` being the LCP element, this is covered because the tag visitors for images are visiting every `IMG` so they are all being captured in the `elements` of a URL Metric. When such an `IMG` is the LCP element, we don't even need the `url` from the LCP Metric because the URL is right there in the `src` and `srcset`. The same goes for an element which has an inline `background-image` style: the tag visitor can see it so optimization can preload what is defined right there in the HTML. 

But for elements with external background images, this is a special case.

plugins/image-prioritizer/class-image-prioritizer-background-image-styled-tag-visitor.php
westonruter: > This is missing the fix we're including via #1723.

I'm not sure this is the case. The fix in #1723 is to allow `fetchpriority=high` to be on the `IMG` element when only the mobile and desktop viewport groups have URL Metrics collected. Preload links are still only being added for `IMG` for the actual viewport groups that have URL metrics collected for them:

https://github.com/WordPress/performance/blob/d13f33c8f561f6d61fe8e5d20c2a2192904af493/plugins/image-prioritizer/class-image-prioritizer-img-tag-visitor.php#L342-L348

Or are you saying that if the mobile and desktop viewport groups have the same LCP element, that we should go ahead and also preload the same URL for that LCP Element even for phablet and tablet when no URL Metrics are yet confirming that they also have the same LCP element?

> This re-emphasizes to me what I mentioned in my previous point: If we were using Optimization Detective's own LCP handling, we would be able to get this kind of correct behavior "for free", rather than having to re-implement it here too. So I'm curious why we're not doing that.

I'm not sure I understand. For elements which are tracked in URL Metrics (and an element appears in `elements` when there is a tag visitor which returns `true` for a tag during iterator), then as part of the stored metadata for that tracked element will be whether it `isLCP`. So here it's looking up which viewport groups have the current tag as the LCP element, and for each one a preload link is added.

plugins/optimization-detective/detect.js
```diff
@@ -424,8 +471,6 @@ export default async function detect( {
 		} );
 	}
 
-	const { onLCP } = await import( webVitalsLibrarySrc );
-
 	/** @type {LCPMetric[]} */
 	const lcpMetricCandidates = [];
```
westonruter, line 476: Now that `onLCP` is typed as `OnLCPFunction` I shouldn't think that `/** @type LCPMetric */` is needed here:

https://github.com/WordPress/performance/blob/47d06dc669b34ef2cffaef01ad8be2d11cbe06ca/plugins/optimization-detective/detect.js#L479-L483

However, when I remove it, TypeScript in PHPStorm complains:

image

It doesn't complain about the usage in Image Prioritizer's `detect.js` here, however:

image

plugins/image-prioritizer/helper.php
westonruter: I've exposed the Web Vitals functions to extensions in 47d06dc669b34ef2cffaef01ad8be2d11cbe06ca.

plugins/image-prioritizer/detect.js
adamsilverstein: thanks for researching!

plugins/image-prioritizer/tests/test-helper.php
adamsilverstein: Ah, makes sense. reading out of context it was confusing.

adamsilverstein (commented):
Looking good!

plugins/image-prioritizer/helper.php
felixarntz: https://github.com/WordPress/performance/commit/47d06dc669b34ef2cffaef01ad8be2d11cbe06ca#diff-ec3ffd3a54ce7df9d216b18a1dee27e4c41785e16698bf3229efcef5e7072fb2R64 looks good to me. It's definitely a nice little enhancement to not have to reimplement that part.

I guess I had not considered this scenario before - so is it correct that, in case the LCP element is not one of the specific elements that are being visited, it will be unknown what the LCP element is for that page load?

In that case, it makes sense that the extension has to implement it. Though I wonder whether we could make this a bit more ergonomic and centralize the implementation, e.g. to get benefits like "common LCP element" out of the box (with what #1723 improved). Doesn't have to be in this PR, but maybe we could implement the LCP element logic centrally so that every URL metric receives the LCP element, whichever it is? If we're worried about overhead of doing this when it's not needed, we could make it opt-in so that it's only recorded if at least one extension opts in to requiring it.

It is a bit specific to implement this directly in OD, but on the other hand my concern is that not doing so will have any extension that needs something LCP related to reinvent the wheel, when instead we could have a general LCP element handler and extensions could only extend what data is included as part of it. WDYT?

westonruter: > I guess I had not considered this scenario before - so is it correct that, in case the LCP element is not one of the specific elements that are being visited, it will be unknown what the LCP element is for that page load?

Yes, that's exactly right. It won't be captured which element is the LCP element since the LCP element is something other than one of the elements being tracked.

> In that case, it makes sense that the extension has to implement it. Though I wonder whether we could make this a bit more ergonomic and centralize the implementation, e.g. to get benefits like "common LCP element" out of the box (with what #1723 improved). Doesn't have to be in this PR, but maybe we could implement the LCP element logic centrally so that every URL metric receives the LCP element, whichever it is? If we're worried about overhead of doing this when it's not needed, we could make it opt-in so that it's only recorded if at least one extension opts in to requiring it.

So you're saying to implement a new root-level property in the URL Metric, like `lcpElement` (alongside the existing `elements`) which captures the properties in the [`LargestContentfulPaint`](https://developer.mozilla.org/en-US/docs/Web/API/LargestContentfulPaint) interface? In this PR that's what is essentially what is being done, although specifically for a background image. And since the `element` can't be captured as-is, this PR is capturing the `id` (also exposed on `LargestContentfulPaint`), `tag`, and `class` as a way to identify the element. Ideally we'd be able to store the server-side XPath to the element, but again it may not be a visited tag, so that `xpath` property could perhaps be a nullable field so that client-side detection can provide it if it so happens that the LCP element has a `data-od-xpath` attribute. In this way, we wouldn't technically need to store `isLCP` with each element, as we could then just look to see if the `xpath` for an element matches the XPath stored with the LCP data.

And if we're capturing these properties of `LargestContentfulPaint`, should we also then capture the other properties like `size`, `renderTime`, and `loadTime`?

And then if we're capturing the LCP metrics, should we then also centralize the capturing of the CLS and INP data as well? 

Or maybe we reconsider for later since this could go to far. For storing the LCP `url`, there's also the question of data validation. We'd want to make sure that we only preload safe URLs which is being addressed in #1713. We can always update Image Prioritizer later to reuse LCP metric data stored by Optimization Detective centrally, especially when/if it comes to light that more use cases need access to this centralized data.

So how about going with an Image Prioritizer-specific implementation but then we open a new issue to discuss capturing these metrics centrally in a future version. This would tie in nicely with the Performance dashboard (#1324) since the data would already be there (although the dashboard would probably want to introduce additional data retention to capture more than the sample size).

felixarntz: > So you're saying to implement a new root-level property in the URL Metric, like `lcpElement` (alongside the existing `elements`) which captures the properties in the [`LargestContentfulPaint`](https://developer.mozilla.org/en-US/docs/Web/API/LargestContentfulPaint) interface?

Yes, exactly.

> And if we're capturing these properties of `LargestContentfulPaint`, should we also then capture the other properties like `size`, `renderTime`, and `loadTime`?

Maybe. I think we should probably keep it light in the default implementation, but allow extensions to expand what is detected. For example there should be an extension point where the LCP data is reported in JS so that a plugin could add more data to the `lcpElement` object that is sent to the REST API, and on the server-side it would need to be possible to extend the schema for what can be in `lcpElement`.

I also would prefer not to include actual performance numbers, because OD is primarily an optimization tool, not a monitoring tool. A separate extension for monitoring could add such data as needed.

> And then if we're capturing the LCP metrics, should we then also centralize the capturing of the CLS and INP data as well?

Potentially, but only if we see a need for it. While it seems to make sense from a metric perspective, I'd argue LCP is different from CLS and INP because there is no "CLS element" or "INP element". The LCP element is directly relevant to things that Optimization Detective allows to do. This also relates to what I mention above about focusing on areas that help optimization, not performance metrics. I find it less applicable for INP and CLS, because what would we send there to help optimization? Possibly there is some useful data (e.g. long tasks), but I feel that this should be separately explored, if at all.

> So how about going with an Image Prioritizer-specific implementation but then we open a new issue to discuss capturing these metrics centrally in a future version. This would tie in nicely with the Performance dashboard (#1324) since the data would already be there (although the dashboard would probably want to introduce additional data retention to capture more than the sample size).

That sounds good to me.

westonruter: > there is no "CLS element"

Well, there are elements that cause CLS. So in theory a top-level CLS property could capture the [`LayoutShiftAttribution`](https://developer.mozilla.org/en-US/docs/Web/API/LayoutShiftAttribution) [sources](https://developer.mozilla.org/en-US/docs/Web/API/LayoutShift/sources). Also, for INP  That said, Embed Optimizer isn't using this and is instead using `ResizeObserver`. Things to explore in the new issue. I've just filed it as #1730.

felixarntz (approved):

PR #1700, opened 2024-11-25
Set development mode to 'plugin' in the dev environment and allow pages to be optimized when admin is logged-in (when in plugin dev mode)

### LINKED ISSUE #1423: Development mode should be set to 'plugin' in the development environment
WordPress has a [development mode](https://make.wordpress.org/core/2023/07/14/configuring-development-mode-in-6-3/) introduced in WP 6.3 which can be set to `core`, `theme`, `plugin`, or `all`. Wouldn't it make sense if this were set to `plugin` in the development environment? Our plugins could behave differently when in plugin development mode. For example, Optimization Detective could by default do the equivalent of the following, a configuration that makes it easier to test during development:

```php
// By default Optimization Detective doesn't optimize responses for admin users.

// By default new URL metrics are only accepted for a given IP once every 60 seconds.
add_filter( 'od_url_metric_storage_lock_ttl', '__return_zero' );

// By default a URL metric is considered fresh for 1 day.
// Reducing this ensures new URL metrics will be collected.
add_filter( 'od_url_metric_freshness_ttl', '__return_zero' );

// By default 3 URL Metrics are collected for a given breakpoint viewport group.
// Reducing this to 1 means the groups can be considered complete with fewer page loads.
add_filter( 'od_url_metrics_breakpoint_sample_size', function () { return 1; } );
```

For example, the plugin could be patched as follows:

```diff
diff --git a/plugins/optimization-detective/optimization.php b/plugins/optimization-detective/optimization.php
index 442a7a62..eb7a9c3e 100644
--- a/plugins/optimization-detective/optimization.php
+++ b/plugins/optimization-detective/optimization.php
@@ -92,7 +92,7 @@ function od_can_optimize_response(): bool {
 		// interfere with the XPath indices. Note that od_get_normalized_query_vars() is varied by is_user_logged_in()
 		// so membership sites and e-commerce sites will still be able to be optimized for their normal visitors.
 		current_user_can( 'customize' )
-	);
+	) || wp_is_development_mode( 'plugin' );
 
 	/**
 	 * Filters whether the current response can be optimized.
diff --git a/plugins/optimization-detective/storage/class-od-storage-lock.php b/plugins/optimization-detective/storage/class-od-storage-lock.php
index 619a041e..40162a7e 100644
--- a/plugins/optimization-detective/storage/class-od-storage-lock.php
+++ b/plugins/optimization-detective/storage/class-od-storage-lock.php
@@ -43,7 +43,7 @@ final class OD_Storage_Lock {
 		 *
 		 * @param int $ttl TTL.
 		 */
-		$ttl = (int) apply_filters( 'od_url_metric_storage_lock_ttl', MINUTE_IN_SECONDS );
+		$ttl = (int) apply_filters( 'od_url_metric_storage_lock_ttl', wp_is_development_mode( 'plugin' ) ? 0 : MINUTE_IN_SECONDS );
 		return max( 0, $ttl );
 	}
 
diff --git a/plugins/optimization-detective/storage/data.php b/plugins/optimization-detective/storage/data.php
index c3395f96..4bf68b3f 100644
--- a/plugins/optimization-detective/storage/data.php
+++ b/plugins/optimization-detective/storage/data.php
@@ -31,7 +31,7 @@ function od_get_url_metric_freshness_ttl(): int {
 	 *
 	 * @param int $ttl Expiration TTL in seconds. Defaults to 1 day.
 	 */
-	$freshness_ttl = (int) apply_filters( 'od_url_metric_freshness_ttl', DAY_IN_SECONDS );
+	$freshness_ttl = (int) apply_filters( 'od_url_metric_freshness_ttl', wp_is_development_mode( 'plugin' ) ? 0 : DAY_IN_SECONDS );
 
 	if ( $freshness_ttl < 0 ) {
 		_doing_it_wrong(
```

ShyamGadde:
## Summary

Fixes #1423 

- Sets `WP_DEVELOPMENT_MODE` constant to `plugin` in development environment.
- Adds plugin development mode configuration for Optimization Detective.

### DIFF
--- a/plugins/optimization-detective/readme.txt
+++ b/plugins/optimization-detective/readme.txt
@@ -148,7 +148,7 @@ add_filter( 'od_url_metric_freshness_ttl', static function (): int {
 
 Filters the minimum allowed viewport aspect ratio for URL Metrics.
 
-The 0.4 value is intended to accommodate the phone with the greatest known aspect ratio at 21:9 when rotated 90 degrees to 9:21 (0.429). During development you may want to set this to 0 to account Dev Tools likely being open and docked to the right:
+The 0.4 value is intended to accommodate the phone with the greatest known aspect ratio at 21:9 when rotated 90 degrees to 9:21 (0.429). During development when you have the DevTools console open on the right, the viewport aspect ratio will be smaller than normal. In this case, you may want to set this to 0:
 
 `
 <?php
@@ -163,7 +163,7 @@ Filters the maximum allowed viewport aspect ratio for URL Metrics.
 
 The 2.5 value is intended to accommodate the phone with the greatest known aspect ratio at 21:9 (2.333).
 
-During development when you have the DevTools console open on the bottom, for example, the viewport aspect ratio larger than normal. In this case, you may want to increase the maximum aspect ratio:
+During development when you have the DevTools console open on the bottom, for example, the viewport aspect ratio will be larger than normal. In this case, you may want to increase the maximum aspect ratio:
 
 `
 <?php
--- a/plugins/optimization-detective/readme.txt
+++ b/plugins/optimization-detective/readme.txt
@@ -135,7 +135,7 @@ add_filter( 'od_url_metric_freshness_ttl', static function (): int {
 } );
 `
 
-During development, this can be useful to set to zero:
+During development, this can be useful to set to zero so that you don't have to wait for new URL Metrics to be requested when engineering a new optimization:
 
 `
 <?php
--- a/plugins/optimization-detective/readme.txt
+++ b/plugins/optimization-detective/readme.txt
@@ -104,12 +104,12 @@ add_filter( 'od_can_optimize_response', '__return_true' );
 
 **Filter:** `od_url_metrics_breakpoint_sample_size` (default: 3)
 
-Filters the sample size for a breakpoint's URL Metrics on a given URL. The sample size must be greater than zero. In plugin development mode, the default is 1. You can increase the sample size if you want better guarantees that the applied optimizations will be accurate:
+Filters the sample size for a breakpoint's URL Metrics on a given URL. The sample size must be greater than zero. You can increase the sample size if you want better guarantees that the applied optimizations will be accurate. During development, it may be helpful to reduce the sample size to 1:
 
 `
 <?php
 add_filter( 'od_url_metrics_breakpoint_sample_size', function (): int {
-	return 10;
+	return 1;
 } );
 `
 
@@ -126,7 +126,7 @@ add_filter( 'od_metrics_storage_lock_ttl', function ( int $ttl ): int {
 
 **Filter:** `od_url_metric_freshness_ttl` (default: 1 day in seconds)
 
-Filters the freshness age (TTL) for a given URL Metric. The freshness TTL must be at least zero, in which it considers URL Metrics to always be stale. In practice, the value should be at least an hour. In plugin development mode, this is set to zero by default. If your site content does not change frequently, you may want to increase the TTL to a week:
+Filters the freshness age (TTL) for a given URL Metric. The freshness TTL must be at least zero, in which it considers URL Metrics to always be stale. In practice, the value should be at least an hour. If your site content does not change frequently, you may want to increase the TTL to a week:
 
 `
 <?php
@@ -135,17 +135,33 @@ add_filter( 'od_url_metric_freshness_ttl', static function (): int {
 } );
 `
 
+During development, this can be useful to set to zero:
+
+`
+<?php
+add_filter( 'od_url_metric_freshness_ttl', static function (): int {
+    return 0;
+} );
+`
+
 **Filter:** `od_minimum_viewport_aspect_ratio` (default: 0.4)
 
 Filters the minimum allowed viewport aspect ratio for URL Metrics.
 
-The 0.4 value is intended to accommodate the phone with the greatest known aspect ratio at 21:9 when rotated 90 degrees to 9:21 (0.429). When in plugin development mode, the default value is 0; this is to account for Dev Tools likely being open.
+The 0.4 value is intended to accommodate the phone with the greatest known aspect ratio at 21:9 when rotated 90 degrees to 9:21 (0.429). During development you may want to set this to 0 to account Dev Tools likely being open and docked to the right:
+
+`
+<?php
+add_filter( 'od_minimum_viewport_aspect_ratio', static function (): int {
+    return 0;
+} );
+`
 
 **Filter:** `od_maximum_viewport_aspect_ratio` (default: 2.5)
 
 Filters the maximum allowed viewport aspect ratio for URL Metrics.
 
-The 2.5 value is intended to accommodate the phone with the greatest known aspect ratio at 21:9 (2.333). When in plugin development mode, the default value is PHP_INT_MAX; this is to account for Dev Tools likely being open.
+The 2.5 value is intended to accommodate the phone with the greatest known aspect ratio at 21:9 (2.333).
 
 During development when you have the DevTools console open on the bottom, for example, the viewport aspect ratio larger than normal. In this case, you may want to increase the maximum aspect ratio:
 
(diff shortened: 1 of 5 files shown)


felixarntz (changes requested):
+1 to set the development mode to "plugin" in our config. For the Optimization Detective specific tweaks, I am not sure it's a good idea to rely on it in all of these places.

The modified condition for `od_can_optimize_response()` makes sense to me as that makes development a lot easier, and it's not very risky as it's a general on/off switch for almost the entire feature. But adjusting the TTLs and sample sizes I find questionable. These kinds of changes will mean that realistically we'll almost never naturally test this in the way that it actually behave, but only how it behaves with the development mode set to "plugin". Why are those tweaks necessary? And if they are, what is our strategy to mitigate that risk?

westonruter:
> But adjusting the TTLs and sample sizes I find questionable. These kinds of changes will mean that realistically we'll almost never naturally test this in the way that it actually behave, but only how it behaves with the development mode set to "plugin". Why are those tweaks necessary? And if they are, what is our strategy to mitigate that risk?

During development I have a plugin active which does this:

```php
add_filter( 'od_url_metric_storage_lock_ttl', '__return_zero' );
add_filter( 'od_url_metric_freshness_ttl', '__return_zero' );
add_filter( 'od_url_metrics_breakpoint_sample_size', function () { return 1; } );
add_filter( 'od_maximum_viewport_aspect_ratio', function () { return PHP_INT_MAX; } );
add_filter( 'od_minimum_viewport_aspect_ratio', function () { return 0; } );
```

With the changes in this PR, such a plugin won't be necessary anymore to start development. These tweaks do a few key things to facilitate development:

1. `od_url_metric_storage_lock_ttl`: Allows a user to submit a new URL Metric to be stored with every page load.
2. `od_url_metric_freshness_ttl`: Makes it so that URL Metrics are always considered stale so that new ones are accepted.
3. `od_url_metrics_breakpoint_sample_size`: Reduces the number of URL Metrics from the sample size of 3 to be considered "complete". Some optimizations only start applying once the gathered URL Metrics all concur that a given element is LCP, so by reducing this from 3 to 1 then fewer reloads are needed to flush out stale URL Metrics to start seeing the expected optimizations apply from the most recent page load.
4. `od_can_optimize_response`: Already discussed above.
5. `od_maximum_viewport_aspect_ratio`/`od_minimum_viewport_aspect_ratio`: Disabling these ensures that metrics will be collected when, for example, Dev Tools is open and the window aspect ratio is not normal.

felixarntz:
@westonruter This only answers my first question, not the primary concern I'm raising. Makes sense that you may use a utility plugin like this for development, and I think for this kind of tweak that's actually better. This means at least that not everyone contributing to OD will get that experience by default, so we'll get some usage of the way this plugin behaves in production.

I don't think the benefit of not needing such a helper plugin outweigh the risks of not running the code in the way it should usually behave out of the box.

westonruter:
@felixarntz Well, then I guess what's the point of plugin development mode if enabling it doesn't get you what you need to start developing? So there's actually two levels of development mode essentially? A shallow development mode and a deep development mode?

felixarntz:
@westonruter 
> Well, then I guess what's the point of plugin development mode if enabling it doesn't get you what you need to start developing? So there's actually two levels of development mode essentially? A shallow development mode and a deep development mode?

I think it's more nuanced, yes. Any of the changes currently in this PR could be implemented as dependent on the development mode config, but tweaking production behavior for development comes with the aforementioned risk, so I think this should be used very selectively.

Examples where I think development mode makes sense:
* Enabling logic to run in a condition where it would otherwise almost never run in development (e.g. the `od_can_optimize_response()` example here, or the Site Kit plugin not printing any Analytics tags when an admin is logged in)
* Skipping various caches to avoid having to always clear caches when debugging

Both of these are more or less on/off switches, but they don't tweak any more granular logic outcomes. There is definitely a fine line between the developer convenience and the risk of running a too altered version of the plugin in development mode, so it makes sense to discuss specific cases. But per the above, here I think tweaking numbers like TTL and sample sizes may go a bit too far.

westonruter:
@felixarntz OK, makes sense. Undone in 6da3ddc.

plugins/optimization-detective/readme.txt
```diff
@@ -125,30 +126,48 @@ add_filter( 'od_metrics_storage_lock_ttl', function ( int $ttl ): int {
...
 
 The 2.5 value is intended to accommodate the phone with the greatest known aspect ratio at 21:9 (2.333).
 
-During development when you have the DevTools console open, for example, the viewport aspect ratio will be wider than normal. In this case, you may want to increase the maximum aspect ratio:
+During development when you have the DevTools console open on the bottom, for example, the viewport aspect ratio larger than normal. In this case, you may want to increase the maximum aspect ratio:
 
 `
 <?php
-add_filter( 'od_maximum_viewport_aspect_ratio', function () {
+add_filter( 'od_maximum_viewport_aspect_ratio', function (): int {
```
mukeshpanchal27:
```suggestion
add_filter( 'od_maximum_viewport_aspect_ratio', static function (): int {
```

plugins/optimization-detective/readme.txt
```diff
@@ -125,30 +126,48 @@ add_filter( 'od_metrics_storage_lock_ttl', function ( int $ttl ): int {
...
 
 `
 <?php
-add_filter( 'od_url_metric_freshness_ttl', '__return_zero' );
+add_filter( 'od_url_metric_freshness_ttl', static function (): int {
+    return WEEK_IN_SECONDS;
+} );
+`
+
+During development, this can be useful to set to zero:
```
felixarntz: Maybe add a brief sentence explanation on why that can be helpful?

plugins/optimization-detective/readme.txt
```diff
@@ -125,30 +126,48 @@ add_filter( 'od_metrics_storage_lock_ttl', function ( int $ttl ): int {
...
+    return 0;
+} );
 `
 
 **Filter:** `od_minimum_viewport_aspect_ratio` (default: 0.4)
 
 Filters the minimum allowed viewport aspect ratio for URL Metrics.
 
-The 0.4 value is intended to accommodate the phone with the greatest known aspect ratio at 21:9 when rotated 90 degrees to 9:21 (0.429).
+The 0.4 value is intended to accommodate the phone with the greatest known aspect ratio at 21:9 when rotated 90 degrees to 9:21 (0.429). During development you may want to set this to 0 to account Dev Tools likely being open and docked to the right:
```
felixarntz: Better use the phrasing like below, since we shouldn't assume people "likely" have it docked to the right:
```suggestion
The 0.4 value is intended to accommodate the phone with the greatest known aspect ratio at 21:9 when rotated 90 degrees to 9:21 (0.429). During development when you have the DevTools console open on the right, the viewport aspect ratio will be smaller than normal. In this case, you may want to set this to 0:
```

plugins/optimization-detective/readme.txt
```diff
@@ -125,30 +126,48 @@ add_filter( 'od_metrics_storage_lock_ttl', function ( int $ttl ): int {
...
+`
 
 **Filter:** `od_maximum_viewport_aspect_ratio` (default: 2.5)
 
 Filters the maximum allowed viewport aspect ratio for URL Metrics.
 
 The 2.5 value is intended to accommodate the phone with the greatest known aspect ratio at 21:9 (2.333).
 
-During development when you have the DevTools console open, for example, the viewport aspect ratio will be wider than normal. In this case, you may want to increase the maximum aspect ratio:
+During development when you have the DevTools console open on the bottom, for example, the viewport aspect ratio larger than normal. In this case, you may want to increase the maximum aspect ratio:
```
felixarntz: Missing part of the sentence:
```suggestion
During development when you have the DevTools console open on the bottom, for example, the viewport aspect ratio will be larger than normal. In this case, you may want to increase the maximum aspect ratio:
```

felixarntz (approved):
@ShyamGadde @westonruter Thanks for the updates, looks great!

Just a few documentation notes that I think should be revised.

plugins/optimization-detective/readme.txt
westonruter:
```suggestion
During development, this can be useful to set to zero so that you don't have to wait for new URL Metrics to be requested when engineering a new optimization:
```

